PR #1645, opened 2020-06-17
Support for tagging user lists

samuli:
This PR adds basic support for assigning tags to user lists. User list tags are saved in the same manner as when assigning tags to resources, i.e. by updating entries in tables `tags` and `resource_tags`. User list tags are separated from resource tags by allowing `resource_id` field in `resource_tags` to be NULL.

Current functionality should not be affected. User list tagging is disabled by default and can be enabled via a new setting. When enabled, tags can be assigned to the list in the edit-mode and the assigned tags are displayed on the mylist-page. Enabling user list tags does not affect Tag-search and Tag-browsing, they are restricted to cover only resource tags.  

Our use case is to allow retrieving user lists by a given tag, for example by adding a `listSearch`method to `tags` table, similar to `resourceSearch`. This could be wired to the `Tag` controller, maybe by passing in a parameter that restricts the search to user lists. I think we can build this to our own code base with the help of this PR. Another option would be to extend VuFinds' current tag-search to cover user list tags as well.

What are your ideas about this?

TODO
- [x] Open JIRA ticket with reminder to write Mink tests to cover new functionality ([VUFIND-1416](https://vufind.org/jira/browse/VUFIND-1416))
- [x] Add PostgreSQL migration
- [x] Test with PostgreSQL
- [x] When merging, update changelog with minor BC breaks (e.g. UserList row constructor signature change)
- [x] When merging, update database changelog and improve documentation of resource_tags table

PR #1697, opened 2020-09-01
Add mink test for list tags

demiankatz:
- Also includes minor style improvement to tag display
- Also fixes order of existing assertions in FavoritesTest
- Resolves [VUFIND-1416](https://vufind.org/jira/browse/VUFIND-1416).


demiankatz:
@samuli, as discussed on #1645, I've added a Mink test for list tags here. This is pretty minimalistic, and I could certainly make it more thorough/robust if I had more time, but this at least gets a baseline in place.

I also added a little bit of styling to the list template: `<strong>` tags around the "Tags:" label, and a `list-tags` class on the actual list of list tags, which makes it easier to extract them for testing purposes, and also provides a potential hook for custom CSS.

I'm going ahead and merging this since the changes are minimal, but I wanted to make you aware of them. Thanks again for all of your work on this feature!

PR #1683, opened 2020-07-27
FOLIO-discoverySuppress

mis306lu:
DiscoverySuppress holding and item records should be excluded.
thank you!

PR #1698, opened 2020-09-01
FOLIO: update getholdings cql query to accomodate optional fields

ihardy:


ihardy:
Hi  @mis306lu would you mind reviewing this one? I had trouble with the suppression changes in #1683 . No items/holdings were being retreived for me. I think that's because on both holdings and items `discoverySupress` is an optional field and none of my records have discoverySupress==false. I tried updating the cql query to look for items/holdings where NOT discoverySupress==true instead and that seems to work. Let me know if that also works for you.

I also removed my first attempt to get around this by just skipping items where discoverySupress==true since that's redundant now. I think your way is probably smarter.

demiankatz:
Thanks, @ihardy; this makes sense to me. I'll be interested to hear what @mis306lu has to say. Once this is approved, I'll backport it to the release-7.0 branch so it can be included in a 7.0.2 release should we decide to issue one before 7.1 comes out.

mis306lu:
Makes sense to me! thank you!

demiankatz:
Thanks, @mis306lu; I have manually merged this to release-7.0 (and from there to the dev branch)!

mis306lu:
@demiankatz - would you believe I just bumped into this issue yesterday indexing 'FOLIO-born' items in VuFind!  Previously I was indexing items we imported that included the field.  Thanks for the fix @ihardy!

demiankatz:
Good to know, @mis306lu! That's the best part of open source development -- when somebody else fixes your problem before you encounter it. ;-)

PR #1699, opened 2020-09-01
Upgrade to League/CommonMark 1.5

demiankatz:
There is a new minor version of the CommonMark Markdown processor, which includes at least one deprecation affecting VuFind's example configuration. This PR applies the updates and provides a space to review/discuss.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
         "laminas/laminas-validator": "2.13.4",
         "laminas/laminas-view": "2.11.4",
         "laminas-commons/lmc-rbac-mvc": "3.0.1",
-        "league/commonmark": "1.4.3",
+        "league/commonmark": "1.5.4",
         "matthiasmullie/minify": "1.3.63",
         "misd/linkify": "1.1.4",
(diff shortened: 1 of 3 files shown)


demiankatz:
@xmorave2, when doing routine dependency updates, I noticed that a new minor version of League/CommonMark was available, but it has some deprecations. I created this PR so you could review the adjustments and let me know if you have any concerns before I merge the updates.

xmorave2:
@demiankatz Looks good to me. Although there are three new extensions (and they look interesting and useful) we could merge this as is and add reflect new extensions on another PR (I could do it...).

xmorave2 (approved):

demiankatz:
Thanks, @xmorave2, I have added a note about the setting deprecation to the [changelog](https://vufind.org/wiki/changelog); I don't think it's likely that anyone has customized this, so I'm not going to devote the effort to building it into VuFind's auto-upgrader. I can revisit that choice if anyone disagrees.

If you'd like to open a PR with support for the new plug-ins at your convenience, that would be great. Thanks!

PR #1700, opened 2020-09-02
KohaRest: Add a comment about OAuth2 authorization header.

EreMaijala:
Also makes the comment block a bit more readable.


demiankatz:
Thank you!

PR #1701, opened 2020-09-02
Delete email authentication token only after initial checks.

EreMaijala:
This prevents the token from being destroyed by e.g. safe link checks by the email server.

This has been affecting e.g. users that have a hotmail address since hotmail checks the links for malicious content.


demiankatz:
Thank you; good catch! I have backported this to release-7.0 so it can be included in release 7.0.2.

PR #1702, opened 2020-09-02
KohaRest: Fix canceling holds.

EreMaijala:
- Allow 204 as valid result code.
- Use item_id as the hold identifier when canceling holds. This allows the template to display failed cancels properly.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
@@ -613,7 +613,7 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
             }
             $holds[] = [
                 'id' => $entry['biblio_id'],
-                'item_id' => $entry['item_id'] ?? null,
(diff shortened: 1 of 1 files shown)


demiankatz:
@EreMaijala, do you want/need anyone to test/review this before it is merged? Assuming you're confident it is ready to go, would you like me to cherry-pick it onto the release-7.0 branch so it can be included in a 7.0.2 release? Given the high interest in this driver, it seems like it might be good to get a bug-fixed version into a release sooner rather than later.

EreMaijala:
@demiankatz I've tested this the best I can, so it should be good to go. And cherry-picking this and other KohaRest fixes to 7.0.x would be great.

I'm sorry to keep these coming, but obviously I didn't quite nail all the details initially, and now the issues pop up when then integration is being tested by libraries with real data.

demiankatz:
@EreMaijala, sounds good, and no worries -- these integrations take some time to get completely right!

demiankatz:
Cherry-picked to release-7.0 and merged to dev from there; closing now!

PR #1708, opened 2020-09-04
FOLIO: Modify notes field and add item_notes, issues, supplements and indexes

bbusenius:
**Changes in this request**:
- Updates the `noteFormatter` to only return notes that are public.
- Eliminates the deprecated `notes` field that was returning item notes and replaces it with the more descriptive `holdings_notes` field.
- Adds return values for `item_notes`, `issues`, `supplements` and `indexes`.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -482,7 +482,8 @@ class Folio extends AbstractAPI implements
             $itemResponse = $this->makeRequest('GET', '/item-storage/items', $query);
             $itemBody = json_decode($itemResponse->getBody());
             $notesFormatter = function ($note) {
-                return !$note->staffOnly && $note->note ? $note->note : '';
+                return !($note->staffOnly ?? false)
+                    && !empty($note->note) ? $note->note : '';
             };
             $textFormatter = function ($supplement) {
                 $format = '%s %s';
@@ -491,7 +492,9 @@ class Folio extends AbstractAPI implements
                 $statement = trim(sprintf($format, $supStat, $supNote));
                 return $statement ?? '';
             };
-            $holdingNotes = array_map($notesFormatter, $holding->notes ?? []);
+            $holdingNotes = array_filter(
+                array_map($notesFormatter, $holding->notes ?? [])
+            );
             $hasHoldingNotes = !empty(implode($holdingNotes));
             $holdingsStatements = array_map(
                 $textFormatter,
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -482,9 +482,25 @@ class Folio extends AbstractAPI implements
             $itemResponse = $this->makeRequest('GET', '/item-storage/items', $query);
             $itemBody = json_decode($itemResponse->getBody());
             $notesFormatter = function ($note) {
-                return $note->note ?? '';
+                return !$note->staffOnly && $note->note ? $note->note : '';
             };
+            $textFormatter = function ($supplement) {
+                $format = '%s %s';
+                $supStat = $supplement->statement;
+                $supNote = $supplement->note;
+                $statement = sprintf($format, $supStat, $supNote);
+                return $statement ?? '';
+            };
+            $holdingNotes = array_map($notesFormatter, $holding->notes ?? []);
+            $hasHoldingNotes = !empty(implode($holdingNotes));
+            $holdingsStatements = $holding->holdingsStatements;
+            $holdingsSupplements = $holding->holdingsStatementsForSupplements;
+            $holdingsIndexes = $holding->holdingsStatementsForIndexes;
+            $issues = array_map($textFormatter, $holdingsStatements ?? []);
+            $supplements = array_map($textFormatter, $holdingsSupplements ?? []);
+            $indexes = array_map($textFormatter, $holdingsIndexes ?? []);
             foreach ($itemBody->items as $item) {
+                $itemNotes = array_map($notesFormatter, $item->notes ?? []);
                 $items[] = [
                     'id' => $bibId,
                     'item_id' => $item->id,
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -482,9 +482,25 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
+                $supNote = $supplement->note;
+                $statement = sprintf($format, $supStat, $supNote);
+                return $statement ?? '';
+            };
+            $holdingNotes = array_map($notesFormatter, $holding->notes ?? []);
+            $hasHoldingNotes = !empty(implode($holdingNotes));
+            $holdingsStatements = $holding->holdingsStatements;
+            $holdingsSupplements = $holding->holdingsStatementsForSupplements;
+            $holdingsIndexes = $holding->holdingsStatementsForIndexes;
+            $issues = array_map($textFormatter, $holdingsStatements ?? []);
```
demiankatz: I think it would be more clear to put the `?? []` fallback up above when initially defining `$holdingsStatements`, etc., rather than down here when you use the variables for the first time. I think there's a greater chance of PHP notices doing it this way.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -482,9 +482,25 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
             $itemBody = json_decode($itemResponse->getBody());
             $notesFormatter = function ($note) {
-                return $note->note ?? '';
+                return !$note->staffOnly && $note->note ? $note->note : '';
             };
+            $textFormatter = function ($supplement) {
+                $format = '%s %s';
+                $supStat = $supplement->statement;
+                $supNote = $supplement->note;
+                $statement = sprintf($format, $supStat, $supNote);
```
demiankatz: Is there always going to be a statement and a note? If not, it might be worth wrapping a `trim` around this to kill any superfluous whitespace.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -482,9 +482,25 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
+            $holdingNotes = array_map($notesFormatter, $holding->notes ?? []);
+            $hasHoldingNotes = !empty(implode($holdingNotes));
+            $holdingsStatements = $holding->holdingsStatements;
+            $holdingsSupplements = $holding->holdingsStatementsForSupplements;
+            $holdingsIndexes = $holding->holdingsStatementsForIndexes;
+            $issues = array_map($textFormatter, $holdingsStatements ?? []);
+            $supplements = array_map($textFormatter, $holdingsSupplements ?? []);
+            $indexes = array_map($textFormatter, $holdingsIndexes ?? []);
             foreach ($itemBody->items as $item) {
+                $itemNotes = array_map($notesFormatter, $item->notes ?? []);
```
demiankatz: Is it worth wrapping an `array_filter` around this to strip out empty entries resulting from staff only notes?

demiankatz (changes requested):
Thanks for this -- looks like a much richer treatment of notes. I've just left a few small suggestions below. Once those are addressed, it might be good to see if @ihardy and/or @mis306lu have any further thoughts.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
bbusenius: Yes, it's more concise too. This is done.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
bbusenius: Great suggestion!

module/VuFind/src/VuFind/ILS/Driver/Folio.php
bbusenius: I don't see why not.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -482,9 +482,33 @@ public function getHolding($bibId, array $patron = null, array $options = [])
             $itemResponse = $this->makeRequest('GET', '/item-storage/items', $query);
             $itemBody = json_decode($itemResponse->getBody());
             $notesFormatter = function ($note) {
-                return $note->note ?? '';
+                return !$note->staffOnly && $note->note ? $note->note : '';
```
demiankatz: Are `staffOnly` and `note` guaranteed to be set, or would it be safer to do something like this instead?

`return !($note->staffOnly ?? false) && !empty($note->note) ? $note->note : '';`

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -482,9 +482,33 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
             };
+            $textFormatter = function ($supplement) {
+                $format = '%s %s';
+                $supStat = $supplement->statement;
+                $supNote = $supplement->note;
+                $statement = trim(sprintf($format, $supStat, $supNote));
+                return $statement ?? '';
+            };
+            $holdingNotes = array_map($notesFormatter, $holding->notes ?? []);
```
demiankatz: I see you added an array_filter to the item notes filtering below; would it make sense to add one here as well for the holding notes?

demiankatz:
Thanks, @bbusenius, this is looking great. I just had a couple small questions after a second review. Regardless of those, I think it would be fine for @ihardy and @mis306lu to take a look at this now, if they're able.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
bbusenius: Done

module/VuFind/src/VuFind/ILS/Driver/Folio.php
bbusenius: Yes, this is better. Done.

demiankatz (approved):

demiankatz:
Thanks! Looks good to me. I'll ping the Slack channel to see if anyone else has time for a double-check before I merge.

ihardy:
Thanks @bbusenius and @demiankatz sorry about the slow response. This looks good to me. Tried it out on one of our test instances and I see the new notes data coming through etc.

demiankatz:
Thanks, @ihardy! Merging now.

bbusenius:
Thanks @ihardy and @demiankatz!

PR #1710, opened 2020-09-08
KohaRest: Update patron validation request path.

EreMaijala:
Old path is deprecated, and new path is included in REST plugin versions considered more stable and secure, so there's no need for back-compatibility.


demiankatz:
Thanks!

PR #1712, opened 2020-09-08
Fix replaceTerm() word boundary bug.

demiankatz:
TODO:
- [x] Figure out how to correctly deal with Boolean operators in query
- [x] Determine whether the fallback clause is needed at all; eliminate if not, add useful test case if so
- [x] Run full test suite
- [x] Resolve [VUFIND-1423](https://vufind.org/jira/browse/VUFIND-1423) when merging.

### DIFF
--- a/module/VuFindSearch/src/VuFindSearch/Query/Query.php
+++ b/module/VuFindSearch/src/VuFindSearch/Query/Query.php
@@ -227,9 +227,8 @@ class Query extends AbstractQuery
         // try again with a less restricted regular expression.
         // TODO: identify a test case where this fallback is actually needed; it
         // may not actually serve a useful purpose.
-        $before = $this->queryString;
         $this->queryString = preg_replace("/\b$from\b/i", $to, $queryString);
-        if ($before === $this->queryString) {
+        if ($queryString === $this->queryString) {
             $this->queryString = preg_replace("/$from/i", $to, $queryString);
         }
     }
--- a/module/VuFindSearch/src/VuFindSearch/Query/Query.php
+++ b/module/VuFindSearch/src/VuFindSearch/Query/Query.php
@@ -93,7 +93,15 @@ class Query extends AbstractQuery
      */
     protected function normalizeText($text)
     {
-        return strtolower($this->stripDiacritics($text));
+        // The input to normalizeText may be a Solr query with Boolean operators
+        // in it; we want to be careful not to turn this into something invalid.
+        $stripped = $this->stripDiacritics($text);
+        $booleans = ['AND', 'OR', 'NOT'];
+        $words = [];
+        foreach (preg_split('/\s+/', $stripped) as $word) {
+            $words[] = in_array($word, $booleans) ? $word : strtolower($word);
+        }
+        return implode(' ', $words);
     }
 
     /**
--- a/module/VuFindSearch/src/VuFindSearch/Query/Query.php
+++ b/module/VuFindSearch/src/VuFindSearch/Query/Query.php
@@ -223,18 +223,15 @@ class Query extends AbstractQuery
         $queryString = $normalize
             ? $this->getNormalizedString() : $this->queryString;
 
-        // If our "from" pattern contains non-word characters, we can't use word
-        // boundaries for matching.  We want to try to use word boundaries when
-        // possible, however, to avoid the replacement from affecting unexpected
-        // parts of the search query.
-        if (!preg_match('/.*[^\w].*/', $from)) {
-            $pattern = "/\b$from\b/i";
-        } else {
-            $pattern = "/$from/i";
+        // Try to match within word boundaries; if that fails to change anything,
+        // try again with a less restricted regular expression.
+        // TODO: identify a test case where this fallback is actually needed; it
+        // may not actually serve a useful purpose.
+        $before = $this->queryString;
+        $this->queryString = preg_replace("/\b$from\b/i", $to, $queryString);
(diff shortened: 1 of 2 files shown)


demiankatz:
@EreMaijala, this bug has quite possibly been lurking here since the early days of VuFind 1, and I don't think even I know all the history here. Please see my comments on VUFIND-1423; I'd value your input if you have a moment to take a look at this.

demiankatz:
...and the plot thickens, because I corrected a small typo in my test case and was surprised when this caused the test to fail; it appears that the query normalization process that replaceTerm() performs has the potential to break Boolean operators. I'm not sure if this affects real-life scenarios or is just a problem for my unit test, but I strongly suspect that it's a real-world problem. I'll investigate further when time permits.

demiankatz:
...and one more update: I've made the normalizeText() method smarter, so it won't convert uppercase Boolean operators to lowercase when it encounters them. This may be an overly specific solution to the problem, but it seems workable at least for the moment. So I now think we're at a point of having all tests passing again, at least. I remain uncertain whether there's actually a valid scenario where we would want to ignore word boundaries.

EreMaijala (commented):
Your changes look good. However, I don't really understand the stripDiacritics thing. We've actually disabled normalization for our spelling suggestions some five years ago, and while I can't remember all the details, I believe the stripping of diacritics was the worst part of it. Actually, I'm not sure I grasp the need to lowercase the terms either. If we case-insensitively replace the terms, why bother?

What I would understand would be to normalize the diacritics etc. to one of the UNICODE normalization forms (we use NFKC everywhere) the data is indexed in. See e.g. https://github.com/NatLibFi/NDL-VuFind2/blob/master/module/FinnaSearch/src/FinnaSearch/Backend/Solr/LuceneSyntaxHelper.php#L95 for what we do for search queries.

demiankatz:
@EreMaijala, thanks for the review! Can you think of any reason we would ever need to omit the word boundary check? Do you think it's better to remove the fallback or leave it with a TODO for the moment?

Regarding the stripDiacritics thing, this is necessary because VuFind's default spelling configuration strips diacritics, and so the suggestions will always come back without the diacritics. Thus, if I search for "tàst" and Solr suggests that I meant "test", the Solr spelling response will include a mapping from "tast" => "test", not "tàst" => "test". This will subsequently cause the term replacement to fail unless we normalize the user input.

Obviously, if you change the Solr spelling behavior, it's also going to break this code... and I think that reveals a design flaw. This logic is really in entirely the wrong place. It makes no sense to have normalization logic that's tied to a specific Solr configuration living in a low-level generic Query class which is shared by every search backend.

I would propose eliminating all of the normalization logic in the Query class, and instead allowing both containsTerm() and replaceTerm() to accept an optional normalization callback function. This would allow the Solr spelling handler to pass in an appropriate normalization function (since this is currently the only place where normalization is needed) while preventing weird side effects from occurring in other contexts. We could also add some new configuration settings to the spelling section which could control if/how text is normalized, so it's easier to work with custom Solr configurations without having to change code.

Of course, one step at a time: I think we should wrap up this pull request in more or less its current form first, since it should address the most obvious existing bugs. Then I think the refactoring I suggested above would be a good candidate for inclusion in VuFind 8 (since it obviously introduces some BC breaks).

EreMaijala:
@demiankatz Fair enough. I couldn't think of a case where word boundary wouldn't be the right thing to do, but perhaps some punctuation or something could trigger the issue?

I agree the Solr-specific issue doesn't need to be fixed here. What's really puzzling is that regardless of what normalization is done in Solr, I would have expected it to return the original term as the mapping source instead of the normalized one.

demiankatz:
@EreMaijala, good idea regarding punctuation. I'll do some experimentation with that when time permits.

Regarding normalization of diacritics, it's possible that either Solr behavior has changed or I'm misremembering something. I'll also put a little more time into reproducing that situation.

demiankatz:
@EreMaijala, I have confirmed that the need for normalization and lowercasing still appears to exist in VuFind's out-of-the-box configuration. I indexed a record with the title "Základy akupunktury" and then I searched for "Základz" (mistyping the last letter).

Here is the spelling response I got back from Solr:

<pre>
  "spellcheck":{
    "suggestions":
    [
      ["zakladz",{
          "numFound":1,
          "startOffset":0,
          "endOffset":7,
          "origFreq":0,
          "suggestion":[{
              "word":"zaklady",
              "freq":1}]}]],
    "correctlySpelled":false}
</pre>

As you can see, Solr returns a lower-cased, diacritic-stripped token instead of my original input. If I don't do matching normalization on the text on the PHP side, then no spelling suggestions are displayed.

I'm not sure why you are seeing different results on your end, though, unless they are related to local configuration changes.

Bottom line: I think we need to keep this behavior for now, but I still feel that my previous suggestions about refactoring the normalization logic from the query class to the spelling processor would make a lot more sense. For now, I'm going to spend a little more time trying to find a situation where the word boundary fallback is needed.

demiankatz:
Thanks also for the punctuation suggestion, @EreMaijala; you were correct: if the token being replaced ends in a comma (and probably other punctuation as well), the fallback clause is needed to properly replace the term. I have added a test case exercising this situation.

I suspect that there may still be some edge cases around this which could lead to problems, but I'm not going fishing for trouble right now. I think the immediate problem has been solved, and this PR leaves the code in a better place than where it started. I'm going to run the full test suite, and if it passes, I will merge this. Further issues can be addressed in separate PRs/tickets.

PR #1714, opened 2020-09-10
Refactor query normalization

demiankatz:
This is work in progress on refactoring text normalization out of the VuFindSearch\Query\Query class. Since the Query class is generic and shared, and normalization details can be backend specific, we should not combine these two things. This refactoring separates the concerns, simplifies the Query interface, and makes the overall system more flexible.

There is certainly room for debate and improvement in terms of naming things and how the classes relate to one another; this is just an initial proof of concept. Feedback is welcome! Since this significantly changes the Query interface, it should not be merged before we begin work on release 8.0.

See also #1712 for some conversation that inspired this work.

TODO
- [x] Discuss
- [x] Run full test suite
- [x] Document backward compatibility breaks when merging


demiankatz:
@EreMaijala, there is no rush since this work is not ready to be merged until we begin work on VuFind 8.0, but when you have time, I would be interested in your thoughts on these changes.

EreMaijala:
@demiankatz This makes sense to me. The one thing bugging me is passing the normalizer around. Would it be too horrible to pass Query the normalizer in constructor? Then e.g. getString could be `getString(bool $normalize = false)` or `getNormalizedString` and normalization wouldn't look so much like a side effect of passing the normalizer. I guess it depends on who is the authority on normalization, but unless there are cases where a Query might be passed different normalizers depending on a particular use, I'd say it would be more clear to let Query just use its own normalizer.

demiankatz:
@EreMaijala, I agree that passing the normalizer as a parameter is not very pretty, but unfortunately, I think it may be the least of the possible evils. The problem is that the code generates Query objects in a lot of different places, and many of those places do not have access to the Search Options object, which is currently the source of the normalizer object. I think that trying to inject search options in all of those places would actually be considerably uglier and make things significantly more complicated. (If you feel differently, by all means feel free to disagree -- I took a somewhat cursory glance at the situation, but what I saw was enough to discourage me).

I suppose another option would be to add a `setNormalizer` method to the Query object, and then inject it "as needed" and use a boolean switch to control whether or not to use it. But that seems more potentially confusing, redundant and indirect than just passing the normalization service directly.

This may all be a sign that we need to reconsider the whole way we create Query objects... maybe we need some kind of central query factory or something... but I'm not feeling particularly inspired to refactor all of that right now. Passing the normalizer when needed still seems like the least disruptive, if not the most elegant, approach to me.

Again, please feel free to disagree -- I'm happy to take more time discussing this to see if we can find a better solution. I'm just not seeing an obvious one right now. :-)

EreMaijala (approved):
I didn't realize we create Query objects everywhere. Your reasoning makes sense, but in the long run a factory would make it more clear.

I agree that adding a setNormalizer method would just complicate things, so this seems like a good interim solution.

demiankatz:
Thanks, @EreMaijala! I did a bit more review and thinking and tried to figure out how to open a JIRA ticket to capture outstanding work, but I couldn't come up with anything. Upon further reflection, I'm not sure we really want to tie a specific normalizer to a specific query object, because there might be situations in the future where we want to use different normalizers in different situations, or even apply a series of normalizers. Passing in a callback maximizes flexibility and separation of concerns. Not to say I'm unwilling to be persuaded otherwise in the future... but I just can't come up with a persuasive task to capture as a ticket right now. So I'm just going to merge this and consider it done for now, while leaving the door open to future conversation. :-)

PR #1719, opened 2020-09-15
Fix handling of nonce to not mess with HeadLink items.

EreMaijala:
Since link elements don't support the nonce attribute, don't try to add it to them. Besides, the $item format didn't use 'attributes' element anyway.


demiankatz:
This makes sense to me; just checking to see if @xmorave2 has any concerns.

xmorave2 (approved):
Thanks for the fix @EreMaijala

demiankatz:
Thanks to both of you!

PR #1721, opened 2020-09-18
Fix return value of getFilterStringParts when the string isn't proper…

EreMaijala:
…ly formatted.

Also adds tests to verify correct functionality.


demiankatz:
Thanks, @EreMaijala; looks like a straightforward fix to me, so I've backported this to release-7.0 and merged it to dev from there.

PR #1722, opened 2020-09-21
Fix mismatched slot names.

demiankatz:


demiankatz:
@crhallberg, I happened to notice that the start/end slot names were mismatched in this template. I assume it's a bug, but this raises a couple of questions:

1.) Why doesn't the mismatch break anything?

2.) Which name is correct?

crhallberg:
Yes! This was a typo. Thank you!

PR #1723, opened 2020-09-22
Alma: Clean up  error messages and always use logger.

EreMaijala:


demiankatz:
@EreMaijala, it looks like you have changed references to "Alma.ini" to "Alma ini" (removing the dot). I'm guessing this may have been the result of an overzealous search-and-replace, but in case it's intentional, I didn't want to change it back without checking first.

Other than that, this looks fine; I'll merge once you've had a chance to address that one small point!

EreMaijala:
@demiankatz That's actually intentional. With MultiBackend the ini file could be something different so I didn't want it to say Alma.ini exactly. Since 'Alma ini' was already used at least in one other place, I thought it would work for all. Ultimately it would be great if the driver actually knew the config file name, but that's something else then.

demiankatz:
Thanks for the clarification, @EreMaijala; that makes sense! I'll merge as-is.

PR #1724, opened 2020-09-22
FeedbackForms: allow extending Form class.

samuli:


demiankatz:
@samuli, this seems like a reasonable change to merge, but before I do so, I want to be sure it's actually necessary. Do you have a situation where you need to be able to use \VuFind\Form\Form in some situations, but a custom subclass in different situations? Or do you simply need to override \VuFind\Form\Form everywhere? If it's the first case, then this approach might help. If it's the second case, I would suggest creating an alias from \VuFind\Form\Form to your local class in the service manager configuration in your local module. This will allow you to override the form class without having to make changes to core code.

(Since I hope to eventually refactor the controllers to reduce internal use of the service locator, I'd prefer not to make this change if there's an easier way to solve the problem, since this will make later refactoring slightly more complicated... but if you really need it, I'll be happy to merge it).

samuli:
> Do you have a situation where you need to be able to use \VuFind\Form\Form in some situations, but a custom subclass in different situations?

Yes. In general we use \VuFind\Form\Form (aliased) but a few use cases are handled via a special feedback-controller that uses a special Form class:
https://github.com/NatLibFi/NDL-VuFind2/pull/1698/files#diff-28508559481a6a535e1404c83144efb2R39

Another option would be to include the special cases to our extended \VuFind\Form\Form, but this would be quite messy.

> Since I hope to eventually refactor the controllers to reduce internal use of the service locator,

Could the new service locator forward information about the caller (module, class, possibly method) to the existing factories? Then the $requesteName could be modified based on the calling context.

demiankatz:
Thanks, @samuli. In that case, I agree that this is the best solution for the moment; if this gets refactored later, the class name definition will just end up moving from the controller itself into the controller's factory, and you'll have to make matching adjustments in your custom code. If that happens, there will be a note about it in the changelog (but it's probably a long way off -- there are currently more pressing concerns than refactoring for the sake of refactoring).

PR #1725, opened 2020-09-22
Fix double colon in MARC links

xmorave2:
According to MARC standard documentation, when there is $i present, it should end with a colon:
https://www.loc.gov/marc/bibliographic/bd76x78x.html

In that case the colon character in MARC links generated in VuFind is doubled.

This patch fixes this and there is one colon for both cases consistantly.


demiankatz:
Thanks, @xmorave2; I've added a comment to the code for a little additional clarity. Merging now!

PR #1726, opened 2020-09-24
SierraRest: Fix a couple of holdings retrieval issues

EreMaijala:
- It wasn't intended that holdings be retrieved for getStatus(es).

- Apparently the location fixed field doesn't necessarily have a label in all Sierra versions (or configurations).

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
@@ -29,7 +29,7 @@ namespace VuFind\ILS\Driver;
 
 use Laminas\Log\LoggerAwareInterface;
 use VuFind\Exception\ILS as ILSException;
-use VuFind\Exception\VuFind\Exception;
+use VuFind\Date\DateException;
 use VuFind\I18n\Translator\TranslatorAwareInterface;
 use VuFindHttp\HttpServiceAwareInterface;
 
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -1733,8 +1734,8 @@ protected function getItemStatusesForBib($id)
             if (!empty($holdingsResult['entries'])) {
                 foreach ($holdingsResult['entries'] as $entry) {
                     $location = '';
-                    foreach ($entry['fixedFields'] as $field) {
-                        if ('LOCATION' === $field['label']) {
+                    foreach ($entry['fixedFields'] as $code => $field) {
+                        if ($code === 40 || $field['label'] === 'LOCATION') {
```
demiankatz: I assume "40" is some kind of system-level identifier; would it be worth defining as a class constant for clarity?

demiankatz (commented):
Makes sense to me! Just one small comment...

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Yes, I added a const.

demiankatz:
Thanks, @EreMaijala! I've cherry-picked the fix to the `use` statement back to the release-7.0 branch since that seems like a particularly critical fix. I'm merging the rest to dev now.

PR #1727, opened 2020-09-24
Use GET instead of POST for limit control.

demiankatz:
- Creates cleaner browser history (no back button problems, etc.).


demiankatz:
It was brought to my attention that the "sort" form uses GET but the "limit" form uses POST. I can't think of any good reason to use POST instead of GET, so I have adjusted the template. @EreMaijala / @crhallberg, any concerns? Anything I'm not thinking of?

EreMaijala:
@demiankatz No problems that I can think of right away. 

There's a bit of a discrepancy between handling of different parameters, though, as far as I can see: while limit and sort use the default values if not specified in the url, view does not. I suppose that's intentional so that a chosen view always sticks, but if you start a fresh session, make a search, change view and go back in history, the newly chosen view still sticks, while any changes to limit and sort would revert to the default values.

demiankatz:
Thanks, @EreMaijala. And yes, the view inconsistency is probably intentional, though that's not to say it couldn't be revisited if it's causing problems for anyone.

crhallberg (approved):

PR #1728, opened 2020-09-24
Update Evergreen.php getStatus() and getHoldings()

Wdcus:
Adds the condition to exclude deleted items in the database.


demiankatz:
Thanks, @Wdcus!

PR #1116, opened 2018-02-16
Better support for Koha status "NOT FOR LOAN"

KeckR:
See also https://vufind.org/jira/browse/VUFIND-1268

In case you like to set up more detailed status information in Koha for items, that are not for loan, you can use the category "NOT LOAN" in the "authorized values" section of Koha's system administration. We are currently using some values (e.g.at the bookbinder, internal administration, circulation, in acquisition, etc.). 

TODO

- [ ] Once standard.phtml template is finalized, make parallel edits to extended.phtml if necessary.

PR #1730, opened 2020-09-28
KohaRest: Allow mapping of different NotForLoan statuses separately.

EreMaijala:
Referring to discussion in #1116, this change improves the item status mapping capability in the KohaRest driver to handle different NotForLoan statuses.


module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -171,6 +175,16 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
...
     ];
 
+    /**
+     * Item status mappings
+     *
+     * @var array
+     */
+    protected $itemStatusMappings = [
+        // NotForLoan with status number 1 is available in Koha by default:
+        'Item::NotForLoan1' => 'On Reference Desk'
```
demiankatz: Just one question: does this status get translated somewhere? Should it? Do we need to add it to the language files somewhere?

EreMaijala: It can be translated as is, but there's no default translation. SierraRest uses the same string, so I took it from there. I'm not sure if it's a really good string, though, since "Not For Loan" could be something else too. Maybe it would be better to leave it only as an example in the ini file and not map it to anything by default.

EreMaijala: Then again, the same text is used for ItemType::NotForLoan in getItemStatusCodes.

demiankatz: Is "On Reference Desk" the text used in Koha? If not, maybe a more generic "Not Available for Loan" would be less specific (but could still be translated to something else as needed). If so, it probably makes sense for VuFind and Koha's language to match up.

xmorave2: There are no hardcoded not for loan status descriptions in Koha at all. It's all fully configurable, with optional defaults: -1:"Ordered", 1:"Not For Loan" and 2:"Staff Collection". (yes, the value could be negative integer)

EreMaijala: Maybe it's better to not have any default mappings. Then the description from Koha will be displayed, whatever it might be. It can still be translated if necessary.

demiankatz: Sounds reasonable to me. Have you tested whether the negative identifier case works reasonably? It might be worth including an example for that just to clarify what the keys would look like to customize those strings. Regarding translation, I can't recall whether we translate status strings at present (though my feeling is that we do not); I wonder if we need to revisit that at a VuFind level, or if it's better to implement translation at the driver level. If you need me to review all the details of current behavior, I'm happy to do so once time permits.

EreMaijala: Latest version should do it all, and I merged the ItemType status handling with Item status handling. While ItemType doesn't have all the same statuses, the NotForLoan status is pretty similar. Negative values work as well, and I added an example to KohaRest.ini.

EreMaijala:
I'll need to do something like this for the LOST statuses as well, but I'll get to that only after this one is done.

demiankatz:
This makes sense to me, and I'll merge it now. A few questions/comments for future reference:

1.) How well is translation currently working for these statuses? Are there any strings being used here as defaults that do not currently have translations defined? There's also some partial work in #1116 to namespace status translations, but I wasn't sure how that related to the work here. I think that could be helpful, but it could also be complicated and have a lot of implications, so I'm not in a huge rush to do more work than is strictly necessary... but it's something we should think about.

2.) I found `getItemStatusCodes()` a bit difficult to review because it is so long. I wonder if we could eliminate the lengthy switch and break this up into smaller parts to make it easier to understand. For example, we could define some `getItemStatusCodeFor[status]` methods, and then call them by doing a `method_exists` check on a dynamically generated string. If no method exists, we could check a class property or array of default statuses, which could condense all the cases that just do an assignment right now (Lost, Held, Waiting). And then there could be a final else clause for the default behavior. This might improve readability as well as making it easier to override individual statuses and behaviors without having to copy the whole switch statement. What do you think?

3.) I assume you'll submit the LOST handling as a separate PR... and maybe that will be easier to do following the refactoring proposed by item 2 (if you agree that's worthwhile, of course).

EreMaijala:
@demiankatz

1.) It's working fine. However, there's no translationPrefix configuration option like there is in the Alma driver. If we decide to put item statuses in a predefined namespace, we need to take into account the Alma driver option and whether it can still provide a namespace. I'd handle the namespace idea as a separate issue. 
Since Koha doesn't mandate any predefined values for the statuses, it's probably better to not include any default translations.

2.) I'll see what I can do. If it works out, I'll post a separate PR.

3.) Yes, I'll provide a separate PR for LOST statuses.

demiankatz:
Thanks, @EreMaijala, sounds good. If the status translations are currently working for you, I don't think we need to do extra work there unless there's a known issue. Good point about the Alma driver -- it would be nice to have uniformity across drivers, no matter where we land.

PR #1732, opened 2020-09-28
Remove laminas/laminas-dependency-plugin

liowalter:
I guess this is not needed any more ?

TODO
- [x] Run full test suite
- [x] Add changelog note when merging (Zend references will no longer be tolerated)


demiankatz:
@liowalter, at this point, it should no longer be required for the core VuFind code. I think we should leave it in place for remaining 7.x releases, because it may help ease the transition from VuFind 6.x if people have local code that still references Zend. However, it makes sense to remove it in VuFind 8.0. I have put an 8.0 milestone on this and will merge it when we begin 8.0 development. In the meantime, I'll run some tests to be 100% sure there are no side effects to removing it. (I'll comment here if there are problems; if not, I'll just check off the TODO list and then merge this at the appropriate time).

Thanks for pointing this out!

liowalter:
Ok, that makes sense !

Thanks !

PR #1733, opened 2020-09-29
Handle null response from Solr properly.

EreMaijala:
At least the similar handler in Solr 8.x can return null in the response element when no similar records are found.


demiankatz:
What happens if you fail to unset this? Is it worth adding a unit test to cover this case? (I'm happy to do so if you like).

EreMaijala:
@demiankatz array_replace_recursive will replace the template response element with a null, and on the following line `$this->response['response']['start']` bombs. I'll add a test.

EreMaijala:
Test added and check fixed to really work.

demiankatz:
Thanks!

PR #1734, opened 2020-09-29
Fix markdown factory not getting config from markdown.ini

xmorave2:
I realize there is a mistake in getting config for markdown service in MarkdownFactory (definitely my fault, sorry).

I fixed the factory and added a regression test for it.


demiankatz:
Thanks for the fix!

demiankatz:
Is it worth backporting this to the release-7.0 branch? I can try cherry-picking it when time permits if you think it's worth the effort.

xmorave2:
I do think it's worth to cherry-pick into 7.0, the patch should apply without big conflicts I think....

demiankatz:
Thanks, @xmorave2, I'll take care of it after the Summit. :-)

demiankatz:
Thanks again, @xmorave2 -- I was able to cleanly cherry-pick this back to release-7.0.

PR #1735, opened 2020-10-05
Reset authentication state on logout.

EreMaijala:
This probably has no effect in real life unless the followup after logout is short-circuited. However, if one would shortcut the redirect and go directly back to login screen, at least ChoiceAuth would get confused.


demiankatz:
Makes sense to me!

PR #1736, opened 2020-10-05
Make it possible to use smaller expiration than 2 days for sessions.

EreMaijala:
Allows defining a smaller minimum age than default for expire commands. Also makes sure that fractional values are printed with '.' as the decimal separator.

The output of the "Expiration age must be at least x days" is slightly changed to include one decimal also in the default case of 2. I don't find that an issue, but I can add logic to work around it if necessary. number_format is used because otherwise comma is used as the decimal separator in the error message with the Finnish locale.


demiankatz:
Makes sense to me!

PR #1737, opened 2020-10-06
Catch HTTP client's exceptions in SierraRest driver.

EreMaijala:
Otherwise e.g. curl adapter's timeout exception goes unhandled.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
@@ -1429,7 +1429,19 @@ class SierraRest extends AbstractBase implements TranslatorAwareInterface,
 
         // Send request and retrieve response
         $startTime = microtime(true);
-        $response = $client->setMethod($method)->send();
+        try {
+            $response = $client->setMethod($method)->send();
+        } catch (\Exception $e) {
+            $params = $method == 'GET'
+                ? $client->getRequest()->getQuery()->toString()
+                : $client->getRequest()->getPost()->toString();
+            $this->error(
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -1429,7 +1429,19 @@ protected function makeRequest($hierarchy, $params = false, $method = 'GET',
...
+        } catch (\Exception $e) {
+            $params = $method == 'GET'
+                ? $client->getRequest()->getQuery()->toString()
+                : $client->getRequest()->getPost()->toString();
+            $this->error(
+                "$method request for '$apiUrl' with params '$params' and contents '"
+                . $client->getRequest()->getContent() . "' caused exception: "
+                . $e->getMessage()
+            );
+            throw new ILSException('Problem with Sierra REST API.');
```
demiankatz: Is it worth passing through the previous exception here and in the other exceptions below -- e.g.`throw new ILSException('Problem with Sierra REST API.', 0, $e);`? That might allow for a more complete backtrace in case of trouble.

EreMaijala: Sure is, thanks. Now changed. Might be useful to modify other drivers too.

demiankatz: Thanks, and good idea: I've opened [VUFIND-1436](https://vufind.org/jira/browse/VUFIND-1436) as a reminder to update other drivers. We can make a to-do list on there as time permits.

PR #1739, opened 2020-10-07
Add pagination support to getUserTransactions

EreMaijala:
Without pagination support, account ajax will display the numbers based on the first page of loans only.

### DIFF
--- a/module/VuFind/src/VuFind/AjaxHandler/GetUserTransactions.php
+++ b/module/VuFind/src/VuFind/AjaxHandler/GetUserTransactions.php
@@ -101,6 +101,18 @@ class GetUserTransactions extends AbstractIlsAndUserAction
     }
 
     /**
+     * Set the ILS pagination helper
+     *
+     * @param \VuFind\ILS\PaginationHelper $helper Pagination helper
+     *
+     * @return void
+     */
+    protected function setPaginationHelper($helper)
+    {
+        $this->paginationHelper = $helper;
+    }
+
+    /**
      * Get the ILS pagination helper
      *
      * @return \VuFind\ILS\PaginationHelper
--- a/module/VuFind/src/VuFind/AjaxHandler/GetUserTransactions.php
+++ b/module/VuFind/src/VuFind/AjaxHandler/GetUserTransactions.php
@@ -57,25 +57,52 @@ class GetUserTransactions extends AbstractIlsAndUserAction
         if (!$this->ils->checkCapability('getMyTransactions')) {
             return $this->formatResponse('', self::STATUS_HTTP_ERROR, 405);
         }
-        $items = $this->ils->getMyTransactions($patron);
+
         $counts = [
             'ok' => 0,
             'warn' => 0,
             'overdue' => 0
         ];
-        foreach ($items['records'] as $item) {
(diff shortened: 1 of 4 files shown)


module/VuFind/src/VuFind/AjaxHandler/GetUserTransactions.php
```diff
@@ -57,25 +57,52 @@ public function handleRequest(Params $params)
...
+
+    /**
+     * Get the ILS pagination helper
+     *
+     * @return \VuFind\ILS\PaginationHelper
+     */
+    protected function getPaginationHelper()
+    {
+        if (null === $this->paginationHelper) {
+            $this->paginationHelper = new \VuFind\ILS\PaginationHelper();
```
demiankatz, line 123: Unless I'm missing something, I think it's necessary (or at least a good idea) to define this property at the top of the class... and maybe even to add a setter for easier customization.

module/VuFind/src/VuFind/AjaxHandler/GetUserTransactions.php
```diff
@@ -57,25 +57,52 @@ public function handleRequest(Params $params)
...
-                break;
-            case 'overdue':
-                $counts['overdue']++;
-                break;
-            default:
-                $counts['ok']++;
-                break;
+        $functionConfig = $this->ils->checkFunction('getMyTransactions', $patron);
+        $page = 1;
+        $pageEnd = 1;
```
demiankatz: Maybe it's just me, but I find it a little confusing to set $pageEnd to 1 above the loop and then to change it inside the loop; would it be clearer to change this to a do..while loop to force the first iteration and then evaluate the condition at the bottom?

module/VuFind/src/VuFind/AjaxHandler/GetUserTransactions.php
```diff
@@ -57,25 +57,52 @@ public function handleRequest(Params $params)
...
+                    break;
+                default:
+                    $counts['ok']++;
+                    break;
+                }
             }
+            $pageEnd = $pageOptions['ilsPaging']
+                ? ceil($result['count'] / $pageOptions['limit'])
+                : 1;
+            $page += 1;
```
demiankatz: Tiny nitpick, but this could be `$page++;`

demiankatz (changes requested):
Thanks, @EreMaijala -- I've cherry-picked the first commit from this PR back to release-7.0 since it seems like a useful fix, but I had a few (very minor and picky) comments about your second commit. Feel free to ignore some of them if I'm being overly detail-oriented. ;-)

module/VuFind/src/VuFind/AjaxHandler/GetUserTransactions.php
EreMaijala: You're absolutely right. I don't know why it worked without for me.

demiankatz: I believe that PHP lets you assign to undefined properties, it just defaults them to public. So it's legal, just not a good idea. :-)

demiankatz:
Thanks for all the adjustments; looks good to me now!

PR #1741, opened 2020-10-08
Refactor creating HTTP client in ObalkyKnih service

xmorave2:
I moved adding header into getHttpClient method. It makes inheritance easier if added method needs also an http client (and it probably does)


demiankatz:
Makes sense to me -- thanks!

PR #1742, opened 2020-10-08
Remove checkRoute=1 parameter from the url when it's no longer needed.

EreMaijala:
This allows one to copy a cleaner url for a record.


demiankatz:
Makes sense to me! Also timely, given that `history.replaceState` explains how you can change the current browser URL without actually reloading the page, a question which came up during the Summit last week when we were discussing SPAs and React.

PR #1743, opened 2020-10-08
Hierarchical facet fix for at least advanced search form

EreMaijala:
I've been running with some invalid values in hierarchical facets and encountered this in the advanced search screen. Related to #1740.


demiankatz:
Makes sense, thanks!

PR #1744, opened 2020-10-08
Alma: Use rawurlencode for path parts.

EreMaijala:
urlencode encodes spaces as pluses, while the path part of url only allows %20. Therefore rawurlencode must be used for path parts. Alma allows spaces in user's primary id, which causes requests for e.g. `/users/+identifier` to fail.


demiankatz:
Thanks! I've backported this one to the release-7.0 branch.

PR #1746, opened 2020-10-09
Catch exception on HTTP request

xmorave2:
We faced an exception with "Unable to enable crypto" on connection to obalkyknih.cz, than we assume we can't get a cover and so just return null

### DIFF
--- a/module/VuFind/src/VuFind/Content/ObalkyKnihService.php
+++ b/module/VuFind/src/VuFind/Content/ObalkyKnihService.php
@@ -166,6 +166,7 @@ class ObalkyKnihService implements \VuFindHttp\HttpServiceAwareInterface,
         try {
             $response = $client->send();
         } catch (\Exception $e) {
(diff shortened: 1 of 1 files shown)


demiankatz:
@xmorave2, the "unable to enable crypto" message can be a sign of a misconfiguration (one possible cause is that PHP can't verify SSL certs because it can't find the certificate store due to a missing sslacapath/sslcafile setting in config.ini). I think it does make sense to swallow these errors so that they don't break the whole system, but it's probably also a good idea to log an error message at the same time so the problem is reported. Since the service already implements the LoggerAwareInterface, this should just be a matter of `$this->logError('Unexpected ' . get_class($e) . ': ' . $e->getMessage());` or something similar inside the catch.

xmorave2:
Thanks @demiankatz. I added logging as you suggested.

demiankatz:
Thanks, @xmorave2, looks good!

PR #1747, opened 2020-10-09
Cache results of class based template resolution.

EreMaijala:
This has the potential to avoid a lot of repeated resolution requests e.g. when rendering search results.

TODO:
- [x] Edit changelog when merging to document BC breaks (replacement of AbstractClassBasedTemplateRenderer with ClassBasedTemplateRendererTrait, plus reworking of resolveClassTemplate method).
- [x] Run full test suite

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/AbstractClassBasedTemplateRenderer.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/AbstractClassBasedTemplateRenderer.php
@@ -29,6 +29,7 @@ namespace VuFind\View\Helper\Root;
 
 use Laminas\View\Exception\RuntimeException;
 use Laminas\View\Helper\AbstractHelper;
+use Laminas\View\Resolver\ResolverInterface;
 
 /**
  * Authentication view helper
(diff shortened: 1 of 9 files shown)


module/VuFind/src/VuFind/View/Helper/Root/AbstractClassBasedTemplateRenderer.php
```diff
@@ -42,25 +43,33 @@
...
+     * placeholder)
+     * @param string            $className    Name of class to apply to template.
+     * @param ResolverInterface $resolver     Resolver to use
+     * @param string            $topClassName Top-level parent class of $className
+     * (or null if $className is already the top level; used for recursion only).
      *
      * @return string
      * @throws RuntimeException
      */
     protected function resolveClassTemplate($template, $className,
```
demiankatz, line 71: Changing the signature and return content of this method potentially introduces a backward compatibility break, which makes me wonder if we should defer this PR until 8.0. I thought about whether we could make the same changes without the BC break, and while we could, I think what you've done actually makes the method naming more accurate, so I don't think it makes sense to go out of our way to prevent change.

Of course, if we are already introducing a BC break, maybe an even better thing to do is to move this functionality into a Trait and deprecate this abstract base class, since the Trait approach is more flexible and cleaner.

module/VuFind/src/VuFind/View/Helper/Root/AbstractClassBasedTemplateRenderer.php
```diff
@@ -93,14 +102,22 @@ protected function resolveClassTemplate($template, $className,
...
-        $contextHelper = $this->getView()->plugin('context');
-        $oldContext = $contextHelper($this->getView())->apply($context);
+        $view = $this->getView();
+        $contextHelper = $view->plugin('context');
+        $oldContext = $contextHelper($view)->apply($context);
 
-        // Render the template for the current class:
-        $html = $this->resolveClassTemplate($template, $className);
+        // Find the template for the current class:
+        if (!isset($this->templateCache[$className][$template])) {
```
demiankatz, line 135: I feel like the cache functionality might be cleaner in its own method -- i.e. `$this->getCachedClassTemplate()`; I think it would improve readability of this method if the caching details were separated.

demiankatz (commented):
This makes a lot of sense to me; just a couple questions about minor details and release timing...

demiankatz:
Also, just for the record: I ran the full test suite and everything is passing with these changes in place.

module/VuFind/src/VuFind/View/Helper/Root/AbstractClassBasedTemplateRenderer.php
EreMaijala: I agree that 8.0 makes sense as the target. I moved the functionality to a trait. I kept it in View/Helper/Root since it assumes AbstractHelper with getView() as the base class. It could be made more generic by adding the view helper as a parameter, but I couldn't think of a need outside view helpers and ensuring that helps keep the semantics sane. Do you agree?

module/VuFind/src/VuFind/View/Helper/Root/AbstractClassBasedTemplateRenderer.php
EreMaijala: Done.

module/VuFind/src/VuFind/View/Helper/Root/AbstractClassBasedTemplateRenderer.php
demiankatz: Yes, this makes sense to me. It might not hurt to add a comment indicating that the trait should only be used in a view helper because of the dependency on getView(), but leaving it in this namespace may be enough to imply that! In any case, this looks good to me, so I'll just leave the PR open for now, and will retest/merge when we're ready to start 8.0 development in the dev branch.

EreMaijala: Thanks, I added a short comment at the top of the trait.

demiankatz:
@EreMaijala, I've merged this to dev-8.0 as well.

PR #1749, opened 2020-10-14
Making the limit of records to retrieve per query in a batch query co…

hajoseng:
…nfigurable. This is because or Solr-Server has set the limit for "ORed" queries to a lower number (50) and we are forced to decrease the amount of records per batch query from 100 to 50.


config/vufind/config.ini
```diff
@@ -558,7 +558,10 @@ default_dismax_handler = dismax
 ; hierarchy. A higher number results in fewer round-trips but may increase Solr's
 ; memory usage. Default is 1000.
 ;cursor_batch_size = 1000
-
+; This limits the number of records to retrieve in a batch e.g. when retrieving
+; records for a list. Default is 100 which should be a good number to avoid
+; memory problems.
+;record_batch_size
```
demiankatz: Can you add `= 100` to the example here to make it more clear?

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
```diff
@@ -182,7 +182,9 @@ public function __invoke(ContainerInterface $sm, $name, array $options = null)
      */
     protected function createBackend(Connector $connector)
     {
+        $config = $this->config->get($this->mainConfig);
         $backend = new $this->backendClass($connector);
+        $backend->setPageSize($config->Index->record_batch_size);
```
demiankatz: If record_batch_size is not set in config.ini, this code will trigger a notice about an undefined value. I would suggest either wrapping the setPageSize() call in an `if (!empty(...)) {` check, or else providing a default value in the set call (i.e. `$config->Index->record_batch_size ?? 100`).

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Backend.php
```diff
@@ -94,6 +101,20 @@ public function __construct(Connector $connector)
...
+    /**
+     * Set the limit for batch queries
+     *
+     * @param int $pageSize Records per Query
+     *
+     * @return void
+     */
+    public function setPageSize($pageSize)
+    {
+        if (!empty($pageSize)) {
```
demiankatz: If you move the validation to the factory, you may not need this... but if you do want to validate, perhaps a better check would be whether the value is > 0, to avoid weird behavior caused by invalid configurations.

demiankatz (changes requested):
Thanks, this makes sense to me... just a few minor suggestions!

Also, it would be nice to add some test coverage for this new functionality (see https://github.com/vufind-org/vufind/blob/dev/module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php) -- I can help with that if you're not already familiar with the test suite. If you can give me write access to the branch, I can add a test directly; otherwise, I can update the test after merging the rest of the PR.

demiankatz:
@hajoseng, one other thought: we already have a setting for limiting maximum boolean clauses:

https://github.com/vufind-org/vufind/blob/dev/config/vufind/config.ini#L550

Maybe the factory should take this into account, and if record_batch_size is larger than maxBooleanClauses, it should automatically adjust it. This would, of course, argue toward accounting for the default value of 100 in the factory.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Backend.php
hajoseng: Thank you Demian
I'm a little bit familiar with the test suite, but in this case I have no idea which test to add. I've given you write access to our repository to show me.

demiankatz:
Thanks, @hajoseng -- I made a couple of small adjustments to the factory to prevent possible problems (see 61196db919bc0b7c4405c0eb74fe2cfaada0fa37) -- please review. I'll take care of the testing in the near future, as time permits.

demiankatz (approved):

demiankatz:
@hajoseng, I've just added the test as promised. Please let me know if you have any questions/concerns. If you're happy with everything, I'll merge this as soon as I hear from you!

hajoseng:
@demiankatz, thank you very much; I'm happy with your adjustments.

demiankatz:
Thanks; merging now!

PR #1752, opened 2020-10-15
return null if lc call number is empty

ihardy:
Returns null for lc call number instead of an empty string in the case where there is no lc call. This is to address this issue from the listserv: https://sourceforge.net/p/vufind/mailman/message/37117867/


demiankatz:
Thanks, @ihardy. I simplified this to remove the else, since I think it's slightly more readable to just drop through to the final default return condition. Any objections? If not, I'll go ahead and merge. Thanks for catching this!

ihardy:
Sounds good to me, much appreciated Demian.

PR #1753, opened 2020-10-15
Return to previous page after printing.

crhallberg:
Automatically return the user to the original page when printing is triggered by a GET parameter.


demiankatz:
@crhallberg, this appears to be breaking the Travis build, and it's not working for me in Firefox...

PR #1754, opened 2020-10-16
Improve record tab hash handling

rajaro:
Fixes an issue where anchor links can trigger click on links inside elements with the same class name as the new hash

### DIFF
--- a/themes/bootstrap3/js/record.js
+++ b/themes/bootstrap3/js/record.js
@@ -283,7 +283,7 @@ function applyRecordTabHash() {
   if (newTab.length <= 1 || newTab === '#tabnav') {
     $initiallyActiveTab.click();
   } else if (newTab.length > 1 && '#' + activeTab !== newTab) {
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks, @rajaro, this makes sense to me. I'm asking @crhallberg and @EreMaijala to double-check in case there are any implications I'm overlooking. I confirmed that the "record-tabs" class is consistently applied in both Record and Collection views, so that potential inconsistency is not an issue. I will also run the full test suite to make sure there are no problems there.

demiankatz:
All tests are passing, which seems like a very good sign... but since there have been some unexpected bugs related to changes to the tab code in the past, I'm going to wait for some additional reviews before merging, out of an abundance of caution.

EreMaijala (approved):

EreMaijala:
@demiankatz Looks good to me. The old code was indeed a bit scary.

demiankatz:
Thanks, @EreMaijala and @rajaro -- I have merged this now!

PR #1756, opened 2020-10-19
KohaRest: Split GetItemStatusCodes into smaller sub-methods.

EreMaijala:
Follow-up from #1730. I'm not quite convinced that this is easier to read, but at least it allows for easier overriding of specific cases if needed.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
@@ -183,6 +183,23 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
     protected $itemStatusMappings = [];
 
     /**
+     * Item status mapping methods used when the item status mappings above don't
+     * contain a direct mapping.
+     *
+     * @var array
+     */
+    protected $itemStatusMappingMethods = [
+        'Item::CheckedOut' => 'getStatusCodeItemCheckedOut',
+        'Item::Lost' => 'getStatusCodeItemLost',
+        'Item::NotForLoan' => 'getStatusCodeItemNotForLoan',
+        'Item::NotForLoanForcing' => 'getStatusCodeItemNotForLoan',
+        'Item::Transfer' => 'getStatusCodeItemTransfer',
+        'Item::Held' => 'getStatusCodeItemHeld',
+        'Item::Waiting' => 'getStatusCodeItemWaiting',
+        'ItemType::NotForLoan' => 'getStatusCodeItemNotForLoan',
+    ];
+
+    /**
      * Whether to display home library instead of holding library
      *
      * @var bool
@@ -1750,9 +1767,7 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
                 }
 
                 // Check for a mapping method for the unavailability reason:
-                $methodName = 'getStatusCode'
-                    . preg_replace('/[^a-zA-Z0-9_]/', '', $code);
-                if (method_exists($this, $methodName)) {
+                if ($methodName = ($this->itemStatusMappingMethods[$code] ?? '')) {
                     $statuses[]
                         = call_user_func([$this, $methodName], $code, $data, $item);
                 } else {
--- a/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
@@ -1862,20 +1862,6 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
     }
 
     /**
-     * Get item status code for NorForLoanForcing status
-     *
-     * @param string $code Status code
-     * @param array  $data Status data
-     * @param array  $item Item
-     *
-     * @return string
-     */
-    protected function getStatusCodeItemNotForLoanForcing($code, $data, $item)
-    {
-        return $this->getStatusCodeItemNotForLoan($code, $data, $item);
-    }
-
-    /**
      * Get item status code for Transfer status
      *
      * @param string $code Status code
@@ -1925,20 +1911,6 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
     protected function getStatusCodeItemWaiting($code, $data, $item)
     {
         return 'On Holdshelf';
-    }
-
-    /**
-     * Get item type status code for NotForLoan status
-     *
-     * @param string $code Status code
-     * @param array  $data Status data
-     * @param array  $item Item
-     *
-     * @return string
-     */
-    protected function getStatusCodeItemTypeNotForLoan($code, $data, $item)
-    {
-        return $this->getStatusCodeItemNotForLoan($code, $data, $item);
     }
 
     /**
--- a/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
@@ -1750,7 +1750,7 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
                 }
 
                 // Check for a mapping method for the unavailability reason:
-                $methodName = 'GetStatusCode'
+                $methodName = 'getStatusCode'
                     . preg_replace('/[^a-zA-Z0-9_]/', '', $code);
                 if (method_exists($this, $methodName)) {
                     $statuses[]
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -1742,67 +1742,27 @@ protected function getItemStatusCodes($item)
...
-                                as $noteKey
-                            ) {
-                                if ('Item::Held' === $noteKey) {
-                                    $onHold = true;
-                                    break;
-                                }
-                            }
+
+                // Check for a mapping method for the unavailability reason:
+                $methodName = 'GetStatusCode'
```
demiankatz: Should this start with a lowercase 'g'? (I realize that method names are case-insensitive in PHP, but it just feels like it would be better to match them up).

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -1742,67 +1742,27 @@ protected function getItemStatusCodes($item)
...
+                if (method_exists($this, $methodName)) {
+                    $statuses[]
+                        = call_user_func([$this, $methodName], $code, $data, $item);
+                } else {
+                    if (!empty($data['code'])) {
+                        $statuses[] = $data['code'];
+                    } else {
+                        $parts = explode('::', $code, 2);
+                        if (isset($parts[1])) {
+                            $statuses[] = $parts[1];
```
demiankatz, line 1780: It looks like this is a change from the old logic, which would have done: `$statuses[] = $parts[1] ?? ''`. It now looks like empty status codes will be ignored rather than set as blank strings. This may well be more correct than before; I just wanted to note the change in behavior in case it was unintentional.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -1821,6 +1781,151 @@ protected function getItemStatusCodes($item)
...
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     */
+    protected function getStatusCodeItemTransfer($code, $data, $item)
+    {
+        $onHold = array_key_exists(
+            'Item::Held',
+            $item['availability']['notes'] ?? []
+        );
+        return $onHold ? 'In Transit On Hold' : 'In Transit';
```
demiankatz, line 1866: Is `'In Transit On Hold'` a typo for `'On Hold'`?

demiankatz (commented):
I've left a few minor comments here. I agree that this is still pretty unwieldy, though as you say, it does enable more granular overriding.

If you don't like having to create multiple methods that wrap around each other to apply the same functionality to multiple codes, maybe it would be cleaner/more straightforward to have a class property that maps codes to method names instead of using a naming convention.

A general problem with most of our ILS drivers is that they all have an unreasonable number of methods in them. The question is whether it's worthwhile to split them up with support classes, which would be more "proper" from an OOP perspective but would eliminate the simplicity of maintaining a single file and introduce more complex factories, etc. It's an architectural issue that's been on my mind for a long time... it's not the end of the world, but it's at least a little bit annoying.

Anyway, happy to brainstorm further if you're not entirely happy with the direction this has taken.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: True. This should never happen, so the change should have no effect, but in case we can't determine a proper status it's better to fall back to the 'Not Available' status below, I think.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: No, it's a string that, I believe, has its roots in Voyager. It's used when the item is on hold but still in transit to the pickup location. I'm open to using something else as the default string, though I recall that libraries had some opinions on the wording including that they don't want it to convey any idea of when the item might be available for pickup.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: Yes, now fixed.

EreMaijala:
@demiankatz Thanks for your comments. I found the method mapping property to be much nicer, now switched to that. I agree that the drivers tend to be a bit huge, but being mostly self-contained has benefits too. Splitting the long ones up would make sense especially if there are parts that could be reused, but I can't see too many of those.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
demiankatz: Okay, great, in that case I'm fine with leaving it in place. If people don't like it, they can always translate it.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -1821,6 +1796,151 @@ protected function getItemStatusCodes($item)
...
+    /**
+     * Get item status code for NorForLoanForcing status
+     *
+     * @param string $code Status code
+     * @param array  $data Status data
+     * @param array  $item Item
+     *
+     * @return string
+     */
+    protected function getStatusCodeItemNotForLoanForcing($code, $data, $item)
```
demiankatz: This can be removed now, can't it?

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -1821,6 +1796,151 @@ protected function getItemStatusCodes($item)
...
+    /**
+     * Get item type status code for NotForLoan status
+     *
+     * @param string $code Status code
+     * @param array  $data Status data
+     * @param array  $item Item
+     *
+     * @return string
+     */
+    protected function getStatusCodeItemTypeNotForLoan($code, $data, $item)
```
demiankatz: This can also be removed, correct?

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -182,6 +182,23 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
      */
     protected $itemStatusMappings = [];
 
+    /**
+     * Item status mapping methods used when the item status mappings above don't
```
demiankatz: If methods are only used when mappings are undefined, why not set mappings instead of defining methods for the situations where a hard-coded value is returned without any logic? Or am I misunderstanding?

demiankatz (changes requested):
Looks like this is moving in the right direction, but I think there's still a little more cleanup to be done. Thanks for the progress!

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
demiankatz: Agreed, that makes more sense!

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: Oops, yes, removed.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: Oops, yes, removed.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: They can be used at the same time. You can e.g. configure a single status differently just by adding it to the mappings in the .ini file. Methods are used for any statuses that don't have a mapping.

demiankatz: Understood, but my question is, if the method is simply returning a hard-coded string value, would it make more sense to make the default implementation a default entry in the mappings array instead of a custom method? If somebody wanted to override with a custom method in a subclass, they could still null out the mappings entry and add a method map entry. Or am I missing a nuance that makes this impractical/inappropriate?

EreMaijala: Sorry, I misunderstood. Totally makes sense. Now simplified.

demiankatz (approved):

demiankatz:
Thanks, @EreMaijala, I think this is a lot simpler and more maintainable than the giant switch statement now! Looks like Travis is slow today, but once the build passes, I think I should be able to merge this.

PR #1758, opened 2020-10-20
Feedbackforms: handle undefined reply-to address.

samuli:
`replyToEmail` can be undefined when not posted via the form and when profile email can not be used as a fallback (profile email not defined or user is not logged-in).


demiankatz:
Thanks! I've cherry-picked this back to the release-7.0 branch since I think it qualifies as a bug fix. :-)

PR #1759, opened 2020-10-20
KohaRest: Take public notes for items from the correct field.

EreMaijala:
I had missed the fact that the itemnotes internal field is now translated to public_notes in the Koha REST API.


demiankatz:
I cherry-picked this one back to release-7.0 since it's a bug fix. Thanks!

PR #1761, opened 2020-10-20
CSV import tool

demiankatz:
This pull request contains work in progress on a tool to import CSV files to Solr (with optional pre-processing).

TODO
- [x] Implement basic command framework
- [x] Design configuration file format (capabilities needed: optional header row support, positional field labeling, field copying support, custom field-specific processing support from calbacks and static classes).
- [x] Implement configuration reading
- [x] Implement column processing
- [x] Implement support for hard-coded field values (not coming directly from input data)
- [x] Implement support for different character encodings
- [x] Eliminate need for defaultOutputDelimiter setting by using JSON to write to Solr instead of CSV
- [x] Confirm end-to-end support
- [x] Make sure Solr errors are handled appropriately
- [x] Write test(s)
- [ ] Complete documentation in [wiki](https://vufind.org/wiki/indexing:csv) when merging


demiankatz:
I believe that I have a minimum viable product here, though I need to spend more time experimenting with character encodings and writing tests. Feedback is welcomed!

demiankatz:
I'm not convinced that the end-to-end CSV approach used here (where we post data to Solr in CSV format) is actually the best or most logical strategy. Before I change it, I have opened #1952, to add the CSV writing capabilities independently of this work. That way, if we need that feature in future, we will have it, but if we stop using it here, it will not be lost. I will resume work on this PR after that one is merged.

demiankatz:
@EreMaijala, I believe that this feature has reached "minimum viable product" stage... but before I polish it up with full test coverage, documentation, etc., I'd like to get opinions on whether the approach is reasonable and the configuration makes sense. I realize that you probably don't need this particular tool, so if you prefer not to take the time to review it, let me know and I can seek other volunteers.

EreMaijala (changes requested):
I think this is mostly good, and the following are just some minor details.

The comments could to be a bit more precise and informative. I first thought that https://github.com/vufind-org/vufind/pull/1761/files#diff-b13db82cdbe7ae917d8176f2fff78203988be3c9896e76af4a9c05a91b764cacR39 refers to the whole section somehow (I know, it's inside the section, but still) and got confused. Saying something the "The callback setting specifies..." would make it more explicit. Also, what happens when there are multiple callbacks? Are they all executed in order and what can they return? Is there a way to drop a value by returning an empty string or null or something? I suppose the wiki page will also help here, and I wonder if it would make sense to actually make the comments shorter and refer to the wiki page for more information.

I haven't checked if things have improved, but fgetcsv may still have some issues e.g. with multiline strings. I once wrote this (a bit ugly, but works) to overcome the issues:
https://github.com/emaijala/MLInvoice/blob/91e01cf1fe9059ad85ce3661fb2652bf5c2bf46f/import.php#L1177

Feel free to borrow at least the idea if you want to. Also if you want to support UTF-16, some extra processing may be needed (see e.g. https://github.com/emaijala/MLInvoice/blob/91e01cf1fe9059ad85ce3661fb2652bf5c2bf46f/import.php#L1110).

demiankatz:
Thanks, @EreMaijala! I have improved the documentation in csv.ini -- let me know if that addresses all of your questions. I have also created a wiki page, and linked to it from the .ini file. It is empty for now, but I will flesh it out after I finish writing tests. At that point, I will decide whether to put more information in the wiki or in the .ini file... but in any case, the two should complement each other one way or another.

Regarding multiline strings, do you have a CSV containing a multiline string you could share for use as a test fixture? I'm not entirely sure what to expect there in terms of formatting, so an example might be helpful.

I'll be sure to include UTF-16 as one of the encoding test fixtures, when I get that far!

EreMaijala (approved):

EreMaijala:
@demiankatz Thanks, it's easier to understand now. Unfortunately I don't have a suitable multiline csv example at hand, but here's one that you could think of importing into VuFind: https://autohotkey.com/board/topic/97147-parsing-csv-files-with-multi-line-fields/

demiankatz:
@EreMaijala, thanks again for your feedback on this. I've built a fairly robust test suite now, which revealed some shortcomings that have now been fixed. I am certain that we'll find more issues when we try to load large amounts of real data with this tool, but I think it's solidly in "minimum viable product" mode now, and we can refine from here as circumstances demand.

A few specific notes:

- I realized that my documentation regarding order of callback processing was incorrect, and that the code was missing the ability the run callbacks to create new values (as opposed to processing existing values), so I added a callbackSeed setting to enable that possibility.
- I was unable to get UTF-16 data to import; I tried to adapt your code, but it ended up going into an infinite loop, and I decided not to spend too much time investigating. For now, I included a recommendation that CSVs be converted to a supported character set prior to ingestion, which may be adequate for most people's needs (rather than creating complex code that we have to maintain ourselves). Happy to revisit if there's an actual use case in the future!
- Multi-line CSVs seem to work natively without any problems... though perhaps there are some more complex cases that work less well than my very simple test case.

Since this is an entirely new feature with low risk of breaking any existing functionality, and since I have your approval on a slightly earlier revision, I'm going to go ahead and merge this soon. If you have any further suggestions or concerns, I'm happy to make further adjustments in a subsequent PR as needed.

PR #1762, opened 2020-10-21
Make the max limit of search API results more flexible

xmorave2:


demiankatz:
Makes sense to me. Any comments, @EreMaijala?

EreMaijala (approved):
Makes sense to me!

It might also be useful to make it configurable in the corresponding ini file, but that's just a thought for the future.

xmorave2:
Thank for review @demiankatz and @EreMaijala 

Could be done configurable in the next step, just one thing we need is to have ability to configure any index (primary, secondary, web, ...) differently.

demiankatz:
Thanks, everyone, I'm merging this as-is for now -- just getting rid of that "magic number" hard-coded in two different places is a useful step forward, and I'll welcome a new PR adding a config setting if the need arises in the future.

PR #1763, opened 2020-10-23
Treat invalid date errors from Solr as parser errors.

EreMaijala:
Currently date parsing errors end up in a 400 Bad Request. This change allows Solr's date parsing errors to be considered query parsing errors. E.g. Bing bot seems to be sometimes double-encoding pluses in a query string resulting in something like this in the VuFind search query string: `lookfor=xyz&filter[]=first_indexed%3A"[NOW-3MONTHS%2FDAY%2BTO%2B*]"`


demiankatz:
Seems reasonable to me!

PR #1764, opened 2020-10-24
Use Psalm for static code analysis

demiankatz:
[Psalm](https://github.com/vimeo/psalm) is a static code analysis tool that we can add to our toolkit for improving VuFind's code quality. This PR introduces it.

TODO:
- [ ] Review default configuration and adjust if necessary
- [ ] Fix outstanding issues (where practical)
- [x] Convert Travis build code into GitHub Actions format
- [x] Consider setting up a [baseline file](https://psalm.dev/docs/running_psalm/dealing_with_code_issues/#using-a-baseline-file), if we want to merge this before all issues are fixed/annotated.
- [ ] Incorporate into Jenkins build (after merge)
- [x] Investigate whether phpstan is a useful complement to this tool (see #1896)
- [ ] Wait for a solution to https://github.com/vimeo/psalm/issues/3621


demiankatz:
I spent some more time today bringing this up to date and trying to resolve issues. One of the most prevalent errors is that Psalm doesn't like factories stating `@throws ContainerInterface` because `ContainerInterface` does not implement `Throwable`. I think it should be possible to overcome this with an intersection type -- see #1923 -- but this unfortunately does not work, due to this apparent known bug in Psalm: https://github.com/vimeo/psalm/issues/3621

We should revisit this in the future and see if the bug is fixed.

demiankatz:
I'm flagging this as "on hold" until https://github.com/vimeo/psalm/issues/3621 is resolved, since we currently can't get the error list to clear out without intersection types working properly.

demiankatz:
I took a few minutes today to bring this up to date. I think the existing problem with union types remains unresolved, but I thought it might be helpful to get this integrated with GitHub Actions, etc.

demiankatz:
I've managed to get the build to pass here by suppressing several of the false-positive errors that have been causing problems. Perhaps the best approach for moving forward here is to either suppress or resolve all remaining errors until we have a clean baseline. Then we can re-evaluate the suppressed checks to see which ones are completely useless and which ones might be valuable in combination with appropriate code annotations. I can already see that some of the issues Psalm can detect and fix will lead to improvements in code quality, so it seems worth getting some of those benefits as soon as we can, even if it means not taking full advantage of everything Psalm has to offer right away.

PR #1765, opened 2020-10-25
Begin fixing return types using Psalm.

demiankatz:
TODO:
- [x] Review auto-generated changes; correct problems


demiankatz:
This PR is a proof of concept and does not correct all identified problems; more work will need to be done later.

PR #1767, opened 2020-10-26
Re-set keyboard focus in lightbox after the content has loaded.

EreMaijala:
The initial call in setFocusToFirstNode() in bindFocus() may be called before modal content has been loaded, which leads the focus to be left in the close button. Calling it after the show command makes sure any actual focusable content gets the focus. I left the call in bindFocus() too since it will allow one to close the dialog directly also before it's fully loaded.

TODO
- [x] Run full test suite


demiankatz:
Thanks, @EreMaijala! This looks fine to me, but I'll wait for @crhallberg's input before proceeding. Assuming he approves it, I'll run the full test suite (just to be sure) and then merge.

crhallberg:
This looks good to me! I don't see why it wouldn't pass the full suite.

crhallberg (approved):

demiankatz:
As predicted, all tests are passing. Merging now!

PR #1770, opened 2020-10-27
Account for more table of contents subfields.

demiankatz:
As previously written, every subfield in a 505 would be treated as a new entry in the TOC list... but some subfields are intended to supplement adjacent subfields rather than be treated separately. This PR changes behavior to split the data more appropriately.

TODO:
- [x] Adjust to meet separator standards from [input conventions](https://www.loc.gov/marc/bibliographic/bd505.html)
- [x] Add test coverage


EreMaijala (approved):
Looks ok to me.

PR #1771, opened 2020-10-28
KohaRest: Add support for mapping different "Lost" statuses.

EreMaijala:
With all the hard work already done, this is a pretty simple one.

### DIFF
--- a/config/vufind/KohaRest.ini
+++ b/config/vufind/KohaRest.ini
@@ -89,11 +89,12 @@ acceptTermsText = "I accept the terms in any language."
 ; VuFind item statuses
 [ItemStatusMappings]
 ;Item::Held = "Held"
-; Custom "Not For Loan" or "Lost" statuses can be mapped by appending the authorized
-; value to the status key like this (first one is negative authorized value, second
-; one is positive):
+; "Not For Loan" or "Lost" statuses can be mapped by appending the authorized value
+; to the status key like this (first one is a negative authorized value, next ones
+; are positive):
 ;Item::NotForLoan-1 = "Ordered"
 ;Item::NotForLoan4 = "At Bindery"
+;Item::Lost4 = "Lost--Library Applied"
 ; Also "Not For Loan" item types can be mapped by appending the authorized value to
 ; the status key like this:
 ;ItemType::NotForLoan1 = "Reference Desk"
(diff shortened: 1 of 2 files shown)


config/vufind/KohaRest.ini
```diff
@@ -89,9 +89,9 @@ acceptTermsText = "I accept the terms in any language."
...
 [ItemStatusMappings]
 ;Item::Held = "Held"
-; Custom "Not For Loan" statuses can be mapped by appending the authorized value to
-; the status key like this (first one is negative authorized value, second one is
-; positive):
+; Custom "Not For Loan" or "Lost" statuses can be mapped by appending the authorized
+; value to the status key like this (first one is negative authorized value, second
+; one is positive):
 ;Item::NotForLoan-1 = "Ordered"
 ;Item::NotForLoan4 = "At Bindery"
```
demiankatz, line 96: Would it make sense to add a Lost example here, just for clarity?

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -182,7 +182,6 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
      */
     protected $itemStatusMappings = [
         'Item::Held' => 'On Hold',
-        'Item::Lost' => 'Lost--Library Applied',
```
demiankatz, line 185: You've deleted this code but not added it anywhere else. Should this be used as my proposed example in the .ini file? Do we need a mechanism for configuring a global fallback string independent of the numeric codes?

demiankatz (commented):
A couple of minor comments...

config/vufind/KohaRest.ini
EreMaijala: Sure, I'll add.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: It's now right below in itemStatusMappingMethods.

demiankatz: Sorry for being unclear; I was referring specifically to the text "Lost--Library Applied" rather than the 'Item::Lost' code itself.

EreMaijala: Right. Now the code takes the default text if there's no mapping so I thought it would make sense to use that. I can add that term as an example, that's a good idea.

demiankatz:
Looks good to me!

PR #1772, opened 2020-10-29
Fix loading of similar records tab.

EreMaijala:
Binding the load event must be done with .on() method in jQuery 3. See [the docs](https://api.jquery.com/load-event/) for more information.

### DIFF
--- a/themes/bootstrap3/templates/RecordTab/similaritemscarousel.phtml
+++ b/themes/bootstrap3/templates/RecordTab/similaritemscarousel.phtml
@@ -65,7 +65,7 @@ function normalizeHeights() {
   });
   items.css('min-height', (tallest+25) + 'px');
 }
-$('#similar-items-carousel img').load(normalizeHeights);
+$('#similar-items-carousel img').on('load', normalizeHeights);
 JS;
 ?>
 <?=$this->inlineScript(\Laminas\View\Helper\HeadScript::SCRIPT, $script, 'SET') ?>


demiankatz:
Thanks, @EreMaijala, this is definitely a needed fix... but I can't tell if it's actually doing anything. We recently had some conversations here about inconsistent thumbnail sizes, and @crhallberg fixed it locally with a CSS adjustment. I had hoped that fixing this bug would resolve the issue without needing special CSS, but as far as I can tell, it doesn't do anything. So I either misunderstand the purpose of this code, or there's some other bug in the normalizeHeights function. I also notice that this event triggers for every single image in the carousel; I wonder if there's a more efficient thing to do.

Bottom line: I'm happy to merge this into release-7.0 as a bug fix, but before I do, I'd like to hear from @crhallberg about whether larger changes are needed to the carousel code.

EreMaijala:
@demiankatz It was indeed just a code fix with no idea of whether it actually does something. As far as I can see the carousel could use some tweaking at least in bootstrap3 theme.

crhallberg (approved):

crhallberg:
I think I should take care of carousel updating in a separate pull request

demiankatz:
I agree, @crhallberg. The question is whether there's any point in merging this in the meantime -- it fixes a technical bug, but if the fixed code still doesn't do anything, it seems pointless. Should we merge this as an incremental step toward the final fix, or should we leave it open as a reminder for now and then close it as "superseded by" your carousel fix PR, when you have time to work on that?

crhallberg:
I think this incremental step should cover until I can fully investigate this issue, even if it's later this month.

demiankatz:
Thanks, @crhallberg, merging now. Should we open a JIRA ticket about the larger issue, or is this already firmly on your radar?

PR #1775, opened 2020-10-30
Fix for several problems found by PHPStan

mtrojan-ub:
https://github.com/phpstan/phpstan

... might be worth a look, there are lots of other problems left (undefined properties / methods / variables, doing stuff with methods that only return null and so on.)


demiankatz:
Thanks, @mtrojan-ub; I'd certainly be open to a PR that adds phpstan to our Travis process, similar to the work I've started to bring Psalm (another static analysis tool) into the process (#1764). There are a couple of these that I want to review a little more closely before merging, which I'll do as soon as time permits -- looks like it caught some significant problems, in any case!

demiankatz:
Okay, this all looks good. I've backported it to the release-7.0 branch since these are all bug fixes, and I've merged to dev from there.

PR #1776, opened 2020-10-30
Add a MarcReader class for lightweight handling of MARC data.

EreMaijala:
Initial results from working on a lightweight alternative for reading MARC record data. Retains the getMarcRecord method for back-compatibility.

A very simple benchmark of creating the class from a MARC record with 42 fields, 1000 iterations:
getMarcReader: 0,416 seconds
getMarcRecord: 1,927 seconds

I haven't yet tested everything I can think of, and I can't test SolrMarcRemote or SolrOverdrive.


module/VuFind/src/VuFind/RecordDriver/MarcReader.php
```diff
@@ -0,0 +1,495 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  RecordDrivers
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:plugins:record_drivers Wiki
+ */
+namespace VuFind\RecordDriver;
```
demiankatz: I don't think this belongs in the RecordDriver namespace. Should we create a new VuFind\Marc namespace? That would make it easier to factor this out into a separate project if we eventually want to reach a common version that is shared via Composer between VuFind and the RecordManager.

module/VuFind/src/VuFind/RecordDriver/MarcReader.php
```diff
@@ -0,0 +1,495 @@
...
+/**
+ * MARC record reader class.
+ *
+ * @category VuFind
+ * @package  RecordDrivers
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:plugins:record_drivers Wiki
+ */
+class MarcReader
```
demiankatz: There's a lot going on in this one class -- it reads and writes two different formats, plus it allows data extraction. I think it might be beneficial to consider abstracting the format-specific logic to separate classes so that it's easier to add additional encodings (such as MARC-in-JSON) as needed and so that it's easier to test discrete units of functionality. I have a few broad ideas of how we might be able to do that, but I haven't thought about it in great depth yet since I realize there may be performance trade-offs to consider. Let me know if you think this is worth discussing in greater depth!

themes/root/templates/RecordDriver/AbstractBase/export-marc.phtml
```diff
@@ -1 +1 @@
-<?=$this->driver->getMarcRecord()->toRaw()?>
+<?=$this->driver->getMarcReader()->toISO2709() ?: $this->driver->getMarcReader()->toXML() ?>
```
demiankatz: What is the purpose of the fallback here? I don't think outputting XML in this context will work right because of the associated HTTP headers.

demiankatz (commented):
Just sharing a few very preliminary thoughts. Have you investigated how much unit test coverage we have on the functionality currently leveraging File_MARC? It seems like adding tests and then confirming that they still work here might be a worthwhile project. I'm happy to help as time permits.

module/VuFind/src/VuFind/RecordDriver/MarcReader.php
EreMaijala: Makes sense, I'll make the changes.

module/VuFind/src/VuFind/RecordDriver/MarcReader.php
EreMaijala: I'll move out the format specific stuff. Can still be refined as needed.

themes/root/templates/RecordDriver/AbstractBase/export-marc.phtml
EreMaijala: There's a chance that the record can't be serialized in ISO2709. If a record has a field longer than 9999 characters or the record itself is more than 99999 characters, it doesn't fit in the constraints of ISO2709. We have some records like that. During testing I couldn't see this to cause other problems than the result being in unexpected format when exporting single records, but bulk export would indeed be problematic. We don't currently have a mechanism for handling the situation where the requested format cannot be provided after all. Throwing an exception or adding error handling that would check for an empty result could help. What do you think?

EreMaijala:
@demiankatz Current tests are only accessing a subset of used MARC handling functions. I'll add a couple of tests to improve this.

EreMaijala:
@demiankatz I added tests for the MarcReader class, but I'll post a separate PR for tests for the Marc traits.

demiankatz:
Thanks, @EreMaijala; note that some of the MARC trait logic is already tested through the \VuFind\RecordDriver\SolrMarc test class, so be sure to take a look at that so you don't waste time duplicating anything we already have covered. :-)

themes/root/templates/RecordDriver/AbstractBase/export-marc.phtml
demiankatz: Yes, I think throwing an exception could make sense here. Another option would be making the logic to detect which export methods are supported a bit more robust (though it may be too expensive to do a lot of analysis there).

EreMaijala:
Merged dev with the added tests.

EreMaijala:
Another completely unscientific benchmark: request 100 MARC results 10 times using curl as the normal /Search/Results page:

dev: 19.914s
marc-reader: 16.679s

Same results from the search API (/api/v1/search):

dev: 8.873s
marc-reader: 5.726s

Warmup done before to minimize the effect of cold Solr caches etc.

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -386,13 +386,32 @@ public function getBreadcrumb()
...
      *
      * @return \File_MARCBASE
      * @throws \File_MARC_Exception
      */
+    public function getMarcReader()
+    {
+        if ($this->getIsMarc()) {
+            return parent::getMarcReader();
+        } else {
+            return new MarcReader('<record></record');
```
demiankatz: Looks like there's a missing close angle bracket here.

module/VuFind/src/VuFind/RecordDriver/MarcReaderTrait.php
```diff
@@ -211,16 +241,13 @@ protected function getSubfieldArray($currentField, $subfields, $concat = true,
...
-                    // Grab the current subfield value and act on it if it is
-                    // non-empty:
-                    $data = trim($currentSubfield->getData());
-                    if (!empty($data)) {
-                        $matches[] = $data;
-                    }
+        foreach ($currentField['subfields'] as $currentSubfield) {
+            if (in_array($currentSubfield['code'], $subfields)) {
+                // Grab the current subfield value and act on it if it is
+                // non-empty:
```
demiankatz: This comment can probably be unwrapped.

module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
```diff
@@ -220,9 +216,12 @@ public function getBibliographyNotes()
      */
     public function getFilteredXML()
     {
-        $record = clone $this->getMarcRecord();
+        $record = clone $this->getMarcReader();
         // The default implementation does not filter out any fields
+        // $marc = new \File_MARCXML($record->toXML(), \File_MARCXML::SOURCE_STRING);
         // $record->deleteFields('9', true);
```
demiankatz: Does this $record need to be changed to $marc for the example to work, or am I misunderstanding what's going on here?

module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
```diff
@@ -635,29 +618,29 @@ public function getAllRecordLinks()
...
      * @return string
      */
     protected function getRecordLinkNote($field)
     {
         // If set, use relationship information from subfield i
-        if ($subfieldI = $field->getSubfield('i')) {
+        if ($subfieldI = $this->getSubfield($field, 'i')) {
             // VuFind will add a colon to the label, so prevent double colons:
-            $data = rtrim(trim($subfieldI->getData(), ':'));
+            $data = rtrim($subfieldI, ':');
```
demiankatz, line 632: Is there a reason you removed the internal whitespace trim? (I'm honestly not sure why that's needed in the first place... but just noticed it as a possible change in logic).

module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
```diff
@@ -725,17 +706,17 @@ protected function getFieldData($field)
...
                         'exclude' => $this->getUniqueId()
                     ];
                 }
                 break;
             case 'issn':
-                if ($issn = $field->getSubfield('x')) {
+                if ($issn = $this->getSubfield($field, 'x')) {
                     $link = [
-                        'type' => 'isn', 'value' => trim($issn->getData()),
+                        'type' => 'isn', 'value' => $issn,
```
demiankatz, line 721: Another eliminated trim.

module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
```diff
@@ -961,19 +939,16 @@ public function getCleanISMN()
...
-            if (!empty($subfields)) {
-                $nbn = [
-                    'nbn' => $subfields[0],
-                ];
-                if (isset($subfields[1])) {
-                    $nbn['source'] = $subfields[1];
-                }
+        $field = $this->getMarcReader()->getField('015');
+        if ($field && $nbn = $this->getSubfield($field, 'a')) {
+            $result = [
```
demiankatz: You could shave a couple of lines with `compact('nbn')` here.

module/VuFind/src/VuFind/Marc/MarcReader.php
```diff
@@ -0,0 +1,264 @@
...
+     * Return leader
+     *
+     * @return string
+     */
+    public function getLeader(): string
+    {
+        // Make sure leader is 24 characters. E.g. Voyager is often missing the last
+        // '0' of the leader.
+        return isset($this->fields['000'])
+            ? str_pad(substr($this->fields['000'], 0, 24), 24) : '';
```
demiankatz: Would it be better to do this normalization logic when setting data in the first place, so that it only needs to be applied once even if the leader is requested multiple times? Or is it actually more efficient to leave it as-is and only do it on demand?

module/VuFind/src/VuFind/Marc/Serialization/Iso2709.php
```diff
@@ -0,0 +1,185 @@
...
+     * Check if this class can parse the given MARC string
+     *
+     * @param string $marc MARC
+     *
+     * @return bool
+     */
+    public static function canParse(string $marc): bool
+    {
+        // A pretty naïve check, but it's enough to tell the different formats apart
+
```
demiankatz: Unnecessary blank line?

module/VuFind/src/VuFind/Marc/Serialization/MarcXml.php
```diff
@@ -0,0 +1,195 @@
...
+     * Check if this class can parse the given MARC string
+     *
+     * @param string $marc MARC
+     *
+     * @return bool
+     */
+    public static function canParse(string $marc): bool
+    {
+        // A pretty naïve check, but it's enough to tell the different formats apart
+
```
demiankatz: Same extra blank line.

module/VuFind/src/VuFind/Marc/Serialization/Iso2709.php
```diff
@@ -0,0 +1,185 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  MARC
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:plugins:record_drivers Wiki
+ */
+namespace VuFind\Marc\Serialization;
```
demiankatz, line 28: Might be worth defining a SerializationInterface both as a clue to help with future serialization implementations and to allow for stronger typing if we pass serializations around in the future (for now, I'm fine with keeping these static, but if we want this to be more extensible, we might rethink that so we can inject the serializations into the reader).

demiankatz (commented):
Thanks again for all of this work. I took time to review it in more detail, and left a few comments (mostly minor nitpicks). I'm also going to see if I can write some quick tests for the two record drivers that you are unable to test, and then perhaps you can merge them in here and make the necessary adjustments.

I think this should probably be targeted for 8.0, since it does change some of the record driver's internal API and thus could prove to be a backward compatibility break. Any objections?

demiankatz:
@EreMaijala, I've opened #1785 to create the necessary SolrMarcRemote/SolrOverdrive tests to ensure these changes are safe. I haven't had time to actually write any meaningful tests there yet, but I've at least started setting up the framework, so they can be filled in the next time I have available time.

module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
EreMaijala: getSubfield method already trims the value, so there's no need to do it again.

module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
EreMaijala: Same reason as above.

themes/root/templates/RecordDriver/AbstractBase/export-marc.phtml
EreMaijala: I added an exception and basic handling for it.

module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
EreMaijala: Yes, the example was borked.

module/VuFind/src/VuFind/Marc/MarcReader.php
EreMaijala: Yes. I moved the leader out of fields altogether, which made the handling a bit cleaner too.

module/VuFind/src/VuFind/Marc/Serialization/Iso2709.php
EreMaijala: Interface added. I thought about the capability to inject serializations, but since we always need ISO2709 and MARCXML (at least the export support even if the internal format was something else), I deemed it unnecessarily complex or full of magic for now. I also wanted to make sure we don't need to instantiate too many classes just to check which one we need.

EreMaijala:
@demiankatz Thanks for the comments. I believe I've addressed them now.

WRT serializations, I'm actually already considering support for storing MARC in MarcReader's array format, but I'm not sure that or MARC-in-JSON (which is actually very similar to the array format) would warrant adding support for injecting serializations. I haven't yet considered if it's worth it to store the array format in Solr since ISO2709 is so efficient.

EreMaijala:
@demiankatz Regarding back-compatibility, I tried to make sure everything stays compatible. What did I break?

I'm planning to take this and put it into action as soon as possible, but getting this merged is not strictly a requirement. I'll just force-push a squashed version that I can pick, if that's alright. However, if we don't merge this before 8.0, I'd consider getting rid of getMarcRecord method altogether.

module/VuFind/src/VuFind/Marc/Serialization/Iso2709.php
demiankatz: Makes sense!

demiankatz:
Thanks for the progress; this all looks great! A few comments and answers:

1.) As a compromise between fully injectable encodings and the current very-static approach, what if the MarcReader class had an array of encoding classes in it? Something like `['iso2709' => Iso2709::class, 'xml' => MarcXml::class]`. Then, the hard-coded `toXML()` and `toISO2709()` methods could be replaced with a generic `toFormat($format)` method, and `setData()` could replace the if-else with an iteration through the format classes. This has at least two advantages: first of all, it requires less "plumbing" to add new formats, and secondly, it opens up the option of easily disabling entire formats to improve efficiency if you know that you do not need to support them.

2.) Regarding the idea of removing the MarcRecord support entirely, I wonder if we should leave that in place indefinitely because some people are using it to edit records on the fly, and the MarcReader doesn't seem to allow that as easily. (Though maybe ease of editing is another reason to consider exposing the MarcReader's internal format as an encoding option).

3.) Regarding backward compatibility breaks in the internal API, I'm referring to things like `getSubfieldArray()`, which has had the first parameter changed from a File_MARC object to a MarcReader array. It absolutely makes sense to do that, but I'm almost certain that doing so is going to break some local custom code for anyone who has customized a MARC record driver. I think trying to avoid that type of BC break is more work than it's worth, but since it's going to require adjustments to local code, I'd be inclined to save it for 8.0 when that kind of change is more likely to be expected.

I'll try to find some time today to finish up #1785 so we can complete the testing of all the affected logic, and then I think we'll be very close to wrapping this up. Given the amount of work we're doing in anticipation of 8.0, I'm also happy to consider creating a dev-8.0 branch so we can start to merge some of these things and clean out the PR queue if you think that would help.

EreMaijala:
@demiankatz 

1.) Makes sense, I'll revise.

2.) Sure, no objection.

3.) You're right about getSubfieldArray, of course.

themes/root/templates/RecordDriver/AbstractBase/export-marc.phtml
```diff
@@ -1 +1,7 @@
-<?=$this->driver->getMarcRecord()->toRaw()?>
+<?php
+$marc = $this->driver->getMarcReader()->toISO2709();
```
demiankatz: I think you missed this `toISO2709()` call in your refactoring; might want to double check if there are any others.

EreMaijala: Now fixed. I think I got the others right.

EreMaijala:
@demiankatz The dev-8.0 branch certainly makes sense to me also because sooner or later we'll have to sort conflicts and it easier to do it incrementally.

demiankatz:
I agree, regarding dev-8.0. I'll get that started once this is ready to merge (soon!).

demiankatz:
@EreMaijala, I've finished up #1785 and merged it here. All tests still seem to be passing. Is there anything outstanding from your perspective, or should I go ahead and merge to dev-8.0?

EreMaijala:
@demiankatz I'm ready for merge. I'll probably post a few follow-ups for using MarcReader also in ILS drivers that handle MARC records and perhaps other places too, but that's intentionally left for later.

EreMaijala:
@demiankatz Oh, and thanks for adding the tests. Good stuff!

demiankatz:
Glad to help, @EreMaijala. I've established a dev-8.0 branch and merged this there.

PR #1778, opened 2020-11-03
KohaRest: Request all results when endpoint uses paging

EreMaijala:
I had not taken into account that the endpoints that allow paging have a default page size in Koha. By default it's 20, so chances were some records would have been left out.


demiankatz:
Thanks, @EreMaijala; since this is a bug fix, I've backported it to release-7.0.

PR #1779, opened 2020-11-03
Improve/update citation generation

demiankatz:
TODO
- [x] Remove place of publication from APA/MLA citations (standards no longer require this)
- [x] Include information about which edition of each citation standard is being used
- [x] Fix problem with years showing up inappropriately when commas are missing ([example](https://library.villanova.edu/Find/Record/1069504))
- [x] Fix problem with extra punctuation after first MLA/Chicago author name ([example](https://library.villanova.edu/Find/Record/1693315))
- [x] Include DOI for APA eBooks.
- [x] Display full page range if possible for all citation formats.
- [x] Do not use p. or pp. in APA citations.
- [x] Include DOI for Chicago-style citations.
- [x] Use "et al" for MLA citations with 3+ authors.
- [x] MLA citations should use p. / pp. prefix before page ranges.
- [x] Await Villanova librarian feedback
- [x] Adjust `<span>` to `<i>` where appropriate


crhallberg:
Unless there's a reason for it that I can't recall, we can replace the `<span style="font-style:italic"></span>` with `<i></i>`.

demiankatz:
@crhallberg, the only reason I can think of for this approach would be to ensure that the citations always use italics, even if `<i>` has been re-styled to look different for some reason... but that seems fairly unlikely to ever happen.

crhallberg:
That was the only thing I could think of. I'm not going to it but if anyone else in the community wants to change it, that's fine.

demiankatz:
Thanks, @crhallberg! I'll wait for the librarian feedback on our end to come through, but this seems like a safe adjustment I can make along with any other requested changes, once we hear back.

PR #1780, opened 2020-11-04
Change default content type of ajax responses to application/json

xmorave2:
According to https://www.ietf.org/rfc/rfc4627.txt


demiankatz:
Thanks, @xmorave2. That `application/javascript` has its roots in old VuFind 1.x code and has been faithfully ported forward since then. I suspect at the time it was written, `application/json` may not have been as well-documented or well-established (though the RFC you site had already been released, so maybe there's no excuse). In any case, this seems like an appropriate correction to make now! I've confirmed that it has no impact on the test suite.

PR #1782, opened 2020-11-05
Deprecate author_fuller / author2_fuller index fields

demiankatz:
A review of the code suggested that we may have some author indexing issues in need of improvement. I am opening this PR to discuss and experiment.

Specifically, indexing the fuller forms of author names to separate author_fuller / author2_fuller fields makes very little sense. While it makes the words searchable, it causes strange displays in some contexts. It seems far preferable to simply use the full MARC name headings everywhere for consistency.

TODO
- [x] Make sure the test records include an example of an author with a $q subfield; test using that author name (yes: record ID vtls000000329 in authoritybibs.mrc is an existing example for author_fuller; record ID testsample5 in journals.mrc is an existing example for author2_fuller).
- [x] Make sure citations generate appropriately for authors with $q
- [x] Evaluate whether author_variant / author2_variant serve a useful purpose (they do: used for supporting search by initials)
- [x] Document need to reindex in changelog; make note to remove deprecated fields in a future release.


demiankatz:
This seems to be a clear improvement based on my own testing. I'm going to put out a call for feedback, and will merge this in a week or two if no one objects.

EreMaijala:
@demiankatz I agree that author_fuller as it has been isn't very useful. I also agree that putting everything in author/author2 makes sense. However, this has the side-effect of polluting the author facets with all the different forms there might be. I'd recommend to stop using copyField in the schema and instead construct the facet fields separately in the indexer. Locally we do that, and it allows much better control in the indexer for what gets put in each field. Getting rid of copyField would also allow one to customize the facet content locally without modifying the schema.

demiankatz:
@EreMaijala, I actually don't think this has the side effect that you think it does. This will not add any additional values to any fields. It just adds more information to the existing values. So, for example, in the old way, an author like:

<pre>
100 $a Person, A. S. $q (Arthur Sample), $d 1910-1980.
</pre>

Would get indexed like:

<pre>
author = Person, A. S. 1910-1980
author_fuller = (Arthur Sample)
</pre>

And now he'll get indexed like:

<pre>
author = Person, A. S. (Arthur Sample), 1910-1980.
</pre>

Obviously this latter approach is better -- especially because the previous approach not only omitted useful information but also messed up punctuation, and sticking the fuller version of the name in a separate field made it searchable but not in a particularly useful or meaningful way.

I hear what you're saying about the copyFields, but I wonder if what we really need is a better way to configure/manage the Solr schema rather than a different default configuration. The advantage of using the copyFields is that it reduces the amount of work the indexer has to do. In the case of authors, the indexing logic is already complicated, and repeating that work for every record would likely have a noticeable performance impact when you spread it across millions of records. That certainly may be worth it in some situations, but if it's not necessary for the majority of users, I think the copyFields are more efficient as a default. In any case, I think that's a matter that should be discussed in a separate place. Do you think it's worth opening a ticket about investigating Solr's schema management API? VuFind is still using the old-fashioned way of providing an XML Solr schema in its default setup, but I wonder if a better approach for the future would be to load that in through the API and then provide tools so that people could dynamically change the schema. This might be the answer to the long-unsolved question of how to manage Solr schema customizations without changing the core VuFind distribution. Of course, as with many things, I have nowhere near the necessary free time to figure this out right now, and the existing approach still works -- but if we create a ticket about it, perhaps somebody will get to it eventually!

EreMaijala:
@demiankatz Right, but author gets copied to author_facet as well. Before the change, if there were two records with the following 100 fields:
`100 $a Person, A. S. $d 1910-1980.
`
`100 $a Person, A. S. $q (Arthur Sample), $d 1910-1980.
`
They both would have been copied to author_facet as "Person, A. S. 1910-1980", right? After the change there would be two facet values, "Person, A. S. 1910-1980" and "Person, A. S. (Arthur Sample), 1910-1980". In addition to the latter being pretty long, I don't think it's desirable for the two to get split.

demiankatz:
I would argue that if you have different headings for the same person in different records, that is a metadata/authority control problem, and not something that VuFind's indexer should be trying to sort out. And while I also agree that using the full name heading is long and not especially attractive, I don't think it's VuFind's place to make editorial decisions about the MARC standard. The heading is really designed to be displayed as a whole single string, and arbitrarily choosing to edit out one part of it causes as many problems as it solves. Again, I'm not defending MARC design here -- I think it's absolutely hideous how punctuation is mixed with data, because it makes these types of things unnecessarily inflexible -- but given these limitations, it seems easier to yield to them than to fight them, since reformatting author names properly would require extra logic beyond just skipping subfield q.

And also, to go back to your example, if we really had two versions of the heading with and without subfield q in real records, they would actually look like this:

<pre>
100 $a Person, A. S., $d 1910-1980.
100 $a Person, A. S. $q (Arthur Sample), $d 1910-1980. 
</pre>

Note that in the first line, the comma is at the end of subfield a, and in the second example, the comma is at the end of subfield q. This is the punctuation issue that drives me crazy! As a result, this would index in the old code as:

<pre>
author = Person, A. S., 1910-1980
author = Person, A. S. 1910-1980
</pre>

...once with a comma, and once without. You'd still have two different values in the index, but it would be even more confusing to look at. So I still think, even in this situation, the change is actually an improvement, since it makes it much more obvious why there is an inconsistency and which records need to be fixed to restore uniformity.

EreMaijala:
@demiankatz We're probably looking at this from a different perpective, and I'm fine with this when it's a conscious decision. And for a single catalog it makes sense. I also stand corrected about my example being wrong.

We're not affected, and we're already doing it differently. That's partly due to the fact that we try to cater for different metadata formats, such as Dublin Core, and what's the (perhaps lowest) common denominator for them (i.e. `LastName, FirstName`). We also have very little control over the metadata that comes from a myriad of different sources.

To make SolrMarc author handling more efficient, perhaps a relator rule set based caching could help: first collect a list of all author fields that match the rules and cache that. Then go through the results and collect requested subfields. When another call for same rules but different subfields comes in, only the latter part would need to be done. The cached field lists would get flushed if the record changes.

demiankatz:
Absolutely, @EreMaijala -- I appreciate your input, and agree that we're coming at this from different perspectives. What I'm hoping we can agree on (in order to get this merged and moved forward) is this:

1.) The current MARC import rules are designed to take advantage of several nuances of the MARC standard, and to index names in a fairly complete way. Omitting subfield $q from author names, while keeping all the other details, was a strange exception to the rule that led to inconsistent/incorrect behavior. Incorporating it back into the author name allows the MARC data to be displayed as originally intended. This may cause problems when mixed with less rich data from other sources, but even without the change, such problems were likely to exist, and making the adjustment will definitely improve behavior in an all-MARC index.

2.) In the case of a "least common denominator" situation, some different custom author indexing code would need to be written to simplify author names to a more uniform format (accounting for punctuation and the meanings of various MARC subfields). And, of course, such simplification would inevitably lead to ambiguities when multiple individuals share the same name -- trade-offs would have to be considered.

Regarding your relator rule cache idea, I agree that something like that would work -- I'm just not familiar enough with SolrMarc internals to know (without doing research) exactly how the scoping and persistence works in order to implement it correctly. It's probably possible, though -- just another thing I don't have time to dig deeper into right now. :-)

EreMaijala:
@demiankatz I'm ok with merging this as is. And I'm not advocating spending time with making indexing faster especially since I'm not going to devote time for that myself. :) I just wanted to jot down the idea I had. Judging from what I did with the work keys it seemed that the scoping would allow for easy caching.

1.) Agreed.

2.) My only gripe is that you have to either change the Solr schema or index an additional field if you want to do something different with the facets. But that's probably an acceptable tradeoff if the other option is slowing down indexing or making it more complicated.

demiankatz:
@EreMaijala, regarding point number 2, I also think that the generation of author facets is potentially a whole separate discussion, and merging this as it is doesn't prevent us from having that conversation later. I'm certainly not trying to shut that one down. But the use of copyFields in the schema is also potentially a broader conversation as well. It does stick out as an inconsistency that we explicitly index most facet fields, but author is handled with copyFields ... but there are also some similar inconsistencies regarding the fields used for browse indexing. If we can find time for a comprehensive review and discussion, I'm always open to that -- but shortage of time seems to be a common theme in my comments these days. I'm less inclined to change things when I don't feel like I have the free time to give them the proper thought first. For something like this, we really need a few more people to step up and join in the process. In the meantime, I always welcome JIRA tickets if there are particular issues you'd like to capture -- I think there are some ideas in this conversation that we might want to capture in a more visible/permanent location.

EreMaijala:
@demiankatz I'll add a JIRA ticket about caching.

demiankatz:
Thanks, @EreMaijala!

PR #1786, opened 2020-11-06
Update dotdotdot jQuery plugin

crhallberg:
Update dotdotdot jQuery plugin to v4.1.0 to fix console error.


demiankatz:
@crhallberg, I notice that this update eliminates the copyright/version/license notice. Is there a version we can use that puts that back in?

demiankatz:
@crhallberg, it looks like the license on this plugin has changed (see [the website](http://dotdotdot.frebsite.nl)), and now has a requirement for payment for commercial use. This could create problems for some VuFind users; I think we need to find an alternative to this component. Given the trend to move away from jQuery, there are probably other good reasons to rework this. What do you think?

PR #1787, opened 2020-11-09
Replace dotdotdot with vanilla JS.

crhallberg:
The jQuery plugin [dotdotdot](https://dotdotdot.frebsite.nl/) changed their licensing, making it a problematic dependency for VuFind. I've replaced dotdotdot with a small vanilla function that performs the same line-clamping utility in Channels.


themes/bootstrap3/js/channels.js
```diff
@@ -59,6 +59,16 @@ VuFind.register('channels', function Channels() {
     }
     record.data('bs.popover').options.content = html;
   }
+
+  // Truncate lines to height with ellipses
+  function clampLines(el) {
+    let words = el.innerHTML.split(" ");
+    while(el.scrollHeight > el.offsetHeight) {
+      words.pop();
+      el.innerHTML = words.join(" ") + "...";
```
xmorave2: I think it should be an ellipsis character (U+2026), not three dots.

demiankatz:
@crhallberg, it also appears that there is an eslint issue here. Can you investigate?

themes/bootstrap3/js/channels.js
crhallberg: I checked using "…" with several fonts didn't see any notable difference in the way the ellipses were displayed or the number of words that were truncated. I'm not against the idea though!

demiankatz: I'd support using … -- if nothing else, it's one character instead of three. :-)

ckaz:
Sorry for chiming in just like that: A closely related issue, which could possibly be fixed in the same go is that in German, ellipsis will be inserted right after any word (without space) only when it is used to show that part of the word is ommitted -- in all other cases, there needs to be a non-breaking space after any word to show that part of the sentence is ommitted. There may be differences in other languages too.

demiankatz:
Thanks, @ckaz, this is a good point. Maybe we should create a new translation string like `eol_ellipsis` (or a better name, if anyone has any good ideas). It would not be too hard to integrate translation into this new code, which would be another improvement upon the original library. What do you think, @crhallberg?

EreMaijala:
Heh, for Finnish it's "never a space before ellipsis". Having an eol_ellipsis or something like that would make me happy.

crhallberg:
I went ahead with this because it also addresses the preference for the Unicode ellipses character vs. three periods.

themes/bootstrap3/js/channels.js
```diff
@@ -86,8 +96,6 @@ VuFind.register('channels', function Channels() {
     $(op).on('swipe', function channelDrag() {
       switchPopover(false);
     });
-    // truncate long titles and add hover
```
demiankatz, line 89: Should the "add hover" part of the comment still be retained here?

demiankatz:
Thanks, @crhallberg... I just left a comment above. Also, it looks like the language files are still using three dots instead of the ellipsis character. Can you adjust that, or am I misinterpreting what I'm seeing? Also sounds like you can make the fi.ini version of the new string the same as the en.ini version based on @EreMaijala's comment.

themes/bootstrap3/js/channels.js
crhallberg: Add hover was out of date before this change

demiankatz: Ahh, right, the popover thing is not a hover. :-)

crhallberg:
I used three periods because I wasn't hearing a preference but since you asked it's now a Unicode ellipses.

demiankatz:
This looks good to me now, so I am merging it. Thanks, everyone!

PR #1789, opened 2020-11-11
Disable deduplication listener when search parameters contains child filter

xrosecky:
Disable deduplication listener when search parameters contains child filter.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Solr/DeduplicationListener.php
+++ b/module/VuFind/src/VuFind/Search/Solr/DeduplicationListener.php
@@ -3,6 +3,8 @@
 /**
  * Solr deduplication (merged records) listener.
  *
+ * See https://vufind.org/wiki/indexing:deduplication for details on how this is used.
+ *
  * PHP version 7
  *
  * Copyright (C) Villanova University 2013.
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks, @xrosecky. I'm asking for @EreMaijala's review of this, since he's the expert on the DeduplicationListener. I also wonder if it might be useful to add a comment to the new method mentioning that merged_child_boolean comes from RecordManager, since it is not used or mentioned anywhere else in the VuFind code, which might cause confusion in the future.

EreMaijala (approved):
Looks good to me, though I'd like to know more of the use case for this. Perhaps a comment referring also refer to the wiki page at https://vufind.org/wiki/indexing:deduplication would be what's needed to clarify this? Perhaps even at the top of the file?

xrosecky:
Our use case is special API for searching in child documents.

demiankatz:
Good idea, @EreMaijala. I've added a comment to the top of the file. @xrosecky, do you want to make any further adjustments, or is this now ready to merge?

xrosecky:
It is ready to merge.

demiankatz:
Thanks! Merging now.

PR #1791, opened 2020-11-13
Configurable truncation of subjects / items on full record display

bbusenius:
## Problem ##
Some of our bib records have a long subject display that pushes critical important item information too far down on the page. In order to make item availability information more visible we need to truncate the subject display by default. Users should be able to expand the listing if they so desire.

This pull requests allows VuFind to be configured with a default number of subjects to show before truncating them. The subject listing can be expanded and collapsed by the user in a similar fashion to sidebar facets. By default, it is configured to display all subjects.

When a number is set in the config, the listing is truncated like this:

image

Notice that there is a "more..." link in the subjects display. Clicking on this expands the list to show all facets and creates a "less..." link that can be used to collapse the display.

## TODO ##
- [x] Complete/incorporate work from https://github.com/demiankatz/vufind/pull/1
- [x] Thoroughly test facet truncation functionality now that it has been updated to use this new mechanism


demiankatz:
Thanks, @bbusenius! I'll try to give this a review next week. I'm also interested to hear what @crhallberg thinks.

bbusenius:
Thanks @demiankatz! Yes, any and all feedback is welcome. I've finally been given time to work through some of the customizations in that spreadsheet you helped us with awhile back. If you like this, I could do something similar for the items display next.

config/vufind/config.ini
```diff
@@ -163,6 +163,10 @@ allowSavedSearches = true
 ; still require Javascript; this simply improves compatibility for certain
 ; features (such as display of hierarchies).
 nonJavascriptSupportEnabled = false
+; Only display x number of subjects in the bib display on the full record by default.
+; Additional subjects are hidden by default and expandable via a "more..." button.
+; Enable by setting to a number, e.g. 3, to display 3 subjects and hide the rest.
+subjectLimit = false
```
demiankatz: Would it make more sense to associate this with the search object (i.e. put in searches.ini or other equivalent configurations)? I think the answer is "no" because it is more likely that this would be desired globally -- but just playing devil's advocate and asking whether there will ever be a situation where we'll want different limits for different search backends.

module/VuFind/src/VuFind/View/Helper/Root/Config.php
```diff
@@ -88,4 +88,15 @@ public function ajaxCoversEnabled()
...
+
+    /**
+     * Should we limit the number of subjects displayed on the full record?
+     *
+     * @return int or float (INF)
+     */
+    public function subjectLimit()
+    {
+        $subjectLimit = $this->get('config')->Site->subjectLimit;
+        return $subjectLimit ? (int)$subjectLimit : INF;
```
demiankatz: The INF constant is neat, but I wonder if PHP_INT_MAX would serve the same basic function while making it easier to have a consistent return type on the function.

themes/bootstrap3/js/record.js
```diff
@@ -289,6 +289,24 @@ function applyRecordTabHash() {
 
 $(window).on('hashchange', applyRecordTabHash);
 
+function moreLessSubjects() {
+  $('.more-subjects, .less-subjects').off('click');
```
demiankatz: Is there a strong reason to tie this specifically to subjects? Would it be better to use a more generic class so this could be replicated to other fields without requiring additional JS code?

demiankatz (commented):
I just left some food for thought...

config/vufind/config.ini
bbusenius: Interesting idea, I certainly wouldn't have thought of this. I think you're correct in that technically it's possible we might want it on some things and not others. It solves a problem for us in the immediate case, however, when searching another data type (e.g. something without availability or less metadata), it might not be needed. I'm still inclined to agree with your, "no" assessment for the following reasons:

1. Even if long subject listings caused a problem on one search backend and not another, I don't think it would harm the other and we might not care so much to configure them differently.
2. Consistency of UI.

But I don't feel strongly about this. We could still configure it the same if it were in `searches.ini` and it's easy enough to change. I'll defer to your judgement.

themes/bootstrap3/js/record.js
EreMaijala: Since we use locally a truncate-field class for several fields, I'd support making this more general. We also use a data-rows attribute to optionally indicate the number of lines to show by default. I'm not implying these should be the attributes to use, but something like it might allow us to move to this mechanism, which would be great!

themes/bootstrap3/js/record.js
```diff
@@ -289,6 +289,24 @@ function applyRecordTabHash() {
...
+  $('.more-subjects').click(function moreSubjects() {
+    $('.subject-hidden').removeClass('hidden');
+    $('#less-subjects').removeClass('hidden');
+    $('#more-subjects').addClass('hidden');
+    return false;
+  });
+
+  $('.less-subjects').click(function lessFacets() {
```
EreMaijala: There could be a more appropriate name for the function. :)

EreMaijala:
Well, now that I look at it, our approach is js-only to avoid putting too much logic in the templates, so maybe we'll keep our current way, but this could be a great alternative for some situations.

config/vufind/config.ini
demiankatz: I'm happy to leave it as it is for now, since it is more straightforward. We can always add another layer later if a compelling need arises.

themes/bootstrap3/js/record.js
demiankatz: I have received occasional complaints from people who think the more/less controls should be more/fewer controls. Depends on the grammatical context you mentally place around the word. ;-) But I agree that maybe function names containing show or hide would be more informative.

demiankatz:
I gather that @bbusenius has modeled this on the way we show and hide facets, which makes sense as a starting point... but maybe that's not the best model. As @EreMaijala points out, the template logic may be unnecessarily complicated, and a pure-JS approach could improve readability. Since the current implementation already requires Javascript, switching to all-JS wouldn't break anything, it would just move the logic to a single place and clarify everything. I know that @crhallberg has been talking about rewriting the facet templates, and this could be a part of that. Maybe the logical thing to do is to build a solid generic, all-JS mechanism here first, and then adapt the facets to use it as a subsequent step.

Sorry to redirect @bbusenius' work in a significantly different direction, but I think this is a good opportunity to think about such things.

bbusenius:
Yes, I just modeled this off what I saw in the facets. We do this in a few places and I know we use `data-` attributes in at least one of them. I'm happy to work on a pure Javascript version if that works better for everyone. I'll start looking at it

demiankatz:
Thanks, @bbusenius. Perhaps @EreMaijala can also provide a link to an example from the finna implementation, in case that saves you any trouble.

EreMaijala:
Yes, certainly. Here's the JS function: https://github.com/NatLibFi/NDL-VuFind2/blob/dev/themes/finna2/js/finna-layout.js#L77. 

You may find it surprisingly long, but it caters for a couple of different situations. There are a few things that could use some cleanup:

- The two hide handlers could be combined.
- I'm not sure if we actually use data-label and even if we did, I think it would be better to have data-more-label and data-less-label.
- data-row-height is not used anymore. Earlier versions had some trouble calculating correct row height, and that's a leftover from those times.

Here are a couple of examples on how it's used:

https://github.com/NatLibFi/NDL-VuFind2/blob/dev/themes/finna2/templates/RecordDriver/DefaultRecord/data-allSubjectHeadings.phtml#L3
https://github.com/NatLibFi/NDL-VuFind2/blob/dev/themes/finna2/templates/ajax/status-full.phtml#L154

bbusenius:
Thanks @EreMaijala this is helpful! I'll use it as a starting point.

bbusenius:
OK, I have an initial implementation of this. First of all @EreMaijala's code was super helpful to look at. It actually slotted right in and worked well as it was. That being said, I did change a few things. I'll outline the bigger changes here.

1. I added a configurable visibility option for items in the holdings tab. I was going to do this as a separate pull request in the future but since we wanted to make this function generic and usable in a wide range of situations, I thought it best to try it on something different.

2. This version uses classes to toggle the visibility instead of calculating height. I made this change because as I tried to apply it to items, I found that the first row height was often not consistent with other item row heights. This was only because at first I was targeting all `<tr>`s in the holdings table and not just the item related ones, however, it made me realize that this would likely be a problem on our institution's VuFind, since I'm pretty sure there are cases where our row heights might be different, especially on smaller screens. I thought it might be more resilient to not rely on height.

3. In my latest commit, I removed the button top functionality that was present in @EreMaijala's version. Though I had initially rebuilt this with my class based version (see previous commit) and it was working fine, I realized that in our institutions version, we put insert the show/hide buttons as siblings at the point of truncation. I don't think this is a better approach necessarily but I thought I'd show it for the sake of discussion. This approach has the benefit of being a little simpler and keeping the show/hide buttons in the same place. When a user toggles, the button isn't pushed down.

image

I don't have a strong preference. I just wanted to show it as an alternative. If you guys prefer the other option (with the hide button always at the bottom and sometimes at the top), I can revert to that version very easily.

Those are the big changes. I'll make any other comments inline.

demiankatz:
Thanks, @bbusenius! I'll take a closer look at this when I return after the Thanksgiving holiday. I'd also be interested to hear @crhallberg's thoughts on the UI decisions/implications.

crhallberg:
Hello! Looks like I have some catching up to do!

- I really like the JS-only approach where data attributes can be used to label or tweak behavior. As long as the new templates are atomic enough for easy customization this is the approach I'll be using with the new theme.
- I'm glad that you've already switched to a button, which is the more accessible and intuitive option. It should be easy to [make the button smaller with Bootstrap](https://getbootstrap.com/docs/3.4/css/#buttons-sizes).
- Does this break the chain highlighting? Hovering any of the links should also underline the items to the left in the chain.
- As far as the button being an in-place toggle, would it be worth trying a dropdown element? More generic for the JS-only approach but maybe less flexible.

bbusenius:
Thanks for the feedback @crhallberg! I agree about button size. I think that since this could be used in different contexts, I'll just go ahead and make the button size configurable via an optional `data-` attribute.

Chain highlighting still works as far as I can tell. I think we're OK there.

Regarding the dropdown element, [can you confirm that you're talking about this](https://getbootstrap.com/docs/3.4/components/#dropdowns)? I still have a slight preference for a button but I couldn't tell you why. I'm going to ask some of our people what they think about that. I don't feel strongly about it, so if you think that would be best, I'm happy to change it. Would we still be able to configure the size of it if we went that way?

bbusenius:
I just talked to our designer, UX person and someone else and they all preferred the toggle as a button. So, I guess that would be our vote. Nobody said a dropdown would be a deal breaker but they prefer the button.

bbusenius:
I think at this point my biggest question is, will this version work for @EreMaijala? Please let me know if there's anything I can do to make it more compliant with your institution's needs.

themes/bootstrap3/js/record.js
```diff
@@ -298,6 +298,56 @@ function removeCheckRouteParam() {
   }
 }
 
+function initTruncate(_holder, _target, _fill = function(m) { return m } ) {
+  var holder = typeof _holder === 'undefined' ? $(document) : $(_holder);
+  var target = $(_target);
+  var targetElemName = target.prop('tagName').toLowerCase();
```
EreMaijala: This will throw an error if target is empty. Needs a check for target.length or something.

themes/bootstrap3/js/record.js
```diff
@@ -298,6 +298,56 @@ function removeCheckRouteParam() {
   }
 }
 
+function initTruncate(_holder, _target, _fill = function(m) { return m } ) {
```
EreMaijala: If I got everything right, _holder is treated as possibly multiple containers, but _target seems to specify the fields for a single container. I think it would make sense to change that so that _holder is always just a single container. Otherwise _target doesn't make sense. Maybe rename this to `truncateField` that handles just a single container? Then we could wrap that into a loop to handle all the .truncate-field containers we have.

themes/bootstrap3/js/record.js
```diff
@@ -298,6 +298,56 @@ function removeCheckRouteParam() {
...
+  var holder = typeof _holder === 'undefined' ? $(document) : $(_holder);
+  var target = $(_target);
+  var targetElemName = target.prop('tagName').toLowerCase();
+  var rowCount = holder.data('rows');
```
EreMaijala: rowCount should have a default (3 would be great :) ) if data-rows is not defined.

themes/bootstrap3/js/record.js
```diff
@@ -343,6 +393,8 @@ function recordDocReady() {
     backgroundLoadTab(el.dataset.tab);
   });
 
+  initTruncate('.truncate-subjects', $('.subject-line'));
```
EreMaijala: As it currently stands, I think the second parameter should be `$('truncate-subjects > .subject-line')` or something like that. However, if the second parameter is always a selector for limiting elements inside the container, how about making it a selector like `'.subject-line'` and then in the actual method do `var target = holder.find(_target);` ?

themes/bootstrap3/js/record.js
```diff
@@ -298,6 +298,56 @@ function removeCheckRouteParam() {
...
 }
 
+function initTruncate(_holder, _target, _fill = function(m) { return m } ) {
+  var holder = typeof _holder === 'undefined' ? $(document) : $(_holder);
+  var target = $(_target);
+  var targetElemName = target.prop('tagName').toLowerCase();
+  var rowCount = holder.data('rows');
+  var moreLabel = holder.data('more-label') ? holder.data('more-label') : VuFind.translate('show_more');
+  var lessLabel = holder.data('less-label') ? holder.data('less-label') : VuFind.translate('show_less');
+  var btnSize = holder.data('btn-size') ? ` ${holder.data('btn-size')}` : '';
```
EreMaijala: What are the backticks here? I've never seen this construct before, so just asking.

EreMaijala (changes requested):
@bbusenius Thanks for the consideration. :) I tried to apply your version but ran into a couple of small issues noted in review comments. All in all, I think with small changes we could handle everything but the leading "show less" button. Maybe it could be togglable with a data attribute? While I agree it doesn't work for all situations, there are cases like super long subject lists that could use the top button especially on mobile.

bbusenius:
@EreMaijala I made all of the changes except for one but I'm still open to discussing it. Regarding the leading "show less" button, I had a configurable version of that in a previous commit, so it was easy to restore. But since you need it, I realized that the "toggle in place" functionality should also be configurable via a data attribute. So there is now a `data-top-toggle` and a `data-in-place-toggle`. This gives us more configurability and will probably cover a wider range of circumstances. These could even be combined if you wanted a leading show less button and still have the original button toggle in place (instead of attaching to the bottom). You'll see that I setup the subjects to display with a leading show less button and _not_ toggle in place by default. I setup the items display to toggle in place with no leading show less button by default.

EreMaijala:
@bbusenius Thanks, I'll check them out as soon as possible.

themes/bootstrap3/js/record.js
crhallberg: Backticks denote a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals), a string that allows for easier formatting with ${var}, similar to PHP.

themes/bootstrap3/js/record.js
```diff
@@ -298,6 +298,74 @@ function removeCheckRouteParam() {
...
+      if(i > truncateAfter && shouldTruncate) {
+        $(this).hide();
+        $(this).addClass('truncate-toggle');
+      }
+    });
+
+    if (shouldTruncate) {
+      var btn = '<button type="button" class="more-link btn' + btnSize + '">' + moreLabel + ' <i class="fa fa-arrow-down" aria-hidden="true"></i></button><button type="button" class="less-link btn' + btnSize + '">' + lessLabel + ' <i class="fa fa-arrow-up" aria-hidden="true"></i></button>';
+      var btnLessTop = '<button type="button" class="less-link-top btn' + btnSize + '">' + lessLabel + ' <i class="fa fa-arrow-up" aria-hidden="true"></i></button>';
+      var btnWrapper = $('<' + targetElemName + ' class="more-less-btn-wrapper' + targetClass + '"></' + targetElemName + '>');
```
crhallberg: I don't have any issues with this section but the community should have a discussion on how to handle sub element generation in the future. A few options include pulling in PHP templates, [HTML5 templates](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_templates_and_slots), or custom web components.

themes/bootstrap3/js/record.js
```diff
@@ -298,6 +298,74 @@ function removeCheckRouteParam() {
...
+        $(this).hide();
+        self.find('.less-link').show();
+        if (hasTopToggle) {
+          self.find('.less-link-top').show();
+        }
+        self.find('.truncate-toggle').toggle();
+      });
+    }
+    self.addClass('truncate-done');
+    self.trigger('truncate-done', [self]);
```
crhallberg: I might recommend using or at least also adding a `VuFind.emit('truncate-done', self);` to tie into the event system in the [VuFind core js](https://github.com/vufind-org/vufind/blob/fbf5028bfa261f5a653e5e54f114447d7d827d46/themes/bootstrap3/js/common.js#L14).

crhallberg (approved):

themes/bootstrap3/js/record.js
EreMaijala: I'm not sure if the event is actually needed. I don't think we're currently using it.

EreMaijala:
@bbusenius I placed a pull request (https://github.com/uchicago-library/vufind/pull/105) on your branch with a change similar to what I said I think needs to be done. It just changes the loop so that all calculations and attribute retrievals are done for each container separately.

themes/bootstrap3/js/record.js
EreMaijala: I think something needs to be done with the backticks since eslint doesn't let them through. Either change the rules or avoid backticks?

bbusenius:
Thanks @EreMaijala. I didn't fully understand. I merged it and it works fine. Thanks for putting that work in!

demiankatz:
There's been a lot of conversation here, and I haven't had time to keep up with all of it. A couple of general questions:

1.) @EreMaijala, do you have any outstanding concerns at this point? I've resolved some of your comments that have obviously been fixed, but I didn't have time to look into all of them in detail.

2.) I notice that eslint is still failing. Is this entirely related to the template literal backticks, or are there other issues that need to be adjusted? Regarding the backticks, it looks like [support is missing from older browsers like IE11](https://caniuse.com/template-literals), so I'm not sure whether we need to worry about that. Thoughts, @crhallberg?

EreMaijala:
@demiankatz Should be all good. @bbusenius removed the backticks, so that should no longer be an issue. I don't know what's up with Travis, however, since it seems to point to wrong lines in record.js, but I'll post a PR for @bbusenius' branch to fix the tests.

EreMaijala:
This should fix the tests: https://github.com/uchicago-library/vufind/pull/106

bbusenius:
Thanks @EreMaijala!

demiankatz:
Thanks, everyone! A couple of questions:

1.) Does anyone object to targeting this for release 8.0? Since the translation work is already underway for 7.1, it's too late to introduce anything that adds new language strings.

2.) If we're going to target 8.0 anyway, should we go ahead and revise the existing side facet code to use this new mechanism, so we can perhaps eliminate some redundant code and make everything more consistent?

I haven't had a chance to do hands-on testing of this yet, but if there's a hurry to get it merged somewhere, I can make time for that and then merge it to the dev-8.0 branch if all is well... but if it's better to get all of the work (including 2 above) finished first, I'll wait for more progress before diving in deeper.

bbusenius:
@demiankatz, regarding number 1, from UChicago's point of view, we weren't expecting this before the 8.0 release. Sooner is always a nice bonus but it's not expected or needed. Regarding number 2, I don't have a strong opinion, however, I can imagine it would be nice to try and use it in more places to ensure that it's as flexible as it needs to be.

demiankatz:
Thanks, @bbusenius, I've set an 8.0 milestone on this. Since there's not a rush, let's go ahead and try to use the logic in other appropriate places before merging. Do you want to try that, or would you like me to find some time on my end (possibly with help from @crhallberg and/or @EreMaijala, if they have a particular desire to get involved). I presently have quite a bit of post-holiday catching up to do, but if nobody else has gotten here by the time I'm done with that, I'll be happy to take a stab at it.

bbusenius:
Sorry @demiankatz, I'm just getting back to this. I'm not able to look at the rest of this right now. I need to continue working on getting some of our other customizations in a state to contribute back in preparation for our launch on Folio. If I find an opening, I'll check back on the status of this and see if I can do something.

demiankatz:
Thanks, @bbusenius, that's no problem -- as I said, this is milestoned for 8.0, so there's plenty of time to work on it in the future, and I'll do what I can to help if/when time permits. For the moment, I'm still digging out from holiday break and trying to get other existing PR's reviewed and merged. :-)

themes/bootstrap3/js/record.js
crhallberg: Since it ties into my thoughts of the future theme, I would be in favor of extra descriptive events that make customization easier. I don't know if that applies to this particular event but the case could be made in general.

crhallberg:
It looks like the outstanding concern is a data option toggling the show less button's position, if I'm reading [the review](https://github.com/vufind-org/vufind/pull/1791#pullrequestreview-545898480) correctly.

demiankatz:
@crhallberg, I think the other big outstanding item here is finding time to use this new mechanism in existing places where we have expand/collapse functionality, like in the side facets. If that allows us to eliminate redundant code and simplify things, that would be great. I think we're all just waiting to see who has time to work on that first, since times are busy! If it's something you can fit in, let us know. If not, I'll continue to keep it on my backburner list for now. There are still older pull requests I want to wrap up first, but this one is getting gradually higher on my own list.

crhallberg:
Is this a case where a view helper would come in handy?

```php
<?=$this->collapse([options])->start(); ?>
    [First three items]
<?=$this->collapse()->fold(); ?>
    [The rest]
<?=$this->collapse()->end(); ?>
```

crhallberg:
Just to put together a full list of the options so far:

| Attribute Option | Description |
| --- | --- |
| tag name | element tag for selector? |
| class | keep this |
| data-rows | items before fold |
| data-more-label | override of "show more" |
| data-less-label | override of "show less" |
| data-btn-size | Bootstrap class insertion |
| data-top-toggle | Show toggle at top of list rather than at fold/truncate |
| data-in-place-toggle | swap more and less in place instead of less at bottom |

demiankatz:
Yes, @crhallberg, I think a view helper might be very ... helpful ... here. :-) I guess the question is whether or not that's really feasible, since I think this logic tends to get applied inside iterations through lists of options, and it might not break down as neatly as your example suggests.

Do you think it's worth trying to apply this new functionality to the existing facet controls and then see if there's a natural view helper breakdown (or not)?

EreMaijala:
As far as I can see, a view helper based approach would be unnecessarily convoluted.

demiankatz:
@EreMaijala, I could imagine a view helper being useful if we're rendering elements with many `data-` attributes, and we need to include defaults, and only some of the defaults need to be overridden in any particular situation. This would allow us to replace a big, ugly block of HTML with a more readable helper call with an array of options. But I'm not sure if that's actually the case here, and I agree that trying to encapsulate all the logic of building one of these blocks into a view helper would probably be overly complicated and harder to read.

EreMaijala:
@demiankatz This could be of inspiration for generic html attributes handling: 
https://github.com/NatLibFi/NDL-VuFind2/blob/dev/themes/finna2/templates/components/molecules/containers/finna-panel/finna-panel.phtml#L5
https://github.com/NatLibFi/NDL-VuFind2/blob/dev/module/Finna/src/Finna/View/Helper/Root/HtmlAttributes.php

demiankatz:
Thanks, @EreMaijala, this does look useful. You keep revealing custom view helpers that seem of general value. I wonder if it would be useful to create a PR of Finna view helpers that might be generally useful, and then @crhallberg might be able to take inspiration from them in the new theme (and perhaps refactor some of the current theme, if appropriate).

EreMaijala:
@demiankatz This last one was intended to be proposed for inclusion in VuFind, but something else came up as usual.. We have a lot that we should comb through.

demiankatz:
I know @crhallberg sometimes pokes around in Finna for inspiration. Perhaps it would be worthwhile for him to take a quick skim of your view helper directory!

EreMaijala:
@demiankatz Oh, yes, I'd love it if someone just went and picked the good parts. :) Anyway, I think I'm causing the discussion to drift from the pull request in hand, so I'll just stop here.

crhallberg:
A general html attributes helper would fit the bill but something like `data-rows` would be more complicated to implement this way.

@EreMaijala I'm happy to do a larger sweep of Finna in the near future! I'm building general tools now for the progressive theme and I'll bet much of this work is already in your theme.

aleksip:
Finna uses truncate functionality all over the place, so I wonder if it could go to `common.js` instead of `record.js`? Or maybe even to a file of its own, `truncate.js` (my vote)?

Regarding the `HtmlAttributes` helper, the reason we did not make a VuFind PR is that we made a laminas-view PR instead, and `HtmlAttributes` is going to be in their next release!

`HtmlAttributes` is especially useful if any kind of conditional attribute manipulation is required, or attributes are passed to another template. So using it in this PR would not bring practical benefits, perhaps only a subjective benefit of the different approach.

Just as an example, here is what

```php
<?php $truncateDiv = '<div class="truncate-subjects" data-rows="%s" data-more-label="%s" data-less-label="%s" data-btn-size="%s" data-top-toggle="%s">'; ?>
<?=sprintf($truncateDiv, $maxSubjectsLimit, $this->transEsc('show_more_subjects'), $this->transEsc('show_less_subjects'), 'btn-xs', '30');?>
```

could look like with `HtmlAttributes` (which handles the escaping):

```php
<?php $attributes = ['class' => 'truncate-subjects', 'data-rows' => $maxSubjectsLimit, 'data-more-label' => $this->translate('show_more_subjects'), 'data-less-label' => $this->translate('show_less_subjects'), 'data-btn-size' => 'btn-xs', 'data-top-toggle' => '30']; ?>
<div<?= $this->htmlAttributes($attributes) ?>>
```

demiankatz:
Thanks for the update, @aleksip! I agree that `record.js` may no longer be the appropriate place for this functionality. I'd certainly be open to adding `truncate.js` rather than making `common.js` even bigger. With the asset pipeline, the files can get combined anyway, so using more granular/meaningful filenames makes sense to me.

Congratulations on the HtmlAttributes merge! I couldn't figure out when the next release is likely, or which version number to expect this in, but I'll definitely keep an eye out for it. Once it's available, there may be a few places where we can use it. As you say, the differences are somewhat subjective, but I would argue that one big advantage of the HtmlAttributes approach is that it makes it easier to improve readability by wrapping lines; when using sprintf, you just have one long, ugly string to look at, but with an array of HTML attributes, it's a bit easier to mentally process.

I think the main question, though, at this point, is whether we want to get this wrapped up in time for the 8.0 release, and if so, what outstanding work is absolutely necessary to get us to that point. Does anyone have any time to work on this further? I can try to make some, but I have a lot on my plate in the run-up to the release!

aleksip:
Looking at this more closely, I think there are two different use cases for truncate functionality: truncating a set of elements and truncating lines of text.

The class-based solution is ideal for truncating things like lists and tables where we want to truncate based on elements which can vary in height.

The Finna implementation is mostly used for truncating text contained in an element like `<p>`, and for those cases we can't use a class-based solution like in this PR.

So for a complete solution I believe we would need support for both use cases. Perhaps a `initTrucateElements()` and a `initTruncateLines()` or something along those lines (pun intended)?

demiankatz:
Yes, that does make sense to me. Would it make sense for you to open a PR that introduces truncate.js with initTruncateElements() inside it (and applying this to existing VuFind code if there are any pre-existing use cases that we can refactor -- I honestly can't remember)? That's probably a quick and easy thing to merge, then we could merge those changes into this PR and refactor/rename a little bit as needed.

aleksip:
Just a heads-up that I have started work on expanding the current implementation to include support for line-height based text truncation and some more configuration options for different use cases. Hoping to have it finished by the end of this week.

demiankatz:
Thanks, @aleksip, that's great! If it's easiest to build on this rather than do the work in a separate PR, perhaps @bbusenius can grant you access to this branch... but if he is unavailable to do that, let me know and we may be able to work out a different way to get additional changes pushed here.

aleksip:
I have now made uchicago-library#111 which contains a rather big refactoring of the truncation functionality and also moves it to a separate `truncate.js` file. I can also make a PR to the main VuFind repository if that would be a better way to move this forward.

I have added a `data-element` attribute, which can be used to specify the elements used in element-based truncation. If left out, and not specified in a JavaScript call, truncation will be based on line height.

I have also added a `data-label` attribute, which can be used if both labels should be the same (we have some use cases for this).

The functionality now also supports a zero rows option, which can be useful (we have many use cases for this).

I have tried to demonstrate some use cases in our [component library](https://natlibfi.github.io/NDL-VuFind-ui-components/?p=molecules-finna-truncate-class-demos). The nice thing is that most features can be used just by applying a common class (`finna-truncate` in our component library) and the data attributes. The only feature that requires an extra JavaScript call is the fill function, but a data attribute could be used for this too if a template was provided instead of a function. But maybe this can be done in a follow-up?

demiankatz:
Thanks, @aleksip, this sounds good (though I confess I haven't had a chance to review the code yet). If @bbusenius has no objections and has time to merge https://github.com/uchicago-library/vufind/pull/111, then we can continue to discuss the work here after the merge is completed. If he is too busy, then perhaps a new pull request here would enable us to move forward more quickly.

bbusenius:
Thanks @aleksip, this seems to be working as expected. I've merged this and will keep looking at it.

themes/bootstrap3/templates/myresearch/mylist.phtml
```diff
@@ -16,6 +16,7 @@
   if ($this->params->getOptions()->getListViewOption() != "full") {
     $this->headScript()->appendFile("record.js");
     $this->headScript()->appendFile("embedded_record.js");
+    $this->headScript()->appendFile("truncate.js");
```
demiankatz: @aleksip, is it intentional that you are only loading truncate.js when the list view option is not full (that is, when the option for ajax-loading records into a tabbed or accordioned control is active, instead of linking to the record page)? If I'm understanding correctly, I think we're more likely to ALWAYS want the truncate.js to be available, but maybe I'm mistaken.

themes/bootstrap3/templates/search/results.phtml
```diff
@@ -38,6 +38,7 @@
   if ($this->params->getOptions()->getListViewOption() != "full") {
     $this->headScript()->appendFile("record.js");
     $this->headScript()->appendFile("embedded_record.js");
+    $this->headScript()->appendFile("truncate.js");
```
demiankatz: Here's another instance of a conditional load of truncate.js. This one is also causing a conflict with the latest dev branch, because some of the script loading got refactored to https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/search/results-scripts.phtml -- should we actually be including this there instead?

demiankatz (changes requested):
I haven't fully reviewed this yet, but I noticed that there's a conflict that needs to be resolved here, and investigating that led to a some questions. Once we get this detail sorted out, I'll devote more time to looking more closely at the remainder of the changes.

themes/bootstrap3/templates/myresearch/mylist.phtml
aleksip: @demiankatz To be honest, I was not at all sure about this approach. I just added `truncate.js` wherever `record.js` was added. So should I just remove all these and add  `$this->headScript()->appendFile("truncate.js");` to `layout/layout.phtml` instead?

aleksip:
@demiankatz @bbusenius There are some small changes still to come in uchicago-library#112, which I have marked as draft for now, so I can add proper loading of `truncate.js` to that PR as well.

themes/bootstrap3/templates/myresearch/mylist.phtml
demiankatz: @aleksip, if we're going to be using the truncation feature globally, the easiest thing is probably to add it to the list of Javascript files in the theme configuration here: https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/theme.config.php#L44

aleksip: @demiankatz Ah, yes, of course! Have now done that in uchicago-library#112 and removed the draft status. @bbusenius

bbusenius:
@aleksip I merged  uchicago-library#112 but I'm seeing a PHP notice. Appears to be a problem with loading Javascript.

image

Any ideas?

demiankatz:
@bbusenius, can you try again? I think the problem was just that this branch needed to have dev merged into it -- the configuration that @aleksip uploaded required newer code in order to be parsed correctly.

demiankatz:
Looks like the updated JS code is failing some of the continuous integration tests. Any ideas, @aleksip?

aleksip:
@demiankatz @bbusenius ESLint did not like a variable name, and there was also another test error due to the dev merge which I hope to have fixed in uchicago-library#113. At least all tests now pass for me locally and on GitHub CI.

Sorry about the extra trouble, I was relying a bit too much on the inspections done by my IDE, and trying to cut some corners. Lesson learned.

demiankatz:
Thanks again, @aleksip, and no worries! I'll wait for @bbusenius to merge this before taking a closer look, so it may have to wait until I return from vacation on July 19. I'll comment further as soon as I am able!

bbusenius:
@demiankatz @aleksip, yes merging dev seemed to fix it. I don't see the problem anymore. I just merged uchicago-library#113. Everything seems to be working as expected.

crhallberg:
This continues to look good to me. The only thing standing out to me (that hasn't been said) is maybe we could consolidate the data attributes via JSON? I've seen this style in a couple of libraries:

```HTML
<div data-truncate="{ 'rows': '%s', 'more-label': '%s', 'less-label': '%s', 'btn-size': '%s', 'top-toggle': '%s' }">
```

Just something to consider that might avoid collisions in the future.

demiankatz:
Thanks, @crhallberg, I think that's a good idea, but I'd be interested to hear what @bbusenius and @aleksip think. From my perspective, it seems like a worthwhile adjustment, since as you say, it reduces possible conflicts with other features, expresses the data more concisely, and reduces the number of data() calls in the Javascript code... but if others think it would be unnecessarily disruptive or has disadvantages I'm not thinking of, I'm certainly open to proceeding with the code as written.

bbusenius:
@crhallberg @demiankatz, I don't have a problem with this. I like the idea that it might reduce possible conflicts with other features and I can see how they could conflict given some of these attributes. I think the advantage of the other way is that it is a little easier to read but I'm not too concerned about it in this case. I'd be fine with either approach.

demiankatz:
Thanks, @bbusenius -- I'll see if I can find time to convert the code to the other model so we can see how it looks and feels.

Unfortunately, I just realized that this PR adds new language strings to the code, and translations are already underway for 8.0, so it will not be possible to merge this into the 8.0 release (language files are now frozen). It may be worthwhile to add the new Javascript functionality to 8.0, however, and then we can add the code that actually utilizes the functionality in 8.1 in order to take advantage of the next translation cycle. I have removed the 8.0 milestone to reflect this, but if others feel that it is urgent to at least get the baseline functionality into 8.0 so that we can take advantage of it, I'll add it back. Apologies for my oversight!

demiankatz:
Okay, as promised, in commit d88d96e9a1f75a908796f37255b2ad6341a92cdc, I have refactored things to use a single JSON-encoded settings element. I think this makes things a bit more readable. Interested in hearing thoughts from @bbusenius, @crhallberg and @aleksip.

demiankatz:
I also did a quick experiment to see if I could get the side facets working with the new mechanism instead of the existing one. I was not successful, but I committed my work in progress here in case anyone would like to pick it up and see if they can have better luck: https://github.com/demiankatz/vufind/commit/dadb583fcec45a6c0fdc3711aba363b793c9e655

I think I got at least part-way there, but it's complicated because based on configuration, the more button might pop open a lightbox instead of expanding the on-screen results, so we'll have to be careful how we implement the details. I was just trying to get the basic default case working as a starting point, but I couldn't even quite get there. I got more/less controls to appear on screen, but they didn't work as expected, and there was lots of weird overlapping text. I suspect that someone with more familiarity with the new mechanism might have better luck than I did picking this up and making it work right. :-)

demiankatz:
@crhallberg has picked up the work I started to incorporate this truncation functionality into side facets (in my https://github.com/demiankatz/vufind/tree/truncate-facets-revisions branch) -- see https://github.com/demiankatz/vufind/pull/1 for the work in progress. I've already cherry-picked some of his work here, to improve the way default settings are handled (see ec16fd7).

demiankatz:
Hello, everyone -- I have merged @crhallberg's work into my own branch, and then my own branch into this PR, to apply this new functionality to existing facet collapsing. It definitely needs more thorough testing because there are a lot of edge cases and configuration settings involved... but at least for the basic case, it seems to be working so far! I'll investigate further when time permits, and if anyone has comments in the meantime, please let me know.

aleksip:
Sorry for the late comment, just back from a five week holiday!

`truncate.js` still works with all my test cases after converting to the JSON settings format. 👍 

This might be a subjective thing, but adding JSON can make the HTML slightly more difficult to read and write, especially for someone familiar with HTML only.

```html
<ul class="finna-truncate" data-element="li" data-rows="5" data-in-place-toggle="true">

<ul class="finna-truncate" data-truncate="{'element': 'li', 'in-place-toggle': true, 'rows': 5}">
```

Another downside is moving away from the HTML standard, which means losing existing HTML attribute related functionality, e.g. the syntax highlighting support above, and CSS attribute selectors:

```css
.finna-truncate[data-rows="0"] {
  background: rebeccapurple;
}
```

However, these could be considered as minor things with limited value or actual use cases, so I'm not opposed to the JSON format.

demiankatz:
@aleksip, thanks for the feedback, and welcome back! :-)

I believe there was some discussion earlier about allowing either the JSON or the data format to be supported, and I think this would be technically feasible (though I would advocate for at least prefixing the data elements with `data-truncate` so that they're less potentially ambiguous -- `data-rows`, for example, could easily collide with other libraries). If you think having that extra flexibility would be valuable, I wouldn't object to adding the functionality... this would give us options for the future, though maybe it just makes things worse by offering the opportunity for inconsistent/mismatched implementations, which could ultimately just increase confusion.

To the point about ease of use, I guess it all comes down to target audience. I think the JSON format makes things easier to read and work with for people familiar with PHP, because we can just json-encode an array instead of building a highly verbose HTML element... but I understand that if you're looking at this through an HTML lens rather than a programming lens, the data approach is easier. Supporting both formats might be helpful for this use case.

To the point about applying styles based on data elements, that is certainly a capability that we lose here (and supporting both JSON and data elements wouldn't help, since this would mean that we couldn't rely on the styles, unless the code actually converted the JSON into data elements on the fly). I guess the main question is whether there are real-life use cases that benefit from this.

Bottom line: it sounds to me like you're not strongly objecting to the current approach, but I'm certainly open to further discussion about this, and interested in input from @bbusenius, @crhallberg and others who might have opinions.

aleksip:
@demiankatz Indeed, those were just some thoughts that came to mind, and I am not objecting to the JSON approach. I agree that allowing both ways would be harder to maintain and would just increase confusion.

This is way out of scope for this issue, but in general I feel that keeping templates as free from VuFind-specific code, and even PHP, as possible is a good thing. This makes the front-end easier and more pleasant to work on for both non-experts who just need to change some HTML and advanced front-end developers not familiar with or not particularly interested in (🤯) PHP.

demiankatz:
Thanks, @aleksip. On the subject of PHP in templates, I agree that this is generally a good goal, though there are some tensions that have to be navigated: most importantly, the fact that view helpers are really valuable for keeping templates concise and offering greater flexibility. In particular, as we try to navigate the process of making VuFind more framework agnostic (i.e. factoring out explicit Bootstrap 3 references from the code, so it's easier to upgrade in the future), I think there's going to be a tendency to add more PHP. As you say, we're getting off topic here, but if this is a subject you're interested in, I encourage you to follow the theme conversations so you can be part of the planning process.

crhallberg:
The style option is a good point but if you're customizing a template to add a data attribute, you might as well add a class while you're at it.

aleksip:
Actually, I was specifically thinking about cases where attribute-based styling could be used to avoid template changes. Since we have more than 170 customizable theme instances, avoiding changes to any files is always a positive thing. But as I mentioned earlier, the real life value of the lost possibility here is probably very small.

demiankatz:
Quick update: work on #2121 revealed a problem when using nested truncation elements; commit 6fe0f7f fixes it. Any concerns, @aleksip?

themes/bootstrap3/js/truncate.js
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        if (shouldTruncate) {
+          element.each(function hideRows(i) {
+            if (i === rowCount) {
+              $(this).addClass('truncate-start');
+            }
+            if (i >= rowCount) {
+              $(this).hide();
+              $(this).addClass('truncate-toggle');
```
aleksip: Maybe we can leave this out, unless we want to provide a class expressly for styling togglable elements? If we do want to provide a class, maybe it could be renamed to `truncate-togglable`?

aleksip (approved):
Looks good to me!

themes/bootstrap3/js/truncate.js
demiankatz: I don't have strong feelings about this; I left it in just in case you were already relying on it for styling in your existing implementation, but if you prefer to change or remove it, please feel free.

demiankatz:
Just an update to note that the work which was extracted to #2121 has now been merged to dev, and dev has been merged back into this branch, so now this has just the new functionality, which we can hopefully review and merge as part of the 8.1 work.

demiankatz:
I'm starting to take inventory of outstanding work that needs to be completed for release 8.1, and I see that this seems to have fallen off the radar. @bbusenius, do you have any time over the next couple of months to collaborate on this to get it finished? If not, I can slide the target forward to 9.0... but if you might be free, I can devote some time to re-testing this and refreshing my memory of where we left it!

languages/en.ini
```diff
@@ -1100,6 +1100,12 @@ Series = "Series"
 Set = "Set"
 shortlink_redirect = "You will be redirected in %%delay%% seconds..."
 show_filters_html = "Show filters (%%count%%)"
+show_less = "Show less"
```
demiankatz: Are the show_less / show_more translations actually used anywhere? If not, can we remove them?

languages/en.ini
```diff
@@ -1100,6 +1100,12 @@ Series = "Series"
...
+show_less_items = "Hide items"
+show_less_subjects = "Hide subjects"
```
demiankatz: Also, any objection to calling these show_fewer_items and show_fewer_subjects to be less grammatically awkward?

module/VuFind/src/VuFind/View/Helper/Root/Config.php
```diff
@@ -88,4 +88,26 @@ public function ajaxCoversEnabled()
     {
         return $this->get('config')->Content->ajaxcovers ?? false;
     }
+
+    /**
+     * Should we limit the number of items displayed on the full record?
+     *
+     * @return int
+     */
+    public function itemLimit()
```
demiankatz: This should probably have a more specific name -- maybe holdingsTabItemLimit?

module/VuFind/src/VuFind/View/Helper/Root/Config.php
```diff
@@ -88,4 +88,26 @@ public function ajaxCoversEnabled()
...
+        $limit = $this->get('config')->Site->itemLimit;
+        return $limit ? (int)$limit : PHP_INT_MAX;
+    }
+
+    /**
+     * Should we limit the number of subjects displayed on the full record?
+     *
+     * @return int
+     */
+    public function subjectLimit()
```
demiankatz: This should probably have a more specific name -- maybe recordViewSubjectLimit?

config/vufind/config.ini
```diff
@@ -163,6 +163,14 @@ allowSavedSearches = true
 ; still require Javascript; this simply improves compatibility for certain
 ; features (such as display of hierarchies).
 nonJavascriptSupportEnabled = false
+; Only display x number of subjects in the bib display on the full record by default.
```
demiankatz: I think it would make sense to move both of these new settings down to the [Record] section, since they both apply to record display. I'd also rename them for greater specificity (see comments on config method names below).

demiankatz (changes requested):
@bbusenius, I took a closer look at this to see if it's in shape to be merged. I'm going to be requesting translations in a little over a week, so I want to get this finalized before then if possible. The functionality seems fine, but I had some thoughts on wording/naming/organization which might be worth addressing.

If you agree with my suggestions but don't have time to implement the changes, let me know and I should be able to take care of it.

EreMaijala (approved):
I'm not sure I see the need for item- or subject-specific more/less texts. I think the normal more_ellipsis/less_ellipsis would work there just as well, but I won't block the merge because of that. :) 

Other than that, I share @aleksip's concern about the json format. I find it unintuitive, but I won't block the merge on that either.

bbusenius:
Sorry everyone, I'm on leave. I didn't mean to block progress on this. I agree with @demiankatz's suggestions on naming conventions and changes to the config. I think that all makes sense. I also share @aleksip's and @EreMaijala's concerns regarding the json format but I do not think it should be a blocker. Regarding the  item- and subject-specific more/less texts, I'm pretty sure I introduced that convention because that's what we do at UChicago and I figured it would be nice to preserve it and allow other institutions to be more flexible. In all honesty though, it's not critical and I wouldn't want that to be a blocker either.

demiankatz:
@EreMaijala, do you mind taking one last look at this when time permits? I've made all of my suggested changes in commit 7c06a401560e07309ebf24c8984d19be994252c1, plus I've reverted to default language to keep the files simple (since this is easy to customize as needed) and I reduced the configuration of the truncation (since some options that were used no longer exist, and others seemed like they might be better reverted to defaults and customized locally as needed, to keep things more consistent). If you think I should add anything back, or if you have any suggestions for improved naming or organization, please let me know.

I also still welcome further feedback from @bbusenius if he happens to see this before it gets merged!

demiankatz (approved):

demiankatz:
Also, for the record, the full test suite is passing.

EreMaijala (approved):
Looks good to me (excluding the JSON that I still don't like)!

PR #1795, opened 2020-11-26
Record toolbar permalink

sipek-mzk:
We found useful to add a button to record toolbar to display the permanent link of the current record. What do you think?
This pull request resolve it.


demiankatz:
Thanks, @sipek-mzk; we're on holiday at the moment for American Thanksgiving, but we'll take a closer look when we're back to work. I've asked @crhallberg for a review when time permits.

crhallberg:
How would you feel about adding a convenience feature to copy the URL? Either next to/in place of this button or in the lightbox? 

```js
navigator.clipboard.writeText("permaLink")
```

demiankatz:
@sipek-mzk, just following up to see if you have any thoughts regarding @crhallberg's question from last month. No rush, but just want to make sure we're all on the same page before considering a merge here.

xmorave2:
I tried to add a button for copying text into clipboard. What do you think about it @demiankatz, @crhallberg, @sipek-mzk ?
If you think it viable way, I would like to generalize (probably as template helper) so this can be used in more places.

crhallberg (approved):
I think your copy button will do nicely!

demiankatz:
Thanks, @xmorave2, I think this makes sense and works well. I do agree that it would make sense to generalize the "copy to clipboard" functionality for easier reuse, either as a view helper or a template. In addition to that, a couple of other specific comments:

1.) The "permanent link" button lacks an icon, which makes it look a little out of place next to the other buttons. I imagine there is probably a good standard icon for permalinking that we could utilize -- [this](https://fontawesome.com/v4.7.0/icon/link), for example. Could you or @crhallberg add something before we merge this?

2.) For some reason, when I open the permalink function in the lightbox, the button gets auto-focused, which makes it look a bit strange:
image
Is this intentional?

xmorave2:
Thanks @demiankatz, I generalise the copy to clipboard button and added compiled css - the link icons is now displayed, also autofocus should be fixed.

themes/bootstrap3/templates/Helpers/copy-to-clipboard-button.phtml
```diff
@@ -0,0 +1,29 @@
+<p>
+    <button type="button" id="copyToClipboard" class="btn btn-primary hidden" role="button" tabindex="0"><?=$this->translate("copy_to_clipboard_button_label")?></button>
```
demiankatz: Do you think there might be a scenario where we have more than one of these on the page at the same time? If so, it shouldn't have a hard-coded ID. Should we add a counter to the view helper class and number them for uniqueness?

demiankatz (commented):
Thanks, @xmorave2, this looks good to me. I have just one question...

themes/bootstrap3/templates/Helpers/copy-to-clipboard-button.phtml
xmorave2: It could be possible and adding a counter does not harm, so I've done it. Thanks for that idea @demiankatz

themes/bootstrap3/templates/Helpers/copy-to-clipboard-button.phtml
```diff
@@ -0,0 +1,32 @@
+<?php
+  $buttonNumber = $buttonNumber ?? 1;
```
crhallberg: This looks good: only recommendation is this: if we do want a fallback here just in case, it should be a random number or a timestamp instead of 1 to avoid collision.

demiankatz: Good idea. I think `random_bytes(5)` would do the trick; I'll give it a try.

demiankatz:
Thanks for the progress on this, @xmorave2! I think we're nearly done. I incorporated Chris' suggestion to use a random identifier for the copy button if none is passed in to the template. I also noticed some ongoing problems with the link icon: for some reason, it was still broken for me until I recompiled the CSS. Even after recompiling, it was missing in the bootprint3 theme because those icons are handled differently. I believe this is all fixed now, but you might want to review your LESS compilation process in case something is wrong there (though perhaps this was caused by a merge error). Are you using the PHP-based compiler or the Node-based compiler? I've noticed that the Node version is sometimes more reliable.

At this point, the only outstanding issue is that I'm still seeing the autofocus issue that I reported above, and which you said that you fixed. Are you no longer seeing it on your end? I don't want to waste a lot of your time troubleshooting this, and I can live with it if it's a pain to resolve... but before I merge this, I want to be sure that a fix didn't get lost somewhere along the way.

xmorave2:
Thanks @demiankatz 

> Are you using the PHP-based compiler or the Node-based compiler? I've noticed that the Node version is sometimes more reliable.

I am using PHP based compiler, so that is probably the root of this kind of issues.
 
> At this point, the only outstanding issue is that I'm still seeing the autofocus issue that I reported above, and which you said that you fixed. Are you no longer seeing it on your end? I don't want to waste a lot of your time troubleshooting this, and I can live with it if it's a pain to resolve... but before I merge this, I want to be sure that a fix didn't get lost somewhere along the way.

I tried to add button.blur after button is showed and clicked, this should do the trick I think

EreMaijala:
As a sidenote, perhaps we should consider retiring the PHP-based compiler or at least make it show a warning.

demiankatz:
Good point, @EreMaijala. I have opened #1925 to add a deprecation warning to the PHP-based compiler, and will use that PR to collect comments/feedback from the community for a little while before deciding whether or not to merge it. I also added some related TODO items on the PR.

demiankatz:
@xmorave2, your latest changes have not made any difference to the behavior that I am seeing (which I have tested across both Firefox and Chrome under both Linux and Windows). When I open the lightbox, before I click anything inside it, the button auto-focuses and looks strange. I suspect that this may be related to some behavior of the lightbox itself, and maybe it is necessary for accessibility. Maybe @crhallberg can weigh in on why this is happening. In any case, perhaps a better solution to this problem would be to adjust the styling of selected buttons to make their appearance less potentially confusing.

Unless @crhallberg thinks that this behavior is specifically related to code within this PR, I would suggest that perhaps we should just revert 50c7fad and merge this as-is, then address other problems in a separate PR. I feel like adding explicit blurs may actually be detrimental to accessibility and may just be hiding a symptom of a larger issue... but I defer to @crhallberg's expertise on accessibility matters (and to you, @xmorave2, if you think the blurs are important for a different reason ... my apologies if I'm making this all more complicated than it needs to be).

crhallberg:
The autofocus is likely coming from the lightbox, which autofocuses the first element: https://github.com/vufind-org/vufind/blob/9663efa4079070e0c6b2a4c9e667c943ea2650de/themes/bootstrap3/js/lightbox.js#L502

xmorave2:
> The autofocus is likely coming from the lightbox, which autofocuses the first element:
> 
> https://github.com/vufind-org/vufind/blob/9663efa4079070e0c6b2a4c9e667c943ea2650de/themes/bootstrap3/js/lightbox.js#L502

Thanks @crhallberg!

@demiankatz So I reverted last commit and this should be ready

demiankatz:
Thanks, everyone! I'm merging this now.

@crhallberg, do you think we should consider different styling for focused buttons, or am I just being too fussy?

PR #1796, opened 2020-11-26
Fix bootstrap3 browse home template

sipek-mzk:
Just fix the closing html tag.


demiankatz:
Thanks! I backported this fix to the release-7.0 branch so it can be included in 7.0.2.

PR #1797, opened 2020-11-27
Fix cache key in ObalkyKnihService class

xmorave2:
Sometimes the cache key could be so long that the cache file is not handled correctly by filesystem.


demiankatz:
Thanks! I backported this fix to the release-7.0 branch so it can be included in 7.0.2.

PR #1798, opened 2020-11-27
KohaRest: Check also for 'errors' element in response.

EreMaijala:
Looks like a low-level error such as 500 results in 'errors' element instead of 'error'.


demiankatz:
I've backported this to release-7.0 so it can be included in release 7.0.2.

PR #1799, opened 2020-11-30
Allow Javascript to be loaded in the header OR the footer

mtrojan-ub:
Progress on: https://vufind.org/jira/browse/VUFIND-1437

Much of the original content of this PR was split off into #1863. What remains is logic for adding Javascript files to not just the header of the page, but also the footer.

TODO
- [x] Reconcile new functionality with asset pipeline


module/VuFindTheme/Module.php
```diff
@@ -106,7 +106,7 @@ public function getViewHelperConfig()
                     View\Helper\TemplatePathFactory::class,
             ],
             'aliases' => [
-                'headThemeResources' => View\Helper\HeadThemeResources::class,
+                'themeResources' => View\Helper\ThemeResources::class,
```
demiankatz: It might be worth including a deprecated HeadThemeResources helper that simply proxies via `$this->view->plugin('themeResources')` and throws a NOTICE in order to reduce backward compatibility problems with custom themes during upgrading. I can help with the details if you like!

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -160,11 +184,46 @@ public function getCss()
...
      * Get Javascript files.
      *
+     * @param string $position Position where the files should be inserted
+     * (allowed values are 'header' or 'footer').
+     *
+     * @return array
+     */
+    public function getJs(string $position=null)
+    {
+        $this->js = array_unique($this->js, SORT_REGULAR);
```
demiankatz: Did you test whether array_unique works correctly with nested/multidimensional arrays? My suspicion is that it may not.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -160,11 +184,46 @@ public function getCss()
...
+        if (!isset($position)) {
+            return $this->js;
+        } else {
+            $js = [];
```
demiankatz: A functional approach might be slightly more efficient here:

<pre>
return array_filter($this->js, function ($jsFile) use ($position) {
    return $jsFile['position'] == $position;
});
</pre>

module/VuFindTheme/src/VuFindTheme/View/Helper/ThemeResources.php
```diff
@@ -56,37 +56,23 @@ public function __construct(\VuFindTheme\ResourceContainer $container)
...
+    public function __invoke($position)
     {
-        $parts = explode(':', $current);
-        // Special case: don't explode URLs:
-        if (($parts[0] === 'http' || $parts[0] === 'https')
-            && '//' === substr($parts[1], 0, 2)
-        ) {
-            $protocol = array_shift($parts);
-            $parts[0] = $protocol . ':' . $parts[0];
+        if (!isset($position) || $position == 'header') {
```
demiankatz: Is there a compelling reason to check for `$position` being unset?

module/VuFindTheme/src/VuFindTheme/View/Helper/ThemeResources.php
```diff
@@ -56,37 +56,23 @@ public function __construct(\VuFindTheme\ResourceContainer $container)
...
+            $this->addMetaTags();
+            $this->addLinks();
+            $this->addScripts('header');
+        }
+        if (!isset($position) || $position == 'footer') {
+            return $this->addScripts('footer');
```
demiankatz: Is a `return` really needed here?

module/VuFindTheme/src/VuFindTheme/View/Helper/ThemeResources.php
```diff
@@ -161,22 +147,36 @@ protected function addLinks()
...
+                    'text/javascript',
+                    $current['conditional'] ?? []
+                );
+            }
+        } elseif ($position == 'footer') {
+            $scripts = '';
+            foreach ($js as $current) {
+                // not sure if this call is correct, there seems to be lots of overhead
+                // no matter if setFile, appendFile, or prependFile is used
+                $scripts .= $this->getView()->plugin('inlineScript')->setFile($current['file'],
```
demiankatz: I would add a line break before `$current['file']` here and indent the contents of the function call a bit less for readability.

demiankatz (changes requested):
Thanks, @mtrojan-ub! This looks like a good start; I've left some initial questions and suggestions.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: It works if SORT_REGULAR is used, according to the documentation (else all elements will be treated as strings per default).

demiankatz: Thanks! I actually checked the documentation yesterday and didn't come away with the same interpretation on the first reading, but I think I see now how this could work. :-) The "Note: Two elements are considered equal if and only if (string) $elem1 === (string) $elem2 i.e. when the string representation is the same, the first element will be used. " confused me -- but I guess SORT_REGULAR makes that note irrelevant, and regular `==` comparison of non-string-cast arrays would be appropriate for our purposes.

module/VuFindTheme/src/VuFindTheme/View/Helper/ThemeResources.php
mtrojan-ub: Yes, due to the differences between Laminas HeadScript and InlineScript View Helpers.

HeadScript just adds the script, but InlineScript returns a JS snippet that needs to be added to the corresponding spot in the template:
- https://docs.laminas.dev/laminas-view/helpers/head-script/
- https://docs.laminas.dev/laminas-view/helpers/inline-script/

demiankatz: Okay, that makes sense... but then the @return tag on the comment for this function needs to be adjusted to indicate that a string may be returned (and maybe we should return an empty string or null the rest of the time for consistency).

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -130,13 +130,100 @@ public function addCss($css)
     public function addJs($js)
     {
         if (!is_array($js) && !is_a($js, 'Traversable')) {
-            $js = [$js];
+            $this->addJsEntry($js);
+        } elseif (isset($js['file'])) {
+            $this->addJsEntry($js);
+        } elseif (isset($js[0])) {
```
demiankatz: Have you tested what happens if you pass an empty array? That probably would never happen in reality, but if somebody had a JS-free theme, we should make sure that it doesn't break. I'd suggest that if the incoming value is empty, you return without doing anything rather than throwing an exception. It's worth adding this to the unit test.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -130,13 +130,100 @@ public function addCss($css)
...
+            for ($i=0; $i < count($array);++$i) {
+                if (isset($entry['priority'])) {
+                    $currentPriority = $array[$i]['priority'] ?? null;
+                    if (!isset($currentPriority)
+                        || $currentPriority > $entry['priority']
+                    ) {
+                        $this->insertEntryAtPosition($entry, $i, $array);
+                        return;
+                    }
+                } elseif (isset($entry['dependency'])) {
```
demiankatz: If we want to keep this simple, I wonder if `load_after` would be a better name for the setting than `dependency` -- `dependency` implies to me more complex/sophisticated dependency resolution, and I don't want us to end up having to reinvent Composer inside VuFind's theme system. :-)

A couple of other questions/concerns:

1.) Do we need to worry about dependencies on multiple files? If one file needs to be loaded after two other files, and we can't predict the order of those files being loaded, how do we deal with that situation?

2.) Do we need a "load_before" setting as a counterpart to "load_after"?

3.) What do we do if the file we're supposed to depend upon is not already present in the Javascript array? Do we throw an exception? If we're not guaranteeing the order if the input, I don't think we can assume that the dependency is already available at the time we load the dependent file.

(And as a side note, all of these types of questions are the reason I tend to prefer a priority system, because it's simple and unambiguous... but of course the downside is that it requires management of priority levels... I just don't think that's likely to be too complex for VuFind, if we just use something like priority 0 for core libraries like jQuery and Bootstrap, priority 50 for VuFind-specific support code, and priority 100 for anything more specific than that... and then local custom code could use higher numbers as needed. I'm not opposed to continuing to work through the issues of a before/after/dependency system, but just warning that I think it will take significant work to cover all the possible edge cases.)

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -130,13 +130,100 @@ public function addCss($css)
...
+     * Helper function to insert an element into an array
+     * at a certain position.
+     *
+     * @param array $entry    The entry to be inserted.
+     * @param int   $position The position to insert the entry.
+     * @param array $array    The array in which the entry shall be inserted.
+     *
+     * @return void
+     */
+    protected function insertEntryAtPosition($entry, $position, &$array)
```
demiankatz: Do you need this support method? I think `array_splice($array, $position, 0, $entry)` will do the same thing with native PHP.

themes/bootstrap3/theme.config.php
```diff
@@ -11,14 +11,14 @@
...
     'js' => [
-        'vendor/jquery.min.js',
-        'vendor/bootstrap.min.js',
-        'vendor/bootstrap-accessibility.min.js',
-        'vendor/validator.min.js',
-        'lib/form-attr-polyfill.js', // input[form] polyfill, cannot load conditionally, since we need all versions of IE
-        'lib/autocomplete.js',
-        'common.js',
-        'lightbox.js',
+        ['file' => 'vendor/jquery.min.js', 'priority' => 100],
```
demiankatz: As I suggested above, I don't think we necessarily need to give every file a unique priority value. I was viewing this as a way of grouping things with common priority levels, rather than a system for explicitly setting the exact order of everything. Thus, I'd group them together in as few levels as possible. If files don't depend on one another, they should get the same level.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -130,13 +130,100 @@ public function addCss($css)
...
+                ];
+            }
         }
-        foreach ($js as $current) {
-            $this->js[] = $current;
+        if (!isset($jsEntry['position'])) {
+            $jsEntry['position'] = 'header';
+        }
+
+        if (!in_array($jsEntry, $this->js, true)) {
```
demiankatz: Just food for thought, but do we want to test for duplicates only using full array matches, or would it be better to match on a subset of attributes, so that a sub-theme can override a parent theme's settings? Or perhaps an easier approach would be to consider an `override` boolean attribute that could cause a child theme's settings for a file to replace those already loaded from a parent theme. Not something we need to actually implement before there's a real use case, but just a possibility to consider while designing this.

demiankatz (commented):
Thanks, @mtrojan-ub! I haven't had an opportunity to actually test this yet, but I left some feedback and ideas based on a quick review of the code. I'm a bit short on time this month, but I'll do more testing as things develop and time permits.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: OK, i added support for this case & added a test case

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: You're right! Due to the documentation, i didn't think it would be allowed to pass 0 as length, but the test case seems to work.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: OK, i just renamed "dependency" to "load_after". Regarding your other questions:

1) We should always be able to predict the order of the other files by adding clear priorities or dependencies.

2) I think that's not really necessary. If there might be a case, this could be indeed solved by using a priority instead.

3) I guess the order should be guaranteed, else it would be necessary to implement a 2-pass-sorting algorithm. So i just added an exception if the "load_after" dependency could not be found.

themes/bootstrap3/theme.config.php
mtrojan-ub: That's true for custom themes, but i think that VuFind's default themes should have priority values anyway. Only if this is the case, the custom themes can set a distinct priority which is e.g. exactly between jquery.min.js and bootstrap.min.js. If a custom theme chooses to add additional values, it may do so without setting any priority.

Also, the current implementation will add all priority-based entries at the top in the correct order, and append all non priority-based entries behind that. If the default themes' entries would have no priority, the custom themes' entries with a priority would be above that, which might lead to runtime errors.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
demiankatz: Thanks!

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
demiankatz: Thanks! This sounds good enough for now, and we can add more complexity if the need arises. :-)

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
demiankatz: Great!

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeResourceContainerTest.php
```diff
@@ -80,10 +80,24 @@ public function testLess()
...
+
+        $expectedResult = [['file' => 'a', 'position' => 'header'],
+                           ['file' => 'b', 'position' => 'header'],
+                           ['file' => 'c', 'position' => 'header'],
+                           ['file' => 'd', 'position' => 'header'],
+                           ['file' => 'http://foo/bar',
+                            'position' => 'header',
+                            'attributes' => ['conditional' => 'lt IE 7']],
+                        ];
+        $this->assertEquals($container->getJs(), $expectedResult);
```
demiankatz: Minor point, but you should reverse the order of these parameters: in assertions, the expected value comes before the actual value (otherwise, error messages may come out in a misleading way).

themes/bootstrap3/theme.config.php
demiankatz: That makes sense, though I wonder if it would at least be helpful to establish groupings through the priority numbering (for example, what if all the vendor dependencies were 1xxx and all the local ones were 2xxx? But maybe I'm overthinking this).

demiankatz (commented):
Thanks for the progress on this! I think the next step may be to write some more exhaustive test cases to cover various scenarios involving the load_after and priority settings... though perhaps we should get more people involved in the conversation before we invest the time in building those, in case any more substantial changes are proposed. My time remains somewhat limited this month, but I'm happy to invest more time in this in January if you need more input, or test-building assistance.

themes/bootstrap3/theme.config.php
mtrojan-ub: If we add the grouping as first number in the priority, this would also affect the sorting and no longer allow a local dependency to be sorted within vendor dependencies. On the other hand, if we would add an additional 'group' key, this would already be possible but only be for used better readability and not have any real effect inside the code.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: You're right! If we allow to override an existing dependency, we should be aware of the consequences. For example => If the overridden settings contain a different priority or dependency, they need to get moved. This also means that all other entries that are depending on it need to be re-evaluated.
If we don't need this right now and the recursion is not implemted, would it be better to throw an Exception in this case (for consistency reasons)?

themes/bootstrap3/theme.config.php
demiankatz: I think we may not be entirely understanding each other. :-) I'm not suggesting that the first digit of the priority number should have any special meaning within the code, or that we need to add a specific group key for different "types" of dependencies, or anything like that. I'm just suggesting that if we're choosing arbitrary integers for setting priority, we could come up with a convention for how the core code uses those numbers... so, for example, vendor dependencies could be 1010, 1020, 1030, etc. and VuFind dependencies could be 2010, 2020, 2030, etc. An end user could feel free to break this pattern if there was a reason to, and the code wouldn't care. I just thought it might be more readable if there was some kind of visible pattern in the built-in defaults.

demiankatz: (And again, maybe this is not a good idea -- I certainly don't insist upon it; I just want to be sure I'm being clear about what my idea actually is!)

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
demiankatz: I agree that it's probably not worth implementing more complex logic here unless we have a use case. I'm fine with throwing the exception as you have implemented, but we might also consider simply ignoring duplicates and logging a warning. (For now, I think the exception is probably a better idea, since it's less ambiguous and forces the user to fix the problem... but I wanted to mention the alternative for the sake of discussion; it might be preferable if we think there are likely to be real-world scenarios where custom themes already have duplicate filenames, since it would be more forgiving during upgrades... but fortunately, I don't think that scenario is actually very likely at all).

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -130,11 +130,102 @@ public function addCss($css)
...
+            if (count($parts) == 1) {
+                $jsEntry = ['file' => $jsEntry];
+            } else {
+                $jsEntry = [
+                    'file' => $parts[0],
+                    'attributes' => ['conditional' => trim($parts[1])],
+                ];
+            }
+        } elseif (isset($jsEntry['priority']) && isset($jsEntry['load_after'])) {
+            throw new \Exception(
```
demiankatz: It would be a good idea to add a small unit test to cover this case; you can use `$this->expectException(\Exception::class);` and `$this->expectExceptionMessage('...the message...');` at the beginning of the test case prior to triggering the exception. That way we can make sure this edge case doesn't break in the future (and we'll be reminded to test it properly if we ever change the behavior).

demiankatz:
Thanks for the progress, @mtrojan-ub! I've looked at your latest commits and responded to your latest replies, but I'll wait for our remaining conversations to be fully resolved before I circle back and do one more full test/review cycle.

themes/bootstrap3/theme.config.php
mtrojan-ub: Thanks for clarifying! I think the main question is this is well understandable by the users, because they need to keep in mind that e.g. 2020 would be higher ranked than 1030, followed e.g. by 40 (or 3040) and so on.

Maybe we should discuss this at the next community call to get some more opinions?

demiankatz: Agreed. This is already on the planned agenda for the next community call, so we can definitely discuss in more detail at that time!

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -130,11 +130,102 @@ public function addCss($css)
...
+     *
+     * @param array $entry The entry to insert.
+     * @param array $array The array into which the entry shall be inserted.
+     *
+     * @return void
+     */
+    protected function insertEntry($entry, &$array)
+    {
+        if (isset($entry['priority']) || isset($entry['load_after'])) {
+            for ($i=0; $i < count($array);++$i) {
```
demiankatz: `foreach (array_keys($array) as $i) {` might be a little more efficient here.

module/VuFindTheme/src/VuFindTheme/View/Helper/HeadThemeResources.php
```diff
@@ -39,144 +39,18 @@
...
-     *
-     * @return void
-     */
-    protected function addMetaTags()
-    {
-        // Set up encoding:
-        $headMeta = $this->getView()->plugin('headMeta');
-        $headMeta()->prependHttpEquiv(
-            'Content-Type', 'text/html; charset=' . $this->container->getEncoding()
+        trigger_error(
```
demiankatz: It would make sense to give this `E_USER_DEPRECATED` as the second parameter, to raise a more appropriate error level.

themes/bootstrap3/theme.config.php
demiankatz: @mtrojan-ub, thanks for the continued work on this! I wonder if it has become more complicated than I intended, though -- my original thought was that the priority values would be sorted as simple integers, but that we'd use a three-digit convention in the core code to improve readability, demonstrate some logical grouping of related files, and keep things unambiguous (since if we use the same number of digits for all values, numeric and alphabetic sort orders work out the same). It looks like you've taken steps to enforce three digit numbers all the time, and I'm not exactly sure what canonizePriority is for -- are you sorting on the second digit rather than the first? Anyway, I apologize if I expressed my ideas unclearly and led you to do more work than was necessary... but on the other hand, if you are happy with what you've come up with, I'm happy to discuss it further and make sure I understand it properly. :-)

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: The test case has been added, however it is not shown automatically as "outdated" because the change has been made in a different line.

themes/bootstrap3/theme.config.php
mtrojan-ub: @demiankatz, Thanks for having another look! To answer your questions:
- canonizePriority is just a function to extract the sorting value from a priority. So, if we have a 3-digit priority with the first digit being ignored for sorting, the function simply strips the first digit. Maybe another name might be better? (something like extractPrioritySortingValue, maybe you have a better idea?)
- After the first digit is stripped, right now the lowest number will be listed first, so lower number means in fact = higher priority
- I thought it might be better to restrict priorities to 3-digit-values to make things clearer (and you immediately get an error message if you chose the wrong scheme). We could also open up the scheme for all kinds of numbers, but what shall happen then? Will the first digit be ignored for sorting as well? What happens if a priority with only 1 digit is given / is it allowed and treated without a group, or should there be an error?

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
demiankatz: Thanks! You can feel free to mark conversations as resolved in situations like this. :-)

themes/bootstrap3/theme.config.php
demiankatz: I think the point I am confused about is why we are not using the first digit for sorting. My thought was that third-party vendor dependencies always need to load first -- that's why we give them 1xx numbers. Then locally-developed libraries load next, because they might depend on third-party code but they do not contain VuFind-specific logic -- that's why we give them 2xx numbers. Then very specific, VuFind-oriented code loads last -- that's why we give them 3xx numbers. My thought was that we'd just sort on the whole number, not ignore the first digit. And of course users would be free to change priorities as appropriate, but for core code, this convention seemed adequate. So I was expecting a numbering scheme like 100, 110, 120, 210, 220, 310, rather than 100, 110, 120, 230, 240, 350. Like I said, I'm certainly happy to discuss the merits of a different solution -- but I just want to be clear about what my original idea was, since I don't think we walked away from that conversation with exactly the same ideas. :-)

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: Ok, i will 😉

themes/bootstrap3/theme.config.php
mtrojan-ub: Thanks for the explanation!

To make sure that i understand correctly this time: Do you mean that a user should be able to e.g. set one of his custom libraries to a priority of 115 if he wants to load it between third-party vendor dependencies with 110 and 120, even if his 115 is not a third-party itself?

demiankatz: Yes, exactly! I intended this to be a convention/best practice but not an enforced requirement.

mtrojan-ub: OK, so now i re-simplified that. Please have another look!
/edit: I had to pull from dev first to avoid merge conflicts, however there seem to be some travis problems now that don't have to do anything with the changes in my code (testMarc.... and so on.)

demiankatz: This looks good to me. I'm not seeing any Travis failures here; am I missing something?

In any case, at this point, is there any outstanding work here, or do you feel this is ready to merge? If you're happy with what we have, then I can give it one more thorough review, and merge it if I find no further concerns.

Thanks again for all of the work on this!

mtrojan-ub: The travis problems seem to be gone, and the version should be complete by now so it should be ready to merge!
Thank you again very much for your support!

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -130,11 +130,102 @@ public function addCss($css)
     public function addJs($js)
     {
         if (!is_array($js) && !is_a($js, 'Traversable')) {
-            $js = [$js];
+            $this->addJsEntry($js);
+        } elseif (isset($js['file'])) {
```
demiankatz: Is there any reason not to combine this condition with the above if clause, since both cases do the same thing?

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -130,11 +130,102 @@ public function addCss($css)
...
+            $this->addJsEntry($js);
+        } elseif (isset($js[0])) {
+            foreach ($js as $current) {
+                $this->addJsEntry($current);
+            }
+        } elseif ($js == []) {
+            return;
+        } else {
+            trigger_error("Invalid JS entry format: " . print_r($js, true));
```
demiankatz: Would it be better to throw an exception rather than trigger an error here?

themes/bootstrap3/theme.config.php
```diff
@@ -11,14 +11,14 @@
...
     'js' => [
-        'vendor/jquery.min.js',
-        'vendor/bootstrap.min.js',
-        'vendor/bootstrap-accessibility.min.js',
-        'vendor/validator.min.js',
-        'lib/form-attr-polyfill.js', // input[form] polyfill, cannot load conditionally, since we need all versions of IE
-        'lib/autocomplete.js',
-        'common.js',
-        'lightbox.js',
+        ['file' => 'vendor/jquery.min.js', 'priority' => 110],
```
demiankatz: Would it make sense to add a comment here documenting the available options?

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -160,11 +251,43 @@ public function getCss()
     /**
      * Get Javascript files.
      *
+     * @param string $position Position where the files should be inserted
+     * (allowed values are 'header' or 'footer').
+     *
      * @return array
      */
-    public function getJs()
+    public function getJs(string $position=null)
```
demiankatz: For some reason, none of the style checkers seem to mind, but I generally put spaces around `=` characters in function definitions. Might be worth changing here for consistency.

demiankatz (changes requested):
@mtrojan-ub, I just did one last review and found a few very minor suggestions and questions. Once you have dealt with these, I'll run the full test suite one more time to be sure everything is passing, and then I think we can merge this.

One other question in the meantime: there's a TODO checkbox here to document this in the wiki, but I'm not sure if there's an obvious place to do that there -- the wiki links back to the example of theme.config.php in the actual code, which is why I suggested perhaps documenting the settings there. But I'm open to other approaches; let me know what you think!

themes/bootstrap3/theme.config.php
mtrojan-ub: I just added a documentation there and tended to the other changes, please have a look!

mtrojan-ub: (also i get some strange error messages again from phpunitfast, related to files i didnt change... might be caused due to problems in the dev branch so i have a hard time executing tests on my local dev machine)

demiankatz: Thanks for the progress, I think we're almost done!

The errors are related to the indentation of your comments in theme.config.php; they're not aligned the way php-cs-fixer expects them to be. You should be able to run `vendor/bin/phing php-cs-fixer` to correct the problem automatically.

If you don't have time, or if you're having trouble getting it to work, let me know and I can do it for you.

In any case, once the style issues are sorted out, I'll run the full test suite one last time and then merge if nothing fails.

EreMaijala:
I've only now had time to check this a bit. I admit that I'm not sure if I grasp this at all, so please bear with me. Anyway, I'll add some comments about things that came to my mind.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -129,12 +129,101 @@ public function addCss($css)
      */
     public function addJs($js)
     {
-        if (!is_array($js) && !is_a($js, 'Traversable')) {
-            $js = [$js];
+        if ((!is_array($js) && !is_a($js, 'Traversable')) || isset($js['file'])) {
```
EreMaijala: It seems that addJs handles quite a few different kinds of input. Would it make sense to add a new method for adding the keyed version and make it and this one call an internal method after massaging the input to a common format?

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -129,12 +129,101 @@ public function addCss($css)
...
+                $jsEntry = ['file' => $jsEntry];
+            } else {
+                $jsEntry = [
+                    'file' => $parts[0],
+                    'attributes' => ['conditional' => trim($parts[1])],
+                ];
+            }
+        } elseif (isset($jsEntry['priority']) && isset($jsEntry['load_after'])) {
+            throw new \Exception(
+                'Using "priority" as well as "load_after" in the same entry '
```
EreMaijala: Nit: double space (end of first line + start of second)

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -129,12 +129,101 @@ public function addCss($css)
...
+                . ' is not supported: "' . $jsEntry['file'] . '"'
+            );
+        }
+
+        if (!isset($jsEntry['position'])) {
+            $jsEntry['position'] = 'header';
+        }
+
+        foreach ($this->js as $existingEntry) {
+            if ($existingEntry['file'] == $jsEntry['file']) {
```
EreMaijala: Does this mean that if I inherit a theme and add the same js as the parent theme in my child theme, it will blow up? If so, that will break the current behavior for us. We list all the scripts in order for inclusion in asset pipeline.

module/VuFindTheme/src/VuFindTheme/View/Helper/ThemeResources.php
```diff
@@ -0,0 +1,186 @@
...
+            );
+        }
+    }
+
+    /**
+     * Add scripts to header or footer.
+     *
+     * @param string $position 'header' or 'footer'.
+     *
+     * @return string|null
```
EreMaijala: I find this dual behavior a bit confusing. Would it make sense to create a FooterScript helper to handle scripts in footer? Not doing so would mean that footer scripts couldn't use the asset pipeline (be minimized), right?

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
demiankatz: I think this probably will cause a problem, then. Do we need some kind of a deeper comparison function so it only fails if the filename is the same but other details are different?

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
demiankatz: Good point, that might increase readability and make it easier to deprecate the old format if we decide to in the future.

module/VuFindTheme/src/VuFindTheme/View/Helper/ThemeResources.php
demiankatz: I hadn't considered this, but I see now you are correct that footer scripts will not go through the asset pipeline since they're using inlineScript instead of headScript (which, of course, makes sense given the context in which they are being rendered). I don't think there's an easy solution to that short of adding asset pipeline support to the inlineScript helper. Would it make sense to address that concern in a separate pull request?

I guess the other question is whether to change HeadThemeResources to ThemeResources as was done here, or to add FooterThemeResources. I don't have a strong opinion -- that approach might be more clear, but the current approach is more flexible if we have additional positions for scripts in the future. One key question is whether there is likely to be a third position in the future, or if there are only two possible scenarios to consider.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: The logic has been split up in the meantime, but this affects the addJsEntry function rather than the addJs function itself.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: The code will now use array_diff and only throw an Exception if the new setting differs from the old one.

module/VuFindTheme/src/VuFindTheme/View/Helper/ThemeResources.php
mtrojan-ub: Since i don't know the asset pipeline that well i can't really take part in the discussion... Please let me know if i can implement any concrete optimizations. Maybe you would like to discuss this topic at the next community call on March 2nd, unfortunately i won't be able to attend this time.

demiankatz: @mtrojan-ub, I wonder if the problem here is that this pull request is solving two problems at once (the ordering of JS files, as well as the header vs. footer rendering of those files). I think it might be easier to discuss, review and ultimately merge this if we break it into two parts. I think we're very close to getting the advanced configuration/ordering logic merged, but I think the issues @EreMaijala raised require some more thought and discussion on the header vs. footer issue.

How would you feel if I opened a new PR containing only the relevant changes to the ResourceContainer and theme configuration? If we can get that fully tested and merged, then it will reduce this PR to only the positioning parts, and those may be easier to talk about if there's less unrelated code coexisting with them. If you think it's worth the effort, I'm happy to give it a try.

mtrojan-ub: That sounds good! Let me know if i can help you in any way.

demiankatz: Great -- see #1863. I'll formally request a review from you there as soon as I finish running the test suite to make sure I did this right... :-)

demiankatz:
Following the merge of #1863, I have renamed this PR and edited the top comment to make it more clear what remains here. @mtrojan-ub and @EreMaijala, perhaps we can talk about this further on the next Community Call to figure out how we can/should reconcile this behavior with the asset pipeline.

demiankatz:
This PR hasn't moved in a while, but I at least took the time to merge in the latest dev code and resolve some conflicts and style issues. Do you have time to look at next steps for this, @crhallberg? I think the main issue here is figuring out whether/how to incorporate asset pipeline functionality into footer-loaded JS code.

themes/root/templates/layout/help.phtml
```diff
@@ -1,12 +1,13 @@
 <?=$this->doctype('XHTML1_TRANSITIONAL')?>
 <html xmlns="http://www.w3.org/1999/xhtml" lang="<?=$this->layout()->userLang ?>" xml:lang="en">
   <head>
-    <?php $this->headThemeResources(); ?>
+    <?php $this->themeResources('header'); ?>
```
demiankatz: This needs to be fixed...

module/VuFindTheme/tests/unit-tests/src/VuFindTest/View/Helper/ThemeResourcesTest.php
```diff
@@ -50,9 +50,29 @@ class HeadThemeResourcesTest extends \PHPUnit\Framework\TestCase
      */
     public function testHelper()
     {
-        $helper = new HeadThemeResources($this->getResourceContainer());
+        $helper = new ThemeResources($this->getResourceContainer());
```
demiankatz: This test needs to run on SetUpThemeResources now that we've renamed the class.

module/VuFindTheme/Module.php
```diff
@@ -89,8 +89,10 @@ public function getViewHelperConfig()
     {
         return [
             'factories' => [
+                View\Helper\FootScript::class =>
+                    View\Helper\PipelineInjectorFactory::class,
                 View\Helper\HeadThemeResources::class =>
-                    View\Helper\HeadThemeResourcesFactory::class,
```
demiankatz, line 93: Since SetUpThemeResources now has the same signature as HeadThemeResources, I don't think we need to build a whole helper for legacy purposes... I think we can probably delete HeadThemeResources, but set up HeadThemeResources as an alias for SetUpThemeResources; this way, outdated templates will continue to work.

demiankatz (changes requested):
Thanks to @crhallberg for solving the outstanding asset pipeline problem; however, now that his work is merged back here, I see there's still a bit of outstanding work remaining. I might be able to get to this next week if nobody else gets there first.

crhallberg:
@mtrojan-ub Are you able to take care of this feedback? Would you like me to open a new PR (or give me write access to the branch)?

module/VuFindTheme/src/VuFindTheme/View/Helper/SetupThemeResources.php
```diff
@@ -141,21 +140,35 @@ protected function addLinks()
...
+
+            $this->getView()
+                ->plugin($helper)
+                ->forcePrependFile(
+                    $current['file'],
+                    'text/javascript',
+                    $current['attributes'] ?? []
+                );
         }
+        return null;
```
demiankatz: Is there a reason for this to return null instead of remaining a void method? @crhallberg, did you do this for a reason, or was it already here?

demiankatz:
I had a little time tonight, so I applied my own suggestions. I also noticed one small question for @crhallberg (above). I'm interested to hear @mtrojan-ub's thoughts when time permits!

module/VuFindTheme/src/VuFindTheme/View/Helper/SetupThemeResources.php
crhallberg: Already there from when footer scripts were returned as an inline string. Void would be fine.

mtrojan-ub:
Thanks for all the progress on this! I just came back from a 2-week-vacation and didn't have the chance to reply earlier, sorry for that.

I think it might be useful to add some footer entries to `ThemeResourceContainerTest.php::testJs()` and split up the `expectedResult` into `expectedResultHeader` and `expectedResultFooter` and checking them seperately. What do you think?

module/VuFindTheme/src/VuFindTheme/View/Helper/SetupThemeResources.php
demiankatz: Fixed, thanks!

demiankatz:
Good idea, @mtrojan-ub -- I've added a test case in 8f841de. Is there any further work we should consider? Any final thoughts from @EreMaijala?

module/VuFindTheme/tests/unit-tests/src/VuFindTest/View/Helper/SetupThemeResourcesTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * HeadThemeResources view helper Test Class
+ * ThemeResources view helper Test Class
```
EreMaijala: *SetupThemeResources*

module/VuFindTheme/tests/unit-tests/src/VuFindTest/View/Helper/SetupThemeResourcesTest.php
```diff
@@ -28,18 +28,18 @@
 namespace VuFindTest\View\Helper;
 
 use VuFindTheme\ResourceContainer;
-use VuFindTheme\View\Helper\HeadThemeResources;
+use VuFindTheme\View\Helper\SetupThemeResources;
 
 /**
- * HeadThemeResources view helper Test Class
+ * ThemeResources view helper Test Class
```
EreMaijala: *SetupThemeResources*

EreMaijala (approved):
I'd fix the two noted comment lines for clarity, but otherwise this looks good to me.

demiankatz:
Thanks, @EreMaijala! I have fixed the incorrect comments and will merge shortly.

demiankatz:
@EreMaijala, I have also resolved [VUFIND-1437](https://vufind.org/jira/browse/VUFIND-1437) since the functional additions discussed there are now implemented. There is also a note about writing documentation/comments/best practices to reduce the number of generated bundles; I know that's something you were thinking about. Do you have anything that should be shared, or is that still work in progress?

PR #1802, opened 2020-12-02
Filter out empty arrays

welblaud:
In the case the `extractEbscoDataFromRecordInfo` method returns an empty array, which is expected behavior (also associated with the `recurseIntoRecordInfo` method), `array_unique` blows up. With additional filtering, it works as expected.

### DIFF
--- a/module/VuFind/src/VuFind/RecordDriver/EDS.php
+++ b/module/VuFind/src/VuFind/RecordDriver/EDS.php
@@ -464,9 +464,7 @@ class EDS extends DefaultRecord
             'BibRecord/BibRelationships/HasContributorRelationships/*/'
                 . 'PersonEntity/Name/NameFull'
         );
-        return array_unique(array_filter($authors, function ($a) {
-            return !empty($a);
-        }));
+        return array_unique(array_filter($authors));
     }
 
     /**
--- a/module/VuFind/src/VuFind/RecordDriver/EDS.php
+++ b/module/VuFind/src/VuFind/RecordDriver/EDS.php
@@ -460,12 +460,13 @@ class EDS extends DefaultRecord
      */
     public function getPrimaryAuthors()
     {
-        return array_unique(
-            $this->extractEbscoDataFromRecordInfo(
-                'BibRecord/BibRelationships/HasContributorRelationships/*/'
+        $authors = $this->extractEbscoDataFromRecordInfo(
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/RecordDriver/EDS.php
```diff
@@ -460,12 +460,13 @@ public function getTitle()
...
     {
-        return array_unique(
-            $this->extractEbscoDataFromRecordInfo(
-                'BibRecord/BibRelationships/HasContributorRelationships/*/'
+        $authors = $this->extractEbscoDataFromRecordInfo(
+            'BibRecord/BibRelationships/HasContributorRelationships/*/'
                 . 'PersonEntity/Name/NameFull'
-            )
         );
+        return array_unique(array_filter($authors, function ($a) {
```
demiankatz: I believe you can simplify this to simply `return array_unique(array_filter($authors));` -- the default behavior of array_filter is to filter out empty elements, so the callback should not be required in this situation. Do you mind giving it a try and updating the PR if it works?

demiankatz (commented):
Thanks! I've just suggested one simplification. I also wonder if it would make sense to do the filtering in extractEbscoDataFromRecordInfo() rather than in getPrimaryAuthors() so that we never pass back empty elements. I can't think of a use case where having empty data would be desirable, though I would have to review the code more carefully to be sure I'm not forgetting something.

module/VuFind/src/VuFind/RecordDriver/EDS.php
janhejzl: Thanks, yes, it works. I have just updated that. As for the placement inside the `getPrimaryAuthors()`, the lower function, `extractEbscoDataFromRecordInfo()`, is used many times in the whole class but this is the only place where its output is deduplicated this way.  Of course, I will be happy with any addition.

demiankatz: Great, thank you! I've also reviewed all the uses of `extractEbscoDataFromRecordInfo` and I agree that it currently makes the most sense to leave this fix where you put it. I'll backport this to the release-7.0 branch so the fix can be included in next week's 7.0.2 release.

demiankatz:
Successfully backported to release-7.0 in time for the 7.0.2 release. Thanks again!

PR #1803, opened 2020-12-04
[Release 7.0] Account ajax fixes

EreMaijala:
This fixes a couple of issues with account ajax.

The first issue with the element check only comes up when the `.fines-status` element is not available but something else is so that _render is called. With the bad check it would call render() method for fines without proper data, and since the render method checks for `if (status.value === 0)` the check is false and icon level becomes DANGER.


demiankatz:
Thanks, and just in time, because I was going to start prepping release 7.0.2 in a couple of minutes. These are simple enough fixes that I'm comfortable including them at the last minute. :-)

EreMaijala:
Nice. :) I refrained from more extensive tweaks, those will be coming later. :)

PR #1804, opened 2020-12-04
Add support for clearing account cache based on status information. U…

EreMaijala:
…se for fines.

We have received several complaints that the cached information remains after the user has paid the fines or staff has written it off. This helps a bit by clearing the cache if it seems the information is no longer correct. If this mechanism is found useful, I can extend it to other functions too.


demiankatz:
Thanks, @EreMaijala, I think this makes sense but I'd be interested to get @crhallberg's feedback as well.

crhallberg (approved):
Looks good! Just to make sure I'm reading this correctly, this correction is only triggered on the fines page, right? That is not an issue, I just wanted to make sure I'm understanding how this all works.

EreMaijala:
@crhallberg Yes, that's correct. I wanted to put this out for comment. I'll add loans, holds etc. either here or in another pull request.

demiankatz:
Thanks for the review, @crhallberg. @EreMaijala, do you have a preference for getting this merged now or leaving this open so you can add additional functionality? I'm fine either way (though in either case, I still need to carve out some time for a final review before merging, since my time remains somewhat limited this month).

EreMaijala:
@demiankatz I'll try to add more checks today, so maybe it's better to wait before merging.

EreMaijala:
In an unexpected moment of clarity I realized it's much better to collect the required information in the controller. Latest iteration now supports updates for all the built-in notifications. I also added a simple way to reset the cache in Demo driver to make testing easier. So now you can call e.g. /MyResearch/Holds?cleardemo and it will reset the cache. This can be left out from the merge since it uses the global $_REQUEST.

demiankatz:
Thanks, @EreMaijala! At this point, my biggest concern is that we're now duplicating significant amounts of logic between the AJAX handlers and the (already unwieldy) MyResearchController. This might be a stupid idea, and I only took a quick glance to check for its feasibility, but what if we refactored the AJAX handlers so that they contain public methods that convert the raw user data into the formatted AJAX response, and then we call the handlers directly in the MyResearchController to process the data? That way, if we need to change something, we don't run the risk of forgetting to synchronize two sets of identical business logic.

I also agree that we shouldn't merge the demo driver change with `$_REQUEST` in it, but it actually wouldn't be too hard to just inject the Laminas request object to the demo driver through its factory, and then do the same check in a framework-approved way. I can throw that together as a separate pull request early next year if you don't have time or inclination in the meantime.

EreMaijala:
@demiankatz That crossed my mind, but I'd like to keep them separate as I have a long-term dream of adding support for optimized retrieval of the required information from the ILS so that the driver wouldn't necessarily do all the work to construct e.g. a complete checkouts list.

~~I'd appreciate it if you could do the Demo changes. I'm running out of time for this year, and won't be back until 4th January. If you don't get to do it by then, I'll see to it.~~

EreMaijala:
@demiankatz I added a cleaner way of clearing the Demo driver cache as suggested and rebased.

demiankatz:
@EreMaijala, what do you think about cc1c1b3? This makes Demo's dependency on HttpRequest optional, so the tests no longer need to be changed, and backward compatibility is greater. This small change makes me a bit more comfortable incorporating this into release 7.1 rather than 8.0. Any objections?

EreMaijala:
@demiankatz Looks good, no objections.

demiankatz:
Thanks, @EreMaijala, I've merged the demo driver changes to dev (and then back to here) to simplify this PR. Did you want to try to get the rest of this done for release 7.1, or should this wait for 8.0? (I confess that I haven't taken the time to refresh my memory on this PR yet, but I can take a closer look soon).

EreMaijala:
@demiankatz It would be nice to have in 7.1 but I'm fine with 8.0 as well. This is now feature-complete so it should be good to go, but if you feel it's too risky, getting it to dev-8.0 soon would be perfectly fine for us.

demiankatz:
This looks fine to me, and all tests are passing, so I am merging it now.

PR #1806, opened 2020-12-10
Make it possible to call renderTemplate for optional templates.

EreMaijala:
This allows renderTemplate to be called even if it's not certain that the given template exists.

We have some driver-specific templates that only need to be rendered if they exist, but the rendering is done by a template shared between all the drivers.

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/ClassBasedTemplateRendererTrait.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/ClassBasedTemplateRendererTrait.php
@@ -66,13 +66,12 @@ trait ClassBasedTemplateRendererTrait
      * (or null if $className is already the top level; used for recursion only).
      *
      * @return string
-     * @throws RuntimeException
      */
     protected function resolveClassTemplate($template, $className,
         ResolverInterface $resolver, $topClassName = null
     ) {
         // If the template resolves, return it:
-        $templateWithClass = sprintf($template, $this->getBriefClass($className));
+        $templateWithClass = $this->getTemplateWithClass($template, $className);
         if ($resolver->resolve($templateWithClass)) {
             return $templateWithClass;
         }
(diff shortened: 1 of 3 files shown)


module/VuFind/src/VuFind/View/Helper/Root/ClassBasedTemplateRendererTrait.php
```diff
@@ -101,19 +96,34 @@ protected function resolveClassTemplate($template, $className,
...
+        $throw = true
+    ) {
         // Set up the needed context in the view:
         $view = $this->getView();
         $contextHelper = $view->plugin('context');
         $oldContext = $contextHelper($view)->apply($context);
 
         // Find and render the template:
-        $html = $view->render($this->getCachedClassTemplate($template, $className));
+        $classTemplate = $this->getCachedClassTemplate($template, $className);
```
demiankatz, line 114: Maybe I'm not reading the code closely enough, but one concern: by moving the throw logic here, do we run the risk that if a template exists but contains no content, it will resolve as false and cause an exception to be thrown even when we do not want one? Would it be safer to keep the exception-throwing where it is in the current code, but wrap it in a try..catch here that returns empty string when $throw is false?

demiankatz (commented):
This looks like a useful addition -- but one question before I merge!

module/VuFind/src/VuFind/View/Helper/Root/ClassBasedTemplateRendererTrait.php
EreMaijala: getCachedClassTemplate just returns the template name, or empty string if it doesn't exist. The actual rendering happens in the next step, `$html = $view->render($classTemplate);`. But I see one problem now, it's not restoring the context properly. Fix coming up.

EreMaijala: Fix done. So to iterate, getCachedClassTemplate returns the template name, if found. Then we check if it's empty and an exception is needed. Otherwise render to $html if template exists, otherwise set it to an empty string.

demiankatz: Thanks!

PR #1807, opened 2020-12-10
Configurable top facet override of alphabetical sort for limits

bbusenius:
# Problem

We need to display our top 8 most frequently used languages at the top of the advanced search limit box. 

image

In order to achieve this, this pull request allows facets to be configured with an override that will place them at the top of the advanced search limit box (See Language in the image above). This implementation allows the static display order to be determined by the configuration. All facets not included in the config are displayed in their natural alphabetical sort order following the static ones from the config. This only works with regular facets, it does not work with hierarchical facets.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Base/Options.php
+++ b/module/VuFind/src/VuFind/Search/Base/Options.php
@@ -982,4 +982,21 @@ abstract class Options implements TranslatorAwareInterface
         $this->autocompleteAutoSubmit = $searchSettings->Autocomplete->auto_submit
             ?? $this->autocompleteAutoSubmit;
     }
+
+    /**
+     * Get advanced search limits that override the natural sorting to
+     * display at the top.
+     *
+     * @param string $limit advanced search limit
+     *
+     * @return array
+     */
+    public function topLimits($limit)
+    {
+        $facetSettings = $this->configLoader->get($this->getFacetsIni());
+        $limits = $facetSettings->Advanced_Settings->limit;
+        $limitConf = $limits && $limits->get($limit) ?
+            preg_replace('/\s+/', '', $limits->get($limit)) : '';
+        return explode(',', $limitConf);
+    }
 }
--- a/module/VuFind/src/VuFind/Search/Base/Options.php
+++ b/module/VuFind/src/VuFind/Search/Base/Options.php
@@ -995,8 +995,10 @@ abstract class Options implements TranslatorAwareInterface
     {
         $facetSettings = $this->configLoader->get($this->getFacetsIni());
         $limits = $facetSettings->Advanced_Settings->limit;
+        $delimiter = $facetSettings->Advanced_Settings->limitDelimiter ?
+            $facetSettings->Advanced_Settings->limitDelimiter : '::';
         $limitConf = $limits && $limits->get($limit) ?
             preg_replace('/\s+/', '', $limits->get($limit)) : '';
-        return explode(',', $limitConf);
+        return explode($delimiter, $limitConf);
     }
 }
--- a/module/VuFind/src/VuFind/Search/Base/Options.php
+++ b/module/VuFind/src/VuFind/Search/Base/Options.php
@@ -994,9 +994,8 @@ abstract class Options implements TranslatorAwareInterface
     public function limitOrderOverride($limit)
     {
         $facetSettings = $this->configLoader->get($this->getFacetsIni());
-        $limits = $facetSettings->Advanced_Settings->limitOrderOverride;
-        $delimiter = $facetSettings->Advanced_Settings->limitDelimiter ?
-            $facetSettings->Advanced_Settings->limitDelimiter : '::';
+        $limits = $facetSettings->Advanced_Settings->limitOrderOverride ?? null;
+        $delimiter = $facetSettings->Advanced_Settings->limitDelimiter ?? '::';
         $limitConf = $limits ? $limits->get($limit) : '';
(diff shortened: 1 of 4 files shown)


module/VuFind/src/VuFind/View/Helper/Root/Config.php
```diff
@@ -88,4 +88,20 @@ public function ajaxCoversEnabled()
...
+
+    /**
+     * Get advanced search limits that override the natural sorting to
+     * display at the top.
+     *
+     * @param string $limit advanced search limit
+     *
+     * @return array
+     */
+    public function topLimits($limit)
```
demiankatz: I think it would be better to tie this functionality to the Search\Base\Options object rather than to the Config helper. This way, we can expand the feature to other search backends in the future if we need to.

module/VuFind/src/VuFind/View/Helper/Root/Config.php
```diff
@@ -88,4 +88,20 @@ public function ajaxCoversEnabled()
...
+    {
+        $limits = $this->get('facets')->Advanced_Settings->limit;
+        $limitConf = $limits && $limits->get($limit) ?
+            preg_replace('/\s+/', '', $limits->get($limit)) : '';
+        return explode(',', $limitConf);
```
demiankatz: It might be wise to have a configurable delimiter, or to take a different approach to multi-valued configuration, since it's not uncommon for facet values to include commas, and that would prevent certain facet fields (like author) from working well with this new feature.

demiankatz (commented):
Thanks for this! Seems like a useful feature to me, but I have a couple of suggestions for making it more flexible and future-proof. I'm happy to help with implementation details if necessary, though it may be a couple of weeks before I have a chance to dive in.

module/VuFind/src/VuFind/View/Helper/Root/Config.php
bbusenius: Yes, absolutely!

module/VuFind/src/VuFind/View/Helper/Root/Config.php
bbusenius: Great idea! I changed the default delimiter to `::` and made it configurable.

bbusenius:
Thanks for the great suggestions @demiankatz! I took a stab at making the changes myself. I'm only "in the office" Wednesday and Thursday of next week and then out for awhile after that, so my availability will be fairly curtailed too. There's no urgency to this so we have plenty of time to iron it out.

demiankatz:
Thanks, @bbusenius, your changes look great. Just one more thought before we finalize this: I think `limit` might be a little too broad a name for the setting... would something like `limitOrderOverride` or `limitPrioritizedValues` (or a better option of your own choosing) be preferable?

bbusenius:
Thanks @demiankatz! Yes, I agree about the config name. I had concern over this and the function name in `Options.php`. I like `limitOrderOverride`, so I went with that. Hopefully it's not a problem that I used the same name for the function in `Options.php`. I think this reads a lot better and is more explanatory than `topLimits` was when it's used in `solr.phtml`.

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -982,4 +982,23 @@ protected function configureAutocomplete(Config $searchSettings = null)
...
+     * @return array
+     */
+    public function limitOrderOverride($limit)
+    {
+        $facetSettings = $this->configLoader->get($this->getFacetsIni());
+        $limits = $facetSettings->Advanced_Settings->limitOrderOverride;
+        $delimiter = $facetSettings->Advanced_Settings->limitDelimiter ?
+            $facetSettings->Advanced_Settings->limitDelimiter : '::';
+        $limitConf = $limits && $limits->get($limit) ?
+            preg_replace('/\s+/', '', $limits->get($limit)) : '';
```
demiankatz: What is the purpose of this preg_replace? It seems to be preventing facet values containing spaces from being priorities (I tried, for example, to move "Conference Proceeding" to the top of the format list, but it doesn't work).

demiankatz: As an alternative, how would you feel about something like:

<pre>
$limitConf = $limits ? $limits->get($limit) : '';
return array_map('trim', explode($delimiter, $limitConf));
</pre>

demiankatz (commented):
Thanks for the revisions; I think this is greatly improved! I just ran into one, hopefully minor, issue that I'd like your feedback on...

module/VuFind/src/VuFind/Search/Base/Options.php
bbusenius: Good catch! Yes, the purpose was just to trim. This was actually pretty sloppy now that you point it out. I've added the patch.

demiankatz: Thanks!

demiankatz:
@bbusenius, I just opened https://github.com/uchicago-library/vufind/pull/107 to make a couple of nearly-trivial optimizations to this (when I see an excuse to use a spaceship operator, I just have to leap on it). If you can merge that, then I think this can be merged on the VuFind side!

(In theory it would also be nice to add support for this feature to other search backends, like Summon and EDS, but that's labor-intensive for EDS, and I don't have Summon access for testing, so it's probably best to leave it alone for now until an actual need arises ... I can open a JIRA ticket to track the need if you think it's worthwhile).

PR #1809, opened 2020-12-14
KohaRest: Check that item_id is set before sending it to Koha.

EreMaijala:
It might not be if it's missing from HMACKeys.


demiankatz:
Thanks!

PR #1810, opened 2020-12-16
Enable ES6 in eslint

xmorave2:
According to Can I use page, ES6 is heavily adopted in browsers now: https://caniuse.com/?search=es6

Some ES6 features could be very useful (const, let, template strings, ...)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,11 +2,8 @@ module.exports = {
   extends: "eslint:recommended",
   env: {
     "browser": true,
-    "es6": true
-  },
-  globals: {
-    "$": false,
-    "jQuery": false
+    "es6": true,
+    "jquery": true
   },
   rules: {
     // errors
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,9 @@
 module.exports = {
   extends: "eslint:recommended",
-  env: { "browser": true },
+  env: {
+    "browser": true,
+    "es6": true
+  },
   globals: {
     "$": false,
     "jQuery": false


demiankatz:
Thanks, @xmorave2. Some of this may tie in with recent conversation on #1791 about backticks. I'd be interested to hear @crhallberg's thoughts on this.

I also suggest that if we do start using more ES6 features, we should update our [browser support statement](https://vufind.org/wiki/installation:requirements#browser_support) appropriately... and we should probably target these changes to release 8.0, since it will likely impact our compatibility with older browsers.

xmorave2:
New browser support after this change will be https://caniuse.com/es6

- dropped support for IE
- support Edge 15+ (released 11. 4. 2017)
- support Chrome 51+ (25. 5. 2016)
- support Firefox 54+ (13. 6. 2017)
- support Opera 38+ (8. 6. 2016)
- Safari 10+ (13. 9. 2016)

partial support even on earlier versions and IE 11

crhallberg:
I'd like to be able to use the full functionality of ES6 going forward so I think this change is timely, even if it doesn't have a large impact on the current themes.

crhallberg:
Also to add a bit of context to the new browser requirements, ES6 support covers over 97% of users:
image

demiankatz:
I'm merging this now; I've updated the [browser support statement](https://vufind.org/wiki/installation:requirements#browser_support), but please let me know if you think the text there should be adjusted.

demiankatz:
@crhallberg, on a related note, are there any polyfills or other configurations we can safely drop when raising the requirements?

PR #1811, opened 2021-01-05
FeedbackForms: require one checkbox option to be selected

samuli:
Currently enabling `required` setting for a checkbox element makes the validation require all options to be selected.

This PR adds `requireOne` setting, that loosens the validation by requiring that at least one (but not necessarily all) options is selected. 

The changes only affect the client-side/browser validation by leaving `require` attributes out from the option-elements. Backend validation still uses `Laminas\Validator\NotEmpty`, that only verifies that some options have been selected.

### DIFF
--- a/config/vufind/FeedbackForms.yaml
+++ b/config/vufind/FeedbackForms.yaml
@@ -47,16 +47,18 @@
 #
 #   fields (array) List of form elements with the following options:
 #
-#     name (string)      Element name
-#     label (string)     Element label (translation key)
-#     required (boolean) Is the element required?
-#     settings (array)   HTML attributes as key-value pairs, for example:
+#     name (string)        Element name
+#     label (string)       Element label (translation key)
+#     required (boolean)   Is the element required? (for checkbox elements this means that
+#                          all options have to be selected.)
+#     requireOne (boolean) Require at least one checkbox option to be selected.
+#     settings (array)     HTML attributes as key-value pairs, for example:
 #       - [class, "custom-css-class another-class"]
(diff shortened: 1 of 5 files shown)


demiankatz:
Thanks! This all makes sense to me. Only one question: is it worth strengthening the back end validation code to be more consistent with the front end validation code? Otherwise, it sounds like in the case of a required set of checkboxes, a user could manipulate the browser to bypass the validation and successfully submit the form without actually checking all of the boxes. I could imagine a required set of checkboxes being useful in a situation where the user must acknowledge a set of terms/conditions before submitting a form. While this kind of thing doesn't seem very likely to happen in a real-world VuFind use case, and just checking some boxes is a lot easier than hacking your way around them, it still might be worth fixing if there's a relatively simple solution, just in the interest of providing secure and consistent code.

(If it's a lot of work or adds a lot of complexity, I could be persuaded to merge this as-is, however). :-)

samuli:
I agree that it's worth to improve also the backend side of the validation. Maybe `\Laminas\Validator\InArray` could be used for this. However, I think this could be merged as-is, and I could try to investigate this further later on.

demiankatz:
Sounds good to me. Thanks, @samuli! I will merge this now. Would you like me to open a JIRA ticket about the validation so we can keep track of it in the long term, or do you expect to work on it soon enough that we don't need to do that? Since this is not at all urgent, I'm happy to take the JIRA route if that's easiest.

samuli:
@demiankatz Continued here: https://github.com/vufind-org/vufind/pull/1814

PR #1813, opened 2021-01-06
makes MARC record field configurable

olli-gold:
We are using a central index provided by the Common Library Network in Germany for VuFind. The Common Library Network wants to introduce a new index field to hold the MARC records in MARCXML instead of raw binary MARC. But they do not want to import the content into the default field in VuFind (fullrecord), as there are very different clients using this index and such a change could cause problems with some of the other clients. That's why they want to introduce a new index field for that. VuFind could deal with MARCXML in the fullrecord field, but it can't easily use a different field to hold the full MARC record.

This PR makes the field with the full MARC record configurable by introducing a new configuration option in 'config.ini'. If this new option is not set, VuFind will keep fullrecord as the default field. I have implemented a check, if the field in the configuration really exists in the record; if not, VuFind will fallback to the default field. This should help avoiding any misconfiguration issues.

It would help us (and probably other VuFind users with the index by Common Library Network), if this PR could be merged into VuFind main and I would not expect it to break anything. But I would also understand, if this would be considered as too specific, though...

### DIFF
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -1721,11 +1721,14 @@ jump_to_single_search_result = false
 ; punctuation, but this can be used when ISBD punctuation is absent (e.g. ", ").
 ;marcPublicationInfoSeparator = " "
 
-; If you have a custom index, you might want to change the field
-; where the full MARC record is supposed to be.
-; You can do that with this configuration setting.
+; If you have a custom index, you might want to change the field where the full
+; MARC record is supposed to be. The field will be checked and ignored, if it is
+; not a valid index field or does not exist in the record or index schema.
+; If you have multiple fields with MARC content, you may add all of them
+; by using a comma sepated list. The first field is the preferred one, if it does not
+; exist, the next ones are taken.
 ; (Default = "fullrecord")
-preferredMarcField = "fullrecord"
+preferredMarcFields = "fullrecord"
 
 ; When displaying publication information from 260/264, this can be set to true
 ; to make 264 information completely replace 260 information. Default is false,
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -1721,6 +1721,12 @@ jump_to_single_search_result = false
 ; punctuation, but this can be used when ISBD punctuation is absent (e.g. ", ").
 ;marcPublicationInfoSeparator = " "
 
+; If you have a custom index, you might want to change the field
+; where the full MARC record is supposed to be.
+; You can do that with this configuration setting.
+; (Default = "fullrecord")
+preferredMarcField = "fullrecord"
+
 ; When displaying publication information from 260/264, this can be set to true
 ; to make 264 information completely replace 260 information. Default is false,
 ; which will display information from 260 AND 264 when both fields are populated.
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/RecordDriver/MarcReaderTrait.php
```diff
@@ -58,7 +58,12 @@ trait MarcReaderTrait
     public function getMarcRecord()
     {
         if (null === $this->lazyMarcRecord) {
-            $marc = trim($this->fields['fullrecord']);
+            // Get preferred MARC field from config, if it is set and is existing:
+            $marcField = (isset($this->mainConfig->Record->preferredMarcField)
```
demiankatz: I think this might be a little more readable in two lines as:

<pre>
$preferredMarcField = $this->mainConfig->Record->preferredMarcField ?? 'fullrecord';
$marc = trim($this->fields[$preferredMarcField] ?? $this->fields['fullrecord']);
</pre>

config/vufind/config.ini
```diff
@@ -1721,6 +1721,12 @@ jump_to_single_search_result = false
 ; punctuation, but this can be used when ISBD punctuation is absent (e.g. ", ").
 ;marcPublicationInfoSeparator = " "
 
+; If you have a custom index, you might want to change the field
+; where the full MARC record is supposed to be.
+; You can do that with this configuration setting.
```
demiankatz: It might be worth adding a comment here to note that the preferred field will only be used if it is present in the index, and otherwise fullrecord will be used.

demiankatz (changes requested):
This seems reasonable to me; I've just left a couple of comments suggesting minor improvements. These comments assume that we want to keep the functionality exactly as implemented here, but I'd suggest one other possibility to consider: what if you instead make the setting `preferredMarcFields` and make it a comma-separated list of fields to search for the MARC record? That would add a little bit more flexibility, if there are more than two possible fields to consider.

module/VuFind/src/VuFind/RecordDriver/MarcReaderTrait.php
olli-gold: Agreed. I will incorporate this in the next version.

config/vufind/config.ini
olli-gold: Agreed. I will incorporate this in the next version.

olli-gold:
> what if you instead make the setting `preferredMarcFields` and make it a comma-separated list of fields to search for the MARC record? That would add a little bit more flexibility, if there are more than two possible fields to consider.

That's a good suggestion. I will review the PR again and incorporate all of your advices. Thank you!

olli-gold:
OK, the requested changes including the improvement suggestion have been implemented now and after a style fix all checks have passed, so it's ready for another review. Thank you!

demiankatz:
Thanks, @olli-gold, this looks good! I'm out of the office for the rest of the week, but I'll give this a final-double check next week and then get it merged.

demiankatz:
Thanks again -- this passes all my tests, so I'm merging it now!

PR #1814, opened 2021-01-07
Checkbox validators

samuli:
Continued from https://github.com/vufind-org/vufind/pull/1811

Improve server-side validation of  checkbox elements:
- when `required` setting is true, posted options are validated with `\Laminas\Validator\Identical`. The validation fails if any of the configured options are missing or if not configured options are posted.
- when `requireOne` setting is true, posted option(s) are compared to the configured options using a custom callback (`InArray`can't be used here, since it doesn't work when the needle is an array).

Both cases are also validated with `notEmpty`.

### DIFF
--- a/module/VuFind/src/VuFind/Form/Form.php
+++ b/module/VuFind/src/VuFind/Form/Form.php
@@ -28,10 +28,10 @@
 namespace VuFind\Form;
 
 use Laminas\InputFilter\InputFilter;
-use Laminas\Validator\EmailAddress;
-use Laminas\Validator\NotEmpty;
-use Laminas\Validator\Identical;
 use Laminas\Validator\Callback;
+use Laminas\Validator\EmailAddress;
+use Laminas\Validator\Identical;
+use Laminas\Validator\NotEmpty;
 use Laminas\View\HelperPluginManager;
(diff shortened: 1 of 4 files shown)


module/VuFind/src/VuFind/Form/Form.php
```diff
@@ -759,14 +761,41 @@ public function getInputFilter()
         ];
 
         foreach ($this->getElements() as $el) {
-            $required = $el['type'] === 'checkbox'
-               ? $el['required'] ?? $el['requireOne'] ?? false
-               : $el['required'] ?? false;
+            $isCheckbox = $el['type'] === 'checkbox';
+            $required = $el['required']
+                ?? ($isCheckbox && ($el['requireOne'] ?? false));
+            $requireOne = $isCheckbox && ($el['requireOne'] ?? false);
```
demiankatz, line 765: Rather than doing this calculation twice, what if you rearranged it to:

<pre>
$requireOne = $isCheckbox && ($el['requireOne'] ?? false);
$required = $el['required'] ?? $requireOne;
</pre>

demiankatz (commented):
Thanks for the progress on this -- just one minor suggestion.

PR #1815, opened 2021-01-07
Refactor expiration commands for improved performance.

EreMaijala:
The old mechanism had performance issues when there were gaps in the tables.

This also simplifies the code so that the maintenance interface uses the same calls as the console utilities.


module/VuFind/src/VuFind/Db/Table/ExpirationTrait.php
```diff
@@ -49,61 +48,51 @@ trait ExpirationTrait
...
-    public function deleteExpired($daysOld = 2, $idFrom = null, $idTo = null)
+    public function deleteExpired($daysOld, $limit = null)
     {
-        // Determine the expiration date:
-        $callback = function ($select) use ($daysOld, $idFrom, $idTo) {
-            $this->expirationCallback($select, $daysOld, $idFrom, $idTo);
+        // Determine the expiration parameters:
+        $lastId = $limit ? $this->getExpiredBatchLastId($daysOld, $limit) : null;
+        $callback = function ($select) use ($daysOld, $lastId) {
+            $this->expirationCallback($select, $daysOld, $lastId);
```
demiankatz: Why are you passing $lastId here when expirationCallback() only appears to accept two parameters?

module/VuFind/src/VuFind/Db/Table/AuthHash.php
```diff
@@ -101,45 +101,17 @@ public function getLatestBySessionId($sessionId)
...
      * @param int    $daysOld Age in days of an "expired" record.
-     * @param int    $idFrom  Lowest id of rows to delete.
-     * @param int    $idTo    Highest id of rows to delete.
      *
      * @return void
      */
-    protected function expirationCallback($select, $daysOld, $idFrom = null,
-        $idTo = null
-    ) {
+    protected function expirationCallback($select, $daysOld)
```
demiankatz, line 112: Unless I'm missing something, all of the expirationCallback() functions are now more or less identical except for the name of the field that we're comparing the dates against. Could we move this function into the trait and use a $this->expirationDateComparisonField property to the classes (which could default to 'created', perhaps) in order to further simplify and reduce redundant code?

demiankatz (commented):
This seems like a good approach, but I had a couple of questions....

module/VuFind/src/VuFind/Db/Table/ExpirationTrait.php
EreMaijala: Oops, that's a leftover from moving the $lastId handling from the callback to the trait. I'll remove it.

module/VuFind/src/VuFind/Db/Table/AuthHash.php
EreMaijala: That'd be nice, but we have different kinds of fields to compare. E.g. the session table has a unix timestamp, while search has a datetime. There can also be additional rules, like with search table where the row must also have saved=0.

demiankatz: Ahh, of course! Too bad. We could still probably make the code less redundant with a default implementation in the trait, but I can see how that could actually make things more confusing, so I agree that it's best to leave it alone.

PR #1817, opened 2021-01-08
Add support for course reserves with KohaRest driver.

EreMaijala:


EreMaijala:
@demiankatz I've tested this the best I can, but it would be great if someone else could verify that things work as expected since we don't use course reserves.

demiankatz:
@EreMaijala, should 3631237 be a separate PR? It seems to have nothing to do with course reserves (and I'm not sure exactly what its purpose is ... might be worth discussing independently).

EreMaijala:
@demiankatz Sorry, I failed to describe the change. Since console utils currently run without theme support, we don't have SafeMoneyFormat available when indexing course reserves with util/index_reserves. This was the smallest change I could think of to make the util work properly.

In the long run it might be useful, and not too complicated, to enable theme support in utils (we already do that locally for some of our console tasks, so it's working pretty well), but that can be discussed separately.

demiankatz:
Thanks, @EreMaijala, I understand now. I was confused because I didn't see the correlation between course reserves and money formatting, but now I see the point. :-)

Would you like me to cherry-pick 3631237 onto the dev-8.0 branch and then re-point this PR there in order to focus this back on the actual course reserves functionality?

EreMaijala:
@demiankatz Sure, thanks. I just included it here since that commit alone is even less meaningful-looking. :)

demiankatz:
@EreMaijala, your comment about the SafeMoneyFormat commit being unclear led me to expand the comment in the factory... but then, when I thought about it a little more, I began to wonder if this is really the best approach. What if you just make SafeMoneyFormat an optional dependency, pass null from the factory if the helper is unavailable, and implement the formatMoney() method to throw an exception if the helper is missing? I think that would have the same effect while making the code a little more straightforward (though admittedly, the current implementation has the advantage of only constructing the helper when it is needed, which is a little more efficient). What do you think? I'm willing to merge as-is, but wanted to offer this alternative before committing to the current approach.

EreMaijala:
@demiankatz I'm fine with either approach, but is there a way to check if the helper exists other than catching an exception? I'd like to avoid the exception in normal circumstances.

demiankatz:
@EreMaijala, yes, you should be able to do something like:

<pre>
$helperManager = $container->get('ViewHelperManager');
$safeMoneyFormat = $helperManager->has('safeMoneyFormat')
    ? $helperManager->get('safeMoneyFormat') : null;
</pre>

Again, if you don't think this is a meaningful improvement, I'm not trying to make more work for you... but it feels to me like it could be slightly cleaner. :-)

EreMaijala:
@demiankatz It's better as you proposed. Now changed, ok?

demiankatz:
Thanks, @EreMaijala, I just made some small adjustments in 20e7fcf to match the PHPdoc with the actual constructor signature and to rearrange some properties that got reordered in the refactoring (just to cut down on unnecessary diffs that might reduce readability of history). Any objections? At this point, if you think there's an advantage to doing so, I'd be reasonably comfortable merging this into the 7.1 release with a changelog note (though it's still probably more technically correct to wait for 8.0, since the signature change is a BC break, even if a nearly-trivial one).

EreMaijala:
@demiankatz Thanks, no objections. If you think the change is small enough, I'd get it in 7.1 to make any further fixes etc. easier, but I'll leave the final call to you.

demiankatz:
Thanks, @EreMaijala, I've gone ahead and merged the SafeMoneyFormat refactoring to dev. Please let me know when you are ready to merge the remaining course reserves functionality. If I hear any further feedback from the Koha users I shared this with, I'll let you know.

EreMaijala:
@demiankatz I'm ready. There's not much more I can do apart from waiting for some bug reports. :)

demiankatz:
Sounds good, @EreMaijala, I've just merged it. Hopefully this will encourage more testing! As you say, it looks like most of the complexity is on the API side rather than the VuFind side in any case.

PR #1818, opened 2021-01-08
Fix 2 undefined variables in search/filters.phtml

mtrojan-ub:
This problem exists since version 6.1

### DIFF
--- a/themes/bootstrap3/templates/search/filters.phtml
+++ b/themes/bootstrap3/templates/search/filters.phtml
@@ -10,9 +10,7 @@
   $lastSort = $this->searchMemory()->getLastSort($this->searchClassId);
   $options = $this->searchOptions($this->searchClassId);
   $hasDefaultsApplied = $params->hasDefaultsApplied();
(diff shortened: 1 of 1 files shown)


EreMaijala:
@mtrojan-ub Both variables should have been relayed by the caller (they could (should?) have been written `$this->checkboxFilters` and `$this->filterList`). See e.g. https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/search/searchbox.phtml#L144

EreMaijala:
@mtrojan-ub To add to that it's done so that the caller has control over whether the filters are displayed.

mtrojan-ub:
Thanks for the hint! I changed it to $this->... to keep it more consistent regarding to the others (like $this->params).

demiankatz:
Thanks! In most situations, the framework should automatically extract all variables from $this so they can be referred to with or without the prefix... but when including sub-templates, it can get more complicated, and it seems like this may be just such a case. Explicitly using $this is always the more reliable (if less enjoyable to read) approach.

PR #1581, opened 2020-03-10
[VUFIND-120] Add support for record "versions" display.

EreMaijala:
Here is an initial implementation of "FRBR" for VuFind. This is based on work identification keys in Solr index.

About naming: I've used the term "work keys" and similar in the backend code, and "versions" in the front end code. This can be revisited if necessary. Like everything here.

Also included is sample indexer configuration for creating work keys. This should be good to go. It's not perfect (in this case I don't think anything will be), but should be at least a good starting point. 

This allows versions to work whether the new Versions tab is enabled or not. If not, versions are always displayed as a search.

Basic principles:

- Uniform titles work as keys alone
- Other titles need author information with them

I won't be able to work on this much more at the moment due to my leave, but feel free to pick up if you find this interesting enough. :)

TODO:

- [x] Should the search link actually contain the record id instead of keys? This would need extra handling in search parameters code, and would also mean that if the record is no longer available, the search wouldn't work either.
- [x] Make the search terms look nicer
- [x] Determine whether or not to turn on by default (currently awaiting vufind-tech feedback until November 18)
- [x] Make indexing ignore non-filing characters
- [x] Make indexing fold the characters to ascii instead of quietly dropping them (see [MetadataUtils.php](https://github.com/NatLibFi/RecordManager/blob/master/src/RecordManager/Base/Utils/MetadataUtils.php#L106))?
- [x] Port work keys logic to PHP so it can be used in XML imports as well.
- [x] Add work keys logic to all appropriate example XSLTs/properties files.
- [x] Test everything (be sure to include testing of Search2 backend to be sure it behaves consistently)
- [x] Do we want to include alternate script support (880 field in MARC 21)? No, not now at least.
- [x] Document in [wiki](https://vufind.org/wiki/configuration:record_versions).
- [x] Resolve #1729 first.
- [x] Stay in versions search when clicking the search button in searchbox if retain_filters_by_default is set
- [x] Fix slow SolrMarc indexing performance, and apply fix to both Unix and Windows import-marc scripts
- [x] Resolve [VUFIND-120](https://vufind.org/jira/browse/VUFIND-120) when merging.

PR #1819, opened 2021-01-12
Explicitly initialize OpenURL embedding in ajax-loaded record tabs.

EreMaijala:
This should fix any timing issues with binding to OpenURL elements as reported in https://github.com/vufind-org/vufind/pull/1581#issuecomment-758250727.


EreMaijala:
Note that as far as I can see it's safe to call openurl's init method twice.

demiankatz:
I agree, this seems pretty safe. I think there might be potential for Hunt to get initialized more than once if init is called repeatedly, but since Hunt is not loaded in the context where record tabs are getting initialized, I think it should be safe in this case.

PR #1820, opened 2021-01-12
Add transEscAttr view helper and use it in templates.

EreMaijala:
While transEsc alone usually works fine, it doesn't hurt to do things as they should, and this shortens the translate+escapeHtmlAttr chain in templates.

TODO:
- [x] Run full test suite


themes/bootstrap3/templates/search/searchbox.phtml
```diff
@@ -1,3 +1,4 @@
+<a href="FOO" title="<?=$this->transEsc('"Phoo"')?>">Pho</a>
```
demiankatz: I'm not sure what this is, but I'm guessing it's not supposed to be here. :-)

demiankatz (changes requested):
This looks good apart from one bit of code that appears to have been added by accident. If time permits later, I may just remove that myself. :-) I'll also run the full test suite on this to be 100% sure everything still works before merging, but a careful code review suggests to me that it should be fine.

themes/bootstrap3/templates/search/searchbox.phtml
EreMaijala: Don't you like it? :D Err.. I'll remove it.

demiankatz: Thanks! All tests are (unsurprisingly) passing, so I'll merge this soon.

PR #1821, opened 2021-01-12
Update breadcrumbs for the record and collection pages.

EreMaijala:
These changes are modeled after https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html, so they're keeping the last aria-current item as a link.

The change to record/view.phtml is pretty straight-forward, but the change to collection/view.phtml tries to also improve it by not including both search and collections home. I'm not quite sure if that works properly or if search always gets added if there's a previous search during the session. I don't have working alphabrowse so I can't actually test that.


demiankatz:
Thanks, @EreMaijala. I'll test the collections functionality when time permits.

In the meantime, it's worth mentioning (since you told me on Slack but didn't write it here) that the removal of the tab from the breadcrumb trail is motivated by the fact that, with AJAX tab loading, including a tab-specific breadcrumb is less relevant than it used to be. We would either need to update the breadcrumb dynamically when the tabs change or (simpler and easier) remove the tab-specific breadcrumb entirely, as proposed here.

I'm interested in hearing whether anyone objects to this -- particularly @crhallberg. I do not personally have any strong preference one way or the other, so I'm happy to opt for the simpler approach if no one feels more strongly.

demiankatz:
@EreMaijala, I double-checked that the breadcrumb trail seems to work reasonably on collections. If I go directly to a collection, I get collection home in the breadcrumbs; if I do a search, I get the search. As you point out, if I perform a search at any point, the breadcrumb trail is going to include the last search instead of the collection home... but I don't think that's a problem we need to solve in this PR; that might just be a matter of clearing the last search memory in more contexts.

EreMaijala:
@demiankatz Thanks for testing!

crhallberg:
I have no opposition to this, despite what my tardiness implies!

PR #1823, opened 2021-01-14
Shibboleth support for library cards

xrosecky:
Shibboleth support for library cards.

TODO: 
- [x] Add a wiki page describing the functionality, including an example configuration


demiankatz:
Thanks, @xrosecky! I will try to give this a review when time permits, though at the moment I am still catching up on some older pull requests due to the holiday backlog. In the meantime, I have also flagged @EreMaijala in case he has any comments, since this touches on a couple of areas that are relevant to his work.

themes/bootstrap3/templates/librarycards/editcard.phtml
```diff
@@ -2,6 +2,8 @@
   // Set up page title:
   $pageTitle = empty($this->card->id) ? 'Add a Library Card' : 'Edit Library Card';
   $this->headTitle($this->translate($pageTitle));
+  // If the card is managed by Shibboleth, enable only renaming the card
+  $rename = ($this->loginMethod == 'shibboleth' || $this->shibboleth);
```
EreMaijala: As far as I can see, `$this->shibboleth` check would break normal library card handling if the user logs in using Shibboleth. You should still be able to handle normal library cards regardless of the login method.

EreMaijala: I'd also rename $rename to $renameOnly.

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
```diff
@@ -266,9 +300,14 @@ protected function processEditLibraryCard($user)
             $username = "$target.$username";
         }
 
+        $card = $user->getLibraryCard($id == 'NEW' ? null : $id);
```
EreMaijala: Should probably include a check that $id is not 'NEW' if $rename is true. This should be the case, but since the parameters come from the user, they should be validated.

themes/bootstrap3/templates/librarycards/home.phtml
```diff
@@ -52,7 +52,11 @@
   <?php endif; ?>
 
   <div class="btn-group">
-    <a href="<?=$this->url('editLibraryCard') ?>NEW" class="btn btn-link"><i class="fa fa-edit" aria-hidden="true"></i> <?=$this->transEsc('Add a Library Card')?></a>
+    <?php if($this->shibboleth): ?>
```
EreMaijala: This prevents adding of a normal library card if the user has logged in using Shibboleth.

EreMaijala (changes requested):
@xrosecky Could you provide a bit of background information on the use case? Normally, it's thought that either the library card is connected during Shibboleth login, or the user can add their own card later using the card number and pin code.

How does this functionality interact with MyResearch::catalogloginAction that's called when a library card is needed?

I've added some review comments, but they might be a bit off since I'm not sure about why this is needed.

themes/bootstrap3/templates/librarycards/editcard.phtml
xrosecky: It is too specific for our use case so I decided to enable editation of library cards when shibboleth is used for authentication.

xrosecky:
@EreMaijala Can you look at it again? I simplified it and enabled editation of library cards when shibboleth is used.

EreMaijala:
@xrosecky Thanks for the work on simplifying the implementation. I'm still a bit confused as to why this is needed, though. Could you describe the use case in a bit more detail? Normally the library card is connected right at login, and without some further information I'm failing to see when this is needed.

xrosecky:
We need to support connecting multiple library cards to user account using Shibboleth authentication from different IdPs instead of relying on entering card number and pin. In our use case we have separate IdP for each connected library.

EreMaijala:
@xrosecky Ok, now I understand. I'll add some comments when reviewing the code. Have you verified that an existing Shibboleth session from the primary login always gets overridden or is there further configuration required in Shibboleth?

module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -723,6 +723,26 @@ public function getILSLoginMethod($target = '')
...
+     *
+     * @param \Laminas\Http\PhpEnvironment\Request $request Request object
+     * containing account credentials.
+     * @param \VuFind\Db\Row\User                  $user    Connect newly created
+     * library card to this user.
+     *
+     * @return void
+     * @throws \Exception
+     */
+    public function connectUser($request, $user)
```
EreMaijala: Maybe connectUserCard?

module/VuFind/src/VuFind/Auth/Shibboleth.php
```diff
@@ -331,6 +331,31 @@ public function logout($url)
...
+     * @param \VuFind\Db\Row\User                  $connectingUser Connect newly
+     * created library card to this user.
+     *
+     * @return void
+     */
+    public function connectUser($request, $connectingUser)
+    {
+        $entityId = $this->getCurrentEntityId($request);
+        $shib = $this->getConfigurationLoader()->getConfiguration($entityId);
+        $username = $shib['prefix'] . '.' . $this->getAttribute(
```
EreMaijala: Should probably check that getAttribute returns a valid value here before adding a library card. If not, indicate the error with a flash message or something.

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
```diff
@@ -240,6 +242,35 @@ public function selectCardAction()
...
 
+    /**
+     * Redirects to Shibboleth authentication to connect a new library card
+     *
+     * @return \Laminas\Http\Response
+     */
+    public function connectNewShibbolethCardAction()
+    {
+        $url = $this->getServerUrl('librarycards-connectshibbolethcard');
+        $redirectUrl = $this->getAuthManager()->getSessionInitiator($url);
```
EreMaijala, line 256: Check that getSessionInitiator doesn't return false

themes/bootstrap3/templates/librarycards/home.phtml
```diff
@@ -52,6 +52,9 @@
   <?php endif; ?>
 
   <div class="btn-group">
+    <?php if($this->shibboleth): ?>
```
EreMaijala: I think this needs to be configurable and default to being disabled. In most cases the option would probably just confuse users. While adding configuration I'd suggest adding a comment on why and how the feature is used, and if any there are any prerequisites e.g. in Shibboleth configuration.

EreMaijala (changes requested):

xrosecky:
Yes, existing Shibboleth session from the primary login always gets overridden.

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
```diff
@@ -78,10 +78,15 @@ public function homeAction()
         // Connect to the ILS for login drivers:
         $catalog = $this->getILS();
 
+        $config = $this->getConfig();
+        $shibboleth = isset($config->Catalog->shibboleth_library_cards) &&
```
EreMaijala: You can combine the two checks into a single `!empty()` check.

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
EreMaijala: While checking for null with the `==` operator works, I'd rather just check `if (!$redirectUrl)` since getSessionInitiator should only return false or a string and the check would cater for both false and an empty string.

module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -723,6 +723,26 @@ public function getILSLoginMethod($target = '')
...
+     * @param \VuFind\Db\Row\User                  $user    Connect newly created
+     * library card to this user.
+     *
+     * @return void
+     * @throws \Exception
+     */
+    public function connectUserCard($request, $user)
+    {
+        $auth = $this->getAuth();
+        if (!is_callable([$auth, 'connectUser'])) {
```
EreMaijala: This should also be connectUserCard.

module/VuFind/src/VuFind/Auth/Shibboleth.php
```diff
@@ -331,6 +331,35 @@ public function logout($url)
...
+     * Connect user authenticated by shibboleth to library card.
+     *
+     * @param \Laminas\Http\PhpEnvironment\Request $request        Request object
+     * containing account credentials.
+     * @param \VuFind\Db\Row\User                  $connectingUser Connect newly
+     * created library card to this user.
+     *
+     * @return void
+     */
+    public function connectUser($request, $connectingUser)
```
EreMaijala: This should also be connectUserCard.

module/VuFind/src/VuFind/Auth/Shibboleth.php
```diff
@@ -331,6 +331,35 @@ public function logout($url)
...
+    {
+        $entityId = $this->getCurrentEntityId($request);
+        $shib = $this->getConfigurationLoader()->getConfiguration($entityId);
+        $username = $this->getAttribute($request, $shib['cat_username']);
+        if ($username == null) {
```
EreMaijala: I'd prefer `if (!$username)` since it covers all "falsy" cases.

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
```diff
@@ -240,6 +245,42 @@ public function selectCardAction()
...
+     * @return \Laminas\Http\Response
+     */
+    public function connectNewShibbolethCardAction()
+    {
+        $url = $this->getServerUrl('librarycards-connectshibbolethcard');
+        $redirectUrl = $this->getAuthManager()->getSessionInitiator($url);
+        if ($redirectUrl == null) {
+            $this->flashMessenger()
+                ->addMessage('authentication_error_technical', 'error');
+            $redirectUrl = '/LibraryCards/Home';
```
EreMaijala: It would be better to return a redirect with `->toRoute()` to avoid hard-coding paths.

module/VuFind/config/module.config.php
```diff
@@ -653,6 +653,7 @@
     'Install/PerformSecurityFix', 'Install/ShowSQL',
     'LibGuides/Home', 'LibGuides/Results',
     'LibraryCards/Home', 'LibraryCards/SelectCard',
+    'LibraryCards/ConnectShibbolethCard', 'LibraryCards/ConnectNewShibbolethCard',
```
EreMaijala: The difference between these actions is not obvious from the names. How about naming ConnectNewShibbolethCard to something like ConnectShibbolethCardLogin? Alternatively use a single action and distinquish between the cases using a query parameter, like `/LibraryCards/ConnectShibbolethCard?redirectToLogin=1`.

EreMaijala (changes requested):

xrosecky:
@EreMaijala Could you look at it again?

languages/en.ini
```diff
@@ -14,6 +14,7 @@ account_normal_checkouts = "Items due later"
 account_requests_available = "Available for Pickup"
 account_requests_in_transit = "In Transit"
 Add a Library Card = "Add a Library Card"
+Add a Library Card using Shibboleth login = "Add a Library Card using Shibboleth login"
```
EreMaijala: This just occurred to me: I think it might make sense to change the default wording to align with the actual login option "Institutional Login". I believe that the term "Shibboleth" is something that a user would not typically be aware of.

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
```diff
@@ -78,10 +78,14 @@ public function homeAction()
...
 
+        $config = $this->getConfig();
+        $shibboleth = !empty($config->Catalog->shibboleth_library_cards) &&
+            ($this->getAuthManager()->getAuthMethod() == 'Shibboleth');
         return $this->createViewModel(
             [
                 'libraryCards' => $user->getLibraryCards(),
-                'multipleTargets' => $catalog->checkCapability('getLoginDrivers')
+                'multipleTargets' => $catalog->checkCapability('getLoginDrivers'),
+                'shibboleth' => $shibboleth,
```
EreMaijala: Name the view variable something more descriptive like `allowShibbolethCards`.

EreMaijala (changes requested):

config/vufind/config.ini
```diff
@@ -354,6 +354,10 @@ title_level_holds_mode = "disabled"
 ; Whether support for multiple library cards is enabled. Default is false.
 ;library_cards = true
 
+; Whether support for connecting multiple library cards using Shibboleth is
+; enabled. You have also to enable Shibboleth authentication. Default is false.
```
EreMaijala: I added a TODO point to the task. I'd suggest creating the wiki page and adding a link to it here. Something like "See https://vufind.org/wiki/configuration:shibboleth_library_cards for more information".

demiankatz: Good idea. Please let me know if you need me to create that page (or increase anyone's wiki rights so that they can do it). (Page creation is a bit restricted right now due to spammers abusing it).

demiankatz:
Thanks for the progress, @xrosecky! Please let me know if you need me to set up a wiki page as discussed above.

xmorave2:
@demiankatz, @xrosecky nor me do not have permissions to create a wiki page, could you please create it? Thanks

demiankatz:
@xmorave2, I have created a placeholder page [here](https://vufind.org/wiki/configuration:shibboleth:library_cards). Please let me know if you have any problems editing it, or if you need anything else. Thanks!

xmorave2:
I wrote a draft of documentation on that page. Any comments are appreciated @demiankatz, @xrosecky. Thanks!

demiankatz:
Thanks, @xmorave2 -- I made some cosmetic changes and adjusted some links so that everything goes somewhere. Feel free to review/revise my work. I decided it was more useful to link directly to config files on GitHub than to create mostly useless wiki pages about those files, but I'm happy to revisit that decision if anyone feels differently. In any case, I'll wait for further feedback before proceeding, but please let me know when everyone is ready for this to be merged!

xmorave2:
Thanks @demiankatz! @xrosecky has reviewed it too. So we are definitely ready for this to be merged.

demiankatz:
Thanks, @xmorave2 -- I think this is nearly ready to go, though I just have one last question: there is a lot of Shibboleth-specific naming still in the code (in the name of the language string, routes, etc.), but the logic actually appears to be fairly generic, as long as the Auth code implements the appropriate method. Should we rename more things with generic names so this does not become confusing if we add support for this feature to a non-Shibboleth auth method? Specifically, I would suggest:

  - Perhaps consider renaming the shibboleth_library_cards setting to auth_based_library_cards (we can still note in the comment that this is currently only supported by Shibboleth, and link to the Shibboleth-specific wiki page).
  - In language files and templates, change not just the value but also the key from "Add a Library Card using Shibboleth login" to "Add a Library Card using institutional login."
  - Remove the word "Shibboleth" from the ConnectShibbolethCard and ConnectShibbolethCardLogin route names; I think this is still appropriate but more concise and generic. (And of course the controller methods should be renamed to match).
  - Rename the allowShibbolethCards variable to something like allowConnectingCards.

I think this makes the code more flexible and future-proof. What do you think?

xrosecky:
Yes, I agree, it is good idea.

xrosecky:
Done - see the latest commit.

config/vufind/config.ini
```diff
@@ -354,6 +354,11 @@ title_level_holds_mode = "disabled"
 ; Whether support for multiple library cards is enabled. Default is false.
 ;library_cards = true
 
+; Whether support for connecting multiple library cards via authentication is
+; enabled. This is currently only supported for Shibboleth. Default is false.
+; See https://vufind.org/wiki/configuration:shibboleth:library_cards for details.
+;auth_based_library_cards = true
```
demiankatz, line 360: I updated the comment and setting name in this file to match the code; let me know if you have any suggested changes.

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
```diff
@@ -78,10 +78,17 @@ public function homeAction()
         // Connect to the ILS for login drivers:
         $catalog = $this->getILS();
 
+        $config = $this->getConfig();
+        $allowConnectingCards = !empty(
```
demiankatz: The formatting of this is a bit strange; I'd suggest collapsing this back to a single line, and moving the `&&` to the beginning of the second line to meet the line length restriction.

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
```diff
@@ -78,10 +78,17 @@ public function homeAction()
...
+            $config->Catalog
+                ->auth_based_library_cards
+        ) &&
+            ($this->getAuthManager()->getAuthMethod() == 'Shibboleth');
```
demiankatz: I don't think we want this hard-coded Shibboleth restriction anymore. Should we check for the callability of the connectUserCard method here, or just trust that the user configured things correctly and let the system fail elsewhere if there's a problem?

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
```diff
@@ -240,6 +247,45 @@ public function selectCardAction()
...
+        if (!($user = $this->getUser())) {
+            return $this->forceLogin();
+        }
+        try {
+            $this->getAuthManager()->connectUserCard($this->getRequest(), $user);
+        } catch (\Exception $ex) {
+            $this->flashMessenger()->setNamespace('error')
+                ->addMessage($ex->getMessage());
+        }
+        return $this->redirect()->toUrl('/LibraryCards/Home');
```
demiankatz: Wouldn't it be better to redirect `->toRoute('librarycards-home')` here?

demiankatz (changes requested):
Thanks for the updates, @xrosecky -- just a few last-minute questions and comments. Once these are addressed, I think we can merge this!

xrosecky:
OK, I fixed them and I also renamed connectUserCard to connectLibraryCard for better naming convention.

module/VuFind/src/VuFind/Auth/AbstractBase.php
```diff
@@ -304,6 +304,19 @@ public function supportsPasswordRecovery()
...
+     * Does this authentication method support connecting library card of
+     * currently authenticated user?
+     *
+     * @return bool
+     */
+    public function supportsConnectingLibraryCard()
+    {
+        // By default, connecting library card of currently authenticated user
+        // is not supported.
+        return false;
```
demiankatz: How about changing this to `return method_exists($this, 'connectLibraryCard');`? That way, by adding the required method, the supports method will automatically return true, we can remove the override method from the Shibboleth class, and there's less work to implement support for other methods in the future. It's still possible to extend the method if additional platform-specific logic is needed, but it makes the default implementation more flexible.

module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -723,6 +737,26 @@ public function getILSLoginMethod($target = '')
         return false;
     }
 
+    /**
+     * Connect user authenticated by shibboleth as library card to his account.
```
demiankatz: I'd change `user authenticated by shibboleth` to `authenticated user` to make this more generic.

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
```diff
@@ -78,10 +78,13 @@ public function homeAction()
         // Connect to the ILS for login drivers:
         $catalog = $this->getILS();
 
+        $config = $this->getConfig();
```
demiankatz: Is $config no longer being used? Can we remove this line?

demiankatz (changes requested):
Thanks -- we are very nearly done, but I spotted a couple more minor issues and one potential simplification. I think after this we can definitely merge the PR!

xrosecky:
OK, fixed in the latest commit.

demiankatz:
Thanks, @xrosecky -- merging now!

PR #1824, opened 2021-01-15
FeedbackForms: fix required/requireOne checkbox validator.

samuli:
Currently validating required checkbox options fails if the options are configured with label and value, for example:
```
 options:
   - label: Option 1
     value: option-1
   - label: Option 2
     value: option-2
```
 $el['options'] is an key-value list with `value` as the key. This PR fixes the validation to use keys.


demiankatz:
Thanks for the fix and the extensive test coverage!

PR #1825, opened 2021-01-15
Fix JDK8 compatibility in WorkKeys generation.

demiankatz:
I was just testing VuFind's compatibility with JDK8 and discovered that we introduced a compatibility break recently through use of the isBlank() method in the work key generator. I believe that this PR switches to a more backward-compatible method, but I haven't had time to test extensively yet.

TODO
- [x] Test


demiankatz:
@EreMaijala, any thoughts on this one? I'll take the time to do some testing later today if you agree with the general approach.

EreMaijala (approved):
Oops. Thanks for finding that one. I can't test this, but the change looks solid.

demiankatz:
Thanks, @EreMaijala, I've done some fairly extensive testing and it seems to be working as expected, so I'm merging this now!

PR #1826, opened 2021-01-15
Fix undocumented behavior in MarcAdvancedTrait::getConsortialIDs

Witiko:
`MarcAdvancedTrait::getConsortialIDs` [calls `$this->getFieldArray('035', 'a', true);`](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php#L934), which has undocumented behavior, because `MarcReaderTrait::getFieldArray` [expects the second argument to be an array](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/RecordDriver/MarcReaderTrait.php#L104) and only [does the right thing by accident](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/RecordDriver/MarcReaderTrait.php#L115).

### DIFF
--- a/module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
+++ b/module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
@@ -901,6 +901,6 @@ trait MarcAdvancedTrait
      */
     public function getConsortialIDs()
     {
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks for the fix!

PR #1828, opened 2021-01-15
[VUFIND-1457] Improve PHP8 compatibility

demiankatz:
TODO

- [x] Review use of `is_callable` due to changed behavior (static call to non-static method now returns false)
- [x] Finish testing in PHP 8
- [x] Run full test suite in PHP 7


demiankatz:
In the interest of getting core code fixes merged as quickly as possible, I am merging this today. We still need https://github.com/laminas/laminas-db/pull/177 to be resolved before the entire VuFind package will be fully PHP 8 compatible. I will leave [VUFIND-1457](https://vufind.org/jira/browse/VUFIND-1457) open to track that detail until it can be resolved.

PR #1829, opened 2021-01-19
Fix breadcrumbs when there is no last search in memory.

EreMaijala:
The old code caused a stray li element to be created when there was no previous search in memory.

### DIFF
--- a/themes/bootstrap3/templates/cart/cart.phtml
+++ b/themes/bootstrap3/templates/cart/cart.phtml
@@ -3,7 +3,7 @@
   $this->headTitle($this->translate('Book Bag'));
 
   // Set up breadcrumbs:
-  $this->layout()->breadcrumbs = '<li>' . $this->searchMemory()->getLastSearchLink($this->transEsc('Search'), '', '</li> ')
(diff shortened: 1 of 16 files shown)


demiankatz:
@EreMaijala, this pattern (concatenating `'<li>'` with a `getLastSearchLink()` call with a `$prefix` parameter of `''`) appears to be repeated in a large number of templates. At a glance, it looks like we probably want to make this fix in all of those places, though I haven't taken the time to look more closely yet. What do you think? Any thoughts from @crhallberg?

EreMaijala:
@demiankatz Oh, I didn't realize that. It looks to me like it has been a simple mistake since the beginning and just pasted everywhere. I'll fix them all.

EreMaijala:
@demiankatz It might have its root in advanced search's layout.phtml.

EreMaijala:
@demiankatz Should be fixed now.

demiankatz:
Thanks, @EreMaijala. This problem really does seem to have existed since the beginning of time; thanks for finally catching it! I've run the full test suite to confirm no obvious negative side effects. I'll merge once I get @crhallberg's go-ahead (just in case he remembers anything I'm forgetting that might be relevant... though that seems extremely unlikely).

crhallberg:
This looks great! Thank you!

PR #1830, opened 2021-01-20
fix bug (bad variable name) in new location handling

ihardy:


demiankatz:
Whoops! Sorry I missed that. :-) A sign that we probably should take some time to increase test coverage when time permits.

PR #1834, opened 2021-01-21
Donor bookplates

bbusenius:
In our VuFind instance, we display donor bookplates from the `927` and `928` fields. These appear as thumbnail images in the right sidebar with the donor name displayed below, like so: 

image

I have implemented this as a proof of concept based on our data as it currently is in order to start a conversation about it but I think there might be things you want to change. Note, I have not included anything in our `.properties` file or whatever else might be needed to get this in Solr. I thought it would be best to get your input before I took those steps (it's also a little bit out of my wheelhouse). Our related Solr looks like this:

image

A couple of things I imagine you might want to change (if it's possible, I'll defer to others who know more than me):
1. Instead of constructing the url for the image from the config, like I do here, would it be better to do this as part of the indexing process so that the whole url is stored in Solr?
2. Instead of having `donor` and `donor_code` as separate fields, can they be combined into 1 data structure in Solr?

Related to number two, you can see that we store these as arrays in Solr. We _do_ need these to be arrays because we actually have records with multiple donor bookplates. I can see that it might be nice to at least have the codes and donor names combined into a single array but I don't know much about that part of the system so I'll defer to others for a reasonable best practice here.

Below is how it currently displays in my proof of concept. Obviously it needs some work done to the display but I thought I'd get your thoughts before I got too detailed with that:

image


demiankatz:
Thanks for sharing, @bbusenius! I do think this is a feature that will be of interest to others, but the challenge is figuring out how best to make it generic and reusable, since I suspect that the details of the use case could vary dramatically from institution to institution.

My first suggestion would be to refactor this as a [Related Records](https://vufind.org/wiki/development:plugins:related_records_modules) module, rather than coding it into the view template. This will make it easier to disable when unused, as well as to control the display order through configuration (if, for example, somebody wanted to put it below the similar records instead of above them).

The other advantage of the "related records" approach is that the Related class could contain helper functions to support the display, and then it would be easy to customize in a subclass. For example, we could build the default example with a getBookplateUrl() method that looks in Solr for a donor_bookplate_url_str_mv field containing a full URL, but you could locally subclass that to construct a URL using a donor code. It would also make it reasonably simple to support a scenario where there is a single bookplate background image, and the template dynamically places text on top of it.

I don't think there is a standard place for this in MARC, so it may be best to put a commented-out fictional example in marc_local.properties but not to have any default indexing rules for this case. I think this is a somewhat niche situation, so while it's good to have it set up in a way that is easy to turn on, it's also good to design it so it doesn't impose any additional processing load on the majority of users who won't actually need it. Thus, a disabled-by-default Related module and some commented-out example import rules fit the bill perfectly.

Regarding how to structure the data in Solr, the options are pretty limited, because Solr is not designed to store complex nested data structures. I think there are really two options: parallel arrays (i.e. have donor_str_mv and donor_bookplate_url_str_mv fields, and make sure they have the same number of entries indexed in the same order) or delimited values (a single field that has to be post-processed to pick the values apart). I tend to refer the parallel arrays option, because it's easier to index that data without writing custom Java code... but each approach has pros and cons. Another advantage to my proposed approach is that we don't really prescribe anything, we just offer an example, and then the most appropriate choice can be made by the end user.

What do you think? Please let me know if there's more you'd like to discuss, or if you need more background on anything I've mentioned!

bbusenius:
Thanks @demiankatz! That's fantastic feedback and it makes good sense to me. I'll get to work on restructuring this and let you know if I have any questions. I appreciate your time and thoughtful feedback.

demiankatz:
Excellent, glad to help, @bbusenius!

bbusenius:
@demiankatz I did an initial refactoring of the code into a Related Records module. It's still not ready for prime time since I'll still need to setup an example in `marc_local.properties`, try it and finish refactoring. On that note, would it make sense to use `bookplate_str_mv` and `bookplate_url_str_mv` for field names? I'm thinking that the "bookplate" taxonomy might be nice since it's more general and I don't know if people will use this for _donor_ bookplates.

demiankatz:
@bbusenius, I agree that using more generic language makes sense -- it increases the flexibility of the code, and it's easy to customize to something more specific if there is a need.

I haven't reviewed the new code yet, since it's probably best for me to wait until you've finished the remaining steps... but please let me know when you are ready, and I'll take a look as soon as you give the word! :-)

todolson:
Coming in late:

There are two things that we are doing at UChicago with `donor` and `donor_code`:
1. enable donor recognition through display of electronic bookplates and donor description, and
2. enable Library Development and donors themselves to easily search for materials that were provided by a specific donor.

The search aspect is missing from the above discussion. My understanding that searching by both code and description is important to us locally as part of our donor relations. If Library Development wants a point a representative of The Martha and Stanton Friedberg Library Fund to all of the things their fund has paid for, they can just search `donor_code:stan` and see everthing. For that outward-facing reason, whatever new fields are added to the stock VuFind Solr schema, we will locally continue to need `donor` and `donor_code`, though they may just become copies of a stock field.

Might this search aspect also be of interest in the broader VuFind community?

Moving on to the display of book plates:

@demiankatz is quite correct that there is no standard place in MARC for these data.

I also agree that the parallel arrays option seems preferable. The edge cases that can mess up parallel arrays simply don't come up in this context.

In terms of `bookplate_url_str_mv`, I find myself resisting requiring full URLs in the Solr data. I always imagined this moving more towards being the consumer of a bookplate service. The bookplates would be able to be retrieved by some sort of service that library has implemented locally, and something in the config would provide a URL template that would tell VuFind how to use a code to format a request to display the bookplate. (A little bit like a book cover service.) For our current way of providing bookplates, with URL conventions, the configuration could look like (and I use`{code}` as a stand-in for whatever templating syntax):

```
bookplate_full: https://www.lib.uchicago.edu/bookplates/{code}-full.jpg
bookplate_thumb: https://www.lib.uchicago.edu/bookplates/{code}-thumb.jpg
```
(We'd have to work out the uppercase/lowercase issues, but I that is just Apache configuration on the other end.)

A site that provisions bookplates through something that looks more like a lookup could do something like:

```
bookplate_full: https://www.lib.uchicago.edu/bookplates?code={code)+size=full
bookplate_thumb: https://www.lib.uchicago.edu/bookplates?code={code)+size=thumb
```

And now that I working this through, I don't think URLs in a single `bookplate_url_str_mv` field could be used to provide both a thumbnail and full image. If the full URLs were added as Solr fields I think there would need to be two fields, one for the thumbnail image and one for the full image.

demiankatz:
@todolson, thanks for the feedback. I think all of this is more argument for taking as non-prescriptive an approach as possible here. If we define two record driver methods to provide bookplate-related information, and a related record module for rendering the bookplates, then users have two choices: they can populate the default fields used by the record driver methods (which I think should be viewed as examples rather than recommendations), or they can override code to meet local needs. By providing clearly defined methods and a display template, this makes those local customizations very specific and focused, and thus easier to maintain... but it remains extremely flexible.

Perhaps there is even an argument to have just one record driver method, getBookplateDetails(), which returns an array of associative arrays. This would make it even easier to add more features and customizations while keeping everything in one simple place.

bbusenius:
Thanks for the great feedback guys. I've refactored this to use url string templates and a `getBookplateDetails` method, though, when I look at it now, I'm seeing I did that in the `RelatedInterface` and not the record driver. Nevertheless, I think I could benefit from some more feedback at this point. @demiankatz, if you want me to move the `getBookplateDetails`, I don't think it should be any trouble. I wasn't originally planning on using thumbnails for us, however, now that you mention it @todolson, I can see that it will be more generally useful to others with thumbnails. So, I made that change. I also took a couple of other liberties. They are the following:

1. After discussing with someone at our institution, I made the display of bookplate titles under the bookplate image, configurable. People can now choose whether or not to display them. This because I was told people were thinking about getting rid of them but hadn't decided for certain. It made me realize that even if they don't get decide to get rid of them, more flexibility would be better. Note, the titles are still required in the Solr because we need them for alt text. It's just the display that's configurable.

2. I made the Solr field names configurable. So, instead of choosing a Solr field to put these in, it's just a configuration option. I realize this adds a little more complexity but I thought it was worth it. Not only is it more flexible but the comments in the config file can be used to explain the feature. That's my hope anyhow.

config/vufind/config.ini
```diff
@@ -1763,10 +1763,27 @@ hide_holdings[] = "World Wide Web"
...
 ;    Channels - Display links to channels of content related to record
+;    Bookplate - Display a bookplate or something similar
 ;    Similar - Similarity based on Solr lookup
 ;    WorldCatSimilar - Similarity based on WorldCat lookup
 related[] = "Similar"
 
+; URLs for bookplates if enabled in the related config above. %s represents an
+; image name or donor code that can be used to identify an image. This will
+; come from a Solr field (donor_code by default).
+; bookplate_full = https://your-institution.edu/bookplates/%s-full.jpg
```
demiankatz: Minor style suggestion: VuFind's convention in configuration files is that we put a space after the semi-colon for textual comments, but we do not put a space after the semi-colon for commented-out example settings. This makes it a little easier to visually distinguish between the two. To follow that convention, I'd recommend removing a few spaces from this file.

module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
```diff
@@ -1722,4 +1722,80 @@ public function getCoordinateLabels()
...
     }
+
+    /**
+     * Get the bookplate names.
+     *
+     * @param $s string name of data to retrieve.
+     *
+     * @return array
+     */
+    public function getBookplateSolrData($s)
```
demiankatz: I would actually move most, and maybe all, of this functionality out of the record driver and into the related record module. Most of it is only going to be used in that context, so it doesn't make sense to carry it around with every record all the time.

You can always call the getRawData() method if you want to access arbitrary fields from external code, should you wish to move getBookplateSolrData() somewhere else. As noted elsewhere, if nothing else, it doesn't make sense to have a Solr-specific method in the DefaultRecord base class, which is used by all records, not just Solr records.

config/vufind/config.ini
```diff
@@ -1763,10 +1763,27 @@ hide_holdings[] = "World Wide Web"
...
+; bookplate_thumb = https://your-institution.edu/bookplates/%s-thumb.jpg
+
+; Solr field with an array of image titles, parallel with bookplate_img_names_field.
```
demiankatz: It may make sense to drop the word Solr from these comments, and build this in a slightly more generic way. While I think this is most likely to be used in combination with Solr, it's conceivable that somebody might want to leverage the component to display data based on a web-scale discovery service or other data source. If you make things just slightly more generic, it should be possible to support that. Most services offer an array of fields, so if you fetch the field data using getRawData and then use the indexes configured here, the solution may be generalizable to at least some other search backends. Further flexibility might be achievable by subclassing, but we can cross that bridge when we come to it.

module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
```diff
@@ -1722,4 +1722,80 @@ public function getCoordinateLabels()
...
+    }
+
+    /**
+     * Get the full bookplate URL string template.
+     *
+     * @return string
+     */
+    public function getBookplateFullUrlTemplate()
+    {
+        return isset($this->recordConfig->Record->bookplate_full) ?
```
demiankatz: Rather than hard-coding the configuration path to these settings, I think it would be better to leverage the configuration options of the "Related" module. See my comment on the constructor for more details.

module/VuFind/src/VuFind/Related/Bookplate.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Display bookplate titles?
+     */
+    protected $displayTitles;
+
+    /**
+     * Establishes base settings for bookplates.
+     *
+     * @param string                            $settings Settings from config.ini
```
demiankatz, line 94: As alluded in an earlier comment, I think it would be useful to leverage $settings for some or all of the module's configuration. Any values attached to the module name in config.ini will be passed here as the $settings parameter, so, for example, if I configured "related[] = Bookplate:foo:bar" then $settings would be set to "foo:bar" here.

The advantage to this is that you can configure multiple instances of Bookplate with different behaviors, and you can turn this on in multiple backends, and each instance can have its own configuration. If you hard-code the configuration paths, you reduce flexibility.

There are two options for how to take advantage of this:

1.) Put all the settings directly into the Related declaration, in a delimited form. Given the large number of config options you are proposing, I don't recommend this, as it would be unwieldy and unreadable.

2.) Pass in the names of a configuration file and section -- i.e. "Related:config:Record" -- and pull settings from there. You just need to create a factory that injects the `VuFind\Config\PluginManager` service to the module's constructor so that you have the ability to load configuration files as needed... then the init() method can pull the appropriate configuration and set everything up. (And, of course, if $settings is empty, you can default to config and Record to keep the existing default behavior).

I clearly have a preference for option 2!

module/VuFind/src/VuFind/Related/Bookplate.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Get bookplate details for display.
+     *
+     * @return array
+     */
+    public function getBookplateDetails()
+    {
+        $sameLen = count($this->bookplateStrs) == count($this->bookplateImgNames)
+            ?? false;
```
demiankatz: I don't think `?? false` does anything here. The `count()` function is never going to return `null`, so the null coalescing operator will never be triggered. Do you instead want to have a `?? []` inside the `count()` calls in case the properties are unset?

themes/bootstrap3/templates/Related/Bookplate.phtml
```diff
@@ -0,0 +1,17 @@
+<?php $bookplates = $this->related->getBookplateDetails(); ?>
+<div id="bookplates">
+  <?php foreach($bookplates as $bookplate): ?>
+    <div class="bookplate">
+      <div class="bookplate-img">
+        <a href="<?=$bookplate['fullUrl']?>">
```
demiankatz: This template needs escaping added everywhere -- i.e. `$this->escapeHtmlAttr()` for values inside HTML attributes, and `$this->escapeHtml()` for the title displayed inline. Otherwise, there's a risk that special characters (`&`, `<`, etc.) might garble the HTML output.

demiankatz: Also, should we account for the possibility of only one URL instead of two? Should it be possible to only specify a full URL, or to have a thumbnail without a full URL? If only one image size is available, how should the code behave?

config/vufind/config.ini
```diff
@@ -1763,10 +1763,27 @@ hide_holdings[] = "World Wide Web"
...
+;    Bookplate - Display a bookplate or something similar
 ;    Similar - Similarity based on Solr lookup
 ;    WorldCatSimilar - Similarity based on WorldCat lookup
 related[] = "Similar"
 
+; URLs for bookplates if enabled in the related config above. %s represents an
+; image name or donor code that can be used to identify an image. This will
+; come from a Solr field (donor_code by default).
+; bookplate_full = https://your-institution.edu/bookplates/%s-full.jpg
+; bookplate_thumb = https://your-institution.edu/bookplates/%s-thumb.jpg
```
demiankatz: Would it be more flexible to offer a set of predefined tokens, rather than a simple `%s` that always expects a specific value? For example, what if the user actually has full URLs for thumbnails and full-size images in their index? They can't express that here by just setting both bookplate_full and bookplate_thumb to %s. If we instead offered a few predefined tokens like %%title%% and %%img_name%% then there would be greater flexibility. I'm not proposing that we should make this super-complicated right now; we could start by just offering the simplest, bare minimum options... but setting up such a mechanism might make this a bit more extensible and flexible for the future.

demiankatz (changes requested):
Thanks for the progress, @bbusenius, this is taking shape nicely. I like the direction you're heading in, and actually most of my comments below suggest going even a little further toward moving things into the Related module and increasing configurability. I hope this is all helpful, and as always, please let me know if you have questions or concerns... and please consider everything a suggestion rather than a demand. :-)

module/VuFind/src/VuFind/Related/Bookplate.php
bbusenius: @demiankatz I don't think I quite know how to create a factory that injects the `VuFind\Config\PluginManager` service to the module's constructor and access the settings. I created a `BookplateFactory` which is executing but I don't really know where to take it from there. Could you give me some more detail or point me to a similar example? Sorry for the hassle.

demiankatz: Sure, no problem! In the factory, you just need to pull the config manager out of the container and pass it to the object being built, e.g.:

<pre>
$configManager = $container->get(\VuFind\Config\PluginManager::class);
return new $requestedName($configManager);
</pre>

Then you can assign that to a property in Bookplate::__construct(), and go from there.

Does that help? I'm happy to dive deeper as needed.

bbusenius: Thanks @demiankatz! That seems to have gotten me over the hump. Much appreciated!

demiankatz: Fantastic, glad I could help! It's definitely a bit unintuitive at first, but once you get the hang of it, the Laminas factory system is pretty pleasant to work with. (At least, it gives me a certain weird "virtual Lego" satisfaction). :-)

config/vufind/config.ini
bbusenius: Fixed.

module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
bbusenius: This makes sense. I've taken a stab at implementation.

config/vufind/config.ini
bbusenius: I did this and I also removed the word Solr from function names.

module/VuFind/src/VuFind/Related/Bookplate.php
bbusenius: I think I have this setup correctly using option 2. I'm not sure if I got any of the details wrong but it's generally setup this way now.

module/VuFind/src/VuFind/Related/Bookplate.php
bbusenius: Silly mistake. Fixed.

themes/bootstrap3/templates/Related/Bookplate.phtml
bbusenius: Great suggestion! I fixed the escaping issue and set it up such that if only thumbnail or full URL is present, it will just put an image in the template. If both are present, it's a thumbnail and a link to the full image. I agree, this is more versatile. We might even use just one of them ourselves, since, that was what I was originally thinking.

config/vufind/config.ini
bbusenius: Fantastic suggestion! I did this in a _very_ simple way. I didn't really have to predefine tokens so much, rather, I divorced the thumbnail logic (previously image name) from the full image logic. This allows them to be _different_. I changed variable names and added a corresponding method to do this. I also setup the infrastructure to use "tokens" going forward. Now we have `%%img%%` and `%%thumb%%` both of which can be part of an image name (e.g. `donor_code` like we use) or the entire URL. If somebody wanted to do like we do, they could have a configuration similar to this:

```
bookplate_full = https://your-institution.edu/bookplates/%%img%%-full.jpg
```

But if someone indexes the full URL, they would simply have a configuration like this:

```
bookplate_full = %%img%%
```

demiankatz:
Thanks for the great progress, @bbusenius! I'm on vacation this week but will give this a closer look next week upon my return.

config/vufind/Related.ini
```diff
@@ -0,0 +1,24 @@
...
+; to the ones below.
+;bookplate_full = https://your-institution.edu/bookplates/%%img%%-full.jpg
+;bookplate_thumb = https://your-institution.edu/bookplates/%%thumb%%-thumb.jpg
+
+; Data field with an array of image titles, parallel with bookplate_img_names_field.
+; Is required for alt text, even if bookplate_display_title is set to false.
+;bookplate_titles_field = "donor"
+
+; Data field with an array of strings that represent the unique parts of image names,
+; of full URLs to images, parallel with bookplate_titles_field.
```
demiankatz: Should "of" be "or" here?

module/VuFind/src/VuFind/Related/Bookplate.php
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * Constructor
+     *
+     * @param VuFind\Config\PluginManager $configLoader PluginManager
+     */
+    public function __construct(\VuFind\Config\PluginManager $configLoader)
+    {
+        $this->configLoader = $configLoader;
+        $this->config = $this->configLoader->get('Related');
```
demiankatz: I think the part of my suggestions that may have gotten lost in transition was the idea that the settings passed to the related module via config.ini could be used to control where additional settings are loaded from. It looks like you've created a new Related.ini file which is hard-coded here. I would suggest that instead of loading the configuration in the constructor here, you instead do it in the `init()` method, so that you can load a different config file/section depending on the user's settings.

If it's still not clear what I'm looking for, I can take a stab at implementing it here so you can see -- just let me know if that would be helpful!

Also, it looks like the comment documenting the existence of this new module still needs to be added to config.ini around here: https://github.com/vufind-org/vufind/blob/dev/config/vufind/config.ini#L1764 -- not sure if that was here previously and got lost in refactoring or if I forgot to suggest it on the first pass. Either way, we should make sure it's there before we merge.

module/VuFind/src/VuFind/Related/Bookplate.php
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Get the full bookplate URL string template.
+     *
+     * @return string
+     */
+    protected function getBookplateFullUrlTemplate()
+    {
+        return isset($this->config->Bookplate->bookplate_full) ?
```
demiankatz: All of these isset checks can be simplified using the null coalescing operator. If you run `vendor/bin/phing php-cs-fixer` it should auto-fix them for you, assuming you have the latest Composer dependencies installed (php-cs-fixer recently improved its ability to auto-correct these issues).

demiankatz (commented):
Thanks for all the progress, @bbusenius -- this is really shaping up nicely. I ran out of time before I could do a thorough re-review of every line of code, but hopefully the comments I've just left will help move things along a little further. I'll take another, closer look once you've had time to respond to these.

config/vufind/Related.ini
bbusenius: Yes, thanks for catching that.

module/VuFind/src/VuFind/Related/Bookplate.php
bbusenius: Beautiful. This is much nicer!

module/VuFind/src/VuFind/Related/Bookplate.php
bbusenius: I see what you're asking for. I took a stab at implementing it. This is now more flexible. The only hitch I ran into was that I couldn't get anything after the second `:` to pass through. Switching it to a `;` worked fine though.

Yes, I had the existence of the module documented in the beginning but lost this in refactoring. I added this back and moved the default bookplate config back to the same section.

demiankatz: Interesting; since none of the existing Related modules use parameters (meaning this has never really been tested), I wonder if there's a bug in the code that's exploding on more `:` characters than necessary. When I have a bit more time, I'll investigate and fix it if necessary!

demiankatz: @bbusenius, as I suspected, I think the problem lies here:

https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/View/Helper/Root/Related.php#L114

That line should be changed to `$parts = explode(':', $current, 2);` and then the ':' separator should work for you.

Would you be willing to make that change here, switch the explode character back from ';' to ':' (for consistency with recommendation modules), and see if that sorts everything out? If it does, then we should adjust the comments in the .ini file to include the optional parameters.

If you don't have time, let me know and I can try to work on this when I carve out a little more of my own... :-)

bbusenius: Thanks @demiankatz! I appreciate your help with diagnosing this. I made these changes. Let me know if there's anything else.

config/vufind/config.ini
```diff
@@ -1763,10 +1763,41 @@ hide_holdings[] = "World Wide Web"
...
+;   %%thumb%% - thumbnail image name, partial thumbnail image name or full URL.
+; If you index the entire path to bookplate images, you will only need to use
+; the token, e.g. bookplate_full = %%img%%, otherwise you can build URLs similar
+; to the ones below.
+;bookplate_full = https://your-institution.edu/bookplates/%%img%%-full.jpg
+;bookplate_thumb = https://your-institution.edu/bookplates/%%thumb%%-thumb.jpg
+
+; Data field with an array of image titles, parallel with bookplate_img_names_field.
+; Is required for alt text, even if bookplate_display_title is set to false.
+;bookplate_titles_field = "donor"
```
demiankatz: Very minor nitpick, but maybe we should make the field examples end in _str_mv to show how they could be added without changing the Solr schema.

module/VuFind/src/VuFind/Related/Bookplate.php
```diff
@@ -0,0 +1,243 @@
+<?php
+/**
+ * Related Records: Bookplates
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2009.
```
demiankatz: Should probably update copyright date to 2021 to reflect the actual vintage of this file.

module/VuFind/src/VuFind/Related/Bookplate.php
```diff
@@ -0,0 +1,243 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Related_Records
+ * @author   Demian Katz <demian.katz@villanova.edu>
```
demiankatz: You can also give yourself author credit if you want. :-)

module/VuFind/src/VuFind/Related/Bookplate.php
```diff
@@ -0,0 +1,243 @@
...
+     *
+     * @param string                            $settings Settings from config.ini
+     * @param \VuFind\RecordDriver\AbstractBase $driver   Record driver object
+     *
+     * @return void
+     */
+    public function init($settings, $driver)
+    {
+        $config = explode(':', $settings);
+        $configFile = 'config';
```
demiankatz: I would simplify lines 102-107 into:

<pre>
$configFile = $config[0] ?? 'config';
$configSection = $config[1] ?? 'Record';
</pre>

In addition to being more concise, this also removes the artificial constraint that there must be exactly two parameters. (Perhaps somebody wants to change the file name but keep the default section name... and maybe we'll add more parameters in the future).

module/VuFind/src/VuFind/Related/Bookplate.php
```diff
@@ -0,0 +1,243 @@
...
+    /**
+     * Get an array of data representing bookplates.
+     *
+     * @param $s string name of data to retrieve.
+     *
+     * @return array
+     */
+    protected function getBookplateData($s)
+    {
+        $data = [
```
demiankatz: It seems overcomplicated to build this $data array and then pull exactly one element out of it every time this method is called. Why not just make this `getBookplateData($field)` and then pass in `$this->getBookPlateTitlesField()`, etc., directly from `init()` instead of using these abstract strings? I think that clarifies the code and improves efficiency (since that way we only call each method once, instead of three times).

module/VuFind/src/VuFind/Related/Bookplate.php
```diff
@@ -0,0 +1,243 @@
...
+
+    /**
+     * Get bookplate details for display.
+     *
+     * @return array
+     */
+    public function getBookplateDetails()
+    {
+        $sameLen = count($this->bookplateStrs) == count($this->bookplateImages);
+        $hasBookplates = !empty($this->bookplateStrs)
```
demiankatz: Technically, you don't need to check that both bookplateStrs and bookplateImages are non-empty if you're also confirming that they're the same length -- you can just check one and safely assume that they're both the same.

module/VuFind/src/VuFind/Related/BookplateFactory.php
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * Related record plugin factory
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2010.
```
demiankatz: Again, you might want to update copyright date here.

themes/bootstrap3/templates/Related/Bookplate.phtml
```diff
@@ -0,0 +1,26 @@
+<?php $bookplates = $this->related->getBookplateDetails(); ?>
+<div id="bookplates">
+  <?php foreach($bookplates as $bookplate): ?>
+    <?php $bpTitle = $bookplate['title'];?>
+    <?php $bpAltText = $bpTitle . ' bookplate';?>
```
demiankatz: We need a translation for this alt text -- I would suggest adding something to en.ini like:

<pre>
bookplate_label = "%%title%% bookplate"
</pre>

And then changing this to something like:

<pre>
$bpAltText = $this->translate('bookplate_label', ['%%title%%' => $bpTitle]);
</pre>

This also raises the question of whether $bpTitle should also be run through $this->translate in case somebody wants to set up a multi-lingual bookplate title.

demiankatz (changes requested):
Thanks, @bbusenius, this is really looking great! Since it's independent of the bookplate functionality, I've cherry-picked the related plugin manager bug fix, so if you merge dev back into this branch, there will be one less changed file in this PR.

Beyond that, I've just left a handful of new comments, mostly minor suggestions for improving documentation and simplifying code, as well as accounting for i18n.

After that, the only thing left to do is to test this. It would be great to have a unit test for at least the Related module itself. I'd be happy to put one together as part of my own testing of the class, but if you'd like to take a stab at it first, let me know.

Once these final adjustments are made, I'll do one last round of testing, and merge this if all is well.

Thanks again!

module/VuFind/src/VuFind/Related/Bookplate.php
bbusenius: Yes, this also no longer makes sense since we allow any combination of thumbnails, full-size images and titles. I had meant to change this and apparently forgot. Since the string is required for alt text, I think it makes sense to anchor to that. That's what I've done.

themes/bootstrap3/templates/Related/Bookplate.phtml
bbusenius: I _really_ like the idea of running `$bpTitle` through `translate`. It adds a much greater degree of flexibility. I never would have thought of this but even even for people who might not want to create a multi-lingual bookplate title, I can imagine an edge case scenario where they might want to control all the titles through a language file. For example, if someone only had donor codes, like we have, and nothing else. It might not be an ideal practice but they could actually do everything with just 1 field if I'm not mistaken. Running `$bpTitle` through `translate` makes that possible.

config/vufind/config.ini
demiankatz: @bbusenius, sorry for not being clear, but I think you misunderstood my intent here -- I think `bookplate_titles_field`, `bookplate_images_field`, etc. make perfect sense as names for configuration settings, and I wasn't asking you to change those. I just meant that the example values in your comments would be more realistic if you used `donor_str_mv` and `donor_code_str_mv` instead of `donor` and `donor_code`, because the `_str_mv` suffixes would allow those versions to work with the existing VuFind schema without requiring any modifications. I'd suggest reverting commit 1b9e290ae46e6e2d27999a40e416d25f397b95b5 and just editing the config file.

Again, sorry for putting you to extra work here! I'd make the fix myself, except that I don't seem to have write access to this branch.

module/VuFind/src/VuFind/Related/Bookplate.php
demiankatz: Right, that makes sense! The only thing to be careful of is that now, if you do have inconsistent numbers of values in the fields, you can no longer assume that `$this->bookplateImages[$i]` and `$this->bookplateThumbnails[$i]` will be set inside the loop, so you should probably add some default values to avoid PHP notices -- e.g. add `?? null` or `?? ''` (depending on what default value makes the most sense).

themes/bootstrap3/templates/Related/Bookplate.phtml
demiankatz: That's great! I wonder if it would actually make sense to use a separate text domain for bookplate titles, to avoid any risk of unexpected translation based on existing values in the language file. You could just prepend the title with "bookplate::" in the translate call, and then the translator would look for translations in `languages/bookplate/en.ini` instead of `languages/en.ini`. (And, of course, in the default case where no translations are defined, the behavior should remain exactly the same as ever).

demiankatz:
Thanks again for all the progress, @bbusenius -- I left just a handful of notes this time around. I'm also interested in whether you'd like me to work on the unit test development for this. If so, it would be more convenient if I could have write access to this branch, but if that's not possible, I can always make a pull request against your pull request.

config/vufind/config.ini
bbusenius: Ah yes, I took that in a bad direction, LOL. This is fixed now.

module/VuFind/src/VuFind/Related/Bookplate.php
bbusenius: I set this up with `?? ''`. I think that makes sense. I tested it to confirm I could force a PHP notice with bad data before and confirmed that this fixed it (it did).

themes/bootstrap3/templates/Related/Bookplate.phtml
bbusenius: Yes, definitely. I didn't know that was an option but I think it makes sense. If someone were to actually manage their bookplate titles in this way, it would give them a nice separate place to keep everything.

bbusenius:
Thanks for the great feedback @demiankatz! I made the changes. Regarding the unit tests, yes, I think it would be helpful if you handled that. I was trying to set some up for you but I haven't managed to do anything useful yet. I'm going to keep playing with it since it's something I'd like to improve on. I have lots of experience doing this in Django but haven't done it on the VuFind code base. It would be helpful for me to see what you do. In the meantime, I think I'm just going to play with a foobar test I have and see what I can learn.

I wasn't sure how to give you write permission to the branch so I just invited you as an outside collaborator to the repo. I should think that would give you the permissions you need.

module/VuFind/src/VuFind/Related/Bookplate.php
demiankatz: Great, thanks!

demiankatz:
Thanks, @bbusenius, I received the invitation, so I think I'm all set to build a test as soon as time permits (probably not before next week, but we'll see). I'll try to put something together that's a useful example, and please feel free to ask questions if there's anything I can help with. PHPUnit takes a little getting used to, but I've grown to appreciate it quite a lot.

demiankatz:
@bbusenius, I actually found time to write a test a little sooner than expected. As usual, it proved to be worth the effort, as it revealed a couple of bugs I hadn't noticed in my eyeball reviews. I also noticed that the PHP template had some short tags (`<?` instead of `<?php`) which were causing php-cs-fixer to fail, so I adjusted those as well.

I'm sure my test could be expanded to cover more edge cases -- I just tried to cover the most obvious cases in the smallest number of tests to get a baseline started. If you feel there are any important cases that we should add, please feel free. Otherwise, if you think this is done, let me know and I'll get it merged.

bbusenius:
@demiankatz sorry, I somehow missed your updates here. Yes, all of these changes make perfect sense. I can't believe I missed those PHP short tags! Anyhow, your test is fantastic. I was trying to write a much more minimal version of a similar test and couldn't figure out how to do it. I wasn't even close. Anyhow, I like this very much. It gives me a better idea about PHPUnit. Thank you very much for this contribution! I think this is ready to merge!

demiankatz:
Thanks, @bbusenius -- I've just merged this. I appreciate all the work (and patience) that got us here. :-)

PR #1835, opened 2021-01-22
XCNCIP ILS driver agency and patron id handling improvement

xmorave2:


xmorave2:
Not ready yet, I'll need to do some more testing

demiankatz:
Thanks, @xmorave2, I'll be happy to review and merge whenever you're ready!

demiankatz:
@xmorave2, is this ready for review yet, or still work in progress?

xmorave2:
@demiankatz I thought it is but, but it is not, converting back to draft, I need to check some more things to make it ready

demiankatz:
Thanks, @xmorave2; I see you've made more progress but left this in draft. I'll keep waiting for you before reviewing.

xmorave2:
@demiankatz We are going to do some deeper testing internally in few weeks, then I will see if there is more work needed, or if this is ready. Thanks

demiankatz:
Great! Good luck with the testing!

xmorave2:
We made some testing (and code changes) and it is ready for review

languages/AccountBlocks/en.ini
```diff
@@ -0,0 +1,9 @@
+Block Check Out = "You can't checkout more items"
+Block Electronic Resource Access = "You can't access electronic resources"
+Block Hold = "You can't place a hold"
+Block Recall = "You can't make a recall"
+Block Renewal = "You can't renew checkouts"
+Block Request Item = "You can't make a request on library material"
+Trap For Lost Card = "You card is reported as lost"
```
demiankatz: I think this should be "your card"

module/VuFind/src/VuFind/ILS/Driver/OAuth2TokenTrait.php
```diff
@@ -100,7 +100,7 @@ public function getNewOAuth2Token(
                 'Problem getting authorization token: Bad status code returned'
             );
         }
-        $tokenData = json_decode($response->getContent(), true);
+        $tokenData = json_decode($response->getBody(), true);
```
demiankatz, line 103: Just curious what motivated this change -- it does seem to make more sense to use the decoded content rather than the raw content, but just wanted to be sure it wasn't intentionally working the other way for some context. Also note that getContent is still being used both earlier and later in error/log messages in this same code. Should this be made more uniform?

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
```diff
@@ -212,11 +273,7 @@ public function loadPickUpLocations()
         } elseif ($this->config['Catalog']['pickupLocationsFromNCIP'] ?? false) {
             $this->loadPickUpLocationsFromNcip();
         } else {
-            throw new ILSException(
-                'XCNCIP2 ILS driver bad configuration. You should set up ' .
-                'one of these options: "pickupLocationsFile" or ' .
-                '"pickupLocationsFromNCIP"'
-            );
+            $this->pickupLocations = [];
```
demiankatz, line 305: Is there a reason to remove this exception? Is it no longer true that pickup location configuration is necessary? If we do change this behavior, should we at least adjust some notes in XCNCIP2.ini?

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
```diff
@@ -509,31 +557,45 @@ protected function getHoldingsForChunk($current, $aggregateId = null,
...
+                ? $isHoldable : false,
             'holdtype' => $this->getHoldType($status),
             'storageRetrievalRequest' => 'auto',
             'addStorageRetrievalRequestLink' => 'true',
         ];
         if (strtolower($status) === 'circulation status undefined') {
             $return['use_unknown_message'] = true;
         }
+        if (!empty($collection)) {
+            $return['collection_desc']= $collection;
```
demiankatz: Missing space in front of = here. Not sure why none of the style checkers complained.

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
```diff
@@ -1189,51 +1237,88 @@ public function getMyHolds($patron)
...
             'ns1:LookupUserResponse/ns1:UserOptionalFields/' .
             'ns1:UserAddressInformation/ns1:PhysicalAddress/' .
-            'ns1:UnstructuredAddress/ns1:UnstructuredAddressData'
+            'ns1:StructuredAddress/ns1:PostalCode'
         );
-        $address = explode("\n", trim((string)$address[0]));
+        $zip = !empty($zip) ? (string)$zip[0] : null;
+
+        if (empty($address1)) {
+            // TODO: distinguish between more formatting types; look
```
demiankatz, line 1318: Is this still a TODO, or have you actually done some of it here?

languages/AccountBlocks/en.ini
```diff
@@ -0,0 +1,9 @@
...
+Trap For Message = "You have a message from your library"
+Trap For Pickup = "Your requested material is available for pickup"
```
demiankatz: Should these "Trap" messages really be blocking users from performing actions in the ILS? Some of these seem more like informational notes rather than reasons to restrict behavior. Have you tested that the system behaves appropriately in these cases?

That aside, though, I do like the idea of having a text domain for block messages. I wonder if there is existing functionality in other drivers that could be improved to use this.

demiankatz (changes requested):
Thanks for this! I've left a few general comments. I can't really speak to the details of the XCNCIP2 implementation, though I certainly appreciate the simplification and standardization that you seem to have done here. I'm going to trust that your test coverage confirms that everything is behaving as expected. It would be nice to compare notes with other XCNCIP2 users, but I'm not sure if there are any at this stage. Let me know if you think it's worth seeking out further feedback, or if you are satisfied once you have responded to my comments here. I'm happy to merge this once you are happy with it -- it seems like an overall improvement, and if it causes problems, we can address them as they are discovered.

languages/AccountBlocks/en.ini
xmorave2: Fixed

module/VuFind/src/VuFind/ILS/Driver/OAuth2TokenTrait.php
xmorave2: json_decode could fail on raw http response content especially when Content-encoding = deflate. So I changed it, because I really experienced this issue... And yes, it should be consistant, also getBody() is always returning string type unlike getContent. It is done now.

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
xmorave2: Yes, there is reason - there are many systems, without any location defined either and still accepting requests. I updated comment in ini file a bit

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
xmorave2: Hm, strange. Anyway I fixed it.

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
xmorave2: Not really. There is pair element for UnstructuredAddressData called UnstructuredAddressType, which can contain almost anything which describes  the structure/formatting of data element. But it describes it in pretty vague way like "HTML", "XML", "New-line delimited text" and so on. So this comment is here for future possiblity of enhancement address parsing, but until we have a live system  using this, it is almost impossible to make concrete implementation.

languages/AccountBlocks/en.ini
xmorave2: "Trap For Message" and "Trap For Pickup" are defined as "Notify the User that ...." But "Trap For Lost Card" is defined as "Do not allow any transactions using this user card to proceed." But it should not block user ta make actions using on-line account...

So theoretically, all traps are more like notifies then blocks. I tweaked the blocks handling a bit.

I do think some drivers could benefit from AccountBlocks text domain. It depends if the block description is free text in given system or something predefined - if it is predefined it makes sense to try to translate the value using this new text domain.

I do not know all systems VuFind has driver for.

The getAccounBlocks method is defined in these driver: Alma, Demo, Koha, KohaILSDI, KohaRest, SierraRest and VoyagerRestful.

Voyager looks like it has some predefined block reasons, so it could be a candidate. Also SierraRest looks like it uses some kind of block reason codes instead of free text. KohaRest looks like a candidate too.

Older Koha and KohaILSDI drivers uses only text comment from the db, so they could not use translation.

xmorave2:
Thanks for quick review @demiankatz. I don not know if anybody else is using ncip too, but we are using it heavily and are going to maintain it in the future. So yes, when problem arises I think it will be fixed ;)

module/VuFind/src/VuFind/ILS/Driver/OAuth2TokenTrait.php
demiankatz: Makes sense -- thanks!

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
demiankatz: Looks good!

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
demiankatz: Makes sense!

languages/AccountBlocks/en.ini
demiankatz: Thanks for the improvements -- this does seem better, and it also makes it easy to subclass and change behavior if local needs differ from defaults.

Regarding the AccountBlocks text domain, I agree that most existing drivers don't look like they would fit with that approach since they're using free-form text pulled from the ILS. It looks to me like the KohaRest driver is the strongest candidate for using this text domain, which might be possible if we just move some strings from the main language files into the new text domain. I'd be interested to hear what @EreMaijala thinks of that idea; I want to be sure we're all on the same page about this approach before merging this PR, but I think once we have discussed that last detail, this is ready to go.

EreMaijala: Superficially the text domain makes sense, but there are several things to consider:

1. I'd like to avoid terminology that's specific to a certain interface and aim for more generic block reasons. And when the translation text is really a code that's not displayable as is, I'd rather use keys that look more like codes, such as 'request_block' or 'message_from_library'.
2. When the text is not a block, AccountBlocks is not the right domain name. ILSMessages or something maybe?
3. I think we should take a broader look on strings coming from ILS drivers and how they're translated. Particularly when connecting to multiple ILS's with MultiBackend, there's the need to translate strings independently per ILS. That's where Alma driver's translation prefix setting is handy, as you can define a per-library prefix or text domain. And the KohaRest driver should support the same mechanism. Things would be more clear if we had a well-defined list of common translation keys that the drivers should use for block statuses etc. that probably don't require library-specific overrides (compared to something like location codes which definitely do).

xmorave2: Thanks for your feedback @EreMaijala 

ad 1) If we could define a set of block reasons that will work with the variety of systems, it would be probably best solution

ad 2) I am not against adding another domain and make it all semantically more precise.. but maybe then would be good idea to add a new method to drivers and reflect it in flash messages that way (maybe not styled as warning, but as info)

ad 3) I agree, that there is room for improvement I am just not sure how both ways would work together... But if we combine these two approaches we must be able to say clearly which strings  should be translated by default/common domain and which could be translated with prefix/domain configured per ILS driver

xmorave2: For now I will try to do something about the 1)

demiankatz: Thanks, @xmorave2 -- I agree that as a starting point, it makes sense to use more code-like keys in the translation file and to rename the text domain to ILSMessages (or something similar) to allow us to incorporate a broader range of ILS-related text in the future. As @EreMaijala points out, it would be really useful to group these messages together so that future driver developers can more easily find and reuse them. Regarding point 3, I think the solution could be as simple as having a configurable text domain in XCNCIP2.ini which defaults to ILSMessages -- then it's easy to create a custom text domain for per-driver messages, but by default, everything just works "normally."

There's clearly a lot more work that we could do here, but I'll be happy if we just set the stage effectively here, and we can do further work in subsequent PRs.

xmorave2: I moved away from direct messages to codes and made translation domain configurable defaulting to 'ILSMessages'

demiankatz:
Thanks, @xmorave2, this looks good to me. @EreMaijala, any final thoughts before I merge this?

languages/ILSMessages/en.ini
```diff
@@ -0,0 +1,7 @@
+available_for_pickup = "Your requested material is available for pickup"
+checkout_block = "You can't checkout more items"
+electronic_resources_block = "You can't access electronic resources"
+lost_card = "Your card is reported as lost"
+message_from_library = "You have a message from your library"
+renewal_block = "You can't renew checkouts"
+request_block = "You can't make a request on library material"
```
EreMaijala: This one is a bit broad. I think it would be better to limit this to something like `"You can't place a request on this material"`, or make the code more specific. Also, I think we need to review the existing messages and align the terminology used (e.g. "make a request" => "place a request", "can't" => "cannot").

languages/ILSMessages/en.ini
```diff
@@ -0,0 +1,7 @@
+available_for_pickup = "Your requested material is available for pickup"
+checkout_block = "You can't checkout more items"
```
EreMaijala: "checkout" => "check out"?

EreMaijala (changes requested):
Generally, it's looking good already, but I think we should align the messages with the existing ones. Since we already have quite a few messages, the difference between them the use cases is not quite clear (e.g. `Default::hold_available`vs. `ILSMessages::available_for_pickip`).

EreMaijala:
Further on, it would probably make sense to move some of the existing messages into ILSMessages domain, but that can be a separate task.

languages/ILSMessages/en.ini
demiankatz: Yes, I agree, "check out" would be better.

languages/ILSMessages/en.ini
demiankatz: I think the question of specificity depends on the scope of the NCIP message that we're translating here -- if it's a block message, it may literally be saying "this user is not allowed to place requests" without being more specific. If that's the case, maybe changing the key from request_block to requests_blocked would make it more clear that it is broad in scope.

Regarding "can't" vs. "cannot," it does appear that our existing language always uses "cannot" instead of "can't." I hadn't realized that, but we might as well change it to be consistent!

I agree that changing "make" to "place" would also improve consistency.

demiankatz:
Thanks, @EreMaijala, for giving this a closer look than I did! Regarding `Default::hold_available` vs. `ILSMessages::available_for_pickup`, I think the scope of the messages is a little different -- one is a status message that may be request-type-specific, while the other is a notification of availability. Maybe it would make sense to rename `ILSMessages::available_for_pickup` to `ILSMessages::available_for_pickup_notification`.

languages/ILSMessages/en.ini
xmorave2: I changed the translations as suggested. And changed the code request_block to  requests_blocked, as it looks like it is probably that case - but hard to say as NCIP is not very specific standard and there could be some responders that behaves differently...

xmorave2:
> Thanks, @EreMaijala, for giving this a closer look than I did! Regarding `Default::hold_available` vs. `ILSMessages::available_for_pickup`, I think the scope of the messages is a little different -- one is a status message that may be request-type-specific, while the other is a notification of availability. Maybe it would make sense to rename `ILSMessages::available_for_pickup` to `ILSMessages::available_for_pickup_notification`.

I changed this too

languages/ILSMessages/en.ini
xmorave2: fixed

EreMaijala:
Thanks, looking fine now. One nit that I'd maybe find better would be to make the pickup notification something like "You have material available for pickup" to better indicate that there's *something* available for pickup in case the user has several requests, but the current one is fine as well. The reason for giving this the scrutiny is that I'd like to avoid any conflicts with any existing messages that could be moved to the new domain.

EreMaijala (approved):

xmorave2:
> Thanks, looking fine now. One nit that I'd maybe find better would be to make the pickup notification something like "You have material available for pickup" to better indicate that there's _something_ available for pickup in case the user has several requests, but the current one is fine as well. The reason for giving this the scrutiny is that I'd like to avoid any conflicts with any existing messages that could be moved to the new domain.

Thanks @EreMaijala, I changed the wording as you suggested

demiankatz:
Thanks again to both of you -- it is now merged!

PR #1836, opened 2021-01-26
Fix problem with i18n of PHTML templates.

demiankatz:


demiankatz:
@xmorave2, while testing the TemplateBased ContentBlock, I discovered that .phtml templates with language suffixes were not getting loaded as expected. I believe this simple fix (and corresponding unit test) should take care of the problem. Do you mind reviewing and letting me know if I've overlooked anything important? Thanks!

xmorave2 (approved):
Good catch @demiankatz, this is really bug. Thanks for patching it.

demiankatz:
Great, thanks for the review, and glad I could help!

PR #1838, opened 2021-01-28
Add configuration for API settings.

demiankatz:


demiankatz:
@EreMaijala, I need to override the max result limit for web results in my local instance, so I thought having an .ini file for some API settings might be useful. Any suggestions or objections?

EreMaijala:
@demiankatz Did you already consider and reject the idea of putting these into the .ini file for each corresponding search type (e.g. searches.ini)? For me it would seem more natural to have everything that controls the search in one place especially since many of the settings already affect the API too.

demiankatz:
@EreMaijala, I hadn't really considered that possibility, but you're right that it could make more sense to have an [API] section in all those existing search configuration files. Some possible advantages of the current approach:

1.) Having all the API settings in a single file makes it easier to globally override things -- you can either just change the [Search] section and have it affect everything, or you can configure different settings for all of the endpoints. If we put API settings in, for example, searches.ini and website.ini, I think it would be necessary to duplicate customizations, because creating the precedent of having website.ini inherit settings from searches.ini seems potentially confusing.

2.) I think if we want to pull these settings from the individual config files, we'll probably want to build some functionality into the search Options classes, and then instantiate those in the controller's constructor or factory to pull settings out. I think that's entirely feasible, but it's a bit heavier than the comparatively simple implementation here.

3.) Having an api.ini advertises the existence of the API, which users may otherwise be unaware of (but so would an [API] section in searches.ini).

Looking at this from a purely API-centric perspective, I think the api.ini approach is slightly preferable because it's simpler and requires less redundant configuration. But in a more holistic view, those advantages may be outweighed by avoiding the creation of yet another .ini file and better utilizing existing conventions. I could really be persuaded either way.

Does any of this influence your opinion one way or another? Unless you now feel that api.ini is the way to go, I'll take some time to reimplement this using Options classes, etc., and see if I run into any significant obstacles that I'm not thinking of right now.

EreMaijala:
@demiankatz I'd be happy to buy your reasoning, but before that I'll raise one minor concern: if we might ever have more than a few API-related settings, having only a single section for them might become a bit convoluted. Do you think a backend-specific API configuration file would work? I also think it would be more clear if each controller would specify the file or section to use instead of the controller class name magic.

xmorave2:
Just for information:
We have raised up limit for Search2 in our code this way: https://github.com/moravianlibrary/Knihovny.cz/blob/master/module/KnihovnyCzApi/src/KnihovnyCzApi/Controller/Search2ApiController.php#L47

Yes, it has concerns like you need to create a subclass and you can't configure it easily in multi-instance environment, but in our case it works good enough ;)

But your solution @demiankatz would be more flexible, so I like it ;)

demiankatz:
Thanks again for the feedback, @EreMaijala and @xmorave2. As discussed yesterday, I tried refactoring the config loading into the search Options class, and it definitely feels better-organized and more flexible this way. I think this is now ready to merge. 

I opted to keep it simple for now, with an [API] section in each search-specific config file, loaded via a method in the Options class. The association between the controllers and the settings are based on the configured search class ID. I think this is better than proliferating new .ini files and extra settings, at least until we have a use case to justify the extra infrastructure. I'm certainly open to making further adjustments, though!

Any objections or remaining ideas?

EreMaijala (approved):
@demiankatz Thanks for revising. I expected it to be more complex than this. Looks very good to me. :)

EreMaijala:
@demiankatz I like it a lot more now, and no further objections. :)

demiankatz:
Great, merging now!

PR #1842, opened 2021-02-02
Disable session writes in getRecordCover AJAX - VUFIND-716

xrosecky:


demiankatz:
Thanks for catching that potential problem!

PR #1843, opened 2021-02-04
Load record drivers in a batch for lists in MyResearch.

EreMaijala:
This can provide a significant performance improvement for a large number of rows especially when Solr is not on the same server but requires a network round-trip.

### DIFF
--- a/module/VuFind/src/VuFind/Controller/MyResearchController.php
+++ b/module/VuFind/src/VuFind/Controller/MyResearchController.php
@@ -1570,20 +1570,7 @@ class MyResearchController extends AbstractBase
 
         if ($driversNeeded) {
             $recordLoader = $this->serviceLocator->get(\VuFind\Record\Loader::class);
-            try {
-                $drivers = $recordLoader->loadBatch($driversNeeded);
-            } catch (\Exception $e) {
-                // Loading the batch failed, try one by one:
-                foreach ($driversNeeded as $i => $current) {
-                    try {
-                        $drivers[$i] = $recordLoader
-                            ->load($current['id'], $current['source'], true);
-                    } catch (\Exception $e) {
-                        // Ignore record loading exceptions...
-                    }
-                }
-            }
-
+            $drivers = $recordLoader->loadBatch($driversNeeded, true);
             foreach ($drivers as $i => $driver) {
                 $fines[$i]['driver'] = $driver;
                 if (empty($fines[$i]['title'])) {
--- a/module/VuFind/src/VuFind/Controller/MyResearchController.php
+++ b/module/VuFind/src/VuFind/Controller/MyResearchController.php
@@ -1126,21 +1126,35 @@ class MyResearchController extends AbstractBase
     }
 
     /**
-     * Get a record driver object corresponding to an array returned by an ILS
-     * driver's getMyHolds / getMyTransactions method.
+     * Get record driver objects corresponding to an of record arrays returned by an
+     * ILS driver's methods such as getMyHolds / getMyTransactions.
      *
-     * @param array $current Record information
+     * @param array $records Record information
      *
-     * @return \VuFind\RecordDriver\AbstractBase
+     * @return array
      */
-    protected function getDriverForILSRecord($current)
+    protected function getDriversForILSRecords($records)
     {
-        $id = $current['id'] ?? '';
-        $source = $current['source'] ?? DEFAULT_SEARCH_BACKEND;
-        $record = $this->serviceLocator->get(\VuFind\Record\Loader::class)
-            ->load($id, $source, true);
-        $record->setExtraDetail('ils_details', $current);
-        return $record;
+        if (!$records) {
+            return [];
+        }
+        $ids = array_map(
+            function ($current) {
+                return [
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/Controller/MyResearchController.php
```diff
@@ -1126,21 +1126,35 @@ public function deletelistAction()
     }
 
     /**
-     * Get a record driver object corresponding to an array returned by an ILS
-     * driver's getMyHolds / getMyTransactions method.
+     * Get record driver objects corresponding to an of record arrays returned by an
```
demiankatz: I think there's a missing word here -- should this be "array of record arrays?"

module/VuFind/src/VuFind/Controller/MyResearchController.php
```diff
@@ -1126,21 +1126,35 @@ public function deletelistAction()
...
+     * ILS driver's methods such as getMyHolds / getMyTransactions.
      *
-     * @param array $current Record information
+     * @param array $records Record information
      *
-     * @return \VuFind\RecordDriver\AbstractBase
+     * @return array
```
demiankatz: I think we could use the more specific typehint `\VuFind\RecordDriver\AbstractBase[]` here. (I realize that we currently use `array` all over the place, but this is one of the things that Psalm may encourage us to improve -- might as well start getting in the habit).

module/VuFind/src/VuFind/Controller/MyResearchController.php
```diff
@@ -1539,29 +1553,48 @@ public function finesAction()
...
+            $recordLoader = $this->serviceLocator->get(\VuFind\Record\Loader::class);
             try {
-                if (strlen($row['id'] ?? '') > 0) {
-                    $source = $row['source'] ?? DEFAULT_SEARCH_BACKEND;
-                    $row['driver'] = $this->serviceLocator
-                        ->get(\VuFind\Record\Loader::class)
-                        ->load($row['id'], $source);
-                    if (empty($row['title'])) {
-                        $row['title'] = $row['driver']->getShortTitle();
+                $drivers = $recordLoader->loadBatch($driversNeeded);
```
demiankatz: Could you just use the `$tolerateBackendExceptions` flag of the `loadBatch()` method here to eliminate the need for the record-by-record fallback, or is there a different specific case that is intended to protect against?

demiankatz (commented):
Thanks for this; I had forgotten how inefficient this was. I've left a few very minor comments. Once you have addressed those, I'll double-check that all the tests still pass.

module/VuFind/src/VuFind/Controller/MyResearchController.php
EreMaijala: Right, and I added type hints as well. Too bad you can't specify the type of array in return hint.

module/VuFind/src/VuFind/Controller/MyResearchController.php
EreMaijala: Fixed.

module/VuFind/src/VuFind/Controller/MyResearchController.php
EreMaijala: The old code tolerates any kind of exception, so I tried to follow that here as well. Our custom version of the old getDriverForILSRecord actually has been added to make any kind of exception tolerable. Do you think it would make sense to incorporate support for that in Loader's load/loadBatch?

demiankatz: I guess the question is whether there's any kind of exception other than a backend exception that would occur inside loadBatch() that would result in some records loading but others being skipped. If there's a realistic, likely scenario that we need to work around, then I'm certainly open to making the record loader more flexible -- perhaps we could replace $tolerateBackendException with a $toleranceLevel parameter that could accept a boolean (for backward compatibility with existing behavior) or a flag constant (to add different behaviors). But if the existing behavior covers the scenario where we would need to try again with one-by-one loading, and any other exception just indicates that we can't load records at all, then I would be in favor of simplifying the code. Either way, I think it would be a good idea to use the logger to report on any exceptions we encounter, to ensure that we don't accidentally mask problems completely.

EreMaijala: You're right, there should only be backend exceptions from the backend, and I verified that with Solr in particular. I simplified the code to just use loadBatch.

module/VuFind/src/VuFind/Controller/MyResearchController.php
demiankatz: Maybe eventually! PHP 8 at least makes aggregate typehints possible, so further improvements may come along eventually. For now, as long we capture things correctly in the PHPDoc, we can automatically fix the code if and when the language supports richer expression in the future.

module/VuFind/src/VuFind/Controller/MyResearchController.php
demiankatz: Thanks!

demiankatz:
All tests are passing -- merging now!

PR #1844, opened 2021-02-05
Improve typehints/comments related to deduplication listener

xrosecky:
We use phpstan for code analysis and we found some QA issues when extending deduplication listener in our module.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Solr/DeduplicationListener.php
+++ b/module/VuFind/src/VuFind/Search/Solr/DeduplicationListener.php
@@ -33,11 +33,11 @@
  */
 namespace VuFind\Search\Solr;
 
+use Interop\Container\ContainerInterface;
 use Laminas\EventManager\EventInterface;
-
 use Laminas\EventManager\SharedEventManagerInterface;
-use Laminas\ServiceManager\ServiceLocatorInterface;
-use VuFindSearch\Backend\BackendInterface;
+
+use VuFindSearch\Backend\Solr\Backend;
 
 /**
  * Solr merged record handling listener.
@@ -54,14 +54,14 @@ class DeduplicationListener
     /**
      * Backend.
      *
-     * @var BackendInterface
+     * @var Backend
      */
     protected $backend;
 
     /**
      * Superior service manager.
      *
-     * @var ServiceLocatorInterface
+     * @var ContainerInterface
      */
     protected $serviceLocator;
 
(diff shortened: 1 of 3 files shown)


module/VuFind/src/VuFind/Search/Solr/DeduplicationListener.php
```diff
@@ -89,18 +89,18 @@ class DeduplicationListener
...
-     * @param string                  $searchConfig     Search config file id
-     * @param string                  $dataSourceConfig Data source file id
-     * @param bool                    $enabled          Whether deduplication is
-     * enabled
+     * @param Backend            $backend          Search backend
+     * @param ContainerInterface $serviceLocator   Service locator
+     * @param string             $searchConfig     Search config file id
+     * @param string             $dataSourceConfig Data source file id
+     * @param bool               $enabled          Whether deduplication is
+     *                                             enabled
```
demiankatz: Whitespace is not required here; PHPDoc does not require this kind of text alignment, as it reduces readability for long descriptions.

module/VuFind/src/VuFind/Search/Solr/DeduplicationListener.php
```diff
@@ -33,11 +33,11 @@
  */
 namespace VuFind\Search\Solr;
 
+use Interop\Container\ContainerInterface;
```
demiankatz: I would recommend using Psr\Container\ContainerInterface here instead of Interop\Container\ContainerInterface; both currently work, but the Interop version is deprecated and will eventually be removed, so using the Psr version will save us time in the future.

demiankatz (commented):
Thanks, @xrosecky, I've left a couple of minor comments.

Some of the work here includes things that I've been thinking about applying more broadly to the code for VuFind 8 -- such as replacing ServiceLocatorInterface references with the more generic/flexible ContainerInterface. That might be a worthwhile project to undertake in a separate PR (though I haven't investigated deeply yet, so I'm not sure how large the scope of the change might be).

I'm also requesting a review from @EreMaijala, since this is largely his code, and I want to be sure he doesn't have any concerns about making some of the typehints more specific and more general.

In any case, thanks for your work on improving code quality -- there's a lot more we can do, but every bit helps!

demiankatz:
I'm also renaming this PR to make its title a bit more specific.

EreMaijala (approved):
Looks good to me. Thanks!

demiankatz:
Thanks again, everyone -- I've merged this.

PR #1845, opened 2021-02-05
Fix fixPrimoFacetValue to not destroy UTF-8 strings

EreMaijala:
fixPrimoFacetValue was actually destroying UTF-8 strings causing the HTML escaper to throw an exception.


demiankatz:
I've backported this to release-7.1 since I think it qualifies as a bug fix. Thanks!

PR #1846, opened 2021-02-06
Simplify and reorganize tests

demiankatz:
This PR is designed to make VuFind's test suite easier to understand and more logical. It replaces a deep and unintuitive test class hierarchy with more focused reusable traits that help with specific testing tasks. It more explicitly separates features that are designed to interact with a live system for integration testing from those that are meant to help with more isolated unit-testing. It also eliminates a lot of outdated service manager setup code, much of which was unnecessary and all of which was confusing.

TODO
- [x] Finish refactoring/simplifying code
  - [x] Eliminate VuFindTest\Unit\DbTestCase
  - [x] Eliminate calls to getServiceManager()
  - [x] Refactor remaining VuFindTest\Unit\TestCase features to traits, if appropriate
  - [x] Eliminate other VuFindTest\Unit base classes in favor of traits, if appropriate
- [x] Ensure that all tests still pass in all contexts (i.e. MySQL & PostgreSQL environments, etc.)
- [x] Check whether wiki testing documentation needs to be updated


demiankatz:
@EreMaijala / @crhallberg -- I don't expect either of you to carefully look at every single line of code here, but I'm asking for a review just to see if you have any suggestions or objections to the work I've done here.

The nice thing about refactoring test code is that it's easy to be confident it worked right since the outcome is measurable. ;-)

In any case, the main work I've done here is designed to make test cases more stand-alone, and to put shared functionality in more clearly-labeled and appropriate places. To that end I've:

- Replaced most of the base classes in the VuFindTest\Unit namespace with traits in the VuFindTest\Feature namespace, and changed the majority of tests to directly extend PHPUnit\Framework\TestCase. (I also adjusted the namespace on existing traits, for better organization). The base classes that I left in place are very specific and are used in fairly limited contexts -- they might be worth revisiting later, but I figured they weren't hurting anything for now.
- I moved the base Mink test class from VuFindTest\Unit to VuFindTest\Integration, since it is not used for unit testing.
- I get rid of all the ugly and confusing "getServiceManager()" logic in the test classes, replacing it with more logical traits. For the limited number of tests that actually require a real running database and Solr instance, there are now "Live" traits that make them easier to identify.

I'll do some final testing (e.g. making sure things still work in a PostgreSQL environment; MySQL is currently fine) and adjust documentation once the approach is finalized... I'd just like a little feedback from others before merging a change this large.

module/VuFind/src/VuFindTest/Feature/LiveDetectionTrait.php
```diff
@@ -23,37 +24,37 @@
...
      *
-     * @return void
+     * @return bool
      */
-    public function setUp(): void
+    public function continuousIntegrationRunning()
     {
-        $this->container = new MockContainer($this);
+        // We'll assume that if the CI Solr PID is present, then CI is active:
+        return file_exists(__DIR__ . '/../../../../../local/vufindtest.pid');
```
EreMaijala: Would it be possible to base this determination on an environment variable or something else more explicit?

EreMaijala (approved):
I didn't give all the changes full attention, but at least i skimmed through it all. This makes sense to me and seems to make things much cleaner. I only added one comment, and it's not very important if you think the solution will work fine as is.

module/VuFind/src/VuFindTest/Feature/LiveDetectionTrait.php
demiankatz: This file gets created here as part of the test instance spin up: https://github.com/vufind-org/vufind/blob/dev/build.xml#L339 -- I can't remember all of the history here, and I'm not sure without looking if this is still really a Solr PID file, or if it used to be and now it's just a relic used to set a flag for the tests. I'm open to changing it to something better, but I'm not sure what the best approach would be; the important thing is that whatever we use as a marker, it has to be set by the startup action and removed by the shutdown action, because most of the tests will run correctly regardless of whether or not the test environment is active, but a few need to know the state. Maybe a good starting point would be to review whether this is really a pid file or not, and at least rename it to something more appropriate if the name is no longer appropriate. I'll take a closer look next week when I'm back from my mini-vacation.

demiankatz: I had a little time to investigate, and it does appear that the real Solr PID file got replaced with a fake placeholder at some point in time. I've opened [this PR](https://github.com/vufind-org/vufind/pull/1853) to restore reality to the situation. Does that make sense to you, or do you think we should come up with a different approach?

demiankatz: I'm resolving this conversation since the work was resolved by PR #1853 (see discussion there for more details, if interested).

demiankatz:
All tests are still passing in both MySQL and PostgreSQL environments! (Apart from the one known broken PostgreSQL test, which is unchanged by this refactoring). I plan to merge this as soon as I have had time to adjust the documentation in the wiki.

PR #1848, opened 2021-02-09
Replace Service Manager references with more generic Containers.

demiankatz:


demiankatz:
Note: `Interop\Container\ContainerInterface` is deprecated and should eventually be replaced with `Psr\Container\ContainerInterface`. I am using it here for consistency with existing code, on the assumption that we'll change all of these references at once in the near future.

PR #1850, opened 2021-02-11
SolrPrefix autocomplete

xrosecky:
SolrPrefix autocomplete using edge n-gram tokenizer in Solr for prefix queries, eg. each term in query is considered as a prefix. Best works for single valued fields.

### DIFF
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -542,6 +542,10 @@ sort = "last_indexed desc"
 ;       call numbers using the custom CallNumber search handler.
 ; Tag
 ;       Provide suggestions from the local database of tags.
+; SolrPrefix:[Autocomplete field]:[Display field]
+;       Perform a search against autocomplete field with edge-n gram tokenizer,
+;       each word in query is handled as a prefix. Display field must be copied
+;       from autocomplete field and be stored and indexed or with docValues.
 ;
 ; You can build your own autocomplete modules if you wish.  See the developer's
 ; guide here:
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -578,6 +578,7 @@ CallNumber = "SolrCN"
 ISN = "Solr:ISN:isbn,issn"
 Coordinate = "None"
 tag = "Tag"
+;Title = "SolrPrefix:title_autocomplete:title"
 
 ; When snippets are enabled, this section can be used to display captions based on
 ; the Solr fields from which the snippets were obtained.  Keys are the names of Solr
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -578,7 +578,7 @@ CallNumber = "SolrCN"
 ISN = "Solr:ISN:isbn,issn"
 Coordinate = "None"
 tag = "Tag"
-;Title = "SolrPrefix:title_autocomplete:title"
+;Title = "SolrPrefix:title_autocomplete:title_sort"
 
 ; When snippets are enabled, this section can be used to display captions based on
 ; the Solr fields from which the snippets were obtained.  Keys are the names of Solr
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -552,7 +552,11 @@ sort = "last_indexed desc"
 ;       field, preferably single-valued, for display purposes, because display
 ;       of values by this option relies on faceting -- be sure to choose an
 ;       appropriate field type, such as a simple _str field, to ensure correct
-;       display.
+;       display. For example, to switch title suggestions to use the edge n-gram
+;       method, you could index titles into a new dynamic title_autocomplete
+;       field, then change the Title configuration in [Autocomplete_Types] to:
+;       "SolrPrefix:title_autocomplete:title_sort" (note the use of title_sort
+;       for display, because it is a string field and thus suitable for faceting).
 ;
 ; You can build your own autocomplete modules if you wish.  See the developer's
 ; guide here:
@@ -578,7 +582,6 @@ CallNumber = "SolrCN"
 ISN = "Solr:ISN:isbn,issn"
 Coordinate = "None"
 tag = "Tag"
-;Title = "SolrPrefix:title_autocomplete:title_sort"
 
 ; When snippets are enabled, this section can be used to display captions based on
 ; the Solr fields from which the snippets were obtained.  Keys are the names of Solr
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -543,9 +543,16 @@ sort = "last_indexed desc"
 ; Tag
 ;       Provide suggestions from the local database of tags.
 ; SolrPrefix:[Autocomplete field]:[Display field]
-;       Perform a search against autocomplete field with edge-n gram tokenizer,
-;       each word in query is handled as a prefix. Display field must be copied
-;       from autocomplete field and be stored and indexed or with docValues.
+;       Perform a search against [Autocomplete field] using the edge n-gram
+;       tokenizer. Each word in the user's query is handled as a prefix. Values
+;       from [Display field] will be displayed for matching records. It is
+;       recommended that you index values for this handler into fields of type
+;       "text_autocomplete" (which are available using the dynamic field suffix
+;       *_autocomplete). These values should also be copied into a separate
+;       field, preferably single-valued, for display purposes, because display
+;       of values by this option relies on faceting -- be sure to choose an
+;       appropriate field type, such as a simple _str field, to ensure correct
+;       display.
 ;
 ; You can build your own autocomplete modules if you wish.  See the developer's
 ; guide here:
--- a/module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
+++ b/module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * Solr Autocomplete Module
+ * Solr Prefix Autocomplete Module
  *
  * PHP version 7
  *
--- a/module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
+++ b/module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
@@ -77,6 +77,13 @@ class SolrPrefix implements AutocompleteInterface
     protected $facetField;
 
     /**
+     * Facet limit, can be overridden in subclasses
+     *
+     * @var int
+     */
+    protected $limit = 10;
+
+    /**
      * Constructor
      *
      * @param \VuFind\Search\Results\PluginManager $results Results plugin manager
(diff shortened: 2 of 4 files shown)


module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
```diff
@@ -0,0 +1,175 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ *
+ * @category VuFind2
```
demiankatz: Category should be VuFind, rather than VuFind2. (We used "VuFind2" while 1.x was still a semi-active project, but now that we've moved on from there, we reverted the package name back to "VuFind").

module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
```diff
@@ -0,0 +1,175 @@
...
+ * @link     http://vufind.org/wiki/vufind2:autosuggesters Wiki
+ */
+namespace VuFind\Autocomplete;
+
+/**
+ * Solr autocomplete module with prefix queries using edge N-gram filter
+ *
+ * This class provides suggestions by using the local Solr index.
+ *
+ * @category VuFind2
```
demiankatz: See above regarding category.

module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ *
+ * @category VuFind2
+ * @package  Autocomplete
+ * @author   Vaclav Rosecky <vaclav.rosecky@mzk.cz>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     http://vufind.org/wiki/vufind2:autosuggesters Wiki
```
demiankatz, line 27: This is also an outdated link; please use `https://vufind.org/wiki/development:plugins:autosuggesters` instead.

module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
```diff
@@ -0,0 +1,175 @@
...
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
```
demiankatz: This is an outdated address; this line should be:

`* Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA`

solr/vufind/biblio/conf/schema.xml
```diff
@@ -232,6 +244,7 @@
    <dynamicField name="*_boolean" type="boolean" indexed="true" stored="true"/>
    <dynamicField name="*_geo" type="geo" indexed="true" stored="true" multiValued="false" />
    <dynamicField name="*_geo_mv" type="geo" indexed="true" stored="true" multiValued="true" />
+   <dynamicField name="*_autocomplete" type="text_autocomplete" indexed="true" stored="true" multiValued="true"/>
```
demiankatz: Would it be more clear to add _mv on the end of this suffix for consistency with other multi-valued suffixes? Or should this only be used with single-valued fields, in which case we should set multiValued to false here.

It seems like this approach might lead to strange results with multi-valued fields, unless I am missing something, because if I have two values, A and B, and my search query matches A, won't the recommendation list include both A *and* B?

config/vufind/searches.ini
```diff
@@ -542,6 +542,10 @@ sort = "last_indexed desc"
 ;       call numbers using the custom CallNumber search handler.
 ; Tag
 ;       Provide suggestions from the local database of tags.
+; SolrPrefix:[Autocomplete field]:[Display field]
+;       Perform a search against autocomplete field with edge-n gram tokenizer,
```
demiankatz: I have some thoughts on making this comment a little more explicit, but before I suggest revised text, let's finish discussing the other details of the PR to make sure I understand the intent correctly.

demiankatz (changes requested):
Thanks for sharing this! I had a few thoughts and suggestions based on a quick first look. Many of these are minor corrections to the comment header on your class. If you have a template set up in your code editor, perhaps it needs to be updated. I apologize if you got these outdated settings from other current VuFind code -- after seeing the issues here, I reviewed the existing code in the dev branch and found and fixed a handful of outdated comments.

solr/vufind/biblio/conf/schema.xml
xrosecky: Yes, you are right, it works best for single-value fields. I will fix it in schema.xml.

config/vufind/searches.ini
demiankatz: How about something like:

Perform a search against [Autocomplete field] using the edge n-gram tokenizer. Each word in the user's query is handled as a prefix. Values from [Display field] will be displayed for matching records. It is recommended that you index values for this handler into fields of type "text_autocomplete" (which are available using the dynamic field suffix *_autocomplete). These values should also be copied into a stored field (or a field with docValues) for display purposes.

demiankatz: One question, though: is the last sentence actually accurate? Should it be "string field" rather than "stored field"?

xrosecky: Yes, field must be stored or with docValues so faceting on that field is working.

demiankatz: Of course, makes sense. So perhaps the comment could be revised to:

Perform a search against [Autocomplete field] using the edge n-gram tokenizer. Each word in the user's query is handled as a prefix. Values from [Display field] will be displayed for matching records. It is recommended that you index values for this handler into fields of type "text_autocomplete" (which are available using the dynamic field suffix *_autocomplete). These values should also be copied into a stored field (or a field with docValues) for display purposes, because display of values by this option relies on faceting.

demiankatz: Actually, wait a moment, maybe it doesn't entirely make sense -- I thought faceting ignored stored values and always used indexed values instead. Am I mistaken?

xrosecky: No, you are right, field used for faceting is not required to be stored. Should I fix the comments accordingly?

demiankatz: Sure. Maybe the easiest solution is to change the last sentence of my proposed revision to something like "These values should also be copied into a separate field for display purposes, because display of values by this option relies on faceting -- be sure to choose an appropriate field type, such as a simple _str field, to ensure correct display."

xrosecky: OK, I fixed the comments in the latest commit.

demiankatz: Thanks, looks good!

demiankatz (approved):

demiankatz:
This looks good to me, but I'm curious if @EreMaijala has any suggestions before I merge it.

module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
```diff
@@ -0,0 +1,175 @@
+<?php
+
+/**
+ * Solr Autocomplete Module
```
EreMaijala: Nit: "Solr Prefix Autocomplete Module"?

EreMaijala (commented):
I might be missing something obvious, but I can't see where and how the [Display Field] in config is handled nor how it would relate to the faceted values. Could you clarify that a bit?

demiankatz:
@EreMaijala, if I'm understanding correctly, this performs a search using [Autocomplete field] and includes [Display field] as a facet within the results. It then returns all of the facet values retrieved in the search results as its suggestions. @xrosecky can correct me if I'm misunderstanding, of course. If that is true, do you think we need to make it more clear in the .ini comment or the PHP code?

xrosecky:
Yes, you are right.

EreMaijala:
@demiankatz Oops, I just misread the code. What's a bit unclear here is what I assume to be the idea behind this. From the comment you get the feeling that the _autocomplete field is the main thing and then you should add a new display field for it, while I assume it's more like the other way round, and you'd normally want to add an autocomplete field for an existing single-valued field. Perhaps an example with e.g. the title field would make it more clear?

demiankatz:
Thanks, @EreMaijala, I agree that a real-world example might make this a little more clear. @xrosecky, would you be willing to add one for clarity?

xrosecky:
OK, I added example configuration.

demiankatz:
Thanks, @xrosecky, but will that example actually work? `title` is a text field, so I would expect that to yield garbled results if used as a facet field. Would it actually be better to use `title_sort`, since that is a string field?

xrosecky:
Sorry, fixed in the latest commit.

demiankatz:
Thanks, @xrosecky! I've moved the example into the description of the handler in b5ff398 to provide a little more context. @EreMaijala, are you comfortable with merging this now?

module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
```diff
@@ -0,0 +1,175 @@
...
+        $results = [];
+        try {
+            $params = $this->searchObject->getParams();
+            $rawQuery = $this->autocompleteField . ':(' .
+                $this->mungeQuery($query) . ')';
+            $params->setBasicSearch($rawQuery);
+            $params->addFacet($this->facetField);
+            $params->setLimit(0);
+            $params->setFacetLimit($this->limit);
+            $this->searchObject->getResults();
```
EreMaijala, line 124: One more thing: you could optimize the query a bit by disabling extra functionality:
```
            $options->disableHighlighting();
            $options->spellcheckEnabled(false);
```

EreMaijala:
@demiankatz It all makes sense now. @xrosecky I added one more comment now that it occurred to me that we probably don't need spellchecking or highlighting so those should be disabled as well. Highlighting of course doesn't do much when no records are returned, but it's still unnnecessary to even ask for it.

xrosecky:
Good catch, I disabled highlighting and spellcheck in the latest commit.

EreMaijala (approved):

demiankatz:
Thanks, @xrosecky and @EreMaijala, for getting this polished to such a nice state! I'm merging now.

PR #1851, opened 2021-02-12
Improved profiler support

EreMaijala:
See https://pecl.php.net/package/xhprof.


demiankatz:
Should we go back to detecting which function is available so we can support both XHProf and Tideways, in case this ends up switching back and forth again in the future?

EreMaijala:
@demiankatz First I thought there's no reason to have both, but of course if you already use Tideways, it's nicer to continue supporting it. Now done.

demiankatz:
Thanks. A couple more thoughts:

1.) Do you think we should use a more specific variable name than `$funcPrefix`? Maybe `$profilerFuncPrefix`?

2.) Would it make sense to move all the profiler code to a separate file? At present, this profiler logic takes up more than 25% of the index.php file. Would it make sense to move everything inside the `if (!empty($vufindProfiler)) {` block to a separate `profiler.php` file that gets `include`d only when needed? I think that might improve readability significantly, though maybe I'm getting carried away.

public/profiler.php
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @return void
+ */
+function enableProfiler($profilerBaseUrl)
+{
+    $profilerEnableFunc = false;
+    $profilerDisableFunc = false;
+    if (extension_loaded('xhprof')) {
+        $profilerEnableFunc = 'xhprof_enable';
+        $profilerDisableFunc = 'xhprof_enable';
```
demiankatz: I think you've got a typo here. :-)

EreMaijala: Whoa, thanks. I still got results when testing. :D

public/profiler.php
```diff
@@ -0,0 +1,83 @@
...
+                $dir = sys_get_temp_dir();
+            }
+            file_put_contents(
+                "$dir/$xhprofRunId.$suffix.xhprof",
+                serialize($xhprofData)
+            );
+            $url = "$profilerBaseUrl?run=$xhprofRunId&source=$suffix";
+
+            // Try to detect how the script was executed to add output appropriately
+            if ('cli' === php_sapi_name()) {
```
demiankatz: In other parts of the code, we use the `PHP_SAPI` constant for this. Is there a reason to prefer the function here (or everywhere)?

public/profiler.php
```diff
@@ -0,0 +1,83 @@
...
+                echo "Profiler output: $url";
```
demiankatz: Should we return after this line to avoid also displaying the HTML version?

public/profiler.php
EreMaijala: Fixed.

demiankatz (commented):
Thanks for working further on this -- it's looking good. Just a few minor notes. Also, a question: do you think we should put this in a subdirectory of $VUFIND_HOME to keep things tidy? I can't think of a particularly logical place for it, so I'm open to just leaving it here -- but if you have any other ideas, it might be worth thinking about before we make it official by merging this.

Regarding file location, if you want some totally brainstormed ideas: we could create a `module/VuFind/functions` directory for holding random support functions... or we could bundle this up in a class and store it in the usual way (though that might be misleading in some ways, since we have to include this before autoloading is activated).

public/profiler.php
EreMaijala: Should be fixed already, was wrong in the initial commit of profiler.php.

public/profiler.php
EreMaijala: No reason, I changed it.

public/profiler.php
demiankatz: Sorry, must have looked in the wrong place! :-)

EreMaijala:
@demiankatz I moved the file to the brainstormed directory since I don't have anything better to offer. Perhaps the other possibility would be to have functions directory outside the module directory, but I like it more this way as it doesn't pollute the top level.

demiankatz:
Yes, I think this looks good! Just one last question: should we perhaps rename `enableProfiler` to `enableVuFindProfiling` to eliminate any possibility of a collision with other code?

EreMaijala:
Good point. Done.

demiankatz:
Thanks again for all the work on this -- I just tested it out and it's working well for me, so I've gone ahead and merged it.

PR #1852, opened 2021-02-12
Primo: Fix url-encoding of filters in search query.

EreMaijala:
There was an attempt at encoding the values properly, but the result was never used. This caused e.g. subjects with ampersands to not be encoded properly in the query string.


demiankatz:
Thanks, good catch! I'll backport this to release-7.1 when I get back.

demiankatz:
Backported as promised.

PR #1853, opened 2021-02-15
Improved Solr handling in CI/test setup

demiankatz:


EreMaijala:
@demiankatz When was vufindtest.pid removed? Is there a possibility that Solr fails to start and the pid file isn't created or is deleted before the tasks complete? Perhaps the fake file isn't that bad, but could be renamed to something more indicative like 'vufind-ci-running'? Or maybe require that VUFIND_CI_RUNNING environment variable is set in ci-tasks target and set it before calling phing? In phing you should be able to check that with something like `<isfalse "${env.VUFIND_CI_RUNNING}"/>` (or isset??), right?

demiankatz:
@EreMaijala, it looks like the vufindtest.pid file stopped being real when we upgraded to Solr 5, which introduced significant changes to VuFind's Solr scripting. I think at that time I just created a fake file as a quick fix to keep the tests running, and then never circled back to clean it up properly.

The idea of the Phing tasks here is that you can run phpunit/phpunitfast/phpunitfaster at any time -- either to do a quick unit test (if Solr is not running) or to run the full integration suite (if Solr is running). In order to make that work, the startup task needs to leave something behind that the PHPUnit tasks can look for to figure out how to behave. An environment variable doesn't really work because if you run `phing startup` it can't permanently change the environment for when you subsequently run `phing phpunitfast`. Or, to put it another way, the environment variable approach could work well if we were only concerned with continuous integration and could chain the tasks together with a meta-task using a shared environment, but I don't think it's workable for the other use case: spinning up a test environment manually, and then running one or more tests one or more times -- which is a major part of how I do VuFind review and development work. (Of course, if you can think of a way to make this work that I'm overlooking, I'd be very interested in hearing it -- I just feel like most solutions would require me to manually set the environment variable or add another `-D` parameter, whereas this solution is fully automatic and thus more convenient).

Assuming you don't have a better suggestion for avoiding the file detection entirely, I do think there are at least two routes we could take from here:

1.) As you say, we could fully separate the CI detection from the Solr PID file by simply renaming the file that we `touch` and detect. This would be a very quick and easy fix, but I do rather like the elegance of looking for a file that other parts of the process need anyway rather than creating a fake marker file.

2.) As you point out, relying on the Solr PID file could create a situation where Solr fails to start, and then tests get skipped, and nobody notices because CI isn't failing. I think the solution to this is simply to make the Solr startup task more bulletproof than it currently is. Before starting Solr, it should check for an existing PID file (and either delete it, or fail -- I'm not sure which approach is better). After starting Solr, it should check for the PID file, and again fail if it does not exist. That way, if Solr fails to start, CI fails. If Solr starts but stops working, the tests will fail. I can't think of a scenario where Solr starts successfully and then deletes the PID file before tests start running -- I suppose that's still a risk, but I can't think of a reason why it would happen.

What do you think?

EreMaijala:
@demiankatz I was thinking of setting the environment variable outside of phing, but then it's not as bulletproof since you could forget it when running phpunitfast.

2.) Yes, bulletproofing the startup would alleviate my concerns. Then the only other case would be that Solr would somehow get shut down in the middle, but I think that can be considered similar to manually deleting the pid file.

demiankatz:
@EreMaijala, do you mind giving this another look? I've set it up to fail if you run `startup` after the system is already started, and to check that the `solr.sh` script returns a successful exit code, and to check if the pid file is missing after Solr finishes starting up. Hopefully that will catch the majority of possible problems.

EreMaijala (approved):
Looks good to me. My only concern is that if Solr port is not 8983, this falls apart. I don't think that's critical or easy to change, though, so I think this is quite acceptable. And other stuff will in any case depend on the port being correct as well.

demiankatz:
Yes, I agree -- it did cross my mind to make the port number behavior more flexible, but I decided not to since I don't expect it to change, and if we need to make adjustments in future, `8983` is a pretty distinctive string to search for within the code base. If we had to, it would likely be possible to make all of the code respect the `SOLR_PORT` environment variable... but it doesn't seem worth adding that complexity without a use case, and since all of this applies only to the self-contained test environment, I feel less pressure to be super-flexible. :-)

demiankatz:
Strangely, I just noticed that this PR has a typo in it -- the filename should be solr-8983.pid in TestCase.php, not vufind-8983.pid. I know that I fixed this error during development of this branch (otherwise, all of my testing would have failed), but it somehow never got committed and pushed up here (probably accidentally got reverted by a `phing shutdown` -- oops). I'll fix it post-merge in the dev branch. Sorry about that!

PR #1854, opened 2021-02-16
Add support for caching Solr queries.

EreMaijala:
TODO (at least): 

- [x] Decide whether just one config in config.ini would be better.

### DIFF
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -563,6 +563,22 @@ default_dismax_handler = dismax
 ; memory problems.
 ;record_batch_size = 100
 
+; This section provides settings for optional caching of search requests. Supported
+; for Solr only.
+[IndexCache]
+; Supported adapters: Memcached, Filesystem. Others may also work, but have not been
+; tested.
+;adapter = Memcached
+; Memcached servers (optional, uses localhost:11211 by default)
+;servers[] = "localhost:11211"
+;servers[] = "otherhost:11211"
+; Other options. See https://docs.laminas.dev/laminas-cache/storage/adapter/ for
+; documentation on options for different adapters.
+; ttl, Time to Live, i.e. caching time in seconds. 300 seconds by default.
+;options['ttl'] = 300
+; Cache directory for Filesystem cache:
+;options[cache_dir] = "/tmp/search-cache"
+
 ; Enable/Disable searching reserves using the "reserves" Solr core.  When enabling
 ; this feature, you need to run the util/index_reserves.php script to populate the
 ; new index.
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -563,22 +563,6 @@ default_dismax_handler = dismax
 ; memory problems.
 ;record_batch_size = 100
 
-; This section provides settings for optional caching of search requests. Supported
-; for Solr only.
-[IndexCache]
-; Supported adapters: Memcached, Filesystem. Others may also work, but have not been
-; tested.
-;adapter = Memcached
-; Memcached servers (optional, uses localhost:11211 by default)
-;servers[] = "localhost:11211"
-;servers[] = "otherhost:11211"
-; Other options. See https://docs.laminas.dev/laminas-cache/storage/adapter/ for
-; documentation on options for different adapters.
-; ttl, Time to Live, i.e. caching time in seconds. 300 seconds by default.
-;options['ttl'] = 300
-; Cache directory for Filesystem cache:
-;options[cache_dir] = "/tmp/search-cache"
-
 ; Enable/Disable searching reserves using the "reserves" Solr core.  When enabling
 ; this feature, you need to run the util/index_reserves.php script to populate the
 ; new index.
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -749,13 +749,12 @@ maxLimit = 100
 ; Supported adapters: Memcached, Filesystem. Others may also work, but have not been
 ; tested.
 ;adapter = Memcached
-; Memcached servers (optional, uses localhost:11211 by default)
-;memcached_servers[] = "localhost:11211"
-;memcached_servers[] = "otherhost:11211"
-; Other options. See https://docs.laminas.dev/laminas-cache/storage/adapter/ for
-; documentation on options for different adapters.
-; ttl, Time to Live, i.e. caching time in seconds. 300 seconds by default.
-;options['ttl'] = 300
+; Servers for Memcached (optional, uses localhost:11211 by default):
+;options[servers] = "localhost:11211,otherhost:11211"
 ; Cache directory for Filesystem cache:
 ;options[cache_dir] = "/tmp/search-cache"
+; Other options. See https://docs.laminas.dev/laminas-cache/storage/adapter/ for
+; documentation on options for different adapters.
+; ttl, Time to Live, i.e. cache entry life time in seconds. 300 seconds by default.
+;options[ttl] = 300
 
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -742,3 +742,20 @@ searchAccessPermission = access.api.Search
 
 ; This is the maximum number of results that can be returned in a single response:
 maxLimit = 100
+
+; This section provides settings for optional caching of search requests. Supported
+; for Solr only.
+[SearchCache]
+; Supported adapters: Memcached, Filesystem. Others may also work, but have not been
+; tested.
+;adapter = Memcached
+; Memcached servers (optional, uses localhost:11211 by default)
+;servers[] = "localhost:11211"
+;servers[] = "otherhost:11211"
+; Other options. See https://docs.laminas.dev/laminas-cache/storage/adapter/ for
+; documentation on options for different adapters.
+; ttl, Time to Live, i.e. caching time in seconds. 300 seconds by default.
+;options['ttl'] = 300
+; Cache directory for Filesystem cache:
+;options[cache_dir] = "/tmp/search-cache"
+
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -750,8 +750,8 @@ maxLimit = 100
 ; tested.
 ;adapter = Memcached
 ; Memcached servers (optional, uses localhost:11211 by default)
-;servers[] = "localhost:11211"
-;servers[] = "otherhost:11211"
+;memcached_servers[] = "localhost:11211"
+;memcached_servers[] = "otherhost:11211"
 ; Other options. See https://docs.laminas.dev/laminas-cache/storage/adapter/ for
 ; documentation on options for different adapters.
 ; ttl, Time to Live, i.e. caching time in seconds. 300 seconds by default.
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -743,8 +743,9 @@ searchAccessPermission = access.api.Search
 ; This is the maximum number of results that can be returned in a single response:
 maxLimit = 100
 
-; This section provides settings for optional caching of search requests. Supported
-; for Solr only.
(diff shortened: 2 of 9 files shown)


config/vufind/config.ini
```diff
@@ -563,6 +563,22 @@ default_dismax_handler = dismax
 ; memory problems.
 ;record_batch_size = 100
 
+; This section provides settings for optional caching of search requests. Supported
+; for Solr only.
+[IndexCache]
```
demiankatz: If we're doing one configuration per index (which I think makes sense, since you might want different caching behavior for different cores), would this make more sense in searches.ini rather than config.ini, since that's a place for Solr-specific settings?

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
```diff
@@ -380,6 +380,31 @@ protected function createConnector()
...
+                    $options['servers'][] = explode(':', $server, 2);
+                }
+            }
+            $settings = [
+                'adapter' => [
+                    'name' => $cacheConfig['adapter'],
+                    'options' => $options,
+                ]
+            ];
+            $cache = \Laminas\Cache\StorageFactory::factory($settings);
```
demiankatz, line 399: Did you consider leveraging the `\VuFind\Cache\Manager` here instead of directly hitting the `\Laminas\Cache\StorageFactory`? I'm not necessarily saying that would be a better solution -- honestly, I think VuFind's Cache Manager could use some work -- but I'm wondering if it makes sense to centralize the logic, and whether this might be an opportunity to make some improvements to caching more generally.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
```diff
@@ -46,8 +47,6 @@
 use VuFindSearch\Backend\Solr\Document\AbstractDocument;
 use VuFindSearch\ParamBag;
 
-use VuFindSearch\Query\Query;
```
demiankatz, line 49: I'm really puzzled why this `use` statement wasn't auto-removed by php-cs-fixer long ago. Am I correct in thinking that this was never needed, and you just cleaned it up as part of the other work here?

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
```diff
@@ -404,26 +423,60 @@ protected function forceToBackendException($ex)
...
+            }
             try {
-                return $this->send($client);
+                $result = $this->send($client);
+
+                if ($cacheKey) {
+                    try {
+                        $this->cache->setItem($cacheKey, $result);
+                    } catch (\Exception $ex) {
+                        if ($ex->getMessage() === 'ITEM TOO BIG') {
```
demiankatz, line 474: Is there a more reliable way to detect this condition? Is a more specific exception thrown, or is there a code that might be more stable over time than the textual message?

demiankatz (commented):
This looks like a great start! I left a few comments.

This also has me thinking about testing -- first of all, we should probably add some test methods to the relevant unit tests, since our VuFindSearch test coverage is high and we should keep it that way. Secondly, and more wildly, I'm wondering if we could use a pre-populated file cache to allow some of the integration tests to run without a real Solr instance (which might be useful for running tests in a resource-constrained environment). That's something to ponder for a while before taking action, but seemed worth mentioning.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
EreMaijala: Yes, at least it wasn't needed when I started working on the file. VSCode intelephense flagged it while php-cs-fixer didn't.

config/vufind/config.ini
EreMaijala: Done.

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
EreMaijala: I did, but since it's pretty much focused on caching configuration etc. in files, it didn't feel like it would make sense. Also, I believe it makes sense to keep configuration such as ttl separate.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
EreMaijala: Well, it's actually Laminas\Cache\Exception\RuntimeException, but it's still just the string from Memcached object's getResultMessage(). Memcached has a getResultCode method as well, but Laminas' adapter doesn't use it.

EreMaijala:
@demiankatz The prepopulated cache is an intriguing idea, but I'm afraid it would be terrible for maintenance, since the cache key is md5. Unless you mock the cache key creation, but then youo could just mock the send method.

I'll think about the tests and try to add some, at least.

EreMaijala:
@demiankatz I've added a test that mocks the cache storage and verifies that it's used properly. Is there anything else that should be added?

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
demiankatz: That's frustrating, but it sounds like there's not much we can do about it. Maybe the best solution for now is simply to add a comment explaining why this magic string is here, so we can revisit in the future if things change, and so people will better understand why this is here.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
demiankatz: Thanks for confirming!

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
demiankatz: Makes sense. I think we should revisit the whole thing in future, but it seems most likely that we'll eliminate/simplify the cache manager in future, if we change anything at all, so I think it does make sense to keep this separate for now.

config/vufind/config.ini
demiankatz: Thanks!

demiankatz:
@EreMaijala, good point regarding the pre-populated cache idea. I had a feeling it was too crazy to work, but it seemed worth mentioning. Oh well!

Thanks for the test progress... I'll take a look now and comment directly on the code if necessary.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/ConnectorTest.php
```diff
@@ -140,6 +140,41 @@ public function testSaveThrowsUnknownFormat()
...
+    public function testCaching()
+    {
+        $conn = $this->createConnector('single-record');
+
+        list(, $expectedBody) = explode("\n\n", $this->response);
+
+        $cache = $this->createMock(\Laminas\Cache\Storage\StorageInterface::class);
+        $cache->expects($this->exactly(3))
+            ->method('getItem')
+            ->willReturnOnConsecutiveCalls(null, $expectedBody, 'foo');
```
demiankatz, line 159: You could add a `with` here to indicate that all three calls have the same input parameter; that might help clarify things slightly.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/ConnectorTest.php
```diff
@@ -140,6 +140,41 @@ public function testSaveThrowsUnknownFormat()
...
+        $conn->setCache($cache);
+
+        $resp = $conn->retrieve('id');
+        $this->assertEquals($expectedBody, $resp);
+        $resp = $conn->retrieve('id');
+        $this->assertEquals($expectedBody, $resp);
+        $resp = $conn->retrieve('id');
+        $this->assertEquals('foo', $resp);
+
+        // Make sure that write() doesn't access the cache
```
demiankatz: Do we need any more assertions in support of this? Sorry if I'm missing something!

demiankatz (commented):
The test looks great -- just a couple of small comments. Once you've addressed those, is there anything else pending? If not, I'll run the full test suite on my end to be sure all is still well, and then we can merge if you're ready.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
EreMaijala: I added a comment and revised the exception handling a bit in anticipation of codes being available at some point. There's the additional complication that any error codes are not translated, so any checks should be adapter-specific. Nevertheless, I posted https://github.com/laminas/laminas-cache-storage-adapter-memcached/pull/6 to add error codes to Memcached exceptions since other adapters already include them.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/ConnectorTest.php
EreMaijala: `with` added. It's still just a string constraint since I think hard-coding the actual key just makes it more complicated to maintain without obvious benefit.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/ConnectorTest.php
EreMaijala: The expectations for number of calls to getItem/setItem will fail if write() calls them. I added a comment about that. If you think it's better to make it more explicit, I can still revise.

EreMaijala: I decided to revise the test to be more explicit. Now it should be easier to read.

EreMaijala:
@demiankatz I think I've addressed all the concerns, and there should be no pending work.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
demiankatz: Thanks! I see that they're asking for test coverage on your Laminas PR, which is a good sign. :-)

config/vufind/searches.ini
```diff
@@ -742,3 +742,20 @@ searchAccessPermission = access.api.Search
...
 maxLimit = 100
+
+; This section provides settings for optional caching of search requests. Supported
+; for Solr only.
+[SearchCache]
+; Supported adapters: Memcached, Filesystem. Others may also work, but have not been
+; tested.
+;adapter = Memcached
+; Memcached servers (optional, uses localhost:11211 by default)
+;servers[] = "localhost:11211"
```
demiankatz: Sorry for the last-minute nitpick, but would it make sense to call this memcached_servers to reduce potential ambiguity? Or can this be addressed directly through the options setting below? (I'm guessing not due to the amount of nesting, but I'd love to be wrong about that). I realize that we agreed earlier that \VuFind\Cache\Manager doesn't have a role here, but I do wonder if it would make sense to have a shared method that translates a VuFind Config object into a cache object in a consistent way, so the settings in the [Cache] section of config.ini could match up with the new settings in [SearchCache] to a greater extent. (I'm not necessarily asking you to do that now -- just brainstorming... but if nothing else, maybe we should open a JIRA ticket about cache consolidation/reorganization).

EreMaijala: So, you _were_ just as wrong as I was, and the configuration has now been simplified. :)

I added https://vufind.org/jira/browse/VUFIND-1467. Feel free to modify as necessary.

demiankatz: Ahh, fantastic, thanks on both counts!

PR #1856, opened 2021-02-18
Fix ISO 2709 MARC reader for improved PHP 8 compatibility.

demiankatz:


demiankatz:
@EreMaijala, this is a pretty trivial change, but it's necessary for PHP 8 compatibility. I think it's uncontroversial, but I'm leaving it for your review just in case I'm failing to consider anything. :-)

EreMaijala (approved):

PR #1858, opened 2021-02-23
Handle repeating IDs in a record loader batch.

demiankatz:
There are some circumstances where the same record may be retrieved multiple times in a batch (for example, circulation transactions where a user has multiple copies/volumes of the same title checked out). Previously, the batch loader would only retrieve the correct record driver once, even if it was requested multiple times. This PR allows the record to be displayed correctly more than once in a list.

TODO
- [x] Decide whether or not to include a deprecated SourceAndIdList::getRecordPosition method for backward compatibility
- [x] Create a unit test for SourceAndIdList


demiankatz:
If anyone wants an easy way to test this, I was able to do so by spinning up the standard test environment, switching to the Demo ILS driver, and changing the `query` setting in `Demo.ini` to `"id:testbug2"` to force all sample records retrieved in the MyResearch area to display the same test record.

EreMaijala (approved):
Looks good to me (and tested also)!

demiankatz:
Thanks, @EreMaijala! I'm merging now -- I've gone ahead and built a unit test for the class. I also decided not to add a deprecated method for backward compatibility, since this is literally only used in one place, and that is code that it seems unlikely anyone would change or override. I have put a note in the changelog, though, just in case!

PR #1860, opened 2021-02-25
Use correct class names as service names (instead of aliases/strings)

demiankatz:


demiankatz:
Psalm turned up a few spots where we're using aliases instead of real class names. This PR fixes them. Since two of the three affected files are in places important to you, @EreMaijala, I'm hoping you can just double-check and make sure I'm not making any mistakes here. Pretty sure this is a safe and trivial fix, though!

EreMaijala (approved):

PR #1861, opened 2021-02-26
Change Vagrant default user to match updated box

aleksip:
This might be one of those problems due to the specific combination of macOS 10.15.7 / VirtualBox 6.1 / Vagrant 2.2.14 / ubuntu/focal64 I am using, but it looks like the recently updated Vagrant box has a different default user (`vagrant`).

This seems to cause some file and directory permission related problems. The `/vagrant/solr/vufind/logs` directory can not be created, and everything in the `/vagrant` directory is owned by `ubuntu:vagrant` and in `/vufindlocal` by `ubuntu:ubuntu`. So changing the user in `Vagrantfile` seemed to be the easiest way to get everything working.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -27,7 +27,7 @@ Vagrant.configure("2") do |config|
   # Network configuration to forward ports.
   config.vm.network :forwarded_port, guest: 80, host: 4567
   config.vm.network :forwarded_port, guest: 8983, host: 4568
-  config.vm.synced_folder ".", "/vagrant", :owner => 'ubuntu'
+  config.vm.synced_folder ".", "/vagrant", :owner => 'vagrant'
 
   # Enable provisioning with a shell script. Additional provisioners such as
   # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks, @aleksip! The last time I tested Vagrant, I ran into some issues but was eventually able to get things working. I was short on time and attributed it to user error, but it was probably this problem. I confirmed that I can get Vagrant up smoothly with these changes on my own test server.

There does appear to be one other issue, which may or may not be new: if you don't run `composer install` prior to `vagrant up`, the dependencies don't install correctly. It appears to be due to some weird issue with the VirtualBox shared folder functionality, and it may be beyond our control to fix. For now, I just updated the documentation to recommend installing Composer dependencies prior to running `vagrant up`, but it might be worth investigating this further in the future.

aleksip:
@demiankatz I came across a `composer install` issue as well. I left the solution I used out of the PR because I was not completely sure whether the issue only occurs with a specific combination of OS and software versions.

It was an error related to `ocramius/package-versions` and possibly due to a bug in VirtualBox. I found a workaround [here](https://github.com/laravel/homestead/issues/1240#issuecomment-555951201), but it did make the installation very slow, so installing dependencies prior to `vagrant up` does seem a better (hopefully temporary) solution.

demiankatz:
Thanks, @aleksip, that's good to know, and as you say, hopefully the problem will sort itself out over time.

If I can ask a favor, would you be willing to test out PR #1862 with your Vagrant environment? I'm working on the transition to Composer 2, and it would be helpful to have more results to compare.

PR #1862, opened 2021-02-26
[VUFIND-1454] Update VuFind to use Composer 2.

demiankatz:
TODO
- [x] Run tests
- [x] Test Vagrant
- [x] Close [VUFIND-1454](https://vufind.org/jira/browse/VUFIND-1454) when merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ env:
 before_script:
   - phpenv config-rm xdebug.ini
   - phpenv rehash
-  - composer selfupdate --1
   - composer install
   - npm install -g eslint@"<5.0.0"
   - npm install -g jshint@"2.9.6"
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -50,7 +50,6 @@ Vagrant.configure("2") do |config|
     php -r "if (hash_file('SHA384', 'composer-setup.php') === trim(file_get_contents('https://composer.github.io/installer.sig'))) { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
     php composer-setup.php
     php -r "unlink('composer-setup.php');"
-    php composer.phar self-update --1
     mv composer.phar /usr/local/bin/composer
 
     # Check out and set up VuFind.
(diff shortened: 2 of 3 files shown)


demiankatz:
I was able to get Vagrant working on my Windows box, and this seems to work fine there!

aleksip:
@demiankatz `composer install` on the guest fails due to the known VirtualBox bug (issue opened 10 years ago!), but I found another workaround [here](https://github.com/composer/composer/issues/9627#issuecomment-780471799) which might be worth trying out.

Unfortunately `composer install` also fails on my host at the moment, because the Solr download is very very slow and keeps timing out. I'm wondering if that could be made optional, since many folks might use a remote Solr backend, like we do. Or is the download required in any case?

aleksip:
@demiankatz Working around the Solr download problem everything else seems to work fine on my Vagrant setup!

demiankatz:
@aleksip, thanks for the confirmation and workaround information. That might be worth pursuing when time permits. Regarding the Solr download, there's definitely no need to download Solr if you're using a remote backend, so I'd be open to making it optional if you have any thoughts on the best way to accomplish that. Ideally it would be nice to install it by default (since that's what most people will be expecting) but to have a switch or option to bypass it. I'm not sure if Composer supports such a thing, though; I haven't gone too far into the advanced capabilities.

In the meantime, if you need a simple workaround for Solr installation, you could always create a dummy empty .tar.gz file in the downloads directory, which should short-circuit the whole installation process. (Not an elegant solution, but a functional one).

EreMaijala:
How about adding a flag file to prevent installsolr from actually doing anything? Something like `solr/disableAutomaticInstall`? This could be set for Travis, and we could also set it in our local repository.

demiankatz:
@EreMaijala, that's a good idea! Would you or @aleksip be interested in opening a PR with that functionality? If not, I can throw it into my queue -- just want to know whether or not I need to remember to do it. :-)

EreMaijala:
@demiankatz I'll post a PR since I already experimented with it a bit.

demiankatz:
Thanks, @EreMaijala!

EreMaijala:
#1869 opened.

PR #1863, opened 2021-02-26
Support richer, more readable Javascript configuration

demiankatz:
This PR extracts some functionality from #1799 in an effort to get the configuration improvements merged separately from other additional features.

TODO
- [x] Run full test suite
- [x] Update and resume work on #1799 after this is merged to dev
- [x] Update changelog to note refactoring of parseSetting() method


demiankatz:
All tests are passing, so I think I've successfully extracted the appropriate code from #1799. @EreMaijala and @mtrojan-ub, can you give this a look and let me know if you have any concerns or if you think it's ready to merge?

Also note that @mtrojan-ub is not currently a member of the vufind-org Developers group, but I think he should be so that he can be assigned as a reviewer, etc. I have sent an invitation!

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -129,12 +129,130 @@ public function addCss($css)
...
     {
-        if (!is_array($js) && !is_a($js, 'Traversable')) {
-            $js = [$js];
+        if ((!is_array($js) && !is_a($js, 'Traversable')) || isset($js['file'])) {
+            $this->addJsEntry($js);
+        } elseif (isset($js[0])) {
+            foreach ($js as $current) {
+                $this->addJsEntry($current);
+            }
+        } elseif ($js == []) {
```
EreMaijala: Is this intentionally not `===`?

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -129,12 +129,130 @@ public function addCss($css)
...
+                'Using "priority" as well as "load_after" in the same entry '
+                . 'is not supported: "' . $jsEntry['file'] . '"'
+            );
+        }
+
+        foreach ($this->js as $existingEntry) {
+            if ($existingEntry['file'] == $jsEntry['file']) {
+
+                // If we have the same settings as before, just skip this entry.
+                if (count(array_diff($existingEntry, $jsEntry)) > 0) {
```
EreMaijala: Is this doing the right thing? I may be missing something, but for the settings to be same array_diff would return an empty array, no? Can we be sure both entries have all the same keys? Why `count` when `!empty` would do?

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -129,12 +129,130 @@ public function addCss($css)
...
+                if (isset($entry['priority'])) {
+                    $currentPriority = $array[$i]['priority'] ?? null;
+                    if (!isset($currentPriority)
+                        || $currentPriority > $entry['priority']
+                    ) {
+                        array_splice($array, $i, 0, [$entry]);
+                        return;
+                    }
+                } elseif (isset($entry['load_after'])) {
+                    if ($entry['load_after'] == $array[$i]['file']) {
```
EreMaijala: Could use `===`?

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -164,7 +282,28 @@ public function getCss()
...
+     * Given a colon-delimited configuration string, break it apart, making sure
+     * that URLs in the first position are not inappropriately split.
+     *
+     * @param string $current Setting to parse
+     *
+     * @return array
+     */
+    public function parseSetting($current)
+    {
+        $parts = explode(':', $current);
```
EreMaijala, line 300: Could we just say that you need to use the array format in theme config if you have a protocol and condition? That would be a minor inconvenience during setup but would avoid doing this check every time.

EreMaijala (changes requested):

demiankatz:
Thanks, @EreMaijala, for the feedback. I'm hoping @mtrojan-ub can answer your questions; I've added him as a reviewer as well so he'll be notified. :-)

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: addJs also accepts strings as parameter; however this should be covered on line 132, so i guess changing it to === shouldn't be a problem to make it clearer.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: You're right! array_diff seems to be a problem if the values are the same but the keys aren't!

After some more tests, i propose we simply use `==` instead to compare the arrays. I would not recommend using `===` because that would only return true if the keys are in the same order.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: Right, strict comparison can be used here.

mtrojan-ub:
Thanks for inviting me to the vufind-org Developers group, i feel truly honored 😃

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: That sounds like the right thing to do, then we could get rid of all the string logic... i just thought that @demiankatz normally likes it if things stay backwards compatible; however we could simplify the use cases and leave a note in the changelog so people know they should change their old entries when upgrading.

demiankatz:
Of course, @mtrojan-ub. Hopefully nobody will be mad at me for adding you before the Project Management Committee is fully up and running, but you've clearly demonstrated that you understand the code base well, so I'm not worried. ;-)

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
demiankatz: We're still going to need parsing for other types of resources (e.g. CSS) that have not been converted to the array format yet. Maybe for now we just add a comment like:

<pre>
// TODO: replace this method with a deprecation warning when all configs have been converted to arrays
</pre>

...and then we can phase it out gradually?

mtrojan-ub: Sounds fine for me

demiankatz:
@EreMaijala / @mtrojan-ub, I think I've made all the changes discussed in the review above as part of commit 8ccbca2 (I also removed an unnecessary else to quiet a warning in my IDE). I don't have time to test any of this today -- but if you review that commit and it looks good to you, I'll run the full test suite in the near future, and merge if you're both happy and all tests pass.

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
mtrojan-ub: Seems like the current Travis test seems to have a problem with it now. This could be caused by the condition in 132 not matching because strings seem to be Traversable... so either this should be changed back to `==` or the Traversable condition needs to be changed (or we explicitly need to check for strings).

demiankatz: No, the problem is that my IDE gave me bad advice, and I should have left that else clause alone! I've just put it back the way it was, and hopefully the tests will pass again. Sorry about that -- dumb mistake on my part. :-)

demiankatz:
Okay, we're back to having the full test suite passing again. Any final thoughts before I merge, @mtrojan-ub or @EreMaijala?

mtrojan-ub:
No objections from my side => i would be glad if we could merge this and make a step forward on this issue!

EreMaijala:
@demiankatz No objections from me either.

demiankatz:
I noticed that the comments mentioned a load_before setting that is no longer supported, so I updated the documentation (with some other minor tweaks) in e103ddbd0e85f2ec08b58e311d0c110ec1c87a71.

demiankatz:
My comment changes didn't break the Travis build, so I'm going ahead and merging now. Thanks, everyone! I'll update #1799 in the near future as well.

PR #1864, opened 2021-03-01
Folio: Fix parameters of getNewItems method.

EreMaijala:
This fixes "Required parameter $limit follows optional parameter $page" error with PHP 8.

With this fix, all tests pass with PHP 8.0.2.


demiankatz:
Thanks, @EreMaijala! I have no idea why those parameters were there in the first place. They've been there since the first commit, but I can't understand what their purpose could ever have been. I'm backporting this to the release-7.1 branch so the fix can be included in release 7.1.1.

demiankatz:
(As a side note, I'm also not sure why I didn't encounter a test failure while I was testing PHP 8 myself... maybe there's a change in 8.0.2 from the version I was testing with).

PR #1866, opened 2021-03-01
Realign collection/view.phtml with record/view.phtml.

demiankatz:
The record/view.phtml and collection/view.phtml templates have gotten significantly out of sync with fixes and new features. This PR attempts to realign them.

TODO
- [x] Run full test suite
- [x] Identify whether there are any potential problems missed by the test suite
- [x] Discuss whether refactoring could make this maintenance easier in future (though refactoring should not be applied to this PR, since this is targeted as a bug fix to the release-7.1 branch, and we don't want to introduce BC breaks there)

### DIFF
--- a/themes/bootstrap3/templates/record/view.phtml
+++ b/themes/bootstrap3/templates/record/view.phtml
@@ -32,7 +32,7 @@
   <?php $sidebarList = $this->related()->getList($this->driver); ?>
   <div class="<?=$this->layoutClass('mainbody')?><?=count($sidebarList) < 1 ? ' solo' : '' ?>">
     <input type="hidden" value="<?=$this->escapeHtmlAttr($this->driver->getUniqueId())?>" class="hiddenId" />
-    <input type="hidden" value="<?=$this->escapeHtmlAttr($this->driver->getSourceIdentifier()) ?>" class="hiddenSource" />
+    <input type="hidden" value="<?=$this->escapeHtmlAttr($this->driver->getSourceIdentifier())?>" class="hiddenSource" />
     <?=$this->flashmessages()?>
     <?=$this->record($this->driver)->getCoreMetadata()?>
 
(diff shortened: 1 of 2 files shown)


demiankatz:
I opened this PR after @EreMaijala observed that these templates had gotten out of sync. The full test suite is passing with these changes in place. I'm curious whether @crhallberg has any thoughts/concerns to add.

EreMaijala (approved):

themes/bootstrap3/templates/record/view.phtml
```diff
@@ -32,7 +32,7 @@
   <?php $sidebarList = $this->related()->getList($this->driver); ?>
   <div class="<?=$this->layoutClass('mainbody')?><?=count($sidebarList) < 1 ? ' solo' : '' ?>">
     <input type="hidden" value="<?=$this->escapeHtmlAttr($this->driver->getUniqueId())?>" class="hiddenId" />
-    <input type="hidden" value="<?=$this->escapeHtmlAttr($this->driver->getSourceIdentifier()) ?>" class="hiddenSource" />
+    <input type="hidden" value="<?=$this->escapeHtmlAttr($this->driver->getSourceIdentifier())?>" class="hiddenSource" />
```
crhallberg, line 35: If no space is the style, I apologize for recent inconsistencies!

crhallberg (approved):

themes/bootstrap3/templates/record/view.phtml
demiankatz: I don't think we have a formal style rule, but I think "no space" is more common. In any case, I made this change simply to make the two view.phtml files consistent with one another, and the space seemed to be the less-used variation, so I removed it. It would be nice if php-cs-fixer could enforce this, because otherwise it's impossible to remember to make it perfectly uniform all the time. I haven't checked to see if there's a rule that applies (no time at the moment) but it might be worth doing.

demiankatz:
I'm merging this now, though there is still a need to discuss unifying the interface. I've opened [VUFIND-1470](https://vufind.org/jira/browse/VUFIND-1470) to track that conversation going forward.

PR #1867, opened 2021-03-02
Catch also RuntimeException when sending email in scheduled searches.

EreMaijala:


EreMaijala:
Turns out, Laminas Mail throws a \RuntimeException derivative on connection issues, and it's not a descendant of \Exception..

EreMaijala:
Actually, it should be a descendant of \Exception. There must be something else going on that I'm missing.

PR #1868, opened 2021-03-02
Disconnect mailer when done processing scheduled searches.

EreMaijala:
This avoids a potential exception from being thrown during destruction if the connection has been idle for too long.

This is the real issue I was mistakenly trying to fix in #1867.

### DIFF
--- a/module/VuFind/src/VuFind/Mailer/Mailer.php
+++ b/module/VuFind/src/VuFind/Mailer/Mailer.php
@@ -58,6 +58,13 @@ class Mailer implements \VuFind\I18n\Translator\TranslatorAwareInterface
     protected $transport;
 
     /**
+     * A clone of transport that can be used for a new connection
+     *
+     * @var TransportInterface
+     */
+    protected $initialTransport;
+
+    /**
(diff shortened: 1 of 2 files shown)


EreMaijala:
There was still a potential issue of Mailer's resetConnection throwing an exception when the SMTP transport tries to send a QUIT command. The only ways to avoid this, as far as I can see, is to use a new connection. I used the clone approach to keep setConnection method working and to avoid having to introduce a connection factory class to Mailer.

EreMaijala:
I think there might still be a case where the old SMTP connection would throw during garbage collection, but I can't see a way around it other than always specifying a connection time limit, which would disable the use of QUIT message in the SMTP connection layer.

demiankatz:
@EreMaijala, this makes sense to me as a solution, but I think it's probably worth adding more comments in the code to explain why we are doing this, because without the context provided by the description of this PR, it may be somewhat mystifying. I think we owe our future selves a little more explanation. ;-) If you like, I can propose some comments and you can tell me if they are correct or not, if you think that would be a useful approach.

demiankatz:
Much clearer! Thanks again!

PR #1869, opened 2021-03-02
Make Solr installation optional. Disable it in Travis.

EreMaijala:


demiankatz:
Thanks, @EreMaijala. I just adjusted a comment in build.xml to match the new logic. Do you think it is also worth removing downloads from the cache, since we don't need them if we don't download Solr?

EreMaijala:
@demiankatz Good point, downloads directory now removed from cache.

demiankatz:
Thanks!

PR #1870, opened 2021-03-02
Add cat_username to getMyProfile method of Aleph driver

xmorave2:


demiankatz:
Makes sense to me -- thanks!

PR #1872, opened 2021-03-05
Catch the exception of unsuccessful hold cancel

welblaud:
There was a logic duplication in the `cancelHolds` method and `doRestDLFRequest`. Inside the `doRestDLFRequest`, if the request fails and the response code is `0000` then `AlephRestfulException` is thrown. This means the repeated check of the response code in `cancelHolds` is never triggered. This way the exception should be gracefully solved.

I made it as small as possible because editing the `doRestDLFRequest` directly would affect a lot of things and because errors of this type are typically thrown during holds cancellation.

Note: Aleph uses a special daemon for harvesting pending requests. For a good user experience, it has a really short delay. After the harvest, it is quite possible a user has the hold list still enriched with the `cancelForm`. If she submits it after the harvest, the error is thrown because the request became non-cancellable in the meantime.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Aleph.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Aleph.php
@@ -1199,8 +1199,10 @@ class Aleph extends AbstractBase implements \Laminas\Log\LoggerAwareInterface,
                     ], null, "DELETE"
                 );
             } catch (AlephRestfulException $e) {
-                $statuses[$id]
-                    = ['success' => false, 'status' => 'cancel_hold_failed'];
+                $statuses[$id] = [
+                    'success' => false, 'status' => 'cancel_hold_failed',
+                    'sysMessage' => $e->getMessage(),
+                ];
             }
             if (isset($result)) {
                 $count++;
--- a/module/VuFind/src/VuFind/ILS/Driver/Aleph.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Aleph.php
@@ -1194,8 +1194,8 @@ class Aleph extends AbstractBase implements \Laminas\Log\LoggerAwareInterface,
             try {
                 $result = $this->doRestDLFRequest(
                     [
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -1191,23 +1191,18 @@ public function cancelHolds($details)
...
+            try {
+                $result = $this->doRestDLFRequest(
+                    [
+                        'patron', $patronId, 'circulationActions', 'requests', 'holds',
+                        $id
+                    ], null, "DELETE"
+                );
+            } catch (AlephRestfulException $e) {
+                $statuses[$id]
+                    = ['success' => false, 'status' => 'cancel_hold_failed'];
```
demiankatz: In the original version of the code, we returned the system message in the 'sysMessage' element of the return array, but that seems to have been dropped in the refactoring here. I believe you could bring it back with `'sysMessage' => $e->getMessage()` if you wanted it. Was it removed intentionally?

janhejzl: Yes, you are right. At first, it was intentional. Now I see it still works with retrieving the message directly from the exception.

demiankatz:
@xmorave2, as another Aleph user, do you mind giving  this a quick review before I merge it?

demiankatz:
@welblaud, I noticed the style checker was complaining about a long line, so I rewrapped it with https://github.com/vufind-org/vufind/pull/1872/commits/ed4348ddc256966ee686f2e74b75716250cf19fb

xmorave2:
> @xmorave2, as another Aleph user, do you mind giving this a quick review before I merge it?

I'd better ask @xrosecky, he know Aleph and its VuFind driver much more than me...

xrosecky:
Yes, it is OK.

demiankatz:
Thanks for the review, @xrosecky!

PR #1874, opened 2021-03-08
Allow translated template rendering via view helper

mtrojan-ub:
As discussed in one of the last community calls, this might be useful if you need to include a translated template in e.g. a content page.

Do you think the functionality is implemented the right way? (Not all tests work yet, i will fix the remaining tests as soon as no other changes are requested.)

TODO
- [x] Improve test coverage
- [x] Note backward compatibility breaks (e.g. changes to ContentBlock) in changelog when merging


demiankatz:
Thanks, @mtrojan-ub! I've taken a first look at this and have a few thoughts:

1.) I don't think this functionality belongs in the Context helper -- honestly, I think that helper, and VuFind's use of renderInContext(), is a bit confusing, and we may be able to eliminate it in the future in favor of core framework functionality, so I'd prefer not to add additional behavior to it. My memory of this is all a bit hazy, though, so maybe it serves more of a purpose than I remember, so feel free to disagree! I do wonder, though, if this actually merits its own separate translateTemplate() helper. It might also be worth looking at the existing HelpText helper (https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/View/Helper/Root/HelpText.php) since that offers similar functionality. Maybe there is an opportunity to unify these things.

2.) Are you only adding the $pageDetails['theme'] value to support the regular expression you're using to strip the leading path off of the template name? If so, I wonder if there's a simpler way to do that... Rather than preg_replacing, it might be better to preg_match a value, using something like `|.*/themes/[^/]+/templates/(.*)$|`. Or would it be better to directly `include` the path instead of using a renderer at all, since otherwise you're triggering duplicate template searches? Again, I haven't really had time to think this through deeply, but just trying to raise ideas/possible issues based on my first look.

Please let me know what you think, and I'll be happy to look more deeply into this and give it more thought when you're ready (and time permits).

demiankatz:
@mtrojan-ub, I've also edited the name of this PR to make it a little more generic, so it remains accurate even if the approach changes.

EreMaijala:
What we've done locally is to follow the Laminas' idea of `$template = $this->resolver('something'); if ($template) { $this->render($template);` with a `findTemplateForLng` method.

See an example of how this is used: 
https://github.com/NatLibFi/NDL-VuFind2/blob/dev/themes/finna2/templates/myresearch/login.phtml#L21

And the implementation:
https://github.com/NatLibFi/NDL-VuFind2/blob/dev/module/Finna/src/Finna/View/Helper/Root/Content.php#L87

While this two-step approach is more complex than a single rendering call, it allows one to act based on whether a template is available (like in the example above, where the result is enclosed in a div). This of course isn't very meaningful when you control the whole thing, but it allows us to define locations where an instition can add their own block with specific instructions etc. by creating a specific template.

demiankatz:
Thanks for sharing, @EreMaijala. I like your approach, though in order to adapt it for mainline VuFind, I think we would have to figure out how to reconcile it with the VuFind\Content\PageLocator, since we want to avoid duplicate logic and also take advantage of MarkDown support. I think this could be accomplished by making the Content view helper a thin wrapper around the PageLocator, with one method to run the template search and another method to render a template based on the return value. This would still allow the two-step approach, it would just force us to access the native render method indirectly because of the MarkDown logic. Also, on a minor side note, I think we should work toward eliminating methods that return a value or false and change them to return a value or null. This will make it easier to manage types within the code as we begin to get more strict about that.

EreMaijala:
@demiankatz Agreed, I didn't mean to imply this would work out of the box.

demiankatz:
Understood! But hopefully we can come up with something useful that you can take advantage of to reduce local custom code.

mtrojan-ub:
Thanks as well for sharing!

As far as i can see, there are 2 problems re-using the HelpText helper:

1) The helper is strictly bound to the HelpTranslations base directory.
In the past, we tried to solve this already with some custom code in which the helper takes additional parameters:
https://github.com/ubtue/tuefind/blob/master/module/TueFind/src/TueFind/View/Helper/Root/HelpText.php#L45

2) The path scheme `<baseDir>/<language>/<template>` is fundamentally different to the PageLocator `<basePath>/<template>_<language>`.

=> So would it be ok to unify these schemes and deprecate the custom HelpText scheme? I don't think it would make sense if the HelpText helper would support the other scheme as well. So either we change the scheme and use the PageLocator inside the HelpText helper, or we create a separate View Helper using the PageLocator and the HelpText Helper stays as-is.

What do you think?

demiankatz:
I agree, @mtrojan-ub, the HelpText helper is too specialized... but if we can create a more general tool, then HelpText might become a subclass of that or even go away entirely. I'm open to backward-breaking changes in the interest of improving consistency (and maybe we can come up with a way to minimize potential disruption).

mtrojan-ub:
OK, so now i moved the logic out of the Context view helper into a new Content view helper. Also, the existing HelpText view helper is now using the content view helper to search for the correct template. The only thing that HelpText still does by itself is to return warnings if the original language has not been found.

Please let me know what you think!

Also, the HelpTranslations template files have not been moved+renamed yet; i will gladly do this as soon as you agree to the code changes.

demiankatz:
Thanks, @mtrojan-ub, I haven't had a chance to look at this in exhaustive detail yet, but I think it's moving in the right direction! It occurs to me that you might not have to move the HelpTranslations around if you create a mechanism for specifying the template patterns to the PageLocator. Did you consider that possibility? Honestly, moving the files around for consistency is probably a better long-term approach, but overriding the series of file patterns used for lookup might be less work and allow greater backward compatibility. What do you think?

mtrojan-ub:
Great idea! So i added a pattern mechanism which is now also used to apply a different pattern when the HelpText view helper is used. Please have another look!

demiankatz:
Thanks, @mtrojan-ub, this looks like good progress. I'll review in more detail as soon as time permits (probably next week).

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,87 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
      *
      * @return array|null Null if template is not found or array with keys renderer
      * (type of template), path (full path of template), page (page name)
      */
-    public function determineTemplateAndRenderer($pathPrefix, $pageName)
-    {
+    public function determineTemplateAndRenderer(
+        $pathPrefix, $pageName, $pattern=null
+    ) {
+        if (!isset($pattern)) {
```
demiankatz: There is now a lot of logic related to creating the `$templates` array. It might make sense to refactor this to a separate support method for easier extension/overriding. Perhaps something like `protected function getTemplateOptionsFromPattern($pathPrefix, $pageName, $pattern)`.

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,87 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
         // 2.) Default language suffix
         // 3.) No language suffix
         $templates = [
-            "{$pageName}_$this->language",
-            "{$pageName}_$this->defaultLanguage",
-            $pageName,
+            'language' => $this->generateTemplateFromPattern(
+                $pathPrefix, $pageName, $pattern, $this->language
+            ),
+            'defaultLanguage' => $this->generateTemplateFromPattern(
```
demiankatz: Should we skip the defaultLanguage option when $this->language == $this->defaultLanguage, to prevent double-checking the same file? (I realize this was an existing issue... just thinking about it while we're changing things anyway).

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,87 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
+                );
+                if (null != $pathDetails) {
                     return [
                         'renderer' => $type,
-                        'path' => $path,
-                        'page' => $template,
+                        'path' => $pathDetails['path'],
+                        'page' => basename($template),
+                        'theme' => $pathDetails['theme'],
+                        'type' => $resultType,
```
demiankatz: There's some potential confusion here, since we have a variable called `$type` which is used to set a key called `renderer` and a key called `type` which receives a value called `$resultType`. It might be worth renaming some variables to make this more consistent, to improve readability.

module/VuFind/src/VuFind/View/Helper/Root/Content.php
```diff
@@ -0,0 +1,100 @@
...
+    ) {
+        if (!str_ends_with($pathPrefix, '/')) {
+            $pathPrefix .= '/';
+        }
+        $pathPrefix = 'templates/' . $pathPrefix;
+        $pageDetails = $this->pageLocator->determineTemplateAndRenderer(
+            $pathPrefix, $pageName, $pattern
+        );
+        $path = preg_replace(
+            '"^.+' . $pageDetails['theme'] . '/templates/"', '',
```
demiankatz: I think my original comment got lost in the refactoring, but I'm still concerned that this may not be entirely reliable, if VuFind is installed in certain weird directory paths. Would it make sense to replace `^.+` with the `APPLICATION_PATH` constant and do a straightforward `str_replace` instead of `preg_replace`? That might serve the same purpose but be much more exact.

module/VuFind/src/VuFind/View/Helper/Root/HelpText.php
```diff
@@ -100,39 +82,29 @@ public function getWarnings()
...
+    public function render($name, $context=[])
     {
-        // Set up the needed context in the view:
-        $this->contextHelper->__invoke($this->getView());
-        $oldContext = $this->contextHelper
-            ->apply(null === $context ? [] : $context);
-
         // Sanitize the template name to include only alphanumeric characters
         // or underscores.
         $safe_topic = preg_replace('/[^\w]/', '', $name);
```
demiankatz: While we're refactoring this anyway, might be a good time to rename `$safe_topic` to `$safeTopic` to be more in line with normal VuFind conventions.

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,87 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
+        if (isset($language)) {
+            $template = preg_replace(
+                $languagePatternExtended,
+                "\\1$language\\3",
+                $template
+            );
+        } else {
+            $template = preg_replace($languagePatternExtended, '', $template);
+        }
+        $template = str_replace('%language%', $language, $template);
```
demiankatz: Should we use `$language ?? ''` to account for the possible null case here?

module/VuFind/src/VuFind/View/Helper/Root/Content.php
```diff
@@ -0,0 +1,100 @@
+<?php
+/**
+ * Content View Helper to resolve translated pages.
+ * This is basically a wrapper around the PageLocator.
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2010.
```
demiankatz: Might want to change the date to 2021 here.

module/VuFind/src/VuFind/View/Helper/Root/HelpText.php
```diff
@@ -100,39 +82,29 @@ public function getWarnings()
...
-            if ($resolver->resolve($tplFallback)) {
-                $html = $this->getView()->render($tplFallback);
-                $this->warnings[] = 'Sorry, but the help you requested is '
-                    . 'unavailable in your language.';
-            } else {
-                // no translation available at all!
-                $html = false;
-            }
+        if (!$html) {
+            $this->warnings[] = 'Sorry, but the help you requested is '
```
demiankatz: When testing, I never saw this message. In [the template](https://github.com/vufind-org/vufind/blob/dev/themes/root/templates/help/home.phtml), when $html is false, the warnings are not displayed at all, and the `help_page_missing` string is displayed instead. Perhaps it is worth changing the message here to `help_page_missing` (so we don't need to translate a new message for i18n), and then simplifying the template to take advantage of the logic here.

demiankatz (changes requested):
I managed to review this a little sooner than anticipated, fortunately!

@mtrojan-ub, I really like this a lot. I just had a handful of questions and suggestions. I also went ahead and made a couple of small changes myself (moving a constructor to the top of a file, and merging the latest dev code into this branch).

It might be worth discussing the actual template syntax further, in case it could be made more flexible or more consistent with other VuFind conventions (for example, we usually use a double-% for tokens instead of a single-%) ... but it works well, so I certainly don't insist on changing it.

After you address this last round of comments, it might be worth asking for further feedback from @EreMaijala and/or @xmorave2, who might have opinions about some of this.

It might also be worth adding a new test case to the PageLocatorTest to exercise the template mechanism... but that can probably wait until we've all settled on the details.

module/VuFind/src/VuFind/View/Helper/Root/HelpText.php
mtrojan-ub: The message has been changed in the meantime, but i'm not sure how exactly you think the template could be simplified. Could you please explain?

module/VuFind/src/VuFind/Content/PageLocator.php
mtrojan-ub: So would "resultType" be the better key name here? Or something else like "languageVariant"?

module/VuFind/src/VuFind/Content/PageLocator.php
mtrojan-ub: The support method has been created, please take another look.

module/VuFind/src/VuFind/View/Helper/Root/HelpText.php
demiankatz: I thought you might be able to simplify to:

```
<?php $this->headTitle($this->translate('Help')); ?>
<?php foreach ($this->helpText()->getWarnings() as $warning): ?>
  <p class="warning"><?=$this->transEsc($warning)?></p>
<?php endforeach; ?>
<?=$this->helpText()->render($topic)) ?>

```

This way, you leverage the existing warnings mechanism for the "no help" message instead of a special case, and you don't bother checking the return value of render() since it will either return something or nothing, and echoing either will be okay as long as the warning is displayed correctly.

module/VuFind/src/VuFind/Content/PageLocator.php
demiankatz: Maybe matchType, since this is really telling us which type of template pattern was matched in the search?

module/VuFind/src/VuFind/Content/PageLocator.php
demiankatz: Looks good to me -- thanks!

module/VuFind/src/VuFind/View/Helper/Root/HelpText.php
mtrojan-ub: The thing is => the way the HelpText view helper is implemented right now, getWarnings just outputs the messages found in render. So this would imply that you pass the $topic also to getWarnings() and either 2 lookups being made, or the lookup results would need to be cached by $topic inside the view helper (or even deeper). Is this what you mean?

demiankatz: @mtrojan-ub, you're right, that won't work... but it could be adjusted to:

```
<?php $this->headTitle($this->translate('Help')); ?>
<?php $helpText = $this->helpText()->render($topic)); ?>
<?php foreach ($this->helpText()->getWarnings() as $warning): ?>
  <p class="warning"><?=$this->transEsc($warning)?></p>
<?php endforeach; ?>
<?=$helpText?>
```

That's still a bit simpler than the existing template.

demiankatz:
Thanks for the progress, @mtrojan-ub! I'm now asking for reviews from @xmorave2 and @EreMaijala in case they have anything to add. I'm most interested in whether the template approach needs to be adjusted, but I think this moves things in the right direction, and it's now just a matter of deciding on a few details.

EreMaijala (commented):
I'm not sure including the help text translation handling here provides enough benefit to justify the added complexity. I had hard time understanding what's happening with the pattern-based approach, and I feel that the old logic for help pages was much easier to understand. I think I'd prefer to keep it separate.

I also think we should try to avoid doing more work than is necessary. Currently getTemplateOptionsFromPattern alwasys generates all versions, when in the best case (or typical?) the first one from the current language would be sufficient.

demiankatz:
@EreMaijala, I think the benefit to this approach is that it centralizes language-based template loading to a single place, rather than having two different implementations of basically the same feature in two different places. It also adds a helper that exposes that functionality in a more open-ended and flexible way. I like the idea of everything boiling down to the same PageLocator class in the end, as I think this will reduce maintenance in the future.

That being said, I do agree that we may be able to make this a little simpler and more efficient. For one thing, I wonder if we could replace all of the separate str_replace calls with a single preg_replace_callback call that maps the tokens (which might improve uniformity and reduce unnecessary work).

I wonder if we could also improve efficiency by using the generator pattern, so that getTemplateOptionsFromPattern() `yield`s results instead of returning a precalculated array.

What do you think?

EreMaijala:
@demiankatz yield would make sense here. And I understand the flexibility that the pattern allows, but I'd still suggest that the helper have something like `renderHelpPage()` rather than the caller having to specify the pattern in this case. I'm not sure if using preg_replace_callback for everything helps since there are two different patterns, but at least the repeated str_replace class before and after the special language case could be combined by using arrays for strings to search and replace. I'll add some specific comments in the code. I'll add some more comments about minor things as review comments as well.

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,102 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
+     *
+     * @param string $pathPrefix Subdirectory where the template should be located
+     * @param string $pageName   Page name
+     * @param string $pattern    Filesystem pattern
+     * @param string $language   Language
+     *
+     * @return string
+     */
+    protected function generateTemplateFromPattern(
+        $pathPrefix, $pageName, $pattern, $language=null
```
EreMaijala: The equals sign should be surrounded by spaces (same with other methods). Default value could be an empty string (see below).

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,102 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
+    ) {
+        $template = $pattern;
+        $template = str_replace('%pathPrefix%', $pathPrefix, $template);
+        $template = str_replace('%pageName%', $pageName, $template);
+        $languagePatternExtended = '"\\{(.*)(%language%)(.*)\\}"';
+        if (isset($language)) {
```
EreMaijala: I think this should be `if ($language) {` to handle an empty string properly.

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,102 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
+            $template = preg_replace($languagePatternExtended, '', $template);
+        }
+        $template = str_replace('%language%', $language ?? '', $template);
+        $template = str_replace('//', '/', $template);
+        return $template;
+    }
+
+    /**
+     * Try to find a template using
+     * 1) Current language suffix
```
EreMaijala: Language is not necessarily a suffix (like in the help page case), so maybe just leave "suffix" out?

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,102 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
+     *
+     * @return string
+     */
+    protected function generateTemplateFromPattern(
+        $pathPrefix, $pageName, $pattern, $language=null
+    ) {
+        $template = $pattern;
+        $template = str_replace('%pathPrefix%', $pathPrefix, $template);
+        $template = str_replace('%pageName%', $pageName, $template);
+        $languagePatternExtended = '"\\{(.*)(%language%)(.*)\\}"';
```
EreMaijala: %language% doesn't need to be captured, so you can leave the surrounding () out and use \\2 in the replacement.

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,102 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
+        if (isset($language)) {
+            $template = preg_replace(
+                $languagePatternExtended,
+                "\\1$language\\3",
+                $template
+            );
+        } else {
+            $template = preg_replace($languagePatternExtended, '', $template);
+        }
+        $template = str_replace('%language%', $language ?? '', $template);
```
EreMaijala: If we only allow strings, ` ?? ''` would not be needed.

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,102 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
+     * @param string $pathPrefix Subdirectory where the template should be located
+     * @param string $pageName   Template name
+     * @param string $pattern    Filesystem pattern
+     *
+     * @return array Array with template options (key equals matchType)
+     */
+    protected function getTemplateOptionsFromPattern(
+        $pathPrefix, $pageName, $pattern
+    ) {
+        $templates = ['language' => $this->generateTemplateFromPattern(
```
EreMaijala: I think this would be more readable as

```
$templates = [
    'language' => $this->generateTemplateFromPattern(
        $pathPrefix, $pageName, $pattern, $this->language
    )
];
```

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,102 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
-        $templates = [
-            "{$pageName}_$this->language",
-            "{$pageName}_$this->defaultLanguage",
-            $pageName,
-        ];
-        foreach ($templates as $template) {
+    public function determineTemplateAndRenderer(
+        $pathPrefix, $pageName, $pattern=null
+    ) {
+        if (!isset($pattern)) {
```
EreMaijala: You can compare with null directly (with `===`).

module/VuFind/src/VuFind/View/Helper/Root/Content.php
```diff
@@ -0,0 +1,101 @@
...
+        $context=[], &$pageDetails=[], $pattern=null
+    ) {
+        if (!str_ends_with($pathPrefix, '/')) {
+            $pathPrefix .= '/';
+        }
+        $pathPrefix = 'templates/' . $pathPrefix;
+        $pageDetails = $this->pageLocator->determineTemplateAndRenderer(
+            $pathPrefix, $pageName, $pattern
+        );
+        $path = str_replace(
```
EreMaijala: How about making `findContainingTheme()` return a relativePath element as well, so we could avoid this rather ugly replace?

EreMaijala (changes requested):
I've added a couple of comments. I'd also suggest adding type hints to parameters and return values of the new methods.

module/VuFind/src/VuFind/Content/PageLocator.php
demiankatz: $language currently defaults to `null` in the function signature, and I think passing null as the second parameter of str_replace may trigger a notice... so I think we either need this, or we need to change the function signature to default to something different (like an empty string).

EreMaijala: Yes, but I proposed making the default an empty string as there's no need to distinguish between null and ''.

demiankatz: Ahh, sorry, I was just viewing this comment in isolation. If we change the default, I agree entirely!

module/VuFind/src/VuFind/View/Helper/Root/Content.php
mtrojan-ub: This has been changed now; however, findContainingTheme is used not only for templates, so i used str_replace anyway to strip the `templates/` prefix.

module/VuFind/src/VuFind/Content/PageLocator.php
```diff
@@ -84,35 +84,107 @@ public function __construct($themeInfo, $language, $defaultLanguage)
...
+     * Try to find a template using
+     * 1) Current language
+     * 2) Default language
+     * 3) No language
+     *
+     * @param string $pathPrefix Subdirectory where the template should be located
+     * @param string $pageName   Template name
+     * @param string $pattern    Filesystem pattern
+     *
+     * @return array Array with template options (key equals matchType)
```
demiankatz: This comment should be adjusted to match the actual return value.

module/VuFind/src/VuFind/View/Helper/Root/Content.php
demiankatz: I think an unconstrained `str_replace` might have issues here in the unlikely but possible situation where the string occurs more than once in the path; maybe it's better to `preg_replace` `|^templates/|` or use a `substr` to ensure greater precision.

demiankatz (commented):
Thanks for the progress. I resolved @EreMaijala's comments where they were clearly addressed, and I added a couple more minor suggestions. I'll wait and see if @EreMaijala has anything to add. If he's satisfied, and after you've addressed the new comments, I can run tests one last time and merge if all is well!

mtrojan-ub:
Thanks for all the feedback! I hope everything is now implemented correctly. The only thing i'm not sure about is if the discussion regarding the `renderHelpPage()` aspect is entirely finished yet?

demiankatz:
Thanks, @mtrojan-ub! I'd be interested to hear if @EreMaijala has further thoughts on the `renderHelpPage()` issue. Personally, I think it makes more sense to leave the specifics about help pages in the help-specific helper, rather than moving that logic into the core content helper. (I might feel differently if we could completely eliminate the help-specific helper and handle everything through the content helper... but because of the way warnings are handled, etc., it seems cleaner to keep them separated, at least for now).

The other issue that was discussed was simplifying the template application logic. I don't have time to test this right now, but at a glance, I think you might be able to refactor this:

```
        $template = $pattern;
        $template = str_replace('%pathPrefix%', $pathPrefix, $template);
        $template = str_replace('%pageName%', $pageName, $template);
        $languagePatternExtended = '"\\{(.*)%language%(.*)\\}"';
        if ($language) {
            $template = preg_replace(
                $languagePatternExtended,
                "\\1$language\\2",
                $template
            );
        } else {
            $template = preg_replace($languagePatternExtended, '', $template);
        }
        $template = str_replace('%language%', $language, $template);
        $template = str_replace('//', '/', $template);
        return $template;
```

into this:

```
        $standardReplacements = [
            '%pathPrefix%' => $pathPrefix,
            '%pageName%' => $pageName,
            '%language%' => $language,
            '//' => '/',
        ];
        $languagePatternExtended = '"\\{(.*)%language%(.*)\\}"';
        $languagePatternExtendedReplacement = $language ? "\\1$language\\2" : '';
        return str_replace(
            array_keys($standardReplacements),
            array_values($standardReplacements),
            preg_replace(
                $languagePatternExtended,
                $languagePatternExtendedReplacement,
                $pattern
            )
        );
```

It's possible I've missed a detail somewhere, but I think that should do the same thing, but in fewer steps.

module/VuFind/src/VuFind/View/Helper/Root/HelpText.php
```diff
@@ -38,26 +38,12 @@
...
-
-    /**
-     * The default fallback language
-     *
-     * @var string
-     */
-    protected $defaultLanguage;
-
     /**
      * The context view helper
```
EreMaijala: This should be changed to `Content` as well.

EreMaijala (changes requested):

EreMaijala:
I left a comment about a comment, but other than that it looks fine to me.

demiankatz:
Thanks for the comment, @EreMaijala. I subsequently noticed that some conflicts have been introduced, so I resolved them. I haven't had a chance to test the results yet. @mtrojan-ub, I'll wait for you to respond to @EreMaijala's comment, and then I can do one more round of testing, and hopefully get this merged.

mtrojan-ub:
Sorry for the late response... i haven't been in the office for 2 weeks, and still got lots of things to catch up, but i hope i can have a closer look at this until the end of the week.

/edit: done now!

module/VuFind/src/VuFind/View/Helper/Root/HelpText.php
mtrojan-ub: done!

xmorave2 (changes requested):
Sorry to get into discussion so late.

I do like the refactoring of PageLocator and the Content helper seem like reasonable idea. I have just on concern: the markdown is not processed when used from template helper:

`<?=$this->content()->renderTranslated('some_markdown_page'); ?>`

demiankatz:
Thanks for catching this, @xmorave2 -- we definitely do not want to merge this before all content types can be displayed through the view helper; I think I must have assumed it worked without testing it.

This also reminds me that we lack tests for the new helper. Perhaps a useful next step would be to build a unit test, include a Markdown case, and then fix the problem from there.

The question, then, is how we should approach getting the work done. I'm out of time for VuFind-related work this week, but I might have time to begin building a test next week if that would help move things along. @mtrojan-ub, what do you think?

mtrojan-ub:
@demiankatz: Feel free to continue on this, i will be out of the office next week and we also got tons of new projects right now, so it might take a while until i can have closer look at this PR again.

demiankatz:
@mtrojan-ub, is there any chance you have time to merge the dev branch into this branch? Because of the recent addition of GitHub Actions configuration to VuFind's repository, I do not have appropriate permissions to merge and push that myself, since it adds a workflow to your repo. If you can do that once, then I should be able to add changes again as a follow-up. If you do not have time, I can begin the work in this slightly outdated branch, but I'd prefer to have the very latest code if possible.

[note: this comment somehow got posted to the wrong pull request on my first attempt, so I deleted it and moved it here... not sure how that happened!]

mtrojan-ub:
Sure! i wont have time today, but i can do that tomorrow.

/edit: The dev branch has been merged into this, and i also fixed a problem with a failing test.

demiankatz:
Thanks, @mtrojan-ub! I hope to have time to look at this more closely next week.

demiankatz:
Sorry it has taken me so long, but I finally managed to take a closer look at this. I think the best solution may be to utilize the logic in the TemplateBased ContentBlock. I refactored to try this in 09cfc91, and it seems to be working (it is now possible to load help pages written in Markdown!), but I had very limited time for testing, so it's possible there are some flaws in my approach. It also required significant refactoring of the ContentBlock itself to support new features. @mtrojan-ub / @xmorave2, what do you think? If you have time to test, I'd appreciate more input. When I'm back from vacation after next week, I can put some more time into this -- if nothing else, I definitely think we should improve our test coverage here before merging.

xmorave2 (approved):

xmorave2:
> Sorry it has taken me so long, but I finally managed to take a closer look at this. I think the best solution may be to utilize the logic in the TemplateBased ContentBlock. I refactored to try this in [09cfc91](https://github.com/vufind-org/vufind/commit/09cfc9100dfddf483ae58c72876a6adf59decb8b), and it seems to be working (it is now possible to load help pages written in Markdown!), but I had very limited time for testing, so it's possible there are some flaws in my approach. It also required significant refactoring of the ContentBlock itself to support new features. @mtrojan-ub / @xmorave2, what do you think? If you have time to test, I'd appreciate more input. When I'm back from vacation after next week, I can put some more time into this -- if nothing else, I definitely think we should improve our test coverage here before merging.

Thanks @demiankatz, I tested it and it looks like it works, but to have more unit tests will be very valuable here.

demiankatz:
Thanks, @xmorave2 -- I believe I have now built tests to cover all of the new components (and the refactored existing ones). Do you see anything missing, or are we ready to merge this now? Also, I'd still be interested in hearing whether @mtrojan-ub has any further input.

xmorave2:
> Thanks, @xmorave2 -- I believe I have now built tests to cover all of the new components (and the refactored existing ones). Do you see anything missing, or are we ready to merge this now? Also, I'd still be interested in hearing whether @mtrojan-ub has any further input.

Thanks @demiankatz, it looks good to me, great job.

mtrojan-ub:
Looks good to me too, thanks!

demiankatz:
Thank you both! Merging now...

PR #1875, opened 2021-03-09
Facets.js::buildFacetNodes without jQuery

LuomaJuha:
Here is a version where the facetnodes are build without using jQuery.

TODO
- [x] Run full test suite.

### DIFF
--- a/themes/bootstrap3/js/facets.js
+++ b/themes/bootstrap3/js/facets.js
@@ -3,74 +3,79 @@
 function buildFacetNodes(data, currentPath, allowExclude, excludeTitle, counts)
 {
   var json = [];
+  var selected = VuFind.translate('Selected');
+  var separator = VuFind.translate('number_thousands_separator');
 
-  $(data).each(function facetNodesEach() {
-    var $html = $('<div/>').addClass('facet');
-    var url = currentPath + this.href;
-    var $item = $('<span/>')
-      .addClass('main text' + (this.isApplied ? ' applied' : ''))
-      .attr('role', 'menuitem')
-      .attr('title', this.displayText);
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks, @LuomaJuha! I'm going to run my test suite on this and see if there are any problems. After that, I'm hoping @crhallberg can take a look and offer feedback.

demiankatz:
Existing tests are passing (though they may not be exercising every line of code). I'll defer to @crhallberg for further comments!

LuomaJuha:
Noticed that I was appending wrong icon at the end but fixed it. There is not really much difference in this one as it is just creating lots of html elements :)

crhallberg (approved):
This is well done! I wouldn't change a thing but I do wonder if a basic HTML element factory and parser would come in handy as we build more advanced JS logic away from jQuery. Thank you for this!

demiankatz:
I'm merging now; thanks everyone! If we need an HTML element builder, we can start a new PR as needed. :-)

EreMaijala:
@LuomaJuha can you check exclude facets? E.g. set 

[Results_Settings]
exclude = "*"

Looks like clicking the x behaves the same as clicking the facet value, and there's no space between the x and the badge.

EreMaijala:
Follow-up: #1878.

demiankatz:
Thanks for catching that, @EreMaijala! Looks like we need to add a Mink test case for excluding hierarchical facets.

demiankatz:
I've opened [VUFIND-1471](https://vufind.org/jira/browse/VUFIND-1471) as a reminder to add the test; when I get to Mink testing in my video series, this might be an example to consider using.

PR #1876, opened 2021-03-09
Fix for a null variable situation in Piwik helper

aleksip:
We want to track deferred recommendations using Piwik, and came across this situation where `$current` may be `null`, causing an exception.


demiankatz:
Thanks, @aleksip, this is clearly a bug fix, so I have backported it to the release-7.1 branch for inclusion in release 7.1.1.

It would be useful to have a unit test for this helper -- if we had written tests in the first place, we would probably have caught this problem much sooner. I'm planning to record a tutorial video on test writing next month; I'll consider this helper as a possible example subject.

demiankatz:
Closing now (fix manually backported).

PR #1877, opened 2021-03-10
Email change tweaks

EreMaijala:
The brackets increase the chances of an email client not messing with the url. There are also some fixes to Finnish and Swedish texts. Also the notification email to the old address will only be sent if the address is valid. This avoids the situation where a bad existing address (easily verified by logging in with the Demo ILS driver) prevents changing the email address.


EreMaijala:
I added a couple more improvements. Redirecting to profile page also fixes a broken login page that was displayed when ChoiceAuth was used.

All commits are standalone and can be merged or rebased to dev.

languages/en.ini
```diff
@@ -1237,10 +1237,10 @@ Username cannot be blank = "Username cannot be blank"
 Username is already in use in another library card = "Username is already in use in another library card"
 verification_done = "Your email address has been verified successfully."
 verification_email_change_sent = "Email address verification instructions have been sent to the new email address. You must verify the address before the change will take effect."
-verification_email_notification = "A request was just made to verify your email address for your account with %%library%%."
+verification_email_notification = "A request was just made to verify your email address for your account with %%library%%. Please ignore this message if you have not added or changed your email address."
```
demiankatz: It might be better to set separate translation strings for the newly added sentences and add them separately in the email template. That will make it easier to ensure that all of the translations remain consistent with each other.

demiankatz (commented):
Thanks, @EreMaijala! I've cherry-picked the email validation commit to dev, since that's very simple and straightforward. I had one suggestion regarding the language file changes. I'll test the rest when I have a little bit more time, but it looks good at a glance!

languages/en.ini
EreMaijala: I decided to roll back most of the changes. I couldn't come up with meaningful translation keys, and we can always tweak the texts locally.

demiankatz: Thanks, @EreMaijala. I think those changes are useful, but it is probably less work to keep them as local string customizations for the moment. Happy to revisit in the future (and/or brainstorm key names) if that would be helpful, though!

demiankatz:
Thanks, @EreMaijala, this looks good to me now!

PR #1879, opened 2021-03-10
Aleph driver: support for administration identifier (adm_id)

xrosecky:
We use administration identifier (adm_id) in our modified MultiBackend driver.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Aleph.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Aleph.php
@@ -1010,10 +1010,12 @@ class Aleph extends AbstractBase implements \Laminas\Log\LoggerAwareInterface,
             $author = (string)$z13->{'z13-author'};
             $isbn = (string)$z13->{'z13-isbn-issn'};
             $barcode = (string)$z30->{'z30-barcode'};
+            $adm_id = (string)$z30->{'z30-doc-number'};
 
             $transList[] = [
                 //'type' => $type,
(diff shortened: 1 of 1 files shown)


demiankatz:
@xrosecky, can you say a little more about how this is used? It certainly won't hurt anything to add another field to the return values, but it might be worth adding a comment to explain why the field is there (in case someone maintaining the code in the future has a question), and if this is something that might apply to other ILSes as well, we should consider whether the field needs to be added to the ILS driver spec, and whether there are other drivers putting an equivalent value in a differently-named field.

xmorave2:
@demiankatz, as far as I know, Aleph is (in some configurations) not able to report the bibliographic identifier. In these cases we use 'adm_id' to identify a record in solr index.

demiankatz:
@xmorave2, if you're using `adm_id` in place of `id`, would it make more sense to just populate `id` with the `adm_id` value based on a configuration, rather than creating a whole separate field for it?

xrosecky:
We need both identifiers.

demiankatz:
@xrosecky, in that case, how about adding a comment above the new code, like:

`// Secondary, Aleph-specific identifier that may be useful for local customizations:`

That way, if we review code when revising the ILS driver spec in future, it will be clear that this is something Aleph-specific, and it won't cause any confusion. If that works for you, then I'm happy to merge this as-is. :-)

xrosecky:
OK, I added the comment.

demiankatz:
Thanks, merging now!

PR #1881, opened 2021-03-10
add enumchron field to getHolding method in FOLIO driver

ihardy:
FOLIO has at least three fields in the item response that come back that are candidates for enumeration information: volume, enumeration, and chronology. I'm just combining them if they exist here. Not 100% sure if this is a typical approach. Looking for input from librarians.

item schema: https://github.com/folio-org/mod-inventory-storage/blob/master/ramls/item.json


module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -549,11 +549,21 @@ public function getHolding($bibId, array $patron = null, array $options = [])
                 );
                 $locationId = $item->effectiveLocationId;
                 $locationName = $this->getLocationName($locationId);
+                // concatenate enumeration fields if present
+                $enum = join(
```
demiankatz: `join` is an alias of `implode`, and VuFind uses `implode` throughout the code. Probably best to switch to that for consistency.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -549,11 +549,21 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
+                            $item->enumeration ?? null,
+                            $item->chronology ?? null]
+                    )
+                );
                 $items[] = [
                     'id' => $bibId,
                     'item_id' => $item->id,
                     'holding_id' => $holding->id,
                     'number' => count($items) + 1,
+                    'enumchron' => $enum ?? '',
```
demiankatz: I'm pretty sure the `?? ''` is unnecessary here, since $enum should never be null -- `implode` will always return a string.

demiankatz (commented):
Thanks, @ihardy! I think it would be helpful to see some real-world data to get a better sense of how this would work. I gather from the schema that "volume" should only be used by monographs, and "enumeration" and "chronology" should only be used by serials... but I wonder if there's ever a record that has all three for some reason. I don't think it will ever hurt to pass this information through the enumchron field, since that is just a display field for clarifying the context of individual items... but depending on the content/formatting of the fields, the way you combine them together might vary. Perhaps it would even make sense to have a configurable separator, in case different cataloging practices at different institutions might lead to different needs.

ihardy:
Thanks for the comments @demiankatz is enumchron meant to be used for serials only? I'll track down some real world examples of different material types.

demiankatz:
@ihardy, I think in the pure world of theory, enumchron only makes sense for a serial, because monographs are one-off publications that have no need for enumeration or chronology. But of course, if you put a monograph in a series, it immediately becomes more complicated. So I'm sure the real world is messy and confusing!

damien-git:
This PR looks good. I tried the same changes (by extending the Folio driver class), and it fixes the "Copy 1...Copy 99" display for periodicals. In our (Michigan State) Folio data, the volume attribute is used but not enumeration or chronology. So it is not properly following the schema, and it is this way because of the imperfect migration from an earlier system. No further migration is planned.
@demiankatz Can this be merged ? I don't see how the PR could possibly lead to a worse display than what we have now.

demiankatz:
Thanks for trying this, @damien-git; while there was some outstanding discussion about possibly making the separator configurable, I don't object to merging this now as incremental progress and continuing that discussion later, since this hasn't moved in quite some time, and as you say, it doesn't make things worse.

I did notice that this PR had a merge conflict with the dev branch, and I've just resolved it. Do you have a moment to test this updated version of the driver to be sure it still works as expected? If you're able to do that and give the thumbs up that it works, I'll go ahead and merge it.

damien-git:
@demiankatz Thanks. I actually don't have an easy way to test Vufind's dev branch, all I can do is copy the changes into our environment and see if it works. It looks like it does now. In terms of features, I feel like a more important one would be sorting the items by enumchron or by volume/chronology/enumeration. I'm not sure how to do that in the general case, but I will probably implement something specific to MSU using the volume. There is something similar in SierraRest (see "statusSortFunction"). In any case, this can be done in a separate PR.

demiankatz:
Thanks, @damien-git! I'll merge this now, and if you want to open a follow-up PR to work on sorting, I'd certainly be happy to continue the conversation there.

demiankatz (approved):

PR #1883, opened 2021-03-12
Update jit-grunt package.

demiankatz:
I had trouble getting `npm install` to work using Node 15 until I updated this package.


demiankatz:
@crhallberg, can you review and comment? If we need to do more than just bump this dependency version, I'm open to repurposing this PR into a larger discussion of Javascript tooling.

crhallberg:
I can't think of any other necessary steps unless we want to update more. Let's start here.

demiankatz:
Great, merging now!

PR #1884, opened 2021-03-15
Initialize search parameters from request in Recommend Ajax handler

aleksip:
It seems useful to initialize the search parameters here in case all request parameters from the original search request are passed to the Ajax handler. This way it is more likely that a recommendation module can use the passed objects in a similar fashion regardless whether it is a deferred call or not.

### DIFF
--- a/module/VuFind/src/VuFind/AjaxHandler/Recommend.php
+++ b/module/VuFind/src/VuFind/AjaxHandler/Recommend.php
@@ -99,7 +99,11 @@ class Recommend extends AbstractBase
         $module = $this->pluginManager->get($params->fromQuery('mod'));
         $module->setConfig($params->fromQuery('params'));
         $paramsObj = $this->results->getParams();
-        $module->init($paramsObj, new Parameters($params->fromQuery()));
+        $request = new Parameters($params->fromQuery());
+        // Initialize search parameters from Ajax request parameters in case the
+        // original request parameters were passed to the Ajax request.
+        $paramsObj->initFromRequest($request);
+        $module->init($paramsObj, $request);
         $module->process($this->results);
 
         // Render recommendations:
--- a/module/VuFind/tests/unit-tests/src/VuFindTest/AjaxHandler/RecommendTest.php
+++ b/module/VuFind/tests/unit-tests/src/VuFindTest/AjaxHandler/RecommendTest.php
@@ -104,12 +104,14 @@ class RecommendTest extends \VuFindTest\Unit\AjaxHandlerTest
             ->will($this->returnValue($mockPlugin));
         $this->container->set(PluginManager::class, $rm);
 
-        // Set up results plugin manager:
+        // Set up results and results plugin manager:
+        $results = $this->getMockResults();
+        $results->getParams()->expects($this->once())->method('initFromRequest');
         $resultsManager = $this->container
             ->createMock(ResultsManager::class, ['get']);
         $resultsManager->expects($this->once())->method('get')
             ->with($this->equalTo('Solr'))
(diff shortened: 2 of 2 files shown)


demiankatz:
Thanks, @aleksip, I think I can see how this could be useful, but do you have a more specific use case to share? I'm trying to wrap my head around the best way to test this.

Also, speaking of tests, what exactly are your changes to the test designed to accomplish here? I think we might need some more assertions somewhere in order to ensure that the test changes are actually proving something.

aleksip:
@demiankatz My use case is that I want to use the `Query` object via `$params->getQuery()`, and if it is a deferred call all variable values in the `Query` object are `null`.

The changes I made to the test were just to get it to pass, because without the added mocks `$paramsObj` was `null`.

demiankatz:
@aleksip, that makes sense. To make the test a little more meaningful, I'd suggest that instead of calling `$this->getMockResults()` directly, you instead set up some expectations, like this:

```
$mockResults = $this->getMockResults();
$mockResults->getParams()->expects($this->once())->method('initFromRequest');
```

At the bare minimum, that will confirm that initFromRequest is actually getting called on the appropriate object. For extra credit, it might be worth adding a `->with()` condition to confirm that the correct request parameters are getting passed in. That might be a little complicated because of the way the request gets wrapped in a Parameters object, so I'll leave it up to you whether or not it's worth the effort.

If you don't mind tweaking that test a little bit, then I can run the full test suite to make sure there are no side effects, and I can probably merge after that. Thanks!

aleksip:
@demiankatz I made an attempt for the extra credit, but it did seem quite complicated indeed, especially as I have not written that many PHPUnit tests before. So here is the bare minimum.

demiankatz:
Thanks, @aleksip! I'm asking for a review from @EreMaijala in case he has anything to add. If he's happy with this, I'll take a few minutes to expand the test (if time permits), run the full test suite, and merge this if everything is behaving as expected.

EreMaijala (approved):
Looks good to me!

demiankatz:
Thanks! For your reference, @aleksip, I looked up the best approach to making the test more specific, and this is what I came up with:

```
        $results = $this->getMockResults();
        $testRequestInitialization = function ($request) {
            // exactly one parameter: mod = foo
            return $request->get('mod') === 'foo'
                && count($request) === 1;
        };
        $results->getParams()->expects($this->once())
            ->method('initFromRequest')
            ->with($this->callback($testRequestInitialization));
```

This uses a callback function to confirm that the request passed to initFromRequest is the same one that gets passed into the request handler.

I would have pushed this up to the PR directly, but I don't have write access to your repo. Instead, I'll just commit the change locally and do a manual merge of the PR including the addition. Right now, tests are running... if all passes, I'll wrap this up in a few more minutes.

aleksip:
@demiankatz Thanks, this was good to learn, I am hoping to get more familiar with PHPUnit in the future.

I do have the "Allow edits by maintainers" setting on for the PR, I thought that would have given you write access? But I am happy to get the change in one way or another.

demiankatz:
That is strange; maybe I did something wrong on my end. I'll let you know if I have trouble again in the future. In any case, all tests are passing, and I have merged this manually to the dev branch as commit 3695e24cda103e9b922b8e205693ea69b68f39f4. Thanks again!

demiankatz:
...except I messed up my Git commands and pushed the wrong commit (looks like I forgot to `git add` before I ran `git commit --amend` -- my most common mistake). I fixed the test with commit 126771241506850c57617d54d8df180811b7b54b. Sorry about that!

PR #1232, opened 2018-09-24
More intuitive merge order of language files

sebkehr:
As already mentioned on the [mailing list](https://sourceforge.net/p/vufind/mailman/message/36241861/) loading/merging of language files may happen in an unexpected manner.

When looking closer at the [ExtendedIni-Loader](https://github.com/vufind-org/vufind/blob/v5.0/module/VuFind/src/VuFind/I18n/Translator/Loader/ExtendedIni.php) the main cause for this can be identified to be the recursive calls of `loadParentData()` and `loadLanguageFile()` possibly leading to resolving a parent language file against all registered base paths (elements of `$pathStack`).

For a more predictable behavior `@parent_ini` statements should be only evaluated w.r.t. the directory of the language file which contains the statement.

Therefore I suggest the attached changes.

TODO

- [ ] resolve VuFindDevTools compatibility issues
- [ ] add comments to newly introduced code
- [ ] briefly summarize introduced changes/features  (esp. `@parent_ini` vs. `@extends` and `fallbacks`)
- [ ] nice-to-have: allow to add language directories with a given priority
- [ ] nice-to-have: provide initialized values as service and factor out view model initialization
- [ ] nice-to-have: provide a way to extend language files by absolute path

PR #1885, opened 2021-03-17
Use slm/locale for Locale detection.

demiankatz:
This PR extracts (and modifies/builds upon) some work that was done as part of the stalled PR #1232, but which may be useful on its own.

TODO:
- [x] Refactor more logic from Bootstrapper into Translator factory, if possible.
- [x] Review whether generator setup in LocaleDetectorFactory::getStrategies() is entirely correct.
- [x] Fix LanguageInitializerTrait to prevent double-initialization when shared between classes.
- [x] Review whether we can pull locale directly from the LocaleSettings object in more places (instead of relying on translator).
- [x] Test that all functionality of old Bootstrapper code is correctly replicated by new logic (browser detection, GET/POST, cookies, etc.).
- [x] Run full test suite.


demiankatz:
@EreMaijala, please take a look at this when time permits -- as noted above, it pulls some ideas from PR #1232 and then takes them a little further. The goal is to further reduce confusing logic in the bootstrapper, to leverage an existing library for locale detection, and to untangle some unnecessary dependencies on the translator. It also makes the code a little more tolerant of a missing PHP intl extension (it will now fill the error log with warnings but run, instead of offering the "white screen of death").

module/VuFind/src/VuFind/Bootstrapper.php
```diff
@@ -190,133 +189,35 @@ protected function initContext(): void
...
-            $langStrip = current(explode("-", $language));
-            if (in_array($langStrip, $validLanguages)) {
-                return $langStrip;
-            }
-        }
-
-        return null;
+        $settings = $this->container->get(LocaleSettings::class);
+        $viewModel = $this->container->get('HttpViewManager')->getViewModel();
+        $viewModel->setVariable('userLang', $locale = $settings->getUserLocale());
```
EreMaijala: I'd take setting of $locale out of the assignment for clarity.

EreMaijala (commented):
- Is there a reason for not converting more translator->getLocale calls to use LocaleSettings (e.g. FacetCacheFactory)?
- Do you think it would make sense to add LocaleAwareInterface/LocaleAwareTrait for easy injection?

demiankatz:
Thanks, @EreMaijala! I thought I replaced the majority of the translator requests with LocaleSettings requests, but it would appear that I missed some. It was not intentional, so I'll look again when time permits.

Regarding auto-injection of locales, I have somewhat mixed feelings at this point about auto-injection based on interfaces; I've been trying to move away from that kind of magic toward shared factories where practical. However, if you have a strong use case for it, I'm happy to consider it.

Also, semi-related: #1232 also has some work to activate the Laminas\I18n module in application.config.php, which I believe takes care of some of the translator setup automatically. I plan to review that more closely in the near future, and may open another PR with related work if it seems useful... but I wanted to get this part done first before making it more complicated. I'm hoping that might enable us to further reduce the complexity of some custom factories, but I have to more closely review exactly what it is doing to be sure.

EreMaijala:
@demiankatz No strong case for the interface/trait. I just thought that since it's available for translator, it might be useful here as well. But I see the point in moving away from those, so let's leave it out.

demiankatz:
@EreMaijala, I finally had time to give this another look, and the PR already includes a change to use LocaleSettings instead of the Translator in FacetCacheFactory. I can't find any references that I missed. Is there something I'm overlooking/forgetting about?

Beyond that, I've just addressed your other small suggestion and have merged the latest dev branch into this one. I'm going to run the test suite again now just to be sure nothing has broken. If all of that passes, and if you can't find more code that needs refactoring, are you okay with having this merged now?

Thanks for the review!

EreMaijala:
@demiankatz Sorry, I must have been looking at wrong branch. It all looks good to me.

EreMaijala (approved):

demiankatz:
Great, thanks; all tests are still passing, so I'm merging this now.

PR #1886, opened 2021-03-18
Add search backend QueryAnalysisInterface and GetSearchTermsCommand

aleksip:
In some cases VuFind code may need to process VuFindSearch backend-specific queries, e.g. a recommendation module might need tokenized query terms.

The main idea behind this feature suggestion is to have all backend-specific query processing in backend-specific classes in VuFindSearch, and accessible via the `Service` class.

`SpellingProcessor` was not trivial to refactor to use this new feature, especially in a backwards compatible way, so for the time being I solved that by using a trait. But if this feature is considered a good idea, it would also be good to refactor `SpellingProcessor` and move `tokenize()` from `QueryTokenizerTrait` to `Solr/QueryHelper`.

I am working on a real use case where I have tested this feature and it would fill the requirements.


demiankatz:
Thanks for this! I guess my first question is whether this really needs to be a backend method. It seems like you're adding quite a few methods that exist solely to proxy other methods. In some cases, this can certainly be justified -- but I wonder if any of it is unnecessary, and it might be easier to simply go directly to the source when needing to access particular pieces of functionality. For example, adding a method to the main search backend adds a lot of complexity around event hooks, etc., but is any of that needed just to retrieve a helper object? It seems like it might be overkill. Would it be easier to just pull it directly from the BackendRegistry?

I'm happy to review this in more depth, but before I do, I wanted to start by seeing whether anything could be easily simplified.

aleksip:
Some of the proxy methods are there just for backwards compatibility, and the implementation could be refactored to remove them. But even taking that into account I agree that there is a lot going on just to get the helper.

I'm very happy to simplify this as much as possible. Can you elaborate on how pulling directly from `BackendRegistry` would work?

demiankatz:
I agree that in some cases, creating methods for backward compatibility may actually be less work than refactoring all the code that depends upon them... but I'm open to refactoring if it's easier, since this will be going into a major version, and we can always note changes in the changelog.

Regarding the BackendRegistry, that's actually just a piece of the BackendManager, so I may have misspoken slightly. But if I'm remembering the details correctly, with the setup you currently have here, you can get to the helper object like this:

```
$container->get(\VuFindSearch\Backend::class)->getQueryHelper('Solr');
```

But if you omit the changes to the Backend, you could still get the same object like this:

```
$container->get(\VuFind\Search\BackendManager::class)->get('Solr')->getQueryHelper();
```

It's slightly more verbose, but it seems manageable to me!

aleksip:
My implementation was based on the assumption that backends should only be retrieved by and accessed via `Service` and the events-based system. Getting them directly from the container or `BackendManager` seems to bypass the events-based system, making replacement of backends via the 'resolve' event impossible?

I have now noticed that `SwitchQuery` for example accesses `BackendManager` this way. Also the events themselves expose direct access to backends, which too allows bypassing the event system and everything depending on it. All this seems somewhat problematic as it seems to make the whole event-based system less robust. Am I just missing/misunderstanding something?

demiankatz:
@aleksip, I suppose the question is "what is the intended scope and purpose of the event system?" I didn't write it (it was a contribution from David Maus, who has moved on to other projects in the intervening years), and so it's possible that my subsequent work hasn't fully reflected its original intentions. Perhaps your observation here reveals the need for a bigger project to standardize some of this one way or another, or at least to better document it.

I confess that I find the event system confusing, and while I recognize its value (and have taken advantage of it on some important occasions), I also try to avoid or circumvent it when it's not absolutely necessary, because I feel that using it makes my code significantly harder to understand. It makes sense to me to hook events when I need to manipulate inputs and outputs of the search process, since there is no other way to do that... but I'm not sure why I would want to use events to override an entire search backend, when the Service Manager system provides a much easier way of doing that through configuration. There may very well be a use case I'm not thinking of here, like a situation where you want to use different backends to handle different search actions for the same system -- but it has never crossed my mind before.

I actually tried to follow the code for the resolve action to try to get a better understanding of how/where/when it comes into play, and I confess that it took me quite a long time to refresh my memory due to a number of unclear things -- the search service is built in a factory but set up in the bootstrapper (I see no obvious reason why we shouldn't move the setup logic to the factory, so we get on-demand initialization). The search service defines an EVENT_RESOLVE constant, but then uses a hard-coded string when hooking the event in the bootstrapper. I realize this is all a tangent, but it would probably be a good idea to clean it up to make this easier to follow in future! (If you agree that these changes might make sense, I can make a separate PR to discuss them further... it would be a good follow-on from my ongoing work to shrink/simplify the bootstrapper).

Anyway, the bottom line is that because I didn't write the search service, and because I'm not entirely comfortable with its design, I tend to be extremely wary about changing or extending it. While past changes have added new services that follow a very similar pattern to existing services -- i.e. retrieving data from the backend -- using the entire weight of the search service to simply retrieve a helper object that exists elsewhere in the service manager seems to me unnecessary. But your point about the resolve event is well-taken and could change my mind. That being said, if we decide that search backends must ONLY be retrieved through the search service, then I think we need to change the interface of the search service to make the resolve method public, so it's possible to fetch backends outside of the service... and then we need to clean up all the existing direct references to the backend manager.

As you can see, you've stepped into a thorny and complicated issue... but thanks for bringing it up! Hopefully through this discussion we can make some improvements to the code's consistency and comprehensibility. It would probably be a good idea to bring in some more opinions, as well -- let me know if/when you'd like me to attract some more attention to this conversation. :-)

aleksip:
@demiankatz Thank you for your long and detailed answer! It took me a quite a while to understand the event-based backend resolving, and I agree that it is not the simplest of designs.

I don't remember coming across this design pattern before and I'm not aware of what the benefits are over the otherwise used service container/dependency injection pattern. Possibly the event pattern allows for more dynamic backend overrides in some situations?

I guess it would be good to survey if anyone is using event-based backend resolving to replace the default backends. If no-one is using it, perhaps it could be replaced with direct `BackendManager` calls. I agree this would make things simpler and bring backend retrieval more in line with the rest of the codebase.

If I have understood things correctly, it would be best to avoid mixing the two approaches, since that makes the event-based system unreliable as events are fired in some situations but not in others.

demiankatz:
I think one aspect of the design that may add clarity is the idea that code in the VuFindSearch namespace was designed to be extremely generic and reusable (and possibly split off to a stand-alone repository in the future, though that obviously hasn't happened), while the VuFind namespace contains code that is very specific to the VuFind application as a whole. It seems likely that the resolve action was set up so that VuFindSearch code could be repurposed in an environment that uses a completely different backend resolution strategy than VuFind's container/dependency injection approach. Thus, code in the VuFindSearch namespace should never refer to the BackendManager, because it shouldn't know about its existence... but code in the VuFind namespace (and other namespaces that depend directly upon VuFind namespace code) can more safely make assumptions about the BackendManager.

That's just a theory, not established fact, but it seems plausible to me and gives me a set of rules I can live with for how/when to use the BackendManager (and which seem to be currently followed in the existing code). I agree that it is probably worth reaching out to the community to see if anyone is using the resolve event in other ways, but I think that is fairly unlikely.

Does that make you any more comfortable with a particular path forward, or do you want me to get the larger conversation started before proceeding further?

aleksip:
That does does make sense to me and indeed seems like a good reason to use the event-based approach.

However, it still seems to me that retrieving backends and calling backend methods directly results in not just the 'resolve' event not being fired, but may also result in 'pre' and 'post' events for backend methods provided by `Service` not being fired?

Your suggestion of making the `Service::resolve()` method public sounds like a good approach, but I believe that it will not remove the problem for 'pre' and 'post' events? Just a quick thought, but maybe the public `resolve()` method could return a proxy object that forwards the event triggering method calls to `Service` and others directly to the backend? Or maybe there is a better way to get the same end result.

demiankatz:
Yes, @aleksip, that's a very good point. This is getting very complicated very quickly, though, and I'm beginning to wonder if we should look at it from a completely different angle. There are really two different reasons we might want to retrieve a backend object through `resolve()`:

1.) To gain access to helpers and other specialized functionality beyond the scope of what the search service can handle. This is a valid and important use case.

2.) To run actions directly on the search backend which are normally meant to be run through the search service. This is a bad idea and will cause problems with events not triggering. We don't want to encourage people to do this. We are currently doing this in at least a couple of places in the code where we probably shouldn't be.

I think for situation no. 1, maybe we should think about whether we can make those resources accessible in a different way, so we don't need to add lots of infrastructure to the search service to get them back out again. For example, with the helper object here, what if we register it as a service in the VuFind module, and then inject it through the factory? That way, we can pull it out of the VuFind service manager when we need it, but we also know it's the same object being used by the search service... but the search service itself can create a default object internally if nothing is injected. It keeps the search service simple and generic, while also giving us a guaranteed safe and reliable way to manipulate/override/access the object in VuFind-land code. Does that make any sense?

For situation no. 2, I think some refactoring may be needed to fix existing problems and inconsistencies. That's probably a project for a separate PR. Perhaps we should open a JIRA ticket to capture the need for the work, since I don't think I'll be able to get to it right away.

demiankatz:
Also, a minor side note related to earlier conversation: I just pushed up commit a9fa617f14, which uses the appropriate constant instead of a string for the resolve event in the Bootstrapper, just to make things a little more clear/explicit. I had second thoughts about moving the bootstrap method to a factory, since it's a little unclear exactly where that logic should live. That might be worth revisiting later, but that's a bit off topic for now.

aleksip:
> For example, with the helper object here, what if we register it as a service in the VuFind module, and then inject it through the factory? That way, we can pull it out of the VuFind service manager when we need it, but we also know it's the same object being used by the search service... but the search service itself can create a default object internally if nothing is injected.

Would the helper be registered as a backend specific query helper, or should there be something like a `QueryHelperManager` and a `QueryHelperRegistry` so all possible backend specific query helpers can be added/requested in a consistent way?

aleksip:
Some additional thoughts: although the PR as it stands seems like a lot just to get a helper object, it only adds one new method and one feature interface to `Service`.

The rest is about adding query helpers in a generic way and refactoring the current implementation accordingly. It could be also be refactored to be simpler if BC is not an issue.

I guess the big question is whether we want to add query helpers in a generic way, or just to go for a simple Solr-specific solution for the time being (similar to how recommendation modules are somewhat Solr-specific)?

If we want to go for a generic implementation, I'm not sure if the service-based approach would be that much lighter. However, it would add a different way to add/request backend-related features, and if there is other VuFindSearch functionality required in VuFind in the future, that would again need its own solution.

I'm happy going forward with any of these approaches.

demiankatz:
@aleksip, apologies for the long silence on this PR; I went on vacation and since returning have been catching up on more recent submissions. I hope to find some time to give this a little more thought soon, and in the meantime I'm definitely interested to hear whether @EreMaijala or others have any thoughts of their own on the best path forward.

EreMaijala:
All in all I think this is a prime example of why modularization is difficult. I'm not claiming to have the clearest thought about all this right now, but I think gaining access to a helper in VuFindSearch from the outside is perhaps not the most optimal way. I'd rather keep any helpers and such an implementation detail of the backend in VuFindSearch without exposing them.

For this case, tokenizing the query, the cleanest approach I can think of is to add that specific function as a feature interface to the backend. To make if cater for the current and potential future needs, I'd call it something like `QueryAnalysisInterface` that would contain method `analyzeQuery`. For now it would take a `\VuFindSearch\Query\Query` as parameter and return an array like:

```
[
  'tokenized' => [
    'first',
    'second'
  ]   
]
```

This would leave tokenization a process that's internal to the backend and it would be free it to do whatever it needs to for accomplishing the desired result, such as calling the actual search engine for tokenization.

demiankatz:
@EreMaijala, I am inclined to agree that if we're going to expose something new through the search backend, it feels a little more natural to expose something that returns data rather than something that returns a helper object -- it's more in keeping with the existing pattern.

That being said, I'm not exactly sure what the cleanest way to design for the present use case might be. The tokenization logic that's being refactored here very explicitly notes that it is only suitable for spell checks because it breaks other search syntax... so maybe we want to describe this as a spelling tokenization feature rather than just a generic tokenization feature. Would it make sense to have the analyzeQuery() method you propose accept a set of constants as input, to request different specific types of outputs? Then we could start with support for just spelling tokenization, but we could expand to support other operations in future... and perhaps define some exception if you request something that a backend doesn't support.

I'm not exactly excited about this idea, but it seems less daunting as a next step than some of the challenges that arose earlier in this conversation, so it might be a route worth pursuing (while I still think that, in parallel, we should examine some of the possible problems/inconsistencies highlighted above).

@aleksip, does this make your next steps any more clear, or do you think this is not the most productive direction to move in?

EreMaijala:
@demiankatz Mine was a feeble attempt at trying to account for future needs, but if it gets too complicated, a dedicated method for getting terms for spelling would be just as good.

aleksip:
I understand the principle of controlling what is exposed through the search backend, specifically through `Service` and the different backend interfaces. However, for a new VuFind developer the current situation can be a bit confusing, since there do seem to be several exceptions to this pattern.

In addition to bypassing the event-based system, as discussed previously, `SwitchQuery::getLuceneHelper()` returns a `LuceneSyntaxHelper` backend helper class.

`VuFind\Search\Solr\Results` and `VuFind\Search\Solr\InjectSpellingListener` both call `$collection->getSpellcheck()`, although the method is not defined in `RecordCollectionInterface`, to get a `Spellcheck` backend class.

A quick search of `VuFind` module code also seems to bring up several more cases where specific backend implementations and other classes from `VuFindSearch` are used directly. There also might be more method calls like `getSpellcheck()`, where the method is not exposed through `Service` or the different backend interfaces.

I do not find helper classes problematic if they are exposed in a backend independent way through an interface. But as we still have a manageable amount of methods I agree that it might be simpler to just add this method directly to the `Service` class as Ere suggested.

I have modified the PR accordingly, but it still is very much WIP until the best way to proceed is decided, which could also affect the current implementations of at least `LuceneSyntaxHelper`, `Spellcheck` and `SpellingProcessor`.

By the way, I still have not figured out why `AbstractQuery` and even `Query` are used in so many places instead of `QueryInterface`. In `BackendInterface`, for example. Could you shed some light on this?

demiankatz:
Thanks, @aleksip, I think that this is definitely a better and more focused approach to this current problem. I am entirely comfortable with polishing this up and getting it merged.

In the longer term, if we're going to add a lot more methods like this to better centralize the logic, I wonder if it would make sense to refactor the search service so that we use a support method to handle most of the event logic, rather than duplicating a very similar pattern in nearly every method. That's a conversation for another day, but I think the whole search service might be less intimidating if we could separate the event handling from the other business logic. I think a shared method accepting a callback and incoming parameters might make this possible, and significantly tighten up the code.

I'm certainly open to refactoring `SwitchQuery` to do things more elegantly.  It probably makes sense to search the whole code base for references to the BackendManager and see how many of those can be eliminated or revised.

I'm also open to adding some additional interface checks and/or changing typehints to make calls to getSpellcheck more appropriate (and to avoid future static analysis problems). Without digging into the code, I'm not sure how difficult/complex that might be, but I'm happy to discuss further as needed.

Regarding the use of `AbstractQuery` instead of `QueryInterface`, I cannot explain it, and I think it should be changed. Since AbstractQuery implements absolutely nothing, there's no functional difference except that using it reduces flexibility. Specific typehint for `Query` might be meaningful, though, to differentiate from `QueryGroup`. I would have to look more closely at all the examples to determine whether or not that was intentional. But I'd certainly support changing `AbstractQuery` typehints as a first step.

Obviously, there's a lot of work highlighted here, and some of it requires more discussion. I think we're better off if we can get it done in small pieces so we can make incremental progress without getting overwhelmed. Thus, I'm definitely not proposing that you do extra work here beyond what is necessary to complete the functionality you need. If you are interested in working on any or all of the items discussed above, I'd encourage you to open additional pull requests. If you think some of these things are important but you do not have time or need more help to make progress on them, feel free to open JIRA tickets and I will assist as time permits.

Thanks again for your work on this and your insightful analysis of the code -- I really appreciate it!

demiankatz:
Some more food for thought, @aleksip and @EreMaijala: while searching through JIRA, I was reminded of [VUFIND-924](https://vufind.org/jira/browse/VUFIND-924), which proposes using the command pattern as a way of allowing greater flexibility in the search backend without a proliferation of new methods. The discussion there even suggests using this in some of the existing problem areas of the code -- I just never got around to implementing it. I think this might be worth considering now. What do you think? If we all agree on the approach, maybe we can figure out a strategy for working together to get it implemented. I'm willing to put some time into it, but any help would be appreciated.

EreMaijala:
@demiankatz I'm not sure I understand the benefit. Maybe I'm missing something, but to me it looks like the command pattern would move backend-specific logic further away from the backend and would make the backend merely a proxy. I'm not sure if that's the right direction, but as I said there might be something I'm not quite grasping here.

demiankatz:
@EreMaijala, maybe the idea I'm getting from VUFIND-924 does not exactly align with David's original intentions, but my thought is this:

The current problem we have is that in order to do anything with the search backends through the search service, we need to add a method to the search service and write fairly redundant code that essentially proxies a specific function call through the search service. This doesn't scale well if we want to have dozens or hundreds of actions we can take, and it gets really ugly if all the backends accept different, incompatible commands. (Obviously, we want to standardize and avoid special-purpose things as much as possible, but sometimes it's going to be unavoidable).

I'm envisioning the command pattern as a way of having a single method in the search service which can handle all of these different possibilities, and moving the more specific logic into separate command classes that get passed to the search service.

So, for example, in the current pull request, rather than having to create a whole new analyzeQuery() method on the search service, we could instead create an AnalyzeQueryCommand that could get passed to an invokeCommand() method on the search service, which could look something like this:

<pre>
    public function invokeCommand(string $backend, CommandInterface $command): array
    {
        $context = __FUNCTION__;
        $args = compact('backend', 'command', 'context');
        $backendInstance = $this->resolve($backend, $args);
        $args['backend_instance'] = $backendInstance;

        $this->triggerPre($backendInstance, $args);
        try {
            $response = $command->execute($backendInstance);
        } catch (BackendException $e) {
            $this->triggerError($e, $args);
            throw $e;
        }
        $this->triggerPost($response, $args);
        return $response;
    }
</pre>

and then the Command class looks something like this:

<pre>
class AnalyzeQueryCommand
{
    protected $query;
    protected $params;

    public function __construct(QueryInterface $query, ParamBag $params = null)
    {
        $this->query = $query;
        $this->params = $params ?: new ParamBag();
    }

    public function execute($backend)
    {
        if (!($backend instanceof QueryAnalysisInterface)) {
            throw new BackendException(
               get_class($backend) . " does not support analyzeQuery()"
            );
        }
        return $backend->analyzeQuery($this->query, $this->params);
    }
}
</pre>

Obviously, that's just meant as an illustration, and there are a lot of details that might be different and points worth discussing more deeply (like how we actually pass arguments and return values around, how we name things, etc., etc.). But I really like the way that this separates the ugly event handling stuff from the backend-specific logic while still allowing us to fully take advantage of the event system... and without bloating the service itself.

aleksip:
@demiankatz Looks like the Command classes would all end up being quite similar? Would they be in `VuFind` or `VuFindSearch`?

demiankatz:
@aleksip, I think the commands might end up looking similar, or we might decide that we could use a very generic command that could cover many use cases (like a CallMethodCommand, for example, which could just take a method name and arguments and effectively allow direct backend access through the search service, with exceptions thrown if the requested method is undefined). That may or may not be a good idea -- I'm just brainstorming possibilities without thinking about best practices yet. But my point is that depending on our approach, there are a variety of ways we could break things down. My inclination would be to put general-purpose, reusable commands in the VuFindSearch namespace, since they would be a part of the search system... but I could imagine very specialized commands that might live in the VuFind namespace because they would not be useful outside of the context of VuFind. I could also imagine having some commands live inside the namespaces of specific backends because they would not be useful outside of that context -- i.e. if for some reason we came up with a command that would only ever apply to Solr-specific backends, that could live in VuFindSearch\Backend\Solr\Command, rather than VuFindSearch\Command.

aleksip:
@demiankatz Feels to me that there would be some overlap between the feature interfaces and commands? My initial feeling is also that it would be nicer if the caller only had to know about interfaces instead of concrete implementation classes.

Could the `invokeCommand()` in `Service` accept interface class and method names as arguments and if applicable call the backend directly, without the need for separate command classes?

demiankatz:
@aleksip, I don't exactly see the feature interfaces and commands overlapping -- the interface is a way that a backend can advertise that it supports a specific feature, and a command class could be a consumer of that feature, using the interface for validation. If we're "duck typing" by just calling methods if they exist, then the interface becomes potentially unnecessary, but our code also becomes less explicit and potentially harder to refactor/test/debug. (I haven't settled on an opinion about how we should proceed yet, so just laying out pros and cons here).

As for whether we could have a backend method that just takes a method name and arguments instead of a command object, that's certainly possible, but I think there are going to be situations where a command class would be justified, because there would be more complex logic. For example, imagine if we refactored the current VuFindSearch\Service::retrieveBatch() method to RetrieveBatchCommand -- that's obviously going to need a place for all the logic to live. For this reason, I view the "just call a method" approach as a subset of the "execute a command" approach.

In other words, I'm pondering the idea of a refactored backend where everything runs through a single method that executes commands, and all the other methods are just convenience wrappers around that single method. That gives us the benefit of consistent event handling and centralized logic, but doesn't necessarily require outside users to directly interact with command objects. In that kind of scenario, we could easily have a callMethod() method which creates a CallMethodCommand and passes it along to the centralized invokeCommand() method. (I also realize that we can't just refactor all the existing functionality to follow this pattern without introducing significant breaking changes for listeners, etc. -- but just pointing out that this is a possible approach we could end up with).

aleksip:
@demiankatz I have made a quick implementation of an internal `callMethod()` without the command object creation, and a refactored `QueryAnalysisInterface`, which can now have separate methods.

This could be further refactored to include command object creation. But if the command object creation would be done in `Service`, and we would still create the convenience wrapper methods, perhaps an alternative to the command objects could be to just put the possible extra logic in the wrapper?

demiankatz:
Thanks, @aleksip, this looks like a good start on a proof of concept -- it's definitely more general-purpose than the previous iteration, though I think more discussion of the possible role of command objects is worthwhile. I'll be interested to hear @EreMaijala's thoughts, too.

My is was that, in the long term, the path forward would not be to create a wrapper method for everything inside the search service, but rather to only use wrappers for backward compatibility and/or for the most commonly-used functionality. We could still potentially create helper code that builds and sends commands but lives at a higher level than the search service itself. My long-term goal is to allow the search service to focus on its core job of routing commands and triggering events, while spreading out the rest of the logic to more appropriate places, to reduce bloat and mental overhead. Right now, the search service is a pretty intimidating (and perhaps baffling) piece of code on first examination, but I think we might be able to significantly improve that situation with a bit of refactoring.

aleksip:
@demiankatz If I have understood everything correctly, adding a new feature to a backend could involve 1) adding a method to an existing or new feature interface, 2) possibly creating or modifying a command class for calling that method, and 3) possibly adding a wrapper method or making other modifications to the helper code living above the search service.

Another case would be implementing something new in a command class only. This makes me wonder about the caller's perspective: what is the exposed API? Seems like it would consist of at least three different things: 1) the helper code living above the search service, 2) the search service and 3) the backend interfaces (if direct backend calls are allowed?).

So it does seem that these changes might not necessarily reduce complexity. It is also still not 100 % clear to me what should go and what should not go to `VuFindSearch`. After all, there is a lot of backend-specific code in `VuFind` too. 

The original motivation for this PR is to implement a common way to process backend-specific query strings. Looking at the [documentation](https://vufind.org/wiki/development:architecture:search_service), the main design goal for the search service was to "separate re-usable classes for *interacting* with a specific remote system" (emphasis mine). Do classes that process query strings fit that description?

This brings to mind what was your very first question: "whether this really needs to be a backend method". I am more than happy to backtrack and continue from a [previous question](https://github.com/vufind-org/vufind/pull/1886#issuecomment-803942116) of mine regarding the discussed alternative.

demiankatz:
@aleksip, I think your assessment is correct. Maybe it's easier to conceptualize this as a set of options, ordered by preference.

The most preferred approach is to offer a feature which is universally supported by all backends. This includes things like 'search' and 'retrieve.' These methods should be exposed directly from the search service, and that should be the preferred way of calling them. (We could implement them internally as commands if that makes the code cleaner, but that is an implementation detail that the caller should not care about).

The next most preferred approach is to offer a feature interface, so we can identify functions supported by some, but not all, backends. This functionality is currently exposed directly through the search service, and maybe we should continue to do that, or maybe we should create a more generic way of requesting access to these features. I don't have a strong opinion about this, except that I'm concerned that the search service will eventually grow to a size where it is harder to manage and understand than it already is.

The least preferred approach would be to send a command directly to the search service. This is not a technique that I would want to see used widely through the VuFind code, but I think it offers powerful flexibility for local customizations and for experimenting with immature functionality that might eventually become a feature interface or standard feature but which needs more use cases or implementations to properly evaluate. I think sending a command to the search service, though perhaps not ideal, offers us a better alternative to either completely working around the service (an existing problem you identified) or adding lots of very backend-specific functionality to a service which is designed to be a generic gateway.

So I'm not really proposing that we drastically change the exposed interface of the search service, but I'm trying to find a way to make it more flexible and extensible for future growth, and to solve some of the problems you have identified.

You're also correct that it's not clear-cut what goes in VuFind and what goes in VuFindSearch. I can only suggest that this may become more clear if we actually start to try some refactoring.

As for your question about whether a QueryHelperManager/QueryHelperRegistry might provide an alternate solution to the problem at hand, I think that's certainly a possibility. Part of the reason I raised concerns about all of this in the first place is that it doesn't seem like we need to full weight of the search service to get at a helper class that processes text in a consistent way. If there's a cleaner way to do that without touching the search service at all, I'd be in favor of it -- as long as it doesn't lead to event-system inconsistencies or any of the other concerns that have been raised through this conversation. If you have ideas about that and would like to put together an alternate proposal, I'd be happy to discuss it... and either way, the work done here so far is definitely not wasted, because I think we are moving toward a significant improvement to the search service. I'm also willing to open a separate PR with a command-based proposal for everyone's consideration, if that would be helpful (though at present I'm not sure exactly when I'll have time, as the next few weeks are going to be very busy).

aleksip:
This PR now depends on #1967.

demiankatz:
@aleksip, now that #1967 has been merged, are you ready to finish this up? I notice that there is a build failure here; does this just need to have dev merged into it so that all the classes exist? (I tried to do the merge myself, but I don't have permission to push to your repository).

demiankatz:
@aleksip, just wondering where we are on this -- seems like it may have fallen off the radar. :-)

aleksip:
@demiankatz Unfortunately I have had to postpone work on this due to more urgent tasks, but I still hope to continue at some point in the not-too-distant future. If you wish I can close this PR for the time being, and re-open when I pick it up again?

Sorry for taking so long to reply!

demiankatz:
@aleksip, no worries! You can feel free to leave this open if it's a useful reminder -- I just like to check in on pull requests every few months if they don't move, just to be sure they're still needed.

demiankatz:
@aleksip, with the VuFind 9.0 release around the corner, I was looking at old pull requests and noticed that you have a few (including this one) that haven't moved in quite a while. Is there any chance you could take a look and see if any are worth either closing or moving forward? I'm also happy to keep waiting, but I think we're due for another review as it's been a year or more in some cases. :-)

Thanks!

aleksip:
@demiankatz I will close this for now because it is related to functionality we have stopped developing. Will reopen if we start working on it again.

demiankatz:
Thanks, @aleksip!

PR #1887, opened 2021-03-18
Add MultiBackend support for getHoldLink method.

demiankatz:
TODO
- [x] Run test to be sure this does not interfere with regular placeHold() / getMyHolds() functionality.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
@@ -590,6 +590,29 @@ class MultiBackend extends AbstractBase implements \Laminas\Log\LoggerAwareInter
     }
 
     /**
+     * Get Hold Link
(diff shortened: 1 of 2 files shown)


demiankatz:
@EreMaijala, a user of the basic Koha driver was having trouble with MultiBackend because it didn't support the getHoldLink method here. It seemed straightforward to implement, so I added it here, and this solved that user's problem. However, I'm not entirely sure whether this might have side effects and interfere with regular hold functionality. Do you have any thoughts on that, or the ability to test easily? If you cannot easily test, I can set up a test scenario here when I have a bit more time.

EreMaijala (approved):
As far as I can see getHoldLink is only used if placeHold is unavailable. I can't see this change interfering, so this looks good to me!

demiankatz:
Thanks, @EreMaijala, I agree -- merging now.

PR #1889, opened 2021-03-19
Fix VuFind case

xmorave2:
Just some typo fixes


demiankatz:
Thanks for catching these!

PR #1891, opened 2021-03-22
Set of email verification related tweaks

EreMaijala:
This is a set of individual commits for improving email verification.

First three are fixes to existing functionality, and the last one adds a flash message to the profile page to reming the user about a pending verification.

### DIFF
--- a/module/VuFind/src/VuFind/Auth/Manager.php
+++ b/module/VuFind/src/VuFind/Auth/Manager.php
@@ -583,9 +583,12 @@ class Manager implements \LmcRbacMvc\Identity\IdentityProviderInterface
         // Depending on verification setting, either do a direct update or else
         // put the new address into a pending state.
         if ($this->config->Authentication->verify_email ?? false) {
-            $user->pending_email = $email;
+            // If new email address is the current address, just reset any pending
+            // email address:
+            $user->pending_email = ($email === $user->email) ? '' : $email;
         } else {
             $user->updateEmail($email, true);
+            $user->pending_email = '';
         }
         $user->save();
         $this->updateSession($user);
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -583,9 +583,12 @@ public function updateEmail(UserRow $user, $email)
         // Depending on verification setting, either do a direct update or else
         // put the new address into a pending state.
         if ($this->config->Authentication->verify_email ?? false) {
-            $user->pending_email = $email;
+            // If new email address is the current address, just reset any pending
+            // email address:
+            $user->pending_email = ($email === $user->email) ? '' : $email;
```
demiankatz, line 588: @EreMaijala, I simplified this with a ternary operator. Any objections?

EreMaijala: Nice, no objections!

module/VuFind/src/VuFind/Controller/MyResearchController.php
```diff
@@ -2258,4 +2246,42 @@ protected function collectRequestAccountStats(array $records): ?array
...
+     *
+     * @param \VuFind\Db\Row\User $user User
+     *
+     * @return void
+     */
+    protected function addPendingEmailChangeMessage($user)
+    {
+        if (!empty($user->pending_email)) {
+            // Check if the message has already been added
+            foreach ($this->flashMessenger()->getInfoMessages() as $message) {
```
demiankatz: Is this necessary? The view helper should already be deduplicating flash messages [here](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/View/Helper/Root/Flashmessages.php#L86). Rather than doing a lot of specific work here, I wonder if we need to fix/improve a pre-existing generic mechanism if things are not working as expected without this.

EreMaijala: Ah, I didn't know about that. Then I believe it's a useless piece of code. I'll remove it.

EreMaijala: @demiankatz Fixed.

demiankatz:
Looks good to me. Rebasing now!

PR #1892, opened 2021-03-22
Add properties definitions to Aleph driver

xmorave2:


demiankatz:
Thanks, this is very helpful!

PR #1355, opened 2019-03-21
PAIA notifications in ILS\Driver\PAIA

domerz:
Incorporates PHP functionality for new PAIA functions concerning notifications (i.e. system messages).
See http://gbv.github.io/paia/paia.html#notifications

PR #1894, opened 2021-03-23
Add missing method to PAIA ILS driver

xmorave2:
I found that PAIA driver calls missing method 'getPaiaNotificationsId'. We are not using it, and don't know how it should work, but to have it not missing, I implemented it just to pass the param messageId.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/PAIA.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/PAIA.php
@@ -2017,6 +2017,18 @@ class PAIA extends DAIA
     }
 
     /**
+     * Get notification identifier from message identifier
+     *
+     * @param string $messageId Message identifier
+     *
+     * @return string
+     */
+    protected function getPaiaNotificationsId($messageId)
+    {
+        return $messageId;
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks for catching this, @xmorave2! It looks like this was added in #1355 by @domerz, and approved by @ndege. This code was added for its future usefulness but was not being used for anything yet, which explains why none of us noticed the omission. Since this comes from the University of Leipzig, I'm also asking for a review from @lahmann. I'm hoping that somebody from University of Leipzig can comment on whether the implementation here is correct, or whether there is something more specific that they can share.

domerz:
That is an obvious mistake from my older PR. Thanks for the fix. I have no objections against it.

For your information: we use that function in our inherited version to add some pre-configured prefixes to the identifier. Having the standard-version return the identifier unchanged, however. seems a sensible approach

demiankatz:
Great! Thanks for the very quick response, @domerz! I'll merge this now.

demiankatz:
Since this is a bug fix, I've backported it to the release-7.1 branch and merged it from there to dev; this means the fix will be included in release 7.1.1.

PR #1895, opened 2021-03-23
MARC support for alternative scripts

EreMaijala:
Related to #1888.


module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
```diff
@@ -951,4 +951,57 @@ public function getCleanNBN()
...
+
+    /**
+     * Get the full title of the record in alternative scripts.
+     *
+     * @return array
+     */
+    public function getTitlesAltScript(): array
+    {
+        return $this->getMarcReader()
+            ->getLinkedFieldsSubfields('880', '245', ['a', 'b']);
```
demiankatz, line 963: It may be impractical due to the way the logic is currently spread across multiple traits (and perhaps this is a sign that the design could be improved), but I wonder if it would be useful to create properties for holding the MARC definitions, so if we redefine which fields/subfields are used for any of these things, we only have to change it in one place in order to affect both the "regular" and 880-based methods. (I'm thinking, in particular, about things like Unimarc support, which might be easier if we could simply override some properties instead of having to rewrite all the methods). Obviously, I'm happy to merge this as-is when you are ready and think about that larger issue later... but this brought it to mind.

EreMaijala: It would make sense, but I'm afraid it won't be easy. I suppose there could be a definition file MARC 21 and architecture that would allow you to implement e.g. UNIMARC support similarly. But there are cases that depend on indicator values, fields that are alternatives etc. that make such definition difficult to build. For simple cases such as "result list title subfields" and "record view title subfields", that would work, but anything more complex, I'm not sure.

I'd like to get feedback from @bbusenius before marking this ready for merging.

demiankatz: Agreed, a simple map couldn't solve all problems... but even if it got us 80% of the way there, it would make it less intimidating to create a SolrUnimarc subclass, since we would only have to override methods with significant functional differences. In any case, a conversation for another day! :-)

I'll await feedback from @bbusenius before proceeding further.

bbusenius: @EreMaijala this is beautiful! I feel like you built a lot of versatility into this with things like the `concat` argument for `getLinkedFieldsSubfields` and the ability to simply get all subfields or specify them in the `getLinkedFields`. I think those subtleties might prove to be useful. I can see how we could override the methods in `MarcAdvancedTrait` if we wanted to change any of the defaults. I can't speak much to what @demiankatz is saying about UNIMARC support but from what I'm seeing here, I feel like this would work for us.

demiankatz:
@EreMaijala / @bbusenius, do either of you have any example MARC records with 880's that we could throw into the `test/data` directory so we have some examples in our standard test set to use for testing/examining this functionality? Do you think any styling is needed for the alt-title class? Any other obstacles to merging this? (I'd like to give it a more thorough review before merging, but I want to be sure it's reasonably stable first). In any case, thanks to both of you for moving this forward!

EreMaijala:
@bbusenius Thanks, that's encouraging.
@demiankatz I'll work on tests a bit.

bbusenius:
@demiankatz I can certainly come up with some MARC records with 880s. How many records would be useful here? Regarding the styling of the `alt-title` class, we do not give it any special styling, however, I was really happy to see the class since it would make it easy if we ever wanted to. For stock VuFind, I don't feel strongly about it but I'm inclined to say that as long as the classes exist, we're good.

EreMaijala:
I added a record along with some more tests yesterday. I think this is ready to go unless we want to add something more.

demiankatz:
@bbusenius, we don't need a huge number of sample records. One would be better than none, and 5-10 would probably be more than adequate. Obviously, if there are a lot of different edge cases to test, or a lot of different fields we want to look at, it might make sense to increase that number... but for starters I'd be happy to have anything that's easy for you to put together.

demiankatz:
Sorry, didn't see @EreMaijala's comment before replying to @bbusenius. If @bbusenius wants to send more test records to cover additional edge cases, that's great, but if not, I'm sure what we have here already is fine as a starting point. I'm swamped with meetings all day today, but I'll try to find time for a more thorough review of this work as soon as practical.

demiankatz:
Looks like this got lost in the shuffle a bit during my vacation. I see that there's a small conflict in one of the tests now. @EreMaijala, would you mind resolving that what time permits? After that, I'll do a final review and merge, unless you or @bbusenius think there's a reason to wait.

EreMaijala:
@demiankatz Conflict fixed. I also added a couple more tests to keep coverage at 100%.

module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
```diff
@@ -951,4 +951,57 @@ public function getCleanNBN()
...
+    }
+
+    /**
+     * Get the full title of the record in alternative scripts.
+     *
+     * @return array
+     */
+    public function getFullTitlesAltScript(): array
```
demiankatz, line 972: This method has exactly the same comment as getTitlesAltScript() above.  Maybe the comment should read "Get the full title (including section/part information) ...." I'm also not seeing a non-AltScript version of this method anywhere. Should we add getFullTitle() for consistency? That might help move us toward simplifying general title handling in the future, too, though maybe I'm getting ahead of myself here.

module/VuFind/src/VuFind/Marc/MarcReader.php
```diff
@@ -277,4 +277,161 @@ public function getFieldsSubfields(string $fieldTag, array $subfieldCodes,
...
+
+    /**
+     * Return first subfield with the given code in the internal MARC field
+     *
+     * @param array  $field        Result from MarcReader::getFields
+     * @param string $subfieldCode The MARC subfield code to get
+     *
+     * @return string
+     */
+    protected function getInternalSubfield(array $field, string $subfieldCode
```
demiankatz, line 429: This function confused me a little bit. I think I understand that this is for working with the internal, abbreviated array, which differs from the more verbose public interface exposed by the class... but the documentation for the $field parameter is identical for this function and for the public `getSubfield()` version. Is one of them incorrect?

module/VuFind/src/VuFind/Marc/MarcReader.php
```diff
@@ -277,4 +277,161 @@ public function getFieldsSubfields(string $fieldTag, array $subfieldCodes,
...
+    public function getLinkedField(string $fieldTag, string $linkedFieldTag,
+        string $occurrence = '', ?array $subfieldCodes = null
+    ): array {
+        $results
+            = $this->getLinkedFields($fieldTag, $linkedFieldTag, $subfieldCodes);
+        if ('' === $occurrence) {
+            return $results[0] ?? [];
+        }
+        foreach ($results as $field) {
+            if ($occurrence === $field['link']['occurrence']) {
```
demiankatz: If you just add `|| empty($occurrence)` here, you can eliminate the '' special case above and get the same behavior... though perhaps it is less clear. Just a thought!

module/VuFind/src/VuFind/Marc/MarcReader.php
```diff
@@ -277,4 +277,161 @@ public function getFieldsSubfields(string $fieldTag, array $subfieldCodes,
...
+        }
+
+        return $result;
+    }
+
+    /**
+     * Return an array of all values extracted from the specified linked
+     * field/subfield combination.  If multiple subfields are specified and $concat
+     * is true, they will be concatenated together in the order listed -- each entry
+     * in the array  will correspond with a single MARC field.  If $concat is false,
```
demiankatz: Extra whitespace after "array" here.

module/VuFind/src/VuFind/Marc/MarcReader.php
```diff
@@ -277,4 +277,161 @@ public function getFieldsSubfields(string $fieldTag, array $subfieldCodes,
...
+     * the return array will contain separate entries for separate subfields.
+     *
+     * @param string $fieldTag       The MARC field that contains the linked fields
+     * @param string $linkedFieldTag The linked MARC field tag to get
+     * @param array  $subfieldCodes  The MARC subfield codes to get
+     * @param bool   $concat         Should we concatenate subfields?
```
demiankatz: Do we need separate $concat / $separator parameters, or could we assume that if you provide a separator, you want concatenation? Maybe that's too obtuse, but it seems like this might be more verbose than necessary.

demiankatz (commented):
@EreMaijala, I'm happy to merge this whenever you think it's ready (just let me know). I did give it a more thorough review, and left these comments -- many of which are just intended as brainstorming and may be best ignored. A few are probably worth acting on, though!

module/VuFind/src/VuFind/Marc/MarcReader.php
EreMaijala: No, we don't need it, but it's just been so that it has the same parameters as getFieldsSubfields, which in turn derived its parameters from MarcReaderTrait::getFieldArray. I'll change the interface of MarcReader while keeping the trait intact, ok?

demiankatz: Once again, I think I'm offering you advice but I'm actually criticizing past me. ;-) I'll leave it entirely to your discretion whether it's worth simplifying this or easier to leave it consistent with the trait. I agree that whatever you do here, we shouldn't mess with the trait -- at least not without opening a whole separate PR to discuss it, which is unlikely to be worth the effort at this stage!

EreMaijala: It's ok to revisit your old decisions. :) I've now simplified the MarcReader methods while leaving the trait alone. I also added a couple of tests to cover the different cases and fixed non-concatenation results of getLinkedFieldsSubfields.

demiankatz:
I'm merging this now! I haven't heard from @bbusenius in a while (guessing that he's busy with other things at the moment), but if he has further feedback, it can be addressed through the completion of #1888.

bbusenius:
Ah yes, sorry. I've been on other things. This looks fantastic though. Thank you @EreMaijala and @demiankatz!

PR #1896, opened 2021-03-23
PHPStan configuration and basic fixes

xmorave2:
This PR contains phing and travis configuration for running PHPStan checks on codebase and also fixes for level 0 of PHPStan analysis

TODO
- [x] Integrate into Jenkins when merging


demiankatz:
Thanks for this, @xmorave2! I've added a TODO to myself to hook this up with Jenkins when we merge it. I plan to merge many of your fixes to dev in advance of this full PR, which I think will make it easier to review what remains here. We may also want to discuss whether this complements or supersedes the work with Psalm in #1764. I'm not opposed to using multiple tools to maximize quality, but if one or the other is clearly better, it might make sense to choose. I'll add a more detailed review here once I've whittled things down by merging fixes.

xmorave2:
@demiankatz From my point of view we could take this as complement to Psalm to maximize code quality. I would not say the one tool is better then the other one, both have their pluses and both are actively developed. Both are extensible, but only phpstan has plugin specific for laminas framework (but I don't try it yet, so maybe, we don't need it)

For me the PHPStan has more clear issues description. If we adopt the PHPStan I am ready to continue on fixing code for more strict rules.

demiankatz:
Great, thank you again, @xmorave2, and I agree that the two tools will likely be complementary. I'll work with you on getting this finished up first since I think it is going to be less work than Psalm, and once we merge the fixes from this work into the Psalm branch, we will see what remaining issues are reported there.

xmorave2:
Great, thanks @demiankatz

composer.json
```diff
@@ -86,7 +86,8 @@
         "phpunit/phpunit": "9.5.2",
         "sebastian/phpcpd": "6.0.3",
         "squizlabs/php_codesniffer": "3.5.8",
-        "dmore/chrome-mink-driver": "^2.7"
+        "dmore/chrome-mink-driver": "^2.7",
+        "phpstan/phpstan": "^0.12.82"
```
demiankatz: It might be worth moving this to the correct alphabetical position. I also like to pin specific versions, so if a new version of a tool is released and detects a new issue, it doesn't inexplicably break the build.

module/VuFind/src/VuFind/Search/UrlQueryHelper.php
```diff
@@ -76,7 +76,7 @@ class UrlQueryHelper
      * on the contents of $query to $urlParams (true) or are they already there
      * (false)?
      */
-    public function __construct(array $urlParams, AbstractQuery $query,
+    final public function __construct(array $urlParams, AbstractQuery $query,
```
demiankatz: Why make this final?

module/VuFind/tests/integration-tests/src/VuFindTest/Auth/DatabaseTest.php
```diff
@@ -38,7 +38,7 @@
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org Main Page
  */
-class DatabaseTest extends \PHPUnit\Framework\TestCase
+final class DatabaseTest extends \VuFindTest\Integration\MinkTestCase
```
demiankatz: This seems to be changing to an incorrect base class; also, why make it final?

module/VuFind/tests/integration-tests/src/VuFindTest/Auth/ILSTest.php
```diff
@@ -40,7 +40,7 @@
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org Main Page
  */
-class ILSTest extends \PHPUnit\Framework\TestCase
+final class ILSTest extends \VuFindTest\Integration\MinkTestCase
```
demiankatz: Another inappropriate base class change.

module/VuFind/tests/integration-tests/src/VuFindTest/Auth/ShibbolethTest.php
```diff
@@ -42,7 +42,7 @@
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org Main Page
  */
-class ShibbolethTest extends \PHPUnit\Framework\TestCase
+final class ShibbolethTest extends \VuFindTest\Integration\MinkTestCase
```
demiankatz: ...and another.

module/VuFindApi/src/VuFindApi/Controller/ApiController.php
```diff
@@ -40,7 +40,7 @@
  */
 class ApiController extends \VuFind\Controller\AbstractBase
 {
-    use ApiTrait;
+    //use ApiTrait;
```
demiankatz: Why comment this out? I think we need this trait, since it sets up event hooks for API security.

demiankatz (changes requested):
Thanks again for all of this work, @xmorave2 -- it's been tremendously helpful. I committed most of the fixes from here to the dev and release-7.1 branches, reducing this just to the actual PHPStan setup, and a few outstanding changes that I had questions about. In a few cases, I solved problems differently, but the vast majority of your work was carried through as-is. Hopefully it won't take too much more work to figure out the issues highlighted below, and then perhaps we can merge this!

tests/phpstan.neon
```diff
@@ -0,0 +1,10 @@
+parameters:
+    paths:
+        - %currentWorkingDirectory%/module
+    tmpDir: %currentWorkingDirectory%/.phpstan_cache
+    ignoreErrors:
+        - '#Class VuFind\\Hierarchy\\Driver\\Hierarchy[a-zA-Z0-9]+ not found#'
+        - '#Class VuFind\\Content\\[a-zA-Z0-9]+\\SyndeticsPlus not found#'
+        - '#Class VuFind\\Content\\[a-zA-Z0-9]+\\Amazon[a-zA-Z0-9]* not found#'
+        - '#Function sybase_[a-z_]+ not found#'
```
demiankatz: One other question: does PHPStan support ignoring errors through annotations in the code? I wonder if it would be better to annotate the offending fiels rather than globally ignoring specific errors. That way, if problems get resolved or go away (for example, if we get rid of all sybase function calls in the future), we don't have to remember to come back here and change this.

composer.json
xmorave2: It does make sense, I'll fix it

module/VuFind/src/VuFind/Search/UrlQueryHelper.php
xmorave2: All the finals are because of using `new static()`, see https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static

another option is to define interface which defines constructor, or ignore this error.

module/VuFind/tests/integration-tests/src/VuFindTest/Auth/DatabaseTest.php
xmorave2: The incorrect class comes probably from some merge, I'll fix it

xmorave2: The base class is changed because of UserCreationTrait, which calls methods `findCss`, `findCssAndSetValue` and `snooze`. These methods are defined in MinkTestCase class

tests/phpstan.neon
xmorave2: Yes, PHPStan supports this. I could rework this and use inline annotations to ignore errors. But PHPStan also has feature to report not used ignoreErrors configuration, so when problem is resolved somehow, You'll see it in analysis report.

module/VuFind/src/VuFind/Search/UrlQueryHelper.php
demiankatz: I'll look into it when I'm back in the office. Thanks for the explanation!

module/VuFind/tests/integration-tests/src/VuFindTest/Auth/DatabaseTest.php
demiankatz: I can probably refactor to solve this. I'll try as soon as time permits.

tests/phpstan.neon
demiankatz: Good to know! I'll leave it to your discretion, then. If there are errors that we might really want to know about (like use of deprecated functions), annotations still might be better... but for things that are widespread and intentional, keeping them here makes sense!

xmorave2: I moved sybase_* related errors ignoring into annotations to source code, when/if the LBS4 driver is removed the problem diseappers completely

module/VuFindApi/src/VuFindApi/Controller/ApiController.php
xmorave2: But the permission checking is done only in SearchApiController (and Search2ApiController and WebApiController), not in ApiController itself... the only thing which is used from ApiTrait is onDispatch method, but do we really need it here? I am not sure, maybe @EreMaijala would know...

The problem found by PHPStan was that it use constant self::STATUS_ERROR, which is defined by ApiInterface, but not in ApiController...

xmorave2: So I think we should remove ApiTrait from ApiController, but if onDispatch method is still needed here, we should refactor it somehow...

EreMaijala: OnDispatch is needed at least for the CORS headers. Otherwise any external Swagger or another in-browser OpenAPI generator would be unable to access the API definitition. The easy, though not particularly pretty, fix would be to move STATUS_OK and STATUS_ERROR to ApiTrait as static variables like so:

```
trait ApiTrait
{
    // define some status constants
    public static $STATUS_OK = 'OK';       // good
    public static $STATUS_ERROR = 'ERROR'; // bad
...
        if (!$auth->isGranted($permission)) {
            return $this->output([], self::$STATUS_ERROR, 403, 'Permission denied');
        }
```

This wouldn't be back-compatible if we get rid of the consts in ApiInterface, but I'd be okay with that.

demiankatz: Could we also solve the problem by replacing `self::STATUS_ERROR` with the more explicit `ApiInterface::STATUS_ERROR`? I don't think we really need to reference the constant using self in this instance.

xmorave2: Thans @EreMaijala for clarification, and @demiankatz for solution ;) The other solution could be to create another trait and move the onDispatch method there, then use this new trait in ApiController and both in SearchApiController, but your solution is easier

demiankatz:
Thanks again, @xmorave2, for all the work on this! I've incorporated the remaining fixes into the dev branch (some with minor adjustments, like refactoring of the UserCreationTrait to simplify matters), and I think this is now just about ready to merge. I'm going to be on vacation until April 6, so I am not going to merge it until after I get back -- but I hope to do it shortly thereafter, along with the necessary adjustments to use this in our Jenkins instance.

demiankatz:
I've gotten clearance to remove the LBS4 driver, so I've gone ahead and done that to further simplify this PR. I'll look at Jenkins integration next; once that is figured out, we're ready to merge. Thanks again for your help!

demiankatz:
I think I've taken care of the Jenkins integration; merging now to see how it goes! Now the question is if/when to start working on higher PHPStan levels. :-)

PR #1897, opened 2021-03-23
Support all flash namespaces

welblaud:
There were two namespaces missing in the array of supported namespaces. This adds the possibility to have additional styling for the WARNING namespace (typically, yellow) and ads the DEFAULT namespace, which has no styling at all.


demiankatz:
@welblaud, for some reason, this PR was showing changes from your earlier Aleph fix as well as the work you just started here. I merged the latest dev branch into it, and now it's showing only the relevant changes. It's possible that you didn't create a fresh branch of dev when you started work on this -- something to watch out for in future, though everything is fine here for now.

I am going to ask @crhallberg for a review in case he has any thoughts on the styling implications.

demiankatz:
@crhallberg, just to fill you in on the background here: @welblaud discovered that if he set flash messages in the 'default' or 'warning' namespaces, they simply never displayed, due to the way VuFind's flash message helper is built. Adding the namespaces to the list makes the messages display. I'm pretty sure warnings are fine as-is, because there are already classes to display them appropriately. I'm less certain about how/whether we should handle default flash messages, because I'm not sure what styling they might need, or what context they would be used in. Any thoughts?

module/VuFind/src/VuFind/View/Helper/Root/Flashmessages.php
```diff
@@ -78,7 +78,7 @@ protected function getClassForNamespace($ns)
     public function __invoke()
     {
         $html = '';
-        $namespaces = ['error', 'info', 'success'];
+        $namespaces = ['default', 'success', 'warning', 'error', 'info'];
         foreach ($namespaces as $ns) {
             $messages = array_merge(
                 $this->fm->getMessages($ns), $this->fm->getCurrentMessages($ns)
```
crhallberg, line 90: It looks like the messages will appear in that order if there are multiple. I don't know any cases where that would happen but is this the order we'd want them to appear in?

crhallberg (changes requested):
We're going to need some styling for the `alert-default` class this creates. Should be fairly simple if we borrow styles from elsewhere in Bootstrap:

```less
.alert-default {
    border-color: @well-border;
    background-color: @well-bg;
}
```

module/VuFind/src/VuFind/View/Helper/Root/Flashmessages.php
demiankatz: Good point, @crhallberg! I would probably list these in order of severity, so maybe `['error', 'warning', 'info', 'success', 'default']` would be better.

It might also make sense to move this from a local `$namespaces` variable to a `$this->namespaces` class property, so it would be easier to reorder or otherwise change in a subclass without rewriting code.

demiankatz:
Thanks for the review, @crhallberg! @welblaud, let us know if you need help with the style adjustments!

module/VuFind/src/VuFind/View/Helper/Root/Flashmessages.php
janhejzl: Ok, I will reorder them. This order is the same as in the FlashMessenger class but it seems reasonable to stick to the 'severity' scale.

demiankatz:
Thanks, @welblaud; merging now!

PR #1898, opened 2021-03-24
SierraRest: Add support for patron authentication methods depending o…

EreMaijala:
…n available API version.

Also adds support for password-less patronLogin by passing null as the password. Testing needed.

### DIFF
--- a/config/vufind/SierraRest.ini
+++ b/config/vufind/SierraRest.ini
@@ -62,8 +62,8 @@ password_field = "pin"
 ; useful when populating your cat_username/cat_password values automatically through
 ; an institutional login. Legal values are "name" (the user's full name) or "email"
 ; (the user's email address). "email" is strongly preferred since it is more likely
-; to be a unique identifier. When activated, this setting will disable PIN checks;
-; this setting is off by default.
+; to be a unique identifier. When patron_validation_field is activated, PIN checks
+; will be disabled. The patron_validation_field setting is disabled by default.
 ;patron_validation_field = "email"
 
 ; This section is used to define library codes and named values which are used by the
--- a/config/vufind/SierraRest.ini
+++ b/config/vufind/SierraRest.ini
@@ -23,21 +23,25 @@ client_secret = "very_secret"
 ;api_version = 6
 ; Timeout for HTTP requests
 http_timeout = 30
-; Redirect URL entered in Sierra for the patron-specific authentication (does not
+; Redirect URL entered in Sierra for patron-specific authentication (does not
 ; need to be a properly working url since the login process is handled completely in
-; the driver without user interaction)
-redirect_uri = "http://localhost/vufind/MyResearch/SierraAuth"
+; the driver without user interaction). Leave commented out when using an API key
+; with global patron data access.
+;redirect_uri = "http://localhost/vufind/MyResearch/SierraAuth"
 ; Set this to true if you are indexing your records with the full .bXXXXXXY bib ID;
 ; set it to false (the default value) if you are only indexing the numeric ID without
 ; the .b prefix or checksum suffix.
 use_prefixed_ids = false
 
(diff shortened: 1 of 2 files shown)


demiankatz:
@EreMaijala, I tried this on my test server and couldn't get it to work -- if I replace the earlier version we had been working on together with the code here, I can't successfully authenticate with Sierra. Unfortunately, I ended up being very short on time today, so I wasn't able to analyze the code in depth to see what the differences were. I'll have to spend more time on it next month when I get back from vacation. Sorry for the delay!

demiankatz:
@EreMaijala, I had more time to test this today. I was able to get it to authenticate reasonably, but not without changing the code. I'm not sure where we draw the line between things that should be configurable, and things that should be local customizations. Let me explain my scenario and see if you have any thoughts on how to proceed.

I have username_field set to name and password_field set to code in my SierraRest.ini. In my current production environment, this lets users log in using their full name as cat_username, and their university-issued ID number as cat_password. This doesn't always work well because "full name" is a terrible field to use as a unique identifier, and there are inconsistencies in the system.

On my test server, I reconfigured this so that cat_username is the university-issued ID number, and cat_password is the user's email address (which of course seems backwards, but I just wanted to have two pieces of information for authentication rather than just one, and these were the elements I could most easily pull from Shibboleth). Obviously, this doesn't work with patron-specific access, but I was able to make it work with global patron data and some code modifications.

Some specific questions, notes and observations:

1.) It's not 100% clear to me whether username_field and password_field in SierraRest.ini are ever used when global patron data is enabled. I feel like they probably are not, given that I wouldn't expect my system to work right if they were still being validated, but it's all a bit confusing, and I didn't take the time to trace the code (yet, anyway). We may want to improve the documentation if necessary.

2.) I'm pretty confident that in my instance, the patrons/validate endpoint is not working correctly (or your assumption that a null return means success is not correct). As I noted above, I'm using university ID as username and email address as password, which I chose arbitrarily... and it never fails. I can send any random values to the endpoint, and I always get null back. Even though a $password parameter gets sent to authenticatePatron(), it never seems to get validated (I'm using API version 5, just to be clear).

3.) You're using a 'varFieldTag' value of 'b' to retrieve users in authenticatePatron(), which uses barcodes... but since I'm using university IDs, I had to change that to 'u' to make things work. Should this be configurable? I feel like there are so many different parallel tracks for authentication that the configuration is already very confusing, and I'm not in a rush to make it even more so... but it would also be nice to have a way to change this without having to hack the code.

4.) Because of the lack of user validation related to item 2 above, I ended up just adding a line of code to validate the user's email address (coming in via the password parameter to authenticatePatron) against the value retrieved from patrons/find. That seems to serve my basic need. That's probably better implemented as a local customization if there's not a more elegant way to handle it, but I suppose there might be configuration-based approaches too.

Anyway, happy to discuss further. I think I'm close to having what I need to move forward, one way or another... but I'm not sure how to make this less confusing and maximize flexibility, and I'm not sure how we know when this PR is done, since there are quite a few different API versions and potential configurations that could impact behavior, and I don't think we have the ability to test them all.

EreMaijala:
@demiankatz Thanks for testing and all the feedback.

1.) username_field and password_field are indeed only used when handling the CAS login form with user-specific data access (redirect_uri is set). I improved the comments in SierraRest.ini to better reflect this.

2.) You were right, makeRequest was masking the actual result. Success is really defined by HTTP status code 204. This should be fixed now.

3.) I added an option (patron_lookup_field). I agree this is getting confusing as there are quite a few branches. I now tried to refactor the code a bit to smaller helper methods in attempt to make it slightly less confusing.

I've been able to test that normal authentication with both v5 and v6 APIs work properly with global patron data and patron-specific data. I think this is the most important thing to test to make sure that login never succeeds with wrong password. And if you can test the passwordless case, I believe we are pretty good here.

PR #1900, opened 2021-03-29
Remove MARC control-characters-as-entities fix

EreMaijala:
It would break any record that contains one of given entity-like strings in the metadata.

For example a record indexed as ISO2709 with this field would fail to parse:

```
		<datafield tag="505" ind1="8" ind2=" ">
			<subfield code="a">Screenwriting Tip #17Screenwriting Tip #18; Screenwriting Tip #19; Screenwriting Tip #20; Chapter three Rebel without a Plan: Outlining Is the Best Present You Can Give Your Future Self; Screenwriting Tip #21; Screenwriting Tip #22; Screenwriting Tip #23; Screenwriting Tip #24; Screenwriting Tip #25; Screenwriting Tip #26; Screenwriting Tip #27; Screenwriting Tip #28; Screenwriting Tip #29; Chapter four Inciting Incident: Away We Go; Screenwriting Tip #30; Screenwriting Tip #31; Screenwriting Tip #32; Screenwriting Tip #33; Screenwriting Tip #34; Screenwriting Tip #35; Screenwriting Tip #36</subfield>
		</datafield>

```
File_MARC would quietly chop the field at #29; while MarcReader would throw an exception. Neither case is desirable. This piece of code seems to originate from 2008 (https://sourceforge.net/p/vufind/svn/906/), but there's no real explanation.

I have tested that SolrMarc indexes a record with the control characters properly in place regardless of whether the input was in ISO2709 or MARCXML format. If there's still a case where the entities get double-encoded or something when they're sent to Solr, I think the fix needs to be done somewhere else. At this point there's no way to determine what's a mis-encoded entity and what's not.

TODO
- [x] Run full test suite


demiankatz:
That code pre-dates my involvement with the project, so I'm also unsure exactly what it was about. It might be worth asking on solrmarc-tech to see if Bob Haschart can shed any light.

Can we add a test record containing these special characters so we can build in a test to check for regressions in the future?

I see that you've changed one of the test fixtures, but I'm guessing that's simply to keep existing tests passing rather than to specifically check this modified behavior.

I'll run the full test suite on this change when I return from vacation.

EreMaijala:
@demiankatz Yes, the original test change was only to use proper JSON escapes. I've added a minimal case to the MarcReader test. It would fail without the change.

I'm not on solrmarc-tech, so I'd appreciate it if you could query on the issue if you believe it's useful.

demiankatz:
Thanks, @EreMaijala! I've sent a message to solrmarc-tech and will report back here if I get any responses.

demiankatz:
Michael Levy pointed out [this blog post](https://bibwild.wordpress.com/2013/06/18/upgrading-a-blacklight-app-from-solr-1-4-to-solr-4-3/) from Jonathan Rochkind, which talks about this in the "Deal with weird escaping of Marc21 binary" section. This doesn't seem to explain why/where this is happening, only that apparently at some point in time, it started happening for some reason. Perhaps it has now stopped again. It sounds like this is a problem unique to storing binary MARC in Solr, so we definitely want to be sure we've tested that scenario (which will get tested when I run the full test suite, in the unlikely case that it hasn't been already). I'd still feel better knowing exactly what changed and when, but it seems like it might take a significant amount of effort to figure that out.

EreMaijala:
@demiankatz I did some more digging and found this: https://github.com/solrmarc/solrmarc/commit/748b69696812f089e5871ef9f642f03e72e2a177. Looking at the current code it looks like choosing to not use a binary writer is no longer possible; the parameter useBinaryRequestHandler still exists at https://github.com/solrmarc/solrmarc/blob/master/src/org/solrmarc/solr/SolrCoreLoader.java#L25 but isn't effective.

demiankatz:
@EreMaijala, yes, that could be a possible explanation -- for a while, SolrMarc wrote directly to Solr via the SolrJ APIs, but it was eventually changed to always use HTTP transactions due to the complexity of maintaining the binary code. If the weird encoding was a side effect of using the binary mode, that could explain both why it started happening and also why it stopped.

demiankatz:
Oh, another important factor which might be in play here: we used to use XML responses from Solr, and now we use JSON. That might also explain the change.

demiankatz:
From Till Kinstler on solrmarc-tech:

<blockquote>
I just tested this: Post a Solr JSON document with a ISO-MARC record without replacing the control characters to a Solr collection in Solr 8.6.3, then retrieve it through the XML response writer. The ISOMARC comes back with the control characters replaced by #29;, #30;, #31; in the XMLish output (see below, relevant snippets of original XML and JSON responses). While the JSON output has the original MARC control characters. I have no idea, when Solr has learned this...
</blockquote>

...so it sounds like the response writer theory is correct, and this is still a problem if you use wt=xml instead of wt=json... but since VuFind isn't doing that any more (to the best of my knowledge/recollection), we're probably safe here... and if we wanted to use an xml handler for some reason, we would want to move the replacement logic into an XML-specific listener instead of its current central location.

EreMaijala:
@demiankatz Right, I just tried wt=xml and got the same results. I did a bit more digging and here's the escaping:
https://github.com/apache/solr/blob/main/solr/solrj/src/java/org/apache/solr/common/util/XML.java#L35
And it all starts to make sense when considering what's valid in XML:
https://en.wikipedia.org/wiki/Valid_characters_in_XML

So there's no way to handle the ISO2709 records in XML (1.0) unless the control characters are escaped somehow. Or the hashes in metadata would be escaped so that they wouldn't be confused with the escaped control characters. Fortunately, like you said, we should be safe here. Or we could try to add XML 1.1 support in Solr...

Concluding this with a quote from http://www.firstobject.com/control-characters-in-xml.htm:
> In its infinite wisdom the XML 1.0 standard excluded the control characters in the range 0x01 to 0x1f except whitespace 0x09, 0x0a, 0x0d, even in escaped form. This was reversed in XML 1.1 but it was too late. This mistake is perfectly understandable if you agree that the purpose of the XML standard was to over-engineer the concept of a simple markup format. ;)

demiankatz:
Thanks, @EreMaijala! Further discussion on solrmarc-tech suggests that SolrMarc was doing this escaping in version 2.x but stopped in 3.x (based on tests where JSON documents indexed with SolrMarc 2 do contain the escaped characters while JSON documents indexed with SolrMarc 3 do not). I theorized that SolrMarc stopped escaping when Solr started it... but that theory is undermined by the fact that when I Git blame the file you shared, the relevant lines are 15 years old. I suppose it's possible that Solr has been escaping these all along, and it was always unnecessary for SolrMarc to do the work... but it used to do it anyway.

demiankatz:
All tests are passing! When time permits, I'm going to do a little more digging through the code to reassure myself that this is totally safe, but I think it's ready to merge.

demiankatz:
I have thoroughly reviewed the code, and the only place where we're still using wt=xml is in retrieving core information for the admin display, which is totally unrelated to MARC retrieval. I'm quite confident this is safe to merge and is, indeed, a bug fix. If anyone is doing XML-based retrieval for some reason (which doesn't seem likely or advisable anyway, since JSON is easier and more efficient), they would need to reintroduce this fix, but it would fit more appropriately in a different layer of the code anyway.

PR #1901, opened 2021-03-29
add fieldType ignored and dynamicField

befehl:
This change add fieldType ignored and a dynamicField using it as *_ignored_mv to schema.xml
The fieldType ignored was up to Solr 7.x included with Solr.

### DIFF
--- a/solr/vufind/biblio/conf/schema.xml
+++ b/solr/vufind/biblio/conf/schema.xml
@@ -112,6 +112,8 @@
     <fieldType name="boolean" class="solr.BoolField" sortMissingLast="true" omitNorms="true"/>
     <!-- add geo field to handle geographic search and display capabilities -->
     <fieldType name="geo" class="solr.SpatialRecursivePrefixTreeFieldType" distErrPct="0.025" maxDistErr="0.000009" distanceUnits="degrees" />
+    <!-- add fieldType ignored to ignore fields, see dynamicField *_ignored_mv -->
+    <fieldType name="ignored" class="solr.StrField" indexed="false" stored="false" docValues="false" />
   </types>
  <fields>
    <!-- Required by Solr 4.x -->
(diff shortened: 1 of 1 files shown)


EreMaijala:
@befehl, Is this for customized copyField directives, or what's the purpose of ignored fields?

befehl:
It is to explicitly ignore fields and its content while loading with solrmarc by adding the field to be ignored to marc_local.properties.

EreMaijala:
@befehl I'm probably missing a bit of knowledge, but could you explain why you'd go through all the trouble of adding such field in marc_local.properties just to ignore it later on? Can you provide an example piece of marc_local.properties to illustrate this?

befehl:
We have the MARC field "700" with subfield "e" in our MARC data "700e", but we don't want to ignore it.
The standard from marc.properties is to use 700abcd which is good for us, but not 700e.
More over, solrmarc was trying add 700e to solr index and solrmarc.log showed exceptions.
Solutions:
1. remove all 700e from all loaded MARC data (not really an option)
2. rewrite the handling of all 700 subfields with solrmarc to eliminate the handling of 700e
3. add an ignored fieldType to schema.xml and just add "ignore_ignored_mv = 700e" to marc_local.properties

We choosed number 3, problem solved.
And is all in one place at marc_local.properties. If someone wants to add it later on, just change the mapping.

EreMaijala:
Thanks for the explanation. I was hoping I could provide feedback on this while @demiankatz is away, but it's probably better to leave this for him when he's back.

demiankatz:
Thanks for joining in, @EreMaijala. I'm also a bit puzzled by this. @befehl, could you share one of the exact errors you were getting from the solrmarc-tech logs? I don't understand why mapping a value to an ignored field would make a difference...

When you first mentioned the "ignored" field, I was thinking of it in terms of a different use case: a situation where you have a tool that indexes data to Solr, and some of the values automatically created by the tool are not needed. You could then customize your schema to treat those fields as "ignored" instead of other values, to avoid storing them in the index. That's a fairly unlikely edge case, but not impossible (based on some of my own past experiences). But a situation where you map values into a dynamic ignored field seems sub-optimal, if there's a way to simply make the tool not send those values at all in the first place. I'm wondering if there might be a better solution for that part of the problem.

befehl:
As I am short in time due to my development in BASE I choosed this solution. Nevertheless I will go through solrmarc config and locate the problem there. Currently I keep my solution as quick fix and keep in our local git, because I had also to fix a coding issue with ICU in schema.xml of solr.
Sorry for wasting your time.

demiankatz:
No worries, @befehl; like I said, I think the field type could be useful in the future, even if the dynamic field may be less likely to be needed. I'm happy to revisit this later if there is a need.

PR #1902, opened 2021-03-30
Initialize id array even if there are no more records.

EreMaijala:


demiankatz:
Thanks! I've cherry-picked this back to the release-7.1 branch and merged to dev from there.

PR #1903, opened 2021-03-31
Prevent double submit of renewal form.

EreMaijala:
Adds a generic mechanism that allows to disable submit buttons after form has been submitted. Also adds CSRF to renewal form to avoid re-renewal if user resubmits the page.

TODO
- [x] Resolve setTimeout debate
- [x] Run full test suite


demiankatz:
@EreMaijala, there are some language file corrections in this PR that seem unrelated to the rest. Do you just want me to cherry-pick those off?

EreMaijala:
Oh, oops. Yes, that'd be nice. I only noticed them as the new translation string was suitably placed but should have made that a separate issue.

I also should have mentioned that I tried to design this in a way that would make it easy to apply in other places too. I think it would make sense to use more csrf checking in actions that modify anything. And apparently it's quite easy to double or triple click buttons especially with touch screens, so preventing double submits would help in those cases.

demiankatz:
Great, I've merged the language fixes to release-7.1 and dev and updated this PR. I've requested a review from @crhallberg in case he has any thoughts on the Javascript. I'll give the rest a closer look when I'm back from vacation.

themes/bootstrap3/js/common.js
```diff
@@ -38,13 +38,30 @@ var VuFind = (function VuFind() {
...
+    $('[data-disable-on-submit]').on('submit', function handleOnClickDisable() {
+      var $form = $(this);
+      // Disable submit elements via setTimeout so that the submit button value gets
+      // included in the submitted data before being disabled:
+      setTimeout(
+        function disableSubmit() {
+          $form.find('[type=submit]').prop('disabled', true);
+        },
+        0
+      );
```
crhallberg, line 52: Opinionated request: Instead of `setTimeout(fn, 0)`, could we use `requestAnimationFrame`? It accomplishes the same thing but feels less hacky to me.

crhallberg (approved):
My only meaningful concern with this PR is whether or not to use this mechanism in more forms. If so, would it be better to make the disable default instead of only when the data attr is present?

themes/bootstrap3/js/common.js
EreMaijala: @crhallberg I don't have a preference, but from reading the spec for requestAnimationFrame it sounds like it's more real-time and would probably fire before next paint (which would be nice), but is it certain that the form submission has completed at that time? In other words, is it possible that requestAnimationFrame gets completed before initDisableSubmitOnClick returns?

EreMaijala:
@crhallberg I'd proceed with caution since we (or downstream users with custom extensions) may not prepared for the situation in all forms. Eventually, I'd like this to cover as much as possible, but enabling by default could mess with AJAX forms, validation etc. so I'd rather approach this more cautiously. In the end we could maybe flip to opt-out.

themes/bootstrap3/js/common.js
crhallberg: Not to my understanding. It uses the event loop to defer same as setTimeout.

EreMaijala: I'd rather be cautious here as well. Using setTimeout is the way recommended all over the web. I didn't find examples using requestAnimationFrame for this purpose. If you have convincing resources, please share. :)

demiankatz: I found [this article](https://javascript.plainenglish.io/better-understanding-of-timers-in-javascript-settimeout-vs-requestanimationframe-bf7f99b9ff9b) fairly informative, though it doesn't really persuade me one way or the other in the setTimeout vs. requestAnimationFrame debate for this particular use case. It does bring up the interesting point that it's a good idea to clear timers that you have created (whichever method you use); I suspect VuFind has a lot of shortcomings on that point. Any thoughts on that, or anything else?

In any case, perhaps the path of least resistance would be to merge this as-is, and then create a separate PR to evaluate our use of timers and optimize/improve where feasible. I confess that I really hate setTimeout, as it almost always feels like a hack, but I recognize that it is often needed for legitimate reasons. :-) For now, as a starting point, I'll run the full test suite on this code as-is and report if it encounters any problems.

demiankatz:
The full test suite is passing, for what it's worth (which probably isn't much, since we don't have any tests actively trying to double-submit forms, etc. ... but at least the normal renewal tests are not broken by these changes).

themes/bootstrap3/js/common.js
EreMaijala: The article still doesn't tell if it's possible that the callback get called before the form is submitted. I'd suppose it is, if it's called whenever the browser is ready to paint, and e.g. button release will require a paint.

crhallberg: The talk in the article is one of the sources for learning about this but I didn't come away with any clearer on `requestAnimationFrame` so I do agree that `setTimeout` is the safer option. My tests show that requestAnimationFrame would work the same.

An even safer and clearer option might be to attach the listener to the btns and add a hidden input to the form to cover the disabling. I messed around with some JS and came up with this proposal:

```JS
$('[data-disable-on-submit]').each(function (i, form) {
  let $form = $(form);
  let $btns = $form.find("[type='submit']");

  $btns.on("click", function submit(e) {
    let $btn = $(e.target);

    let $input = $('<input type="hidden"/>');
    $input.attr("name", $btn.name);
    $input.val($btn.value);
    $form.append($input);

    $btns.prop("disabled", true);
    $form.submit(); // I don't like this part
  });
});
```

The `setTimeout` above is definitely more eloquent.

demiankatz: Thanks, @crhallberg! Sounds like we can move forward with the existing solution, though future exploration may be worthwhile. I'll wait to hear what @EreMaijala thinks about merging now vs. doing more development before taking further action.

EreMaijala: @demiankatz I'd move forward with the existing solution. We can always revisit this. 

@crhallberg There's a potential problem with attaching to the buttons in that if something, such as late validation, blocks the form submit, the buttons get disabled prematurely and would prevent submitting the form after correction.

demiankatz: Great, merging now! I welcome a follow-up PR to improve the solution and/or extend it to more forms, if and when there's a need.

demiankatz:
@EreMaijala, on my final last-minute review I caught a problem that we've all overlooked up to this point -- I fixed it with 31c4d29. I think this is uncontroversial, so I'm going ahead with the merge.

EreMaijala:
@demiankatz Oops, great catch! Thanks!

PR #1904, opened 2021-04-01
Extend sitemap generation to index and content pages.

EreMaijala:
Also adds support for including language-specific alternative URLs.

The idea is to make the base sitemap creation more automatic, or to avoid the manual work altogether.

I've tried to follow the recommendations at https://developers.google.com/search/docs/advanced/crawling/localized-versions when adding the language support.

TODO:
- [x] Get feedback on whether this is a good idea.
- [x] Add support for including something else?
- [x] Add wiki page for new plugin type
- [x] Add https://vufind.org/wiki/development:plugins:sitemap_generators to https://vufind.org/wiki/development:plugins
- [x] Update https://vufind.org/wiki/administration:search_engine_optimization
- [x] Add changelog note about BC breaks


demiankatz:
I'll look at this more closely when I'm back from vacation, but a couple of quick thoughts:

1.) I wonder if there's an opportunity for a plugin system here for adding URLs to the sitemap... rather than embedding all different kinds of URL lists into the main generator class, could we delegate this to plug-ins? That would make it easier to add different kinds of custom URLs to the sitemap in the future, and might also allow us to better separate properties and dependencies. I realize that requires a bit more conceptual design work, but I think it might pay off in terms of flexibility and maintainability, rather than just adding lots of additional functionality to the existing class.

2.) Should we make the alternate language logic a little more flexible, so you can potentially specify a list of languages to include, in case you want to highlight two or three instead of the full complement?

EreMaijala:
@demiankatz 

1.) I think the idea is sound, but I'm having a bit of difficulty trying to think of how the plugins would tie with the generator. There are so many different aspects:
 - Plugin needs to be able to access getNewSitemap
 - The results of a plugin affect the sitemap index file (one file, multiple files, access to getFilenameForPage, how to make sure there are no conflicts)
 - Plugin should be able to output messages (access to verboseMsg)
 
 All of these can of course be solved, but I'm not quite sure how to make this clear and concise. 

2.) I've improved the language handling, but I ended up making the language list work like a filter so that languages not enabled in VuFind won't be included. Makes sense?

demiankatz:
Thanks, @EreMaijala. I like the language improvements!

Regarding the plugin design, I was thinking that the plugins could be "SitemapUrlProviders." The interface could be something like:

getSitemapName(): string

getUrls(): ?array

setMessageCallback(): void

So each plugin returns a base name for the sitemap it's being used to build, and returns batches of URLs... it can also be given a callback method for displaying output, in case you need it to provide real-time feedback.

Then the Sitemap Generator's job becomes managing the plug-ins, collecting URLs from them, and creating the appropriate files.

This could be very simple, or very complicated, depending on whether you do a simple "every plugin must use a unique base name, and every page returned from getUrls() equates to the contents of a single sitemap file" or a more complex "plugins can stream URLs into the same filenames, and their pages can be any size, so it's up to the generator to appropriately route/merge/split data". But I think either of those approaches is potentially feasible.

Does that help, or am I overlooking key details? My main point is that my view is the separation of concerns here is that plugins should provide URLs from various sources, and the generator should care about all the details of putting those URLs in the right file using the right format. (And, of course, whether the plugins just return URL strings, or arrays representing the various sitemap elements, is open to debate... I'd perhaps propose supporting either format for flexibility).

EreMaijala:
Plugin support added. Start page and content page support moved to their respective plugins. I ended up making the main method addUrls that takes a Sitemap for maximum flexibility. I also made it so that multiple plugins can output to the same Sitemap.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -232,4 +232,56 @@ public function findContainingTheme($relativePath, $returnType = false)
...
+     *
+     * Note that for any matching file the last match in the theme hierarchy is
+     * returned.
+     *
+     * @param string|array $relativePathPattern Relative path pattern (or array of
+     * patterns) to search within themes
+     *
+     * @return array
+     */
+    public function findInThemes($relativePathPattern)
```
demiankatz: This new method seems like a good candidate for a unit test (and potentially for merging separately, prior to the rest of this code). If you'd like me to split it off and write the test, I'm willing to do that if you don't get there first.

EreMaijala: Test added. I'd be fine with merging it separately, though it feels a bit odd to add a method that's seemingly unused..

demiankatz: When a new method is of general use and stands alone, I like to commit it separately just to keep the history cleaner (and to reduce cognitive load when reviewing commits). I'd rather see a small commit with a message like "Add new method to do X, to be used by feature Y" than to just have an "Add feature Y" commit on a shared utility class, which provides less context about why the file was touched when browsing history. But I'm just explaining myself, not asking for anything to be done differently; I'm happy to manually pull these things out as needed.

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
```diff
@@ -179,6 +179,38 @@ public function testFindContainingThemeWithMixin()
...
+    {
+        $ti = $this->getThemeInfo();
+        $ti->setTheme('child');
+        $files = $ti->findInThemes(
+            [
+                'templates/content/*.phtml',
+                'templates/content/*.md'
+            ]
+        );
+        $this->assertEquals('parent', $files[0]['theme']);
```
demiankatz: Feel free to disagree if you took this path intentionally, but I'd be inclined to be a little more specific about the assertions, either by doing one big assertEquals on the whole $files variable, or else by adding `$this->assertEquals(3, count($files))`, just to be sure we get what we expect, and ONLY what we expect. I'm happy to make the change if you agree with the principle. :-)

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
EreMaijala: @demiankatz Ok, makes sense!

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
EreMaijala: Adding the count check makes sense. I decided to chicken out from one big assert since one of the returned fields is an absolute path. I suppose I can dig it up though, if it would be better that way.

demiankatz: I'm fine with just adding the count for now, rather than spending a lot of time recreating that path. :-)

EreMaijala: Done. :)

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
demiankatz: I've gone ahead and done this as commit 6a90d39. Thanks again!

config/vufind/sitemap.ini
```diff
@@ -50,6 +50,27 @@ index[] = "Solr,/Record/"
 ;    other limiting options and requires that the index has terms enabled.
 retrievalMode = search
 
+;  This settings controls whether different language versions are added to the
+;  sitemap. Default is false.
+;  Note that including language versions will make the sitemap files substantially
+;  larger and may not be very useful if language detection is enabled in VuFind.
+indexLanguageVersions = false
+
+;  This setting can be used to filter the included language codes. By default all
```
demiankatz: Is it worth adding "when indexLanguageVersions is true," or is that already implied strongly enough?

module/VuFind/src/VuFind/Sitemap/Generator.php
```diff
@@ -276,14 +317,49 @@ public function generate()
...
 
+    /**
+     * Generate sitemaps with any enabled plugins
+     *
+     * @return array
+     */
+    protected function generateWithPlugins(): array
+    {
+        $additionalSitemaps = [];
+        if ($plugins = $this->config->Sitemap->plugins) {
```
demiankatz: Does this behave appropriately if the setting is missing entirely? Just want to be sure there's no risk of notices.

module/VuFind/src/VuFind/Sitemap/Generator.php
```diff
@@ -483,31 +565,37 @@ protected function buildIndex($totalPages)
 
             // Add a <sitemap /> group for a static sitemap file.
             // See sitemap.ini for more information on this option.
-            if (isset($this->config->SitemapIndex->baseSitemapFileName)) {
-                $baseSitemapFile = $this->config->Sitemap->fileLocation . '/' .
-                    $this->config->SitemapIndex->baseSitemapFileName . '.xml';
+            $baseSitemapFileName = $this->config->SitemapIndex->baseSitemapFileName
```
demiankatz, line 618: It looks like the .xml extension has been lost here. Is that intentional? I realize that it may be more flexible to not force the extension, but changing this behavior will break existing configurations (unless the extension is still getting added somewhere and I'm just overlooking it).

module/VuFind/src/VuFind/Sitemap/Generator.php
```diff
@@ -530,19 +618,31 @@ protected function getNewSitemapIndex()
      */
     protected function getNewSitemap()
     {
-        return new Sitemap($this->frequency);
+        $sitemap = new Sitemap($this->frequency);
+        if ($this->config->Sitemap->indexLanguageVersions) {
```
demiankatz: This logic is potentially going to get repeated many times; should we do the work once and store it in a property?

module/VuFind/src/VuFind/Sitemap/Generator.php
```diff
@@ -276,14 +317,49 @@ public function generate()
...
+                );
+                if (!isset($pluginSitemaps[$sitemapName])) {
+                    $pluginSitemaps[$sitemapName] = $this->getNewSitemap();
+                }
+                $plugin->addUrls($pluginSitemaps[$sitemapName]);
+            }
+
+            foreach ($pluginSitemaps as $sitemapName => $sitemap) {
+                if (!$sitemap->isEmpty()) {
+                    $filename = $this->getFilenameForPage($sitemapName);
```
demiankatz: This may be fine for now, but do we potentially need to worry about plugins that generate huge numbers of URLs? Should we have a mechanism for splitting a huge list into multiple files?

module/VuFind/src/VuFind/Sitemap/Plugin/ContentPages.php
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @param Sitemap $sitemap Sitemap to add to
+     *
+     * @return void
+     */
+    public function addUrls(Sitemap $sitemap): void
+    {
+        $files = $this->themeInfo->findInThemes($this->includedFiles);
+        $nonLanguageFiles = [];
+        $languages = array_keys($this->config->Languages->toArray());
```
demiankatz: Now that the slm/locale PR has been merged, it might be better to use LocaleSettings::getEnabledLocales() here instead of direct config access.

demiankatz: Since $this->config isn't used for anything else, I'd suggest just taking in the language list through the constructor and putting the logic to fetch the list from LocaleSettings into the factory... that keeps the class itself simple and easier to test.

module/VuFind/src/VuFind/Sitemap/Sitemap.php
```diff
@@ -38,13 +38,22 @@
  */
 class Sitemap extends AbstractFile
 {
+    const XHTML_NAMESPACE = 'xmlns:xhtml="http://www.w3.org/1999/xhtml"';
```
demiankatz: We should set this public or protected to keep PHPStan happy when we raise its level in the future.

module/VuFind/src/VuFind/Sitemap/Sitemap.php
```diff
@@ -65,11 +104,31 @@ public function __construct($frequency = 'weekly')
      */
     protected function getEntry($url)
     {
-        $loc = htmlspecialchars($url);
+        $alternativeLinks = '';
+        if ($this->alternativeLanguages) {
+            $lngParam = strpos($url, '?') === false ? '?lng=' : '&lng=';
```
demiankatz: If we have a sitemap plugin that provides URLs for code that lives outside of VuFind, do we need to worry about this logic inappropriately manipulating URLs? I haven't traced all the relationships thoroughly to understand exactly how things might behave... but this seems like some very VuFind-specific behavior being added to a class that was meant to be a fairly abstract representation of an XML sitemap. Would it be better to implement this with a callback function for applying language codes to URLs, or something like that?

demiankatz (commented):
I've now taken the time to review this more thoroughly, and I've left a few comments. I haven't actually tried to run the code yet, but I'll do some hands-on work (and maybe write some additional unit tests) once you've had a chance to respond.

EreMaijala:
@demiankatz Thanks, I'll fix and revise.

module/VuFind/src/VuFind/Sitemap/Generator.php
EreMaijala: Yes. As long as $this->config->Sitemap is \Laminas\Config\Config, it works and just returns null. The constructor is already doing that.

module/VuFind/src/VuFind/Sitemap/Generator.php
EreMaijala: Oops, no. Fixed.

module/VuFind/src/VuFind/Sitemap/Generator.php
EreMaijala: Done.

module/VuFind/src/VuFind/Sitemap/Generator.php
EreMaijala: Good point. Now added and refactored plugins to use yield.

module/VuFind/src/VuFind/Sitemap/Plugin/ContentPages.php
EreMaijala: Done.

module/VuFind/src/VuFind/Sitemap/Sitemap.php
EreMaijala: Done.

module/VuFind/src/VuFind/Sitemap/Sitemap.php
EreMaijala: Added option for a plugin to disable language versions. Plugins can now also define frequency that overrides the default one. And they can now return per-url languages and/or frequencies for maximum flexibility.

EreMaijala:
@demiankatz I think this is now much better and addresses your comments.

module/VuFind/src/VuFind/Sitemap/Generator.php
```diff
@@ -142,41 +163,53 @@ class Generator
      *
      * @param BackendManager $bm      Search backend manaver
      * @param SearchService  $ss      Search manager
-     * @param string         $baseUrl VuFind base URL
```
demiankatz, line 145: It appears that during the course of all the refactoring, the order of these two parameters got swapped. Was that intentional? Seems like it might be easier to keep the existing order to reduce potential BC breaks, though I'm fine with the change if there's a compelling reason.

config/vufind/sitemap.ini
```diff
@@ -43,13 +43,37 @@ fileLocation   = /tmp
...
 ;  'search' (the default, most compatible but slower method), or 'terms' (the
 ;    faster option). Note that 'terms' method does not support hidden filters or
 ;    other limiting options and requires that the index has terms enabled.
 retrievalMode = search
 
+;  This settings controls whether different language versions are added to the
+;  sitemap. Default is false.
+;  Note that including language versions will make the sitemap files substantially
+;  larger and may not be very useful if language detection is enabled in VuFind.
+indexLanguageVersions = false
```
demiankatz: Changing this to true is having no effect on the main record sitemaps generated by my test instance, whether or not I also set allowedLanguages below. It does work for content pages, though. Is this intentional?

demiankatz (commented):
Thanks for the improvements; this is shaping up very nicely. I found a couple of issues, though, noted below. Also, due to the large quantity of code to review here, it might also be a good idea to add more tests. If you'd like me to write a couple as part of my own review process, let me know; I'm definitely willing!

module/VuFind/src/VuFind/Sitemap/Generator.php
EreMaijala: Oops. Now refactored back. It's not perfectly back-compatible though, since the remaining new parameters are still required.

config/vufind/sitemap.ini
EreMaijala: Oops. I forgot to update the relevant part when changing the language handling mechanism. Should work now.

module/VuFind/src/VuFind/Sitemap/Generator.php
demiankatz: True. I've added a note to update the changelog when I merge this... but keeping the order the same at least reduces the cognitive load of bringing things up to date. :-)

config/vufind/sitemap.ini
demiankatz: Hmm, now I always get the alternate languages, regardless of how I set indexLanguageVersions.

demiankatz: Messing around with this makes me wonder whether we really need indexLanguageVersions AND allowedLanguages. Could we just have a single setting that is true for all, false for none, and an array for selected values? (I'm fine with keeping it as it is when you prefer... but I'm wondering if having multiple settings is making it easier to introduce bugs).

EreMaijala: Fixed the setting (another casualty of bad refactoring), but I think you're right. I'll refactor a bit more!

EreMaijala: I'm having a bit of problem detecting the type of the setting. Do we have such settings somewhere else? Or should it be empty for none, `*` for all and a comma-separated list for specific ones (like e.g. `orFacets` in facets.ini)?

demiankatz: Yes, I like that idea! Locale names should never contain commas, so there's no technical need to format them as an array in the .ini file anyway.

EreMaijala: Should be done now. I also moved the x-default logic from Sitemap to Generator to keep the logic out of Sitemap. And fixed a mistake in fallback language handling to avoid duplicate hreflang entries.

EreMaijala:
@demiankatz I updated the related wiki pages mentioned in the TODO points. Let me know if there's anything further you'd like to see in them.

config/vufind/sitemap.ini
```diff
@@ -43,13 +43,37 @@ fileLocation   = /tmp
...
 retrievalMode = search
 
+;  This settings controls whether different language versions are added to the
+;    sitemap. Possible values:
+;    empty or undefined - Language versions are not added
+;    "*"                - Language versions are added for all languages enabled in
+;                         config.ini
+;    "en,de,..."        - A comma-separated list of languages to include from the
+;                         ones enabled in config.ini
+;  Note that including language versions will make the sitemap files substantially
```
demiankatz, line 62: @EreMaijala, the two notes here seem to have gotten deleted during your latest refactoring. I put them back because they still appeared to be relevant, but please feel free to revert if you had removed them intentionally.

demiankatz:
Thanks for the work, @EreMaijala! This seems to be working as expected now. I'm going to take a little time to write more tests, and then will merge as long as those turn up no major issues (and you don't object to my recent config file edits).

demiankatz:
@EreMaijala, update: I added tests for the plugins in 59f72fd, which revealed the need for some minor improvements, which I have made.

module/VuFind/src/VuFind/Sitemap/Generator.php
demiankatz: Actually, if the entire [Sitemap] section is missing, this does cause a notice, as I discovered when beginning to build a Generator test class. I addressed this in 4b4499994b, along with using a class property in place of a duplicate config lookup. Please let me know if you have any concerns about this!

demiankatz:
@EreMaijala, I've run out of time for today, but I at least got a very small start on building a GeneratorTest -- enough to find and fix a couple more issues (see reopened thread above). I'd love to make the GeneratorTest more complete in the future (and also to perhaps improve some of the weird behavior of the Generator class' setters, which doesn't make much sense to me, and in some cases seems inconsistent with the comments -- e.g. `setVerbose()`)... but for now, I'm comfortable with merging this as a step in a positive direction, and we can always open another PR for additional work.

I'm just going to wait for your approval of my recent changes and comments before merging. Just let me know if/when you think we're ready to go!

EreMaijala:
@demiankatz Thanks, your changes look good to me and thanks for working on the tests also.

PR #1906, opened 2021-04-06
Demo driver: Add fake IDs to transactions when idsInMyResearch is false.

EreMaijala:
This allows renewals to work properly.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Demo.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Demo.php
@@ -472,7 +472,6 @@ class Demo extends AbstractBase
                 "expire"   => $this->dateConverter->convertToDisplayDate(
                     'U', strtotime("now + 30 days")
                 ),
-                "reqnum"   => sprintf("%06d", $i),
                 "item_id" => $i,
                 "reqnum" => $i
             ];
--- a/module/VuFind/src/VuFind/ILS/Driver/Demo.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Demo.php
@@ -472,6 +472,7 @@ class Demo extends AbstractBase
                 "expire"   => $this->dateConverter->convertToDisplayDate(
                     'U', strtotime("now + 30 days")
                 ),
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -472,6 +472,7 @@ protected function createRequestList($requestType)
                 "expire"   => $this->dateConverter->convertToDisplayDate(
                     'U', strtotime("now + 30 days")
                 ),
+                "reqnum"   => sprintf("%06d", $i),
```
demiankatz: @EreMaijala, this line got removed during static analysis because there's a duplicate "reqnum" property two lines further down. If we need this version for things to work, can we get rid of the other one?

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -1078,6 +1079,7 @@ protected function getRandomTransactionList()
                         = $this->getRandomBibIdAndTitle();
                     $transList[$i]['source'] = $this->getRecordSource();
                 } else {
+                    $transList[$i]['id'] = "demo_nonexistent_$i";
```
demiankatz, line 1081: The purpose of the idsInMyResearch property checked above is explicitly to exclude ids when it is set to false, so it seems wrong to add an id in this case. I honestly can't remember what use case idsInMyResearch was originally created for, so maybe we need to have a discussion about its usefulness, but I think that's a separate concern. I can dig in and refresh my memory on this topic as needed!

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: Oops. That was accidentally added when I merged the changes from an old working copy. Now removed.

EreMaijala:
Uh oh, now that I tested again, renewals actually work fine without the id. If there's anything further needed, I'll open a new pull request.

PR #1908, opened 2021-04-07
Fix invalid leader handling in ISO2709 serialization, add tests

EreMaijala:
If the leader was too short, the resulting record would have been invalid. Added tests bring coverage of VuFind\Marc to 100%.


demiankatz:
Thanks!

PR #1910, opened 2021-04-07
Add default values for possibly undefined variables

xmorave2:
I used rector tool do make same code update.

I tried to keep every kind of change in own commit for chance we finally not adopt some of the changes.

Some changes (at least in commit "Add default values for possibly undefined variables") also helps to raise phpstan level.


demiankatz:
Thanks, @xmorave2 -- this is great! As with the previous PHPStan PR, I'll review this a commit at a time, and cherry-pick commits to dev, then rebase here as time permits. I've started by plucking out the PHP version constant commit, which is both a bug fix and a simplification. I really appreciate all the effort you're putting into code quality improvement!

module/VuFindTheme/tests/unit-tests/src/VuFindTest/View/Helper/HeadThemeResourcesTest.php
```diff
@@ -99,7 +99,7 @@ protected function getMockHeadMeta()
         $mock->expects($this->once())->method('prependHttpEquiv')
             ->with($this->equalTo('Content-Type'), $this->equalTo('text/html; charset=utf-8'));
         $mock->expects($this->once())->method('appendName')
-            ->with($this->equalTo('Generator'), $this->equalTo('fake-generator'));
+            ->with($this->equalTo(\Generator::class), $this->equalTo('fake-generator'));
```
demiankatz: This is referring to the 'Generator' meta-tag, not the \Generator PHP class, so I don't think this is an appropriate change. I'll merge the other class constants, but I think we should revert this part.

module/VuFind/src/VuFind/Controller/SearchController.php
```diff
@@ -88,7 +88,8 @@ public function editmemoryAction()
             }
         } elseif ($removeFacet) {
             $defaults = ['operator' => 'AND', 'field' => '', 'value' => ''];
-            extract($removeFacet + $defaults);
+            extract($removeFacet);
+            extract($defaults);
```
demiankatz: This changes the meaning of the code -- it will cause the defaults to always override the incoming values. I think it probably makes sense to totally rewrite this as:

<pre>
$query = $initialParams->removeFacet(
    $removeFacet['field'] ?? '',
    $removeFacet['value'] ?? '',
    $removeFacet['operator'] ?? 'AND'
);
</pre>

That seems a lot more straightforward.

module/VuFind/src/VuFind/ILS/Driver/KohaILSDI.php
```diff
@@ -1511,6 +1511,7 @@ public function getMyTransactions($patron)
             "GetPatronInfo&patron_id=$id" . "&show_contact=0&show_loans=1"
         );
         $end = microtime(true);
+        $requestTimes = [];
```
demiankatz: This one is a rabbit hole. I'd change this to the single line `$requestTimes = [$end - $start];` to save a line of code... except I question why we are echoing out debug information inside an ILS driver at all! Shouldn't we be using $this->debug() instead... or maybe just removing this code entirely? I suspect this may be an unwanted relic that's just been in place since the driver was originally contributed many years ago. What do you think?

demiankatz:
@xmorave2, I'm out of time for today and still have quite a bit more reviewing to do, but I've cherry-picked quite a few commits and rebased this PR. I'll do more as soon as time permits!

module/VuFindTheme/tests/unit-tests/src/VuFindTest/View/Helper/HeadThemeResourcesTest.php
xmorave2: I agree, this is mistake

module/VuFind/src/VuFind/Controller/SearchController.php
xmorave2: I amended the commit

module/VuFind/src/VuFind/ILS/Driver/KohaILSDI.php
xmorave2: I amended the commit and also change echo to debug. That is probably a relic from ancient times, and we may revisit it in the future. But I am not sure if it is worth it, as using Koha REST driver does make more sense in long term

module/VuFind/src/VuFind/Controller/SearchController.php
demiankatz: Great, I'll just drop this commit the next time I rebase, since all the remaining parts have already been merged to dev.

module/VuFind/src/VuFind/ILS/Driver/KohaILSDI.php
demiankatz: Thanks! I agree that this is a good enough solution for now, and because of KohaRest, it probably doesn't make sense to invest too much time in KohaILSDI, except to fix problems.

module/VuFindTheme/tests/unit-tests/src/VuFindTest/View/Helper/HeadThemeResourcesTest.php
demiankatz: Great, I'll just drop this commit the next time I rebase, since all the remaining parts have already been merged to dev.

module/VuFind/src/VuFind/Controller/SearchController.php
demiankatz: Thanks!

demiankatz:
Thanks again for the progress, @xmorave2! I've rebased this again, so now it's just down to two longer commits that need to be reviewed. I'll work on that as soon as time permits -- I expect that most of this will be able to be merged as-is, but I think I see a few spots where deeper problems may be getting revealed, and there may be better fixes than the automated ones, so I want to allow myself some time to sift through it all more thoroughly.

demiankatz:
I managed to get through one more commit this afternoon, so now this is down to just one commit. Do you plan to add any more here? If not, I might add the PR, and then close it out by merging that final commit when it's ready. I'll try to review the remaining code soon!

xmorave2:
Thanks @demiankatz. No, i am not going to add more commits to this PR. After this commit is pushed to dev, I'm going to work on phpstan level 1 fixes.

demiankatz:
I've gotten about halfway through the review here and have run out of time, so I'll have to finish this off next week. There have been a few instances where the auto-corrected code has actually introduced bugs, so I'm taking it slow... though there have been more bugs fixed than introduced so far, so it's certainly worth the effort. More soon!

Regarding phpstan level 1, I tried it a couple of days ago just to see how many errors there would be and noticed that there were a lot of warnings about undefined constants. Since most of VuFind's constants are set up in public/index.php, I wonder if we need to include that file in analysis to quiet those problems....

demiankatz:
Okay, I've finished reviewing and testing this, and I think we're all done here. I'll look forward to seeing what PHPStan level 1 has to say!

PR #1911, opened 2021-04-07
Voyager: Add a workaround for changed PDO signature in PHP 8.

EreMaijala:
The changed PDO method signature in PHP 8 makes the PHP 7 compatible Yajra\PDO\Oci class incompatible with PHP 8. This incompatibility makes phpstan crash during analysis on PHP 8, so Voyager.php has been excluded from analysis.

I no longer have access to Voyager to really test this, but at least the setting should be working.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Voyager.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Voyager.php
@@ -37,7 +37,6 @@ use PDOException;
 use VuFind\Date\DateException;
 use VuFind\Exception\ILS as ILSException;
 use VuFind\I18n\Translator\TranslatorAwareInterface;
-use Yajra\Pdo\Oci8;
 
 /**
  * Voyager ILS Driver
@@ -61,7 +60,7 @@ class Voyager extends AbstractBase
     /**
      * Lazily instantiated database connection. Use getDb() to access it.
      *
-     * @var Oci8
+     * @var \Yajra\Pdo\Oci8
      */
     protected $lazyDb;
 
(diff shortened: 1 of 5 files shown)


module/VuFind/src/VuFind/ILS/Driver/Voyager.php
```diff
@@ -187,7 +186,29 @@ protected function getDb()
...
+Voyager.ini or VoyagerRestful.ini:
+
+[Catalog]
+forceOCI8Support = true
+
+EOT
+                    );
+                    throw new ILSException('Unsupported PHP version');
+                }
+                // @phpstan-ignore-next-line
```
demiankatz: Since you're ignoring the whole file via PHPstan configuration, is it worth removing this line and reverting the removal of the use statement, to minimize diffs? (I'm happy to make the changes while testing this, if that would save you time).

EreMaijala: The line can be removed, but I'm not sure about the use statement. Will PHP 8 throw a fatal when including the file which wouldn't give us the chance to write the log message?

demiankatz: Use statements have no run-time impact; they just set up aliases that are used by the language parser. That's why they always have to come before executable code. A use statement on its own won't load anything, so I think the code should be functionally equivalent either way.

EreMaijala: In that case, sure. I'd appreciate it if you could make the changes. :)

demiankatz: Will do!

demiankatz:
This was broken because of some unwanted whitespace, which I fixed in cead99e  -- once that was sorted out, I was able to confirm that Voyager connections still work correctly in PHP 7, and PHPStan now passes in PHP 8. Merging now -- thanks for the help!

PR #1913, opened 2021-04-09
Set phpstan memory limit to 2G.

EreMaijala:
This will avoid running out of memory when PHP's default is too small.

For reference, our branch with our custom module ran out of memory with Travis' 1G limit.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -108,10 +108,10 @@
 
   <!-- Phpstan -->
   <target name="phpstan-checkstyle">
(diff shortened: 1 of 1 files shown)


xmorave2 (approved):
I do agree, thanks @EreMaijala

demiankatz:
Thanks! If we find that this needs to be adjusted again in future, we should consider adding a phpstan_memory_limit property to build.xml with a default value... but for now, the hard-coded is an easy starting point.

PR #1914, opened 2021-04-09
Fix case where Overdrive connector returned null.

demiankatz:
This seems to me like a bug that needs fixing, but I want to get confirmation from @bpalme before merging it.


demiankatz:
@bpalme confirmed that this fix is needed, so I've backported it to the release-7.1 branch and merged it to dev from there.

PR #1916, opened 2021-04-13
Add support for custom Google Analytics create options.

demiankatz:
Because of Same-Site cookie changes, Google Analytics can cause browser warnings. The workaround is to add some custom options to the initialization. This PR adds a mechanism for setting these custom options.

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/GoogleAnalytics.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/GoogleAnalytics.php
@@ -62,8 +62,11 @@ class GoogleAnalytics extends \Laminas\View\Helper\AbstractHelper
     /**
      * Constructor
      *
-     * @param string|bool $key       API key (false if disabled)
-     * @param bool|array  $universal Are we using Universal Analytics?
+     * @param string|bool $key     API key (false if disabled)
+     * @param bool|array  $options Configuration options (supported options:
+     * 'universal' and 'create_options_js'). If a boolean is provided instead of
+     * an array, that value is used as the 'universal' setting and no other options
+     * are set (for backward compatibility).
      */
     public function __construct($key, $options = [])
     {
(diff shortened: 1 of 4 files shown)


module/VuFind/src/VuFind/View/Helper/Root/GoogleAnalytics.php
```diff
@@ -52,16 +52,32 @@ class GoogleAnalytics extends \Laminas\View\Helper\AbstractHelper
...
+     * are set (for backward compatibility).
      */
-    public function __construct($key, $universal = false)
+    public function __construct($key, $options = [])
     {
+        // The second constructor parameter used to be a boolean representing
+        // the "universal" setting, so convert to an array for back-compatibility:
+        if (!is_array($options)) {
+            $options = ['universal' => (bool)$options];
+        }
```
crhallberg, line 77: Not a huge concern but does casting as boolean turn anything into a boolean? Like a string? Would it be better to explicitly test for a boolean here?

crhallberg (approved):
Besides my one small concern, this looks good!

module/VuFind/src/VuFind/View/Helper/Root/GoogleAnalytics.php
demiankatz: Yes, casting to a boolean turns anything into a boolean... this is useful for values that come out of config files, since depending on how the config was created, they could be just about anything -- ones and zeroes, real booleans, etc. This obviously leads to the potential bad side effect that the string "false" evaluates as true because it's a non-empty string, but I think that is fairly unlikely to be an issue in this particular case, especially since I suspect that the "universal" setting may be obsolete due to changes on Google's side, and I may be removing it entirely in the near future if that proves to be true.

demiankatz: I'm going to go ahead and merge for now; making processing of boolean config settings more uniform might be a good project for another PR in the future (and may be more necessary as we get stricter about typing), but I think that can be safely handled as a separate matter from this one.

PR #1918, opened 2021-04-14
SierraRest: Load records in a batch for transactions.

EreMaijala:
The old mechanism of loading an item and its bib separately for each transaction (current or historic) is very inefficient. According to my testing it takes roughly the same time to load a single record or a batch of 10 records. With this change, all items are loaded in a single request and their bibs in another one, and all the logic is in a shared helper method.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
@@ -564,6 +564,8 @@ class SierraRest extends AbstractBase implements TranslatorAwareInterface,
                 'records' => []
             ];
         }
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks, @EreMaijala, I have requested that some books be checked out to my local Sierra account so that I can test this. I'll follow up once I am able to try it out.

demiankatz:
I have received the requested test records and can confirm that this is significantly faster (with 3 checked out items, average load time for the checked out screen went down by a full second, from about 3 seconds to about 2 seconds; I imagine the savings are even more significant with longer lists). I'm merging this now!

PR #1919, opened 2021-04-14
FOLIO: use date converter for requiredBy formatting in placeHold

demiankatz:
@EreMaijala pointed out that FOLIO was not using the date converter in placeHold(), which I think would make the code a bit more reliable and consistent with other drivers. This PR is an attempt to standardize the date formatting, but I have not had an opportunity to test it. I am hoping that @ihardy and/or @mis306lu might have time to take a look. Please let me know what you think!


demiankatz:
Thanks to @ihardy for confirming that this works.

PR #1920, opened 2021-04-15
FOLIO: Access to Location Codes in ILS Driver

### LINKED ISSUE #110: Need Access to Location Codes in FOLIO Driver
In order to implement the conditional logic required for the display of service links in VuFind, we need access to the location codes in FOLIO. These codes follow the [4 tier location hierarchy in FOLIO](https://wiki.folio.org/display/FOLIOtips/Settings+-+Tenant+-+Location+Setup) and look like this: `UC/HP/ASR/LawASR`. Codes are more granular and precise than the discovery display name that is used for the string literal. For example, `LawASR` which maps to the discovery display name of `Mansueto` is just one of many others all of which map to `Mansueto`. In some cases we choose to display service links for some codes but not others even though they might both fall under the same discovery display name. Codes are also more reliable since they are less likely to change when a display string might.

To make this possible, we will need to request a new optional `location_code` key in the output of the VuFind [getHolding ILS driver method](https://vufind.org/wiki/development:plugins:ils_drivers#getholding).

bbusenius:
Fixes [issue #110](https://github.com/uchicago-library/vufind/issues/110).

**Changes in this request:**
- Add a new optional `location_code` key to the output of `getHolding` in the FOLIO ILS driver
- Change the name of the `getLocationName` method to `getLocationData` and alter the signature. Instead of returning a location name string, return an array with location name and code.
- Add location code to location caching.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -461,7 +461,7 @@ class Folio extends AbstractAPI implements
             ) as $location) {
                 $name = $location->discoveryDisplayName ?? $location->name;
                 $code = $location->code;
-                $locationMap[$location->id] = [$name, $code];
+                $locationMap[$location->id] = compact('name', 'code');
             }
         }
         $this->putCachedData($cacheKey, $locationMap);
@@ -478,11 +478,10 @@ class Folio extends AbstractAPI implements
     protected function getLocationData($locationId)
     {
         $locationMap = $this->getLocations();
-        $locationName = '';
-        $locationCode = '';
+        $name = '';
+        $code = '';
         if (array_key_exists($locationId, $locationMap)) {
-            $locationName = $locationMap[$locationId][0];
-            $locationCode = $locationMap[$locationId][1];
+            return $locationMap[$locationId];
         } else {
             // if key is not found in cache, the location could have
             // been added before the cache expired so check again
@@ -491,12 +490,12 @@ class Folio extends AbstractAPI implements
             );
             if ($locationResponse->isSuccess()) {
                 $location = json_decode($locationResponse->getBody());
-                $locationName = $location->discoveryDisplayName ?? $location->name;
-                $locationCode = $location->code;
+                $name = $location->discoveryDisplayName ?? $location->name;
+                $code = $location->code;
             }
         }
 
-        return [$locationName, $locationCode];
+        return compact('name', 'code');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -459,8 +459,9 @@ protected function getLocations()
             foreach ($this->getPagedResults(
                 'locations', '/locations'
             ) as $location) {
-                $locationMap[$location->id]
-                    = $location->discoveryDisplayName ?? $location->name;
+                $name = $location->discoveryDisplayName ?? $location->name;
+                $code = $location->code;
+                $locationMap[$location->id] = [$name, $code];
```
demiankatz: If you use `compact('name', 'code')` here, you'll get an associative array, which might make the rest of the code more readable (instead of using hard-coded 0/1 indexes).

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -489,10 +492,11 @@ protected function getLocationName($locationId)
             if ($locationResponse->isSuccess()) {
                 $location = json_decode($locationResponse->getBody());
                 $locationName = $location->discoveryDisplayName ?? $location->name;
+                $locationCode = $location->code;
             }
         }
 
-        return $locationName;
+        return [$locationName, $locationCode];
```
demiankatz: Again, it might be worth considering returning an associative array here to improve readability. In fact, up above, you could just return `$locationMap[$locationId]` directly up on line 484 instead of decomposing it and then putting it back together again.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -572,6 +578,7 @@ public function getHolding($bibId, array $patron = null, array $options = [])
                     'indexes' => $holdingsIndexes,
                     'callnumber' => $holding->callNumber ?? '',
                     'location' => $locationName,
+                    'location_code' => $locationCode,
```
demiankatz, line 580: I notice that location_code is not a documented field for getHolding(), yet two existing ILS drivers are already using it. I have added it to the spec as an optional field.

demiankatz (commented):
Thanks, @bbusenius; I've left a couple of suggestions and comments. I'd also be interested to hear if @ihardy or @mis306lu have any thoughts on this, though it looks pretty uncontroversial to me.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
bbusenius: Ah yes. I was on the fence about doing this anyhow. This confirms it for me. Done.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
bbusenius: Thanks! This is what I was hoping you would do. We did this with OLE as well.

demiankatz:
Thanks for the updates, @bbusenius, this looks great to me now. I'll post it on the FOLIO VuFind integration Slack channel (which I can also invite you to, if you're on the FOLIO Slack and not already in the channel) to see if there is any other feedback, but I suspect it can be merged soon.

PR #1921, opened 2021-04-16
Do not code various non-Book formats as Book.

demiankatz:


demiankatz:
This has been reviewed and tested by Villanova's resource management team, so I'm feeling confident enough to merge it now.

PR #1922, opened 2021-04-19
Alma resolver: Add support for displaying Unpaywall service.

EreMaijala:
Also includes any other getOpenAccessFullText type services and changes the default display title of packages to package_display_name with package_public_name as the fallback.

Looks like Unpaywall is a it special and only has a few fields. It also uses a new service type.

### DIFF
--- a/module/VuFind/src/VuFind/Resolver/Driver/Alma.php
+++ b/module/VuFind/src/VuFind/Resolver/Driver/Alma.php
@@ -109,9 +109,8 @@ class Alma extends AbstractBase
                     continue;
                 }
             }
-            $serviceType = $this->mapServiceType(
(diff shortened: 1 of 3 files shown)


demiankatz:
Thanks, @EreMaijala; this looks straightforward to me, and I have no objection to merging it, but I also have no means of testing it directly. Is there anyone else you'd like to ask for a review, or should I go ahead and merge?

EreMaijala:
@demiankatz Sorry, I've no idea who could test this. That said, I believe this should be quite safe.

demiankatz:
No worries. I'll merge now, and if somebody runs into a problem, they can let us know. :-)

PR #1923, opened 2021-04-19
Indicate that ContainerExceptions must also be Throwable.

demiankatz:
This is a minor documentation improvement intended to reduce Psalm warnings (see #1764).

### DIFF
--- a/module/VuFind/src/VuFind/AjaxHandler/AbstractIlsAndUserActionFactory.php
+++ b/module/VuFind/src/VuFind/AjaxHandler/AbstractIlsAndUserActionFactory.php
@@ -56,7 +56,7 @@ class AbstractIlsAndUserActionFactory
      * @throws ServiceNotFoundException if unable to resolve the service.
      * @throws ServiceNotCreatedException if an exception is raised when
      * creating a service.
-     * @throws ContainerException if any other error occurs
+     * @throws ContainerException&\Throwable if any other error occurs
      *
      * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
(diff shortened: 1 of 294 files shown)


demiankatz:
@xmorave2, your work with PHPStan has inspired me to try to make more progress with Psalm as a complement. This PR is an attempt to fix one of the most prevalent issues reported by Psalm... but the fix doesn't work (see my recent notes on #1764). I think that's due to a bug in Psalm rather than an error in this change, but I'm interested in your opinion. Do you see anything wrong with what I am doing here? Is it worth merging this regardless of whether it resolves Psalm errors? I'm hopeful that if we do this now, it will benefit Psalm in the long run once the relevant bug gets fixed.

xmorave2:
@demiankatz 


https://docs.phpdoc.org/3.0/guide/references/phpdoc/types.html

xmorave2 (approved):
@demiankatz, according to PHPdoc there not a thing like intersection type: https://docs.phpdoc.org/3.0/guide/references/phpdoc/types.html

But PHPstan and Psalm both do recognize them, so it is OK for us to use them. And as this change is done in the manner the intersection types were designed, I am for merging this.

demiankatz:
Thanks for the review, @xmorave2; I'll go ahead and merge this now, and hopefully Psalm will fix their support in `@throws` soon (feel free to watch that issue if you want to add a little more pressure). Interestingly, I also see that an RFC just opened for native intersection type support in PHP: https://github.com/Girgias/intersection-types

xmorave2:
@demiankatz That's really interesting, thanks for the info

PR #1924, opened 2021-04-21
Add ability to use own (local) identifiers for getting covers

xmorave2:
ObalkyKnih.cz could serve covers also for records without any standard identifier -  in that case the identifier is in format "SIGLA-recordid" where is sigla is library identifier used in Czech Republic and recordid is record identifier in local library system


demiankatz:
This looks good to me! Thanks!

PR #1925, opened 2021-04-21
Deprecate PHP-based LESS compiler.

demiankatz:
The PHP-based LESS compiler is less accurate than the Node-based one. We have maintained it for many years to avoid adding an additional development dependency, but it may be causing more problems than it is solving at this point. This PR adds a deprecation message to encourage users to switch to Grunt.

TODO:
- [x] Discuss whether this deprecation is appropriate, or whether we can/should improve the reliability of the PHP-based tool
- [x] Review wiki documentation about Grunt and LESS; update/clarify as needed
- [x] If we want to remove this after deprecation, we should add a note to [VUFIND-1474](https://vufind.org/jira/browse/VUFIND-1474)

### DIFF
--- a/module/VuFindConsole/src/VuFindConsole/Command/Util/CssBuilderCommand.php
+++ b/module/VuFindConsole/src/VuFindConsole/Command/Util/CssBuilderCommand.php
@@ -27,7 +27,7 @@
  */
 namespace VuFindConsole\Command\Util;
 
-use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use VuFindTheme\LessCompiler;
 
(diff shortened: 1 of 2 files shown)


ekilfeather:
Happy to use Grunt as default

crhallberg (approved):
I'm in favor of this! There will need to be a future conversation about how to improve the node-based tooling but deprecating PHP can happen now and is a good first step.

EreMaijala (approved):

ckaz:
We use grunt and SCSS which works very well. Cheers!

liowalter:
grunt is fine for us as well !

xmorave2:
We've just switched to grunt too, so no problem here

demiankatz:
There have been no objections to this deprecation, so I am going to go ahead and merge it. I have also updated the relevant JIRA ticket (so we remember to remove this in 9.0) and made some minor updates to the wiki.

I am hoping that as he reviews VuFind's use of Javascript tooling, @crhallberg can further revise the LESS page in the wiki; since we may be overhauling things significantly, it doesn't seem worth drastically revising the documentation until our plans are more firm.

I also wonder if we should remove the "active" compilation option for LESS, which I believe relies on the PHP compiler (and which has not been heavily used or tested, as far as I know). What do you think, @crhallberg? Perhaps that merits another separate PR.

demiankatz:
One last thought: should we add Node version to the [requirements page](https://vufind.org/wiki/installation:requirements), noting that it is only required for development purposes?

PR #1926, opened 2021-04-21
Icon Helper

crhallberg:
Starting a new theme atom to allow easier use and customization of icons throughout VuFind.

- [x] Abstracting similar methods (ie. using icon font or [SVG sprites for Font Awesome](https://fontawesome.com/how-to-use/on-the-web/advanced/svg-sprites))
    - [x] maybe `use=sprite:FontAwesome`
- [x]  Dependency insertion (ie. font-awesome.css)
- [x] Comment theme.config.php
- [x] Write test for new ThemeInfo functionality
- [x] Add remix link to icon wiki page, if using
- [x] Write unit test for new helper


module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Constructor
+     *
+     * @param Config $config Icons config (icons.ini)
+     */
+    public function __construct(Config $config)
+    {
+        $this->config = $config->Config;
+        $this->iconMap = $config[$config->Config->use];
```
demiankatz: Might be worth having some defaults/fallbacks here, like:

<pre>
$this->config = $config->Config ?? new Config([]);
$this->iconMap = $config[$config->Config->use ?? 'FontAwesome'] ?? new Config([]);
</pre>

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,93 @@
...
+        $mappedIcon = $this->iconMap[$icon] ?? $icon;
+
+        $attrs = '';
+        $escAttr = $this->getView()->plugin('escapeHtmlAttr');
+        foreach ($extra as $key => $val) {
+            $attrs .= ' ' . $key . '="' . $escAttr($val) . '"';
+        }
+
+        return $this->getView()->render(
+            'icons/' . $this->config->use,
```
demiankatz: Might be worth moving icon templates under the Helpers/ folder (i.e. Helpers/icons/*.phtml) so there's no conflict if somebody somewhere has an IconsController.

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
demiankatz: Of course, I also wonder if the configuration belongs somewhere else entirely -- like in theme.config.php, since I could imagine we might want different icon systems in different themes.

Also, is it possible that picking one icon library isn't flexible enough? Would it make more sense to rearrange the configuration so that there's a default icon library, but you can override individual icons to use different systems?

module/VuFind/src/VuFind/View/Helper/Root/IconFactory.php
```diff
@@ -0,0 +1,74 @@
+<?php
+/**
+ * Captcha helper factory.
```
demiankatz: Copied and pasted comment.

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,93 @@
+<?php
+/**
+ * Make link view helper
```
demiankatz: Copied and pasted comment.

module/VuFind/src/VuFind/View/Helper/Root/IconFactory.php
demiankatz: (Yes, it is true that I copied and pasted my comment about your copied and pasted comments in order to comment on them twice).

themes/root/theme.config.php
```diff
@@ -1,6 +1,7 @@
 <?php
 return [
     'extends' => false,
+    'css' => ['icons.css'],
```
demiankatz: While I see the value in having some kind of global default, I also wonder if it would be better for now to move this into bootstrap3. Adding an additional HTTP transaction for a separate CSS file on every page just for icons seems unnecessary, and out of keeping with other aspects of our CSS management.

themes/root/templates/icons/Remix.phtml
```diff
@@ -0,0 +1,3 @@
+<svg class="icon--svg"<?=$this->attrs ?> aria-hidden="true">
```
demiankatz: If we go forward with including Remix icons, we should update our icons page in the wiki to give them credit, and perhaps also include a copy of their license in the directory with the icons. Might be worth adding a TODO comment above.

themes/root/templates/icons/FontAwesome.phtml
```diff
@@ -0,0 +1 @@
+<i class="fa fa-<?=$this->icon ?>"<?=$this->attrs ?> aria-hidden="true"></i>
```
demiankatz: We should probably escapeHtmlAttr() the icon value here (and also in Remix.phtml) just to be on the safe side.

demiankatz (changes requested):
Thanks, Chris, this is a great idea. I have a whole bunch of thoughts, some big and some small. This is just a brainstorm dump, so feel free to disagree with me on some of these ideas. :-)

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
crhallberg: In consideration related to both of these points:
- icons.ini per theme (like languages per theme)
- `use=FontAwesome,Bootstrap,Tall Terry's Excellent Icons`

We could also look in `theme.config.php` for overrides but I thought the ini file would be cleaner and easier to edit.

themes/root/templates/icons/Remix.phtml
crhallberg: I'm picking on Remix right now because they have a lot of options for import (svg, sprite, font) so I can work with two robust fonts. I do like it though and crediting them is the least we could do.

themes/root/templates/icons/FontAwesome.phtml
crhallberg: Is it enough to escape them in the helper? https://github.com/vufind-org/vufind/pull/1926/files#diff-b10064b350845d825e6677a96d797f7c11bc5950d0b7ffe061176b4d2e440d14R83-R86

demiankatz: Yes, the attributes are fine... it's $this->icon that doesn't appear to be getting escaped anywhere. (Admittedly, it shouldn't NEED escaping, but it doesn't hurt in case of weird edge cases or misconfigurations).

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
demiankatz: I'm not completely opposed to having an icons.ini per theme, but putting .ini files in themes would be a new thing we haven't done before. Having a section in themes.config.php would be more consistent with everything else we currently do in themes. You could likely expose the settings through the VuFindTheme\ResourceContainer.

I don't think a comma-separated list of icon sets is going to work, because how would the code know which one to use? I was thinking of something more like:

<pre>
[General]
defaultIconSet = FontAwesome

[Overrides]
favorite = Bootstrap
delete = "Tall Terry's Excellent Icons"
</pre>

So when we look up a particular icon, the request gets routed to the appropriate library. But maybe I'm making things more complicated than they need to be... just trying to aim for maximum flexibility.

themes/root/theme.config.php
crhallberg: I agree this is problematic. I also worry about including external dependencies (like FontAwesome's CSS). It would be safe to bundle thru theme.config.php because it will affect nearly pages but if we can make it a one-stop config edit, I'd rather do that.

demiankatz: In theory, you could dynamically include CSS through the icon-specific templates, since those inclusions will be deduplicated by the framework, and then we'd only load the styles when they are needed... but that might also be incredibly inefficient on pages with lots of icons, since it would trigger lots of deduplication work. Just a thought for consideration!

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
demiankatz: You marked this one as resolved, but it still says "make link" instead of "icon." Did something not get pushed?

config/vufind/icons.ini
```diff
@@ -0,0 +1,16 @@
+[Config]
+use=FontAwesome
+
+[FontAwesome]
+favorites=Remix:star-fill
```
demiankatz: I think this is an effective way of making things more functional, but I'm still not convinced that we want this to come from an .ini file. What if themes.config.php just included a section like:

<pre>
'icons' => [
    'defaultSet' => 'FontAwesome',
    'mappings' => [
        'favorites' => 'Remix:star-fill',
        'my-account' => 'user',
        'search' => 'Remix:search-line',
    ]
]
</pre>

As long as we set up the theme config loading so that it appropriately merges settings through the theme inheritance tree, this would make it possible to swap out individual icons or make larger-scale changes on a theme by theme basis. People could even publish mix-ins designed ONLY to change the icon system. It just feels to me more in keeping with how themes work. But again, happy to discuss further. :-)

crhallberg: Can't argue with that! I've moved it to theme.config.php.

themes/root/theme.config.php
crhallberg: I've moved the dependency loading to the icon templates but I was also concerned about deduplication work. It turns out, `appendStylesheet` just appends to the end of Container/ArrayObject and the deduplication only happens once when it's time to render.

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,100 @@
...
+    protected $iconMap;
+
+    /**
+     * Constructor
+     *
+     * @param Config $config Icons config (icons.ini)
+     */
+    public function __construct(\VuFindTheme\ThemeInfo $themeInfo)
+    {
+        $themeConfig = $themeInfo->getThemeInfo()[$themeInfo->getTheme()];
```
demiankatz: I think we're going to need to make this more robust (with proper inheritance/merging), or else it's only going to work when the full configuration exists in the active theme (and not a parent of it). This should be fine for prototyping purposes, but it's worth adding a TODO checkbox about.

themes/bootstrap3/theme.config.php
```diff
@@ -61,5 +61,13 @@
             'layoutClass' => 'VuFind\View\Helper\Bootstrap3\LayoutClass',
             'search' => 'VuFind\View\Helper\Bootstrap3\Search'
         ]
+    ],
+    'icons' => [
```
demiankatz: Another TODO item: once we're happy that we have all the functionality we need, we should add comments here (or somewhere) explaining the config format.

themes/root/theme.config.php
demiankatz: That sounds like good news, though it probably wouldn't hurt to do some profiling when this is all ready.

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
crhallberg: Does `getThemeInfo()` not handle inheritance? It seems I misread my test, I thought inheritance was being handled. Is there no mechanism to get a fully inherited theme config of the active theme?

demiankatz: `getThemeInfo` retrieves information about all of the themes in the inheritance chain, but it doesn't automatically merge them together. When you call `$themeInfo->getThemeInfo()[$themeInfo->getTheme()]`, that is going to give you the contents of theme.config.php from the current active theme... but if the data you're looking for is not in there, you'll need the use the 'extends' and 'mixins' values from that array to look up the next parent/mixin in line, and so on.

It might make sense to add a utility method to ThemeInfo like getMergedConfig() to make it easier to fetch this kind of data in one call. I can take a stab at it if that would be helpful.

crhallberg: That would be helpful! I think there's some code in the theme compiler or PHP less compiler that may be helpful.

demiankatz: I'll see what I can do!

demiankatz: See 74c09bf -- I think this does the trick!

demiankatz: Now I just need to write a unit test so our coverage in VuFindTheme doesn't drop. I've added a TODO checkbox and will take care of it closer to merge time.

demiankatz:
Thanks for all the progress on this, @crhallberg! Let me know if/when you'd like me to give it another review.

demiankatz:
(Also note that there currently appears to be a very minor CI failure).

crhallberg:
One of the potential major steps is whether or not we want to apply this to our current themes (just bootstrap3 essentially) or if we should use this for future features and templates.

demiankatz:
@crhallberg, I would be in favor of applying this to the existing themes, because I think it will make the templates more readable (not to mention easier to port when the new theme is under development).

crhallberg:
Should that be a separate pull request?

demiankatz:
@crhallberg, I don't have a strong opinion about whether it's better to do that here or as a separate PR, but of course we definitely want to be sure the helper is finalized and we're happy with the interface before you take the time to apply it to a ton of templates. If you'd like another review from me, just let me know when you're ready.

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
```diff
@@ -213,6 +213,47 @@ public function testFindInThemes()
...
+        // Child with parents merged in
+        $ti->setTheme('child');
+        $childJS = $ti->getMergedConfig('js');
+        $this->assertEquals(['extra.js', 'hello.js'], $childJS);
+        // recursive
+        $childHelpers = $ti->getMergedConfig('helpers');
+        $this->assertEquals(
+            ['fooOverrideFactory', 'fooFactory'],
+            $childHelpers['factories']['foo']
+        );
```
crhallberg, line 243: It made sense to me that if we're merging values (like with JS and CSS) that the child ones should come before the parent ones. Is there a reason why we should switch this?

demiankatz: I would normally expect the parent values to come before the child values, since the child values are often adding to the parent values, and may depend upon the things that precede them. However, I'm not sure if this expectation is relevant to what we're doing here, since I think in the context of icons, we're either overriding specific values or else adding things into a list where order is unimportant. I'm just speaking to what I would expect in terms of CSS/JS behavior, which may not really apply to this.

themes/bootstrap3/theme.config.php
crhallberg: How does the below sound?

demiankatz: This seems good to me. One suggestion, though: I don't think prefix needs to be a font-specific setting. It could be a prefix that gets applied to all icon names within the set. This could apply to filenames for files, or anchors for SVGs. It just defaults to empty. This way, you can use it to improve readability when everything has the same prefix, but you can also ignore it if you don't need it. You could potentially apply the prefix to the icon name inside the view helper so that you don't have to deal with it in the templates at all. Of course, I don't insist on this -- just a thought. What do you think?

demiankatz:
@crhallberg, the bug fixes @EreMaijala just made in #1953 raise a couple of interesting edge cases that may be worth thinking about in relation to this work: some icons may need to be combined with extra attributes (like `fa-spin`), and some icon values may need to get passed up into Javascript code, so we may need a mechanism for that somewhere. I don't think any of that necessarily changes anything you've done so far -- but it just happened to be a timely reminder of these icon-related issues.

crhallberg:
I have a few ideas on how to resolve the `fa-spin` issue that I'm sure will come up during the icon replacement. Making a data-icon-spin is my top choice right now.

crhallberg:
One potential issue is switching the default icon set can lead to missing icons. For example, BS3 uses FontAwesome and Bootprint uses Fugue image icons. If Bootprint doesn't specify/override an icon, it will look for the same icon in the default set:
`` which doesn't exist.

This can be solved by specifying `FontAwesome:icon` in Bootstrap3, catching all icons in a verbose safety net. Unless there's a clever code way to map all icons to the default set in the theme config they were set in.

demiankatz:
@crhallberg, I agree that changing the default icon set is likely to be a risky operation. I don't really envision somebody going in and overriding the default in a sub-theme, because it's too hard to control. I think the main purpose of the default icon set is to document what is being used within a theme, and to provide a useful baseline when building a whole new theme from the ground up.

Obviously it would be great if we could create a prioritized list of icon sources and have the code magically pick the most appropriate one, but I don't think there's any practical way to implement that since in most cases, it's not easy to check whether or not an icon is present (at least, not without doing some pretty expensive logic to parse files, which would be overkill).

demiankatz:
@crhallberg, looks like the build is failing due to a minor php-cs-fixer thing. Is this otherwise ready for another review? Let me know when you're ready and I'll give it another fresh look.

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,151 @@
...
+     */
+    public function __construct(ThemeInfo $themeInfo)
+    {
+        $this->config = $themeInfo->getMergedConfig('icons', true);
+        $this->defaultSet = $this->config['defaultSet'] ?? 'FontAwesome';
+        $this->iconMap = $this->config['aliases'] ?? [];
+    }
+
+    /**
+     * Map icon to set. Add prefix, return with set and template
```
EreMaijala: Add periods.

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,151 @@
...
+     * @param string $name  Which icon?
+     * @param array  $extra Just extra HTML attributes for now
+     *
+     * @return string
+     */
+    public function __invoke(string $name, $extra = []): string
+    {
+        [$icon, $set, $template] = $this->mapIcon($name);
+
+        // Compile attitional HTML attributes
```
EreMaijala: *additional

themes/bootstrap3/theme.config.php
```diff
@@ -61,5 +61,42 @@
...
+                'template' => 'font',
+                'prefix' => 'fa fa-',
+                'src' => 'vendor/font-awesome.min.css',
+            ],
+            'Remix' => [
+                'template' => 'svg-sprite',
+                'prefix' => 'ri-',
+                'src' => 'remixicon.symbol.svg',
+            ],
+        ],
```
EreMaijala: An example (even commented-out) for image-based set would be nice.

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * Returns inline HTML for icon
+     *
+     * @param string $name  Which icon?
+     * @param array  $extra Just extra HTML attributes for now
+     *
+     * @return string
+     */
+    public function __invoke(string $name, $extra = []): string
```
EreMaijala: This method does potentially quite a bit of work and could use some caching of results. Otherwise e.g. the format icons on a results list can cause quite a bit of repeated work.

EreMaijala (changes requested):
This all looks pretty good to me. I only noticed some minor issues, but then there's the question of performance, and since the helper does quite a bit of work to render an icon, I'd strongly suggest caching the results (with a key that takes into account the extras) to avoid repeating it.

themes/bootstrap3/theme.config.php
crhallberg: This is something I'll do over in the application PR (#1962).

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,184 @@
...
+    {
+        $cacheKey = $this->cacheKey($name, $extra);
+        $cached = $this->cache->getItem($cacheKey);
+
+        if ($cached == null) {
+            [$icon, $set, $template] = $this->mapIcon($name);
+
+            // Compile additional HTML attributes
+            $escAttr = $this->getView()->plugin('escapeHtmlAttr');
```
LuomaJuha: Would it be better to assign $escAttr during construct so it would not be required as a parameter to a function? like $this->escAttr = ...

demiankatz: I agree; since we have a custom factory already anyway, we might as well make this a hard dependency. That will make test-writing easier, too.

crhallberg: Done!

themes/root/templates/icons/Remix.phtml
demiankatz: I see that the Remix file itself contains licensing information, so I don't think we need to add a separate license file. However, where did we land on the use of Remix? I see that right now, the configuration is actually using Remix for a couple of things. Was this only for testing/demo purposes, or is this a real change that we want to merge?

themes/bootstrap3/theme.config.php
demiankatz: I see that the Fugue example in bootprint3 in #1962 covers this situation, so I am resolving this thread now.

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,219 @@
...
+     * Prevent extra work by only appending the stylesheet once
+     *
+     * @var boolean
+     */
+    protected $styleAppended = false;
+
+    /**
+     * Constructor
+     *
+     * @param ThemeInfo      $themeInfo    Theme info helper
```
demiankatz: Since we are only using ThemeInfo to fetch configuration, I don't think we should make it a dependency of this class. Instead, we should fetch the configuration in the factory, and pass it here. That will make it easier to write a unit test, and will make things generally more flexible.

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,219 @@
...
+     * @param CacheManager   $cacheManager Cache manager instance
```
demiankatz: Similarly, since we are just using the CacheManager to fetch a cache, we could also do that work in the factory and pass only the cache object here, which once again gives us a less specific dependency for easier testing and greater flexibility in the future.

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,219 @@
...
+        }
+        ksort($extra);
+        return $name . '+' . md5(json_encode($extra));
+    }
+
+    /**
+     * Returns inline HTML for icon
+     *
+     * @param string $name  Which icon?
+     * @param array  $extra Just extra HTML attributes for now
```
demiankatz: Should we make this comment more final and confident, or do you still think this might be used for something more in future?

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -189,6 +189,56 @@ public function getThemeInfo()
...
+        $currentTheme = $this->getTheme();
+        $allThemeInfo = $this->getThemeInfo();
+
+        /**
+         * Assume a parent value 'a' and a child value 'b'
+         *
+         * Using array_merge (default) will merge them into ['b', 'a']
+         * Using array_replace ($flatten = true) will merge them into 'b'
+         *
+         * We're using an anonymous funcyion here to swap the arguements in the
```
demiankatz: Typos: funcyion and arguements

themes/bootstrap3/templates/header.phtml
```diff
@@ -18,7 +18,8 @@
           <?php $cart = $this->cart(); if ($cart->isActive()): ?>
             <li id="cartSummary">
               <a id="cartItems" data-lightbox title="<?=$this->transEscAttr('View Book Bag')?>" href="<?=$this->url('cart-home')?>">
-                <i class="fa fa-suitcase" aria-hidden="true"></i> <strong><?=count($cart->getItems())?></strong> <?=$this->transEsc('items')?>
+                <?=$this->icon('cart') ?>
```
demiankatz: Is there a reason this change is here, rather than in #1962?

themes/bootstrap3/templates/search/searchbox.phtml
```diff
@@ -94,7 +94,8 @@
     <?php elseif ($handlerCount == 1): ?>
       <input type="hidden" name="type" value="<?=$this->escapeHtmlAttr($handlers[0]['value'])?>" />
     <?php endif; ?>
-    <button type="submit" class="btn btn-primary"><i class="fa fa-search" aria-hidden="true"></i> <?=$this->transEsc("Find")?></button>
+    <button type="submit" class="btn btn-primary">
```
demiankatz: Should this change also be moved to #1962?

themes/bootstrap3/theme.config.php
```diff
@@ -61,5 +61,42 @@
...
+            /**
+             * Icons can be assigned or overriden here
+             *
+             * Format: 'icon' => [set:]icon
+             *
+             * Icons assigned without set will use the defaultSet (above)
+             */
+            'favorites' => 'Remix:star-fill',
+            'my-account' => 'user-circle',
+            'search' => 'Remix:search-line',
```
demiankatz: I think I already mentioned this in a different thread, but just to highlight it: do we intend to use Remix here "for real" or is this just an example? If it's an example, we should remove it or comment it out. If it's for real, we should add Remix to our icons wiki page. I've added a TODO checkbox to track the need to decide one way or the other.

themes/root/templates/Helpers/icons/font.phtml
```diff
@@ -0,0 +1,10 @@
+<?php
+    $this->headLink()->appendStylesheet($this->src);
+
+    // Append extra classes
+    if (isset($this->extra['class'])) {
+        $this->icon .= ' ' . $this->extra['class'];
+        $this->attrs = preg_replace('/ class="[^"]+"/', '', $this->attrs);
+    }
+?>
+<span class="icon--font <?=$this->icon ?>"<?=$this->attrs ?> role="img" aria-hidden="true"></span>
```
demiankatz: Should we `escapeHtmlAttr` the `$this->icon` value here, just to be on the safe side?

themes/root/templates/Helpers/icons/svg-sprite.phtml
```diff
@@ -0,0 +1,3 @@
+<svg class="icon--svg"<?=$this->attrs ?> aria-hidden="true">
+    <use xlink:href="<?=$this->imageLink($src) ?>#<?=$this->icon ?>"></use>
```
demiankatz: Another case where I wonder if we should `escapeHtmlAttr` the `$this->icon` value out of caution.

demiankatz (changes requested):
Thanks for all the work on this, @crhallberg -- I think it's looking really good, but I have a few more, mostly minor, suggestions. I think once these are addressed, the last detail is to write a unit test... and doing that should be a bit easier if you make the constructor changes I suggest here.

themes/bootstrap3/theme.config.php
crhallberg: I do intend to make a Remix icon set but in the other PR.

themes/bootstrap3/templates/header.phtml
crhallberg: These were for tests and were accidentally committed. Removing now.

themes/bootstrap3/theme.config.php
demiankatz: Thanks!

demiankatz:
@crhallberg, thanks for the changes! Should we also delete remixicon.symbol.svg from this PR since we're not using it? Beyond that, am I correct in thinking that the last outstanding item is writing a unit test for the Icon helper? If so, do you want me to work on that when time permits?

demiankatz:
@crhallberg, I've just created a unit test that I think covers all the major cases, plus I've adjusted the constructor parameters as proposed earlier as well as fixing typos. The only outstanding question from me relates to the $extra parameter comment -- see above.

demiankatz:
@crhallberg, it looks like your latest commit has somehow broken my unit test. Can you investigate?

demiankatz:
Thanks for the fixes, @crhallberg. The build was still failing due to some minor php-cs-fixer issues, so I merged in the latest dev branch and applied style fixes. Now that everything is passing, I think this may be ready to merge. Do you agree? Any final thoughts from @EreMaijala, @LuomaJuha, @aleksip or others?

module/VuFind/src/VuFind/View/Helper/Root/IconFactory.php
```diff
@@ -0,0 +1,83 @@
...
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options sent to factory.');
+        }
+        $config = $container->get(\VuFindTheme\ThemeInfo::class)
+            ->getMergedConfig('icons', true);
+        $cache = $$container->get(\VuFind\Cache\Manager::class)
```
EreMaijala: extra $

module/VuFind/src/VuFind/View/Helper/Root/Icon.php
```diff
@@ -0,0 +1,219 @@
...
+     */
+    protected $styleAppended = false;
+
+    /**
+     * Constructor
+     *
+     * @param array            $config   Icon configuration
+     * @param StorageInterface $cache    Cache instance
+     * @param EscapeHtmlAttr   $escAttr  EscapeHtmlAttr view helper
+     * @param HeadLink         $headLink For stylesheet appending
```
EreMaijala: Nit: other comments tell what the parameter is, this one only what it's for.

EreMaijala (changes requested):
When I asked about caching I only thought about a non-persistent runtime cache (similarly to what was done with ClassBasedTemplateRendererTrait), though failed to convey this idea to you. Now I'm a bit concerned that the lookups from the persistent cache may actually be slower than computing the result especially for icons that are not repeated.

demiankatz:
@crhallberg and @EreMaijala: I wonder if it's worth changing the factory to create an [in memory cache adapter](https://docs.laminas.dev/laminas-cache/storage/adapter/#memory-adapter) instead of pulling from the cache manager. This would reduce any storage/performance overhead of using a more persistent cache, but it would keep all the code as-is and allow flexibility in the future.

EreMaijala:
@demiankatz That'd work for me!

demiankatz:
Great! If we go that route, we need to be sure to take into account the work done in #2054 to modernize the use of laminas-cache. We should make sure the latest dev code is merged here, then explicitly add the appropriate adapter package to package.json and use the `\Laminas\Cache\Service\StorageAdapterFactory::class` service to build the adapter.

I can help with all this next week if @crhallberg doesn't have time.

demiankatz:
I've addressed @EreMaijala's review comments and converted to in-memory caching.

I had trouble testing this in real-life since the configuration is all over in #1962. When I just used the icon helper in a random template without adding any configuration, I got a fatal error because it tried to use the set name as the template name, which seems wrong to me. I addressed this in 49f3546, but I'm not sure if that's really the best approach or not.

@crhallberg and @EreMaijala, are you happy with this now? Should we merge it and continue work on #1962 now, or do you think further work is needed first?

crhallberg:
I think your solution makes sense!

EreMaijala (approved):
Looks good to me!

demiankatz:
Thanks, @crhallberg and @EreMaijala -- I'm going ahead and merging this now, since it just adds new functionality and is unlikely to cause any problems. If further revisions are needed, we can work on those as part of #1962.

PR #1928, opened 2021-04-22
Simplify/standardize I18n configuration.

demiankatz:
This pull request extracts more useful ideas from #1232:

- It loads standard Laminas\I18n configurations, exposing some additional view helpers and filters
- It makes the translator factory more granular (e.g. by giving the ExtendedIni loader its own separate factory), and applies VuFind customizations through a delegator rather than a subclass
- Configuration now applies customizations directly to the lower-level Translator object instead of the MVC-specific proxy
- It simplifies some unnecessary complexities in the DisplayLanguageOption view helper and its factory
- It relies on SlmLocale to set up the translator locale instead of setting it explicitly in VuFind-land code

Note one minor annoyance which I don't particularly like: the standard Laminas `translate` view helper has a different interface than VuFind's `translate` view helper. VuFind's configuration should override the Laminas helper with the VuFind version, but due to the existence of some differently-cased aliases, it might be possible to access the wrong helper and become confused. I'm not sure if it's worth any effort to eliminate this possibility, but it's at least something we should be aware of in the future.

TODO:
- [x] Run full test suite.
- [x] Add note to changelog when merging.

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/DisplayLanguageOptionFactory.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/DisplayLanguageOptionFactory.php
@@ -65,6 +65,9 @@ class DisplayLanguageOptionFactory implements FactoryInterface
         if (!empty($options)) {
             throw new \Exception('Unexpected options sent to factory.');
         }
-        return new $requestedName($container->get(Translator::class));
+        $translator = $container->get(Translator::class);
+        // Add a special locale used just for this plugin:
+        $translator->addTranslationFile('ExtendedIni', null, 'default', 'native');
+        return new $requestedName($translator);
     }
 }
--- a/config/application.config.php
+++ b/config/application.config.php
@@ -2,8 +2,9 @@
 
 // Set up modules:
 $modules = [
-    'Laminas\Form', 'Laminas\Router', 'LmcRbacMvc', 'SlmLocale', 'VuFindTheme',
-    'VuFindSearch', 'VuFind', 'VuFindAdmin', 'VuFindApi'
+    'Laminas\Form', 'Laminas\Router', 'LmcRbacMvc', 'Laminas\I18n',
+    'Laminas\Mvc\I18n', 'SlmLocale', 'VuFindTheme', 'VuFindSearch', 'VuFind',
+    'VuFindAdmin', 'VuFindApi'
 ];
 if (!extension_loaded('intl')) {
     // Disable SlmLocale module if intl extension is missing:
--- a/module/VuFind/config/module.config.php
+++ b/module/VuFind/config/module.config.php
@@ -441,10 +441,12 @@ $config = [
             'VuFindSearch\Service' => 'VuFind\Service\SearchServiceFactory',
             'Laminas\Db\Adapter\Adapter' => 'VuFind\Db\AdapterFactory',
             'Laminas\Http\PhpEnvironment\RemoteAddress' => 'VuFind\Http\PhpEnvironment\RemoteAddressFactory',
-            'Laminas\Mvc\I18n\Translator' => 'VuFind\I18n\Translator\TranslatorFactory',
             'Laminas\Session\SessionManager' => 'VuFind\Session\ManagerFactory',
         ],
         'delegators' => [
+            'Laminas\I18n\Translator\TranslatorInterface' => [
+                'VuFind\I18n\Translator\TranslatorFactory',
+            ],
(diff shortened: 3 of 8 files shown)


demiankatz:
@EreMaijala, I'm interested to hear your thoughts on this one. So far, it seems to be working fine both for my manual tests and the full automated suite. I like the fact that this allows us to work with, rather than against, the framework in a few places... and the changes to the DisplayLanguageOption helper are a clear improvement.

module/VuFind/src/VuFind/View/Helper/Root/DisplayLanguageOptionFactory.php
```diff
@@ -65,10 +65,6 @@ public function __invoke(ContainerInterface $container, $requestedName,
         if (!empty($options)) {
             throw new \Exception('Unexpected options sent to factory.');
         }
-        // We want to construct a separate translator instance for this helper,
-        // since it configures different language/locale than the core shared
-        // instance!
-        $factory = new \VuFind\I18n\Translator\TranslatorFactory();
-        return new $requestedName($factory($container, Translator::class));
+        return new $requestedName($container->get(Translator::class));
```
EreMaijala: This apparently breaks the language selection.

EreMaijala (changes requested):
This looks pretty good to me apart from the issue I flagged with a comment.

module/VuFind/src/VuFind/View/Helper/Root/DisplayLanguageOptionFactory.php
demiankatz: Can you elaborate on how it is broken? It works for me in my test environment (displaying the drop-down list with the text from native.ini). Is it possible you are seeing a cache-related problem? (Of course, I don't rule out the possibility that it's actually broken... but I want to be sure I test the right thing, since I thought I had this covered).

EreMaijala: After clearing the cache I'm seeing all choices in English.

EreMaijala: With dev they're in the native language.

demiankatz: Oops, I should take my own advice: I was tricked by the cache! Thanks for catching this bug; I believe it is now fixed by c23ef01. Please let me know if it's sorted out on your end now, when time permits.

module/VuFind/src/VuFind/View/Helper/Root/DisplayLanguageOptionFactory.php
```diff
@@ -65,10 +65,9 @@ public function __invoke(ContainerInterface $container, $requestedName,
...
             throw new \Exception('Unexpected options sent to factory.');
         }
-        // We want to construct a separate translator instance for this helper,
-        // since it configures different language/locale than the core shared
-        // instance!
-        $factory = new \VuFind\I18n\Translator\TranslatorFactory();
-        return new $requestedName($factory($container, Translator::class));
+        $translator = $container->get(Translator::class);
+        // Add a special locale used just for this plugin:
+        $translator->addTranslationFile('ExtendedIni', null, 'default', 'native');
```
demiankatz, line 70: Perhaps it makes more sense to put this in the actual TranslatorFactory; I'm torn between doing that (which is probably more straightforward) and leaving it here (since it ties the loading of the translations to the use of the helper that needs them). Any thoughts/preferences, @EreMaijala?

module/VuFind/src/VuFind/View/Helper/Root/DisplayLanguageOptionFactory.php
EreMaijala: Works great now. :)

module/VuFind/src/VuFind/View/Helper/Root/DisplayLanguageOptionFactory.php
EreMaijala: @demiankatz I like it the way it is now. As you said it ties the logic with the helper that needs it.

module/VuFind/src/VuFind/View/Helper/Root/DisplayLanguageOptionFactory.php
demiankatz: Great, thanks again! Merging now!

PR #1929, opened 2021-04-23
Make install command return 0 on success.

EreMaijala:
Simple and obvious-looking change, but I haven't actually tested the command.

### DIFF
--- a/module/VuFindConsole/src/VuFindConsole/Command/Install/InstallCommand.php
+++ b/module/VuFindConsole/src/VuFindConsole/Command/Install/InstallCommand.php
@@ -928,5 +928,6 @@ class InstallCommand extends Command
 
         // Report success:
         $this->displaySuccessMessage($output);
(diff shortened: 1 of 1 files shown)


demiankatz:
I tested the code with and without the change, and the return value is the same in both cases -- looks like when a function returns a default null value, the framework normalizes it to 0 anyway. However, this was definitely an oversight on my part, and it makes sense to be explicit. Merging now -- thanks for the catch!

demiankatz:
(Also backported to release-7.1, since I think this qualifies as a bug fix even though it makes no apparent functional difference -- seems better safe than sorry in case the behavior differs from platform to platform).

demiankatz:
See commit https://github.com/vufind-org/vufind/commit/e6f0e52ed64c1ed2e652f62895e2d88ae64fd8e3 for the version that made it into the codebase.

PR #1930, opened 2021-04-26
Don't override login template in collection controller.

EreMaijala:
This fixes a crash that occurs when a collection record is loaded with the ?login=true parameter. Note that there's currently nothing doing so, as far as I can tell, but some crawlers seem to be happy to try everything.


demiankatz:
Thanks, @EreMaijala. This is likely related to the removed VuDL module, but it might come in handy in future, so I'm glad you were able to fix it. I've backported the fix to the release-7.1 branch and merged it to dev from there.

PR #1931, opened 2021-04-26
Improved accessibility for results / favorites lists.

al-pu:
To improve accessibility it is important to use ul / ol lists if the content itself represents lists.
Especially the search result lists and record listings at the account pages has only been divs.

to be discussed (maybe):
* naming and use of the new css class
* templates/RecordDriver/DefaultRecord/list-entry.phtml: display of cover moves into the result-li

TODO
- [x] Fix broken cart tests
- [x] Fix margins on search results and favorites
- [x] Fix background color striping of favorite lists


demiankatz:
Thank you for this proposal, @al-pu!

I'm requesting a review from @crhallberg before I take a closer look at this myself. Once everyone is on the same page about approach, please let me know and I can go ahead with my own review and testing!

crhallberg (approved):
This checks all the boxes I was looking for. Thanks very much!

crhallberg:
While we have you, I would be curious if you have any opinions on how to make the record listings inside the `li`s more accessible. I used to think putting the metadata in tables was the way to go but lately, I think using a definition list would be best. Any thoughts?

al-pu:
I hope we can share some more accessibility improvements in the future. There are some issues we are working on.

al-pu:
Just had to compare some catalogues. I like the idea of using definition lists. I think it`s more self-explanatory than displaying key / value by using tables.

demiankatz:
@al-pu, I had some time to look at this.  A few notes/comments:

1.) I notice that there are some margins or padding on the search results which we probably ought to remove. Here is the current dev code (in the bootprint3 theme):

image

...and here is yours:

image

(Similar problems exist for both search results and favorite lists).

2.) The Mink integration tests for the cart are consistently failing on this branch (and I confirmed that they still pass on the dev branch). For some reason, only one item is getting added to the cart where they expect to add two. The cart does seem to be working right when I manually test it, so I'm not exactly sure what is going wrong. I can investigate this further as time permits; if you have any theory in the meantime, please let me know!

3.) Something seems to be wrong with the logic for alternating the background color of favorite list rows. In dev, every row is gray. In your branch, every row is white. I think both of these are wrong. Screen shots:

Dev:

image

This PR:

image

(I wonder if @crhallberg should try to fix this in dev before we address it here... but maybe that would just be a duplication of effort; let me know if you have any thoughts on that!)

I will add some TODO boxes to the top comment so we don't lose track of these issues.

demiankatz:
To follow up a little further on the failing tests, this is the code where things are going wrong:

<pre>
        foreach ($page->findAll('css', '.cart-add') as $button) {
            $button->click();
        }
</pre>

The search sets up a result list with two results, and then uses a loop over the `.cart-add` elements to click all the "add to cart" buttons. If I add some time delays and watch this happening in a real browser, I can see the first cart button highlight, but not activate... and then the second one gets clicked and works normally. I'm not sure why that first click is not registering correctly, or how the changes here are breaking this test (since I don't see an obvious direct relationship between the changes and the cart functionality). Also strange is that if I manually click on the first cart button while the test is running, I can make the test pass.

Any ideas? Is it possible there's an HTML error somewhere that we're not noticing that is leading to this strange behavior? I wonder if running the page through a validator might be worthwhile....

al-pu:
1) Strange. On my local machine it's working fine. there are no margin / paddings.
2) I will have a look at this within the next days (including your last comment). Not looked into it yet.
3) I found the problem. Between the `<li>`-blocks are JS-blocks (comes from themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml). If I remove them within firefox dev tools the html stucture is proper again and and the background is switching between the items. JS has to occur outside the list or inside the li. I think it's the best to solve it within this issue. I can try to solve this.

demiankatz:
Thanks, @al-pu! Some further comments:

1. I tested with the bootprint3 and sandal themes and saw the issue in both. I didn't try the base bootstrap3 theme. Could that be a factor?
2. I'm happy to help with more test support as needed. I definitely appreciate any help you can offer, but please don't waste too much of your time if an obvious solution does not present itself. The problem may be related to some strange quirk of the test framework or the way things are set up; if my familiarity with the stack and setup saves you time, please don't hesitate to ask for backup. :-)
3. Great, I'll look forward to seeing your solution to this part!

al-pu:
2.) Build failed here: Figured out by @RLangeUni
```
  <!-- Run LessToSass, error if there are uncommitted changes (used by Travis) -->
  <target name="checkLessToSass">
    <exec command="grunt lessToSass" checkreturn="true" passthru="true" />
    <exec command="git diff --exit-code *.scss" checkreturn="true" passthru="true" />
  </target> 
```
I commited only the scss before. Because it satisfy your demand. Now the less styles are available too. The build shoudln't fail here any more (I hope so).

1.) I suppose, new less styles was missing and lists displayed incorrectly therefore.

3.) I added the small improvement.

demiankatz:
Thanks, @al-pu! All tests are passing now, though I had to run the `grunt less` task to recompile the CSS to get everything working on this end. I think this is ready to merge, though I'm hoping @crhallberg can give it one last look first, to be sure we haven't overlooked any details. (He's out of the office today but will be back next week).

demiankatz:
This seems to have gotten lost in the shuffle somewhere; @crhallberg, can you give this a re-review and make sure it's ready to go?

A couple of comments from my end:

1.) Due to subsequent changes to the dev branch, a conflict was introduced. I have resolved it, and I think everything is okay on that front.

2.) For stylistic consistency, it would probably be a good idea to indent all of the content inside the added list tags. Any objections?

crhallberg:
The CSS styles all look good and consistent to me! I agree that we should indent the code for readability and consistency.

demiankatz:
This is looking good to me now -- I've adjusted the indentation and done another round of testing. I found a problem in the bootprint3 theme (see [VUFIND-1492](https://vufind.org/jira/browse/VUFIND-1492)), but it is a pre-existing bug not related to the changes here. I'm merging this now -- thanks again to everyone who helped get this done!

PR #1933, opened 2021-04-27
enable overriding scss variables from bootstrap accessibility plugin

al-pu:
Setting the bootstrap accessibility plugin scss variables to '!default' enables global overriding of them from other files like compiled.scss

Why as PR?
* bootstrap accessibility itself does not seemed to be maintained any more
* we coudn't override the default plugin styles without changing something within the bootstrap3/scss/vendor folder. because a clean VuFind is our base layer we coudn't edit there.

Further question:
* Can we backport the change to 5.1? And if yes, shall I create an additional PR?

### DIFF
--- a/themes/bootstrap3/scss/vendor/bootstrap-accessibility/partials/_variables.scss
+++ b/themes/bootstrap3/scss/vendor/bootstrap-accessibility/partials/_variables.scss
@@ -1,21 +1,21 @@
 //Set the variables for use in stylesheet
 
 // Colors
-$black: 		#000;
-$green-dark: 	#2d4821;
-$blue-dark: 	#214c62;
-$red-dark: 		#6c4a00;
-$red: 			#d2322d;
-$yellow: 		#f9f1c6;
+$black: 		#000 !default;
+$green-dark: 	#2d4821 !default;
+$blue-dark: 	#214c62 !default;
+$red-dark: 		#6c4a00 !default;
+$red: 			#d2322d !default;
+$yellow: 		#f9f1c6 !default;
 
 // Outline
-$outline-default-style: 	dotted;
-$outline-default-color: 	$black;
+$outline-default-style: 	dotted !default;
+$outline-default-color: 	$black !default;
 
 // Alerts Colors
-$alert-success-color: 			$green-dark;
-$alert-info-color: 				$blue-dark;
-$alert-warning-color: 			$red-dark;
-$alert-warning-bkg: 			$yellow;
-$alert-danger-color: 			$red;
-$alert-danger-hover: 			darken($red, 10%);
+$alert-success-color: 			$green-dark !default;
+$alert-info-color: 				$blue-dark !default;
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks, @al-pu. A few thoughts/comments:

1.) As a matter of principle, we try to avoid changing vendor code unless absolutely necessary. It seems like this may be a case where it is actually absolutely necessary... but I'd be interested to know whether @crhallberg has any alternative suggestions.

2.) Since VuFind currently uses LESS for its primary CSS compilation, with SCSS as an alternative, we should make sure that any changes are synchronized between both versions of the stylesheets. (Right now this fix seems to be SCSS-only, which will then fail to get compiled in via the normal `grunt less` process we use to build compiled.css).

3.) If backporting this to release-5.1 will simplify your local maintenance, I'm willing to do it. No need to open a separate PR; let's just figure out our solution here, and then I can backport it as part of the merge process.

crhallberg:
I think the removal of `!default` was introduced by `grunt lessToSass`. There is an [official SASS port of Bootstrap 3.4](https://github.com/twbs/bootstrap-sass/tree/master/assets/stylesheets) so maybe a better way to go about this in the future is to include a grunt step that copies it out of another directory or zip file to prevent changes like this.

*Note: future version of Bootstrap (after 3) use SASS.*

demiankatz:
@crhallberg, I don't think this has anything to do with lessToSass, unfortunately (since if it did, that would be easier to fix). Here's the original source SCSS -- as you can see, it is also missing the `!default`: https://github.com/paypal/bootstrap-accessibility-plugin/blob/main/src/sass/partials/_variables.scss

I don't think lessToSass touches the vendor directory contents... and if it did, this PR would be causing Travis to fail, since there would be a mismatch in the lessToSass output.

al-pu:
1.+3.) I am aware that it is bad practice to change the vendor code. Could not figure out a better way. I am thankful that you consider to make a exception in this case (except there are technical issues).
2.) Yes, this solution is SCSS-only. I had a short investigation for the less equivalent but forgot to mention it: https://lesscss.org/features/#variables-feature-default-variables. It seems that there are no changes needed  on the less side (The value overriding should work out of the box - if I get it right).

demiankatz:
I wonder if the original plugin was written in LESS first and then inaccurately ported, which could be what @crhallberg was saying in the first place and I misunderstood.

@crhallberg, do you have any objections to making this change to the SCSS at this point, assuming you can't think of a solution that avoids changes to the vendor directory?

crhallberg:
I agree editing the accessibility vendor is the best way to go about this.

Sorry for the delay, I had to do some research to make sure this was the best way and it wouldn't be overridden accidentally.

demiankatz:
Thanks, @crhallberg! I've gone ahead and backported this to release-5.1 (and merged from there to release-6.1, release-7.1, and dev). Hopefully this entire thing will be superseded in the not-too-distant future.

PR #1934, opened 2021-04-28
FOLIO: Add callnumber_prefix to the output of getHolding.

bbusenius:
We use call number prefixes for some of the stuff we do behind the scenes. We also display them in some cases. I'm hoping this will be an acceptable tweak to the driver though I don't see this attribute in the ILS driver API documentation. Let me know what you think.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -499,6 +499,24 @@ class Folio extends AbstractAPI implements
     }
 
     /**
+     * Choose a call number and callnumber prefix.
+     *
+     * @param string $hCallNumP Holding-level call number prefix
+     * @param string $hCallNum  Holding-level call number
+     * @param string $iCallNumP Item-level call number prefix
+     * @param string $iCallNum  Item-level call number
+     *
+     * @return array with call number and call number prefix.
+     */
+    protected function chooseCallNumber($hCallNumP, $hCallNum, $iCallNumP, $iCallNum)
+    {
+        if (empty($iCallNum)) {
+            return ['prefix' => $hCallNumP, 'callnumber' => $hCallNum];
+        }
+        return ['prefix' => $iCallNumP, 'callnumber' => $iCallNum];
+    }
+
+    /**
      * This method queries the ILS for holding information.
      *
      * @param string $bibId   Bib-level id
@@ -565,6 +583,12 @@ class Folio extends AbstractAPI implements
                 $locationCode = $locationData['code'];
                 $itemCallNumber = $item->itemLevelCallNumber ?? '';
                 $itemCallNumberPrefix = $item->itemLevelCallNumberPrefix ?? '';
+                $callNumberData = $this->chooseCallNumber(
+                    $holdingCallNumberPrefix,
+                    $holdingCallNumber,
+                    $itemCallNumberPrefix,
+                    $itemCallNumber
+                );
                 $items[] = [
                     'id' => $bibId,
                     'item_id' => $item->id,
@@ -579,10 +603,8 @@ class Folio extends AbstractAPI implements
                     'issues' => $holdingsStatements,
                     'supplements' => $holdingsSupplements,
                     'indexes' => $holdingsIndexes,
-                    'callnumber' => $itemCallNumber
-                        ? $itemCallNumber : $holdingCallNumber,
-                    'callnumber_prefix' => $itemCallNumberPrefix
-                        ? $itemCallNumberPrefix : $holdingCallNumberPrefix,
+                    'callnumber' => $callNumberData['callnumber'],
+                    'callnumber_prefix' => $callNumberData['prefix'],
                     'location' => $locationName,
                     'location_code' => $locationCode,
                     'reserve' => 'TODO',
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -511,9 +511,9 @@ class Folio extends AbstractAPI implements
     protected function chooseCallNumber($hCallNumP, $hCallNum, $iCallNumP, $iCallNum)
     {
         if (empty($iCallNum)) {
-            return ['prefix' => $hCallNumP, 'callnumber' => $hCallNum];
+            return ['callnumber_prefix' => $hCallNumP, 'callnumber' => $hCallNum];
         }
-        return ['prefix' => $iCallNumP, 'callnumber' => $iCallNum];
+        return ['callnumber_prefix' => $iCallNumP, 'callnumber' => $iCallNum];
     }
 
     /**
@@ -581,15 +581,13 @@ class Folio extends AbstractAPI implements
                 $locationData = $this->getLocationData($locationId);
                 $locationName = $locationData['name'];
                 $locationCode = $locationData['code'];
-                $itemCallNumber = $item->itemLevelCallNumber ?? '';
-                $itemCallNumberPrefix = $item->itemLevelCallNumberPrefix ?? '';
                 $callNumberData = $this->chooseCallNumber(
                     $holdingCallNumberPrefix,
                     $holdingCallNumber,
-                    $itemCallNumberPrefix,
-                    $itemCallNumber
+                    $item->itemLevelCallNumberPrefix ?? '',
+                    $item->itemLevelCallNumber ?? ''
                 );
-                $items[] = [
+                $items[] = $callNumberData + [
                     'id' => $bibId,
                     'item_id' => $item->id,
                     'holding_id' => $holding->id,
@@ -603,8 +601,6 @@ class Folio extends AbstractAPI implements
                     'issues' => $holdingsStatements,
                     'supplements' => $holdingsSupplements,
                     'indexes' => $holdingsIndexes,
-                    'callnumber' => $callNumberData['callnumber'],
-                    'callnumber_prefix' => $callNumberData['prefix'],
                     'location' => $locationName,
                     'location_code' => $locationCode,
                     'reserve' => 'TODO',
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -551,6 +551,8 @@ class Folio extends AbstractAPI implements
                 $textFormatter,
                 $holding->holdingsStatementsForIndexes ?? []
             );
+            $holdingCallNumber = $holding->callNumber ?? '';
+            $holdingCallNumberPrefix = $holding->callNumberPrefix ?? '';
             foreach ($this->getPagedResults(
                 'items', '/item-storage/items', $query
             ) as $item) {
@@ -561,8 +563,6 @@ class Folio extends AbstractAPI implements
                 $locationData = $this->getLocationData($locationId);
                 $locationName = $locationData['name'];
                 $locationCode = $locationData['code'];
-                $holdingCallNumber = $holding->callNumber ?? '';
-                $holdingCallNumberPrefix = $holding->callNumberPrefix ?? '';
                 $itemCallNumber = $item->itemLevelCallNumber ?? '';
                 $itemCallNumberPrefix = $item->itemLevelCallNumberPrefix ?? '';
                 $items[] = [
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -561,6 +561,10 @@ class Folio extends AbstractAPI implements
                 $locationData = $this->getLocationData($locationId);
                 $locationName = $locationData['name'];
                 $locationCode = $locationData['code'];
+                $holdingCallNumber = $holding->callNumber ?? '';
+                $holdingCallNumberPrefix = $holding->callNumberPrefix ?? '';
+                $itemCallNumber = $item->itemLevelCallNumber ?? '';
+                $itemCallNumberPrefix = $item->itemLevelCallNumberPrefix ?? '';
                 $items[] = [
                     'id' => $bibId,
                     'item_id' => $item->id,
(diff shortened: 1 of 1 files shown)


demiankatz:
@bbusenius, I guess the main question is how to define and use this.

First of all, I double checked, and it does not appear that there is a precedent for this... so I think the naming you have chosen makes sense. We just need to clearly define what the difference is between callnumber and callnumber_prefix so that we can update the driver spec.

In order to display this in existing VuFind code, it might be easier to concatenate the prefix and call number together inside the driver... but I'm guessing that would not be workable because you only need to display the prefix in certain circumstances (and it would likely introduce inconsistencies with things like AlphaBrowse of call numbers). Does it make sense to introduce any core code or configuration around this functionality, or is it easier to just leave it for local customization for now?

bbusenius:
@demiankatz great questions. So, I may have misspoken because in all likelihood, I think we always display the call number with the prefix and then use the one without it for generating links to the AlphaBrowse and sorting, etc. I only implied otherwise because I don't trust my memory well enough to say we always display it with 100% certainty (it's likely though). In OLE, where we had a totally forked driver, we would actually return a `callnumber_display`, `callnumber_sort` and `callnumber_prefix`. In some cases we actually need the prefix separately. For example, when we generate a link to a map of where to get the item, we need the prefix separately.

Just as a point of reference, some of the more common prefixes that we have are `f` (for folio), `CD` and `DVD`.

Anyhow, I'm trying to simplify things here so I'm questioning the need for a display call number and a sort call number in the driver. I'm thinking I can just tack the prefix on wherever we want to show it and let everything else use the stock "prefixless" functionality.

I'll also note, I have this marked as a customization that we need to maintain locally. This is based off that conversation we had with you eons ago when we went over our customizations in a large spreadsheet. I'm not saying I'd be opposed to turning this into a bigger contribution, rather, just trying to explain the minimal approach. As long as I can get this data, I should be able to do what I need.

demiankatz:
Thanks for the clarification, @bbusenius; if the simplest thing is to just add this as another optional data element in the spec (and note that it is not currently used by core code), I'm happy to do that.

I can't actually think of a reason why you would need the callnumber_sort in the front end, though if that needs to be added for a compelling reason, I'd certainly be open to doing that as well. If it's actually needed somewhere, I would think it might be possible to generate the sort version algorithmically on demand too if that were actually needed, which would avoid the need to implement it in all the different drivers (many of which likely do not support that option natively).

If it spares you from maintaining extra local code, I'm also open to adding some default usage of the call number prefix in displays. I suspect there may be a fairly uncontroversial approach possible. It's just a question of making sure we're consistent in how the Solr index, browse index, and ILS data correlate with one another.

bbusenius:
Thanks @demiankatz. Just adding it to the spec would get us over the hump but I'm happy to do more with it if it will make it more useful and keep us from maintaining local code. I agree we don't need a `callnumber_sort`. I think we were just doing it because we could. The name is misleading. As far as I know we were only using it as a proxy for `callnumber` to generate links.

Currently, the way I'm dealing with this is to "override" the `getUniqueCallNumbers` method in `HoldingsILS`. Rather than directly override the method, however, I give it a unique name and use it selectively where I need it. I'm just doing this out of an abundance of caution to be sure I don't unwittingly break some callnumber related piece of functionality I might not be thinking about. The big change here is that instead of returning an array of callnumbers, I return an array of keyed arrays with `prefix` and `callnumber` keys in order to keep the data matched.

If you think this is a fair approach, I would be inclined to just edit the `getUniqueCallNumbers` method directly and update the templates where it's used. If memory serves me, that would be in the holdingsils template, in one of the templates for the results view and somewhere else for My Account. Is there a better way?

demiankatz:
Thanks, @bbusenius, that sounds like a reasonable approach, though your proposal only covers the holdings tab of the record view; getUniqueCallNumbers is a helper method of the tab class and is not used anywhere else. That means that changing it is a pretty low-risk operation, but it also means that you need to do additional work if you want to affect the call numbers in the search results view. That all gets processed in the [GetItemStatuses handler](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php), which is quite a complex bit of code because there are three different code paths for formatting the statuses in different ways depending on config.ini settings. If you think it's worth the effort, I'm certainly happy to work with you to try to get all the details right here... but if you don't have the time, we can just start by adding the field and revisit all the details later as time permits.

bbusenius:
I think this might be worth the effort. We currently do show call number prefixes in our result view. I could probably convince stakeholders to do without it for awhile but I'd need to circle back on it pretty quickly. Also, I'm probably going to have to do the same work if I try to override in a local module anyhow.

demiankatz:
Sounds good, @bbusenius. I'll leave this open for now so you can work on it as time permits. If you'd like me to do anything else in the meantime, just let me know!

bbusenius:
Sounds good @demiankatz! I'm making progress on this and will likely ask for your input on the holdings tab piece before I move onto the results. I do have another question first though.

In working on adding the prefix to the holdings tab as outlined above, I was reminded that we also display an item level call number in cases where it is different from the holdings level call number. This is displayed at the item level like so:

image

Folio has a field expressly for this purpose called "effective call number string". Glancing at the API it looks like Folio returns `itemLevelCallNumberPrefix` and `itemLevelCallNumber`. So my two-part question is 1. Would this be something that could be contributed back? and 2. If yes, would it be appropriate to do it as part of this PR or better to do it as another PR? I'm imagining that if this would only be used in the holdings tab and would not be added to the results view if we did it.

demiankatz:
@bbusenius, the getHolding() method is returning an array of item-level information, so if FOLIO returns item-level values, those should be getting used instead of the holding-level values for the purposes of this function (or maybe the holding value is needed as a fallback if the item-level values are undefined, depending on whether or not the API already does the fallback work for you).

As far as how to submit the PR's, maybe it would make sense to open two: this one, which can contain just FOLIO-specific changes (for review/discussion by the FOLIO community), and a second one (yet to be opened), which can contain the core VuFind changes to make use of the callnumber_prefix field. That second PR can include new code in the VuFind\ILS\Driver\Demo class to generate random prefixes for testing purposes, and then it will completely stand alone. That way we can keep the VuFind and FOLIO conversations separate to make things less confusing... and we can merge the PR's in either order, depending on which gets done first, as long as we agree on the name of the prefix field (which I think we already do).

bbusenius:
Thanks @demiankatz. That makes sense and that's what I'll do. It looks like the Folio driver has a bug then, since it's not using the item call number when available. I don't think the API does that automatically and that's OK, it's probably even preferable for us.

This does present a problem though. I need to disentangle what I can contribute back and what we need to maintain locally. From our perspective it's better to display the holdings call number at the holdings level and the item call number separately with the item. The reason is, we have lots of [records like this one](https://catalog.lib.uchicago.edu/vufind/Record/2361348) that were confusing for our users when the item call number displayed at the top with the holdings level call number (you'll have to click "View More Items" and scroll down to `c.1 : v.495` to see the item level call number).

Regarding the above example:

1. We're adding the copy number to the end of the call numbers. The copy number is not actually part of the call number. This helped but it was not good enough for our users when they all displayed at the top. The copy number as I recall, is also not always available.
2. In this case it's only a prefix that makes them different. This is not always the case. Sometimes the numbers are very different.

The problem we had was that on some records, users had no way of knowing which call number belonged to which item, so they would go off looking for the item with the wrong call number. In cases where only the prefix was different, they wouldn't know the item was large format or audio format and they'd be looking for normal print.

Would it make sense for me to add the prefix field and fix the item call number bug in the Folio driver and then override the Folio driver locally such that I revert the bug fix for the holdings call number and add a new field for item call number locally?

I apologize for getting so far into the weeds on this one. I had no idea all this was going to come up.

demiankatz:
@bbusenius, no worries, this is how it goes sometimes -- you start with one line of code and end up with a short novel. ;-)

I confess that I can no longer remember the exact motivation behind it, but I wonder if the holdingsTemplate setting might come into play here -- this allows a choice between the "standard" default holdings view, and a different "extended" version. Like I say, I can't remember exactly why this was done in the first place and don't currently have enough time to review the differences, but I wonder if call numbers are handled differently in these two places (or if they should be).

It definitely seems like, one way or another, we need a better way of differentiating between holding-level and item-level call numbers when differences exist... but I've been away from these templates long enough that I don't remember how they work without spending some time digging in. It's also possible we need to revisit the data structures behind the templates, since right now we basically build the holdings by examining the items, without a real mechanism for providing holding-level data by itself.

I'm certainly open to brainstorming further, and I'm also willing to refresh my memory on some of these details if you need me to... I'm just having an unusually busy day at the moment. :-) Also happy to put this on a future Community Call agenda if you think a real-time conversation would be valuable.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -561,6 +561,10 @@ public function getHolding($bibId, array $patron = null, array $options = [])
                 $locationData = $this->getLocationData($locationId);
                 $locationName = $locationData['name'];
                 $locationCode = $locationData['code'];
+                $holdingCallNumber = $holding->callNumber ?? '';
```
demiankatz: I think you could actually move the two holding-related variable declarations above the foreach loop that's iterating over items in order to avoid a little bit of redundant work.

bbusenius: :facepalm: Yes, of course. This is done.

demiankatz:
@bbusenius, thanks for the progress on this. I also took a little more time to refresh my memory on how call number displays work. It looks to me like right now, each holding location will include a list of all the different unique call numbers found at that location. This could be potentially confusing if there are a lot of them, but it should at least express that multiple call numbers exist rather than just picking one and sending people to the wrong place. It's possible this behavior has changed since you first started working with VuFind, though I can't remember the history without digging into Git logs.

If it would be more useful to display the call number from the holding record, I think it would be fairly easy to incorporate this by adding a new optional top-level key to the output of getHolding() which could include information about particular holding locations. I'm not really sure whether or not this would be useful (and I'm sure we would need to make it configurable, because I suspect even if it's useful for some institutions, it might be detrimental to others)... but I think it's technically feasible without a lot of effort. Right now, FOLIO is returning the old-style, simple format from getHolding, which is just an array of items, and then \VuFind\ILS\Connection is upgrading this to the new, more advanced format (which allows for more top-level keys). It wouldn't be hard to change the FOLIO driver to use the keyed format and then add more data at the top level, as needed.

Let me know if you think this is worth discussing further and I'm happy to do so! Otherwise, I'll wait for you to open the other call number PR we discussed before taking further action here.

demiankatz:
@bbusenius, I also edited the title of this PR to add "FOLIO:" since the non-FOLIO-specific bits are going to end up elsewhere.

bbusenius:
Thanks @demiankatz! The way you describe call numbers, is the way I always recall it being. I'm not sure if we'll want to pursue the optional top-level key to the output of getHolding() or not. I'm inclined to think we'll be doing something custom for this, no matter what. This is one where I'm pretty sure they're going to want to keep it as is in our current catalog. I would be open to discussing this more in the future but for right now, if we're able to get prefixes displaying along with the other changes here, I think it's a big win. I opened PR #1951 for the other. So far it only deals with the display of prefixes in the record tab. I think it would be good to iron that out before I try to do anything with results.

demiankatz:
Sounds good to me, @bbusenius!

demiankatz:
@bbusenius, now that we're wrapping up #1951, I thought I'd give this another look. One question before we merge it -- I'm wondering if the logic for inheriting from the holding call number is entirely correct. Right now, we're looking at prefix and call number independently... so if either one of those values is empty, we take the value from the holdings. It seems to me that this could cause a problem, if the holdings call number has a prefix, and the item call number does not -- in that situation, there's no way to prevent the holdings prefix from carrying forward. I wonder if instead it would be better to inherit both values from the holdings if the item call number is empty (or maybe only inherit the holdings prefix if BOTH the item prefix AND the item call number are empty). Does that make any sense? Or does some detail of how FOLIO works make these concerns irrelevant?

bbusenius:
Sorry I missed your last comment @demiankatz. This seems really obvious when I look at it now. I think in practice, it doesn't happen with our data which is probably why it was easy for me to miss. I'm not sure if there's anything about how FOLIO works that would make these concerns irrelevant, however, the way our institution uses these would make it irrelevant to us (I don't think this bug would ever effect us with our data). It seems like something we should fix though. I should be able to do this shortly.

demiankatz:
Thanks, @bbusenius!

bbusenius:
@demiankatz I took a pass at this. There are lots of ways to do this. I hope you don't think the way I did it is overkill. The more I thought about it, I felt like the way I did it would be a little more clear and easy to override if needed. Let me know what you think.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -561,6 +581,14 @@ public function getHolding($bibId, array $patron = null, array $options = [])
                 $locationData = $this->getLocationData($locationId);
                 $locationName = $locationData['name'];
                 $locationCode = $locationData['code'];
+                $itemCallNumber = $item->itemLevelCallNumber ?? '';
+                $itemCallNumberPrefix = $item->itemLevelCallNumberPrefix ?? '';
```
demiankatz: I don't think the `$itemCallNumber` and `$itemCallNumberPrefix` variables are especially useful -- you could just put the direct references like `$item->itemLevelCallNumber ?? ''` into the `chooseCallNumber` call below to save a couple of lines of code. It's possibly worth keeping the `$holdingCallNumber` and `$holdingCallNumberPrefix` variables since that avoids repeated null checking inside the loop, but that's a pretty trivial optimization, so I could really go either way on those.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -575,7 +603,8 @@ public function getHolding($bibId, array $patron = null, array $options = [])
                     'issues' => $holdingsStatements,
                     'supplements' => $holdingsSupplements,
                     'indexes' => $holdingsIndexes,
-                    'callnumber' => $holding->callNumber ?? '',
+                    'callnumber' => $callNumberData['callnumber'],
```
demiankatz: If you make `chooseCallNumber` return a `callnumber_prefix` key instead of just `prefix`, then you could eliminate these two lines by using `$items[] = $callNumberData + [ ...` above.

demiankatz (commented):
Thanks, @bbusenius, this makes a lot of sense to me -- as you say, this approach makes it really easy to customize the call number behavior in a subclass if there is a need. I've just left a few suggestions that might reduce the length of the code, since the getHolding method is already quite long. If you don't think these are worth the effort, though, I would be happy to merge as-is. Please let me know your preferences!

bbusenius:
Thanks for the feedback @demiankatz! I made those changes.

demiankatz:
Looks good, @bbusenius, thanks for the final tweaks; merging now!

PR #1937, opened 2021-05-03
Enable split-on-whitespace in Solr.

EreMaijala:
From Solr 6.5 the splitting on whitespace was disabled by default in Solr. This enabled multi-word synonyms to work properly. However, it also results in the queries being handled differently based on whether the field analysis for query fields is the same for all fields or not. From user perspective this can look really inconsistent and cause seemingly correct searches to fail to match anything. See e.g. the following links for more information:

 https://opensourceconnections.com/blog/2018/02/20/edismax-and-multiterm-synonyms-oddities/
 https://lucene.472066.n3.nabble.com/Problem-with-eDisMax-and-multi-word-synonyms-td4471917.html

Note that this can be revisited if/when https://issues.apache.org/jira/browse/SOLR-12779 is resolved. 

TODO: 
- [x] Decide if this is worth it given the downside.
- [x] Consider making this configurable in VuFind so that no schema change is required.


demiankatz:
Thanks for doing the research on this one; it gets into pretty deep waters!

I think it probably does make sense to allow this to be configured from within VuFind, so it's easier to adjust without schema hacking. I think multi-word synonyms are a rarely used feature, but it would be good to have a strategy prepared to support them if/when needed.

I suppose this might also be an indication that we should re-evaluate our schema and see if it's feasible to reduce inconsistencies in field types. Might we be able to solve this by creating `allfields_xxx` variations of key fields that get copyFielded values from differently-formatted fields in order to enforce uniformity? That seems like a brute force approach, but it might help in the multi-word synonym scenario.

EreMaijala:
@demiankatz What do you think about the latest commit?

I'm not sure I understand how adding a new kind of `allfields` field would help unless you'd target only `allfields` or `allfields_xxx` and never both at the same time.

config/vufind/searches.ini
```diff
@@ -112,6 +112,23 @@ retain_filters_by_default = true
...
+;
+; Parameters can be defined for the following search contexts:
+; *                 All contexts
+; search            Search
+; retrieve          Retrieve records
+; similar           Find similar records
+; terms             Return terms
+; alphabeticBrowse  Return information from the browse index
+; workExpressions   Return work expressions (record versions)
+;default_parameters[*] = "echoParams=none&debugQuery=false"
```
demiankatz, line 133: I like the idea of passing the parameters as a query string, but maybe it's worth commenting about the format in the notes above. Have you done any testing to see how multi-valued parameters need to work? Does `parse_str` work if you repeat the name, or do you need to add brackets? (I'm suspecting the latter, which is inconsistent with how Solr works and might cause confusion if we don't comment on it... though maybe there's never a need for repeating default parameters).

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
```diff
@@ -542,4 +552,17 @@ protected function getInjectConditionalFilterListener(Config $search)
...
+
+    /**
+     * Get a default parameters listener for the backend
+     *
+     * @param Backend $backend Search backend
+     * @param array   $params  Default parameters
+     *
+     * @return DeduplicationListener
+     */
+    protected function getDefaultParametersListener(Backend $backend, array $params)
```
demiankatz, line 570: This reminds me that I'd really like to come up with a more elegant way of managing listeners, since this factory just keeps getting bigger and bigger... but that's a problem for another day. :-)

demiankatz (commented):
I like this approach, and it seems reasonable to me (though I haven't had time to actually test it yet -- happy to do so in the near future if you'd like me to). Once we're both happy with the approach, we should also add some test coverage.

To clarify what I was saying about `allfields_xxx` fields, I was thinking about how the most common source of mismatched analysis problems is the AllFields search defined in searchspecs.yaml, which does a DisMax query over a large list of fields. Most of them are analyzed the same, but a few are outliers. I'm suggesting that we could solve the inconsistency taking all of the outliers, copyFielding them to `allfields_xxx` fields of type `text`, and then using the `allfields_xxx` copies instead of the raw `xxx` fields in the AllFields definition in searchspecs.yaml. Of course, we would have to do this for all of the search definitions that have inconsistencies, and I'm sure AllFields is not the only one. This is a brute force approach and would grow the index, so I don't claim it's a particularly _good_ solution... but I think it might be a relatively _simple_ one in some situations where there are only a small number of outliers that need special DisMax boosts. I wouldn't propose it as the default approach, but maybe a workaround we could suggest in situations where multi-word synonym support is critical.

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
EreMaijala: Indeed. A plugin manager might be nice...

config/vufind/searches.ini
EreMaijala: Right, multivalued params would have required the bracket syntax. I decided to just parse manually to avoid needing it. I can't think of a real case where parameters would need to be repeated, but it's still a good idea to support that.

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
demiankatz: I think the obstacle to a plugin manager is that many of the listeners are constructed with backend-specific constructor parameters that are specific to the factory. We need a distinct instance of each listener for each backend. That would either require a separate plugin manager per backend, a non-standard naming convention for retrieving listeners from the plugin manager, or a refactoring of the listeners to have standardized constructor parameters and accept configuration after construction. None of these solutions seem ideal, though some might be better than what we currently have!

EreMaijala: You're of course right on all accounts.

config/vufind/searches.ini
demiankatz: Makes sense!

demiankatz:
Thanks, @EreMaijala, I think this looks good from my perspective. I'd suggest two things before merging:

1.) It might be worth sending an email to vufind-tech to make sure others have a chance to consider this, and offer feedback about multi-word synonyms.

2.) As noted above, we should add some test coverage for the new listener class.

I'm happy to help with either or both of these, though I may not have time until next week. Please let me know what you'd like me to do, and I'll put it on the list.

EreMaijala:
@demiankatz I'll see what I can do with the test coverage. If you could formulate and send the email, that'd help a lot!

EreMaijala:
@demiankatz Tests added.

demiankatz:
Thanks, @EreMaijala! I've sent the email, so I'll give this a couple of weeks to collect feedback and then will merge if no one objects or suggests changes.

EreMaijala:
@demiankatz Where are we now with this?

demiankatz:
Thanks for reminding me about this one, @EreMaijala -- no one responded to my email request for input, so I lost track of it. I think at this point, I'm comfortable interpreting that as a lack of objections, especially since it's easy to configure to restore the old behavior should the need arise.

PR #1940, opened 2021-05-04
Update dev tools plus related dependencies/configs.

demiankatz:
This PR upgrades some of VuFind's development related dependencies to newer versions. Some notes:

- This update was motivated by deprecation warnings related to ppito/laminas-whoops depending on laminas/laminas-console; this issue is fixed in the new version.
- It is not yet possible to upgrade php-cs-fixer to the newly-released version 3.0, due to a dependency conflict with PHPMD. This should be fixed by the next release of PHPMD. In the meantime, I have corrected some php-cs-fixer deprecations so it will continue to work once we are able to perform the update to 3.0.
- I had to bump the version of symfony-console slightly for compatibility with one of the updates; there is a new major version available, but I opted to make a smaller jump in this PR. I will address all other dependency updates separately after this is merged.
- All other updates are minor, and I have tested that the tools still behave as expected (all tests passing, etc.).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,11 +62,11 @@
         "pear/http_request2": "2.4.2",
         "pear/validate_ispn": "dev-master",
         "phing/phing": "2.16.4",
-        "ppito/laminas-whoops": "2.0.0",
+        "ppito/laminas-whoops": "2.1.0",
         "scssphp/scssphp": "1.4.1",
         "serialssolutions/summon": "1.3.0",
         "slm/locale": "0.5.0",
-        "symfony/console": "4.4.18",
+        "symfony/console": "4.4.22",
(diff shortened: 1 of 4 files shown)


demiankatz:
@EreMaijala, with our new policy of "no unreviewed commits," I'm asking for a review of this (hopefully uncontroversial) change. I do think it's a helpful process, since the documentation I've added in this PR is probably more useful than what I would have put in just a commit message, and should be considerably more informative if somebody is doing code archaeology in the future. If you don't have time to review, feel free to say so and perhaps we can get somebody else to give it a look.

EreMaijala (approved):
Looks good to me and everything still works.

PR #1941, opened 2021-05-04
Add getBuildings() method to retrieve building field.

demiankatz:
For some reason, we never added a record driver method to retrieve the building field from Solr; however, some code existed which assumed the existence of such a method. This PR creates the method and normalizes references to it into a plural form.

### DIFF
--- a/config/vufind/SearchApiRecordFields.yaml
+++ b/config/vufind/SearchApiRecordFields.yaml
@@ -44,6 +44,12 @@ bibliographyNotes:
   type: array
   items:
     type: string
+buildings:
+  vufind.method: getBuildings
+  description: Buildings where the record is held
+  type: array
+  items:
+    type: string
 callNumbers:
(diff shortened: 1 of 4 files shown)


demiankatz:
@EreMaijala, I've implemented this as we discussed earlier today.

EreMaijala (commented):
Looks good to me, but tests are failing. Should this be added to the API as well?

demiankatz:
@EreMaijala, sorry about that -- I was trying to save time by skipping the integration tests, but I accidentally skipped all of the API tests as well, so I didn't catch the break. Should be fixed now!

I also somehow overlooked the fact that the building field was being defined in a test, but not anywhere real. That should be addressed now as well.

Do you mind taking another look?

EreMaijala (approved):

EreMaijala:
@demiankatz Thanks, looks good now!

PR #1942, opened 2021-05-04
Tests: Replace deprecated setMethods() with onlyMethods().

demiankatz:
PHPUnit has deprecated the setMethods() method in the mock builder. Fixing this reveals some problems with our test suite, which I will finish fixing as time permits. For now, this PR is a work in progress!

TODO:
- [x] Fix broken tests


demiankatz:
@EreMaijala, since all of these changes are in the test suite, and all tests are still passing, I think this should be pretty safe to merge, but if you don't mind giving it a quick look, I'd appreciate it.

Once again, this is a change to PHPUnit that feels like a real improvement. I never could quite remember/understand the purpose of `setMethods` and I am glad to see it deprecated. Now we have `onlyMethods` (only allow the specified methods to be mocked) and `addMethods` (allow mocking of methods that don't actually exist in the class being mocked). This better validates that the methods being mocked actually exist in the original classes, so you can't refactor code and have tests pass because the mock objects keep working even when the underlying functionality is gone.

Most of the changes here were pretty painless. In a few cases, I had to use `addMethods` instead of `onlyMethods` because mock objects are mocking behavior that is actually implemented with the `__call` magic method. The only place where I found myself fighting against this instead of working with it was in the MultiBackendTest, which is very convoluted and has a lot of execution paths. I ended up just cheating and sorting methods into add/only buckets to make it run. That might be worth revisiting in the future, but for now I think it's adequate.

EreMaijala (approved):
@demiankatz All in all looks good to me, but the MultiBackend tests bothered me enough to prompt a little cleanup. What do you think?

demiankatz:
Thanks, @EreMaijala, that's great and makes me completely happy with this change. I appreciate you taking the time to improve the tests!

PR #1943, opened 2021-05-05
Add Matomo view helper for Matomo 4.x.

EreMaijala:
The old Piwik helper remains for back-compatibility.

Compared to old Piwik helper this switches to the new tracker JavaScript style recommonded by Matomo. It also tries to improve handling of different rendering contexts so that actions get better names. It also adds support for using custom dimensions instead of or in addition to custom variables. Custom dimensions are recommended by Matomo.


module/VuFind/src/VuFind/View/Helper/Root/Matomo.php
```diff
@@ -0,0 +1,678 @@
...
+        $current = $viewModel->getCurrent();
+        if (null === $current || 'layout/lightbox' === $current->getTemplate()) {
+            return null;
+        }
+        $children = $current->getChildren();
+        if (isset($children[0])) {
+            $template = $children[0]->getTemplate();
+            if (!strstr($template, '/home') && !strstr($template, 'facet-list')) {
+                $results = $children[0]->getVariable('results');
+                if (is_a($results, 'VuFind\Search\Base\Results')) {
```
demiankatz: I realize that this probably just crept in from the existing Piwik helper, but I'd suggest changing `is_a()` calls to `instanceof` so that we can actually refer to the class directly in the code instead of using its name in a string. That improves static analysis and allows advanced IDEs to work better.

demiankatz (commented):
One small comment, but beyond that, I don't see obvious problems (beyond the potential fragility of some of the code for retrieving details from the view, which is an existing issue with the Piwik helper, and which likely has no easy solution). I'm going to ask our local Matomo expert to take a look at this in case he has additional thoughts.

At some point, we should update the [usage stats wiki page](https://vufind.org/wiki/configuration:usage_stats) with notes on the new helper, and suggestions for migrating from the old one without introducing inconsistencies to data. (I'm assuming you could just move identical settings over if you wanted to keep things the same, though switching to dimensions would be a good idea at some logical boundary like the start of a new year).

module/VuFind/src/VuFind/View/Helper/Root/Matomo.php
EreMaijala: Forgot to mention that I changed this.

demiankatz: Thanks!

demiankatz:
This gets the thumbs-up from our side! Merging now, though we may have more feedback once we actually try it out in production.

PR #1944, opened 2021-05-05
Alma: Return pickup location fields as strings.

EreMaijala:
Also adds a missing parameter.


demiankatz (approved):
I think this qualifies as a bug fix, so I have backported it to the release-7.1 branch and merged it to dev from there. A quick review of all the drivers revealed that the missing parameter problem also impacted the FOLIO driver, so I fixed that too as commit 48f949a1ffc4975ba5ae4330b3903c7b19978dd6. Thanks!

PR #1945, opened 2021-05-05
[VUFIND-1464] Add \Memcached support to Memcache session handler

demiankatz:
TODO:
- [x] Resolve [VUFIND-1464](https://vufind.org/jira/browse/VUFIND-1464) when merging.

### DIFF
--- a/module/VuFind/src/VuFind/Session/Memcache.php
+++ b/module/VuFind/src/VuFind/Session/Memcache.php
@@ -66,13 +66,13 @@ class Memcache extends AbstractBase
     /**
      * Set up the connection to Memcache.
      *
-     * @param Config                 $config Session configuration ([Session] section
+     * @param ?Config                $config Session configuration ([Session] section
      * of config.ini)
      * @param ?\Memcache|?\Memcached $client Optional Memcache client object
      *
      * @return void
      */
-    protected function connect(Config $config, ?object $client): void
+    protected function connect(?Config $config, ?object $client): void
     {
         // Set defaults if nothing set in config file.
         $host = $config->memcache_host ?? 'localhost';
--- a/module/VuFind/src/VuFind/Session/Memcache.php
+++ b/module/VuFind/src/VuFind/Session/Memcache.php
@@ -53,25 +53,62 @@ class Memcache extends AbstractBase
     /**
      * Constructor
      *
-     * @param Config    $config Session configuration ([Session] section of
-     * config.ini)
-     * @param \Memcache $client Optional Memcache client object
+     * @param Config                 $config Session configuration ([Session] section
+     * of config.ini)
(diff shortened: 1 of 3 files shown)


demiankatz:
@EreMaijala, I opened VUFIND-1464 a few months ago because you mentioned that PHP 7 and the \Memcache extension were incompatible. However, that does not seem to be the case -- I was able to successfully install both \Memcache and \Memcached with PHP 7 on my test server. In any case, while investigating this, I updated the Memcache session handler to support either client, which should help future-proof the code in case something really does break or get deprecated.

There's just one thing which I don't understand: one of the unit tests fails unless \Memcache::get() receives exactly three parameters, even though the documentation indicates that the method accepts two parameters, one of which is optional. I was able to make the test pass by sending some extra null parameters to the method, which should be harmless... but this is obviously an ugly hack. My best theory is that something in the \Memcache library is misreporting data to PHP's Reflection API, which is in turn making PHPUnit's mock library misbehave... but if that's the case, there's nothing I can do about it except the existing workaround. If you have any alternative theories or better solutions, though, I'd love to hear them!

demiankatz:
Also, a side note: I notice that the Travis build did not run here. I am getting an error message about a negative credit balance, even though my account shows a positive balance. Perhaps Travis is having some technical difficulties right now, but if this does not resolve itself in a day or two, I will try to remember to contact support.

demiankatz:
Actually, I take it back: I have a positive credit balance in my personal account, but the vufind-org account has run out. I'll see what I can do to resolve this with support!

demiankatz:
Update: I've opened a ticket, but following a bit more research, I think the [prospects don't look good](https://news.ycombinator.com/item?id=25338983).

EreMaijala:
@demiankatz Thanks, I'll take a look as time permits. I was indeed mistaken, and there is a current version of Memcache supporting PHP up to version 8.

module/VuFind/src/VuFind/Session/Memcache.php
```diff
@@ -122,7 +122,12 @@ protected function connect(?Config $config, ?object $client): void
      */
     public function read($sessId)
     {
-        $value = $this->connection->get("vufind_sessions/{$sessId}");
+        // For some reason, Memcache tests fail if we do not pass exactly three
+        // parameters to the get method, even though this seems inconsistent with
+        // the documentation. This mechanism makes the tests pass, but may be worth
+        // revisiting in the future.
+        $extraParams = $this->connection instanceof \Memcache ? [null, null] : [];
+        $value = $this->connection->get("vufind_sessions/{$sessId}", ...$extraParams);
```
EreMaijala: This line seems to be too long.

EreMaijala (approved):

PR #1947, opened 2021-05-06
Add ID to MARC subject heading data

aleksip:
Would be useful and performant to do it here.

### DIFF
--- a/module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
+++ b/module/VuFind/src/VuFind/RecordDriver/MarcAdvancedTrait.php
@@ -144,7 +144,8 @@ trait MarcAdvancedTrait
                         $retval[] = [
                             'heading' => $current,
                             'type' => $fieldType,
-                            'source' => $source
+                            'source' => $source,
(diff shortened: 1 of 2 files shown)


demiankatz:
@aleksip, this seems reasonable to me, but it is causing a test failure due to the addition of the new value. Do you think it is reasonable for `'id` to be set to `''` when no value exists, or would it be better to omit the key entirely? (I don't really have a preference; just asking whether you had considered both options).

aleksip:
@demiankatz Ah, I already updated the test before noticing your comment! I don't have a preference either, but to be honest I did not consider both options.

demiankatz:
Thanks, @aleksip. Since other fields here are defaulting to the blank string when unset, it makes sense to be consistent, so I'm happy to merge this as-is. Thanks!

aleksip:
@demiankatz Wow, that was fast, thank you!

demiankatz:
Well, it's the least I can do after how long #1886 is taking. (And speaking of that one, please let me know if you need me to do any more on it -- I've been waiting to see your response to Ere's comments).

PR #1949, opened 2021-05-07
Add a new cover image provider: Orb

fredericd:
Orb does stand for Outil de Recherche Biblbiographique (tool for
bibliographic search). It's a french database of bibliographic
information available by subscription.

- Web site: https://www.base-orb.fr
- API: http://doc.api.base-orb.fr/

### DIFF
--- a/module/VuFind/src/VuFind/Content/Covers/Orb.php
+++ b/module/VuFind/src/VuFind/Content/Covers/Orb.php
@@ -1,0 +1,131 @@
+<?php
+/**
+ * Orb cover content loader.
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2021.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Content
+ * @author   Frédéric Demians <f.demians@tamil.fr>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+namespace VuFind\Content\Covers;
+
+/**
+ * Orb cover content loader.
+ *
+ * @category VuFind
+ * @package  Content
+ * @author   Frédéric Demians <f.demians@tamil.fr>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:plugins:content_provider_components
+ */
+class Orb extends \VuFind\Content\AbstractCover
+    implements \VuFindHttp\HttpServiceAwareInterface
+{
+    use \VuFindHttp\HttpServiceAwareTrait;
+
+    /**
+     * Base URL for Orb API
+     *
+     * @var string
+     */
+    protected $url;
+
+    /**
+     * API user for Orb
+     *
+     * @var string
+     */
+    protected $apiUser;
+
+    /**
+     * API key for Orb
+     *
+     * @var string
+     */
+    protected $apiKey;
+
+    /**
+     * Constructor
+     *
+     * @param string $url     Base URL for Orb
+     * @param string $apiUser API key for Orb
+     * @param string $apiKey  API key for Orb
+     */
+    public function __construct($url, $apiUser, $apiKey)
+    {
+        $this->url = $url;
+        $this->apiUser = $apiUser;
+        $this->apiKey = $apiKey;
+        $this->supportsIsbn = $this->cacheAllowed = true;
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


fredericd:
@demiankatz is there a way to run test suite directly from command line ?

EreMaijala:
@fredericd See here: https://vufind.org/wiki/development:recommended_tools#checking_everything

fredericd:
Thanks @EreMaijala

module/VuFind/src/VuFind/Content/Covers/Orb.php
```diff
@@ -0,0 +1,129 @@
...
+
+        $url = 'https://' . $this->apiUser . ':' . $this->apiKey . '@' .
+               $this->url . '/products?eans=' . $ean . '&sort=ean_asc';
+        $result = $this->getHttpClient($url)->send();
+        if ($result->isSuccess()) {
+            $data = $result->getBody();
+            $json = json_decode($data, true);
+            foreach ($json['data'] as $title) {
+                if ($title['ean13'] == $ean) {
+                    $size = $size == 'small' ? 'thumbnail' : 'original';
```
demiankatz: `$size =` is doubled here for some reason.

module/VuFind/src/VuFind/Content/Covers/Orb.php
```diff
@@ -0,0 +1,129 @@
...
+                    return $title['images']['front'][$size]['src'];
```
demiankatz: Is it worth adding an `isset` check here, just in case the API returns something unexpected?

demiankatz (changes requested):
Thanks, @fredericd, this looks like a useful addition, and I appreciate your patience with all the linting requirements. I noticed just one small issue and had one other minor suggestion. Beyond that, just a couple of questions... Have you reviewed the service's terms of service to be sure that caching of images is okay? Are there any attribution requirements?

I expect this can get merged pretty quickly; I just want to be sure we have all the details in place and have the most reasonable defaults.

Thanks again!

module/VuFind/src/VuFind/Content/Covers/Orb.php
fredericd: It's the ternary operator. If $size == 'small', transform $size into 'thumbnail', otherwise 'original'. Is there anything wrong syntactically ?

module/VuFind/src/VuFind/Content/Covers/Orb.php
fredericd: Well, this API endpoint is supposed to always return ['data'], even whether the ean/isbn-13 is not found in their database. In this case, it return: 

```json
{
  "data":[]
}
```

fredericd:
> Thanks, @fredericd, this looks like a useful addition, and I appreciate your patience with all the linting requirements.

The linting is great.

> Have you reviewed the service's terms of service to be sure that caching of images is okay? Are there any attribution requirements?

No attribution is required. I have a customer who have a valid subscription to this service. I had various exchanges with "Orb". They agree with the way it is implemented.

module/VuFind/src/VuFind/Content/Covers/Orb.php
demiankatz: I think the problem is that you should avoid reassigning `$size` inside the loop... because if there is more than one value in `$json['data']`, and `$size` is set to `small`, then the first iteration will change it to `thumbnail`, and the second iteration will see that it no longer equals `small` and unexpectedly change it to `original`. I'd suggest using a different local variable, like `$sizeKey` just to ensure that doesn't happen.

module/VuFind/src/VuFind/Content/Covers/Orb.php
demiankatz: I'm just wondering what happens if for some reason the API endpoint returns a data array that lacks an images or front element. Maybe this is impossible. Just wondering if it would be worth having a safety like:

<pre>
if (!isset($title['images']['front'][$size]['src'])) {
    throw new \Exception("Unexpected missing $size image from Orb.");
}
</pre>

...and then if something changes or goes wrong, there will be a record in the error log, rather than a PHP notice and possible unexpected behavior.

demiankatz:
Great, thanks for confirming that Orb is happy with the implementation. Once the two issues we're discussing are resolved, I'll get this merged.

module/VuFind/src/VuFind/Content/Covers/Orb.php
fredericd: After the re-assignation, there is a return. So there is not risk... The return data structure returned by the API looks like this:

```json
{
  "data": [
    {
      "ean13": "9782247197217",
      "images": {
        "front": {
          "original": {
            "src": "https://products-images.di-static.com/image/orb-ec3af334e23ff377fc14ad32c1497305/9782247197217-475x500-1.jpg"
          },
          "thumbnail": {
            "src": "https://products-images.di-static.com/image/orb-ec3af334e23ff377fc14ad32c1497305/9782247197217-120x160-1.jpg"
          }
        }
      }
    }
  ]
}
```

I will add a test to be sure that a URL is set.

module/VuFind/src/VuFind/Content/Covers/Orb.php
fredericd: > I'm just wondering what happens if for some reason the API endpoint returns a data array that lacks an images or front element. Maybe this is impossible. Just wondering if it would be worth having a safety like:

> if (!isset($title['images']['front'][$size]['src'])) {
>     throw new \Exception("Unexpected missing $size image from Orb.");
> }

I will test as suggested, but in order to avoid returning an invalid url.

demiankatz:
Thanks for the latest adjustments. This is looking good to me! Merging now.

demiankatz (approved):

fredericd:
🍪 Thanks @demiankatz for your precious help, comments and quick integration.

PR #1950, opened 2021-05-10
Fix Stamen tile URLs for SSL compatibility.

demiankatz:
- Thanks to Leila Gonzales for the fix.
- See https://github.com/stamen/maps.stamen.com/issues/167 for details.


demiankatz:
Thanks for pointing this out on the vufind-tech mailing list, @lmgonzales. I've created this PR so we can get the fix into release 7.1.1. Do you mind reviewing my work to be sure this is exactly right?

demiankatz:
(Note that the one seemingly unrelated diff here is just the result of my editor cleaning up trailing whitespace in the file).

lmgonzales:
Thanks for updating the file @demiankatz. I checked the fix against our production server and the updated URL is correct.

demiankatz:
Thanks for the confirmation -- merging now!

demiankatz:
(Note that I also cherry-picked this back to the release-7.1 branch so it can be included in the next maintenance release).

PR #1951, opened 2021-05-10
Add call number prefixes to templates

bbusenius:
This pull request is for adding call number prefixes to the templates. Prefixes are supported by some integrated library systems. At our institutions we us prefixes such as `f` and `xf` to denote large format items. We also use prefixes such as `CD` and `DVD` for audio formats.

Changes in this request:

- Make the output of `getUniqueCallNumbers` return a keyed array with a display version or the call number and a prefix that can be used separately when needed.
- Update the `holdingsils.phtml` template to use the display version of the call number where appropriate.

Related to PR #1934


bbusenius:
The unintuitive thing here is that I'm only using the call number and the display version in the template. I never actually use the prefix but I still return it. The reason for this is because `display` is needed in `getUniqueCallNumbers` in order to ensure that we get all the unique call numbers, otherwise prefixed call numbers are ignored if the prefix is the only difference. Then, the display version is cleaner and easier to use in the templates than concatenating the prefix. But all that being said, we actually need the prefix separately in our custom code, which is why I still return it. Since we need it at our institution and it's data representing a richer form of the call number, I thought this was appropriate.

module/VuFind/src/VuFind/RecordTab/HoldingsILS.php
```diff
@@ -100,13 +100,31 @@ public function getDescription()
     public function getUniqueCallNumbers($items)
```
demiankatz: I would add a second `$fullDetails` parameter here to request the more detailed return format -- so if it's true, it returns an array of arrays, and if it's false, it returns plain call numbers as before. It can default to false. This way, if people have custom templates expecting the old return format, this change won't break them.

There are a few different ways you could approach this -- perhaps the least complicated (though also least efficient) would be to always build the full detailed array, but when only call numbers are requested, filter it down to just the call numbers. I have no objection to this approach if it seems best to you, but feel free to try something different if you have other ideas.

module/VuFind/src/VuFind/RecordTab/HoldingsILS.php
```diff
@@ -100,13 +100,31 @@ public function getDescription()
...
-        return array_unique($callNos);
+
+        uasort(
+            $callNos,
+            function ($a, $b) {
+                return $a['display'] <=> $b['display'];
+            }
+        );
+
+        $unique = [];
```
demiankatz, line 139: I wonder if it would be more efficient to deduplicate before you sort, since the sort operation is probably more expensive than the deduplication, especially if there are large numbers of values to compare with one another.

module/VuFind/src/VuFind/RecordTab/HoldingsILS.php
```diff
@@ -100,13 +100,31 @@ public function getDescription()
     public function getUniqueCallNumbers($items)
     {
         $callNos = [];
-        foreach ($items as $item) {
+        foreach ($items as $i => $item) {
             if (isset($item['callnumber']) && strlen($item['callnumber']) > 0) {
-                $callNos[] = $item['callnumber'];
+                $prefix = $item['callnumber_prefix'] ?? '';
+                $callnumber = $item['callnumber'];
+                $callNos[$i]['callnumber'] = $callnumber;
```
demiankatz: I think indexing directly into `[$i]` without first initializing an array may cause some warnings in some of the static analysis tools we are introducing to VuFind. It's usually better to initialize things explicitly. I wonder, in this case, if it would actually be more readable to do the work in variables and then use `compact` to turn them into an array... e.g.:

<pre>
$prefix = $item['callnumber_prefix'] ?? '';
$callnumber = $item['callnumber'];
$display = $prefix ? $prefix . ' ' . $callnumber : $callnumber;
$callNos[] = compact('callnumber', 'display', 'prefix');
</pre>

(This also eliminates some redundant calculations around prefix).

demiankatz: While we're messing with this code anyway, I also see that this existing line:

`if (isset($item['callnumber']) && strlen($item['callnumber']) > 0) {`

could be shortened to:

`if (strlen($item['callnumber'] ?? '') > 0) {`

If we want to slightly improve readability. (But maybe I'm just addicted to null coalescing). ;-)

demiankatz (changes requested):
I've added some thoughts and suggestions below. In addition to those, a couple other things we should do:

- Add callnumber_prefix random generation to the Demo driver to test the feature with
- Write a unit test for VuFind\RecordTab\HoldingsILS

I'm happy to help with either or both of these if you do not have time.

bbusenius:
Great feedback @demiankatz. I think the changes you suggested make it much nicer. I'm curious to see what you think about the way I handled the `$fullDetails` switch. I handled it a little differently than what you suggested. There were a few reasons but mainly it was because the sorting and deduping would have needed to be redone with the switch anyhow. In our data it is very common to have the same call number with different prefixes. This would have caused duplicate call numbers and to avoid it, I would have needed to put a bunch of `if` statements in the code or duplicate the loop (which I kind of do anyhow in the other function). Anyhow, I thought this was the simplest solution and easiest to read, given all those concerns. It will also be easy to rip out in the future if we ever want to. That was my thinking behind it.

I'm looking at the other stuff you mentioned now.

themes/bootstrap3/templates/RecordTab/holdingsils.phtml
```diff
@@ -80,15 +80,15 @@
 </h3>
 <table class="table table-striped">
   <caption class="sr-only"><?=$this->transEsc('holdings_details_from', ['%%location%%' => $this->transEsc($holding['location'])]) ?></caption>
-  <?php $callNos = $this->tab->getUniqueCallNumbers($holding['items']); if (!empty($callNos)): ?>
+  <?php $callNos = $this->tab->getUniqueCallNumbers($holding['items'], $fullDetails=true); if (!empty($callNos)): ?>
```
demiankatz: I think you just want `true` here rather than `$fullDetails=true`.

demiankatz (changes requested):
Thanks, @bbusenius, this looks good to me, and your approach to the legacy compatibility makes sense to me. As you say, it'll be easier to deprecate/prune out in future this way. Just one small comment, which I think is contributing to the CI failure we're currently seeing. After that, I'll stay tuned for your progress on next steps... just say the word if you need anything!

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -276,6 +276,17 @@ protected function getFakeCallNum()
...
+    /**
+     * Generate a fake call number prefix.
+     *
+     * @return string
+     */
+    protected function getFakeCallNumPrefix()
+    {
+        $codes = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+        return substr(str_shuffle($codes), 1, rand(0, 3));
+    }
```
bbusenius: @demiankatz I'm not sure that this is very useful. I just did it to get started. If we had a small handful of predetermined prefixes that were randomly chosen, it would be easier to distinguish them from the rest of the call number. Would that be preferable?

demiankatz: The test data doesn't need to be realistic, so maybe you could even do something like "Prefix" followed by a random digit.

bbusenius:
@demiankatz I think this is ready for another look. I've taken a stab at getting prefixes to display for the various configurations on the results page.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
```diff
@@ -171,7 +171,7 @@ protected function pickValue($rawList, $mode, $msg, $transPrefix = false)
         } elseif (count($list) == 0) {
             // Empty list?  Return a blank string:
             return '';
-        } elseif ($mode == 'all') {
+        } elseif ($mode == 'all' || $mode == 'group') {
```
demiankatz: Why is this change necessary? I'm just trying to figure out how it relates to everything else. According to the function comment, 'group' is not a $mode value that this method is expecting... but maybe the comment is wrong. :-)

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
```diff
@@ -230,6 +230,22 @@ protected function reduceServices(array $rawServices)
...
+     * @param string $prefix     Callnumber prefix or empty string.
+     * @param string $callnumber Main call number.
+     *
+     * @return string
+     */
+    protected function formatCallNums($prefix, $callnumber)
+    {
+        $displayCallnumber = !empty($prefix)
+            ? $prefix . ' ' . $callnumber : $callnumber;
+        return $callnumber . '::::' . $displayCallnumber;
```
demiankatz: Might be worth adding `$separator = '::::'` as a third parameter to the function, so it's easier to override if somebody wants to use something different.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/HoldingsILSTest.php
```diff
@@ -0,0 +1,88 @@
+<?php
+/**
+ * RecordTab Manager Test Class
```
demiankatz: You should update the comment to actually describe the class. :-)

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/HoldingsILSTest.php
```diff
@@ -0,0 +1,88 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Tests
+ * @author   Demian Katz <demian.katz@villanova.edu>
```
demiankatz: You can give yourself author credit here if you like!

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/HoldingsILSTest.php
```diff
@@ -0,0 +1,88 @@
...
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
+ */
+namespace VuFindTest\RecordTab;
+
+use VuFind\RecordTab\HoldingsILS;
+
+/**
+ * RecordTab Manager Test Class
```
demiankatz: Another comment to update.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
demiankatz: Or is it necessary to use a separator at all? Can you just return arrays here? I think the whole response gets JSON encoded at the end, so you may be able to work with a more complex data structure. (I apologize if I'm mistaken, though -- it's been a while since I've worked with this code!)

demiankatz (commented):
Thanks, @bbusenius, this is looking good! I left a few comments based on a review of the latest code. I haven't done a hands-on test yet, but I'll do so once you've had a chance to address these items.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
bbusenius: Ah yes, I meant to call this out and then forgot to. I was going to ask you about this. I felt like this might be fixing an unrelated bug but I wasn't sure. I might have misunderstood what I was seeing. Without this change I was getting an "Undefined" message (or something to that effect) on some holdings with multiple call numbers. This is probably unrelated to what I'm doing with the rest of the PR but I left it so we could have this conversation. From my point of view, it's fine to remove this but if you want me to leave it, that's fine to.

demiankatz: Thanks, @bbusenius. If you can remember what setup caused the error, that would be helpful, but if not, I'll do some experimentation and see if I can get to the bottom of whatever is going on. If this truly is a bug (which seems entirely possible), maybe I can open a separate PR to deal with that issue, add tests to reproduce it, etc.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
bbusenius: So, I thought about both of those options when I was working on this too... Here's why I didn't do them. The main reason was that due to the complexity of the code, I wanted to keep my footprint here as minimal as possible.

Regarding a more complex data structure, yes, everything is being JSON encoded. It's not happening here but it is happening at some point in the stack, however, everything seems to be expecting a string throughout the code, not just in the PHP but also the Javascript where this is being used. For example, here in the PHP `htmlentities` is being used on call numbers in various places and the `pickValue` method is imploding multiple call numbers into a string with a `,\t` separator as well. So, I was trying to be delicate but also keep the changes minimal. I think we can change all this but it will require more changes throughout. My fear is that there might be a reason for keeping these as strings that I'm not aware of.

Regarding the addition of the separator as a third parameter to the function, I didn't do this for two reasons, 1. that `,\t` was already being used in a prescriptive manner on call numbers and 2. (more importantly) whatever we use here must also be used in the Javascript. I thought it might make more sense to be prescriptive in this case for that reason because if someone overrides it here, they're also going to have to override it in the Javascript.

So that was my thinking when I did this. In summary, I think that adding the separator as a parameter to the function would be really easy but of less value due to the Javascript. Changing the data to an array should be doable but will require changes throughout. 

I'll defer to your judgement on how you think we should proceed.

demiankatz: Thanks, @bbusenius, that all makes sense. This may be worth revisiting in the future, but I agree that for the moment, your current approach is probably the simplest and least risky.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
bbusenius: No problem. When I remove `|| $mode == 'group'` and configure it like this:
```
multiple_call_nos = group
multiple_locations = group
```
I get something that looks like this:

image

When I add the `||` clause above, I get the following result which looked more correct to me:

image

This particular record is pretty crazy but I think this is right. There were lots of others when I did this though. Anyhow, we don't use this configuration and I don't know how common this would be in practice.

demiankatz: Thanks, @bbusenius, that's very helpful! I'm in a conference all day today, but I'll take a closer look early next week.

demiankatz: @bbusenius, I took a closer look at this, and I think it's an unnecessary change: the [documentation in the .ini file](https://github.com/vufind-org/vufind/blob/dev/config/vufind/config.ini#L1726-L1727) indicates that the "group" setting is only valid for the multiple_locations setting. It is not supported for multiple_call_nos. Perhaps we should actually change this so that an exception is thrown if you provide an unsupported option for this setting. What do you think?

bbusenius: Oh right. Gosh, now that you mention it, I remember reading that at some point. I think I might have forgotten after I let this sit for awhile. I reverted that part of the code.

I'm agnostic about the exception. I'm not sure if that would be helpful or not. I'm happy to do it if you think it would be best though.

demiankatz: The exception might have saved you the trouble of digging into this in the first place, but other than that, I think the benefit is pretty limited. I'm happy to just leave it alone for now. :-)

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/HoldingsILSTest.php
bbusenius: This should be done as of `914397652e`, though it's reluctant to update here for some reason.

demiankatz: Yes, I'm seeing the changes. Maybe you are looking at a link to a specific old commit, but if you go to the files tab of the PR, it should be up to date.

demiankatz:
Thanks again for the progress, @bbusenius. I'm working on updating the wiki to add the callnumber_status field so that the documentation will match the code when we get this merged. I have also just run the full integration test suite, and I have encountered several failures:

<pre>
There were 10 failures:

1) VuFindTest\Mink\BasicTest::testAjaxStatus
--- Expected
+++ Actual
@@ @@
-'A1234.567'
+'A1234.567::::A1234.567'

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/BasicTest.php:75
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

2) VuFindTest\Mink\CallnumberBrowseTest::testFirstAndMsg
Failed asserting that '/vufind_test/Alphabrowse/Home?source=dewey&from=CallNumberOne' contains "Prefix: 9 CallNumberOne".

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:103
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:160
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:183
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:201
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

3) VuFindTest\Mink\CallnumberBrowseTest::testAllAndMsg
Failed asserting that '/vufind_test/Alphabrowse/Home?source=dewey&from=CallNumberOne' contains "Prefix: 2 CallNumberOne".

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:103
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:160
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:183
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:212
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

4) VuFindTest\Mink\CallnumberBrowseTest::testFirstAndGroup
Failed asserting that '/vufind_test/Alphabrowse/Home?source=lcc&from=CallNumberOne' contains "Prefix: 6 CallNumberOne".

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:103
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:160
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:185
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:234
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

5) VuFindTest\Mink\CallnumberBrowseTest::testAllAndGroup
Failed asserting that '/vufind_test/Alphabrowse/Home?source=dewey&from=CallNumberOne' contains "Prefix: 6 CallNumberOne".

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:103
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:160
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:183
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:245
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

6) VuFindTest\Mink\CallnumberBrowseTest::testStatusFull
Failed asserting that '/vufind_test/Alphabrowse/Home?source=dewey&from=CallNumberOne' contains "Prefix: 2 CallNumberOne".

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:103
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:160
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:183
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:266
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

7) VuFindTest\Mink\CombinedSearchTest::testCombinedSearchResults
--- Expected
+++ Actual
@@ @@
-'A1234.567'
+'A1234.567::::A1234.567'

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CombinedSearchTest.php:87
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CombinedSearchTest.php:113
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

8) VuFindTest\Mink\CombinedSearchTest::testCombinedSearchResultsAllAjax
--- Expected
+++ Actual
@@ @@
-'A1234.567'
+'A1234.567::::A1234.567'

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CombinedSearchTest.php:87
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CombinedSearchTest.php:136
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

9) VuFindTest\Mink\CombinedSearchTest::testCombinedSearchResultsMixedAjax
--- Expected
+++ Actual
@@ @@
-'A1234.567'
+'A1234.567::::A1234.567'

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CombinedSearchTest.php:87
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CombinedSearchTest.php:158
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

10) VuFindTest\Mink\IlsActionsTest::testCancelIllRequest
Selector not found: .alert.alert-info a

/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:251
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:262
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:395
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

FAILURES!
Tests: 1528, Assertions: 5844, Failures: 10.
</pre>

It looks like we may have a combination of tests that need to be updated to reflect the new functionality, and tests that are finding actual problems caused by the changes here. I haven't had a chance to sift through it all yet. (I also think that last test failure, in VuFindTest\Mink\IlsActionsTest::testCancelIllRequest, may be a red herring; possibly just a random timing-related failure, though we'll have to re-run the suite to see if it consistently appears).

I am definitely seeing a problem in my test environment where the entire delimited call number string is coming through to the user interface instead of getting processed properly; see this screen shot:
image

Any idea what might be wrong here? Are you not seeing the same thing?

I should also note that I did merge the latest dev branch into yours before running tests, so that may account for any inconsistencies in line numbers in test classes, etc. I wasn't able to push the merge up to your repo, because I don't have permission to add workflows, and the merge includes the changes where we transitioned from Travis to GitHub Actions for CI.

Let me know if and when you want me to dig further into any of this and/or re-run the test suite. For now, I'm going to spend my remaining time cleaning up the ILS driver spec in the wiki, following Ere's lead on converting some of it into a more readable tabular format. Adding the new field is a good excuse to do this work!

demiankatz:
Regarding the bug in the screen shot, I should also note that this is with the "Sample" ILS driver in place, as it is in a clean out-of-the-box installation. I wonder if perhaps there is a bug when the callnumber_prefix is unset rather than empty, or something like that... but that's just a hunch, and not supported by a consultation of the code (haven't had time to refresh my memory on all the details today... just wanted to get these tests run so we could see the results and address any issues).

bbusenius:
The bug in the screenshot is strange. I have not seen anything like that and I cannot reproduce it with the Demo driver or the Folio driver. I had `calnumber_prefix` hacked into the Folio driver to test with. When I remove it entirely, I still do not see the the entire delimited call number string, not in any browser or in any theme. I'll keep looking into it. In the meantime, should I just merge the dev branch in myself and push it up? I'm happy to do that. It will answer 1 question anyhow.

bbusenius:
I should also add, I don't see it when I use the sample driver either.

demiankatz:
Thanks, @bbusenius! Yes, if you can merge in dev and push it up, that would be helpful, and then I should once again be able to push here if necessary.

Regarding the bug in my screen shot, let me rebuild my test environment just to be sure it wasn't a glitch related to something weird that I did. (I'm pretty sure I did the right thing, but I've been juggling a lot of projects today, so maybe I just did something dumb ;-) ).

bbusenius:
Merged and pushed. I'm still not seeing the problem but I'll keep looking.

demiankatz:
@bbusenius, I tried again, and I'm still seeing the same problem. I just checked out your branch and spun up a standard VuFind test instance, as described in the first [VuFind Testing video](https://vufind.org/wiki/videos:testing_1). I also double-checked that it's not a caching-related problem: when I view the source of the Javascript code, I see the :::: delimiter logic. If you're still unable to reproduce the problem, please let me know -- I'm happy to check things on this end if it would help.

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -2,14 +2,16 @@
...
       var cns = callnumber.split(',\t');
       for (var i = 0; i < cns.length; i++) {
-        cns[i] = '<a href="' + VuFind.path + '/Alphabrowse/Home?source=' + encodeURI(callnumber_handler) + '&amp;from=' + encodeURI(cns[i]) + '">' + cns[i] + '</a>';
+        var pair = cns[i].split(delim);
+        cns[i] = '<a href="' + VuFind.path + '/Alphabrowse/Home?source=' + encodeURI(callnumber_handler) + '&amp;from=' + encodeURI(pair[0]) + '">' + pair[1] + '</a>';
       }
       return cns.join(',\t');
     }
-    return callnumber;
+    return callnumber.split(delim)[1];
```
bbusenius: :man_facepalming:

bbusenius (commented):
@demiankatz thanks for the testing video. That's super useful! I was able to look at this today. The problem was because I wasn't handling for when there isn't a call number handler! This last push should fix the visual problem. It may or may not fix the tests. I spun up the test environment with Vagrant, which got me to the point where I could see the error you were seeing, however, it's not able to run the tests properly. I might need to build a new environment before I can look at the test failures.

demiankatz:
Thanks, @bbusenius, that makes sense, and I'm glad that the video was useful. I'm out of time for this week but will re-run the tests next week and let you know what I find!

bbusenius:
@demiankatz, I was able to run the CI tests in a new test environment and can see that they're still failing due to the call number change. I'm looking at updating the tests now.

bbusenius:
@demiankatz I may have spoken too soon. I think my other VuFind environment was interfering with my CI environment. I shut everything down and started fresh today and now the CI tests appear to be passing. When you have time, please re-run them with the new code to confirm.

demiankatz:
@bbusenius, I merged the latest dev branch into your branch and ran the full test suite; I'm still getting quite a few failures. I'm going to re-run the tests on dev, just to be 100% sure that everything still passes there (though I have no reason to expect it not to, since I've been testing quite frequently). I'm not sure how/why some of these failures are related to your changes, so it's once again possible that there are some red herrings in the mix... but it also seems like some of them really are problems, whether they indicate bugs or suggest that we need to adjust the tests to account for the new reality.

Here's the list of failures -- let me know if this gives you any ideas, or if you need me to spend some more time with this. I'm happy to help, though my backlog keeps getting bigger, so it may take me a while to catch up. :-)

<pre>
There were 10 failures:

1) VuFindTest\Mink\CallnumberBrowseTest::testFirstAndMsg
Failed asserting that '/vufind_test/Alphabrowse/Home?source=dewey&from=CallNumberOne' contains "Prefix: 3 CallNumberOne".

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:103
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:160
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:183
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:201
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

2) VuFindTest\Mink\CallnumberBrowseTest::testAllAndMsg
Failed asserting that '/vufind_test/Alphabrowse/Home?source=lcc&from=CallNumberOne' contains "Prefix: 7 CallNumberOne".

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:103
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:160
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:185
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:212
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

3) VuFindTest\Mink\CallnumberBrowseTest::testFirstAndGroup
Failed asserting that '/vufind_test/Alphabrowse/Home?source=dewey&from=CallNumberOne' contains "Prefix: 1 CallNumberOne".

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:103
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:160
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:183
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:234
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

4) VuFindTest\Mink\CallnumberBrowseTest::testAllAndGroup
Failed asserting that '/vufind_test/Alphabrowse/Home?source=dewey&from=CallNumberOne' contains "Prefix: 2 CallNumberOne".

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:103
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:160
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:183
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CallnumberBrowseTest.php:245
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

5) VuFindTest\Mink\CartTest::testCartExportToThirdParty
--- Expected
+++ Actual
@@ @@
-'https://www.google.com/'
+'http://localhost/vufind_test/Search/Results?lookfor=id%3A(testsample1+OR+testsample2)'

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CartTest.php:703
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

6) VuFindTest\Mink\CombinedSearchTest::testCombinedSearchResultsAllAjax
--- Expected
+++ Actual
@@ @@
-'A1234.567'
+'Loading...'

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CombinedSearchTest.php:87
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CombinedSearchTest.php:136
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

7) VuFindTest\Mink\FavoritesTest::testAddSearchItemToFavoritesLoggedIn
Failed asserting that 3 matches expected 1.

/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/FavoritesTest.php:402
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

8) VuFindTest\Mink\HoldsTest::testCancelHold
Selector not found: .alert.alert-success

/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:251
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php:266
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php:303
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

9) VuFindTest\Mink\IlsActionsTest::testStorageRetrievalRequest
Selector not found: .modal-body select

/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:251
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:174
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:293
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:414
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

10) VuFindTest\Mink\IlsActionsTest::testCancelStorageRetrievalRequest
Selector not found: .alert.alert-info a

/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:251
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:285
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:443
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

FAILURES!
Tests: 1532, Assertions: 5974, Failures: 10.

BUILD FAILED
/home/dkatz/vufind3/build.xml:202:385: Task exited with code 1

Total time: 19 minutes  42.04 seconds
</pre>

demiankatz:
Just a quick confirmation that the dev branch has no failing tests, so these do all somehow seem to be related to the changes here.

bbusenius:
Thanks @demiankatz. I don't know why tests passed for me yesterday but I was able to recreate some errors today. I should think that the change I just pushed will fix errors 1-4. I think error 10 is a red herring. 5-9 look like real problems to me, however, I am not able to recreate those errors in my environment. I think they might be getting skipped for some reason.

demiankatz:
Thanks for the progress, @bbusenius -- when I have a chance, I'll re-run the tests and see if I can figure out what's going on with those remaining ones. This week has gotten unexpectedly crazy, so it may take me a few days to get back around to it. Apologies in advance!

demiankatz:
@bbusenius, thanks for fixing the broken tests. The other failures were indeed red herrings, and I think I have addressed them in the dev branch now. (Not sure why the timing was working out differently here than there, but that seems to be all it was).

I gave this one more review, and I was still a little bit uncomfortable with always using the delimited format in the AJAX return value, since it feels like a pretty dramatic change from past behavior that could break locally customized code... then I had a brainstorm: if we only use the delimiter to separate the prefix from the main call number, then the code is significantly more backward-compatible, since the output only changes if we take advantage of the newly-added feature. I put in commit 2cba9b3 to implement things this way, and all tests are still passing. Do you mind reviewing my work and letting me know if you have any concerns about it? If you're happy with the changes, I think we can merge this now.

demiankatz:
Sorry, I broke the build with a minor style problem in my JS -- should be fixed by 8bf46b2.

bbusenius:
@demiankatz I really like this change. Not only is it more backwards compatible, it's more in line with what I did in the `status-full` template. I don't see any problems with the code and it's passing CI tests. I also tested it with the demo driver and Folio driver (with call number prefixes hacked in) and everything seems to be working well. This has my approval. I appreciate all your patience on this one.

demiankatz:
Thanks, @bbusenius! I realized that following my changes, the function name and comment no longer entirely made sense, so I've just adjusted those in commit 6f32732. I'm re-running the test suite out of an abundance of caution, but I expect it to pass, and then I'll get this merged! I appreciate all of your work on this.

demiankatz:
...and it's all good. Merging now!

PR #1952, opened 2021-05-11
More flexible Solr update support

demiankatz:
VuFind's code for sending updates to Solr is fairly XML-centric. This pull request increases flexibility by making it easier to send CSV and JSON data directly to Solr.

TODO
- [x] Run full test suite
- [x] Manually test various documents/encodings


demiankatz:
@EreMaijala, this code is designed to make it easier to build VuFind tools that send different types of commands to Solr, beyond the assumed XML framework. It includes some work that I did for #1761 but which I don't expect to ultimately use there (and I expanded upon that to add better JSON support here, which I think that I will end up using there). Do you have any concerns or objections to merging this?

demiankatz:
(I also wonder if we should design something better than this `asCSV`/`asJSON`/`asXML` pattern... it seems like it might make more sense if the document objects simply returned a body and a mime type, instead of doing this dance with different format-specific methods.... but that's a pretty fundamental redesign, and probably a conversation for another day).

EreMaijala:
@demiankatz Do you expect to support format conversions, e.g. asking for JSON from an XML document? If not, I think it would make it more clear if the document itself would dictate the content type, so you wouldn't need to pass two parameters that need to be compatible with each other.

demiankatz:
@EreMaijala, I can't think of a compelling reason to do format conversion -- seems likely to introduce problems, and I can't think of a real problem that it solves. But I also haven't thought very hard. :-) In any case, I assume that conversation applies to future developments -- did you have any concerns with the PR at hand?

EreMaijala:
@demiankatz In that case I'd propose a radical departure from the current AbstractDocument. Rather make it something like:

```
abstract class AbstractDocument
{
    /**
     * Return content mime type.
     *
     * @return string
     */
    abstract public function getContentType();

    /**
     * Return serialized representation.
     *
     * @return string
     */
    abstract public function serialize(); // Could be getContent() or toString() or something else as well
}
```

Then you wouldn't need to pass the content type around alongside the document or handle cases where wrong format was requested.

Apart from that the changes make sense. I'd just like to avoid having two parameters to describe one thing.

demiankatz:
@EreMaijala, I think your proposal is a sound approach, and it will allow us to get rid of a whole lot of unfinished TODO items in the process. A few thoughts/questions, though:

1.) If we're making this drastic a change, I would get rid of AbstractDocument entirely and create DocumentInterface for greater flexibility. Maybe we could end up with a new inheritance chain with RawDocument at the root (which takes content and mime type in the constructor), which could have XMLDocument/JSONDocument/CSVDocument as children (which have the mime type hard-coded in and just take content in the constructor), and then the more specific documents (update/delete/etc.) could stand alone. That might also be overkill since this doesn't really need to be that complicated... but it could work.

2.) For XML, we're currently using the MIME type `text/xml; charset=UTF-8` -- do you think we need to manage the content type and encoding separately within our classes for greater flexibility? Maybe that becomes a third constructor parameter to RawDocument.

3.) Outside of the obvious changes to the document classes, the biggest potential backward compatibility break here is the removal of the `$format` parameter from `\VuFind\Solr\Connector::write()`. I can think of a few possible approaches here, but I don't have a really strong preference for which one to use:

a.) We can just remove the parameter, update the changelog, and deal with the consequences.

b.) We can add some transitional logic so that the $format parameter becomes semi-optional (i.e. default it to 'auto', issue a deprecation warning if 'xml' or 'json' is passed in, and for any other value, shift the parameters over so that `$format` is used as `$handler` and `$handler` is used as `$params`). This is not pretty, but it would reduce breakage, and then we could fully rip it out in the next major version.

c.) We can deprecate the whole `write()` method and move the new version into a different method (maybe `sendDocument()`, which perhaps offers a clearer distinction than `send` vs. `write`, as far as existing methods go).

4.) Do you think we should solve all this here in this PR, or should I merge this as a transitional step and then open a new one for the bigger refactoring?

EreMaijala:
@demiankatz I'd like to keep the document class such that it can be built gradually if necessary. If the constructor has to take the raw string, the document has to be complete at the time it's created, right? That's why I was thinking that serializing the results just in time would provide more flexibility, but then I'm not sure if that's a plausible use-case.

I'm not sure which option to prefer with the write method as I'm not sure how probably it would have been used in a customized way. We're not using any writes within VuFind, but others could, so I'm not sure. If you think we could get away with breaking back-compatibility, I'd just do that now. I like the idea of using a better name, though, so leaving write for back-compatibility would be fine.

demiankatz:
@EreMaijala, I agree that we want flexibility with the document classes. I'm not proposing that ALL document classes should take their document in through the constructor, just that ones that are used for sending pre-existing, pre-built documents (the "Raw" family of classes) should do so. That's why I propose defining an interface and using that as the typehint everywhere, instead of continuing to have every document class inherit from the same abstract base class. That way the raw classes can use shared functionality, but the more special-purpose ones can just implement the interface in their own, more appropriate way. If we need something more dynamic but general-purpose, we could also define DynamicJsonDocument, DynamicXmlDocument, etc. as opposed to the Raw versions... but I'd wait for a use case before trying to design those, of course.

Regarding breaking changes, now that I think about it a little more, it occurs to me that if we do go forward with changing the typehint on `write()` from a base class to an interface, that's already a backward compatibility break (and one which will have obvious consequences, rather than potentially subtle/confusing ones like removing a parameter). That makes me a lot more comfortable with simply removing the parameter as part of the same change.

EreMaijala:
@demiankatz Right, I somehow dropped the change to an interface from my thinking. Now it all makes sense to me!

demiankatz:
@EreMaijala, what do you think of this? I think I've incorporated all the ideas we discussed, plus I added typing to everything for good measure. The full test suite is passing, but I'd like to do some more direct hands-on testing when time permits (probably some time next week).

demiankatz:
Hmm, looks like something I did (probably related to typing) doesn't work in PHP 7.3. (PHP 7.4 and 8.0 seem fine). I'll investigate further next week!

demiankatz:
Update: I fixed the broken build. Apparently PHP 7.3 doesn't like type declarations on class properties; I hadn't realized you need 7.4 for that! I haven't done any hands-on testing yet beyond what the test suite checks, but I'll wait for feedback before spending more time checking all the edge cases.

EreMaijala (commented):
@demiankatz All in all this looks much better than before. The only issue I could find is that the `@author` tags may be out of date in several files.

demiankatz:
Thanks again for all the input, @EreMaijala. I've manually tested everything, and my hands-on results confirm what the test suite is already saying: everything seems to be fully functional. I'm merging this now and will work on updating #1761 next.

PR #1954, opened 2021-05-13
Update dependencies (including many PHP 8 fixes).

demiankatz:
This PR brings more of VuFind's dependencies up to date; many of these updates involve PHP 8 fixes.

Some specific notes:
  - The "composer/package-versions-deprecated" package was added to allow compatibility between Composer 2 and PHP 7.3; it can be removed when we raise our minimum requirement to PHP 7.4.
  - The endroid/qr-code package has fixed PHP 8 compatibility, but it has also introduced many breaking changes; while its changelog did not make this especially clear, I ended up having to significantly rewrite our integration code.
  - The laminas/laminas-cache package now includes a mechanism for excluding unneeded cache adapters. When it upgrades to 3.0.0 in the future, adapters will switch to being "opt-in" rather than "opt-out" (and I expect that some will be deprecated). For now, all adapters are still being automatically included with VuFind for maximum flexibility, but we will have to revisit this soon.
  - The laminas/paginator package has deprecated the DbSelect adapter (which we use in just one place), moving it to a separate package (laminas/laminas-paginator-adapter-laminasdb). I have included this package and switched from the deprecated class to the supported one. Obviously, this is another sign that the whole laminas/laminas-db package is being phased out and will need to be replaced eventually.
  - I opted to only upgrade league/commonmark to 1.5.8 instead of the latest 1.6.x, since that newer release introduces new deprecations and may require documentation adjustments. That can be addressed later as a separate PR.

TODO:
- [x] Run full test suite
- [x] Confirm that QR codes generate correctly
- [x] Confirm that tag pagination still behaves correctly
- [x] Update changelog to note QR code BC breaks (when merging)


demiankatz:
@EreMaijala, as a follow-up to the tool updates in #1940, I spent some time updating core dependencies here. I think this is all reviewed, tested, and ready to go, but some of these changes turned out to be more significant than I had anticipated -- see notes above. Please let me know if you have any concerns before I merge this. Thanks!

EreMaijala (approved):
@demiankatz Thanks! No concerns, and with the new QRcode library the generation works also on my PHP 8 installation, a nice surprise.

demiankatz:
Thanks, @EreMaijala -- this is merged now.

PR #1955, opened 2021-05-14
de.ini: Fix incorrect translation of delete_account_description_html

mtrojan-ub:


demiankatz:
Thanks, @mtrojan-ub, since this is a fix, I have backported it to the release-7.1 branch and merged it to dev from there.

One note: it appears that you pushed this branch directly to the main vufind-org/vufind repository, instead of to your personal one. I suspect this was accidental, but you have this power now that you are a committer. It is not a problem, but I am trying to keep the branches in the main repository as clean/simple as possible, so I try to keep pull request branches outside to reduce clutter. In any case, I will delete this branch now that it has been merged!

mtrojan-ub:
Thanks for merging! Sorry for the effort. Since this was a very minor change, instead of creating a new branch in our repository, i noticed that there was the possibility of using the edit mask directly on github and setting an option to create a branch there, which seemed to be the easier way to do it. In the view, there has been an option called "Create a new branch for this commit and start a pull request" which i haven't seen before, that's why i decided to try it out. Of course, if you would like to keep pull requests outside, i will go back to the old procedure & create a branch in our own repository in the future.

demiankatz:
Thanks for the explanation, @mtrojan-ub, that makes sense! If you have another simple fix in the future, and this approach saves you time, that's fine with me since I can merge these things quickly and delete the branches when done. As long as we don't create long-lived pull requests with a lot of discussion based on branches in the main repo, everything should be fine.

PR #1956, opened 2021-05-17
Add support for hold start date and hold modification

EreMaijala:
(N.B. A set of separate but related commits)

This set of changes adds the following:

- Support for specifying start date for a hold 
- Validation of dates before sending them to the ILS driver
  - There's some variation on how things are handled in each driver, and this allows them to be simplified.
- Modification of holds

Notes:

- The markup for the Select all checkbox in holds.phtml is a bit odd, but it's the same as in the checkouts page.
- Form handling is a bit complicated due to the no-js support for canceling requests.
- Update won't work without JS, so the button is only displayed via JS.
- Creation of $safeId is changed to work properly when there are multiple holds on a single bibliographic record.
- The date handling cleanup is not yet reflected in ILS drivers.
- "Start date" and "Frozen until" are mostly the same at least for the drivers involved at the moment, but they're handled as different ones based on what's logical.

TODO: 

- [ ] Update https://vufind.org/wiki/development:plugins:ils_drivers
- [ ] Clean up more drivers?


languages/en.ini
```diff
@@ -514,6 +514,15 @@ hold_cancel_success = "Your request was successfully canceled"
 hold_cancel_success_items = "%%count%% request(s) were successfully canceled"
 hold_date_invalid = "Please enter a valid date"
 hold_date_past = "Please enter a date in the future"
+hold_edit_failed_items = "Updating failed for %%count%% hold(s)"
+hold_edit_frozen = "Frozen Status"
+hold_edit_frozen_set = "Freeze (inactivate)"
```
demiankatz: I wonder if "suspend" or "temporarily disable" would be better wording than "inactivate."

config/vufind/KohaRest.ini
```diff
@@ -69,7 +69,8 @@ HMACKeys = item_id:holdtype:level
 defaultRequiredDate = 0:0:2
 
 ; extraHoldFields - A colon-separated list used to display extra visible fields in the
-; place holds form. Supported values are "requiredByDate" and "pickUpLocation"
+; place holds form. Supported values are "startDate", "requiredByDate" and
+; "pickUpLocation"
```
demiankatz, line 73: I wonder if we should link to a wiki page with definitions of these values, or something -- I think we're getting away from everything being totally self-explanatory, but I don't want to have to maintain the same documentation in dozens of different .ini files. Open to suggestions!

module/VuFind/src/VuFind/Controller/HoldsTrait.php
```diff
@@ -192,6 +226,7 @@ public function holdAction()
                 'homeLibrary' => $allowHomeLibrary
                     ? $this->getUser()->home_library : '',
                 'extraHoldFields' => $extraHoldFields,
+                'defaultStartDate' => $defaultStartDate,
```
demiankatz: It wouldn't take too much to convert this whole array assignment into a `compact` statement. I wonder if it's worth doing that since we're adding even more to the list.

module/VuFind/src/VuFind/Controller/MyResearchController.php
```diff
@@ -1185,13 +1185,34 @@ public function holdsAction()
             return $view->cancelResults;
         }
 
-        // By default, assume we will not need to display a cancel form:
+        // Process update requests if necessary:
+        $updateResultsContainer = $this->getHoldUpdateResultsContainer();
+        $holdUpdateResults = $updateResultsContainer->results ?? null;
+        if ($holdUpdateResults) {
+            $view->updateResults = $holdUpdateResults;
+            $updateResultsContainer->results = null;
```
demiankatz, line 1193: It took me a while to understand why you would want to do this (I didn't realize at first this was a session container). Might be worth adding a comment.

module/VuFind/src/VuFind/Controller/MyResearchController.php
```diff
@@ -1220,6 +1253,155 @@ public function holdsAction()
         return $view;
     }
 
+    /**
+     * Edit holds
+     *
+     * @return mixed
+     */
+    public function editHoldsAction()
+    {
```
demiankatz, line 1263: There's an enormous amount of code here... given that the MyResearchController is already unreasonably huge, I'd love to avoid making it dramatically larger still. Maybe refactoring MyResearchController is a project for another day and we should just live with this for now, but if there's a logical way to move more code to a helper or support class, that might save us time down the road.

themes/bootstrap3/js/lightbox.js
```diff
@@ -327,7 +327,7 @@ VuFind.register('lightbox', function Lightbox() {
     _lightboxTitle = submit.data('lightbox-title') || $(form).data('lightbox-title') || false;
     // Get Lightbox content
     ajax({
-      url: $(form).attr('action') || _currentUrl,
+      url: $(form).attr('action') || _currentUrl || window.location.href,
```
demiankatz, line 330: Just curious what this change is for...

demiankatz (commented):
I didn't have time for an especially thorough review, and I haven't actually tested anything yet, but here are a few initial comments based on a fairly cursory examination of the code.

config/vufind/KohaRest.ini
EreMaijala: That'd make sense. We'd still have to list the supported values in each .ini file as they differ per driver, but further explanation could be on a wiki page. Maybe a sub-page of https://vufind.org/wiki/configuration:ils ?

module/VuFind/src/VuFind/Controller/MyResearchController.php
EreMaijala: I tried initially to put this in the holds plugin like `cancelHolds` is done, but interfacing it with the action method became too complicated. I suppose something like separate controllers for different parts of MyResearch (like LibraryCardsController now) would be a more natural way of splitting things up.

themes/bootstrap3/js/lightbox.js
EreMaijala: I forgot to mention: Lightbox didn't work on forms that don't have an action attribute (the hold selection form is such one). _currentUrl is not yet set at the moment the initial request is made, so we need to use the current page url like the form would have used if submitted normally.

module/VuFind/src/VuFind/Controller/MyResearchController.php
EreMaijala: Comments added.

module/VuFind/src/VuFind/Controller/HoldsTrait.php
EreMaijala: Good point. Now done.

languages/en.ini
EreMaijala: Updated, ok?

config/vufind/KohaRest.ini
demiankatz: Yes, that makes sense to me. Then we could link to this page from the driver spec as well as from the .ini files, and it would define a common vocabulary of settings that the individual drivers could choose from as appropriate. Let me know if you need me to create a stub page for you.

module/VuFind/src/VuFind/Controller/MyResearchController.php
demiankatz: Yes, I have been wanting to split up MyResearchController for years, but I just never find the time since it's not broken, it's just ugly. Ideally, at the same time we split this up, we could also switch to dependency-injected controllers to move away from explicit reliance on the service manager (as discussed in [VUFIND-966](https://vufind.org/jira/browse/VUFIND-966)). So many improvement projects, so little time! Unfortunately, I suspect the increasing urgency of replacing laminas-db is going to take up a lot of my time before I get around to working on this important but less-vital goal.

In any case, getting back down to the task at hand, if it would make sense to establish a HoldsController and start splitting some functionality away now, maybe that would offer an excuse to begin some of the necessary refactoring... but if that makes the rest of this project harder, we can just leave it as-is for now and deal with the refactoring later.

themes/bootstrap3/js/lightbox.js
demiankatz: Thanks, that makes sense. I wonder if we should split off this one-liner into its own pull request for separate testing/merging. (If you think it's worth it but don't have the time to set it all up, I'm willing to do the work).

EreMaijala: That'd make sense, and I'd appreciate help on this.

demiankatz: Okay, extracted to #1959.

demiankatz: I merged #1959 to dev and then merged dev back to here, so these changes are no longer part of the diffs in this PR. Thanks!

EreMaijala: Thanks!

module/VuFind/src/VuFind/Controller/MyResearchController.php
EreMaijala: Introducing a HoldsController now would make sense. I'll just do that and move HoldsAction there as well. Do you have a preference on whether to keep the holds.phtml template in the current location for back-compatibility? I'm tempted to move it as well to make its location standard for the future, but that'd be an incompatible change. I think that keeping the old route is necessary.

demiankatz: I don't object to moving holds.phtml as long as we document it in the changelog. Would it make sense to move the existing Holds functionality as a separate PR and then merge that work here, to keep things broken into smaller chunks? If that makes the work harder, though, I'm fine with having it done here -- we'll just want to add a TODO checkbox about the changelog so I don't lose track of it.

EreMaijala: I almost put it here, but decided it's so much cleaner to do the refactoring separately and posted PR #1961. Now I just need it in to continue here...

config/vufind/KohaRest.ini
EreMaijala: Added page https://vufind.org/wiki/configuration:ils:holds

demiankatz: Thanks, this is great! The only problem was that the extraHoldFields text was not displaying neatly, but I reformatted it as a table and now it looks fine.

EreMaijala: Awesome, the table is much nicer!

demiankatz: ...so now the last step is to link to this page from the relevant .ini files. Let me know if you'd like help getting that done!

EreMaijala: I can add the links, but if you'd like to update the comments in the ini files as well, I'd appreciate help.

demiankatz: Why don't you add it to KohaRest.ini and Demo.ini as part of this pull request, since you're already modifying those two files, and I'll add it to all the other ones that aren't part of the PR?

demiankatz: ...actually, I'll take care of KohaRest.ini too, since I'll just add a comment above the `[Holds]` section which won't conflict with anything else. I can also paste the same comment into Demo.ini here. I'll let you know when it's all done!

EreMaijala: Thanks!

demiankatz: Okay, all done -- since it's a trivial (comments-only) change, I went ahead and committed directly here: https://github.com/vufind-org/vufind/commit/1751f32411eeb6265eded161bdd31c7a581c85e4

EreMaijala:
To make things easier to manage, I'll close this one and post separate PR's for start date + validation and hold modification when #1960 and #1961 are done.

PR #1957, opened 2021-05-17
Fix bug in loading custom notification schedule config.

demiankatz:
The processing of the scheduled_search_frequencies setting does not appear to be working correctly in all situations, which sometimes results in a failure of the command line tool for sending notifications (complaining about an object being passed where an array is expected). This PR makes the processing more explicit and ensures that the setting is always an array.

### DIFF
--- a/module/VuFind/src/VuFind/Search/History.php
+++ b/module/VuFind/src/VuFind/Search/History.php
@@ -139,7 +139,7 @@ class History
         if (!($this->config->Account->schedule_searches ?? false)) {
             return [];
         }
-        // If custom frequences are not provided, return defaults:
+        // If custom frequencies are not provided, return defaults:
         if (!isset($this->config->Account->scheduled_search_frequencies)) {
             return [
                 0 => 'schedule_none', 1 => 'schedule_daily', 7 => 'schedule_weekly'
--- a/module/VuFind/src/VuFind/Search/History.php
+++ b/module/VuFind/src/VuFind/Search/History.php
@@ -135,15 +135,19 @@ class History
      */
     public function getScheduleOptions()
     {
+        // If scheduled searches are disabled, return no options:
         if (!($this->config->Account->schedule_searches ?? false)) {
             return [];
         }
-        if ($this->config->Account->scheduled_search_frequencies instanceof Config) {
-            return $this->config->Account->scheduled_search_frequencies->toArray();
+        // If custom frequences are not provided, return defaults:
+        if (!isset($this->config->Account->scheduled_search_frequencies)) {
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/Search/History.php
```diff
@@ -136,7 +138,12 @@ public function getScheduleOptions()
         if (!($this->config->Account->schedule_searches ?? false)) {
             return [];
         }
-        return $this->config->Account->scheduled_search_frequencies
-            ?? [0 => 'schedule_none', 1 => 'schedule_daily', 7 => 'schedule_weekly'];
+        if ($this->config->Account->scheduled_search_frequencies instanceof Config) {
```
EreMaijala: Is it ok to check instanceof with a potentially missing element? I suppose this works most of the time as the Account section is usually defined, but if it wasn't, wouldn't this balk? So it would be safer to move this check inside the isset block below?

demiankatz: If the account section isn't set, then the check for scheduled_searches will evaluate to false, and the code will never get this far. I think that makes this safe, though perhaps in need of a clarifying comment. However, your observation also reminds me that our test coverage for this method is not complete, since otherwise this bug would never have gotten through in the first place... thus, I'll expand the tests today to make sure all the edge cases are checked, and make adjustments as needed.

EreMaijala: You're right. There's still a chance that a later change could introduce a bug since it's sort of a side-effect, but I suppose it's fine.

demiankatz:
@EreMaijala, I've added the missing tests and reformatted the code for improved clarity. Mind taking one last look? Thanks!

module/VuFind/src/VuFind/Search/History.php
```diff
@@ -133,10 +135,19 @@ public function getSearchHistory($userId = null)
      */
     public function getScheduleOptions()
     {
+        // If scheduled searches are disabled, return no options:
         if (!($this->config->Account->schedule_searches ?? false)) {
             return [];
         }
-        return $this->config->Account->scheduled_search_frequencies
-            ?? [0 => 'schedule_none', 1 => 'schedule_daily', 7 => 'schedule_weekly'];
+        // If custom frequences are not provided, return defaults:
```
EreMaijala: Typo: frequences

EreMaijala (approved):
Looks good, apart from a small comment typo.

demiankatz:
Thanks for the catch, @EreMaijala! Typo fixed, so merging now!

PR #1958, opened 2021-05-18
Add Solr-based reserves support to Demo driver.

demiankatz:
I added this during the testing of #1952; I believe it should be of general value. You can now use the Demo driver to test either the ILS-based or Solr-based version of course reserves.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Demo.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Demo.php
@@ -124,6 +124,27 @@ class Demo extends AbstractBase
     protected $failureProbabilities = [];
 
     /**
+     * Courses for use in course reserves.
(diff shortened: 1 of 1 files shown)


demiankatz:
@EreMaijala, I think this is pretty trivial and low-risk; it just adds some more fields to the simulated reserves data so that you can use the Demo driver with the util/indexReserves.php command-line utility in order to build a reserves index. The existing non-Solr-based functionality should continue to work exactly the same as before.

EreMaijala (approved):
@demiankatz Looks good to me. And definitely low risk yet very useful. I've found the Demo driver invaluable when testing VuFind functionality, so the more comprehensive it is, the better.

demiankatz:
Thanks, @EreMaijala!

PR #1959, opened 2021-05-18
Use window.location.href as fallback for form action.

demiankatz:
- Lightbox didn't work on forms that don't have an action attribute. _currentUrl is not yet set at the moment the initial request is made, so we need to use the current page URL like the form would have used if submitted normally.
- Extracted from #1956.

TODO
- [x] Run full test suite.


crhallberg (approved):
This makes sense to me.

demiankatz:
All tests are passing, @crhallberg gives the thumbs up, and it makes sense to me, so I'm merging now!

PR #1960, opened 2021-05-19
Add patron to getCancelHoldDetails. Use in MultiBackend.

EreMaijala:
This unifies the getCancel*Details methods and makes things easier with the MultiBackend driver.
Also fixes issues with Demo+MultiBackend where the record id may be anything and not related to the patron.

I encountered this while working on #1956.

TODO
- [x] Run full test suite
- [x] Document possible BC breaks
- [x] Update ILS driver wiki page

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Demo.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Demo.php
@@ -1599,13 +1599,17 @@ class Demo extends AbstractBase
      * separated by a pipe, which is then submitted as form data in Hold.php. This
      * value is then extracted by the CancelHolds function.
      *
-     * @param array $holdDetails An array of item data
+     * @param array $hold   An array of hold data
+     * @param array $patron Patron information from patronLogin
      *
      * @return string Data for use in a form field
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function getCancelHoldDetails($holdDetails)
+    public function getCancelHoldDetails($hold, $patron)
     {
-        return $holdDetails['reqnum'];
+        return empty($hold['available']) && empty($hold['in_transit'])
+            ? $hold['reqnum'] : '';
     }
 
     /**
@@ -1663,13 +1667,16 @@ class Demo extends AbstractBase
      * separated by a pipe, which is then submitted as form data in Hold.php. This
      * value is then extracted by the CancelHolds function.
      *
-     * @param array $details An array of item data
+     * @param array $request An array of request data
+     * @param array $patron  Patron information from patronLogin
      *
      * @return string Data for use in a form field
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function getCancelStorageRetrievalRequestDetails($details)
+    public function getCancelStorageRetrievalRequestDetails($request, $patron)
     {
-        return $details['reqnum'];
+        return $request['reqnum'];
     }
 
     /**
@@ -2259,13 +2266,16 @@ class Demo extends AbstractBase
     /**
      * Get Cancel ILL Request Details
      *
-     * @param array $details An array of item data
+     * @param array $request An array of request data
+     * @param array $patron  Patron information from patronLogin
      *
      * @return string Data for use in a form field
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function getCancelILLRequestDetails($details)
+    public function getCancelILLRequestDetails($request, $patron)
     {
-        return $details['reqnum'];
+        return $request['reqnum'];
     }
 
     /**
--- a/module/VuFind/src/VuFind/ILS/Driver/Demo.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Demo.php
@@ -1606,7 +1606,7 @@ class Demo extends AbstractBase
      *
      * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function getCancelHoldDetails($hold, $patron)
+    public function getCancelHoldDetails($hold, $patron = [])
     {
         return empty($hold['available']) && empty($hold['in_transit'])
             ? $hold['reqnum'] : '';
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -1099,7 +1099,7 @@ class Folio extends AbstractAPI implements
      *
      * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function getCancelHoldDetails($hold, $patron)
+    public function getCancelHoldDetails($hold, $patron = [])
     {
         return $hold['reqnum'];
     }
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -1092,12 +1092,14 @@ class Folio extends AbstractAPI implements
     /**
      * Get FOLIO hold IDs for use in cancelHolds.
      *
-     * @param array $hold An single request
-     * array from getMyHolds
+     * @param array $hold   A single hold array from getMyHolds
+     * @param array $patron Patron information from patronLogin
      *
      * @return string request ID for this request
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function getCancelHoldDetails($hold)
+    public function getCancelHoldDetails($hold, $patron)
     {
         return $hold['reqnum'];
     }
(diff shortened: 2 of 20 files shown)


demiankatz:
Thanks, @EreMaijala, this makes sense to me. One question before I take a really close look at everything: is there a reason you chose not to make the new parameter optional? Many of the other ILS driver methods default the patron array to false or null (an inconsistency we can discuss another day); I'm just wondering if it would make sense to do something similar here for potentially greater backward-compatibility (though the changed method signature is a BC break no matter what we do).

EreMaijala:
@demiankatz It's not really optional for MultiBackend with the included change, that's why I added it as non-optional. I can still add a fallback to MultiBackend, just a sec.

EreMaijala:
@demiankatz Done.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1092,12 +1092,14 @@ public function placeHold($holdDetails)
...
-     * array from getMyHolds
+     * @param array $hold   A single hold array from getMyHolds
+     * @param array $patron Patron information from patronLogin
      *
      * @return string request ID for this request
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function getCancelHoldDetails($hold)
+    public function getCancelHoldDetails($hold, $patron)
```
demiankatz: I think a default parameter was missed on this one.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -1663,13 +1667,16 @@ public function cancelStorageRetrievalRequests($cancelDetails)
...
-     * @param array $details An array of item data
+     * @param array $request An array of request data
+     * @param array $patron  Patron information from patronLogin
      *
      * @return string Data for use in a form field
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function getCancelStorageRetrievalRequestDetails($details)
+    public function getCancelStorageRetrievalRequestDetails($request, $patron)
```
demiankatz, line 1677: Should getCancelStorageRetrievalRequestDetails have a default value for $patron for consistency?

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
```diff
@@ -1235,21 +1242,23 @@ public function cancelILLRequests($cancelDetails)
...
-    public function getCancelILLRequestDetails($details)
+    public function getCancelILLRequestDetails($request, $patron)
     {
-        $source = $this->getSource($details['id'] ?? $details['item_id'] ?? '');
+        $source = $this->getSource($patron['cat_username']);
         $driver = $this->getDriver($source);
         if ($driver
             && $this->methodSupported(
-                $driver, 'getCancelILLRequestDetails', compact('details')
+                $driver, 'getCancelILLRequestDetails', compact('request', 'patron')
```
demiankatz, line 1256: I'm reasonably sure that methodSupported() doesn't care that the first key of the compacted array has changed from 'details' to 'request' but just double-checking since it's a subtle behavior change.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -2259,13 +2266,16 @@ public function cancelILLRequests($cancelDetails)
...
-     * @param array $details An array of item data
+     * @param array $request An array of request data
+     * @param array $patron  Patron information from patronLogin
      *
      * @return string Data for use in a form field
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function getCancelILLRequestDetails($details)
+    public function getCancelILLRequestDetails($request, $patron)
```
demiankatz, line 2276: Same question here: should we add a default $patron value to all of these too?

demiankatz (commented):
Just a few more comments and notes.

I also wonder if we should consider defaulting $patron to `null` rather than `[]` since `patronLogin` can return null, and this will ensure that the parameter is nullable. (Otherwise we'll want to do an `?array` typehint to ensure that this doesn't get broken by future work related to static analysis).

As always, sorry to be a pain about these details. Happy to help with editing files if you agree with my intent but don't want to do all the work. :-)

module/VuFind/src/VuFind/ILS/Driver/Folio.php
EreMaijala: Oops. Fixed.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: I don't think so. It has always had the patron parameter, and it doesn't make sense to call it unless there are requests to display and thus a known patron.

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
EreMaijala: It doesn't care, but I think is actually completely borked and should not be using compact at all! The parameter is passed as the second argument to supportsMethod, which returns "True if the method can be called with the given parameters, false otherwise", so I believe this should be `[$request, $patron]` instead.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: Same answer.. :)

EreMaijala:
@demiankatz I don't think defaulting $patron to null makes sense since this will never be called unless patronLogin is successful. I wouldn't add a default parameter at all, but it makes sense for backwards-compatibility since getCancelHoldDetails hasn't always had the parameter.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
demiankatz: Thanks!

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
demiankatz: Yes, that's my impression too. We can leave it alone here since it's really a separate concern, but maybe we need a separate PR to eliminate the unnecessary compacts in the MultiBackend driver. Would you like me to put that together?

module/VuFind/src/VuFind/ILS/Driver/Demo.php
demiankatz: I don't think it's true that this has always had the patron parameter, because you're adding the patron parameter here. However, I agree with your point that it isn't necessary to make $patron optional in this context. :-)

demiankatz: Actually, though, it does seem to be a problem that there are several instances of getCancelStorageRetrievalRequestDetails that have not had the patron parameter added yet in this PR (KohaRest, for example).

module/VuFind/src/VuFind/ILS/Driver/Demo.php
demiankatz: ...and same comment: there are some drivers that have not had the $patron parameter added in this PR (possibly just PAIA and VoyagerRestful in this case).

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: Sorry, I wasn't quite clear: The parameter has always been passed to getCancelStorageRetrievalRequestDetails even if it hasn't been declared for all the instances. See https://github.com/vufind-org/vufind/commit/3bec0eb7d3cb9971fe446cf00f14c7f32348d09a#diff-c2869125639e00db24fef78a00af0fa67f1d25a7770931b86c0118902437ef77R85

demiankatz:
@EreMaijala, I agree that the null default doesn't make sense since under normal circumstances the patron value should always be an array -- that might also be an argument for changing the various null and false defaults on other methods to an empty array as well, though that's a project for another day.

There does seem to be some more work to do here with making storage retrieval and ILL consistent with holds, though -- but I think once that's done, I can finalize testing and get this merged.

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
EreMaijala: I think it's my bad, so I'll volunteer to clean it up. :)

demiankatz: Thanks! As we discussed on Slack, that can be addressed in a separate PR since the compact problem is largely independent of what's being worked on here.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
demiankatz: Ahh, I see, that makes sense. I still think it's worth making these method signatures consistent across drivers, in case we implement feature interfaces in the future (and just to prevent potential confusion). I've gone ahead and added $patron to the drivers that were missing it. (See commit f6d1968707c94103d8c8ce61b78f26a282cf2151).

module/VuFind/src/VuFind/ILS/Driver/Demo.php
demiankatz: ...and same action: commit f6d1968707c94103d8c8ce61b78f26a282cf2151

demiankatz:
I added some additional tests to the Mink ILS test suite to ensure that all types of canceling are tested properly, and everything is passing -- merging now.

PR #1961, opened 2021-05-19
Add HoldsController for holds-related functionality.

EreMaijala:
Moves holds list from MyResearch/Holds to Holds/List. Keeps old route for compatibility.

Spun from #1956.

TODO
- [x] Run full test suite
- [x] Document BC breaks: removal of myresearch-holds route, refactoring of MyResearchController getDriversForILSRecords and collectRequestAccountStats methods.


module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -0,0 +1,312 @@
...
+            ->collectRequestStats($view->recordList);
+        return $view;
+    }
+
+    /**
+     * Edit holds
+     *
+     * @return mixed
+     */
+    public function editAction()
```
demiankatz: Should we omit editAction here since it is not being used yet?

demiankatz (commented):
This refactoring makes sense to me. I'll review and test more thoroughly when time permits (and after you've had a chance to answer my question below). I've added a TODO item about documenting the BC breaks.

module/VuFind/src/VuFind/Controller/HoldsController.php
EreMaijala: Oops, yes.

EreMaijala: Now removed.

demiankatz: Thanks!

module/VuFind/config/module.config.php
```diff
@@ -78,6 +78,16 @@
                     ]
                 ]
             ],
+            'legacy-holds' => [
+                'type' => 'Laminas\Router\Http\Literal',
+                'options' => [
+                    'route'    => '/MyResearch/Holds',
+                    'defaults' => [
+                        'controller' => 'Holds',
```
demiankatz: How would you feel about continuing to point this at the MyResearchController's holdsAction, but instead changing that action to force a redirect to the new holds-list route? That way, people will get sent to the new URL instead of having two different URLs that do the same thing.

demiankatz: (My thought is that eventually, after we complete refactoring, the MyResearchController will be nothing but a collection of redirects... and then we can deprecate and remove the whole thing).

module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -0,0 +1,160 @@
...
+     * Constructor
+     *
+     * @param ServiceLocatorInterface    $sm  Service locator
+     * @param \Laminas\Session\Container $urc A session container for hold update
+     * results
+     */
+    public function __construct(ServiceLocatorInterface $sm,
+        \Laminas\Session\Container $urc
+    ) {
+        $this->serviceLocator = $sm;
```
demiankatz: I'd be inclined to change this to `parent::__construct($sm);` -- since we don't use the serviceLocator directly in this class at all, I think it's more clear to show that it is a dependency of the parent class... and when we eventually refactor that, the relationship will be more clear.

module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -0,0 +1,160 @@
...
+        $cancelStatus = $catalog->checkFunction('cancelHolds', compact('patron'));
+        $view = $this->createViewModel();
+        $view->cancelResults = $cancelStatus
+            ? $this->holds()->cancelHolds($catalog, $patron) : [];
+        // If we need to confirm
+        if (!is_array($view->cancelResults)) {
+            return $view->cancelResults;
+        }
+
+        // Process any update request results stored in the session:
```
demiankatz: This block of code isn't used yet; should we pull it out here and add it back in #1956?

module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -0,0 +1,160 @@
...
+                    return $this->getRefreshResponse();
+                }
+            } else {
+                return $this->forwardTo('Holds', 'Edit');
+            }
+        }
+
+        // By default, assume we will not need to display a cancel or update form:
+        $view->cancelForm = false;
+        $view->updateForm = false;
```
demiankatz: Another line we don't need before #1956.

module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -0,0 +1,160 @@
...
+ * @link     https://vufind.org Main Site
+ */
+class HoldsController extends AbstractBase
+{
+    /**
+     * Hold update results container
+     *
+     * @var \Laminas\Session\Container
+     */
+    protected $updateResultsContainer;
```
demiankatz: We also don't need the updateResultsContainer yet, though if it's less work to leave it here with the custom factory, it's not hurting anything.

module/VuFind/src/VuFind/Controller/Plugin/IlsRecords.php
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Action helper to perform ILS record related actions
+ *
+ * @category VuFind
+ * @package  Controller_Plugins
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Page
+ */
+class IlsRecords extends AbstractRequestBase
```
demiankatz: Why does this extend AbstractRequestBase? That seems to add a bunch of dependencies that aren't actually used by either of the methods in this helper. Is there something I'm missing that prevents you from just directly extending \Laminas\Mvc\Controller\Plugin\AbstractBase instead?

demiankatz (changes requested):
I ran the test suite and everything is passing, so I took a closer look at the code and had a few more comments and suggestions. Apologies for asking you to rip out things that you just put in, but I'm worried this will cause confusion if we commit too much "future stuff" before we're ready for it.

Do you think it would be worthwhile to follow this pattern for the other request-related actions (storage retrieval, etc.)? I can probably find time to do that once this is finalized.

module/VuFind/config/module.config.php
EreMaijala: I'm ok with that if it's easier to deprecate and drop in future. I just tried to remove one hop from the execution, but your point is valid.

EreMaijala:
@demiankatz Thanks for the review comments. I certainly did this too hastily, but the issues should now be sorted.

I'm itching to rewrite the cancel form handling to target an edit action, but I'm trying to keep this simple for now.

demiankatz:
@EreMaijala, I removed the HoldsControllerFactory because we don't need it yet. I'd suggest that when you add it back in, it might be simpler to revise VuFind\Controller\AbstractBase so that it no longer throws an exception for incoming $options, but instead expands them as additional constructor parameters to the controller being constructed -- e.g.:

<pre>
    public function __invoke(ContainerInterface $container, $requestedName,
        array $options = null
    ) {
        return $this->applyPermissions($container, new $requestedName($container, ...($options ?? [])));
    }
</pre>

...and then the HoldsControllerFactory can just construct the session container and pass it through the $options array to parent::__invoke(). This allows you to avoid duplicating the permission-related logic, and will make future refactoring easier.

EreMaijala:
@demiankatz Thanks, makes sense.

demiankatz:
All tests are passing, and I've added notes to the changelog. Merging this now, and may try to do more MyResearchController refactoring when time permits!

EreMaijala:
@demiankatz Thanks! Now that the groundlaying work has been done, I believe further refactoring will be more straightforward.

PR #1963, opened 2021-05-20
Add support for hold start date. Validate dates.

EreMaijala:
Dates are validated before sending them to the ILS drivers. The drivers receive the dates as unix timestamps in addition to the existing display format.

This pull request only simplifies the date handling for the Demo driver. Other drivers continue to work the same, but at least some benefit from the early validation of dates.


module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
```diff
@@ -65,16 +67,27 @@ abstract class AbstractRequestBase extends AbstractPlugin
...
+     */
+    protected $dateConverter;
+
     /**
      * Constructor
      *
      * @param HMAC           $hmac           HMAC generator
      * @param SessionManager $sessionManager Session manager
+     * @param DateConverter  $dateConverter  Date converter
+     * @
```
demiankatz: There's a stray `@` here.

module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
```diff
@@ -303,6 +316,76 @@ public function getDefaultRequiredDate($checkHolds, $catalog = null,
...
+     * problems.
+     *
+     * @param string $startDate       User-specified start date
+     * @param string $requiredBy      User-specified required-by date
+     * @param array  $extraHoldFields Hold form fields enabled by
+     * configuration/driver
+     *
+     * @return array
+     */
+    public function validateDates($startDate, $requiredBy, $extraHoldFields)
```
demiankatz: Rather than having two parameters (`$startDate` and `$requiredBy`) here, what if you took an associative array (`['startDate' => ..., 'requiredBy' => ...]`), and then appended "TS" onto the keys for the output? This might be more flexible if different types of requests have different types of dates associated with them.

demiankatz: You might also rename `$extraHoldFields` to something like `$enabledFormFields` to make this more generic for use with other types of requests in the future.

module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
```diff
@@ -303,6 +316,76 @@ public function getDefaultRequiredDate($checkHolds, $catalog = null,
...
+                $errors[] = 'hold_start_date_invalid';
+            }
+        }
+
+        if (in_array('requiredByDate', $extraHoldFields)) {
+            try {
+                if ($requiredBy) {
+                    $result['requiredByTS'] = $this->dateConverter
+                        ->convertFromDisplayDate('U', $requiredBy, true)
+                        ->setTime(23, 59, 59)
```
demiankatz: I guess the need to set a time adds an inconvenient wrinkle to making this more generic, but maybe we could pass timestamp specific flags somehow, either through the input array or a separate options array, to indicate which timestamps need special treatment.

module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
```diff
@@ -303,6 +316,76 @@ public function getDefaultRequiredDate($checkHolds, $catalog = null,
...
+                        ->getTimestamp();
+                } else {
+                    $result['requiredByTS'] = 0;
+                }
+                if ($result['requiredByTS'] < time()) {
+                    $errors[] = 'hold_required_by_date_invalid';
```
demiankatz: And these error messages also would need to be passed in somehow -- which I think further suggests the idea of a more complex input array, like `['requiredBy' => ['value' => ...,, 'error_msg' => ..., 'options' => ...]`. Maybe that's over complicated, but I think it could pay off in the long run.

module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
```diff
@@ -303,6 +316,76 @@ public function getDefaultRequiredDate($checkHolds, $catalog = null,
...
+                }
+            } catch (DateException $e) {
+                $errors[] = 'hold_required_by_date_invalid';
+            }
+        }
+
+        if (!$errors
+            && in_array('startDate', $extraHoldFields)
+            && in_array('requiredByDate', $extraHoldFields)
+            && $result['startDateTS'] > $result['requiredByTS']
```
demiankatz: The plot keeps getting thicker as I read my way down. Maybe an optional validation callback function?

demiankatz: ...and of course, feel free to tell me this isn't worth the effort. But if that's the case, I think this function belongs in the Holds helper rather than AbstractRequestBase, because as written it seems very holds-specific.

demiankatz (commented):
Thanks, @EreMaijala, this looks good to me, but I did have a few suggestions. (Pardon the developing nature of some of these comments -- I wrote as I read, which led to some evolution of ideas along the way). :-)

module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
EreMaijala: Fixed.

module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
EreMaijala: Parameter renamed.

module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
EreMaijala: This isn't worth the effort. :)

I don't think it's worth it to make the input an array with rules. I think it would just move the logic from the code in the plugin to the array in the caller and would make things more complicated without providing obvious benefits. If we end up needing a more generic validation of a group of dates, we could of course make validateDates a wrapper for a rule-based validation mechanism, but I'd rather leave it to a later point if the need arises. Hold modification will bring the need to validate a "frozen until" date, but I'd still rather add a simple method for that.

I have moved the method to Holds plugin to alleviate the issue of it not being generic. As you rightly pointed out, it's pretty hold-specific anyway. We still have other hold-specific methods (at least the request group -related ones) in AbstractRequestBase, but that's another issue.

demiankatz: Thanks, @EreMaijala! It's still on the to-do list to get rid of controller plugins entirely ([VUFIND-1285](https://vufind.org/jira/browse/VUFIND-1285)) since there have been threats of their removal in a future Laminas release... so we can address other refactoring of inappropriately-located methods as part of that project, if and when we get there. (I think this is all really related to the overall goal of making controllers more focused and less "magical").

For now, I'm happy with the relocation of the method. Just one last question before I do the final testing (and hopefully merging) of this work: we've now added the DateConverter as a dependency of all of the request processing helpers, even though Holds is the only one that uses it. Are you planning to apply similar logic to the other types of requests? If so, I'm fine with leaving it where it is so it's ready for future work... but if it's only ever going to be used by Holds, it would probably be cleaner to move it there and give Holds a separate factory.

EreMaijala: I think it would make sense to add similar checking to other requests as well, so if it's ok to leave it there for now, I'll try to actually make use of it later on.

demiankatz: Great, thanks! Not sure if I'll get this tested before next week, but hopefully it won't take too long once I have a free moment.

demiankatz: @EreMaijala, I ran the full test suite today, and two integration tests are failing:

<pre>
There were 2 failures:

1) VuFindTest\Mink\IlsActionsTest::testPlaceHold
Selector not found: .modal-body a

/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:251
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:143
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:260
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98

2) VuFindTest\Mink\IlsActionsTest::testHoldsAll
Selector not found: .modal-body a

/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:251
/home/dkatz/vufind3/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:658
/home/dkatz/vufind3/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:98
</pre>

I'm quite short on time today, so I haven't been able to investigate why. If you have a theory, please let me know; otherwise, I'll dig into this further when I have a bit more time.

EreMaijala: I suppose placing hold fails, but I'm not sure why. The config change means there'll be the start date field as well, but it should be prepopulated. Other option is that date validation fails, but if the script doesn't change anything, that shouldn't be the case either.

EreMaijala: I found the issues, will post fixes.

EreMaijala: The issues should be fixed now. My original intention was to modify VuFindDate to make it support returning a DateTime object, but decided to work around that need for now. Also the start date and required date validation didn't consider today as valid...

demiankatz:
@EreMaijala, I found another minor issue: if I place a hold with a start date in the future, I get this:

image

I think there's at least one missing translation in this PR.

EreMaijala:
@demiankatz Oops. I lost track of the translations when splitting the original PR. Should be fixed now.

demiankatz:
Thanks, @EreMaijala -- I think this is nearly ready to go, but one question (which may be partially a pre-existing problem): when I try to cancel a frozen hold and the operation fails, the error message appears above the message about the freeze date, which looks strange. Do you think we should move the error messages to the bottom of the hold blocks, rather than having them in the middle as we currently do?

EreMaijala:
@demiankatz I moved the error message to the bottom. It was a pre-existing problem, but since the information below the failure message would typically not be available for holds that can be cancelled (or, canceling usually works fine), it didn't come up.

demiankatz:
Thanks, @EreMaijala -- everything looks good to me now; merging!

PR #1964, opened 2021-05-20
KohaRest: Trim fine description before comparing with translation.

EreMaijala:
The fine description may include trailing whitespace, which would make the comparison fail.


demiankatz:
Is there any chance there might be trailing whitespace on $type as well? Should we trim that as well to be safe?

EreMaijala:
That doesn't seem to be the case, but I added a trim as it won't hurt either.

demiankatz:
Thanks! Merging now...

PR #1965, opened 2021-05-24
Make MultiBackend driver work better with the Demo driver.

EreMaijala:
Since the Demo driver can take any record source e.g. for placing holds, MultiBackend driver shouldn't restrict it. This improves testing capability with MultiBackend+Demo. Normal drivers should notbe affected.

Tests with getHolding and checkRequestIsValid are extended to verify the functionality.


demiankatz:
This makes sense to me. If we think there will ever be another multi-source-capable driver we should probably define an interface or other means of identifying the capability, rather than explicitly looking for the Demo driver... but since this is a very specific edge case, I think it's okay as-is for now.

PR #1966, opened 2021-05-24
Display an error flash message if canceling of any of the holds fails.

EreMaijala:


demiankatz:
This looks good to me. I also ran the full test suite to be sure that no existing tests were impacted by the addition of the new flash message, and all is well. (I suppose we could add a test including a forced failure to cover the addition of the message, but at the moment I do not have time, and I'm sure there are other, higher-priority testing tasks at the moment in any case). Thanks for adding this!

PR #1967, opened 2021-05-24
Use Command Pattern in search system

aleksip:
Here is an implementation of the command pattern as described in [VUFIND-924](https://vufind.org/jira/browse/VUFIND-924) and discussed in #1886.

I was initially a bit reserved about the idea as I thought it might add unnecessary overhead and make things more complex. But now I have grown to really like it, and think that the simplest thing to do could actually be sending commands directly to the search service.

The existing code was rather easy to refactor into commands. Using commands also makes the events system simpler, since events can just have the command as a parameter instead of several method specific parameters.

The current event system expects the target of the `post` event to be the return value of the requested operation. However in the command pattern the result is contained in the command object, so again it seems simpler to have the command as the target for the `post` event as well.

Making changes to the event system also provided an opportunity to remove the backend instance from event parameters. It can of course be left as a parameter, but as direct access to backend instances bypasses the search service and the events system, removing it is probably a good thing.

Since modifying the events system is a breaking change, I have also included a protected `legacyInvoke()` method, which uses commands but leaves the event system unchanged. Using `legacyInvoke()` in the current service methods and deprecating them would provide a backwards compatible way to move to using the new command-based service and events.

If the decision is to avoid direct command calls, `legacyInvoke()` can be refactored to be the only way to invoke commands. However, in addition to simpifying the event system, exposing commands to callers would avoid the overhead of having to maintain helper code.

I found adding copyright statements and author annotations to the new command classes a bit problematic. I don't want things to look like everything is just my work, but figuring out the proper copyrights and authors is no easy task in a refactoring like this.

This would be a case for another issue, but I wonder if the boilerplate license and copyright blocks could be simplified by just referring to a file or files in the root folder, and authors perhaps just dropped altogether, as they can be determined from version control.


module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
```diff
@@ -0,0 +1,202 @@
...
+        }
+        if (!($backendInstance instanceof $this->interface)
+            || !method_exists($this->interface, $this->method)
+        ) {
+            throw new BackendException(
+                "$this->backend does not support $this->method()"
+            );
+        }
+        $callArgs = $this->args;
+        $callArgs[] = $this->params;
```
demiankatz: Is it safe to assume the last argument will always be a ParamBag? Maybe an $addParamsToArgs boolean property, defaulting to true, would balance flexibility and convenience, in case of a future divergence from this pattern.

demiankatz (commented):
@aleksip, thank you for this; as you say, the code refactored very neatly!

As far as whether to use Commands exclusively or not, my suggestion is to transition toward a Commands-only approach, but not to fully commit to it yet. This gives us time to evaluate whether or not there are any significant disadvantages to that approach, and it also makes it easier to minimize backward compatibility breaks.

I say this mainly because, if we refactor all of VuFind's core code to use Commands exclusively right away, then we will break everybody's custom listeners, since they will not yet expect Commands and will have to be rewritten. Instead, I would suggest that we rewrite all of VuFind's internal listeners to operate on Command objects instead of legacy parameters, but keep VuFind's internal calls using the wrappers around legacyInvoke. This way, we can announce with 8.0 that a listener change is underway, and people will have time to rewrite their listeners to use Commands without it being an emergency. Then, in 9.0, if we still like the Commands-only approach, we can rip out all of the legacyInvoke stuff, and people's listeners won't break if they upgraded during the 8.x line. Also, the experience of rewriting all of the internal listeners to use Commands may reveal some issues or ideas that we haven't thought of yet.

If you'd like any help with the listener refactoring, let me know! I probably won't be able to do much for the next couple of weeks (due to multiple conference-related commitments), but I also don't want to force you to do all the work if you don't want to. :-)

Regarding your comments about copyright statements and author annotations, I certainly wouldn't mind simplifying those things if we could do so without creating any legal issues -- but I'm not sure exactly what requirements exist as far as what must be included in each file, and I haven't taken the time to research it. Regarding the `@author` tag specifically, there are two reasons I would prefer not to simply rip it out: first of all, it's required by the code standards enforced by PHP-CodeSniffer, and while we certainly could customize the standard, I prefer to keep things as close to defaults as reasonably possible. Secondly, and more importantly, some of the authors acknowledged in the comments did foundational work on VuFind 1.0, but that work lives in an old Subversion repository, so their contributions are not apparent in the current Git repository that was established for VuFind 2.0 and above. While I'm not too particular about the author attribution in individual files (and I certainly don't care whether or not my name is on every single thing), I do try to give credit to the people who established files or did major work on them when practical... so in this instance, you'd certainly be justified to add your name to everything, and you could copy over names from the main backend if you felt it would clarify that the work is not entirely your own. I definitely wouldn't worry about making sure everything is comprehensive and perfect as long as no credits are getting completely dropped along the way.

aleksip:
@demiankatz Your suggestion on how to proceed sounds good to me. Should the listener refactoring be done in this PR or a separate one?

I have now added the `addParamsToArgs` boolean property. I have also added also some copyright and author information taken from the respective interfaces, based on the assumption that whoever wrote the interface most likely wrote at least the initial `Service` method implementation as well.

demiankatz:
Thanks, @aleksip, this all looks good to me. I would be inclined to do the listener refactoring here, since that way if it inspires any changes/adjustments/bug fixes, we haven't already committed immature code to the dev branch... but if you feel that it would be cleaner to merge this first, I'm not completely opposed to the idea. Whether or not we split this in two or do it all here, I'll want to do a bit more testing just to be 100% sure everything is still working correctly... so if you prefer to merge this as-is, I'll try to do that testing sooner rather than later; otherwise, I'll wait until the listener work is done and test it all at once.

aleksip:
@demiankatz I agree that it is good to do the listener changes in this PR. I will start working on the refactoring.

aleksip:
@demiankatz I have now refactored listeners and tests to use command-based events. As you can see the refactoring was quite straightforward and there was no need for significant modifications. And indeed I found a few things to improve in the search backend command pattern implementation as well.

demiankatz:
@aleksip, I agree, this looks like really clean and clear refactoring, and I think it without exception improves the quality and comprehensibility of the code. I'm going to mention this work on Tuesday's Community Call so that others have a chance to review it. I also want to do a bit more hands-on testing myself, just to be sure we didn't miss anything (though I don't think we did). I think we should be able to merge this in another couple of weeks, barring unexpected complications. Can you think of any other reasons to wait?

Thanks again for this excellent work!

aleksip:
@demiankatz Can't think of other reasons to wait than thorough testing and asking for comments, which I agree are good to do.

demiankatz:
Thanks, @aleksip! The full test suite is passing for me, and I can't think of anything else to break, so I think we're in good shape. I'll just allow a little more time for feedback from others.

demiankatz:
There have been no objections raised, so I have just merged this. Thanks again for the work! I have also opened [VUFIND-1482](https://vufind.org/jira/browse/VUFIND-1482) to track the need to utilize the search service in places where we are currently bypassing it. I will try to do some work on that when time permits, but I have quite a few projects going on right now, so it may take a little while. If you have time/interest in doing any work on it, please feel free to start the ball rolling, and I will assist as I can. If you do not have time either, it can wait until I get there.

PR #1968, opened 2021-05-25
Patch chrome-mink-driver 2.8.0 on install for PHP 8 compatibility.

EreMaijala:


demiankatz:
Makes sense to me, and doesn't break anything on my end. Hopefully this will get fixed upstream soon!

PR #1971, opened 2021-06-02
Better validation of limit parameter.

demiankatz:
Entering limit parameters with non-numeric portions can result in PHP notices under some circumstances; this change forces limit values to be integers for improved validation. This PR also adds test coverage of the limit functionality more broadly.


EreMaijala (approved):

EreMaijala:
@demiankatz Makes sense to me.

PR #1972, opened 2021-06-02
Improvements to AlphabrowseController

demiankatz:
- Validate source parameter before sending to Solr
- Refactor code for improved efficiency, readability and extensibility

### DIFF
--- a/module/VuFind/src/VuFind/Controller/AlphabrowseController.php
+++ b/module/VuFind/src/VuFind/Controller/AlphabrowseController.php
@@ -109,10 +109,10 @@ class AlphabrowseController extends AbstractBase
         // Load config parameters
         $types = $this->getTypes($config);
         $extras = $this->getExtras($config);
-        $rows_before = is_numeric($config->AlphaBrowse->rows_before ?? null)
+        $rows_before = ctype_digit($config->AlphaBrowse->rows_before ?? '-')
             ? (int)$config->AlphaBrowse->rows_before : 0;
         $highlighting = $config->AlphaBrowse->highlighting ?? false;
-        $limit  = is_numeric($config->AlphaBrowse->page_size ?? null)
+        $limit  = ctype_digit($config->AlphaBrowse->page_size ?? '-')
             ? (int)$config->AlphaBrowse->page_size : 20;
 
         // Process incoming parameters:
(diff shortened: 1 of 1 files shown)


demiankatz:
@EreMaijala, somebody is throwing malformed requests at my VuFind instance, which is revealing a couple of places where we could stand to improve our input validation. (That was the inspiration for #1971 as well as this one). It has always bothered me how long the code is in this class, so while I was in the area adding validation, I also did a little bit of refactoring to improve readability. There's still a lot more that could be done, but this is a small start.

module/VuFind/src/VuFind/Controller/AlphabrowseController.php
```diff
@@ -53,50 +106,15 @@ public function homeAction()
...
-            && is_numeric($config->AlphaBrowse->rows_before)
+        // Load config parameters
+        $types = $this->getTypes($config);
+        $extras = $this->getExtras($config);
+        $rows_before = is_numeric($config->AlphaBrowse->rows_before ?? null)
             ? (int)$config->AlphaBrowse->rows_before : 0;
         $highlighting = $config->AlphaBrowse->highlighting ?? false;
-        $limit  = isset($config->AlphaBrowse->page_size)
-            && is_numeric($config->AlphaBrowse->page_size)
+        $limit  = is_numeric($config->AlphaBrowse->page_size ?? null)
```
EreMaijala: Since `is_numeric` allows more than just digits, I'd rather use something like `ctype_digit($config->AlphaBrowse->page_size ?? '-')`, but since the value comes from config, it's not critical.

EreMaijala (approved):

EreMaijala:
@demiankatz Nice, makes the code so much cleaner!

module/VuFind/src/VuFind/Controller/AlphabrowseController.php
demiankatz: Thanks for the suggestion; I've adjusted this accordingly.

EreMaijala: Do you know if $config always returns the value as a string? If not, it would be safest to cast to string before checking with ctype_digit, as it would return false for an integer.

demiankatz: I addressed this in #1977.

PR #1973, opened 2021-06-02
Refactor OAuth2 authorization and use it in XCNCIP2 ILS driver

xmorave2:


demiankatz:
Thanks, @xmorave2! Since this touches on the KohaRest code, I'm going to wait for @EreMaijala's feedback before taking a close look at this myself... but I'm happy to pitch in if and when my input is needed!

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
```diff
@@ -44,6 +46,7 @@ class XCNCIP2 extends AbstractBase implements \VuFindHttp\HttpServiceAwareInterf
 {
     use \VuFindHttp\HttpServiceAwareTrait;
     use \VuFind\Log\LoggerAwareTrait;
+    use \VuFind\ILS\Driver\CacheTrait;
```
EreMaijala, line 48: You should add a getCacheKey method that makes sure any cached values are unique for the driver instance.

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
```diff
@@ -173,6 +193,11 @@ public function init()
         }
         $this->disableRenewals
             = $this->config['Catalog']['disableRenewals'] ?? false;
+
+        $this->useOAuth2 = ($this->config['tokenEndpoint'] ?? false)
+            && ($this->config['clientId'] ?? false)
+            && ($this->config['clientSecret'] ?? false);
+        $this->config['tokenBasicAuth'] = $this->config['tokenBasicAuth'] ?? false;
```
EreMaijala: Perhaps make `$tokenBasicAuth` a member variable like `$useOAuth2`, so you don't need to modify $config.

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
```diff
@@ -272,19 +297,36 @@ public function loadPickUpLocationsFromNcip()
...
+            $result = $client->send();
         } catch (\Exception $e) {
             throw new ILSException($e->getMessage());
         }
 
+        // If we get a 401, we need to renew the access token and try again
+        if ($this->useOAuth2 && $result->getStatusCode() == 401) {
+            $client->getRequest()->getHeaders()
+                ->addHeaderLine('Authorization', $this->getOAuth2Token(true));
+
```
EreMaijala: Extra line.

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
```diff
@@ -272,19 +297,36 @@ public function loadPickUpLocationsFromNcip()
...
+            try {
+                $result = $client->send();
+            } catch (\Exception $e) {
+                throw new ILSException($e->getMessage());
```
EreMaijala: I think it might be a good idea to log the actual errors and return something that doesn't potentially reveal too much information the ILSException.

module/VuFind/src/VuFind/ILS/OAuth2Service.php
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * @category VuFind
+ * @package  VuFind\ILS
+ * @author   Josef Moravec <moravec@mzk.cz>
+ * @license  https://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:plugins:ils_drivers Wiki
+ */
+namespace VuFind\ILS;
+
+use Laminas\Http\Client\Exception\RuntimeException as HttpException;
```
EreMaijala: Consider adding a new exception class in the \VuFind\Exception namespace.

EreMaijala (changes requested):
Tested to work with KohaRest, so functionality-wise it's all good.

Is there a compelling reason for making the OAuth2 authenticator a service? I think it would simplify things if it was a trait that any ILS driver could simply use if needed.

If it's kept as a service, it would be better to make it more generic and live e.g. in \VuFind\Service namespace. In that case I'd also remove logging from it and make it just throw exceptions to allow the caller to handle errors in a way appropriate for the situation.

Alternatively, this might be the time to consider switching to a proper OAuth2 client such as league/oauth2-client. I see that you've already done a lot of work with the refactoring and tests etc, but I still think this is worth considering. I haven't checked if it's easy to plug our own HTTP client to league/oauth2-client, so that might be one obstacle. What do you think?

I'd also welcome @demiankatz to comment especially reagrding the architecture.

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
EreMaijala: To clarify: This wouldn't normally be an issue unless you use the MultiBackend driver to access two or more XCNCIP services, but it makes sense to prepare for that in any case.

demiankatz:
Thanks for the review, @EreMaijala -- I agree with all of these observations.

I do like the idea of using a generic oauth2 client rather than inventing one, if that is feasible without making the work significantly harder. Regarding compatibility between a potential oauth2 client and our HTTP library, it's also worth noting that laminas-http has effectively reached end of life; see: https://github.com/laminas/laminas-http/blob/2.15.x/README.md -- it sounds like this is not an immediate emergency, but sooner or later we're going to need to figure out how to move to something PSR-7 compliant. Perhaps this situation is the motivation to begin thinking about that, or perhaps we can decide to defer use of a library until we are able to reach that point (a decision that we could capture as a JIRA ticket, perhaps).

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
xmorave2: I understand. I added getCacheKey method now

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
xmorave2: done

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
xmorave2: fixed

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
xmorave2: fixed

module/VuFind/src/VuFind/ILS/OAuth2Service.php
xmorave2: I added AuthToken exception class

xmorave2:
> Tested to work with KohaRest, so functionality-wise it's all good.

Thanks for testing!

> Is there a compelling reason for making the OAuth2 authenticator a service? I think it would simplify things if it was a trait that any ILS driver could simply use if needed.

There is no particular reason for that, I refactored it to trait now

> If it's kept as a service, it would be better to make it more generic and live e.g. in \VuFind\Service namespace. In that case I'd also remove logging from it and make it just throw exceptions to allow the caller to handle errors in a way appropriate for the situation.
> 
> Alternatively, this might be the time to consider switching to a proper OAuth2 client such as league/oauth2-client. I see that you've already done a lot of work with the refactoring and tests etc, but I still think this is worth considering. I haven't checked if it's easy to plug our own HTTP client to league/oauth2-client, so that might be one obstacle. What do you think?
> 
> I'd also welcome @demiankatz to comment especially reagrding the architecture.

I make a bit of investigation. And there is only one obstacle for this I can see till now: It is possible to inject own http client into Provider class of league/oauth2-client library. But the client must implement GuzzleHttp\ClientInterface and we are using laminas interface.

xmorave2:
So as @demiankatz pointed, migration to the PSR compliant HTTP client is probably the first step here...

EreMaijala:
> I make a bit of investigation. And there is only one obstacle for this I can see till now: It is possible to inject own http client into Provider class of league/oauth2-client library. But the client must implement GuzzleHttp\ClientInterface and we are using laminas interface.

So that makes the choice for the new HTTP client obvious. :) Guzzle would have been the one I'd consider in any case. I'm ok with leaving the switch to a proper OAuth2 library to a later stage.

xmorave2:
And it looks like league/oauth2-client could be PSR-18 compliant in the meantime: https://github.com/thephpleague/oauth2-client/pull/885

So we could choice any PSR-18 client we want ;)

demiankatz:
Thanks! I'm going to merge this as-is, and we can revisit the Oauth2 library question in the future. I have opened two JIRA tickets to capture this: [VUFIND-1481](https://vufind.org/jira/browse/VUFIND-1481) for the HTTP client library update, and [VUFIND-1482](https://vufind.org/jira/browse/VUFIND-1482) for the use of an Oauth2 client library.

xmorave2:
Thanks  @demiankatz!

PR #1974, opened 2021-06-02
Add support for modifying holds.

EreMaijala:
This PR adds support for updating selected holds in lightbox depending on what the ILS driver supports. 

Notes:
- The markup for the Select all checkbox in holds.phtml is a bit odd, but it's the same as in the checkouts page.
- The changes in common.js is so that the "Select all" checkbox triggers the change events for the checkboxes whose values it changes. This is required for the button activation.
- Form handling is a bit complicated due to the no-js support for canceling requests.
- Update won't work without JS, so the button is only displayed via JS.
- Creation of $safeId is changed to work properly when there are multiple holds on a single bibliographic record.
- While for canceling requests the item_id field is used for identification, I decided to require reqnum with update support since that's most appropriate and should be unique.
- "Start date" and "Frozen until" are mostly the same at least for the drivers involved at the moment, but they're handled as different ones based on what's logical.
- There's one potential issue: holds may have different possible pickup locations, and modifying them in a batch will fail for some. I believe this is something that can be lived with since more often than not the pickup locations are the same, but if deemed necessary, making pickup location editable from the holds list for each hold individually could be implemented as well.

TODO:
- [x] Update the driver documentation wiki page


module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -66,8 +68,29 @@ public function listAction()
             return $view->cancelResults;
         }
 
-        // By default, assume we will not need to display a form:
+        // Process any update request results stored in the session:
+        $updateResultsContainer = $this->getHoldUpdateResultsContainer();
+        $holdUpdateResults = $updateResultsContainer->results ?? null;
+        if ($holdUpdateResults) {
+            $view->updateResults = $holdUpdateResults;
+            $updateResultsContainer->results = null;
```
demiankatz, line 97: Should this logic come AFTER the POST-processing logic? Otherwise, it seems there might be some edge case somewhere where we consume results from the container, but then never display them because the update code forwarded somewhere else.

demiankatz: (Or could we make the whole thing more clear by actually creating a separate action to be the target of the form, so we don't need to have so many conditional paths inside this one method?).

module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -102,4 +138,170 @@ public function listAction()
...
+            $this->flashMessenger()
+                ->addErrorMessage('error_inconsistent_parameters');
+            return $this->redirect()->toRoute('holds-list');
+        }
+
+        // Get list of pickup locations based on the first selected hold. This may
+        // not be perfect as pickup locations may differ per request, but it's the
+        // best we can do.
+        $holds = $catalog->getMyHolds($patron);
+        $firstDetails = reset($details);
```
demiankatz: I think this block of code might break out nicely into a separate getPickupLocations() method.

module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -102,4 +138,170 @@ public function listAction()
...
+                    $this->flashMessenger()
+                        ->addErrorMessage('ils_connection_failed');
+                }
+                break;
+            }
+        }
+
+        $gatheredDetails = $this->params()->fromPost('gatheredDetails', []);
+        if ($this->params()->fromPost('updateHolds')) {
+            $validPickup = true;
```
demiankatz: Might it be feasible to move all the logic that populates gatheredDetails to a separate method, again just to make this action easier to read? Not sure how woven together all the variables are, but it seems like it might be possible to split a big chunk of this processing off.

module/VuFind/src/VuFind/Controller/Plugin/Holds.php
```diff
@@ -268,4 +300,42 @@ public function validateDates($startDate, $requiredBy, $enabledFormFields)
...
+    ): array {
+        $result = [
+            'frozenUntilTS' => null,
+            'errors' => [],
+        ];
+        if (!in_array('frozenUntil', $extraHoldFields) || empty($frozenUntil)) {
+            return $result;
+        }
+
+        $errors = [];
```
demiankatz: Since you initialize $result['errors'] above, would it be easier to just add messages directly to that instead of introducing another $errors variable and reassigning it later?

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -1674,6 +1677,77 @@ public function getCancelHoldDetails($hold, $patron = [])
             ? $hold['reqnum'] : '';
     }
 
+    /**
+     * Get Update Hold Details
+     *
+     * Get required data for updating a hold. This value is relayed to the
```
demiankatz: Should the comment make more clear that an empty string means the hold is not eligible for modification? Or am I overlooking/misunderstanding something?

themes/bootstrap3/js/common.js
```diff
@@ -422,20 +422,26 @@ $(document).ready(function commonDocReady() {
   setupQRCodeLinks();
 
   // Checkbox select all
-  $('.checkbox-select-all').change(function selectAllCheckboxes() {
+  $('.checkbox-select-all').on('change', function selectAllCheckboxes() {
```
demiankatz, line 425: Are the changes in this file directly related to the rest of the PR, or are these general modernizations/fixes that stand alone?

themes/bootstrap3/templates/holds/edit.phtml
```diff
@@ -0,0 +1,97 @@
...
+  // Set page title.
+  $this->headTitle($this->translate('hold_edit_title'));
+
+  // Set up breadcrumbs:
+  $this->layout()->breadcrumbs = '<li><a href="' . $this->url('myresearch-home') . '">' . $this->transEsc('Your Account') . '</a></li>'
+    . '<li><a href="' . $this->url('holds-list') . '">' . $this->transEsc('My Holds') . '</a></li>'
+    . '<li class="active">' . $this->transEsc('hold_edit_title') . '</li>';
+?>
+<h2><?=$this->transEsc('hold_edit_title')?></h2>
+<?php if ($this->helpText): ?>
```
demiankatz: Where does this come from? Is it intentionally not escaped or translated?

themes/bootstrap3/templates/holds/edit.phtml
```diff
@@ -0,0 +1,97 @@
...
+    <div class="form-group hold-required-by">
+      <label for="required_by_date" class="control-label"><?=$this->transEsc('hold_required_by')?>:</label>
+      <input id="required_by_date" type="text" name="gatheredDetails[requiredBy]" value="<?=(isset($this->gatheredDetails['requiredBy']) && !empty($this->gatheredDetails['requiredBy'])) ? $this->escapeHtmlAttr($this->gatheredDetails['requiredBy']) : ''?>" size="10" class="form-control"/>
+      (<?=$this->dateTime()->getDisplayDateFormat()?>)
+    </div>
+  <?php endif; ?>
+
+  <?php if (in_array('pickUpLocation', $this->fields) && $this->pickupLocations): ?>
+    <?php
+      if (isset($this->gatheredDetails['pickUpLocation']) && $this->gatheredDetails['pickUpLocation'] !== "") {
```
demiankatz: You could use `($this->gatheredDetails['pickUpLocation'] ?? '')` to eliminate the isset check...

demiankatz (commented):
I've taken a first scan through the code, without necessarily looking at everything in depth, just to get the conversation started. Here are a few initial thoughts!

I'm also going to do a little hands-on testing, so I'll add further comments if I run into any problems there.

It's probably worth adding Mink testing of this functionality once it's finalized; I can help with that when the time comes.

demiankatz:
I just tried it out and it seems to work as expected. I do wonder if we need to worry about the pickup location problem when the holds come from multiple libraries and the locations are not compatible -- perhaps that is unlikely, but the results could be very confusing when it happens. Maybe that's a problem to solve another day, though.

I also definitely understand why bulk editing is useful here, but I also wonder if it would be useful to have an "edit" button on each hold which essentially has the same functionality as checking the box and clicking the edit button, so people can easily change just one hold without as many clicks. But again, that's just a thought and not something we need to implement right now.

module/VuFind/src/VuFind/Controller/HoldsController.php
EreMaijala: A separate path is compelling, but it makes things difficult if the form needs to be displayed again due to validation issues. But perhaps it makes sense to branch to sub-methods internally.

There should be no such case where we have update results but are still redirected to the update form. If that were to happen, it would still make sense to clear the old results as otherwise displaying them later might be confusing as they wouldn't necessarily relate to the changes submitted.

demiankatz: That makes sense. I do agree that branching to sub-methods might be a good compromise solution. In the MyResearchController, there were too many methods, so we made the methods very long to keep related things together. Now that we have a whole controller for holds, I think we can get away with spreading out to more methods for the sake of improved readability and maintenance.

EreMaijala: I think it's now better already with your proposed changes of splitting blocks out of editAction. What remains is mostly just about displaying the form and results. What do you think?

module/VuFind/src/VuFind/Controller/HoldsController.php
EreMaijala: Done.

module/VuFind/src/VuFind/Controller/HoldsController.php
EreMaijala: Done.

module/VuFind/src/VuFind/Controller/Plugin/Holds.php
EreMaijala: Simplified.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: Now the method doesn't exist anymore. Need to document the field returned in getMyHolds.

themes/bootstrap3/js/common.js
EreMaijala: The changes in common.js is so that the "Select all" checkbox triggers the change events for the checkboxes whose values it changes. This is required for the button activation. While at it, I changed the deprecated `.change()`.

themes/bootstrap3/templates/holds/edit.phtml
EreMaijala: Oops, that was a copy-paste from the new hold form that I used as the base.

themes/bootstrap3/templates/holds/edit.phtml
EreMaijala: It was actually possible to simplify further since it doesn't need to handle all the options that the hold creation form does.

module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -102,4 +157,213 @@ public function listAction()
...
+                $gatheredDetails,
+                $pickupLocations
+            );
+            if ($updateFields) {
+                $results
+                    = $catalog->updateHolds($selectedIds, $updateFields, $patron);
+                $successful = 0;
+                $failed = 0;
+                foreach ($results as $result) {
+                    if ($result['success']) {
```
demiankatz, line 217: I think you could be clever with `$result['success'] ? ++$successful : ++failed;` here if you want to shave some lines of code. Feel free to disregard if you think that's too clever for our own good.

module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -102,4 +157,213 @@ public function listAction()
...
+    protected function getPickupLocationsForEdit(array $patron, array $selectedIds
+    ): array {
+        $catalog = $this->getILS();
+        $holds = $catalog->getMyHolds($patron);
+        $firstDetails = reset($selectedIds);
+        $pickupLocations = [];
+        foreach ($holds as $hold) {
+            if ((string)($hold['updateDetails'] ?? '') === $firstDetails) {
+                try {
+                    $pickupLocations = $catalog->getPickUpLocations($patron, $hold);
```
demiankatz: Since this is a method now, I think you can slightly simplify things by just directly returning here, eliminating the break and the $pickupLocations variable, and returning an empty array at the bottom of the function as a default. I think that's functionally equivalent but slightly simpler.

module/VuFind/src/VuFind/Controller/Plugin/Holds.php
```diff
@@ -250,22 +255,57 @@ public function validateDates($startDate, $requiredBy, $enabledFormFields)
...
+        $result = [
+            'frozenUntilTS' => null,
+            'errors' => [],
+        ];
+        if (!in_array('frozenUntil', $extraHoldFields) || empty($frozenUntil)) {
+            return $result;
+        }
+
+        try {
+            $result['frozenUntilTS'] = $frozenUntil
```
demiankatz: If $frozenUntil is a "falsey" value, won't we return because of the empty check above? Is there any reason to check its value again here?

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -1674,6 +1683,56 @@ public function getCancelHoldDetails($hold, $patron = [])
...
+    ): array {
+        $results = [];
+        $session = $this->getSession($patron['id']);
+        foreach ($session->holds as &$currentHold) {
+            if (!isset($currentHold['updateDetails'])
+                || !in_array($currentHold['updateDetails'], $holdsDetails)
+            ) {
+                continue;
+            }
+            if ($this->isFailing(__METHOD__, 25)) {
```
demiankatz, line 1707: Since we have a new method that can fail, we should add this method to the Failure_Probabilities section of Demo.ini for consistency.

themes/bootstrap3/templates/holds/edit.phtml
```diff
@@ -0,0 +1,89 @@
...
+
+<form action="<?=$this->url('holds-edit')?>" class="form-edit-holds" method="post" name="editHolds">
+  <input type="hidden" name="csrf" value="<?=$this->escapeHtmlAttr($this->auth()->getManager()->getCsrfHash())?>">
+  <?=$this->flashmessages()?>
+
+  <?php foreach ($this->selectedIDS as $id): ?>
+    <input type="hidden" name="selectedIDS[]" value="<?=$this->escapeHtmlAttr($id)?>">
+  <?php endforeach; ?>
+
+  <?php if (in_array('frozen', $this->fields)): ?>
```
demiankatz, line 20: I confess that I haven't tried to work this out from either looking at the code or interacting with the interface, but I'm curious what the relationship is between frozen and frozenUntil. Are there any interdependencies between them? I can imagine scenarios where an ILS supports one or the other or both... but this may lead to some edge cases where one takes precedence over the other, or some extra rules might come into play. Do we need to worry about any of that?

demiankatz (commented):
Thanks for the progress, @EreMaijala, this is coming along very nicely. I've left a few more comments based on a closer reading of the revised files. Let me know if/when you'd like me to try to add some mink tests to the mix!

module/VuFind/src/VuFind/Controller/HoldsController.php
EreMaijala: Simplified. I decided to return default pickup locations at the bottom as a last resort, though I'm not quite sure if that's better than not providing the choice in the first place.

module/VuFind/src/VuFind/Controller/Plugin/Holds.php
EreMaijala: Right, simplified.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: Done.

themes/bootstrap3/templates/holds/edit.phtml
EreMaijala: This needs to be documented, but as I see it frozenUntil can be supported alongside frozen, but not without it.

EreMaijala:
@demiankatz I've updated the relevant parts of the ILS drivers wiki page.

demiankatz:
@EreMaijala, I did some refactoring (see #1979) to create a separate test class for holds, so we can add some more tests for the functionality here without having to sit through the entire IlsActionsTest. If you can approve that PR, we can merge it in here and make some further progress. However, I also notice that IlsActionsTest is failing in this branch for some reason. I'm out of time for today, so I haven't had an opportunity to investigate further. If you're in a position to look into this, please let me know what you find. Otherwise, I'll take a closer look later in the week when I'm caught up again. :-)

demiankatz:
@EreMaijala, I've fixed the broken tests, but now I've run out of time to actually write more. I'll try to get to that later in the week (or maybe even later today, if I'm lucky and get through my other tasks...).

demiankatz:
@mtrojan-ub, is there any chance you have time to merge the dev branch into this branch? Because of the recent addition of GitHub Actions configuration to VuFind's repository, I do not have appropriate permissions to merge and push that myself, since it adds a workflow to your repo. If you can do that once, then I should be able to add changes again as a follow-up. If you do not have time, I can begin the work in this slightly outdated branch, but I'd prefer to have the very latest code if possible.

demiankatz:
@EreMaijala, I had a little bit more time to work on tests, and I think I may have found a problem (though perhaps it's unavoidable). Take a look at commit 3fe72fe -- this adds a test which sets a start date in the future, and then confirms that the hold shows up as frozen. However, when the test sets the start date to two days in the future (06-10-2021), the holds screen displays a date one day in the future (06-09-2021). Not sure if this is always going to happen, or if it's dependent on time of day. Maybe the test just needs to be made a little more flexible, so it accepts a range of dates in the output... but I first wanted to check and see if this might be a simple bug somewhere.

I was also going to proceed to add some tests to update holds, but I was having trouble finding the right combination of settings to make editing available. I know I was able to do this manually a couple of days ago, but today I had no luck. Is there anything I need to do beyond uncommenting the relevant line in Demo.ini?

EreMaijala:
@demiankatz The day is one off because you enter a start date (the first day it will be active), but the list displays the last day it's frozen (not active). I'm not sure if there is a better way without making the message more complicated.

For editing the updateFields setting in Holds section should be enough. I can re-check if it doesn't work for you.

demiankatz:
Thanks for the clarification, @EreMaijala. I adjusted the test to pass based on the current behavior, but I'm not sure if I agree with the English wording. If we say "frozen until 6/9" that suggests to me that the freeze is over at the beginning of 6/9, not the end. If we wanted to suggest that the freeze lasts for the whole day, I would say "frozen through 6/9." This is somewhat up for debate, but that's my personal interpretation. I don't have strong feelings about this, but perhaps it's worth adjusting either behavior or wording for greater clarity (in which case, we'll need to make the test match).

I'm also still having trouble getting the edit controls to appear. I have committed my work in progress on an edit test, which just sets up the Demo.ini file and then has a TODO comment where I would like to do the actual editing. If you put a `die();` into the test right before the TODO comment, then you can get everything set up and interact with it in your browser. Do you mind taking a look at this and letting me know if I'm missing something obvious?

Thanks!

EreMaijala:
@demiankatz Thanks for all the work. I had forgotten to add updateDetails to the new hold placeHold, so it didn't work for newly added holds. Should be fixed now.

I've also revised terminology to use "frozen through".

demiankatz:
@EreMaijala, during testing, I ran into this after clicking on the "edit" button for a hold:

image

I had stepped away for quite some time, so I suspect the problem is that my session timed out and I was no longer logged in... but regardless of the cause, there seems to be a problem with lightbox error handling in this situation. Not sure if it's specific to this PR or a more generalized issue. Let me know if you want me to dig deeper into this and/or get @crhallberg involved.

EreMaijala:
@demiankatz Oh, that seems to happen if the hold id is no longer valid. Expired session works fine and so does a lost CSRF check during editing, but this is obviously an issue. I'll check.

demiankatz:
Thanks, @EreMaijala... and it definitely seems like this is more than a session expiration problem, because after repeated testing, I can't ever get it to work. I'm out of time for today, but I've pushed up a little more progress on the test suite... once the edit dialog works again, it should only take a few minutes to finish up!

EreMaijala:
@demiankatz Should be fixed now. Can be tested by manipulating the id in the edit link or with a lucky clear_demo=1 parameter in another tab.

EreMaijala:
@demiankatz Thanks for the steps to reproduce, I was finally able to find out the problem. When getting rid of getUpdateHoldDetails method I forgot that it also added the IDs to the list of valid ones. Thus, when canceling holds was disabled, editing wouldn't work either. Should be fixed now.

demiankatz:
Thanks, @EreMaijala, everything seems to be working now! Once again, the effort of writing tests has more than justified itself!

I've added tests for all the permutations of cancel-enabled/updates-enabled in combination with hold freezing. I tried to do the same things in different ways when I repeated actions across tests, to maximize the coverage of functionality. There is obviously more we could do -- particularly in terms of testing editing of multiple holds at the same time -- but I'm out of time, and I think this is a good start. If we find problems in the future, we can add tests to cover them more specifically.

I think this is now ready to merge, unless you have any remaining concerns. The only question I had is whether you intend to add edit support for request groups. I'm guessing you omitted this because it would add significant complexity, and I'm fine with leaving it out for now (I don't use this feature in any of my own instances)... but if anybody is actually using request groups, I imagine that being able to edit the pickup location but not the group could be confusing, so maybe this is something we should revisit in the future if that need really exists.

EreMaijala:
@demiankatz I think that changing request group would work when a single hold is being edited, but it could also be considered such fundamental part of the hold request (comparable to moving a hold to another item) that it wouldn't be possible to modify it without practically replacing the hold. We currently use request groups to select between normal hold and a regional hold fulfilled by any of several regional libraries, and switching between those is not supported. I think this is something that could be revisited as necessary.

Another option I have considered is to check pickup locations for all the selected holds at least when there are less than X of them. Then we could do something nice like display only the ones available for all of them or not allow selection and display message instead requesting to change the holds individually. But I think that can also be a future enhancement, and at least some systems have fairly complicated rules for determining available pickup locations, so I wouldn't want to slow things down too much.

demiankatz:
@EreMaijala, that makes sense to me. I do wonder if the request groups could be used as a shorthand check for possible problems -- i.e. we already have the request group information when we're displaying the hold list. If the user checks holds from two different request groups, we know without further checking that adjusting pickup location isn't going to work, so we could suppress that control. Maybe that would be a place to start if we want extra validation, and we can worry about getting fancier later, if we have to. (And with your idea of checking multiple holds, it could be a configurable "max_holds_to_check" kind of setting, so institutions that have simple rules could use a high number, and institutions where it would not be beneficial at all could turn it off). In any case, do you want me to merge it as-is now, and then we can revisit these other things in a separate PR in the future if and when necessary?

EreMaijala:
@demiankatz I think a follow-up with the enhancements would be better than dragging this along any further, if that's ok for you.

demiankatz:
Agreed -- merging now!

PR #1975, opened 2021-06-02
Reduce logging severity of RecordMissing to debug.

EreMaijala:
This reduces logging in production environments where e.g. crawlers may cause a lot of hits on non-existent records.


demiankatz:
This makes a lot of sense to me -- thanks!

PR #1978, opened 2021-06-04
show open fines/fess instead of closed Folio.php

rtwilson:


demiankatz:
Thanks, @rtwilson! Since we already discussed this with @ihardy on Slack, I think it's safe to merge immediately. I'm going to backport it to the release-7.1 branch so that it can be included in the next minor release, since I think this qualifies as a bug fix. If anyone can think of a reason why we might want to make this configurable, so that Closed fines can also be included, let me know... I imagine there might be some value in being able to view Fine history, but that's not what we normally do in VuFind (in the Voyager driver, for example, we only show fines with an outstanding balance).

demiankatz:
Backported and merged to dev!

PR #1979, opened 2021-06-07
Split holds tests into their own class.

demiankatz:
This PR splits the Holds-related tests out of IlsActionsTest in order to make the test suite more granular and to reflect the separation of HoldsController from MyResearchController.

### DIFF
--- a/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php
+++ b/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php
@@ -1,0 +1,383 @@
+<?php
+/**
+ * Test class for holds-related functionality.
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2021.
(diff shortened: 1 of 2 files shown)


EreMaijala:
@demiankatz There might be a side-effect that's causing testCancelIllRequest to fail for me with `Selector not found: a.placeILLRequest` when all tests are run, but it works when only IlsActionsTest is run.

demiankatz:
@EreMaijala, this branch is passing all tests in my environment when I run the full suite, so something must be different on your end. Is it possible it's a timing issue of some sort? I'd be interested to know if adding `-Dsnooze_multiplier=2` to the Phing command line makes a difference to your outcome. (This will cause the tests to run more slowly).

EreMaijala:
@demiankatz Right, this seems to be a timing problem on my end.

EreMaijala (approved):

demiankatz:
Thanks, @EreMaijala!

PR #1980, opened 2021-06-08
Fix KohaRest driver alias issue with LoggerAwareTrait.

EreMaijala:
The problem was caused by both the driver and OAuth2TokenTrait using the same trait and ambiquity with the error alias.

This is quite trivial change, but requesting review just in case (the first one will suffice).


xmorave2 (approved):
This is pretty straightforward. Thanks for catching this @EreMaijala

demiankatz:
Thanks, @EreMaijala and @xmorave2; I have merged this now.

This is clearly a sign that we need a unit test for the KohaRest driver... even if it doesn't have full coverage, a bare minimum test would have caught this issue for us. (If memory serves, there's an abstract base test for ILS drivers that can be extended to get the ball rolling). Might be worth establishing one the next time KohaRest changes are submitted.

EreMaijala:
@demiankatz Agreed.

PR #1981, opened 2021-06-10
Add support for getting JS translations statelessly.

EreMaijala:
Use the new mechanism where appropriate instead of hand-crafted JSON. Also adds tests for the jsTranslations helper.

Note that for visual facets there was custom code that added some of the translations without escaping and some with escaping. I renamed two translation keys to account for that, but while the results in the page source look fine, I couldn't get visual facets to work to verify (it's only telling me to enable javascript and update my browser).


module/VuFind/src/VuFind/View/Helper/Root/JsTranslations.php
```diff
@@ -88,25 +88,35 @@ public function addStrings($new)
...
+     *
+     * @param array $strings Strings to translate (key = js key, value = string to
+     * translate)
+     *
+     * @return string
+     */
+    public function getJSONFromArray(array $strings): string
+    {
+        $view = $this->getView();
+        $translate = $view->plugin('translate');
```
demiankatz: It seems a little strange to inject transEsc through the constructor but to pull translate out of the view here. I think we should choose one method or the other and be consistent. (I somewhat prefer the injection approach because it's more explicit and easier to test).

themes/bootstrap3/templates/Recommend/VisualFacets.phtml
```diff
@@ -34,26 +34,23 @@
...
-      . "});\n";
-
-    $script .= <<<JS
+    $translations = $this->jsTranslations()->getJSONFromArray(
+      [
+        'More Topics' => 'More Topics',
+        'visual_facet_parent' => 'visual_facet_parent',
+        'items' => 'items',
+        'more_topics_html' => 'more_topics',
+        'on_topic_html' => 'on_topic'
```
demiankatz: While I appreciate the way you were able to avoid editing the language files here, I think it might be worthwhile to just rename the more_topics and on_topic keys in the source files, rather than simply mapping them here. If we're using them as HTML keys, we might as well label them correctly in the files for consistency with our conventions. I assume they're not being used in any other context, though it's worth double-checking if you haven't already. I'm happy to take care of the language file cleanup if you like -- should be very quick with the help of the command line tools.

demiankatz (changes requested):
Thanks for this -- a couple of minor suggestions. I haven't had a chance to actually test this yet, but I will after you've made your next pass at the code. I'm not sure why you're having trouble with visual facets -- there is a Mink test for this, so they must be working on your system sometimes if the test suite is passing. Maybe the error message is misleading if some part of the configuration is missing (since setting this up is a little weird). I'll look into that while I'm trying this later on!

module/VuFind/src/VuFind/View/Helper/Root/JsTranslations.php
EreMaijala: Agreed. I tried to keep the constructor back-compatible, but I'll change that.

themes/bootstrap3/templates/Recommend/VisualFacets.phtml
EreMaijala: This is a bit complicated. In JsTranslations the key having _html suffix means that it's not html-encoded regardless of whether the translation key in the ini file has an _html suffix. Such string can be directly put into e.g. `$('.something').attr('title', VuFind.translate('title_html'));` whereas a normal encoded string would need to be decoded first. So, the translations key in the ini file does not have an _html suffix to indicate that you can't just include markup in it and expect it to show, while the js translation key has it to indicate that we don't want it encoded.

This means we have a dual meaning for the _html suffix, which is really confusing. I would have preferred something like _noenc suffix for JsTranslations, but if changing it would break existing functionality, I'm a bit worried to do it. What we could do is add _noenc as a preferred alternative to _html in JsTranslations while keeping _html for back-compatibility.  What do you think?

Note that there's no existing example of this being used in core code, so this is all something I've tried to deduce from the code itself, and I could have it wrong. But I made sure the tests pass with the original version as well. The _html suffix was introduced in commit 75654a980a98e1336292669e3c12845aa0772c42.

demiankatz: I'm pretty sure that the thought behind the commit you cite was simply, "if a translation string contains HTML and we encode it, it will display incorrectly, so we must not encode it." The assumption was that we would be using the same keys in the JS and in the language files. The potential side effect of being able to use the _html suffix in an alias to change encoding behavior was not something I ever considered. Indeed, the entire jsTranslations() mechanism is just supposed to be a dictionary of translations that will be needed, mapped directly from the existing translation system. If we are changing keys in some places, that is not intentional or desired (at least according to my original design ideas). The only reason we're not just dumping the entire translation file into the Javascript is to save bandwidth and memory. And if there's any common theme to the things I've just said, it's probably that I never really thought this through very deeply. :-)

I didn't write the original visual facets code (though I may have contributed to the translation logic while it was being developed; I honestly can't remember). Regardless of whether or not it's all my fault, I'm not entirely sure why the translations are being handled the way they are there. I wonder, though, if we're trying to do something that's inconsistent with the default behavior of the translator, whether it's actually better to have a separate method for getting unescaped translations. We have separate translate and transEsc methods in the PHP for a reason, so why not do the same thing in the Javascript, and then we do the right thing based on context rather than magic? I realize that's a bigger project, and maybe not worth tackling right now... but I'd consider that before I'd consider more special-purpose suffixes. :-)

Is that any help?

EreMaijala: If that indeed is the original idea, the helper should be checking the mapped translation key for the _html suffix instead of the JS array key since the actual translation key in the ini file is the authority. Usually they match, sure, but still. Nevertheless, there's still the problem that when using jQuery methods you don't want the strings to be pre-escaped, and it's not about them having HTML content but about HTML encoding of normal text. I made a silly little example branch trying to illustrate the issue, see the last commit in https://github.com/EreMaijala/vufind/tree/translation-issue.

The problem as I see it is that JS translations shouldn't have been escaped by the translator at all. That should have been left to the JS that uses the translation to handle any ecaping needed per case. But changing that after the fact would be a huge back-compatibility break I wouldn't want to do. So, what we're left with is sub-optimal, but hard to change.

Would it make it any better if there was another suffix, say _no_enc for the JS translation key that would indicate that the translation is normal text but not pre-encoded? I'll add that so you can judge the results.

EreMaijala:
@demiankatz I posted #1982 to fix visual facets in certain situations.

module/VuFind/src/VuFind/View/Helper/Root/JsTranslations.php
EreMaijala: Done. Note that the view and translate view helper are still needed in the tests since TransEsc pulls Translate helper from the view.

themes/bootstrap3/templates/Recommend/VisualFacets.phtml
EreMaijala: I added an `_unescaped` suffix. It works the same as _html but makes the semantics a bit more clear, I think.

demiankatz: I think this is an acceptable compromise under the circumstances, because I agree that changing the system more fundamentally might be unreasonably disruptive... but it would be great to simplify this in future. Perhaps we can add this to the to-do list for the next theme. Do you think it's worth opening a JIRA ticket about it?

demiankatz:
@EreMaijala, I noticed that there was an opportunity to further simplify some ugly code, and also that we had no test cases for the (possibly unused?) edge case where an array is included as a value in the array provided to the JsTranslations helper. I addressed both of these things in commit 07adb68. Do you mind reviewing that and letting me know if you have any concerns? If you're okay with that change, I think we can merge this (as long as all tests are passing -- I'm running the suite now).

demiankatz:
All tests are passing, so I'll just wait for a final okay before merging.

EreMaijala:
@demiankatz Thanks, looks good to me! A JIRA ticket would be good if indeed there's a way to remember to take it into account when creating the new theme.

demiankatz:
I've opened [VUFIND-1484](https://vufind.org/jira/browse/VUFIND-1484) to track the need for jsTranslations improvements, and I've linked it to the ticket about theme redesign as a reminder that these projects could be connected.

PR #1982, opened 2021-06-15
Fix visual facets to enable faceting

EreMaijala:
Visual facets didn't work with deferred side facets or no side facets as it didn't enable faceting. Also the JavaScript message and browser update message were left under the visualization, and if there wasn't anything to display, they were left on screen.


demiankatz:
Thanks, @EreMaijala! I've updated the test suite to cover this case, and confirmed that it was broken before your changes and fixed afterward. Merging now!

demiankatz:
I backported the actual fix to the release-7.1 branch since it's a bug fix, but I'm only merging the test revisions to dev to keep things simple.

PR #1983, opened 2021-06-15
Make EBSCO URLs configurable; fix for HTTPS compatibility.

demiankatz:
EBSCO will be shutting down HTTP URLs soon; this PR makes all URLs configurable and switches to HTTPS by default. I am targeting this against release-7.1 instead of dev so that we can get the change released more quickly, so users can more easily get up to date before service is potentially disrupted.


EreMaijala (approved):
Looks good to me, and everything works.

demiankatz:
Thanks!

PR #1984, opened 2021-06-16
Remove leftover OAuth2Service class reference

xmorave2:
It was not removed during refactoring from service class to trait


demiankatz:
Thanks for the catch!

PR #1989, opened 2021-06-16
Aleph driver: parsing of duedate in getHolding

xrosecky:
Duedate in getHolding should contain date or null if not checked out.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Aleph.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Aleph.php
@@ -335,13 +335,6 @@ class Aleph extends AbstractBase implements \Laminas\Log\LoggerAwareInterface,
     use \VuFindHttp\HttpServiceAwareTrait;
 
     /**
-     * Duedate configuration
-     *
-     * @var array
-     */
-    protected $duedates = false;
-
-    /**
      * Translator object
      *
      * @var AlephTranslator
@@ -532,9 +525,6 @@ class Aleph extends AbstractBase implements \Laminas\Log\LoggerAwareInterface,
             $this->dlfbaseurl = $this->config['Catalog']['dlfbaseurl'];
         }
         $this->sublibadm = $this->config['sublibadm'];
-        if (isset($this->config['duedates'])) {
-            $this->duedates = $this->config['duedates'];
-        }
         $this->available_statuses
             = explode(',', $this->config['Catalog']['available_statuses']);
         $this->quick_availability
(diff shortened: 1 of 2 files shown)


demiankatz:
I'm not sure I entirely understand what's happening here, but I'm inclined to trust you. Nonetheless, should we ask for input from any other Aleph users before merging this one, since it seems to remove a significant amount of (apparently unnecessary) logic?

xrosecky:
Yes, I would wait for the input from the other Aleph users.

If I understand the documentation on wiki correctly, the duedate in getHolding should contain only date. Is that so?

demiankatz:
Thanks, @xrosecky, I'll see if I can get another opinion or two on this. Regarding the 'duedate' field, it is only used for display purposes, so in theory it can be anything, but it is intended to be a human-readable date. Other values will work in the sense that the text will display in the status area, but it would not be the intended use of the field.

welblaud:
Wouldn't it be also good to remove the parts associated with the `duedates` property [here](https://github.com/vufind-org/vufind/blob/bf8efcc349c4c9480af0cb256bb8050f8309193e/module/VuFind/src/VuFind/ILS/Driver/Aleph.php#L342) and [here](https://github.com/vufind-org/vufind/blob/bf8efcc349c4c9480af0cb256bb8050f8309193e/module/VuFind/src/VuFind/ILS/Driver/Aleph.php#L535) with the associated settings [in the config](https://github.com/vufind-org/vufind/blob/bf8efcc349c4c9480af0cb256bb8050f8309193e/config/vufind/Aleph.ini#L65)? They are probably not needed any more.

xrosecky:
Fixed.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -942,7 +932,7 @@ public function getHolding($id, array $patron = null, array $options = [])
             }
             $matches = [];
             $dueDateWithStatusRegEx = "/([0-9]*\\/[a-zA-Z]*\\/[0-9]*);([a-zA-Z ]*)/";
-            $dueDateRegEx = "/([0-9]*\\/[a-zA-Z]*\\/[0-9]*)/";
+            $dueDateRegEx = "/([0-9]*\\/[a-zA-Z0-9]*\\/[0-9]*)/";
```
demiankatz, line 936: I'm guessing that this change is needed because of some kind of date formatting difference between Aleph systems -- i.e. dates may be 06/Jun/2021 in some places, and 06/06/2021 in others. If that's the case (and maybe it's not, because I'm just guessing), does the same change need to be made to $dueDateWithStatusRegEx above, in order to correctly extract request status where appropriate? Or is all of that obsolete too?

xrosecky: Yes, you are right, fixed.

demiankatz:
Thanks, @xrosecky and @welblaud! I think this is now ready to merge. I double-checked the history of the functionality that is removed here, and it appears to have been carried over from VuFind 1.x and not touched since... therefore, I think it is very likely that it is no longer used/needed at this point... and if anyone complains, we should reintroduce the functionality in a more appropriate way, rather than abusing the due date field.

PR #1990, opened 2021-06-16
Aleph driver: return isbn as string instead of single value array

xrosecky:


demiankatz:
Thanks -- merging now!

PR #1992, opened 2021-06-17
Reset ChoiceAuth state when any exception occurs.

EreMaijala:
E.g. the email authenticator may throw an exception other than AuthException.


demiankatz:
Makes sense to me!

PR #1993, opened 2021-06-17
Ignore RecordMissing exceptions in Whoops.

demiankatz:
- This restores functionality that was lost when migrating from zf2-whoops to laminas-whoops.


EreMaijala (approved):
Looks good and works great!

demiankatz:
Thanks!

PR #1994, opened 2021-06-18
Aleph driver: z304 mapping of address can be overriden for REST DLF API

xrosecky:


module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -475,6 +475,12 @@ class Aleph extends AbstractBase implements \Laminas\Log\LoggerAwareInterface,
      */
     protected $defaultPatronId;
 
+    /**
+     *
```
demiankatz, line 482: There's a comment missing here.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -567,6 +573,27 @@ public function init()
...
+        $this->addressMappings = $this->getDefaultAddressMappings();
+
+        if (isset($this->config['AddressMappings'])) {
+            foreach ($this->config['AddressMappings'] as $key => $val) {
+                $this->addressMappings[$key] = $val;
+            }
+        }
+    }
+
+    protected function getDefaultAddressMappings() {
```
demiankatz: This function needs a comment too.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -567,6 +573,27 @@ public function init()
...
+        return [
+            'barcode' => 'z304-address-5',
+            'fullname' => 'z304-address-1',
```
demiankatz, line 612: fullname and barcode are not documented values supported by getMyProfile. I wonder if we should add official support for either or both of these; they have been part of this driver for quite some time, but presumably are not getting used anywhere in core code.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -567,6 +573,27 @@ public function init()
...
+            'address1' => 'z304-address-3',
```
demiankatz: Looks like address-3 is getting used here twice, and address-2 is not used at all. Is that intentional?

demiankatz (changes requested):
Thanks for this contribution! I've left a few suggestions and questions; I haven't thoroughly reviewed all the mappings -- I looked enough to spot a couple of possible problems. Once these are fixed, please let me know, and I'll give it another, more thorough, review.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
xrosecky: We parse firstname and lastname from fullname attribute because Aleph does not store them in separate fields.

demiankatz: Makes sense, but my question is, would it be better if the VuFind core looked for a fullname field, and displayed that instead of firstname/lastname when present? If you just use that value for processing internal to the ILS driver, it doesn't necessarily need to be exposed through the data returned by the driver... but if using that exposed data would be useful, it's something we could look into supporting. In any case, it certainly doesn't have to change right now, just something I was thinking about when reviewing the code.

demiankatz:
Thanks for the progress, @xrosecky. Looks like there are some conflicts with the dev branch, likely related to the other recent Aleph PRs that were merged. Can you resolve those? I'll review again once the PR is in a mergable state.

demiankatz:
@xrosecky, looks like this needs a run of `vendor/bin/phing php-cs-fixer` to adjust some minor whitespace issues; that should get the build passing again, and then I can complete my review. Thanks for the continued progress!

demiankatz:
Thanks, @xrosecky. This looks good to me, though quite a few things have changed. Let me summarize them to confirm that they are all intentional and so that others can review if they are interested:

  - The barcode, dateFrom, dateTo and id fields have been removed from the getMyProfile response. None of these are standard fields expected by VuFind, so this should have no impact unless somebody has custom code that relies on them.
  - The expire field has been changed to expiration_date -- this switches a legacy non-standard field to the standard field introduced in release 4.1. Again, this should be fine unless somebody relied on this in custom code.
  - The city field has been added. This is harmless, and also possibly helpful.
  - Names are now extracted from the address-1 field, instead of the address-2 field. Because the old code both extracted names from address-2 AND used address-2 as the 'address1' field in the getMyProfile response, I suspect that this is actually a bug fix.

Does that summary match your intentions? If so, and if no one else objects, I think this can be merged. Thanks for the work!

xrosecky:
Yes, it matches my intentions.

demiankatz:
Thanks! I have @welblaud's approval as well, so I'm merging this now.

PR #1995, opened 2021-06-18
Fix ApiTrait jsonPrettyPrint value

aleksip:
Convert value to boolean, since PHP interprets all string values other than "" and "0" as true.


demiankatz:
Makes sense to me -- thanks!

PR #1996, opened 2021-06-18
Aleph driver: send user's language with REST DLF requests

xrosecky:


module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -355,6 +355,13 @@ class Aleph extends AbstractBase implements \Laminas\Log\LoggerAwareInterface,
      */
     protected $cacheManager;
 
+    /**
+     * Cache manager
```
demiankatz: I think this comment should say "Translator" :-)

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -482,10 +489,11 @@ class Aleph extends AbstractBase implements \Laminas\Log\LoggerAwareInterface,
      * @param \VuFind\Cache\Manager  $cacheManager  Cache manager (optional)
      */
```
demiankatz, line 501: Missing doc comment for new parameter here.

module/VuFind/src/VuFind/ILS/Driver/AlephFactory.php
```diff
@@ -66,7 +66,10 @@ public function __invoke(ContainerInterface $container, $requestedName,
         return parent::__invoke(
             $container,
             $requestedName,
-            [$container->get(\VuFind\Cache\Manager::class)]
+            [
+                $container->get(\VuFind\Cache\Manager::class),
+                $container->get('Laminas\Mvc\I18n\Translator')
```
demiankatz: I would suggest using `\Laminas\Mvc\I18n\Translator::class` here instead of the string; this will allow static analysis to behave more effectively.

demiankatz (changes requested):
This makes sense to me -- just a few minor stylistic comments. Once those are addressed, I'll ask for further input from other Aleph users.

demiankatz:
Thanks for the updates, @xrosecky! Looks like there are just a couple of outstanding style issues now -- most can be auto-fixed with `vendor/bin/phing phpcbf`, but there's one long line that may need to be reformatted. Beyond that, I'll check in with @welblaud to see if he has further input.

demiankatz:
@xrosecky, there was a typo that was causing tests to fail (I think a variable got accidentally double-pasted); I fixed it, and now this looks good to me and tests are passing. I'll just wait to hear from @welblaud before merging.

demiankatz:
@welblaud gives his approval, so I'm merging now. Thanks again!

PR #1997, opened 2021-06-21
Added a check if this plugin exists within the given namespace and a …

hajoseng:
…fallback to VuFind if not. This should fix the issue Helge Ahrens raised last week.


demiankatz:
Thanks, @hajoseng, this is great!

PR #1998, opened 2021-06-22
Fix class name to reflect the field name.

EreMaijala:
I had missed this one when changing the terminology.


demiankatz:
Thanks for catching that!

PR #1999, opened 2021-06-22
Avoid error in MultiBackend driver when a hold has no id field.

EreMaijala:
Since the id field is not mandatory, use item_id as fallback and handle gracefully the situation where neither exists. The error can be reproduced by setting 

```
[Catalog]
idsInMyResearch = false
```
in Demo driver ini when MultiBackend is enabled and trying to edit holds.


demiankatz:
Makes sense to me! Thanks!

PR #2000, opened 2021-06-23
Set preferred MARC field by configuration as well in getMarcReader()

hajoseng:
Method getMarcReader() still uses hardcoded fulltext-field as MARC field - different to getMarcRecord(). This is fixed by setting the preferred MARC field as a property.


demiankatz:
Thanks for catching this, @hajoseng -- it makes sense to me, but I'm also tagging @EreMaijala since this touches on code he has worked on, just in case he has any further suggestions. I just made one small adjustment (aef5184) which slightly simplifies the code to take advantage of the fact that the preferred field determination is now taking place in its own function.

EreMaijala:
@hajoseng @demiankatz Pull request #2000, time to celebrate! This makes sense, but I'd rather have the new method also return the MARC data (with a name like `getMarcData`).

hajoseng:
@EreMaijala @demiankatz What about the method getMarcRecord() which is declared deprecated and only used twice in SolrMarcRemote and SolrOverdrive? Is there a need of a method which returns the complete MARC record? If so should the MarcReader class populate the lazyMarcRecord property?

EreMaijala:
@hajoseng I meant that I'd keep the getMarcReader and getMarcRecord methods otherwise as is, but make them something like:

```
        if (null === $this->lazyMarcReader) {
            $this->lazyMarcReader
                = new $this->marcReaderClass($this->getMarcData());
        }
```
So the proposed getMarcData method would do what setMarcField here does, but would just return the raw MARC data from `$this->fields`.

demiankatz:
Yes, good point, @EreMaijala -- we could also call it getRawMarcData() and make it public if we wanted to, since it's exposing something fairly clearly-defined.

EreMaijala:
@demiankatz Good idea, I like that.

hajoseng:
Unfortunately push my changes is now restricted, because I don't know how to pull Demians changes; do you have any hints to solve this? // Got it.

demiankatz:
This looks great! I just adjusted the function comment, which was no longer accurate, in 38d65e4. Merging now.

(Sorry for the inconvenience caused by my earlier commit -- glad you were able to sort it out).

PR #2001, opened 2021-06-23
Make dates optional and escape values when editing a hold.

EreMaijala:
If both start date and required by date are editable, the validation would required both to be set (that'd be right if we were placing a hold). Skip validation if neither is set to keep the values optional. Also, user-entered values should be escaped.


demiankatz:
Makes sense to me!

PR #2002, opened 2021-06-23
Fix format and record-versions commas

crhallberg:
Fix these commas, which come from CSS:

image

TODO:
- [x] Run lessToSass when styles are finalized.


demiankatz:
Thanks, @crhallberg -- just looks like this needs a lessToSass in order to pass the build.

Also, exactly which case is `.record .format:not(.badge)::after` used for? I assume the screen shot above is what is fixed by `.record-versions .format::after`, but I wasn't sure what the other one was doing.

crhallberg:
The `:not(.badge)` is an additional case that would cover undiscovered combinations that would result in the same erroneous commas.

demiankatz:
@crhallberg, when I test this branch on my development box, I'm still seeing unwanted commas inside badges inside the Versions tab of the record view:

image

I'm still unclear about the `:not(.badge)` scenario, though -- this seems to be saying "do not display commas on the .format class if it is not a .badge" but that seems to be the opposite of what we want. Don't we want to omit commas when it IS a badge?

crhallberg:
Oops! I definitely mixed up my cases and made the selector the opposite of what it should be. Can you give this update a shot?

demiankatz:
Thanks for the adjustments, @crhallberg. It makes more sense now, but it still wasn't quite right. I moved these changes over to a copy of the release-7.1 branch and changed the base of this PR, since this is a bug fix that we probably want sooner rather than later. I then added 835b081, which adds a line to fix the real cause of the unwanted commas in the versions tab. I'm now uncertain whether or not the previous line that you added is actually necessary. Any thoughts? I can do some more testing and adjustment as needed, but wanted to get your opinion before investing more time.

PR #2005, opened 2021-06-29
SierraRest: Don't try to fetch items when there are no checkouts.

EreMaijala:
This fixes a PHP notice e.g. when trying to retrieve historic loans and there are none. Avoids an invalid API call as well.


demiankatz:
Makes sense to me; merging now!

PR #2006, opened 2021-06-30
League/Commonmark update to 1.6.5

xmorave2:
This PR updates league/commonmark dependency to version 1.6.5, updates MarkdownFactory according to https://commonmark.thephpleague.com/1.6/upgrading/ and refactors setting of extensions to be more flexible by removing hardcoded GithubFlavoredMarkdownExtension, but make the parts of github extension a default set.

### DIFF
--- a/config/vufind/markdown.ini
+++ b/config/vufind/markdown.ini
@@ -36,7 +36,7 @@
 ;   HeadingPermalink, Mention, SmartPunct, Strikethrough, Table, TableOfContents,
 ;   TaskList.
 ; More about extensions: https://commonmark.thephpleague.com/1.6/extensions/overview/
-; Some of them could have a configuration, see sections bellow.
+; Some of them could have a configuration, see sections below.
 extensions = Autolink,DisallowedRawHtml,Strikethrough,Table,TaskList
 
 ; See https://commonmark.thephpleague.com/1.6/extensions/external-links/
--- a/config/vufind/markdown.ini
+++ b/config/vufind/markdown.ini
@@ -1,6 +1,6 @@
 ; This file could be used to define configuration of markdown to HTML converter.
 ; More detailed configuration documentation could be found here:
-; https://github.com/thephpleague/commonmark/blob/latest/docs/1.5/configuration.md
+; https://commonmark.thephpleague.com/1.6/configuration/
 [Markdown]
 ; How to handle HTML input. Options are: strip, allow, escape. Defaults to strip
 ;html_input = allow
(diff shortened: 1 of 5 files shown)


config/vufind/markdown.ini
```diff
@@ -35,8 +35,11 @@
 ;     Attributes, ExternalLink, Footnote, HeadingPermalink, Mention, SmartPunct,
 ;     TableOfContents
 ; More about extensions: https://commonmark.thephpleague.com/1.5/extensions/overview/
```
demiankatz: Should we change the various comments and links that reference 1.5 to reference 1.6 instead?

config/vufind/markdown.ini
```diff
@@ -35,8 +35,11 @@
...
-; Some of them could have a configuration, see sections bellow
-;extensions = Attributes,ExternalLink,Footnote,HeadingPermalink,Mention,SmartPunct,TableOfContents
+; Some of them could have a configuration, see sections bellow. Available extensions
+; are: Attributes, Autolink, DisallowedRawHtml, ExternalLink, Footnote,
+; HeadingPermalink, Mention, SmartPunct, Strikethrough, Table, TableOfContents,
+; TaskList.
+extensions = Autolink,DisallowedRawHtml,Strikethrough,Table,TaskList
```
demiankatz, line 40: Now that we have a default list of extensions, do user who have customized this setting previously need to make any changes for things to work correctly? I doubt that this is being heavily used, but we should at least put a note in the changelog if some action may be necessary.

demiankatz: Also, if the setting is commented out, should we now default to this default set, or should we continue to default to an empty array in the factory code?

demiankatz (changes requested):
Thank you very much for this -- I was delaying upgrading this dependency because I suspected it would require some extra work, so I appreciate your initiative to take care of it before I had a chance to ask for help. :-)

In any case, just a couple of minor questions/suggestions regarding the configuration....

config/vufind/markdown.ini
xmorave2: A note in changelog would be fine. To continue thing work as before, this set of extensions should be enabled.

config/vufind/markdown.ini
xmorave2: Definitely, I updated the links and also fixed the list of extensions, which was partly duplicated...

config/vufind/markdown.ini
xmorave2: I made these extensions a default set if no extensions are set. To remove the default, and use raw commonmark, there is posibility to set the 'extensions' option to empty string

demiankatz: Note added -- thanks!

demiankatz:
@xmorave2, I see that you're still doing a bit of work on this -- please let me know when you're ready for it to be merged; I think we're nearly there!

config/vufind/markdown.ini
demiankatz: (And I revised my note to indicate that manual action is only needed for users who already customized the extensions setting in markdown.ini -- probably a very small number, if not zero). :-)

xmorave2:
> @xmorave2, I see that you're still doing a bit of work on this -- please let me know when you're ready for it to be merged; I think we're nearly there!

Thanks @demiankatz, I am done now, it is ready for your final look ;)

demiankatz:
Looks good to me -- thanks again!

PR #2007, opened 2021-06-30
Fix handling location level in XCNCIP2 ILS driver

xmorave2:
One more small enhancement of XCNCIP2 driver


demiankatz:
Makes sense to me -- thanks!

PR #2011, opened 2021-07-01
[VUFIND-1466] Improve Solr Connector HTTP client behavior

demiankatz:
TODO
- [x] Test everything.
- [x] Update changelog when merging (new constructor signature for Connector; removal of setAdapter/setProxy methods).
- [x] Resolve [VUFIND-1466](https://vufind.org/jira/browse/VUFIND-1466) when merging.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
+++ b/module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
@@ -365,20 +365,16 @@ abstract class AbstractSolrBackendFactory implements FactoryInterface
             array_push($handlers['select']['appends']['fq'], $filter);
         }
 
+        $httpService = $this->serviceLocator->get(\VuFindHttp\HttpService::class);
+        $client = $httpService->createClient();
+
         $connector = new $this->connectorClass(
-            $this->getSolrUrl(), new HandlerMap($handlers), $this->uniqueKey
+            $this->getSolrUrl(), new HandlerMap($handlers), $this->uniqueKey, $client
         );
-        $connector->setTimeout(
-            $config->Index->timeout ?? 30
-        );
+        $connector->setTimeout($config->Index->timeout ?? 30);
 
         if ($this->logger) {
             $connector->setLogger($this->logger);
-        }
-        if ($this->serviceLocator->has(\VuFindHttp\HttpService::class)) {
-            $connector->setProxy(
-                $this->serviceLocator->get(\VuFindHttp\HttpService::class)
-            );
         }
 
         if (!empty($searchConfig->SearchCache->adapter)) {
--- a/module/VuFind/src/VuFindTest/Feature/LiveSolrTrait.php
+++ b/module/VuFind/src/VuFindTest/Feature/LiveSolrTrait.php
@@ -66,6 +66,11 @@ trait LiveSolrTrait
             $container, $config['vufind']['config_reader']
         );
         $container->set(\VuFind\Config\PluginManager::class, $configManager);
+        $httpFactory = new \VuFind\Service\HttpServiceFactory();
+        $container->set(
+            \VuFindHttp\HttpService::class,
+            $httpFactory($container, \VuFindHttp\HttpService::class)
+        );
         $container->set(SearchSpecsReader::class, new SearchSpecsReader());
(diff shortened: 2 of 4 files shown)


demiankatz:
@EreMaijala, I haven't fully tested this yet because I wanted to have a conversation about it first. The goal here is to simplify the Solr connector, which is doing some weird proxy stuff that seems entirely unnecessary. The changes I've made so far focus on getting rid of the proxy logic but keeping other behavior intact. I think it might be best to give the Connector a hard dependency on the HTTP Service and get rid of all the internal client building and adapter overriding. In addition to simplifying the code, this will reduce the amount of HTTP code we'll have to change in the future if we move away from the laminas-http library.

I think the biggest reason to leave things alone would be if there was a use case where it was necessary to use a special adapter only in the context of the Solr connector... but even in that case, maybe adding a "setOverrideAdapter" method in place of "setAdapter" would make this more clear, and we could still get rid of the code that builds clients natively.

I think the original intent of this code was to be more VuFind-independent, but as we've discussed elsewhere, I don't know that that is really a realistic goal at this point anyway... if we wanted to extract VuFindSearch into a stand-alone library, we'll have some work to do no matter what.

I imagine @aleksip might also have some thoughts on this; if so, I'd be interested to hear them.

EreMaijala:
@demiankatz Reusability is a nice idea, but practically, I believe it's better not to try to do that. There's already Solarium for a PHP Solr library, and I don't see a reason to choose VuFindSearch over it unless, perhaps, you wanted to access same backends as VuFind does. As a matter of fact, at some point it might make sense to look into using Solarium in VuFind as well. It would already support e.g. the JSON APIs.

Do you think it would make sense to go one step further here and make HttpService a constructor parameter here? It could still default to null. And SetHttpAdapter could be left as the method to override/not require the service.

demiankatz:
@EreMaijala, I'm not familiar with Solarium, but it does sound like it might be worth considering, especially if laminas-http changes require significant reworking of our connector in the future.

Regarding making HttpService a constructor parameter, I agree that that would make sense as long as we agree on it becoming a mandatory dependency; that was one of my planned next steps if nobody objected to that approach. I'll wait and see if @aleksip or others have more input, but I plan to work on that next week unless somebody stops me. :-)

EreMaijala:
@demiankatz I'd vote for making it mandatory.

demiankatz:
@EreMaijala, since no one has objected, I've taken this all the way to the extreme of making the HttpService a mandatory constructor parameter. If users want to manipulate the HTTP adapter, it makes more sense to me to customize the factory to inject a custom-configured HTTP service rather than to customize the Connector class itself; I think it's probably better to just remove all of that potential complexity.

I still need to do a little more work to get the full test suite passing... but if you think I'm going in the wrong direction in the meantime, please let me know!

demiankatz:
All tests are now passing, so I think this is ready to merge with @EreMaijala's approval (though of course I'm still happy to take it in a different direction as needed).

EreMaijala (approved):
@demiankatz While this is okay as is, I just realized that most other Connector classes take a \Laminas\Http\Client instead of HttpService. I wonder if we should do that with Solr's Connector as well. We currently create a new client for each request attempt, but the client could also be `->reset()` instead. I don't think this makes a huge difference as long as we don't use a PSR-18 interface for the HTTP client, but this would provide a possibility to prepare for that and unify the connectors a bit.

demiankatz:
Thanks, @EreMaijala, that's an excellent suggestion -- it makes the constructor signature change significantly less disruptive, reduces the number of test classes that need changing, and generally feels cleaner. I've completely redone my changes to take this new approach; do you mind re-reviewing?

The only thing I'm not entirely happy about is the fact that the Laminas\Http\Client has no getOptions() method, which means (unless I'm missing something) that there is no easy way to find out the current timeout setting. If we could do that, then the Connector's getTimeout/setTimeout methods could just proxy the client... but for now, we're forced to store the timeout value in a property in the connector, and do some extra work in the constructor, just to be sure everything is consistent. Hopefully we can improve that in the future when we eventually replace laminas-http with something else.

I'm currently re-running the full test suite to be sure I haven't broken anything. I'll report back in a few minutes when that is done.

demiankatz:
All tests are passing!

EreMaijala (approved):
Looks very good now. :)

PR #2012, opened 2021-07-02
Run all tests on all PHP versions. Pin Composer version.

EreMaijala:
The composer version doesn't make difference at the moment, but I think it would be safer to use the known "good" version.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,11 +14,11 @@ jobs:
         php-version: ['7.3', '7.4', '8.0']
         include:
           - php-version: 7.3
-            phing_tasks: "phpunitfast"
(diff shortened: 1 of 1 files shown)


demiankatz:
Seems reasonable to me, but I'd suggest that we could perhaps at least omit eslint, jshint and checkLessToSass from two of the three versions, since these are Javascript and CSS-related checks and the PHP environment really shouldn't impact them at all. That might help reduce the build time at least a little bit.

EreMaijala:
@demiankatz Right, of course. I changed it so that the extended tests are run in the PHP 8 case as it's the quickets one to complete the other tests.

demiankatz:
Thanks, @EreMaijala, this looks good to me. I've just added a comment to explain the inconsistency in case anyone wonders... as long as that doesn't break the build, I'll merge this in a moment.

PR #2013, opened 2021-07-05
Add new ShowFacetExcept configurable filter option

hajoseng:
Extending the hide-facet-value feature by hiding all values except of those of a positive list of values to keep.

### DIFF
--- a/config/vufind/facets.ini
+++ b/config/vufind/facets.ini
@@ -303,5 +303,5 @@ visual_facets = "callnumber-first,topic_facet"
 
 ; Prevent all but specific facet values from being displayed to the user.
 ; Use facet field names as keys and untranslated facet values as values.
-[HideAllFacetValuesExcept]
+[ShowFacetValue]
 ;format[] = "Book"
--- a/config/vufind/facets.ini
+++ b/config/vufind/facets.ini
@@ -300,3 +300,8 @@ visual_facets = "callnumber-first,topic_facet"
 ; Use facet field names as keys and untranslated facet values as values.
 [HideFacetValue]
 ;format[] = "Book"
+
+; Prevent all but specific facet values from being displayed to the user.
+; Use facet field names as keys and untranslated facet values as values.
+[HideAllFacetValuesExcept]
+;format[] = "Book"
--- a/module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
+++ b/module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
@@ -482,14 +482,14 @@ abstract class AbstractSolrBackendFactory implements FactoryInterface
         BackendInterface $backend,
         Config $facet
     ) {
-        if (!isset($facet->HideFacetValue)
-            || ($facet->HideFacetValue->count()) == 0
-        ) {
-            return null;
-        }
+        $hideFacetValue = (isset($facet->HideFacetValue))
+            ? $facet->HideFacetValue->toArray() : [];
+        $hideAllFacetValuesExcept = (isset($facet->HideAllFacetValuesExcept))
+            ? $facet->HideAllFacetValuesExcept->toArray() : [];
         return new HideFacetValueListener(
             $backend,
-            $facet->HideFacetValue->toArray()
+            $hideFacetValue,
+            $hideAllFacetValuesExcept
         );
     }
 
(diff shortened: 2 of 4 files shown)


demiankatz:
Thanks, @hajoseng! This makes sense to me, but I've made a few minor adjustments to the comments, and I've expanded the test suite to use the new setting both by itself and in combination with the existing setting, to cover more possible scenarios (see commit 63055251217bd5494f1c385b0579d406b8ea7c93).

I think this is ready to merge now, but please let me know if you have questions or concerns about my changes. I'm also interested in whether @EreMaijala has any further comments.

hajoseng:
@demiankatz: Thank you very much; I've no concerns about your changes.

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
```diff
@@ -482,14 +482,14 @@ protected function getHideFacetValueListener(
...
     ) {
-        if (!isset($facet->HideFacetValue)
-            || ($facet->HideFacetValue->count()) == 0
-        ) {
-            return null;
-        }
+        $hideFacetValue = (isset($facet->HideFacetValue))
+            ? $facet->HideFacetValue->toArray() : [];
+        $hideAllFacetValuesExcept = (isset($facet->HideAllFacetValuesExcept))
+            ? $facet->HideAllFacetValuesExcept->toArray() : [];
```
demiankatz: We don't want to create the listener if it's not going to do any work, so I think there should be a check here that says:

```
if (empty($hideFacetValue) && empty($hideAllFacetValuesExcept)) {
    return null;
}
```

Do you have time to add and test that? Thanks!

demiankatz (changes requested):
Actually, I noticed one small issue as I reviewed the code again....

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
hajoseng: tested and added

demiankatz: Thank you!

config/vufind/facets.ini
```diff
@@ -300,3 +300,8 @@ visual_facets = "callnumber-first,topic_facet"
 ; Use facet field names as keys and untranslated facet values as values.
 [HideFacetValue]
 ;format[] = "Book"
+
+; Prevent all but specific facet values from being displayed to the user.
+; Use facet field names as keys and untranslated facet values as values.
+[HideAllFacetValuesExcept]
```
EreMaijala: I'd rather name this ShowFacetValue or FacetValueFilter. Same with the parameter and variable names.

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
```diff
@@ -482,14 +482,17 @@ protected function getHideFacetValueListener(
         BackendInterface $backend,
         Config $facet
     ) {
-        if (!isset($facet->HideFacetValue)
-            || ($facet->HideFacetValue->count()) == 0
-        ) {
+        $hideFacetValue = (isset($facet->HideFacetValue))
```
EreMaijala: unnecessary () around isset.

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
```diff
@@ -482,14 +482,17 @@ protected function getHideFacetValueListener(
...
+            ? $facet->HideFacetValue->toArray() : [];
+        $hideAllFacetValuesExcept = (isset($facet->HideAllFacetValuesExcept))
```
EreMaijala: unnecessary () around isset.

EreMaijala (changes requested):

config/vufind/facets.ini
hajoseng: Ok; I started with ShowFacetValue and renamed it afterwards because it's part of a "HideFacet" context and I felt it should have "hide" in it's name. Anyway I agree that ShowFacetValue sounds better and renamed it back.

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
hajoseng: done

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
hajoseng: done

demiankatz:
Thanks again, @hajoseng! I have updated the test suite to match the other wording changes, and I think this is now ready to merge -- doing so now.

PR #1729, opened 2020-09-25
Bug fix: AJAX facet calls must support all UrlQueryHelper features

demiankatz:
TODO
- [x] Test thoroughly
- [x] Investigate whether similar changes are needed for other facet-related AJAX calls (e.g. get tree)

PR #2015, opened 2021-07-06
[VUFIND-1434] Change UrlQueryHelper::setDefaultParameter override default to false

demiankatz:
TODO
- [x] Test everything
- [x] Note change in changelog
- [x] Resolve [VUFIND-1434](https://vufind.org/jira/browse/VUFIND-1434) when merging

### DIFF
--- a/module/VuFind/src/VuFind/Search/UrlQueryHelper.php
+++ b/module/VuFind/src/VuFind/Search/UrlQueryHelper.php
@@ -193,12 +193,11 @@ class UrlQueryHelper
      * @param string $name          Name of parameter
      * @param string $value         Value of parameter
      * @param bool   $forceOverride Force an override of the existing value, even if
-     * it was set in the incoming $urlParams in the constructor (defaults to true for
-     * backward compatibility)
+     * it was set in the incoming $urlParams in the constructor (defaults to false)
      *
      * @return UrlQueryHelper
      */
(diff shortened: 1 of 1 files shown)


demiankatz:
This is a follow-up to #1729, which was a bug fix introduced in 7.0.2, a time when we had to be extra careful about maintaining backward compatibility. Now that we're working on a new major release that allows backward compatibility breaks, it is a better time to switch to more reasonable/logical defaults. Unfortunately, I can't remember all of the details of that work at this point. It definitely seems to make more sense for the $forceOverride parameter to default to false; what I'm less sure about is whether any of the existing calls to setDefaultParameter now need to be changed to explicitly set the $forceOverride parameter to true, or if the new default behavior is equally appropriate. I'll start by running the full test suite to see if anything fails with just the initial change of the default, but I suspect there are some relevant edge cases that don't have test coverage, so this is going to require some additional hands-on testing (and perhaps the creation of some new tests). @EreMaijala, is this fresher in your mind than in mine? Do you have any thoughts? In any case, I'll report back with more details once tests finish running.

demiankatz:
All existing tests are passing... so the question is if there are edge cases that need to be covered.

EreMaijala (changes requested):
~~I think the "if it is not already defined." part of the method doc needs to be changed as well.~~ Oops, no, now it makes sense.

EreMaijala (approved):

PR #2017, opened 2021-07-06
[VUFIND-1426] Rename VuFind\Db\Table\Tags::getForUser

demiankatz:
Use the more accurate and descriptive getListTagsForUser() instead.

TODO
- [x] Test everything
- [x] Update changelog to note BC break
- [x] Resolve [VUFIND-1426](https://vufind.org/jira/browse/VUFIND-1426) when merging


demiankatz:
Because this method is only used in one place and is unlikely to be subject to customization, I don't think it's worth the effort of putting in a deprecated compatibility method as a stop-gap; that just creates more work down the road. However, feel free to disagree, @EreMaijala (or others). :-)

EreMaijala (approved):

PR #2019, opened 2021-07-06
[VUFIND-1007] MySQL 4-byte UTF-8 support

demiankatz:
TODO
- [x] Investigate whether any column widths need adjustment
- [x] Investigate whether changes are needed to the database upgrade tool
- [x] Decide how to handle text field upgrading
- [x] Decide how to handle key length limits
- [x] Test everything
- [x] Include notes in the changelog about this change (new encoding default; dropped support for Latin-1 legacy databases)
- [x] Resolve [VUFIND-1007](https://vufind.org/jira/browse/VUFIND-1007) when merging


demiankatz:
@mtrojan-ub / @EreMaijala, I've opened this pull request to start the ball rolling on utf8mb4 support in MySQL. My first commit is a simple string replace on the mysql.sql file; I haven't tested anything yet. I just thought it might be helpful to give us something to begin working on, and I'll make more progress as time permits (and report it here).

xmorave2:
There might be a need of reduce the varchar lenghts from 255 to 191 for some columns, see: https://stackoverflow.com/questions/6172798/mysql-varchar255-utf8-is-too-long-for-key-but-max-length-is-1000-bytes

EreMaijala:
@xmorave2 There was some discussion on this during yesterday's community call (see https://vufind.org/wiki/community_call:minutes20210706). The problem was identified, and it was proposed that we drop support for older MySQL versions that can't handle the longer indexes, but it's still up to further analysis whether it's possible and feasible to upgrade legacy databases. As far as I can see, the following are required for longer keys:
- InnoDB row format is dynamic
- innodb_large_prefix is enabled
- InnoDB file format is Barracuda (required for the above settings)

I believe that these requirements are fulfilled by MySQL 5.7.7 and later, but:
- I'm not sure about MariaDB versions (maybe need MariaDB 10.2.2 as minimum, see https://mariadb.com/kb/en/innodb-file-format/)
- I'm not sure about whether it's straight-forward enough to upgrade databases created with MySQL 5.5 or older MariaDB versions automatically.

I also considered the possibility of keeping the long key columns as utf8 while the tables otherwise would be utf8mb4, but that'd leave e.g. user.username with utf8, which is not desirable.

I'd like to avoid having to make the columns shorter if possible. We also have a compound key in the auth_hash table, and it would require making the hash even shorter.

EreMaijala:
As far as I can see `ALTER TABLE <tablename> ROW_FORMAT=DYNAMIC;` would also change the file format as necessary, though I haven't been able to verify this. But there's another option: limit the keys to shorter prefixes instead of modifying the columns. For example:

```
 alter table auth_hash drop key `hash_type`;
 alter table auth_hash add key `hash_type` (`hash`(140), `type`(50));
```

Side note: We don't really need the backticks everywhere, but I kept them for consistency.

The downside is that now there's a possibility that two identical hashes are inserted, but that should never happen in practice. For other tables I can't see a problem since we luckily don't use any of those long varchars as the primary key.

EreMaijala:
I'm working on fixing the database upgrade mechanism.

demiankatz:
Thanks, @EreMaijala! Feel free to push work here if that's the easiest thing to do.

EreMaijala:
@demiankatz I've added functionality to update keys, and also fixed some other issues that I discovered while testing. Note that the collation is now different from the previous one (utf8_general_ci is not a good default as it gets most non-ascii things wrong).

Unfortunately I've run out of time to do proper testing apart from verifying changes being made properly on one database, but this should be functionally complete now. Feel free to continue fixes, improvements and critique, everyone!

demiankatz:
Thanks, @EreMaijala, for the progress on this -- it looks great! I've made a couple of small adjustments to simplify character set extraction and to fix some flash messages that were showing up when they weren't supposed to (which were actually pre-existing problems that I noticed while testing this).

I did run into one significant issue, however: running the `CONVERT TO CHARACTER SET utf8mb4` operation actually changes the types of all of the text fields. MySQL upgrades all of the text fields to the next-largest type (e.g. mediumtext becomes largetext) in order to avoid the possibility of losing data based on the larger size of the utf8mb4 characters. This then causes the database to get out of sync with the mysql.sql file, and if you run the upgrade process twice in a row, it will make changes twice. It seems like we have two possible approaches:

1.) Restructure the database upgrade process so that it downgrades the text fields back to expected types after the character set conversion. Obviously, the disadvantage to this approach is the possible loss of data that the upgrade was designed to prevent.

2.) Change the text fields in mysql.sql to use larger types. If we do this, we need to be careful about order of operations to be sure that we don't upgrade the types based on the mysql.sql input, and then upgrade them AGAIN with the character set conversion. (I'm not sure if this is a risk or not, but it could be depending on how the code flows).

I suppose that a hybrid solution could be a mix of the two: upgrade some columns in mysql.sql where we think there is a legitimate risk of data loss, and leave others where they are... then make sure that the upgrade process gets everything into the right state by the time it's done. This is pretty tricky, and it also means that the output of the "SQL preview" mode (when you skip credentials) will not match the real behavior, since some of the changes will be related to side effects of prior queries. Ugly! We might at least need to pop up some kind of warning about that scenario.

I'm interested in whether @mtrojan-ub has any thoughts on this situation, or anything else related to the update.

In any case, regardless of the upgrade issue, I am currently running the full test suite on a clean instance created using the modified mysql.sql in this branch; I'll report back in a little bit with the results of that, though I don't expect anything to fail.

demiankatz:
As expected, all tests are passing -- so we just need to figure out a solution to the text field issue, and then I think we're in pretty good shape to merge this.

module/VuFind/sql/mysql.sql
```diff
@@ -227,17 +227,17 @@ CREATE TABLE `user` (
   `auth_method` varchar(50) DEFAULT NULL,
   `last_language` varchar(30) NOT NULL DEFAULT '',
   PRIMARY KEY (`id`),
-  UNIQUE KEY `username` (`username`),
-  UNIQUE KEY `cat_id` (`cat_id`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+  UNIQUE KEY `username` (`username`(190)),
```
mtrojan-ub, line 230: I'm worried about limiting UNIQUE keys to a specific amount of characters, because this might lead to side effects.
For example, if you have a user with a very long user name and another user would like to register with similar starting characters, this would not be possible because the DB would think  that the user already exists. So if there is a limit to a field you need a UNIQUE key on, the field length itself should be reduced to the max size of the key to avoid problems.
However, i can also understand that there will be existing installations for which we do not want to cause problems with existing users. And long user names should be the exception and not the regular case, so it might be worth taking that risk (or take the other way and remove the limit, but lose compatibility with old MySQL versions at the same time.)

mtrojan-ub:
Option 3) would be to use a MODIFY syntax instead of a CONVERT TO CHARACTER SET syntax to avoid data type changes, see also: https://dev.mysql.com/doc/refman/8.0/en/alter-table.html
I do not think this would lead to problems because the real existing text is only utf8mb3 at most so far, so there should be neither conversion problems nor data loss like in option 1).

module/VuFind/sql/mysql.sql
demiankatz: It's probably worth doing an experiment with creating 191-character usernames that only differ in the last character, and then running the database migration to see what happens. My suspicion is that the update will fail because MySQL won't let you create a unique key violation. If that is the case, I think we can probably just put a warning in the changelog so that people understand the situation if it occurs, and assume that this is extremely unlikely to happen in a real-world scenario. (Of course, we should also test that unique key violations are handled gracefully in the VuFind UI -- i.e. if a user tries to register with a duplicate key violation, they should get the "your username is already taken" type of message).

On the other hand, if the database upgrade DOESN'T fail in this scenario and instead creates some kind of problem/corruption, then we obviously need to change our approach and/or add some additional checks/validations to the upgrade logic.

In any case, I can probably find some time to test this next week... but if anyone else has time sooner, just let me know.

demiankatz:
@mtrojan-ub, I think I understand what you are saying: modifying a field from utf8mb3 to utf8mb4 won't actually change the size of the data, because the difference is not how the data is encoded, but rather which characters are allowed. So there are no existing utf8mb3 characters that will suddenly take up four bytes. If that's the case, then I agree that we can probably do a modify here instead. It would be a good idea to test conversion of utf8mb3 text (preferably a combination of 1-byte, 2-byte and 3-byte characters) that hits the limit of an existing text field, just to confirm this theory. Again, that is something I can try to find time for next week.

Thanks for the helpful feedback!

module/VuFind/sql/mysql.sql
demiankatz: I found some time today to experiment with this.

First, I applied a shorter unique key to the user table and experimented with creating users whose names only differ past the key length limit. Duplicate keys initially caused a fatal error, but I added 36fd982 to handle them more gracefully (treating them as duplicate usernames, as proposed above). I imagine the check for duplicate key exceptions is a bit fragile (and perhaps region-specific), so I gave it a separate method to make it easier to locally override, or make more complex in the core.

Next, I tried creating multiple long usernames in a utf8mb3 database, and then applying the upgrade process to bring that up to utf8mb4. On my first attempt, it "just worked" with no problems, which didn't make sense... until I realized that the upgrade process was not applying unique key changes correctly. I added commit 1212660 to fix this problem. Now, when I upgrade the database and create a key violation, the upgrade process fails with a reasonably clear message:

image

So, bottom line, I've fixed some problems and added some more robust error handling to the code, and I think these things are going to be useful regardless of our final decisions on other details... but the question is, what is our final decision?

Omitting the key length limit appeared to cause no problems for me when upgrading my database using MySQL 8.0.26. Is the limit related to a 5.x issue, or does the key length only become potentially problematic when usernames contain four-byte characters? Is there some other extreme scenario we should be testing?

Assuming that the key length change IS necessary for whatever reason, I'm not too concerned about it. I can't think of a realistic scenario where a username is longer than 190 characters AND it only differs from another existing username past character 190 (though please let me know if you think this really could happen). Even if that scenario happens, we can now handle it with some grace, either by treating it as a duplicate username or by reporting a problem during the upgrade process. If a user encounters the upgrade problem, they can either choose to edit mysql.sql to eliminate the key restriction (and deal with the possible consequences of that), or they can clean up their data and try again.

Any thoughts on how to proceed next? As you can see, I don't have particularly strong opinions one way or the other, but I'm also not as well-informed on all the details as others might be.

demiankatz:
@mtrojan-ub, on the subject of character set conversion: if I'm understanding [this documentation](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html#alter-table-character-set) correctly, if we don't want to use CONVERT TO CHARACTER SET, we instead need to run MODIFY commands on every individual column in the table. I think this means that `fixCharsetAndCollationProblems` would have to become significantly more complex, since we'd need to get a list of columns in the table, and only apply character set changes to the appropriate data types. Certainly doable, but I want to be sure we're on the same page that this is the right solution before I take the time to implement it. What do you think?

mtrojan-ub:
> ... we'd need to get a list of columns in the table, and only apply character set changes to the appropriate data type

Basically i think that's the correct approach.

> I think this means that fixCharsetAndCollationProblems would have to become significantly more complex

I don't know the code that well, but looking at methods like `getCharsetAndCollationProblemsForTable` and `getEncodingProblemsForTable`, it somehow seems like these methods basically have the same purpose. Maybe they should be merged somehow?

demiankatz:
@mtrojan-ub, you make a good point that the changes here make things a little more confusing. Originally, we had separate methods for dealing with character sets and collations... but @EreMaijala updated the collation method to also include character sets, while still leaving the separate character set method in place.

At this point, `getEncodingProblemsForTable` is designed to look specifically for latin1 encoding so that it can convert it to utf8 (or update config.ini to indicate that latin1 is desired). This is all related to a VuFind 1.x bug (see [VUFIND-294](https://vufind.org/jira/browse/VUFIND-294)). However, the revised `getCharsetAndCollationProblemsForTable` is probably going to undo or further change the work done by `getEncodingProblemsForTable`.

Perhaps we have a bit more work to do here. My proposal is that we drop support for latin1 MySQL databases. This feature was retained for compatibility with VuFind 1.x, and I think after more than a decade, there is really no reason to continue supporting this, since it adds significant complexity to the code with no real benefit. I think we should simply remove `getEncodingProblemsForTable` and related functionality, and adjust the code to throw an exception if the charset setting in the [Database] section of config.ini is set to latin1.

I think after that work is done, we can return to modifying the `getCharsetAndCollationProblemsForTable` and related routines to modify relevant columns instead of doing a table-level character set conversion.

This is quite a bit of work, so I want to be sure we're on the same page before I devote more time to it. What do you think? Would it make sense to do the latin1 deprecation as a separate PR in order to prevent this one from getting too complicated?

mtrojan-ub:
I don't see any particular reason why we should keep up latin1 support. So if you say that dropping support for this makes things easier => sure! Same thing for doing it in a separate PR first.

demiankatz:
Thanks, @mtrojan-ub -- I tried to do a separate PR to remove Latin-1 support, but the changes here are so entangled with the changes that would need to go there that I decided it was easier to do all the work in one place. The result is commit cae6644, which causes VuFind to throw an exception if database encoding is set to Latin-1, removes Latin-1 specific upgrade logic, and moves the useful parts of getEncodingProblemsForTable into getCharsetAndCollationProblemsForTable. Now there is just one method dealing with character sets and collations, and hopefully the whole process is a bit more clear again.

These changes allowed me to run MODIFY commands on the individual columns of tables that need collation changes, which in turn fixes the earlier problem we discussed with TEXT data types getting unexpectedly changed.

During the course of testing this, I discovered that there were a couple of multi-part keys which were getting re-converted every time I ran the database upgrade script. This was a combination of inconsistent whitespace (SHOW CREATE TABLE does not put spaces after commas in multi-part keys, but our mysql.sql script does) and inconsistent key length output (SHOW CREATE TABLE does not show a length limit in the key field if the limit is equal to the full length of the field... I adjusted our mysql.sql to match this in order to satisfy the logic).

I believe, with these changes in place, that this is now fully functional. The only question is whether the current approach to key lengths is acceptable, or if there's a better alternative.

For now, I'll wait until @EreMaijala is back and can take one more look. If the two of you are both satisfied with this, then I believe it is ready to be merged.

demiankatz:
If we don't make any further changes, I think we can check off the "test everything" box... but I'll leave it unchecked until the conversation is completed. The full test suite passes if I spin up a fresh instance using this branch. It also passes if I spin up a fresh instance using the existing dev branch, then switch to this one and run the upgrade process. I also confirmed, as discussed above, that if I run the database upgrade twice in a row, the second attempt does not change anything, because it recognizes that all changes have been successfully applied.

EreMaijala:
@demiankatz Thanks for picking this up from where I left it. I agree that the latin1 conversion can be removed. I also think that all the fixes and changes you've done make sense. As you noticed MySQL 8 wouldn't have any issues unless the database is inherited from an older version with settings that don't allow large keys. If it wasn't for legacy database configuration, we might be able to just say that you need at least MySQL 5.7.7 or MariaDB 10.2.2.

demiankatz:
Thanks, @EreMaijala! I believe this is complete, so I have gone ahead and merged it. If you, @mtrojan-ub, or anyone else, has any outstanding concerns about any of this, please let me know and they can be addressed by a subsequent PR.

PR #2020, opened 2021-07-07
Normalize display texts for "Favorites" and "Saved Items".

mtrojan-ub:
...as discussed in yesterday's community call. I wonder if it's necessary to normalize uppercase/lowercase too?

TODO
- [x] Run full test suite (and adjust tests as necessary) once wording is finalized.
- [ ] Add note to changelog when merging


languages/en.ini
```diff
@@ -390,7 +390,7 @@ Excerpt = "Excerpt"
 exclude_facet = "Exclude matching results"
 exclude_newspapers = "Exclude newspaper articles"
 Export = "Export"
-Export Favorites = "Export Favorites"
+Export Favorites = "Export Seleted Items"
```
demiankatz: I think this should just be "Export Items," because this is used as the title of the dialog box where the user makes the selection. Also note that, if you really feel we want the word "Selected," there is a typo here (missing "c"). :-)

demiankatz (changes requested):
Thanks, @mtrojan-ub, for getting this started. I had one small suggestion (see below). Also note that a number of these translations are also found in en-gb.ini and can likely be deleted now because the spelling is standardized when we remove the word "Favorites" / "Favourites".

We should also run the full test suite, because I think there's a good chance that some of the Mink tests have hard-coded text that may need to be updated. If you don't have time for that, I can work on it when time permits. I will add a TODO checkbox so we don't forget.

demiankatz:
Regarding case sensitivity, I think it probably makes sense to use lowercase except in page titles... which I think means everything can stay as it is, except add_to_favorites_html  should have "Saved Items" lowercased to "saved items." What do you think?

mtrojan-ub:
Thanks for your feedback! Everything is done now, please have another look when time permits!

demiankatz:
Thanks, @mtrojan-ub -- only one test was impacted by the change, and I have fixed it and pushed up the changes.

I believe that this is now ready to merge, but if @crhallberg, @EreMaijala or others have a moment to double-check, I'd be happy to hear more input. I'll wait a couple of days before merging, unless there's an immediate outpouring of support!

languages/en.ini
crhallberg: This string is only used as the [title for the cart](https://github.com/vufind-org/vufind/blob/1716403969bc00d0ad4bdaf9aa3602e790567ec8/themes/bootstrap3/templates/cart/export.phtml#L3) as far as I can tell, so I think Export Items would be more appropriate. Maybe we could change the key to Export Cart, actually. Unless I'm mistaken.

demiankatz: Since this PR is just changing values and not keys, I'm going to merge it as-is since we all seem to agree on the "Export Items" wording. If we want to change keys, we can deal with that in the future; I prefer not to without really strong reason, since that's more likely to break people's local customizations.

PR #2022, opened 2021-07-12
Add contentinfo role & aria label for screen readers in feedback form

tmikkonen:
When using a screen reader this enables them to inform the user of the possible help text when entering name/email field whichever comes first depending of the direction (up or down).

### DIFF
--- a/themes/bootstrap3/templates/feedback/form.phtml
+++ b/themes/bootstrap3/templates/feedback/form.phtml
@@ -73,7 +73,7 @@
       <?php if (in_array($el['type'], ['checkbox', 'radio'])): ?>
         <div class="field-set" role="group"<?= !empty($el['label']) ? ' aria-labelledby="' . $this->escapeHtmlAttr($el['name']) . '"' : ''?>>
       <?php else: ?>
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks for this! Just one question: if $elementHelpPre is empty, would it make more sense to omit the aria-label entirely, rather than to create an empty attribute? (It's entirely possible that it's fine as-is; I'm just curious). I'm also requesting a review from @crhallberg in case he has different suggestions.

tmikkonen:
> if $elementHelpPre is empty, would it make more sense to omit the aria-label entirely, rather than to create an empty attribute? (It's entirely possible that it's fine as-is; I'm just curious).

I did thought about it briefly earlier but as the W3C Recommendation for aria-label doesn't mention anything against the label being empty I kept the modification minimal.
https://www.w3.org/TR/wai-aria/#aria-label

crhallberg:
I agree that the aria-label should be omitted instead of empty. Empty aria-labels usually cause elements from being skipped over (such as a purely decorative image). You can see a compact way to do it a few lines above.

demiankatz:
Thanks for the feedback, @crhallberg and for the adjustments, @tmikkonen -- merging now!

PR #2026, opened 2021-07-13
corrects use of limitations in DAIA driver

olli-gold:
We have noticed, that the DAIA driver does not always handle limitations of services properly (after an update of our DAIA provider we got wrong availablities for many items). The reason for that mess was, that it could happen, that DAIA offers two services (for example "interloan" and "presentation"), while "presentation" does not have any limitations, but "interloan" has. As interloan is not respected by the DAIA driver, the availability for interloan does not show up in the DAIA response, but if it has a limitation, this limitation will be merged into the response of the driver (which might be confusing for the display, as a limitation might show up, while there is none for this service).

I have corrected that in this PR by checking, if the limitation service is matching the same whitelist of services as the availability check itself. Maybe @lahmann or @hajoseng can double check this PR for integrity before this could get merged.


olli-gold:
Finally made it a clean PR...
I had to change a test result to match the new pattern, so I would like to explain, why I think, that the expected result in https://github.com/vufind-org/vufind/blob/dev/module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/DAIATest.php#L100 is not correct. In https://github.com/vufind-org/vufind/blob/dev/module/VuFind/tests/fixtures/daia/response/daia.xml#L27 the service interloan has the limitation "Nur Kopie", which is supposed to be passed to the DAIA driver. But the DAIA driver does not respect the interloan service. As long as that's the case I would not expect the limitation of the interloan service to be in the DAIA response. 
Of course it would also be possible to add the interloan service to the driver (in that case I would expect to see the limitation in the DAIA response) - but this could mean, that we had to make a couple of more changes, for instance in the display logic of availabilities in the frontend...

demiankatz:
Thanks, @olli-gold! I'd definitely appreciate input from @lahmann or @hajoseng when time permits.

In the meantime, one thought: the `['loan', 'presentation']` array is used in many places through the code. Would it be more clear if we created a class property containing this value, so that if the service list needs to be customized or revised, it can be done in a single place rather than throughout the code base.

olli-gold:
Thanks for sharing that thought, @demiankatz!
I like the idea to put the services into an array, which probably can be easier to maintain. In DAIA, we have 5 services: loan, presentation, interloan, remote and openaccess. The first three services cover physical media, while "interloan" describes the conditions for an interlibrary loan and "presentation" and "loan" the conditions to use the media on site. "remote" and "openaccess" are addressing electronic media. Having said that, we could build the class property ONSITE for the often used array.

demiankatz:
Thanks, @olli-gold. I think it could make sense to have a `$this->onSiteServices` property which could take the place of the hard-coded arrays. It might even make sense to add an `isOnSiteService()` method to replace all the `in_array` checks, which might increase readability of the code.

I'm still waiting for feedback from @lahmann or @hajoseng before moving forward with the logic changes here (I think they are both on holiday right now), but let me know if you'd like to do any refactoring before this is merged, or if you'd prefer to tackle that in a separate PR, either now or later.

hajoseng:
Thank you very much @olli-gold for addressing this issue. I think the data-handling of the driver is a bit cumbersome. DAIA sets a limitation related to a specified service (like "loan" or "presentation"). The driver picks instead the first service which is marked as "available" and merges it with the first limitation in the answer. As far as I understand is this the source of the addressed error. Restricting limitations to the services "loan" and presentation" would skip limitations on the other services (eg "remote", see ppn 873184661 or 771400322). I think best would be not to break the structure of the answer of DAIA and keep the relations between service and limitation (and maybe other properites). An easier approach - though a bit unclean - would be to skip limitations on the service "interloan" which is anyway skipped within the driver. @olli-gold: What do you think about this thought?

hajoseng:
I dived a little deeper into this and realized that it's quite complex due to some inconsistencies in the translation process from DAIA-Services to holdings or storage retrieval informations. Moreover there have been some updates of the DAIA-ILS-connector for which the DAIA-driver could or should be adjusted. I need some more time for investigations and I plan to come back with one or two pull requests on those issues next and/or the following week. These will cover the correct use of limitations as well.

demiankatz:
Thanks, @hajoseng, I will look forward to seeing your work, and perhaps @olli-gold can help test/review, since I am not in a position to do so (I'll certainly review, but I can't test).

olli-gold:
Thanks for helping to work on this, @hajoseng! I agree, that handling the limitations in the DAIA driver is a bit cumbersome and probably a source of some problems. I do also agree on your idea to keep the structure of the DAIA response and to set the limitations related to the service. 
However, I still think, it makes sense to distinguish onSite services (loan and presentation) from the others - and so I guess we should basically stick with the practice of skipping certain services in the driver. Maybe it would be an idea to implement new methods to get availabilities (and limitations) for specific services only. But I do not have a specific idea about how to implement that yet.

Of course I am willing to review and test, once you have a PR addressing these questions. And let me know if I can help somehow on that.

hajoseng:
@olli-gold, there's one issue I'm pretty unsure about: Imagine an item having two services available: presentation and loan. The driver keeps in mind that these services are available but just overwrites href-attributes and merges limitations if they exist. Actually I haven't found examples in our catalogue where this could be "harmful". At least I've found examples both having an href-attribute but the same. As far as I understand the DAIA specification different attributes should be allowed as well - in which case the driver would just drop one of them. These services are also allowed to have different limitations - which are simply merged losing the information, which limitation belongs to which service. To put it in a nutshell: There is a loss of information and I'm not sure if DAIA provides too much information, so it's correct to throw some of it away, or if the holdings section processes too little information and should be enhanced.  

In my opinion best would be to distinguish between different services if they provide different informations (such as limitations). But, as VuFind shows only one status per item, this enhancement would not only effect the DAIA driver. @demiankatz, what do you think: Does it make sense enhancing the holdings-section to include such - maybe rare - cases of an item provided with a loan and a presentation service each with different limitations (or other significant informations)?

demiankatz:
@hajoseng, the ILS driver spec does include facilities for including various types of notes at both the item and holdings level (if I'm remembering correctly). I wonder if part of the question here is whether something currently single-valued needs to be made multi-valued, or if perhaps in some cases we can translate things into notes so they are visible to users (or whether that's not good enough, because we need things to be machine-readable/actionable).

Is that any help, or am I missing the point? :-)

olli-gold:
Continued in #2060

PR #2027, opened 2021-07-15
Aleph driver: record ID can contain bibliographic base even if only o…

xrosecky:
…ne base is configured


demiankatz:
Thanks, @xrosecky, this makes sense to me, but I'm going to ask @welblaud to also take a look since he's in a better position to actually test this.

demiankatz:
Merging with approval from @welblaud -- thanks to both of you! :-)

PR #2028, opened 2021-07-22
[VUFIND-1468] Normalize creator punctuation.

demiankatz:
TODO
- [x] Design configuration
- [x] Read configuration
- [x] Apply configuration
- [x] Test
  - [x] Regex application
  - [x] Paired character matching
  - [x] Abbreviation special case
  - [x] Initial special case
  - [x] Run full test suite
- [x] Resolve [VUFIND-1468](https://vufind.org/jira/browse/VUFIND-1468) when merging


demiankatz:
This is my first attempt at a solution to VUFIND-1468. It is heavily inspired by @EreMaijala's RecordManager work, though not identical -- I tried to move more of the behavior into the configuration to increase flexibility. It seems to work reasonably well for the test data included with VuFind, and for a variety of edge cases I tried throwing at it manually. I'd be interested to hear @mtrojan-ub's thoughts on this, particular from a performance perspective, since I know he has been working with the creator indexing code recently.

In order to make all of this work, I also had to make some adjustments/additions to the Java code for loading configurations. I added a TODO comment for improved comment handling when reading the .ini, since the solution I came up with to make things work here is not comprehensive... but I didn't have enough time to perfect a better regex-based solution.

I'm also not really sure that author-classification.ini is the best place for these configuration settings and am open to moving them elsewhere... but since this config is already used for creator indexing, it seemed marginally more efficient to expand it than to create a new one.

I welcome feedback, but I think this is, at very least, a step forward!

import/index_java/src/org/vufind/index/PunctuationContainer.java
```diff
@@ -0,0 +1,78 @@
...
+        new ThreadLocal<PunctuationContainer>()
+        {
+            @Override
+            protected PunctuationContainer initialValue()
+            {
+                return new PunctuationContainer();
+            }
+        };
+
+    private Set<String> punctuationRegEx = new LinkedHashSet<String>();
```
mtrojan-ub: If we store patterns here and use `name.replaceAll()` later, the patterns need to be compiled every single time.
For performance reasons it would be better to change the type from `Set<String>` to `Set<java.util.regex.Pattern>` and to call `Pattern.Compile(string)` for each entry only once when populating the set.

config/vufind/author-classification.ini
```diff
@@ -6,6 +6,34 @@ firstAuthorRoles = adp,aut,cmp,cre,dub,inv
...
 
+; When extracting roles, it may also be necessary to adjust punctuation on the
+; corresponding names. The following three sections contain settings related to
+; punctuation stripping.
+
+; This is a list of regular expression patterns to remove from names. These will
+; be applied in the order provided. In addition to the patterns specified here,
+; trailing periods will always be stripped (subject to abbreviation rules below).
+[PunctuationRegExToStrip]
+0 = "[ /:;,=(\\[]*$"
```
mtrojan-ub: Instead of hardcoding characters like spaces, it might be better to use character classes like `\s` to also match special kinds of unicode spaces that do not match a "normal" space.

It might be necessary to set the UNICODE_CHARACTER_CLASS flag when using / compiling patterns for this to work, depending on the Java version.

demiankatz:
Thanks for the review, @mtrojan-ub, this is very helpful; I'll try to find time to make these changes early next week, if not sooner.

import/index_java/src/org/vufind/index/PunctuationContainer.java
demiankatz: Good idea. Does 5eb3082  look good to you?

config/vufind/author-classification.ini
demiankatz: Excellent point -- I have addressed this in 20d2a67 (and confirmed that the UNICODE_CHARACTER_CLASS flag is indeed necessary to properly handle unicode spaces).

demiankatz:
I have made the suggested changes and will now re-run the test suite to be sure all is still well. I'll also leave this open until @EreMaijala has an opportunity to comment, in case he has any additional suggestions.

demiankatz:
All tests still passing!

import/index_java/src/org/vufind/index/PunctuationContainer.java
mtrojan-ub: Yes that looks better, thanks!

EreMaijala (approved):
@mtrojan-ub Those performance-related comments were great! I've nothing further, looks good to me.

demiankatz:
Thanks again, @EreMaijala and @mtrojan-ub -- merging now.

PR #2030, opened 2021-07-23
Fix checking of metadata emptyness in cover router

xmorave2:
Cover handlers getMetadata always rerurn array, never boolean false


demiankatz:
Thanks for the catch!

PR #2032, opened 2021-07-27
Focus to element after closing modal

LuomaJuha:
Small accessibility adjustment, where the user is returned to its previous element when the lightbox modal is closed.

### DIFF
--- a/themes/bootstrap3/js/lightbox.js
+++ b/themes/bootstrap3/js/lightbox.js
@@ -466,7 +466,8 @@ VuFind.register('lightbox', function Lightbox() {
       });
     });
   }
-
+  // Element which to focus after modal is closed
+  var beforeOpenElement;
   function reset() {
     _html(VuFind.translate('loading') + '...');
     _originalUrl = false;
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks, @LuomaJuha; this makes sense to me, but I'll wait for @crhallberg's review before merging, just to be sure we're not overlooking any details. He's out of the office right now but should be back next week.

crhallberg (approved):
This looks very good to me! The only thing that comes to mind is that I might add an underscore to the variable name since it's internal to the Lightbox and not used outside of it. That is a nitpick and should not hold this PR back! Thanks for your work!

demiankatz:
Thanks, @crhallberg -- I agree and have adjusted the variable name as you suggested. I also added an initialization to the variable declaration for the sake of formality. I've committed the modified version of the code as d9c71d2c1d18807f0d37b0c4355847794deb0598. (I don't have write access to this branch, so I was unable to update the pull request; I'm simply going to close this now since the logic has been merged). Thanks again, @LuomaJuha!

PR #2033, opened 2021-07-27
Aleph: support for administration identifier (adm_id) in getMyHolds

xmorave2:
Similar to the same change in getMyTransactions. See #1879


demiankatz:
Thanks! Makes sense to me to be consistent with the other PR.

PR #2034, opened 2021-07-28
[VUFIND-1458] Improve XML batch-loading support

demiankatz:
This pull request makes VuFind's XSLT more consistent and adds better support for batch-loading records.

TODO:
- [x] Replace ojs.xsl with ojs-multirecord.xsl (we don't need both, because the multirecord version handles both single and multi-record cases)
- [x] Add multi-record support to other existing examples
- [x] Test everything
- [x] Update changelog to note removal of ojs-multirecord.xsl
- [x] Re-record or annotate [XML Indexing video](https://vufind.org/wiki/videos:indexing_xml_records) to reflect improved multi-record handling
- [x] Close [VUFIND-1458](https://vufind.org/jira/browse/VUFIND-1458) when merging


demiankatz:
I have now updated all of the XSLT examples to support `<collection>` tags, with two exceptions: nlm_ojs.xsl, because this data format no longer appears to be used "in the wild," and so I couldn't find any examples to test with (I added a note about this fact to the .properties file, since this likely requires future work if anyone still has a related need); and vudl_FOXML.xsl, since that is long obsolete and should probably be removed (though for now I will leave it, since it may still be a useful example of a complex XSLT, even though it's no longer actively used by anyone).

demiankatz:
@xmorave2, I'm hoping somebody can give this a quick review before I merge it, and since you made an XSLT correction not too long ago, I thought you might be a good candidate. If you don't feel comfortable, though, let me know and I'll see if I can find another volunteer.

I would also be interested in feedback from @sbujam if he has any time to take a look, since he originated the ojs-multirecord.xsl that inspired all the rest of this work. (Also of note: I found a bug in the previous ojs-multirecord.xsl which was breaking the allfields indexing, and that has been corrected here).

import/xsl/dspace.xsl
```diff
@@ -19,7 +19,7 @@
                 <!-- ID -->
```
EreMaijala, line 33: Is there a reason for not including the collection handling here as is done in other files?

demiankatz: Good catch! An oversight on my part. Without the collection handling, this XSLT created a separate <add> element for each record, so I think it looked correct to me when I tested (some of the other XSLTs would mash everything together into a single record). However, it was missing the top-level collection tag, so it was definitely not right without this logic.

EreMaijala (approved):

import/xsl/dspace.xsl
EreMaijala: Ok, now it looks good to me!

demiankatz:
Thanks again, @EreMaijala!

PR #2036, opened 2021-07-28
[VUFIND-1483] Do not circumvent search service

demiankatz:
This PR is intended to make better use of the new command pattern behavior of the search service.

TODO
- [x] Update VuFind\Solr\Writer to use search service
- [x] Identify and fix other direct accesses to BackendManager
- [x] Finish subsidiary pull request (#2037)
- [x] Test everything (and fix broken tests)
- [x] Resolve [VUFIND-1483](https://vufind.org/jira/browse/VUFIND-1483) when merging

### DIFF
--- a/module/VuFind/src/VuFind/Record/FallbackLoader/Summon.php
+++ b/module/VuFind/src/VuFind/Record/FallbackLoader/Summon.php
@@ -29,8 +29,8 @@ namespace VuFind\Record\FallbackLoader;
 
 use SerialsSolutions\Summon\Laminas as Connector;
 use VuFind\Db\Table\Resource;
-use VuFindSearch\Backend\Summon\Backend;
 use VuFindSearch\ParamBag;
+use VuFindSearch\Service;
 
 /**
  * Summon record fallback loader
@@ -51,22 +51,22 @@ class Summon implements FallbackLoaderInterface
     protected $table;
 
     /**
-     * Summon backend
+     * Search service
      *
-     * @var Backend
+     * @var Service
      */
-    protected $backend;
+    protected $searchService;
 
     /**
      * Constructor
      *
-     * @param Resource $table   Resource database table object
-     * @param Backend  $backend Summon search backend
+     * @param Resource $table         Resource database table object
+     * @param Service  $searchService Search service
      */
-    public function __construct(Resource $table, Backend $backend)
+    public function __construct(Resource $table, Service $searchService)
     {
         $this->table = $table;
-        $this->backend = $backend;
+        $this->searchService = $searchService;
     }
 
     /**
@@ -105,7 +105,7 @@ class Summon implements FallbackLoaderInterface
                 $params = new ParamBag(
                     ['summonIdType' => Connector::IDENTIFIER_BOOKMARK]
                 );
-                return $this->backend->retrieve($bookmark, $params);
+                return $this->searchService->retrieve('Summon', $bookmark, $params);
             }
         }
         return new \VuFindSearch\Backend\Summon\Response\RecordCollection([]);
--- a/module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
@@ -73,13 +73,15 @@ class WriteDocumentCommand extends \VuFindSearch\Command\AbstractBase
      */
     public function execute(BackendInterface $backend): CommandInterface
     {
+        $this->validateBackend($backend);
         $connector = is_callable([$backend, 'getConnector'])
             ? $backend->getConnector() : null;
         if (!($connector instanceof Connector)) {
-            throw new \Exception('Unexpected connector: ' . get_class($connector));
-        }
-        if (!isset($this->context['doc'])) {
-            throw new \Exception('Missing document');
+            throw new \Exception(
+                $connector === null
+                    ? 'Connector not found'
+                    : 'Unexpected connector: ' . get_class($connector)
+            );
         }
         // If we have a custom timeout, remember the old timeout value and then
         // override it with a different one:
--- a/module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
@@ -28,10 +28,8 @@
  */
 namespace VuFindSearch\Backend\Solr\Command;
 
-use VuFindSearch\Backend\BackendInterface;
-use VuFindSearch\Backend\Solr\Connector;
+use VuFindSearch\Backend\Solr\Backend;
 use VuFindSearch\Backend\Solr\Document\DocumentInterface;
-use VuFindSearch\Command\CommandInterface;
 use VuFindSearch\ParamBag;
 
 /**
@@ -43,7 +41,7 @@ use VuFindSearch\ParamBag;
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org
  */
-class WriteDocumentCommand extends \VuFindSearch\Command\AbstractBase
+class WriteDocumentCommand extends \VuFindSearch\Command\CallMethodCommand
 {
     /**
      * Constructor.
@@ -61,63 +59,12 @@ class WriteDocumentCommand extends \VuFindSearch\Command\AbstractBase
         string $handler = 'update',
         ?ParamBag $params = null
     ) {
-        parent::__construct($backend, compact('doc', 'timeout', 'handler'), $params);
-    }
-
-    /**
-     * Execute command on backend.
-     *
-     * @param BackendInterface $backend Backend
-     *
-     * @return CommandInterface Command instance for method chaining
-     */
-    public function execute(BackendInterface $backend): CommandInterface
-    {
-        $this->validateBackend($backend);
-        $connector = is_callable([$backend, 'getConnector'])
-            ? $backend->getConnector() : null;
-        if (!($connector instanceof Connector)) {
-            throw new \Exception(
-                $connector === null
-                    ? 'Connector not found'
-                    : 'Unexpected connector: ' . get_class($connector)
-            );
-        }
-        // If we have a custom timeout, remember the old timeout value and then
-        // override it with a different one:
-        $oldTimeout = null;
-        if (is_int($this->context['timeout'] ?? null)) {
-            $oldTimeout = $connector->getTimeout();
-            $connector->setTimeout($this->context['timeout']);
-        }
-
-        // Write!
-        $connector->write(
-            $this->context['doc'],
-            $this->context['handler'] ?? 'update',
-            $this->params
+        parent::__construct(
+            $backend,
+            Backend::class,
+            'writeDocument',
+            [$doc, $timeout, $handler],
+            $params
         );
-
-        // Restore previous timeout value, if necessary:
-        if (null !== $oldTimeout) {
-            $connector->setTimeout($oldTimeout);
-        }
-
-        // Save the core name in the results in case the caller needs it.
-        return $this->finalizeExecution(['core' => $this->getCore($connector)]);
-    }
-
-    /**
-     * Extract the Solr core from a connector's URL.
-     *
-     * @param Connector $connector Solr connector
-     *
-     * @return string
-     */
-    protected function getCore(Connector $connector)
-    {
-        $url = rtrim($connector->getUrl(), '/');
-        $parts = explode('/', $url);
-        return array_pop($parts);
     }
 }
--- a/module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
@@ -1,0 +1,121 @@
+<?php
+
+/**
+ * Command to write a document object to Solr.
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2021.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Search
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org
+ */
+namespace VuFindSearch\Backend\Solr\Command;
+
+use VuFindSearch\Backend\BackendInterface;
+use VuFindSearch\Backend\Solr\Connector;
+use VuFindSearch\Backend\Solr\Document\DocumentInterface;
+use VuFindSearch\Command\CommandInterface;
+use VuFindSearch\ParamBag;
+
+/**
+ * Command to write a document object to Solr.
+ *
+ * @category VuFind
+ * @package  Search
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org
+ */
+class WriteDocumentCommand extends \VuFindSearch\Command\AbstractBase
+{
+    /**
+     * Constructor.
+     *
+     * @param string            $backend Search backend identifier
+     * @param DocumentInterface $doc     Document to write
+     * @param ?int              $timeout Timeout value (null for default)
+     * @param string            $handler Handler to use
+     * @param ?ParamBag         $params  Search backend parameters
+     */
+    public function __construct(
+        string $backend,
+        DocumentInterface $doc,
+        int $timeout = null,
+        string $handler = 'update',
+        ?ParamBag $params = null
+    ) {
+        parent::__construct($backend, compact('doc', 'timeout', 'handler'), $params);
+    }
+
+    /**
+     * Execute command on backend.
+     *
+     * @param BackendInterface $backend Backend
+     *
+     * @return CommandInterface Command instance for method chaining
+     */
+    public function execute(BackendInterface $backend): CommandInterface
+    {
+        $connector = is_callable([$backend, 'getConnector'])
+            ? $backend->getConnector() : null;
+        if (!($connector instanceof Connector)) {
+            throw new \Exception('Unexpected connector: ' . get_class($connector));
+        }
+        if (!isset($this->context['doc'])) {
+            throw new \Exception('Missing document');
+        }
+        // If we have a custom timeout, remember the old timeout value and then
+        // override it with a different one:
+        $oldTimeout = null;
+        if (is_int($this->context['timeout'] ?? null)) {
+            $oldTimeout = $connector->getTimeout();
+            $connector->setTimeout($this->context['timeout']);
+        }
+
+        // Write!
+        $connector->write(
+            $this->context['doc'],
+            $this->context['handler'] ?? 'update',
+            $this->params
+        );
+
+        // Restore previous timeout value, if necessary:
+        if (null !== $oldTimeout) {
+            $connector->setTimeout($oldTimeout);
+        }
+
+        // Save the core name in the results in case the caller needs it.
+        return $this->finalizeExecution(['core' => $this->getCore($connector)]);
+    }
+
+    /**
+     * Extract the Solr core from a connector's URL.
+     *
+     * @param Connector $connector Solr connector
+     *
+     * @return string
+     */
+    protected function getCore(Connector $connector)
+    {
+        $url = rtrim($connector->getUrl(), '/');
+        $parts = explode('/', $url);
+        return array_pop($parts);
+    }
+}
--- a/module/VuFindSearch/src/VuFindSearch/Command/GetQueryBuilderCommand.php
+++ b/module/VuFindSearch/src/VuFindSearch/Command/GetQueryBuilderCommand.php
@@ -1,0 +1,67 @@
+<?php
+
+/**
+ * Command to fetch a query builder object from a backend (if supported).
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2021.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Search
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org
+ */
+namespace VuFindSearch\Command;
+
+use VuFindSearch\Backend\BackendInterface;
+
+/**
+ * Command to fetch a query builder object from a backend (if supported).
+ *
+ * @category VuFind
+ * @package  Search
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org
+ */
+class GetQueryBuilderCommand extends AbstractBase
+{
+    /**
+     * Constructor.
+     *
+     * @param string $backend Search backend identifier
+     */
+    public function __construct(string $backend)
+    {
+        parent::__construct($backend, []);
+    }
+
+    /**
+     * Execute command on backend.
+     *
+     * @param BackendInterface $backend Backend
+     *
+     * @return CommandInterface Command instance for method chaining
+     */
+    public function execute(BackendInterface $backend): CommandInterface
+    {
+        $result = is_callable([$backend, 'getQueryBuilder'])
+            ? $backend->getQueryBuilder() : null;
+        return $this->finalizeExecution($result);
+    }
+}
--- a/module/VuFindSearch/src/VuFindSearch/Command/GetQueryBuilderCommand.php
+++ b/module/VuFindSearch/src/VuFindSearch/Command/GetQueryBuilderCommand.php
@@ -60,6 +60,7 @@ class GetQueryBuilderCommand extends AbstractBase
      */
     public function execute(BackendInterface $backend): CommandInterface
     {
+        $this->validateBackend($backend);
         $result = is_callable([$backend, 'getQueryBuilder'])
             ? $backend->getQueryBuilder() : null;
         return $this->finalizeExecution($result);
--- a/module/VuFind/src/VuFind/Record/FallbackLoader/SummonFactory.php
+++ b/module/VuFind/src/VuFind/Record/FallbackLoader/SummonFactory.php
@@ -66,10 +66,9 @@ class SummonFactory implements FactoryInterface
         if (!empty($options)) {
             throw new \Exception('Unexpected options passed to factory.');
         }
-        $backendManager = $container->get(\VuFind\Search\BackendManager::class);
         return new $requestedName(
             $container->get(\VuFind\Db\Table\PluginManager::class)->get('resource'),
-            $backendManager->get('Summon')
+            $container->get(\VuFindSearch\Service::class)
         );
     }
 }
--- a/module/VuFind/src/VuFind/Solr/Writer.php
+++ b/module/VuFind/src/VuFind/Solr/Writer.php
@@ -28,13 +28,13 @@
 namespace VuFind\Solr;
 
 use VuFind\Db\Table\ChangeTracker;
-use VuFind\Search\BackendManager;
-use VuFindSearch\Backend\Solr\Connector;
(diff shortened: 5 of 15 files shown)


demiankatz:
@aleksip, I have opened this PR to get the ball rolling on fixing situations that bypass the search service. So far, I have just updated the VuFind\Solr\Writer (and I've done a small amount of hands-on testing to confirm that it at least works for the basic commit, optimize and delete cases). I'll do more testing, convert more code, and fix broken tests as time permits... but if you have any thoughts in the meantime, I'd love to hear them, since I don't want to spend the time polishing everything if I'm not already on the right track. Thanks again for helping to push us in this direction!

demiankatz:
I had time to make a little more progress today; I fixed the broken tests for the Solr Writer, and I updated the SwitchQuery module to use the search service. Still lots more to do; I'll keep chipping away as time permits!

demiankatz:
I made some more progress today, including opening a separate PR (#2037) to deal with the sitemap generator, which got fairly complex and seemed better handled separately.

demiankatz:
@aleksip, I'm interested in your thoughts on b248d88; I changed the CallMethodCommand class to be concrete instead of abstract. It seemed unnecessary to build a separate command class to call a method that is only called in one place... but if you feel this is too undisciplined, I'm happy to revisit the solution.

demiankatz:
I believe that all BackendManager direct accesses have now been replaced with search service calls, between the work here and in #2037. I have tested all of the refactored code except for the Summon-related logic (since I no longer have access to a Summon instance), though some of this could certainly use more thorough exercising, and there are several opportunities for new unit tests.

I'm now going to pause work on this until others have an opportunity to offer feedback. Once that is received and responded to, I can do some more thorough testing and polishing before merging this.

demiankatz:
I've resolved merge conflicts in this branch, so it is now ready for more review/testing. As noted in my previous comment, I haven't thoroughly tested everything yet, so please review for general approach and be alert for mistakes. Once I'm sure there are no major revisions needed to my strategy here, I'll test more thoroughly.

module/VuFind/src/VuFind/Controller/Feature/AlphaBrowseTrait.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Proxy Backend::alphabeticBrowse using the CallMethodCommand.
+     *
+     * @return array
+     */
+    protected function alphabeticBrowse()
+    {
+        $service = $this->serviceLocator->get(\VuFindSearch\Service::class);
+        $command = new \VuFindSearch\Command\CallMethodCommand(
+            'Solr',
```
EreMaijala: Should the backend be configurable like in Channels?

module/VuFind/src/VuFind/GeoFeatures/Command/GetResultCoordinatesCommand.php
```diff
@@ -0,0 +1,76 @@
...
+        if (!($backend instanceof \VuFindSearch\Backend\Solr\Backend)) {
+            throw new \Exception('Unexpected backend: ' . get_class($backend));
+        }
+        $context = $this->getContext();
+        $queryBuilder = $backend->getQueryBuilder();
+        $params = $this->getSearchParameters();
+        $params->mergeWith($queryBuilder->build($context['searchQuery']));
+        $params->set('fl', 'id, ' . $context['geoField'] . ', title');
+        $params->set('wt', 'json');
+        $params->set('rows', '10000000'); // set to return all results
```
EreMaijala: Looks like something that should use a cursor mark like GetTreeDataCommand. I understand it's been this way before, but I still believe it should be refactored here.

EreMaijala (changes requested):
After glancing through this looks fine to me. I've only a general concern that might be just my lack of understanding of the Command pattern: How are situations where a backend can't handle the command handled? E.g. if someone sets `source = Primo` in channels.ini (I know the comment says to use a capable backend, but who knows which one is..), what's the result?

demiankatz:
Thanks, @EreMaijala -- your question about incompatible backends is a good one, and I will experiment and see what happens when I have time to test this more thoroughly (I might wait and see what @aleksip has to say first). Regarding your other comments, I think the ideas are good, but any changes to existing behavior/functionality should be done in separate PRs after this refactoring is done, just to prevent things from getting too confusing here. I'll be sure not to lose track of them!

module/VuFind/src/VuFind/ChannelProvider/AlphaBrowse.php
```diff
@@ -230,15 +228,16 @@ protected function summarizeBrowseDetails($details)
...
                     'thumbnail' => false, // TODO: better thumbnails!
                     'id' => $id
                 ];
             }
         }
         // If we have a cover router and a non-empty ID list, look up thumbnails:
         if ($this->coverRouter && !empty($ids)) {
-            $records = $this->searchService->retrieveBatch('Solr', $ids);
+            $records = $this->searchService
+                ->retrieveBatch($this->source, $ids);
```
aleksip: Should we be adding new code that uses deprecated methods?

module/VuFind/src/VuFind/Autocomplete/Eds.php
```diff
@@ -87,7 +90,16 @@ public function getSuggestions($query)
         $results = null;
         try {
             // Perform the autocomplete search:
-            $results = $this->backend->autocomplete($query, $this->domain);
+            $command = new \VuFindSearch\Command\CallMethodCommand(
```
aleksip: This is more of a general question than specific to this PR: I have been wondering about when to use fully qualified class names instead of imports and unqualified names? I am used to always using imports.

module/VuFind/src/VuFind/GeoFeatures/Command/GetResultCoordinatesCommand.php
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Command to fetch map coordinates for search results.
+ *
+ * @category VuFind
+ * @package  GeoFeatures
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org
+ */
+class GetResultCoordinatesCommand extends \VuFindSearch\Command\AbstractBase
```
aleksip: Not sure about this but I wonder if it would be good to extend from `CallMethodCommand` whenever possible? This would mean more consistent exceptions, and would provide another place for adding common functionality in addition to `AbstractBase`.

Edit: looks like this command calls several methods, so my comment does not apply here.

module/VuFind/src/VuFind/Record/FallbackLoader/Summon.php
```diff
@@ -105,7 +105,7 @@ protected function fetchSingleRecord($id)
                 $params = new ParamBag(
                     ['summonIdType' => Connector::IDENTIFIER_BOOKMARK]
                 );
-                return $this->backend->retrieve($bookmark, $params);
+                return $this->searchService->retrieve('Summon', $bookmark, $params);
```
aleksip, line 108: Use `RetrieveCommand` instead of deprecated method?

module/VuFind/src/VuFind/Autocomplete/Eds.php
```diff
@@ -87,7 +90,16 @@ public function getSuggestions($query)
...
+                $this->searchClassId,
+                Backend::class,
+                'autocomplete',
+                [$query, $this->domain],
+                null,
+                false
+            );
+            $this->searchService->invoke($command);
+            $results = $command->getResult();
```
aleksip: Method chaining could be used here and in the other numerous places where we are only interested in the results: `$results = $this->searchService->invoke($command)->getResult();`

module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
```diff
@@ -42,7 +42,7 @@
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org
  */
-abstract class CallMethodCommand extends AbstractBase
+class CallMethodCommand extends AbstractBase
```
aleksip: I guess I was thinking of a 1:1 relationship between methods and commands when I made this abstract. But it does not have to be so, of course. 

One downside to making this a concrete class is that if the same method is called by several different "anonymous" `CallMethodCommand`s, it will be harder to track the places where the method is called.

Also, if there are several specialized commands with a lot of extra logic calling the same method, it could become a bit unclear what should go in a command class and what should not.

In any case, I am not sure if these are real problems and am not objecting to this change.

aleksip:
> How are situations where a backend can't handle the command handled?

There isn't a common way of handling these situations. `CallMethodCommand` throws a `BackendException`. `BackendException`s are catched by `Service`, which will then trigger an error event for all possible listeners, and then it throws the error to the caller.

But a command can also do something else in this situation, e.g. return another type of result.

The caller can catch exceptions, and it could also use `CommandInterface::isExecuted()`.

aleksip:
Now that there are quite a few commands written, the way `execute()` implementations always need to end with `return parent::execute($backendInstance);` is starting to look slightly odd. Or is it just me? I have always considered it best practice to call the parent constructor or method as early as possible.

Also just noticed that `AbstractBase` is not actually declared abstract! 🤦‍♂️ 

Maybe `execute()` should be removed from `AbstractBase` and replaced with a `protected setExecuted(): CommandInterface` method?

module/VuFind/src/VuFind/Controller/Feature/AlphaBrowseTrait.php
demiankatz: There's not a use case yet, but it doesn't hurt to be flexible. I think a `$alphabrowseBackend` property in the trait might be the easiest way to manage this. Would you agree?

module/VuFind/src/VuFind/GeoFeatures/Command/GetResultCoordinatesCommand.php
demiankatz: Since this PR is meant to be focused on eliminating direct backend access, changing the logic doesn't seem appropriate in this context. Maybe the best course of action is to leave it as-is here but open a JIRA ticket to improve it as a separate step. What do you think?

EreMaijala: Agreed.

module/VuFind/src/VuFind/Controller/Feature/AlphaBrowseTrait.php
EreMaijala: Yes.

module/VuFind/src/VuFind/ChannelProvider/AlphaBrowse.php
demiankatz: It's not exactly new code, since it was already here -- I'm just changing how the backend parameter is provided. I do have the task of eliminating calls to deprecated backend methods captured as part of [VUFIND-1474](https://vufind.org/jira/browse/VUFIND-1474) (eliminate 8.x deprecations in release 9.0). Since there are already so many changes here, I'd prefer to stay focused on eliminating direct backend access in this PR, and I think the switch from wrapper methods to commands is probably better handled in a separate PR.

module/VuFind/src/VuFind/Autocomplete/Eds.php
demiankatz: There's no consistent rule for this, though it is something I have been thinking of. Generally speaking, if a class is only used once, it sometimes feels more clear and concise to me to use the fully qualified class name instead of an import. However, that's not a strong opinion, and doing things consistently may be preferable to saving a line of code here and there. I think as we add more static analysis, using imports more consistently may become more important. Ideally, it would be nice to have a php-cs-fixer rule to take care of this so we don't need to think about it. (I took a quick look at the Import section of the [available rules](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/doc/rules/index.rst) but didn't see anything that seemed like an easy solution... but maybe I didn't look closely enough). I suppose it's also possible that if php-cs-fixer can't help, phpstan or some other tool can.

module/VuFind/src/VuFind/GeoFeatures/Command/GetResultCoordinatesCommand.php
demiankatz: Yes, it does make sense to me to reuse things as much as possible, but as you say, that probably doesn't apply in this particular instance.

module/VuFind/src/VuFind/Record/FallbackLoader/Summon.php
demiankatz: At first I was going to repeat my "it's not really new code" argument from above here, but I guess in this case, it probably would make more sense to use a command. I'll change it next time I have a chance to work on this PR.

module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
demiankatz: As an alternative to making this non-abstract, what if we used anonymous subclasses to extend it? e.g. `new class(params) extends CallMethodCommand`. I haven't really worked with anonymous classes before, but it seems like a simple workaround if we want to keep the base class abstract, and I imagine this might make different instances easier to differentiate from each other. Might be worth experimenting with, just to see what backtraces look like when you do it this way. What do you think?

demiankatz:
> Now that there are quite a few commands written, the way `execute()` implementations always need to end with `return parent::execute($backendInstance);` is starting to look slightly odd. Or is it just me? I have always considered it best practice to call the parent constructor or method as early as possible.

I don't think there's necessarily a strict rule about calling the parent early -- I think it depends on whether the child method is doing pre-processing of parameters for the parent, or post-processing of output from the parent. In these instances, obviously the child is doing work that has to happen before the parent finalizes everything, so it seems reasonably natural to me.

That being said, I agree that having to call the parent's execute method doesn't feel right -- I just think the reason it feels wrong to me is probably unrelated to the order of operations.

> Also just noticed that `AbstractBase` is not actually declared abstract! 🤦‍♂️
> 
> Maybe `execute()` should be removed from `AbstractBase` and replaced with a `protected setExecuted(): CommandInterface` method?

I do like the idea of having an abstract `execute()` method, since that makes `AbstractBase` legitimately abstract and points out exactly what needs to be filled in to create a new command. The question is how best to make it completely clear to the author of a new command that they need to call `setExecuted()` at the end. I suppose it could go into the function's header comment. I also wonder if there's a better name than `setExecuted()` for the method, since it's really doing two unrelated things: setting a flag and returning a value. Would something like `finalizeExecution()` make more sense? Maybe I'm overthinking...

In any case, I have a few other projects to work through today, but I hope to spend more time on this PR early next week.

module/VuFind/src/VuFind/GeoFeatures/Command/GetResultCoordinatesCommand.php
aleksip: A problem with non backend method specific commands like this is that they too bypass the event based system.

This command for example is calling `$backend->getQueryBuilder()` when there is a `GetQueryBuilderCommand` and `$backend->getConnector()->search($params)` when there is a `SearchCommand`. So if I were to write a listener for `SearchCommand` calls, the search calls made by this command would be missed by the listener.

I wonder if this command for example could be made into a trait utilizing `GetQueryBuilderCommand` and `SearchCommand`?

And in general perhaps it would be safest to restrict commands to the `VuFindSearch` namespace, to keep all direct backend method access there?

module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
aleksip: I'm not sure if anonymous subclasses are an answer to the problem I'm thinking of: easy discovery of all usages of a specific backend method? It is much easier to find usages of a `RetrieveCommand`, for example, than to scan for non specific classes created with specific arguments.

Allowing concrete `CallMethodCommand`s or anonymous subclasses also allows for bypassing the event system, which might actually be a good reason to not allow them?

module/VuFind/src/VuFind/GeoFeatures/Command/GetResultCoordinatesCommand.php
demiankatz: Very good points, and this does get tricky. I really want to isolate extremely specialized logic from the backend, so I find the idea of building another trait into the backend itself doesn't particularly appeal to me; it doesn't feel like it scales well in the long run... but obviously putting too much logic in the command makes it extremely easy to be undisciplined and miss situations where the search system should be utilized... and as I think I've said elsewhere, having a command that calls the search system repeatedly feels a little weird to me, though I'm not yet convinced that it's an entirely bad thing. When I get back to working on this PR, I'll look at this example in particular and see if a favorable solution stands out for me. And if I'm misunderstanding your trait suggestion, and you're not suggesting just building this logic directly into the backend, then please correct me. :-)

module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
demiankatz: Yes, all good points. I think I'll just go ahead and make concrete classes for everything; I was trying to cut down on the number of files created, but adding some more files is really not a huge price to pay for the benefits you point out.

aleksip:
> I don't think there's necessarily a strict rule about calling the parent early

Indeed, the pre-processing example you gave is a good example of a late parent call. I guess it is just that the current implementation does not really give any other option is what makes it feel a bit strange. Also that there really isn't anything related to actual command execution going on in the parent! 😄

> I do like the idea of having an abstract execute() method

Now that you said that I like the idea too! 🙂 

> Would something like finalizeExecution() make more sense?

I think that would work!

module/VuFind/src/VuFind/GeoFeatures/Command/GetResultCoordinatesCommand.php
aleksip: I was thinking the trait could go in `VuFind\GeoFeatures`.

demiankatz: What class were you envisioning to use the trait? I was thinking you were proposing adding a trait to the Solr backend, but if you refer to something else, I may just be confused. :-)

aleksip: It looks like `GetResultCoordinatesCommand` is currently only used once? In that case the logic might not even need to go into a trait, but could perhaps stay in `MapSelection::getSearchResultCoordinates()`?

demiankatz:
See #2080 for an implementation of the above-discussed changes. I'll incorporate them here once that is merged.

module/VuFind/src/VuFind/GeoFeatures/Command/GetResultCoordinatesCommand.php
demiankatz: Ahh, yes, of course, I understand what you are saying now. I will look into this when time permits!

demiankatz: I think the key issue here is that on my first pass of refactoring, my goal was simply to find code that made backend accesses and encapsulate it inside commands... but that's not really correct in many cases; rather than making big, complex commands, each individual backend access should be a separate command.

module/VuFind/src/VuFind/Record/FallbackLoader/Summon.php
aleksip: I've just realized we should only stop using the deprecated service method calls in 9.0, because there might be code relying on the old style events. So even new code should use them until 9.0 instead of commands. Sorry about causing confusion here.

aleksip: And if we wanted to be truly fully compatible even with new commands (not replacing deprecated `Service` methods) I guess they too should trigger legacy style events until 9.0...?

demiankatz: Ahh, thanks for remembering. That also applies to my work in #2037 -- I've just committed a small fix (83216014509000dca39ccd27acd883aeab8ed9ac) to make sure the sitemap generator still uses the deprecated method (for now).

demiankatz: I don't think it's necessary to trigger legacy-style events for new commands; in 8.0 we are announcing that people should be updating code to use the new style of events in preparation for 9.0. If they are writing new code for new events, they should use the new style, not the deprecated style. Or am I missing a larger implication?

aleksip: The event target changes, which unfortunately cannot be avoided if we want to remove the possibility of direct backend access.

All `VuFind\Search` event listeners receive all events, and if they attempt to call a `BackendInterface` method on the target, that would cause an exception with the new events, I believe. Not sure how likely is is to have this kind of code out there. At least most of our code follows the pattern

```php
$backend = $event->getTarget();
if ($backend === $this->backend) {
```

which would not be a problem even if the target is a command object.

demiankatz: Yes, that makes sense. My gut feeling is that the risk here is relatively low, and it's probably better to deal with problems as they arise rather than do a lot of work that we'll almost immediately have to undo anyway... but if you feel otherwise (or if others do), please let me know.

demiankatz:
Because this PR has a lot going on, I've decided to break it into smaller pieces for ease of review, and so we can get parts of it merged before all the work is done. I've started with #2084 and #2085, both of which I believe are ready to be merged.

demiankatz:
See also #2086 for more work split off from here. I'm going to pause on this until those three PRs can be merged, and then I will resume work.

module/VuFind/src/VuFind/Record/FallbackLoader/Summon.php
aleksip: I agree, and in any case 8.0 is a major version change so breakage is allowed, even if we are trying our best to avoid it until 9.0. Maybe a short note regarding possible code that handles the target as a `BackendInterface` instance could be added to the release notes?

aleksip: And I guess there might also be breakage if the code expects that some parameters exist, so maybe the note should be something along the lines that "much effort has been made to ensure that in most cases there is no breakage, but best to check your listener code just to be sure".

demiankatz: Here is the current changelog note:

The VuFindSearch\Service has been refactored to use the Command Pattern; you can now send command objects to the service's invoke method instead of calling more specific search/retrieve/etc. methods. These command objects are also shared with search listeners. The command system makes it easier to customize search behavior without having to rewrite the search service itself. This has been implemented in a mostly-backward-compatible way, but the old way of doing things is deprecated in this release and will be removed in VuFind 9.0. If you make calls to the search service in your code, and/or if you have written any custom search listeners, you should update your code to use command objects instead of the old techniques. When using command objects, some event details will be different (some arguments may not be present, and the event target will also be different). See pull request #1967 for details and examples.

Feel free to edit this directly if you want to improve it.

I also think it may be a good idea to include some content about the command pattern refactoring in this year's VuFind Summit, to make sure the community is aware of this and so that we have a recording we can share in the future. I'd be happy to collaborate with you on such a thing if you might be interested.

demiankatz:
@aleksip, I plan to run the full test suite one more time after our conversation is complete, but I believe that this is ready for another review. I've broken off the remaining more complex pieces into pull requests #2090, #2091, #2092 and #2093. The first three of these are ready for review, and I plan to revise #2093 based on the outcome of #2092. Everything that remains here is relatively simple and straightforward.

The only point that I'm not entirely sure about is whether GetLuceneHelperCommand is justified, as that is providing access to an implementation detail of the backend, which I think is generally undesirable. The question is whether it's better to write commands to perform specific analysis checks/tasks, or if that's overkill for this fairly limited use case. I'm open to revising if you think it's worth the effort, but I'm being lazy as the default state. :-)

Anyway, thanks again for all the help discussing and reviewing this code. I think we're nearing the end now!

module/VuFind/src/VuFind/GeoFeatures/Command/GetResultCoordinatesCommand.php
demiankatz: I've moved this logic to #2092 and put a TODO checkbox there to open a JIRA ticket about cursor mark support after the other work there is completed.

hajoseng:
@demiankatz I think the LuceneSyntaxHelper includes methods which address more general sanity checks and corrections as well as quite lucene-specific methods. Maybe these two aspects should be separated into a solr-specific class and another one doing general syntax checks and corrections. I'm not sure if this could make the code easier to follow; at least it could increase the use of those general query analysis (as checking brackets, quotes, case-sensitivity issues ...) and therefore legitimate the use of a command. If you think this could be useful I would come up with a suggestion next week.

demiankatz:
@hajoseng, it certainly might make sense to build something like an AbstractSyntaxHelper which might contain some general-purpose checks that are not Lucene-specific, for which LuceneSyntaxHelper could be a child. I'm not sure if this is worth doing before there is a use case for a second child class, though, since it's sometimes easier to do this type of refactoring when there are multiple use cases to consider.

From my perspective, the biggest problem with the LuceneSyntaxHelper is that it is currently serving two interrelated purposes: it is checking syntax, and it is acting as a container for configuration settings related to syntax checking. Maybe that's not actually a problem, but it does feel a bit like a violation of the single responsibility principle. I don't really have an idea that makes it more elegant, though -- if we move the settings to a separate container object, then we still have to either inject that container into the helper, or else make the helper's functions take more parameters and build a third class to encapsulate the first two. This might be better in terms of theoretical design principles, but it just seems unnecessarily complicated from a pragmatic perspective.

If you have different ideas along these lines, I'd certainly be interested to hear them, though. In the meantime, maybe it is just more elegant to create a CheckQueryForBooleansCommand and GetBackendQuerySettingsCommand (which could return an array of configurations from the backend, for greater flexibility) to meet the current needs of the SwitchQuery module. Still interested in hearing @aleksip's thoughts as well.

module/VuFind/src/VuFind/Record/FallbackLoader/Summon.php
aleksip: The note looks perfect to me. Will get back to you on Slack re VuFind Summit.

demiankatz:
@aleksip, I've split off the SwitchQuery recommendation module to #2106 since there was still some ongoing discussion about that (which I will try to address very soon). For everything left here, I've written tests and double-checked that it works (with the exception of the Summon fallback loader, which I cannot test "for real" -- though the unit test seems okay). Do you have any last thoughts on any of this before I merge it?

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * @param BackendInterface $backend Backend
+     *
+     * @return CommandInterface Command instance for method chaining
+     */
+    public function execute(BackendInterface $backend): CommandInterface
+    {
+        $this->validateBackend($backend);
+        $connector = is_callable([$backend, 'getConnector'])
+            ? $backend->getConnector() : null;
```
aleksip: Although it would not be as compact and efficient as this code here, I wonder if it would make sense to have an `AbstractBase::validateMethod()` or similar as well? But this is not something that needs to be done before 8.0.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Execute command on backend.
+     *
+     * @param BackendInterface $backend Backend
+     *
+     * @return CommandInterface Command instance for method chaining
+     */
+    public function execute(BackendInterface $backend): CommandInterface
+    {
```
aleksip: Could this code not go into a `Backend::writeDocument()` method?

module/VuFindSearch/src/VuFindSearch/Command/GetQueryBuilderCommand.php
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Command to fetch a query builder object from a backend (if supported).
+ *
+ * @category VuFind
+ * @package  Search
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org
+ */
+class GetQueryBuilderCommand extends AbstractBase
```
aleksip, line 42: I can't seem to find where this command is used apart from the tests? 

Do we want to expose the query builder if we are not exposing the connector?

demiankatz: It is used in the DevToolsController; that's an odd special case because it is a development-oriented tool specifically designed to expose the internal behavior of the backend. (You can paste in a minified search query from VuFind's search table, and it will dump out a whole bunch of useful information about it, including the output of the query builder).

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
demiankatz: What signature/behavior did you have in mind? I'm not sure if I follow exactly what you have in mind.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
demiankatz: True, I'll refactor.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
aleksip: I don't think I had anything specific in mind 😄, but if you are refactoring this in any case then we can just forget about this...

demiankatz: Good point!

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Command/WriteDocumentCommand.php
demiankatz: Okay, this is done in 56766dc (and f330f42, since I inadvertently broke some tests). I think it's much better now! Do you mind taking one last look? If you're satisfied now, I think this refactoring project is complete (unless there's more to do with #1886 before the 8.0 release).

demiankatz: (And I see now that you're proposing we wait on #1886, so that's one less thing to worry about this week :-) ).

aleksip (approved):
Looks good to me!

PR #2037, opened 2021-07-29
Refactor sitemap generator to use search service more consistently.

demiankatz:
This contains sitemap-specific refactoring related to the work in #2036. I separated it out here since it is a substantial change that might require discussion and revision in its own right, and I didn't want it to get mixed up with too many other unrelated changes.

TODO
- [x] Create support method in Generator class to generate command object (for easier extensibility/overriding)
- [x] Rename command classes to something better
- [x] Resolve TODO related to file naming of index-based files
- [x] Improve processing of plugin configuration (and mandatory inclusion of Index plugin)
- [x] Test that verbose messages appear as expected
- [x] Test that breaking sitemaps into chunks works as expected
- [x] Change public generator methods back to protected where possible
- [x] Clean up style issues
- [x] Test thoroughly


demiankatz:
This is a first pass, and I haven't thoroughly tested it, so I'm sure more work is needed -- but I wanted to get feedback on the general approach before I spend a lot of time polishing.

The idea here is to take advantage of the new search service command pattern functionality to eliminate the need to pass the BackendManager to the Sitemap Generator. Instead, the Sitemap Generator sends commands directly to the search service.

This has some obvious advantages:

  * The Sitemap Generator code was huge, and this breaks it into smaller chunks that can be independently tested and managed
  * The Sitemap Generator code had a lot of conditional branches to behave differently based on the selected data retrieval method (terms vs. cursors). The new code uses a different command object for each method, and a common abstract base class to store the shared behavior. This feels a lot more elegant.

This also has some potential flaws and open questions:

  * It's not 100% clear to me which values make sense in the context variable, and which make sense being passed directly to the command's constructor. For now, I put almost everything into the context but passed the search service directly, since I'm certain we don't want listeners manipulating the search service, but perhaps some of the other values might be up for grabs.
  * The command needs access to some functionality of the generator. For now, I made some generator methods public and changed the code to pass the generator into the command as part of its context. This may not be the most elegant approach.
  * I feel a little weird about the fact that in some situations, the command is triggered by the search service, but then makes additional calls to the search service. I think this should work, and maybe it's not a problem... but perhaps something else to revisit.

I have not extensively tested this, but I can confirm that unit tests are passing, and when I run the command-line generator tool, it does create output (though I haven't checked that for correctness, only that it looks valid at a glance).

I think with this refactoring, we can more easily increase our unit test coverage on the generator code since we can approach each of the new classes independently. I also plan to do more thorough hands-on testing after the general approach is finalized.

@EreMaijala or @aleksip, any thoughts?

EreMaijala (changes requested):
Looks good and works fine when Solr is running, but if Solr is down, the error handling doesn't seem to be working correctly. Instead I get the following error:

`Fatal error: Uncaught TypeError: VuFind\Search\Solr\AbstractErrorListener::listenForBackend(): Argument #1 ($backend) must be of type VuFindSearch\Backend\BackendInterface, null given, called in /Users/eremaijala/src/VuFind2/module/VuFind/src/VuFind/Search/Solr/V3/ErrorListener.php on line 58 and defined in /Users/eremaijala/src/VuFind2/module/VuFind/src/VuFind/Search/Solr/AbstractErrorListener.php:96`

A question: during last round of sitemap refactoring the idea of making index-based sitemap generators plugins like the ContentPages. Do you think it would make sense to do here?

demiankatz:
@EreMaijala, I believe the error you mentioned is now addressed by #2066.

As far as refactoring the index-based generation to a plugin, that is a very appealing idea -- I'll see if I have time to take a stab at it.

demiankatz:
@EreMaijala, refactoring to a plugin seems to have worked very nicely. I only had enough time to take a first stab at it, and there's a lot more polishing that needs to happen... but I think this is another major step forward in clarifying and simplifying the code. I'll try to find time to work on this more next week. In the meantime, I've created lots of additional TODO checkboxes to remind me of outstanding issues.

demiankatz:
@EreMaijala, I couldn't wait until next week to finish this up. :-) I think I've polished and tested everything, and I'm very happy with how it turned out. I think it would now be very feasible to get unit test coverage for everything, and I may try to work on that next week if time permits. I'll wait and see if you have any further suggestions first, though!

module/VuFind/src/VuFind/Sitemap/Generator.php
```diff
@@ -347,62 +275,55 @@ protected function generateWithPlugins(): array
...
-                                'frequency' => $frequency,
-                            ]
-                        );
-                    } else {
-                        $sitemap->addUrl($url);
-                    }
-                }
-                // Unset the reference:
-                unset($sitemap);
+        $mandatoryPlugins = ['Index'];
```
EreMaijala: How about making this `$defaultPlugins` that's only used if `$this->config->Sitemap->plugins` isn't set, and include 'Index' in `plugins[]'  in sitemap.ini? This would require a config change for sites currently using plugins, but would also make it possible to define the order the plugins are executed in the sitemap.ini file, or disable Index plugin altogether.

If you don't think that's a good idea, I'd at least change the default order so that Index plugin is processed last as was previously the case (this would put the index page etc. at the beginning of the sitemap set), and make it so that if it's defined in `plugins[]`, that will override the default order and not run Index plugin twice as the current code would.

module/VuFind/src/VuFind/Sitemap/Plugin/IndexFactory.php
```diff
@@ -0,0 +1,117 @@
...
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options passed to factory.');
+        }
+        $configLoader = $container->get(\VuFind\Config\PluginManager::class);
+        $sitemapConfig = $configLoader->get('sitemap');
```
EreMaijala, line 70: I'm slightly concerned that the Index plugin is not completely controlled by the generator but reads the config on its own. This makes it less controllable than I'd find ideal, but since I can't come up with a real world example of where this would be an issue, and I also don't can't come up with a really clean way of providing the necessary config from the Generator, I think it's okay.

EreMaijala (changes requested):
All in all this looks awesome, and makes things much more clear. :)

module/VuFind/src/VuFind/Sitemap/Generator.php
demiankatz: I was worried that this would cause backward-incompatibility headaches with #1904, but since #1904 actually hasn't been released yet (it's new to 8.0), this actually is no problem at all -- a warning note to vufind-tech should more than cover it. I'll make the change as soon as time permits!

module/VuFind/src/VuFind/Sitemap/Plugin/IndexFactory.php
demiankatz: I thought it was actually a benefit to separate concerns here, and allow the Generator to know as little as possible about the plugin. This allows custom configuration to be injected with a custom factory, instead of having to modify the Generator... and it wouldn't be too hard to have multiple subclasses with multiple factories if there were somehow a need to have multiple instances of the Index plugin running with different configurations.

All that being said, if you do think of a reason to change the approach, I'm always open to further discussion.

EreMaijala: Haha, I see! And that's valid reasoning as well, so no further objections..

module/VuFind/src/VuFind/Sitemap/Generator.php
demiankatz: Okay, I took care of this in 70975c3 -- if you have no further suggestions, I'll see if I can find time to write a couple of unit tests, and then will merge after that.

EreMaijala (approved):
No further issues, looks good to me. :)

module/VuFind/src/VuFind/Sitemap/Command/GetIdsWithCursorMarkCommand.php
```diff
@@ -0,0 +1,124 @@
...
+     */
+    protected function getIdsFromBackend(
+        Backend $backend,
+        string $cursorMark,
+        int $countPerPage
+    ): array {
+        // If the previous cursor mark matches the current one, we're finished!
+        static $prevCursorMark = '';
+        if ($cursorMark === $prevCursorMark) {
+            return ['ids' => [], 'nextOffset' => $cursorMark];
```
demiankatz: @EreMaijala, I started work on tests today and discovered that this was returning a `cursorMark` key here, when I'm pretty sure it's supposed to have a `nextOffset` key instead. I changed it, but please let me know if I've made a bad assumption. (I suspect it doesn't really matter because the value is not used when the `ids` array is empty, but it seems best to be consistent).

EreMaijala: @demiankatz Quite right, good catch! It didn't make a difference with the current code since it breaks out when ids is empty. Which leads me to think, would it make more sense to break when nextOffset is empty or not returned at all? At least in theory there could be a backend that could return an empty set for an iteration while not being complete. It could of course handle that internally just as well, so this is not important, but I'm trying to think of what would be the most logical contract here.

demiankatz: Yes, I think I like the idea of not returning a nextOffset at all, since there is no next offset when you've hit the last page of results. I'll adjust this at the same time that I finish writing the tests.

module/VuFind/src/VuFind/Sitemap/Command/AbstractGetIdsCommand.php
```diff
@@ -0,0 +1,140 @@
...
+     * CallMethodCommand constructor.
+     *
+     * @param string  $backend       Search backend identifier
+     * @param mixed   $context       Command context
+     * @param Service $searchService Search service
+     */
+    public function __construct(string $backend, $context, Service $searchService)
+    {
+        parent::__construct($backend, $context);
+        $this->searchService = $searchService;
```
aleksip: Passing the search service to a command that is then passed to the search service seems like an odd thing to do. I'm in no way sure about this, but seems like an indication that this could be done differently.

module/VuFind/src/VuFind/Sitemap/Command/GetIdsWithCursorMarkCommand.php
```diff
@@ -0,0 +1,124 @@
...
+                'start' => 0, // Always 0 when using a cursorMark
+                'wt' => 'json',
+                'sort' => $key . ' asc',
+                // Override any default timeAllowed since it cannot be used with
+                // cursorMark
+                'timeAllowed' => -1,
+                'cursorMark' => $cursorMark
+            ]
+        );
+        $results = $this->searchService->getIds(
```
aleksip: This is a deprecated method, could `GetIdsCommand` be used instead?

module/VuFind/src/VuFind/Sitemap/Command/GetIdsWithCursorMarkCommand.php
```diff
@@ -0,0 +1,124 @@
...
+     */
+    protected function setupBackend(Backend $backend): void
+    {
+        // Set up the record factory. We use a very simple factory since performance
+        // is important and we only need the identifier.
+        $recordFactory = function ($data) {
+            return new \VuFindSearch\Response\SimpleRecord($data);
+        };
+        $collectionFactory = new RecordCollectionFactory($recordFactory);
+        $backend->setRecordCollectionFactory($collectionFactory);
```
aleksip: So as discussed in #2036 perhaps the commands in this PR too could be further refactored into non command code in `VuFind/Sitemap` with only the actual `$backend->` calls like the one here being commands?

I'm not really familiar with the code here, and not sure how this is related to `GetIdsInterface` and the related existing `GetIdsCommand`, but perhaps that type of approach could be used here too?

module/VuFind/src/VuFind/Sitemap/Command/GetIdsWithCursorMarkCommand.php
demiankatz: Yes, probably a good idea in this context.

module/VuFind/src/VuFind/Sitemap/Command/AbstractGetIdsCommand.php
demiankatz: Good point, I'm going to rethink this next week and see how I can better separate the pluggable logic from the command.

module/VuFind/src/VuFind/Sitemap/Command/GetIdsWithCursorMarkCommand.php
demiankatz: The issue at hand here is that there are two different methods of getting a list of IDs for the sitemap, each of which interacts with the search system differently... so I want to have a way of handling this in a pluggable way. I think I just got overambitious and pushed too much logic into the command classes... I suspect I can pull back a little and end up with something more reasonable.

demiankatz:
Thanks again for the feedback, @aleksip. I'm out of time for now, but I hope to do more on this next week (and since it's churning in my brain, I might get back on it later tonight if I can't sleep ;-) ).

demiankatz:
@aleksip and @EreMaijala: I have refactored this code yet again, and I think it is now even better. Please take a look at the last two commits and let me know what you think! If you're both happy, I'll finish up testing tomorrow and perhaps get this merged (after reconciling with #2080).

demiankatz:
Quick update: now that #2080 has been merged, I've added a couple more commits to reconcile the new commands here with the functionality added there and to add test coverage.

module/VuFind/src/VuFind/Sitemap/Command/GetIdsWithCursorMarkCommand.php
demiankatz: This change is incorporated into the latest round of refactoring.

module/VuFind/src/VuFind/Sitemap/Command/AbstractGetIdsCommand.php
demiankatz: Done, as noted below.

module/VuFind/src/VuFind/Sitemap/Command/GetIdsWithCursorMarkCommand.php
demiankatz: Done!

EreMaijala:
@demiankatz I admit I haven't been able to properly keep track of all the changes, so while this looks good to me, I'll be happy to let @aleksip say the final word. :)

module/VuFindSearch/src/VuFindSearch/Command/GetUniqueKeyCommand.php
```diff
@@ -0,0 +1,62 @@
...
+
+use VuFindSearch\Backend\BackendInterface;
+
+/**
+ * Get the backend's unique search key field (currently only supported by Solr)
+ *
+ * @category VuFind
+ * @package  Search
+ * @author   David Maus <maus@hab.de>
+ * @author   Aleksi Peebles <aleksi.peebles@helsinki.fi>
```
aleksip: Looks like copy pasted authors here?

module/VuFindSearch/src/VuFindSearch/Command/GetUniqueKeyCommand.php
```diff
@@ -0,0 +1,62 @@
...
+     * @param BackendInterface $backend Backend
+     *
+     * @return CommandInterface Command instance for method chaining
+     */
+    public function execute(BackendInterface $backend): CommandInterface
+    {
+        if (is_callable([$backend, 'getConnector'])) {
+            $connector = $backend->getConnector();
+            if (is_callable([$connector, 'getUniqueKey'])) {
+                return $this->finalizeExecution($connector->getUniqueKey());
```
aleksip, line 56: Should there be a `GetConnectorCommand`? Perhaps not, and then all connector method calls should be commands like this one here? Or if there is a `GetConnectorCommand`, would it be ok to just make direct connector method calls after obtaining the connector? Or should there be connector method specific commands as well?

In any case I guess we want to avoid a situation where there is a `GetConnectorCommand` and also `$backend->getConnector()` calls in other commands...

module/VuFindSearch/src/VuFindSearch/Command/SetRecordCollectionFactoryCommand.php
```diff
@@ -0,0 +1,65 @@
...
+use VuFindSearch\Backend\AbstractBackend;
+use VuFindSearch\Response\RecordCollectionFactoryInterface;
+
+/**
+ * Perform a search and return record collection command.
+ *
+ * @category VuFind
+ * @package  Search
+ * @author   David Maus <maus@hab.de>
+ * @author   Aleksi Peebles <aleksi.peebles@helsinki.fi>
```
aleksip: More copy pasted authors? 🙂

module/VuFindSearch/src/VuFindSearch/Command/TermsCommand.php
```diff
@@ -0,0 +1,70 @@
...
+use VuFindSearch\Backend\Solr\Backend;
+use VuFindSearch\ParamBag;
+
+/**
+ * Fetch terms from the backend (currently only supported by Solr)
+ *
+ * @category VuFind
+ * @package  Search
+ * @author   David Maus <maus@hab.de>
+ * @author   Aleksi Peebles <aleksi.peebles@helsinki.fi>
```
aleksip: And here

aleksip (commented):
I have concentrated on the command pattern related parts, since I am not that familiar with other parts of the code.

module/VuFindSearch/src/VuFindSearch/Command/GetUniqueKeyCommand.php
demiankatz: My thinking on this is that the connector is an implementation detail of the backend; there is no requirement that a backend have a connector, it is simply a pattern that is common and useful. There is also nothing like a ConnectorInterface, so there's no guarantee of connector consistency across backends. I think it makes more sense for commands to focus on retrieving specific pieces of information rather than on gaining access to specific pieces of functionality, and if we ever need to expand something like this to apply to more backends, we just need to add an interface and move more logic out of the command and into the relevant backend(s). (I think this same philosophy applies to your work in #1886, where you wrote a command to returns the results of analysis rather than retrieving the object that does the analysis... which is also a reminder that we should get that finished up soon :-) ). Since the backend code already makes heavy use of the connector all over the place, trying to create an event hook to intercept all those calls would be a losing battle anyway. I think if you want to override connector behavior, you would be much better off doing it with a custom factory rather than a custom event listener. Would you agree, or am I missing something?

module/VuFindSearch/src/VuFindSearch/Command/GetUniqueKeyCommand.php
demiankatz: I thought "I'll fix this in my first command, and then copy it to the other two..." except I missed the second instance and then cloned the mistake! ;-) Thanks for catching it, I've just fixed all three instances.

module/VuFindSearch/src/VuFindSearch/Command/GetUniqueKeyCommand.php
aleksip: Yes, I agree. And thanks for the reminder! 😁

demiankatz: Great, thanks! I believe this is ready to merge now! Onward to make more progress on #2036...

PR #2038, opened 2021-07-30
Small fixes for updateHolds changes

LuomaJuha:
Hide edit properly, if the hold can not be edited. Function call to lower case


demiankatz:
Thank you!

PR #2040, opened 2021-08-03
Add link on publisher into detail record

xmorave2:


demiankatz:
Thanks, @xmorave2, this is a good idea; however, I think it needs to be implemented a little differently. The DefaultRecord templates are shared across records from all search backends, but not every backend supports publisher searching. We don't want to hard-code a link to Solr that could potentially show up in non-Solr contexts.

I would instead recommend using the `getLink` view helper. You could follow the example of how series links work; see:

https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/RecordDriver/DefaultRecord/link-series.phtml

and:

https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/RecordDriver/DefaultRecord/data-series.phtml

We then may have to add additional backend-specific templates to make the links work right in other contexts (WorldCat, EDS, etc.). Alternatively, we could set up the DefaultRecord template to have no link, and then add the link only in places where we know it will work (SolrDefault, etc.).

What do you think?

xmorave2:
I make a bit of investigation and divide record drivers into 4 groups according to implementation needs:

1) Don't support record formatter, so are not relevant to this work: Browzine, EDS, LibGuides, SolrAuth, SolrWeb
2) Do use record formatter, but not support any publisher searching: Pazpar2, Primo
3) Support publisher searching on field 'Publisher': DefaultRecord, Search2Default, SolrOverdrive, Summon
4) Supports publisher searching on other field: WorldCat (srw.pb)

Note: I implemented Pazpar2 link the same way as other Pazpar2 links are implemented - paramater type is silently ignored in Pazpar2 search backend

Note2: Primo uses AllFields for publisher searching, as well as link-journaltitle for Primo does

themes/bootstrap3/templates/RecordDriver/DefaultRecord/data-publicationDetails.phtml
```diff
@@ -4,7 +4,11 @@
     <span property="location"><?=$this->escapeHtml($pubPlace)?></span>
   <?php endif; ?>
   <?php $pubName = $field->getName(); if (!empty($pubName)): ?>
-    <span property="name"><?=$this->escapeHtml($pubName)?></span>
+    <span property="name">
```
demiankatz: I'm not a microdata expert, but based on the [RDFa Lite](https://www.w3.org/TR/rdfa-lite/#vocab-typeof-and-property) documentation, I think you can put the `property="name"` attribute directly onto the `<a>` tag instead of wrapping it in a `<span>` in order to simplify the markup. Would you agree?

demiankatz (commented):
This looks great, @xmorave2 -- thanks for taking the time to work through all the backends! I just have one very minor question/suggestion.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/data-publicationDetails.phtml
xmorave2: I do agree. This is done now

demiankatz:
Thanks -- looks good to me now; I've just merged it!

PR #2041, opened 2021-08-03
Add use command into generated sql

xmorave2:
When copying SQL from VuFind installer, there is need to add USE command manually. This PR should fix it.


demiankatz:
Thanks for catching this! Merging now....

PR #2042, opened 2021-08-04
show profile menu (incl. toggler) for editing favorite lists in account

al-pu:
Displaying profile menu (and toggler) while editing / create new favorite list coming from account features. Account behavior feels more consistently.

Editing / create new favorite lists in lightbox still works fine.

Co-authored-by: Robert Lange <robert.lange@uni-leipzig.de>

### DIFF
--- a/themes/bootstrap3/templates/myresearch/editlist.phtml
+++ b/themes/bootstrap3/templates/myresearch/editlist.phtml
@@ -10,6 +10,13 @@
 
 <?=$this->flashmessages()?>
 
+<?php if ($this->auth()->isLoggedIn() && $this->layout()->getTemplate() !== 'layout/lightbox') :?>
+  <div class="<?=$this->layoutClass('mainbody')?>">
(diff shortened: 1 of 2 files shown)


demiankatz:
Thanks, @al-pu, this makes sense to me, but I'm asking for a review from @crhallberg in case he has further suggestions. My biggest question is whether there is a better way to do the lightbox-conditional logic without having to check the template name. While we do that check in a couple of other places in the code, so it is certainly not unprecedented, I wonder if it would be better handled with CSS, or if we perhaps should consider adding a view helper for lightbox detection to better encapsulate the logic.

crhallberg:
I don't have any better detection methods to recommend (after trying a few things). I think a view helper would allow for clearer use and customization over CSS. The CSS would be fairly simple however. Something to the effect of:

```CSS
#modal .mainbody.left {
    float: none;
    width: auto;
}
#modal .sidebar {
    display: none;
}
```

crhallberg (approved):

demiankatz:
Thanks, @crhallberg and @al-pu. I'm merging this as-is, but I think I will open a separate PR when time permits to create a new "inLightbox()" view helper to better abstract the template checking.

PR #2043, opened 2021-08-04
Improve sanitization of facetexclude parameter.

demiankatz:
This reduces the potential footprint of XSS-style attacks against this parameter. (They don't work even without this change, but this reduces the potential that weird values will get propagated forward).


demiankatz:
@crhallberg, this relates to the facet list functionality that you helped build a while back; just looking for a double-check that my fix here is reasonable.

crhallberg (approved):

PR #2044, opened 2021-08-06
Add ability to filter/reorder limiters on EDS advanced search form

xmorave2:
EDS.ini says for [Advanced_Facets] section:
> These facets will be displayed as limiters on the advanced screen. Only values that are present in the EDS API Available Limiters will be used. Leave blank to use all available values from the INFO method

Which in fact is not true, and Advanced_Facets setting is ignored

I am not sure this is the best solution of this, but for start it is at least working ;)


demiankatz:
Thanks for catching this oversight, @xmorave2! I'm a little short on time right now, but I'll give this a test and review next week, and I suspect I'll be able to merge it after that.

demiankatz:
@xmorave2, I've finally had time to take a closer look at this, and it's actually a bit more complicated and confusing than I had thought.

The existing comment is a bit unclear; I'm not actually sure if the [Advanced_Facets] section is meant to be used for filtering or not. In fact, I'm not sure I understand what the intention of that section is at all. It's currently only used for applying labels to facets, but it doesn't even work for that purpose. Because the limiters are handled in a special way (i.e. "LIMIT|" is prepended to the field name), I can't figure out a way to relabel the limiters, even though the section is supposedly used for that purpose.

I wonder if a better solution would be to remove the [Advanced_Facets] section entirely, because as far as I can tell, it has never worked correctly and may cause confusion. Then we can implement something new to meet whatever needs you have, and it may all be more clear. I can think of two possible options:

1.) An [Advanced_Limiters] section, which is formatted with key = limiter code, value = limiter label. This would then have to be processed in such a way that the labels could be overridden based on the configuration.

2.) A limiter_filter[] setting in [Advanced_Facet_Settings] which could be used to create a restricted list of limiters to be displayed. This would be simpler to implement if we don't want to deal with overriding labels, but also a bit less powerful than the other option.

What do you think? Does this approach make sense, or do you think we should continue down the road of making the existing section work better?

xmorave2:
Thank for review @demiankatz. I tried to implement the first option. I removed the Advanced_Facets section (and references on it) from EDS.ini and added ability to use custom labels.

demiankatz:
Thanks, @xmorave2, this is a big improvement, and the functionality to restrict the available limiters is working for me. However, overriding the labels does not seem to work in my case for some reason. Is it working for you? Is there some other setting that could be interfering with it?

xmorave2:
> Thanks, @xmorave2, this is a big improvement, and the functionality to restrict the available limiters is working for me. However, overriding the labels does not seem to work in my case for some reason. Is it working for you? Is there some other setting that could be interfering with it?

Thanks for testing @demiankatz. I did try to change label on 'LA99' field which of type 'multiselect'. But for field with type 'select' it was not working, because there is inconsistency in template. For multiselect the label is show using this code: `$this->transEsc($facet['Label'])`, but for select type it is `$this->transEsc('eds_limiter_' . $field, [], $facet['Label'])` so the label is used only as default value. I did change it to be the same. When you copy the default definition from [Facets] section it will continue to work with current translations with the possibility to change the label to whatever you want.

demiankatz:
@xmorave, please see commit 1f70e8f. I took another look at this and realized that limiter translation had become very confusing and inconsistent, since if you edited labels in the `[Advanced_Limiters]` section, those changes would apply to the advanced search screen but not the filter applied in the search results. It seemed easier to simplify things to a new `advanced_limiters` setting modeled on the existing `common_limiters` setting. This also suggested to me that we should rename `getEnabledLimiters()` to `getAdvancedLimiters()` since that function is only used for advanced searches. What do you think? I believe this can be merged now, but if you disagree with my approach, I am happy to discuss further.

xmorave2:
> @xmorave, please see commit [1f70e8f](https://github.com/vufind-org/vufind/commit/1f70e8f92916833c773669e4870475621decf300). I took another look at this and realized that limiter translation had become very confusing and inconsistent, since if you edited labels in the `[Advanced_Limiters]` section, those changes would apply to the advanced search screen but not the filter applied in the search results. It seemed easier to simplify things to a new `advanced_limiters` setting modeled on the existing `common_limiters` setting. This also suggested to me that we should rename `getEnabledLimiters()` to `getAdvancedLimiters()` since that function is only used for advanced searches. What do you think? I believe this can be merged now, but if you disagree with my approach, I am happy to discuss further.

Looks good to me, just one more thing: should we rename enabledLimiters property of Options class into something more specific like 'enabledAdvancedLimiters' or 'advancedLimiters' for even more consistancy? What do you think?

demiankatz:
Thanks @xmorave2, I agree! I'll make this edit later today and will then merge the PR.

xmorave2:
Thanks @demiankatz

demiankatz:
Property renamed, so I'm merging now -- I also added a few comments for clarity. Thanks again for catching this problem!

PR #2048, opened 2021-08-12
Update league/commonmark to 2.x

xmorave2:
TODO
- [x] Wait for release of pear/validate including https://github.com/pear/Validate/pull/4 before merging, so we can remove the "dev-master" dependency workaround from composer.json


demiankatz:
Thanks, @xmorave2! I'll circle back to this one as soon as we are ready to raise the minimum PHP version for release 9.

demiankatz:
I also notice that Composer is failing right now due to issues with pear/validate. That issue has been fixed by https://github.com/pear/Validate/pull/4, but we have to wait for that to be formally released, which I believe will happen very soon. By the time we're ready to merge this, it should be all sorted out.

demiankatz:
@xmorave2, I'm thinking about establishing a dev-9.0 branch soon to begin merging early work on 9.0. In preparation for that, I reviewed this PR again and found that some conflicts had developed with the dev branch. I've made an attempt at resolving them, but could you please double-check my work? Unit tests are passing, but I haven't tried hands-on testing yet.

demiankatz:
@xmorave2, I noticed that the build was failing here... this is due to an upstream bug in the pear/validate library. I submitted a PR to get it fixed a couple of months ago, but the fix hasn't been released yet. I've resorted to a "dev-master" version of this dependency as a temporary workaround, but we should ideally wait for that to get fixed before merging this. I'll add a TODO checkbox.

demiankatz:
@xmorave2, thanks to @EreMaijala's work on #2252, the PEAR-related problem that was preventing this from moving forward has finally been resolved. I have also opened #2292 so that we can raise the PHP version independently from other changes/upgrades. I have merged the latest dev-9.0 into this branch and resolved conflicts. However, so much time has passed that there is now a new version of league/commonmark released! Do you want me to merge this as-is and call it incremental progress, or do you want to bring it further up to date first?

xmorave2:
thanks @demiankatz, ti is great. I'll look into it in next days, but I think I would like to update to current version If it is possible.

demiankatz:
Thanks, @xmorave2, there is no rush, so feel free to work on it whenever it's convenient!

xmorave2:
@demiankatz I think this is ready for review. I updated commonmark to latest version 2.2.3 and refactor markdown factory to avoid using deprecated method mergeConfig

config/vufind/markdown.ini
```diff
@@ -12,6 +12,30 @@
...
+; More about extensions: https://commonmark.thephpleague.com/2.2/extensions/overview/
+; Some of them could have a configuration, see sections below.
+; Note: FrontMatter does enable to inject the title other then main heading. Just add
+; something like this at the very beginning of your markdown file:
+;---
+;title: "Some specific title"
+;---
+; See more about FrontMatter extension here:
+; https://commonmark.thephpleague.com/2.2/extensions/front-matter/
+
```
demiankatz: I would remove this blank line to clarify the relationship between the comment above and the setting below.

config/vufind/markdown.ini
```diff
@@ -12,6 +12,30 @@
...
+extensions = Autolink,DisallowedRawHtml,Strikethrough,Table,TaskList
+
+; CommonMarkCore extension is always enabled by default
+[CommonMarkCore]
```
demiankatz, line 37: Adding this section effectively move some settings out of [Markdown] and into [CommonMarkCore]. Do we need to worry about backward compatibility? (I imagine the chances of people being affected by this change are relatively low... but maybe if nothing else a changelog note is justified).

module/VuFind/config/module.config.php
```diff
@@ -338,7 +338,7 @@
     'service_manager' => [
         'allow_override' => true,
         'factories' => [
-            'League\CommonMark\MarkdownConverterInterface' => 'VuFind\Service\MarkdownFactory',
+            'League\CommonMark\ConverterInterface' => 'VuFind\Service\MarkdownFactory',
```
demiankatz, line 341: Should we create an alias for the old service name for backward compatibility, or just make a note in the changelog?

module/VuFind/src/VuFind/Service/MarkdownFactory.php
```diff
@@ -93,104 +111,194 @@ public function __invoke(
...
+            ->get('markdown')->toArray();
+        $this->extensions = isset($this->config['Markdown']['extensions'])
+            ? array_map(
+                'trim',
+                explode(',', $this->config['Markdown']['extensions'])
+            )
+            : self::$defaultExtensions;
+        $this->extensions = array_filter(
+            $this->extensions,
+            function ($item) {
```
demiankatz: The default behavior of array_filter is to drop empty entries. Do you need this callback? (If there's potentially a value in the array that PHP considers empty but which is not a blank string, then it is justified... but otherwise maybe the code can be simplified).

module/VuFind/src/VuFind/Service/MarkdownFactory.php
```diff
@@ -93,104 +111,194 @@ public function __invoke(
...
+                return $item !== '';
+            }
+        );
+        unset($this->config['Markdown']['extensions']);
```
demiankatz: Is there a strong reason to unset this here rather than to filter it out when building the configuration below? I confess I haven't studied the code in close detail yet, but just wondering if the same effect could be achieved in a way that relies less on mutable data.

themes/bootstrap3/templates/content/markdown.phtml
```diff
@@ -1,7 +1,18 @@
...
-$found = preg_match('/#+(.*)/', $this->data, $matches);
-if ($found) {
-    $this->headTitle($matches[1]);
+$content = $this->markdown($this->data);
+$frontMatter = [];
+$title = '';
+if ($content instanceof League\CommonMark\Extension\FrontMatter\Output\RenderedContentWithFrontMatter) {
+  $frontMatter = $content->getFrontMatter();
+  $title = $frontMatter['title'] ?? '';
+} else {
```
demiankatz: What situation causes this else clause to trigger? Will it impact our ability to call `$content->getContent()` below?

demiankatz (changes requested):
Thanks for the progress, @xmorave2! I'm a bit short on time this week so I wasn't able to do hands-on testing or to review the code quite as closely as I would like. However, I was able to do a quick review and left a few comments. I'll try to circle back for a closer look next week.

config/vufind/markdown.ini
xmorave2: I've removed it.

config/vufind/markdown.ini
xmorave2: That is handled in `MarkdownFactory::getConfigForCoreExtension` and is covered by test.

module/VuFind/config/module.config.php
xmorave2: This does make sense. I've added an alias.

module/VuFind/src/VuFind/Service/MarkdownFactory.php
xmorave2: I just wanted to be explicit here. But you are right, I can't imagine that is filtered out by default behavior could be usable as extension name. I've simplified it.

module/VuFind/src/VuFind/Service/MarkdownFactory.php
xmorave2: This `unset` is even unnecessary. The config is being build from scratch in `getBaseConfig` method, so extensions key is not transferred into final configuration. I've removed it.

themes/bootstrap3/templates/content/markdown.phtml
xmorave2: The else branch is used when FrontMatter extension in disabled. There is no impact on `getContent`. `RenderedContentWithFrontMatter` is subclass of `RenderedContent` which only adds  `getFrontMatter` method. But thanks for pointing on it. In case there is FrontMatter enabled and no title in frontmatter of the page. Then title remains empty. I've changed it to use the first heading as title, just as before this patchset.

xmorave2:
Thank you @demiankatz for spending time on this and for all your comments. I tried to address them ;)

config/vufind/markdown.ini
demiankatz: Thanks for the clarification!

themes/bootstrap3/templates/content/markdown.phtml
demiankatz: Great, thanks!

demiankatz:
@xmorave2, I think perhaps that extensions unset in the configuration processing was needed after all. When I set up a clean environment and test rendering a Markdown file in the content template directory, I get an `Unexpected item 'extensions'.` exception:

image

Can you take a look when time permits?

demiankatz:
Another thought: it would be a good idea to have a Mink test that attempts to view rendered Markdown content, since right now, the full test suite is passing yet Markdown rendering is broken. I wonder if it would be justified to create a Markdown example page for testing purposes, or if there's a way we could use Markdown for a "real" purpose and be able to test it as a bonus.

Side note: while testing this, I also discovered what appears to be incorrect behavior... We have asklibrary.phtml and asklibrary_en.phtml templates set up as examples of a language fallback... but VuFind always seems to be rendering asklibrary_en.phtml no matter what language gets selected. (I think this is because VuFind's general translation system chooses English as a last resort if nothing better matches... but it seems like we had originally intended the content system to behave differently).

xmorave2:
I fixed `Unexpected item 'extensions'.` error. It was due to leftover from previous versions.

demiankatz:
Thanks, @xmorave2, I'll retest soon. In the meantime, I've opened [VUFIND-1541](https://vufind.org/jira/browse/VUFIND-1541) and [VUFIND-1542](https://vufind.org/jira/browse/VUFIND-1542) to capture the other issues I highlighted above, since we don't need to hold up this PR on account of those items.

demiankatz (approved):
Working well for me now! Thanks again!

PR #2050, opened 2021-08-13
Upgrade to php-cs-fixer 3 (and apply fixes).

demiankatz:
Note that I also had to upgrade phpmd here to avoid a conflict related to a shared dependency.

TODO
- [x] Review style changes; fix template formatting where needed
- [x] Run full test suite


demiankatz:
@EreMaijala / @xmorave2, I'm interested in your opinions on the changes here. Upgrading to php-cs-fixer 3 changes some default settings, which changes a large number of files. The changes fall into three general categories:

1.) Function signatures and argument lists that take up more than one line have been reformatted to have one argument per line. This is by far the most prevalent change here. There are a few places where this may increase verbosity unnecessarily, but in the vast majority of cases, I think it improves readability and clarity, and so I am inclined to accept the changes.

2.) Class constants are explicitly declared public if visibility is not already specified. This seems fine to me.

3.) Some other minor whitespace adjustments were made; all trivial. In some cases (all in template files), I slightly refactored code to get a cleaner layout.

Bottom line: I think this is all fine. I have reviewed every diff in this PR, so I don't expect either of you to do the same... but I'd be interested to hear whether you have any concerns about the broad trends on display, and a double-check of my template work in commit eca7241e30ac73656eeab206bea5d554666c01ca might not hurt (though I'm not too worried as long as the test suite is passing; also note that I fixed one small template refactoring mistake in 94df418ba62133dd5693db5149921a12a6ed30fd).

demiankatz:
(And, for the record, the full test suite is passing).

xmorave2 (approved):
Great work @demiankatz, and to your questions:

> 1.) Function signatures and argument lists that take up more than one line have been reformatted to have one argument per line. This is by far the most prevalent change here. There are a few places where this may increase verbosity unnecessarily, but in the vast majority of cases, I think it improves readability and clarity, and so I am inclined to accept the changes.

I do agree that even with increased verbosity, it's worth to make this change. From my point of view it is more readable this way

> 2.) Class constants are explicitly declared public if visibility is not already specified. This seems fine to me.

I agree, we've done some work here before: https://github.com/vufind-org/vufind/commit/46248f941a6b2378cbd23adc2258a165af212ed9

> Bottom line: I think this is all fine. I have reviewed every diff in this PR, so I don't expect either of you to do the same... but I'd be interested to hear whether you have any concerns about the broad trends on display, and a double-check of my template work in commit [eca7241](https://github.com/vufind-org/vufind/commit/eca7241e30ac73656eeab206bea5d554666c01ca) might not hurt (though I'm not too worried as long as the test suite is passing; also note that I fixed one small template refactoring mistake in [94df418](https://github.com/vufind-org/vufind/commit/94df418ba62133dd5693db5149921a12a6ed30fd)).

I didn't found any problems here.

EreMaijala:
I agree with both of you, and I like it that we'll have consistency in how argument lists are formatted.

EreMaijala (approved):

demiankatz:
Thanks, both of you -- I'm feeling comfortable with merging this now!

PR #2052, opened 2021-08-13
Overdrive: use class constants instead of strings.

demiankatz:
These should be trivial changes to modernize the code and improve static analysis.


demiankatz:
I'm not in a position to test this, so if @bpalme has an opportunity to take a look, I would be grateful!

bpalme:
It all looks fine to me.  It's not directly related to Overdrive functionality.  I probably just mimicked something I found in another class.

demiankatz:
Thanks, @bpalme -- I'm sure you are correct, and I think what probably happened here is that I refactored some of the code this was based on while your pull request was still in progress, and then didn't notice that there were some outdated styles in the pull request when it finally got merged. In any case, happy to have it sorted out now, and it should make no functional difference to the code. :-)

PR #2054, opened 2021-08-13
Upgrade to laminas-cache 2.13.0

demiankatz:
This PR updates to the latest laminas-cache release and makes use of up-to-date cache features (including use of the standard Blackhole adapter instead of the custom, VuFind-specific NoCache adapter).

TODO
- [x] Switch to non-deprecated factory method for adapters
- [x] Test all supported adapters
- [x] Adjust composer configuration to omit deprecated adapters
- [x] Run full test suite
- [x] Note relevant changes in changelog (removal of VuFind-specific NoCache adapter, removal of APC support, removal of deprecated adapters, change of VuFind\Cache\Manager constructor signature)


demiankatz:
@EreMaijala, do you mind double-checking my work on this one? I'm particularly interested in whether your Solr query caching still works as expected, as I haven't had a chance to test that specific detail yet (though my hands-on testing of other caches, plus a run of the full test suite, seem to suggest that the general approach is valid).

Specific notes on what is changed here:

- The static factory method for building cache adapters is deprecated, so this uses the new recommended method. Some configuration arrays had to be restructured to work with this method.
- The APC cache adapter has been irrelevant since at least PHP 5.5, so I have removed support for it (we might want to add back support for APCu in the future, but since nobody has complained about APC breaking, I'm guessing there is no current demand).
- The VuFind-specific NoCache adapter has been removed, because the Laminas-provided BlackHole adapter does the same thing.
- I have disabled the inclusion of several storage adapters in composer.json to eliminate deprecation warnings, since we don't use any of them anyway (now that APC support is removed).

EreMaijala (approved):
Solr caching is still working fine, and all looks good to me!

demiankatz:
Thanks, @EreMaijala!

hajoseng:
@demiankatz, @EreMaijala: After updating VuFind and running composer I get a fatal error saying: 'Unable to resolve service "Laminas\\Cache\\Service\\StorageAdapterFactory" to a factory; are you certain you provided it during configuration?'. Have I missed something?

EreMaijala:
@hajoseng Have you made sure to clear the caches? I've only seen the error when I failed to run composer update, but if you've done that, I'd suspect that something old is still cached.

hajoseng:
@EreMaijala Thank you; I indeed missed clearing the config-caches.

PR #2055, opened 2021-08-16
Do not include redundant traits.

demiankatz:
We do not need to include the LoggerAwareTrait, since it is also included automatically by the OAuth2TokenTrait. This redundant inclusion was causing PHPMD reporting to fail.


demiankatz:
Thanks for correctly diagnosing this one, @EreMaijala. Any objections to merging this?

EreMaijala (approved):
While I'd like to be more explicit about how traits are included, I can't see a better solution that doesn't get too convoluted.

demiankatz:
Thanks, @EreMaijala -- I suppose we could always add comments on `use` statements for traits that include other traits if that would help with clarity... but that seems like potentially more of a maintenance burden than a help. :-)

PR #2056, opened 2021-08-16
Update SASS and tools for scssphp 1.6.0

crhallberg:
- [x] Remove deprecated extends alls.


demiankatz:
@crhallberg, it's still not happy for me. I'm getting:

<pre>
Processing bootstrap3
	compiled.scss
WARNING: on line 59 of themes/bootstrap3/scss/components/channels.scss:
Compound selectors may no longer be extended.
Consider `@extend a, :hover` instead.
See http://bit.ly/ExtendCompound for details.
</pre>

demiankatz:
Thanks for the fix, @crhallberg! I've confirmed that compiling with 1.4.1 and 1.6.0 now yields identical output from the SCSS compiler. I also confirmed that the channel style you removed is obsolete, so I removed it from the LESS as well and recompiled everything for consistency. Merging now!

PR #2059, opened 2021-08-18
Daia driver services

hajoseng:
Here is the first part of my pull request; it should solve PR #2026. I'd like to add electronic holdings to the driver's functionality but I have to work a little on it. The changes only touch the DAIA driver and DAIA configuration; following my remark on #2026 the driver still merges limitations regardless to which service they belong.


hajoseng:
I skip this and try a new one.

PR #2060, opened 2021-08-18
DAIA: make accepted service list configurable

hajoseng:
…uration is in DAIA.ini, default loan and presentation (interloan can be added). Limitations are only regarded if the belong to an accepted service. There is as well a bit of code restructured in order to be able to include electronic holdings. Holdings without holding services are skipped.
This is the first part which should solve pr #2026. I'd like to add the electronic holdings functionality as well, but I have to do some more considerations on this topic.


hajoseng:
There is a failure of the CI-tests I don't understand; can anybody give me a hint?

demiankatz:
Thanks, @hajoseng. I have requested a review from @olli-gold since I think he is better qualified to comment on this than I am... but if either of you need my input, just let me know and I'll take a closer look. I will certainly follow your conversation as it unfolds.

Thanks also for fixing the CI failure. Sorry I didn't get a chance to look before you solved it. Was it just a php-cs-fixer failure, or something different? Let me know if you had to spend a lot of time on the solution, and perhaps I can offer you a shortcut for future reference.

hajoseng:
@demiankatz: Thank you; it was a php-cs-fixer failure. I was a bit irritated because it only showed a diff-output instead of pointing on errors or warnings und - moreover - it showed these diffs at placed where I haven't changed anything. I guess the rules have become more strict.

demiankatz:
@hajoseng, when you get php-cs-fixer diff output, you can just run `vendor/bin/phing php-cs-fixer` to apply those diffs; you don't have to do anything by hand. I hope that saves you trouble in the future!

(And it's likely that the new problems were caused by the recent upgrade to php-cs-fixer 3, which is more strict about argument lists than php-cs-fixer 2 was).

config/vufind/DAIA.ini
```diff
@@ -56,6 +56,10 @@ daiaContentTypes['xml'] = "application/xml"
 ; daiaContentTypes['json'] = "application/json, application/javascript"
 daiaContentTypes['json'] = "application/json"
 
+; accepted on-site services
+; (separate multiple values by colons)
+daiaOnSiteServices = loan:presentation:interloan
```
olli-gold: As the default for onSiteServices in DAIA.php is loan and presentation, I guess this should be reflected here as well. So I would not put interloan into the configuration setting by default.

hajoseng: You are right; I fixed it.

olli-gold: Thank you :-)

hajoseng:
@demiankatz, thank you for your hint on php-cs-fixer as command tool. Just tried it.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/DAIATest.php
```diff
@@ -49,93 +49,97 @@ class DAIATest extends \VuFindTest\Unit\ILSDriverTestCase
...
+                    'reserve' => "N",
+                    'callnumber' => "DEF 34",
+                    'location' => 'Example Library for DAIA Tests',
+                    'locationid' => 'http://uri.gbv.de/organization/isil/DE-000',
+                    'locationhref' => 'http://www.example-library.edu',
+                    'storage' => 'Abteilung III',
+                    'storageid' => '',
+                    'storagehref' => '',
+                    'item_notes' => ['mit Zustimmung'],
+                    'services' => ['presentation', 'loan'],
```
olli-gold, line 104: It's not clear, to which service the item_note is related. Maybe we need generic item_notes and should add a new field for service-related item_notes?

olli-gold (commented):
Thank you @hajoseng! Overall I am happy with that PR, although it breaks with some of our customizations in our VuFind instance (which are expecting to have electronic holdings in the DAIA response, if electronic holdings are applicable at all). But I guess, those customizations can easily deal with the new array key for electronic holdings instead of the holdings key.

Before eventually approving this, I would like to discuss the effects of the introduction of the new array key, though. The DAIA driver would be the only driver handling it this way so far and I guess this might be a reason to postpone these changes to VuFind 9, as it might affect some display logics and templates as well...

I do also not see, that this PR really solves the problem of mixed limitations for different services, which was the original problem in #2026 - it just skips limitations, if they are related to services, which are not in the onSiteServices, as I noted here: https://github.com/vufind-org/vufind/pull/2060/files#r691956985... For instance, it's still not clear, to which service an item_note is related. Maybe we need a new item_note field to indicate relations between services and notes (ie. limitations)?

hajoseng:
Just did another commit to include electronic holdings as well. Those holdings can be connected to several remote services and several remote-links for each item - each with different limitations (e.g. one for within university scope and one for externals provided with a login). I just put those links and limitation remarks into item_notes; maybe - or hopefully - there show up nicer solutions. Showing remote services is configurable as well; it is skipped by leaving the daiaRemoteServices list empty.

hajoseng:
@olli-gold, thank you for your answer. Important issues which we have to work out. Maybe you have a look at the way the driver handles remote services after my last commit.
I'm going to work more on this pr from next week monday. Until then I'll make up my mind and hopefully get more ideas.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/DAIATest.php
demiankatz: Could this also be handled by appending/prepending service-specific text to the notes in the general field? (I don't mind adding something new, but if we can keep the ILS driver spec simpler and less specific without losing functionality, that's probably preferable).

demiankatz:
@olli-gold, the `electronic_holdings` key is not new to this PR or unique to the DAIA driver; it is also supported by the Alma driver and the Demo driver. It's definitely worth confirming that what is added here works consistently with the other implementations. (I haven't studied the code closely, so I have no reason to believe that there are any problems -- just saying that on general principle).

olli-gold:
Ah, ok, thanks for that notice - I did not know, that this electronic_holdings key already exists. That definitely changes my mind about approving this PR :) However, I'll test the latest development again.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/DAIATest.php
olli-gold: Yes, I thought about that as well and it may be a workaround if we do not want to introduce a new field .

demiankatz: I'll be interested to hear what @hajoseng has to say! The other consideration with either solution is that, since translation for 8.0 is already underway, we don't have the ability to add new translation strings right now... (Though in this case, if it's absolutely necessary, I could live with adding new strings in just English and German for now, since I don't think DAIA is widely used in other language communities yet... but correct me if I'm wrong about that!)

hajoseng:
@demiankatz: It doesn't look like a real merge conflict; or have I missed something?

themes/bootstrap3/templates/RecordTab/holdingsils/electronic.phtml
```diff
@@ -23,7 +40,7 @@
             <b><?=$this->transEsc("Item Notes")?>:</b>
             <ul>
               <?php foreach ($item['item_notes'] as $item_note): ?>
-                <li><?=$this->escapeHtml($item_note) ?></li>
+                <li><?=$item_note ?></li>
```
demiankatz: Was this reverted intentionally, or just a merge error?

themes/bootstrap3/templates/RecordTab/holdingsils/electronic.phtml
```diff
@@ -3,11 +3,28 @@
...
+            <?=$locationTextEsc?>
+          <?php endif; ?>
         <?php else: ?>
-          <?=$locationTextEsc?>
+          <?php foreach ($item['location'] as $index => $location): ?>
+            <?php $locationTextEsc = $this->transEscWithPrefix('location_', $location); ?>
+            <?php if ($item['locationhref'][$index] ?? false): ?>
+               <p>
+               <a href="<?=$this->escapeHtmlAttr($item['locationhref'][$index])?>" target="_blank"><?=$locationTextEsc?></a>
+               <?=$item['locationnote'] ?? ''?>
```
demiankatz: This probably needs HTML escaping.

themes/bootstrap3/templates/RecordTab/holdingsils/electronic.phtml
```diff
@@ -3,11 +3,28 @@
...
+               </p>
+            <?php else: ?>
+              <p>
+              <?=$locationTextEsc?>
+              <?=$item['locationnote'] ?? ''?>
```
demiankatz: This probably needs HTML escaping.

demiankatz (commented):
@hajoseng, there were some merge conflicts related to @EreMaijala's recent work to improve consistency of HTML escaping. I think I've resolved the direct conflicts, but there also seem to be some other related issues in the template, which I've commented on below.

I also suspect it may be possible to simplify this code (perhaps by casting the data structure to an array) but I didn't analyze it in depth to confirm this suspicion.

themes/bootstrap3/templates/RecordTab/holdingsils/electronic.phtml
hajoseng: This is a merge error.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/DAIATest.php
```diff
@@ -49,93 +49,97 @@ class DAIATest extends \VuFindTest\Unit\ILSDriverTestCase
...
+                    'barcode' => "1",
+                    'reserve' => "N",
+                    'callnumber' => "DEF 34",
+                    'location' => 'Example Library for DAIA Tests',
+                    'locationid' => 'http://uri.gbv.de/organization/isil/DE-000',
+                    'locationhref' => 'http://www.example-library.edu',
+                    'storage' => 'Abteilung III',
+                    'storageid' => '',
+                    'storagehref' => '',
+                    'item_notes' => ['mit Zustimmung'],
```
olli-gold, line 103: I just notice, that this might be a limitation_type rather than an item_note...

themes/bootstrap3/templates/RecordTab/holdingsils/electronic.phtml
```diff
@@ -3,12 +3,20 @@
...
       <td class="fullLocation">
-        <?php $locationTextEsc = $this->transEscWithPrefix('location_', $item['location']); ?>
-        <?php if ($item['locationhref'] ?? false): ?>
-          <a href="<?=$this->escapeHtmlAttr($item['locationhref'])?>" target="_blank"><?=$locationTextEsc?></a>
-        <?php else: ?>
-          <?=$locationTextEsc?>
+        <?php if (!is_array($item['location'])): ?>
+          <?php $item['location'] = [$item['location']]; ?>
         <?php endif; ?>
+        <?php foreach ($item['location'] as $index => $location): ?>
```
demiankatz: Thanks for the improvements, @hajoseng. You can further simplify this by removing the `is_array` check and just changing your loop to `<?php foreach ((array)$item['location'] as $index => $location): ?>` -- this does exactly the same thing through type-casting and saves you three lines of code. :-)

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/DAIATest.php
hajoseng: @olli-gold: a limitation_type contains the id of a limitation, which is in any case should be a purl. Only the contents of limitations are mapped to item_notes. In fact I've never seen a limitation id in real DAIA; limitations seem to be announced only in a free-text style.

olli-gold: I see the problem, that we do not get an ID from our DAIA provider, but the specification recommends to set IDs (cf. https://gbv.github.io/daia/daia.html#limitations). Maybe we should ask for IDs at our DAIA provider, but it will certainly take much time until we really have them. In the meantime we could choose to either add IDs in the driver (at least for some known free text limitations) or to keep the status of just showing the limitations as item_notes (if this is common practice for the most drivers in VuFind)...

olli-gold: But maybe we should open a different PR for that development, as it really does not need to be addressed here.

hajoseng:
I would keep this pr like this. I'm still not sure if there's a need to map item_notes to services. Maybe it's ok just to prevent notes related to service which are not included from being collected. @olli-gold @lahmann : Would you like to do some checks to approve this pr (or to find some more issues)?

olli-gold (approved):
I think I'm ok with the PR as is. Thank you for that development, @hajoseng!

demiankatz:
Thanks, @hajoseng and @olli-gold -- would you like me to merge this now, or should I wait for @lahmann? (I believe he is out of the office right now, so it might be another week or two).

hajoseng:
I think it would be good, if @lahmann had a look on it.

lahmann:
Sorry for overseeing this request for review and thanks for working on the driver. As far as I can tell the code is fine, but as @olli-gold already mentioned will break some code in customized DAIA drivers which is ok for major releases.

What I'm wondering is, if the comments in DAIA.ini need some more information on how the handling of `on-site` and `remote` services differ (also the newly introduced functions `getItemOnSiteStatus` and `getItemRemoteStatus` miss comments in the function description what they do and how they differ). As far as I know DAIA does not distinguish between `on-site` or `remote` services regarding limitations and availability (there are some differences e.g. in handling of `delay` though). E.g. the DAIA service `remote` is also not necessarily bound to electronic availability (cf. http://gbv.github.io/daia/daia.html#services) but can relate to interloan services as well. So in order to configure the driver in a predictable way it would be necessary to at least mention the different ways of how limitations are handled for the two classes of services and why this distinction matters.
Besides that, I am not really convinced that the missing handling of limitations and unavailability information for remote services is something on would expect from a DAIA client.

Sorry again for posting my questions late, but especially because this pull request introduces breaking changes it would be good to keep it as close to the DAIA specifications as possible.

demiankatz:
Thanks, @lahmann -- sounds like we will not be merging this for 8.0, just to be on the safe side!

demiankatz:
I'm putting a 9.0 milestone on this so we don't lose track of it, but I'm certainly open to discussing the possibility of an earlier release if it turns out to be appropriate.

olli-gold:
I'm fine with postponing this PR to VuFind 9, but would like to recall the original problem I had (PR #2026), which is still not solved as I have closed my PR in favor of this one. My original problem was, that limitations from the interloan service were passed to the result, even though they were related to the interloan service only, which is not considered as relevant to the driver. I still believe, that this can cause confusion for some records, as the limitation could be displayed, although it is not relevant to the item at all.
I still think, that this should not wait to be fixed in VuFind 9, but earlier. The developments in this PR are going further as they are addressing to make the accepted services configurable at all and to use the different array keys for electronic holdings and physical holdings as (how I learned during the conversation here) it's done by other drivers as well.

demiankatz:
Thanks, @olli-gold. Certainly, if there's a way to address the problem from #2026 incrementally leading into the full changes found here, that might be good to include in an earlier release. It's also possible, if we consider this a bug fix that requires a breaking change, that we could include changes in an 8.x release as long as we note them in the changelog. Since the DAIA driver is a fairly specialized piece of the project, I would expect the impact to be comparatively low.

In any case, since I do not use DAIA and have limited knowledge of it, I don't have particularly strong opinions... just let me know how I can help to facilitate this work so we meet everyone's needs!

hajoseng:
@lahmann, @olli-gold Thank you very much for your comments. Originally I only wanted to add some improvements, as the driver currently doesn't consider remote and openaccess services at all. Anyway I've started to investigate a little more in this and will proceed with this pull request.

hajoseng:
Now on-site and remote services are no longer distiguished. Instead electronic holdings are identified by a method using service-related criteria. As well there are some more information gathered as item_notes (e.g. about issues of newspapers or magazines and licence notes). Additionally some minor changes (empty notes shouldn't be shown as well as holding informations lacking basic items). See comments.
I'll try to fix the unit-text issue tomorrow.

hajoseng:
Changes so far are:
- which services to analyze is made configurable
- if service loan is available and href is given, addLink is set to true (see 2.6 od DAIA documentation)
- if service remote is given and delay is set to PT0S, this service is assumed to be an electronic service (see 2.5)
- content of chronology is gathered as it provides crucial informations in many cases (see 2.8)
- item['about'] is gathered because sometimes it contains licence-notes (not documented)
- if no available service, no duedate and no service link is found the item returns just the information that it is not holdable
- if no item_notes are found, no item_notes are displayed

But anyhow there's maybe more work to do:
- if several services are available there is still no link between a service and the related notes
- the configuration of  preferred service has still no effect on displayed services in a record view
- in my installation I get quite weird results in the result list view; I haven't investigated in this issue yet

@lahmann @olli-gold: I would be glad, if you have another look at this PR and leave some comments.

demiankatz:
@lahmann / @olli-gold, I just happened to notice that this PR has been inactive for several months. Did either of you have a chance to look at it? I realize that @hajoseng is not available right now, but I just wanted to see if we can do anything to keep this moving forward.

module/VuFind/src/VuFind/ILS/Driver/DAIA.php
```diff
@@ -182,6 +194,16 @@ public function init()
...
             $this->debug('No ContentTypes for response defined. Accepting any.');
         }
+        if (isset($this->config['DAIA']['daiaServices'])) {
+            $this->daiaServices = explode(
+                ':',
+                $this->config['DAIA']['daiaServices']
+            );
+        } else {
+            $this->debug(
+                'Accepting loan, presentation, remote and openaccess as services.'
```
lahmann, line 204: Improved debugging message: "No services configured - accepting loan, presentation, remote and openaccess as services by default."

module/VuFind/src/VuFind/ILS/Driver/DAIA.php
```diff
@@ -781,15 +803,50 @@ protected function parseDaiaArray($id, $daiaArray)
...
+                $status = $this->getItemStatus($item);
+                // chronology provides further information and is an item_note
+                if (isset($item['chronology'])) {
+                    $status['item_notes'][] = $item['chronology']['about'];
+                }
+                // check if this is an electronic holding
+                $isElectronic = $this->serviceIsElectronic($status);
+                // add result_item to the result array: either to electronic_
+                // holdings if it is electronic or to holdings otherwise
+                if ($this->serviceIsElectronic($status)) {
```
lahmann, line 815: Check can be done on `$isElectronic`.

module/VuFind/src/VuFind/ILS/Driver/DAIA.php
```diff
@@ -885,54 +931,82 @@ protected function getItemStatus($item)
...
             }
         }
 
-        /*'returnDate' => '', // false if not recently returned(?)*/
+        // gather more information about the provided service
+        if (isset($item['about'])) {
+            $item_notes[] = $item['about'];
+        }
+
+        // If no useful availability service is found, return only the
```
lahmann, line 977: The added logic block below will introduce very strict requirements for DAIA items in terms of availablity modelling: according to this if-conditional the unavailability information of an unavailable item will be shown to the user only if either a link, a duedate or an available service exists. Otherwise this item will simply be shown as unavailable. This is too strict as according to DAIA specification (http://gbv.github.io/daia/daia.html#availability) the only required field for defining an unavailable service is `service`. E.g. the availability of an unavailable item could be further specified by a given limitation, but any limitation will be ignored if the above mentioned conditions are not met.

lahmann:
I finished my review and would not recommend merging this pull request in its current state as the introduced changes will require a very specific and strict way of DAIA availability modelling which is not defined by the DAIA specification (see https://github.com/vufind-org/vufind/pull/2060#pullrequestreview-1146074268).

My suggestion is to postpone this pull request to the VuFind 10 release.

It would be interesting to hear if @olli-gold has some thoughts on this.

demiankatz:
Thanks, @lahmann, I have changed the milestone to 10.0 for now. We can, of course, always merge this early if things get resolved sooner/more easily than anticipated.

demiankatz:
@hajoseng, @olli-gold and @lahmann: I see that this PR now has some conflicts that need to be resolved. Is there still a need for this work, and does anyone have time/interest to move it forward? Unfortunately, I am not in a good position to help resolve this one, but of course I'm always happy to support progress however I can!

lahmann:
I still have objections (see above) which I would like to discuss before merging this PR - feedback by @olli-gold and @hajoseng would be very much appreciated.

demiankatz:
Since both @olli-gold and @hajoseng may have limited availability right now, I'm going to push this forward to the 10.1 milestone to give us more time for this conversation. Thanks, @lahmann!

(Of course, I'd still be happy to merge this in time for the 10.0 release if the work gets done faster than expected).

demiankatz:
I'm moving this to the 11.0 milestone since significant work is still needed, and the 10.1 release is getting close.

demiankatz:
Reassigning milestone as per discussion on the 1 July Community Call.

dmj:
We can close this one. We are running fine with the default DAIA ILS driver.

dmj:
@demiankatz I think we can also delete the branch. Could you double-check that this is the case?

demiankatz:
Thanks, @dmj, it is safe to delete the branch; I have taken care of it.

PR #2063, opened 2021-08-18
Make QRCode loader validate params and handle errors better.

EreMaijala:
This allows e.g. /QRCode/Show?text=something&level=L&size=FOO&margin=4 to return the "QRCode not available" image instead of failing with an exception.


demiankatz:
Looks good -- thanks!

PR #1617, opened 2020-05-04
ShibbolethWAYF support

xrosecky:
ShibbolethWAYF implementation (where are you from) - support of multiple IdPs for authentication for use in combination with MultiBackend driver. 

- Common functionality moved to AbstractShibboleth
- Discovery service with WAYF is required - see [IdPDiscovery](https://wiki.shibboleth.net/confluence/display/CONCEPT/IdPDiscovery).
- Mapping of attributes can be overriden for each IdP

PR #2064, opened 2021-08-18
Fix Shibboleth Factory.

demiankatz:
- Fix incorrect case-sensitive config filename
- Fix outdated config service name


demiankatz:
@xrosecky, I noticed some problems while reviewing code, and I think a bug slipped through during our work on #1617. Do you mind reviewing these fixes and letting me know if you have any concerns before I merge them? Thanks!

demiankatz:
(If you think it would be better to rename Shibboleth.ini to shibboleth.ini instead of changing the constant in the factory, that's an option too!)

xrosecky:
@demiankatz I have no objections.

demiankatz:
Thanks, @xrosecky!

PR #2065, opened 2021-08-19
EDS: Use div containers instead of spans to give each link its own line.

EreMaijala:
…ir own line.

Otherwise the links would be concatenated to a pretty much unreadable line.


demiankatz:
Thanks, @EreMaijala, I see that this is clearly an improvement. (I hadn't noticed this issue in our own instance, because we chose to completely suppress those links in that context in favor of directing users exclusively to our link resolver).

PR #2066, opened 2021-08-20
Improve Solr error listener.

demiankatz:
- Do not cause unexpected errors if the backend fails to initialize.
- Use constant for tag more consistently.


demiankatz:
@EreMaijala, this PR fixes the problem you identified while testing #2037; the error listener was not checking to be sure the backend had initialized properly when handling errors. This problem was not previously visible in the sitemap generator because that code was fully bypassing the search service and thus not triggering listeners at all.

While fixing that bug, I also took a little time to clean up some inconsistent usage of a constant for the "parse error" tag, and made some minor cosmetic changes like removing useless whitespace and eliminating an unused "use" statement.

I did test to confirm that the detection and auto-correction of parse errors still works as expected.

EreMaijala (approved):
Thanks, tested with #2037 and now all good. :)

demiankatz:
Thanks, @EreMaijala!

PR #2067, opened 2021-08-23
Add a constraint to record tab route parameter.

EreMaijala:
Also make sure that the url helper can handle a case where there's no proper route match (the url doesn't match the default route either).

The url helper issue would come up with e.g. /Record/id/* where the tab doesn't match the constraint and the url doesn't match the default route.


EreMaijala:
Verified that Mink tests still pass.

demiankatz:
Thanks, @EreMaijala, looks good to me!

PR #2069, opened 2021-08-24
Fix handling of an empty hold updateFields setting.

EreMaijala:
This avoids displaying of edit button and links when `updateFields = ""`.


demiankatz:
Thanks!

PR #2071, opened 2021-08-24
KohaRest: Fix handling of frozenThrough/startDate hold parameter.

EreMaijala:
Somehow I totally failed to test the functionality, and Koha has required a date-time instead of a date for at least the last three years. Now I've actually verified that changing hold frozen through date or setting the initial start date to future works.


demiankatz:
Thanks for the fix!

PR #1695, opened 2020-08-27
Add makeLink and makeTag view helpers

crhallberg:
To clean up areas in the templates where we do:

```php
<?php if($something): ?>
    <a href="<?=$url ?>"...
<?php endif; ?>
<?=$text ?>
<?php  if($something): ?>
    </a>
<?php endif; ?>
```

Now:

```php
<?=$this->makeLink($text, $something ? $url : false); ?>
```
or
```php
<?php if($something): ?>
    <?=$this->makeLink($text, $url); ?>
<?php else: ?>
    <?=$text ?>
<?php endif; ?>
```

PR #2072, opened 2021-08-24
Improve escaping behavior of RecordLink helper.

demiankatz:
This was created to support discussion on #1695.

TODO
- [x] Create RecordLinker helper as discussed
- [x] Replace all recordLink() references with properly-escaped recordLinker() calls.
- [x] Run full test suite.
- [x] Update changelog when merging (renaming of helper, removal of getHoldUrl method, renaming of getBreadcrumb to getBreadcrumbHtml)
- [x] Update deprecation cleanup ticket in JIRA


demiankatz:
@EreMaijala, on #1695 you expressed concern about the RecordLink helper returning pre-escaped output. I suspect that this is somehow possible, but I can't easily reproduce it. Do we need to add something else to this test suite to cover the potential problem case? I thought the trigger might be multiple query parameters, but those do NOT appear to be getting HTML escaped by default.

EreMaijala:
@demiankatz At least the following methods escape the the result (some allow to bypass escaping): related, getHoldUrl/getRequestUrl, getChildRecordSearchUrl, getVersionsSearchUrl. They all use escapeHtml instead of escapeHtmlAttr, though, so using the results in an attribute usually works but is not quite right. The last two include url parameters by default as well, so you should see the results in the "Show all versions" link if you have e.g. record versions enabled.

EreMaijala:
@demiankatz So, yes, I think we need some more tests, and we need to fix this...

demiankatz:
@EreMaijala, since these methods are not used in very many places, I don't think it will be especially difficult to fix this... but the question is how we want to do it. I notice that the `related()` method has a parameter to disable escaping, but none of the others do. I can think of a few possible approaches:

1.) Disable all escaping in the helper, and move all escaping to the templates that call the helper. This simplifies everything considerably, but also introduces a large number of backward compatibility breaks (though I suspect they are unlikely to impact many people, if any).

2.) Add a parameter to all the methods to disable escaping, and default it in a way that maintains backward compatibility. Perhaps this parameter could accept a callback as well as a boolean, so instead of having the helper pull things out of the view helper plugin manager, we just call things like `$this->recordLink()->related($link, [$this, 'escapeHtmlAttr'], $source)`. This might be a little overcomplicated, but if we add a support method to handle all the escaping based on the parameter, I think it might actually be a fairly clean solution.

3.) Add a `setEscaper()` method to the helper, so we can inject a helper to do escaping... then we could set up the helper so that it uses an injected escaper if found, but does not escape if there is no injected helper. I don't really like this solution, but I thought of it, so I mention it here for completeness.

I initially favored option 1, but as I think about it more, I think I prefer option 2, as it potentially gives us both backward compatibility and a clean path forward. What do you think? I'll implement something once I've heard your input.

demiankatz:
I guess another factor to consider is, even if we choose option 2, whether we should change default behavior to "do not escape." I suspect the impact of this potentially backward-breaking change would be low, and the resulting consistency might improve clarity.

EreMaijala:
@demiankatz I'm torn between the options. While I can't see many cases where switching the default to not escape would cause problems, that's the kind of change I'm still hesitant to do. I think passing an escaper is overly complicated since one can just wrap the result in the desired escaper in the template.

There's one more option: add a new helper that doesn't escape anything, make the current one a wrapper around it and deprecate it. With this approach you wouldn't need to cause backward-compatibility issues while the new version would be clean. The only problem with this approach is that I can't think of a good name for the helper, but I'm sure that can be resolved.

demiankatz:
@EreMaijala, regarding option 2, I guess the question is whether there's a significant readability difference between `$this->escapeHtmlAttr($this->recordLink()->method())` and `$this->recordLink()->method([$this, 'escapeHtmlAttr'])`. The answer is probably yes. :-)

Regarding a new name, the best I can come up with is `recordLinker` instead of `recordLink`, since it's a helper and could perhaps benefit from a more active name. What do you think?

EreMaijala:
@demiankatz I find the first one more readable. :) Also, I'm thinking of adding some sort of static analysis to catch potential problems, and in that case having the escaping method in there makes the analysis easier. `RecordLinker` sounds good to me. I could only come up with `RecordLink2`, which is nowhere near as dynamic. :)

demiankatz:
@EreMaijala, I've made some progress on this -- I believe I've done the bare minimum work to introduce the new helper and keep all the tests passing. I've also eliminated all references to the old helper in the core module code, and adjusted a small sampling of calls in the themes. The next step is to check and fix all the template references to the old helper, of which there are >100. If you're satisfied with the work so far, I'll do that as soon as I am able... but this seems like a good stopping point to get further feedback.

Also noteworthy and potentially confusing: the record() helper's getLink method returns pre-escaped URLs (see, for example, my changes to data-containerTitle.phtml). Changing this could be quite disruptive, but something to think about for the future...

EreMaijala:
@demiankatz Great progress, this looks good to me!

demiankatz:
Thanks, @EreMaijala. I think I've cleaned up the remaining references, but that was a lot of edits, so a double-check would be greatly appreciated. :-)

demiankatz:
Update: looks like at least one integration test is failing; not sure if I'm going to have time to investigate that before going to bed, but I'll try to finish up tomorrow if necessary....

demiankatz:
No, never mind, I think it was just a timing glitch -- I repeated the offending test and everything passed this time.

EreMaijala (approved):
Looks great to me!

demiankatz:
Thanks, @EreMaijala -- I've reviewed the diffs one more time for good measure, and having found no obvious problems, am merging now.

PR #2073, opened 2021-08-25
Fix HierarchicalFacetListener comparison 

aleksip:
Fixes the comparison which was mistakenly changed in the refactoring.


EreMaijala (approved):

demiankatz:
Thanks for the fix, and sorry for missing this in review! This is probably a sign that we should add test coverage for this listener, when time permits. (Unfortunately, I'm a bit overwhelmed with other projects at the moment).

PR #2074, opened 2021-08-26
Add support for help text on the hold update form.

EreMaijala:
This is just like the help text that is displayed on the hold form.

I just realized that while the helpText setting has been universally supported for a long time, it's not in the default config files. @demiankatz Do you think it should be added?

TODO:
- [x] Update wiki
- [x] Decide whether to include the help texts in all the ILS sample configs.

### DIFF
--- a/config/vufind/Alma.ini
+++ b/config/vufind/Alma.ini
@@ -73,7 +73,7 @@ updateFields = pickUpLocation
 ; Optional help texts that can be displayed on the hold update form. Displayed as is;
 ; HTML tags can be used, and everything needs to be properly escaped.
 ;updateHelpText[*] = "Hold update default help text used if not overridden."
-;updateHelpText[en] = "Hold update help text for British English localization."
+;updateHelpText[en-gb] = "Hold update help text for British English localization."
 
 ; The "NewUser" section defines some default values that are used when creating an account
 ; in Alma via its API. This is only relevant if you use the authentication method "AlmaDatabase"
--- a/config/vufind/Alma.ini
+++ b/config/vufind/Alma.ini
@@ -65,6 +65,16 @@ defaultPickUpLocation = ""
 ; available for pickup). The only supported value is "pickUpLocation".
 updateFields = pickUpLocation
 
+; Optional help texts that can be displayed on the hold form. Displayed as is;
+; HTML tags can be used, and everything needs to be properly escaped.
+;helpText[*] = "Default help text used if not overridden."
+;helpText[en-gb] = "Help text for British English localization."
+
+; Optional help texts that can be displayed on the hold update form. Displayed as is;
+; HTML tags can be used, and everything needs to be properly escaped.
+;updateHelpText[*] = "Hold update default help text used if not overridden."
+;updateHelpText[en] = "Hold update help text for British English localization."
+
 ; The "NewUser" section defines some default values that are used when creating an account
 ; in Alma via its API. This is only relevant if you use the authentication method "AlmaDatabase"
 ; in the "Authentication" section of the "config.ini" file.
--- a/config/vufind/Demo.ini
+++ b/config/vufind/Demo.ini
@@ -140,13 +140,15 @@ defaultRequiredDate = 'driver:0:2:0'
 ; "pickUpLocation", "startDate" and "requiredByDate"
 extraHoldFields = "comments:requestGroup:pickUpLocation:startDate:requiredByDate"
 
-; Optional help texts that can be displayed on the hold form. HTML allowed.
+; Optional help texts that can be displayed on the hold form. Displayed as is;
+; HTML tags can be used, but everything needs to be properly escaped.
 helpText = "Help text for all languages."
 ;helpText[en-gb] = "Help text for English language."
 
-; Optional help texts that can be displayed on the hold update form. HTML allowed.
+; Optional help texts that can be displayed on the hold update form. Displayed as is;
+; HTML tags can be used, but everything needs to be properly escaped.
 updateHelpText = "Hold update help text for all languages."
-;updateHelpText[en-gb] = "Hold update help text for English language."
+;updateHelpText[en] = "Hold update help text for English language."
 
 ; Fields that can be updated for holds (unless the hold is already in transit or
 ; available for pickup). Supported values are "frozen", "frozenThrough" and
--- a/config/vufind/Demo.ini
+++ b/config/vufind/Demo.ini
@@ -140,6 +140,14 @@ defaultRequiredDate = 'driver:0:2:0'
 ; "pickUpLocation", "startDate" and "requiredByDate"
 extraHoldFields = "comments:requestGroup:pickUpLocation:startDate:requiredByDate"
 
+; Optional help texts that can be displayed on the hold form. HTML allowed.
+helpText = "Help text for all languages."
+;helpText[en-gb] = "Help text for English language."
+
+; Optional help texts that can be displayed on the hold update form. HTML allowed.
+updateHelpText = "Hold update help text for all languages."
+;updateHelpText[en-gb] = "Hold update help text for English language."
+
 ; Fields that can be updated for holds (unless the hold is already in transit or
 ; available for pickup). Supported values are "frozen", "frozenThrough" and
 ; "pickUpLocation"
--- a/config/vufind/Demo.ini
+++ b/config/vufind/Demo.ini
@@ -148,7 +148,7 @@ extraHoldFields = "comments:requestGroup:pickUpLocation:startDate:requiredByDate
 ; Optional help texts that can be displayed on the hold update form. Displayed as is;
 ; HTML tags can be used, and everything needs to be properly escaped.
 ;updateHelpText[*] = "Hold update default help text used if not overridden."
-;updateHelpText[en] = "Hold update help text for British English localization."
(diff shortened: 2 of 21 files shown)


config/vufind/Demo.ini
```diff
@@ -140,6 +140,14 @@ defaultRequiredDate = 'driver:0:2:0'
 ; "pickUpLocation", "startDate" and "requiredByDate"
 extraHoldFields = "comments:requestGroup:pickUpLocation:startDate:requiredByDate"
 
+; Optional help texts that can be displayed on the hold form. HTML allowed.
+helpText = "Help text for all languages."
+;helpText[en-gb] = "Help text for English language."
```
demiankatz: I would probably use [en] instead of [en-gb] in the example, since the en-gb language is disabled by default.

module/VuFind/src/VuFind/ILS/Connection.php
```diff
@@ -367,11 +367,10 @@ protected function checkMethodHolds($functionConfig, $params)
                     explode(':', $functionConfig['updateFields'])
                 );
             }
-            if (isset($functionConfig['helpText'])) {
-                $response['helpText'] = $this->getHelpText(
-                    $functionConfig['helpText']
-                );
-            }
+            $response['helpText']
```
demiankatz, line 370: Other calls to getHelpText() are wrapped in an `isset()` check. Would it make sense to be consistent, and either restore that here or remove it elsewhere?

themes/bootstrap3/templates/holds/edit.phtml
```diff
@@ -8,6 +8,9 @@
     . '<li class="active">' . $this->transEsc('hold_edit_title') . '</li>';
 ?>
 <h2><?=$this->transEsc('hold_edit_title')?></h2>
+<?php if ($this->helpText): ?>
+  <p class="helptext"><?=$this->helpText // Intentionally left unescaped?></p>
```
demiankatz: Minor nitpick, but I'd put a period and a space on the end of the comment, because otherwise it looks like a question due to the end-PHP tag.

themes/bootstrap3/templates/record/hold.phtml
```diff
@@ -9,7 +9,7 @@
 ?>
 <h2><?=$this->transEsc('request_place_text')?></h2>
 <?php if ($this->helpText): ?>
-<p class="helptext"><?=$this->helpText?></p>
+  <p class="helptext"><?=$this->helpText // Intentionally left unescaped?></p>
```
demiankatz: Same as above.

config/vufind/Demo.ini
demiankatz: Also, should we say something stronger than "HTML allowed," since the way we are rendering this also puts responsibility on the end user to escape everything properly where applicable.

themes/bootstrap3/templates/holds/edit.phtml
demiankatz: Alternatively, would changing the template variable to $this->helpTextHtml make this more clear everywhere?

demiankatz (changes requested):
Thanks for this! A few comments... and I would agree that it probably makes sense to put the same config comments in all the driver configs, once we've settled on the best example text.

EreMaijala:
@demiankatz Thanks for the good comments. I believe the latest commit addresses them all. I kept the old helpText view variables for back-compat. I'll copy the help text settings to the other files when you've approved the comments.

config/vufind/Demo.ini
demiankatz: Looks like you changed [en-gb] below, but not above... Intentional, or an oversight?

module/VuFind/src/VuFind/Controller/HoldsTrait.php
```diff
@@ -227,7 +227,8 @@ public function holdAction()
         $config = $this->getConfig();
         $homeLibrary = ($config->Account->set_home_library ?? true)
             ? $this->getUser()->home_library : '';
-        $helpText = $checkHolds['helpText'] ?? null;
+        // helpText is only for for backward compatibility:
```
demiankatz: Typo: "for for"

demiankatz:
Thanks, @EreMaijala, just just left a couple more trivial comments. Once you have a chance to address those, I'll run the full test suite (unless you already have) and merge if everything is passing. Certainly looks fine to my eyes.

demiankatz:
It also looks to me like you've made the wiki updates, so I'm assuming that box can now be checked... unless there's something still in progress on your end.

config/vufind/Demo.ini
EreMaijala: Oversight, now fixed. I see that en-gb is in some other ini files as well, so I'll need to take care of them as well.

module/VuFind/src/VuFind/Controller/HoldsTrait.php
EreMaijala: Fixed.

EreMaijala:
@demiankatz Thanks. I've not run the test suite yet. I'll start updating other ini files next. I updated the wiki page, and if the texts are ok, the box can be checked.

demiankatz:
@EreMaijala, the wiki looks good to me. I have my test environment up, so I'll run the tests now and let you know how it goes. If the only commits after that point are adding comments to .ini files, I don't think we need to worry about breaking anything.

config/vufind/Demo.ini
demiankatz: It's certainly not the end of the world to have en-gb in an example; actually, the more I think about it, the more I realize that it might actually be a more useful example than [en], since in this example, English is taken care of by the default case. It might honestly be worth changing this back, and I certainly wouldn't bother changing all the other ones. I apologize for not really thinking this through before commenting.

demiankatz: Maybe the best fix is to say `;helpText[en-gb] = "Help text for British English localization."` in the example. I think that makes everything more clear.

demiankatz: And maybe change "Help text for all languages" to "Default help text used if not overridden."

demiankatz: (If you agree with this in principle, I'm happy to make the changes as penance for changing my mind ;-) )

demiankatz:
@EreMaijala, all tests are passing, so I think we can merge this once we settle on the final shape of the comments. Let me know if you want me to work on any of that!

config/vufind/Demo.ini
EreMaijala: ~Since you can't have both, I wouldn't call it "default". Though that would be a useful feature. Perhaps we need * for default in the array?~

demiankatz: Ahhhh, right! Yes, I think adding support for `*` would make sense and perhaps clarify the whole situation. It seems like having a single hard-coded help text option would be more useful if it went through the translator, but I guess now is not the time to further overcomplicate things. ;-)

EreMaijala:
@demiankatz I've added support for the `*` default key and updated the examples for all relevant drivers. Tests are still passing, but I haven't added a test that would exercise the new default help text support.

config/vufind/Alma.ini
```diff
@@ -65,6 +65,16 @@ defaultPickUpLocation = ""
...
 
+; Optional help texts that can be displayed on the hold form. Displayed as is;
+; HTML tags can be used, and everything needs to be properly escaped.
+;helpText[*] = "Default help text used if not overridden."
+;helpText[en-gb] = "Help text for British English localization."
+
+; Optional help texts that can be displayed on the hold update form. Displayed as is;
+; HTML tags can be used, and everything needs to be properly escaped.
+;updateHelpText[*] = "Hold update default help text used if not overridden."
+;updateHelpText[en] = "Hold update help text for British English localization."
```
demiankatz: This is a recurring typo: `[en]` instead of `[en-gb]` -- seems to have been copy-pasted to several .ini files. I can do a global search-and-replace if that would be helpful.

config/vufind/KohaRest.ini
```diff
@@ -113,7 +123,7 @@ HMACKeys = item_id:holdings_id:StorageRetrievalRequest
 extraFields = item-issue:acceptTerms:pickUpLocation
 
 ; Optional help texts that can be displayed on the request form
-helpText = "Help text for all languages."
+;helpText = "Help text for all languages."
```
demiankatz: Does it make more sense to use `helpText[*]` here now?

demiankatz (commented):
This looks good to me, apart from a couple of very minor observations.

Regarding testing, I'm happy to take a little time (probably next week) to establish a test class for the ILS Connection class so we can test this (and maybe start working on some other tests of this important core class).

config/vufind/Alma.ini
EreMaijala: Would have thought I had been capable of fixing this properly after all the discussion we had... Now it should be right.

config/vufind/KohaRest.ini
EreMaijala: Yes. Changed in VoyagerRestful.ini as well.

EreMaijala:
Note: the removed acceptTermsText in KohaRest.ini is our local thing that had crept in the example.

demiankatz:
Thanks, @EreMaijala -- this looks good and is working correctly for me. It would be nice to get some test coverage on this feature, but I think given current priorities, it's okay to merge as-is, so that's what I'm going to do.

PR #2076, opened 2021-08-26
Update cancelHolds method comments.

EreMaijala:
Makes the comments reflect the way the driver in question actually works.


demiankatz:
Thanks!

PR #2077, opened 2021-08-27
Always unset updateDetails unless the value is usable.

EreMaijala:
Previously an empty string would have been kept and used for the id in the list.


demiankatz:
Makes sense; thanks for the fix!

PR #2078, opened 2021-08-27
corrects logic for first and last name detection for space seperated …

olli-gold:
…names from PAIA

We have noticed, that the PAIA driver handles space seperated names wrong, if someone has more than one first name. I have changed the logic, so that only the last part of the name is taken as the last name, while everything before that is handled as the first name. I know, that this is not 100% accurate, but I guess it's better than the treatment right now.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/PAIA.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/PAIA.php
@@ -1378,12 +1378,8 @@ class PAIA extends DAIA
             $lastname = $nameArr[0];
         } else {
             $nameArr = explode(' ', $username);
-            $firstname = '';
             $lastname = array_pop($nameArr);
-            foreach ($nameArr as $value) {
-                $firstname .= ' ' . $value;
-            }
-            $firstname = trim($firstname);
+            $firstname = trim(implode(' ', $nameArr));
         }
 
         // TODO: implement parsing of user details according to types set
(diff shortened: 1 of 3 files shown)


module/VuFind/src/VuFind/ILS/Driver/PAIA.php
```diff
@@ -1378,13 +1378,12 @@ protected function paiaParseUserDetails($patron, $user_response)
...
-            $lastname = '';
-            array_shift($nameArr);
+            $firstname = '';
+            $lastname = array_pop($nameArr);
             foreach ($nameArr as $value) {
-                $lastname .= ' ' . $value;
+                $firstname .= ' ' . $value;
             }
-            $lastname = trim($lastname);
+            $firstname = trim($firstname);
```
demiankatz: I think you could condense this code considerably by getting rid of the initialization and foreach loop and simply saying:

`$firstname = trim(implode(' ', $nameArr));`

What do you think?

demiankatz (commented):
Thanks, @olli-gold! One suggestion below. Also, would it be relatively easy to add some test coverage for various name formatting cases? There's an existing test suite for PAIA, so it might be good to get some coverage on the cases you're concerned about.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
olli-gold: You are right, that sounds like a great simplification and is absolutely equivalent.

olli-gold:
Right, adding this test case to the test suite sounds reasonable to me.

demiankatz:
Thanks, @olli-gold, this makes sense to me and the test looks good, so I'm going ahead and merging it now.

PR #2079, opened 2021-08-27
Make CacheTrait generally available, use in HoldsController

EreMaijala:
Note: Two separate commits to keep the refactoring clear from the added code.

Tests are passing.

### DIFF
--- a/module/VuFind/src/VuFind/Controller/HoldsController.php
+++ b/module/VuFind/src/VuFind/Controller/HoldsController.php
@@ -29,6 +29,7 @@
  */
 namespace VuFind\Controller;
 
+use Laminas\Cache\Storage\StorageInterface;
 use Laminas\ServiceManager\ServiceLocatorInterface;
 use VuFind\Exception\ILS as ILSException;
 use VuFind\Validator\Csrf;
@@ -45,6 +46,8 @@ use VuFind\Validator\Csrf;
  */
 class HoldsController extends AbstractBase
 {
+    use \VuFind\Cache\CacheTrait;
+
     /**
      * CSRF validator
      *
@@ -55,13 +58,18 @@ class HoldsController extends AbstractBase
     /**
      * Constructor
      *
-     * @param ServiceLocatorInterface $sm   Service locator
-     * @param Csrf                    $csrf CSRF validator
+     * @param ServiceLocatorInterface $sm    Service locator
+     * @param Csrf                    $csrf  CSRF validator
+     * @param StorageInterface        $cache Cache
      */
-    public function __construct(ServiceLocatorInterface $sm, Csrf $csrf)
-    {
+    public function __construct(
+        ServiceLocatorInterface $sm,
+        Csrf $csrf,
+        StorageInterface $cache
+    ) {
         parent::__construct($sm);
         $this->csrf = $csrf;
+        $this->setCacheStorage($cache);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -281,7 +305,9 @@ protected function getPickupLocationsForEdit(
         int $checkLimit = 0
     ): ?array {
         $catalog = $this->getILS();
-        $holds = $catalog->getMyHolds($patron);
+        // Get holds from cache if available:
+        $holds = $this->getCachedData($this->getCacheId($patron, 'holds'))
```
demiankatz, line 305: Do we need to invalidate this cache when the user cancels or edits a hold?

demiankatz (commented):
Thanks, @EreMaijala! I've gone ahead and cherry-picked the first commit to dev (with minor modifications, to eliminate the accidental `sleep` that was added and some unnecessary annotations in comments) -- see https://github.com/vufind-org/vufind/commit/18f82cbfff93d1c3b1a987dc9c3b2b49bd933919. I had one question about this part before we merge it.

module/VuFind/src/VuFind/Controller/HoldsController.php
EreMaijala: The presumption is that any edit action gets initiated from the holds list. Otherwise we wouldn't have the correct update or cancel details remembered either. And the list is refreshed after an edit or cancel, so it shouldn't be necessary to invalidate the cache manually. Unless, of course, the user is doing something weird such as using multiple browser tabs to manage stuff, but even then the edit action redirects to the list and refreshes things.

demiankatz: Okay, that makes sense to me!

module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -147,10 +155,26 @@ public function listAction()
 
             $driversNeeded[] = $current;
         }
+        // Cache the current list of requests for editing:
+        $this->putCachedData(
+            $this->getCacheId($patron, 'holds'),
+            $driversNeeded,
+            600
```
demiankatz: Is it worth using a class property here and below instead of the magic number "600"?

module/VuFind/src/VuFind/Controller/HoldsController.php
```diff
@@ -147,10 +155,26 @@ public function listAction()
...
+        );
 
         // Get List of PickUp Libraries based on patron's home library
         try {
-            $view->pickup = $catalog->getPickUpLocations($patron);
+            $view->pickup = $this->getCachedData(
+                $this->getCacheId($patron, 'pickup')
```
demiankatz: Is it worth putting the cache key in a variable since we calculate it twice close together here and in line 173?

demiankatz (changes requested):
A couple very minor suggestions before merging this....

module/VuFind/src/VuFind/Controller/HoldsController.php
EreMaijala: Done.

module/VuFind/src/VuFind/Controller/HoldsController.php
EreMaijala: I ended up setting the cacheLifetime property in constructor. Ok or would you prefer a more explicit property?

EreMaijala:
@demiankatz Thanks for fixing the delay as well. I thought I removed that, but obviously I didn't. Optional delay might be a useful thing, but for another PR..

module/VuFind/src/VuFind/Controller/HoldsController.php
demiankatz: That seems fine for this case; we can sort it out later if we ever need other types of caching in this controller.

demiankatz:
Thanks -- looks good to me now!

PR #2080, opened 2021-08-27
Make abstract search command truly abstract; improve tests.

demiankatz:
As @aleksip pointed out in #2036, the AbstractBase command was not actually abstract. I've refactored according to our conversation there. I also fixed several tests that were weirdly relying on command subclasses to use mocks instead, which seems significantly more appropriate.

Once this is approved, I'll go ahead and incorporate this work into #2036 and #2037.

### DIFF
--- a/module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
+++ b/module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
@@ -128,9 +128,8 @@ abstract class CallMethodCommand extends AbstractBase
         if ($this->addParamsToArgs) {
             $callArgs[] = $this->params;
         }
-        $this->result
-            = call_user_func([$backendInstance, $this->method], ...$callArgs);
-
-        return $this->finalizeExecution();
+        return $this->finalizeExecution(
+            call_user_func([$backendInstance, $this->method], ...$callArgs)
+        );
     }
 }
--- a/module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
+++ b/module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
@@ -131,6 +131,6 @@ abstract class CallMethodCommand extends AbstractBase
         $this->result
             = call_user_func([$backendInstance, $this->method], ...$callArgs);
 
-        return parent::execute($backendInstance);
+        return $this->finalizeExecution();
     }
 }
--- a/module/VuFind/src/VuFindTest/Feature/MockSearchCommandTrait.php
+++ b/module/VuFind/src/VuFindTest/Feature/MockSearchCommandTrait.php
@@ -1,0 +1,80 @@
+<?php
+
+/**
+ * Trait for tests requiring mock search Command objects
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2021.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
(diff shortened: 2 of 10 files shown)


module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
```diff
@@ -131,6 +131,6 @@ public function execute(BackendInterface $backendInstance): CommandInterface
         $this->result
             = call_user_func([$backendInstance, $this->method], ...$callArgs);
 
-        return parent::execute($backendInstance);
+        return $this->finalizeExecution();
```
aleksip: I think that this change needs to be done to all current commands. And that the change is unfortunately a BC breaking one, although I wonder if anyone has actually written their own commands yet. 😄

demiankatz: Unless I'm missing something, `CallMethodCommand` is the only subclass of `AbstractBase`, and everything else extends `CallMethodCommand`. I think it makes sense to use `parent::execute` in subclasses of `CallMethodCommand`, because that's actually doing the "call method" action -- some subclasses can use the default logic while others wrap it with extra details, but in all cases, it makes sense in the context of inheritance. It was less logical to have a defined execute method in `AbstractBase` because that command has no specific purpose.

As for a backward breaking change, I suspect that we're safe from anyone having defined commands just yet. Since this isn't officially released code yet, I think we're free to change it -- and now is the time to get it right, before 8.0 comes out!

Do you still think that further changes are needed, or would you be satisfied to merge this as-is?

Even if you do want more changes, if you're only concerned about changing subclasses of `CallMethodCommand`, perhaps we can merge this first as an incremental step so I can start making progress on updating the other command-related pull requests before doing additional refactoring. What do you think?

aleksip: Ah indeed, seems I wasn't thinking clearly.

Some existing commands (`RandomCommand`, `RetrieveBatchCommand`) end with 

```php
$this->executed = true;
return $this;
```

which could be replaced with 

```php
return $this->finalizeExecution();
```

Also now looking at the commands more closely, `finalizeExecution()` could perhaps also have an optional `$result` parameter, and set the result as well? Since this is probably a very common thing to do in the final stage of execution.

If we are to enforce a one command per backend method rule and/or a one backend method call per command rule, some of the current commands are a bit problematic, since `GetIdsCommand`, `RandomCommand`, `RetrieveBatchCommand` and `WorkExpressionsCommand` call other backend methods as well.

I guess that applying the command pattern now makes it even more clear that the extra logic was not just bloating the `Service` class, but should not have been in the backend in the first place. That is, the only logic in the backend should be in the actual backend classes and their helper classes. I'm not sure about this, but seems to make sense?

In any case, in regards to your last question, I have no objections to merging in small steps!

demiankatz: Thanks, @aleksip, I like your suggestions regarding finalizeExecution, and I'll apply them before merging this. Right now I'm working through the refactoring of #2037 and should have something to show there fairly soon. The proper way to apply the command pattern is beginning to become more clear, but as you say, it's taking a bit of experience to really get a feel for it.

Regarding commands that call other backend methods, you're right that it may require some additional thought and work. The idea for at least some of these was to have seamless fallbacks in situations where a method is missing... but as you point out, this could have implications for triggering of listeners, etc.

demiankatz: Improvements applied in 0852d24. Running test suite now; will merge if everything passes.

PR #2081, opened 2021-08-30
Fix limitOrderOverride

xmorave2:
This setting does not use facet values, as it should

TODO
- [x] Merge #2088 before this, so the test passes

### DIFF
--- a/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AdvancedSearchTest.php
+++ b/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AdvancedSearchTest.php
@@ -28,6 +28,7 @@
 namespace VuFindTest\Mink;
 
 use Behat\Mink\Element\Element;
+use Behat\Mink\Session;
 
 /**
  * Mink test class to test advanced search.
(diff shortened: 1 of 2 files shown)


demiankatz:
Thanks, @xmorave2 -- could you give me an example of how to reproduce the problem that this is fixing? I just want to be sure I understand exactly what is going on (and maybe I can write a test to prevent future regressions).

xmorave2:
> Thanks, @xmorave2 -- could you give me an example of how to reproduce the problem that this is fixing? I just want to be sure I understand exactly what is going on (and maybe I can write a test to prevent future regressions).

Of course. I found it in this context:
- language facet field being translated
- language facet field used as limit on advanced search
- limitOrderOverride set to some values from index For example "Czech::English::German"
- UI language set to Czech so the used displayText is "Čeština", "Angličtina" and "Němčina"

then language limit on advanced search form is strictly alphabetically sorted

demiankatz:
Thanks for the clarification, @xmorave2 -- I have confirmed that your fix works, and I have added an integration test to prevent regressions. Unfortunately, writing the test revealed some bugs/limitations in VuFind's config writer (which is used to set up the test environment). I fixed those problems in #2088, but that will have to be merged before this, to ensure that tests pass. Do you mind reviewing #2088 so I can merge it? If you're not comfortable with making a call on it, I can also seek another reviewer.

PR #2082, opened 2021-08-30
WorkKeys.java: Optimize caching behaviour

mtrojan-ub:


import/index_java/src/org/vufind/index/WorkKeys.java
```diff
@@ -67,16 +67,9 @@
...
-            if (!this.transliterators.containsKey(transliterationRules)) {
-                this.transliterators.put(
-                    transliterationRules,
-                    Transliterator.createFromRules("workkeys", transliterationRules, Transliterator.FORWARD)
-                );
-            }
-        }
         final Transliterator transliterator = transliterationRules.isEmpty()
-            ? null : this.transliterators.get(transliterationRules);
+            ? null : this.transliterators.computeIfAbsent(transliterationRules, value ->
```
demiankatz: Would it be slightly more clear to use `transliterationRules` in place of `value` here? I realize that this will likely work the same whether the key is passed in from the callback function's parameter or by just inheriting it from the scope above it... but it feels more clear to me if we use the function parameter. (But feel free to correct me if my rusty Java is leading me to make inappropriate suggestions). :-)

demiankatz: ...and maybe it would be less confusing/ambiguous if we used something shorter, like just `rules` instead of `value` in this line and in place of `transliterationRules` below -- that makes it more explicit exactly what is happening, I think.

EreMaijala (commented):
I'll happily defer the review to @demiankatz even with rusty skills since my Java skills don't allow for proper review here.

import/index_java/src/org/vufind/index/WorkKeys.java
mtrojan-ub: If `value ->` is not specific enough, i suggest we should instead change it to `transliterator ->`, because that is what the value represents.

mtrojan-ub: Oh no, sorry, of course that doesnt work because `transliterator` is the name of the outer variable, but `rules` is definitely not what the value represents. Any other suggestions?

demiankatz: Would `key` make sense, since the reason for the callback is to use a key to create a new value? Then I think it's pretty clear that `transliterationRules` is the key being used, but that the callback is a function to derive a new value based on a key.

mtrojan-ub: I'm not sure if the naming problem might be caused by misunderstanding this piece of code :wink:

The notation `value -> ...` is not the same as in PHP, where `key => value` is used to describe a key and a value in an associative array. In Java it's a so-called Lambda Expression. `value` is an alias for the function call to `Transliterator.createFromRules()`, which will be passed as a lambda callback to the computeIfAbsent function. The lambda callback will only be executed to generate the value if the key doesn't yet exist in the Hashtable. That's why neither `key` nor `transliterationRules` would be correct in this case.

demiankatz: @mtrojan-ub, my point here is that you are using a lambda function, but you are ignoring the parameter to that function and instead pulling a variable from the calling scope, and I thought it might be clearer to actually use the function parameter instead. Perhaps that is not the preferred best practice, though, or maybe I am still misunderstanding something.

And just to make absolutely clear what I'm talking about, assume this hypothethical example:

<pre>
Set<int> squares = new LinkedHashSet<int>();
int two = 2;
int squareOfTwo = squares.computeIfAbsent(two, i => i * i);
</pre>

In this case, I provide a function to square a number, and if the result is not already in the cache, it gets computed. I think this is more clear than saying:

<pre>
int squareOfTwo = squares.computeIfAbsent(two, i => two * two);
</pre>

...even though I think the two pieces of code will behave exactly the same.

Does that make any sense? And in any case, apologies for taking up so much time on what I think is just a minor question of style. :-)

mtrojan-ub: Thanks @demiankatz, now i finally realized what you mean! I basically just added the lambda expression in the first place because without it, it would lead to a "no suitable method found" error... but you're correct that it might be clearer to use the alias as well in the inner function call. I hope it's correct now.

demiankatz:
Thanks, @mtrojan-ub, this looks good to me now, and tests are passing, so I'm going to merge it!

PR #2083, opened 2021-08-31
SierraRest: Disable keepalive as it may cause trouble with Sierra.

EreMaijala:
@bpalme Can you try to see if this eliminates the issue you reported with the Sierra connection?


EreMaijala:
According to feedback this did not solve the issue. Closing.

PR #2085, opened 2021-08-31
Improve search service command object error handling.

demiankatz:
This was split off from #2036 to simplify review/merging.

### DIFF
--- a/module/VuFindSearch/src/VuFindSearch/Service.php
+++ b/module/VuFindSearch/src/VuFindSearch/Service.php
@@ -356,11 +356,19 @@ class Service
                 $args
             );
             if (!$response->stopped()) {
+                // We need to construct our error message differently depending
+                // on whether or not we have a command object...
+                $commandClass = isset($args['command'])
+                    ? get_class($args['command']) : 'null';
+                $context = $args['context'] ?? $commandClass;
+                $backend = isset($args['command'])
+                    ? $args['command']->getTargetBackendName()
+                    : ($args['backend'] ?? $backend);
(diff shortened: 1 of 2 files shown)


demiankatz:
@aleksip, I think this addresses some error handling details that we overlooked in the command pattern refactoring. Any objections to merging it?

module/VuFindSearch/src/VuFindSearch/Service.php
```diff
@@ -356,11 +356,19 @@ function ($o) {
                 $args
             );
             if (!$response->stopped()) {
+                // We need to construct our error message differently depending
+                // on whether or not we have a command object...
+                $commandClass = isset($args['command'])
+                    ? get_class($args['command']) : 'null';
+                $context = $args['context'] ?? $commandClass;
```
aleksip: I might be missing something here, but why `$commandClass` instead of `$args['command']->getContext()`?

aleksip (commented):
Ah, well spotted!

module/VuFindSearch/src/VuFindSearch/Service.php
demiankatz: Probably because when I discovered this problem, I was debugging an issue and hadn't fully refreshed my memory on how any of this works. :-) I'll correct it shortly; thanks for the correction!

demiankatz:
Okay, I have significantly improved this code based on @aleksip's comments, and I have added a test to cover the command object case, so I am confident this is now ready to merge.

PR #2086, opened 2021-08-31
Refactor EDS code to use search service.

demiankatz:

TODO

### DIFF
--- a/module/VuFind/src/VuFind/Autocomplete/Eds.php
+++ b/module/VuFind/src/VuFind/Autocomplete/Eds.php
@@ -29,6 +29,8 @@
  */
 namespace VuFind\Autocomplete;
 
+use VuFindSearch\Service;
+
 /**
  * EDS Autocomplete Module
  *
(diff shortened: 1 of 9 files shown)


demiankatz:
@aleksip, I think this one should be pretty uncontroversial (and I've tested it with my own EDS account, so I know that the refactoring was successful). The only point that might be up for discussion is where to put backend-specific commands. I've put them inside the `VuFindSearch\Backend\EDS\Command` namespace, but I think arguments could be made for other locations (like `VuFindSearch\Command\EDS`, for example). I don't really have a strong feeling; I'm just trying to separate things that seem potentially general purpose (like some of the currently-Solr-only commands in other PRs that might theoretically extend to other backends in the future, which I have left in the `VuFindSearch\Command` namespace) from things that are clearly tied to a specific backend (like these very EDS-oriented commands).

aleksip (approved):
I don't really have a strong opinion on the location, since modern IDEs make it easy to find subclasses wherever they are, and both possibilities you mentioned are logical in their own way. So I am happy with the current `VuFindSearch\Backend\EDS\Command`.

demiankatz:
Great, thanks, @aleksip... and after a bit more thought, I do favor this current approach over my proposed alternative, if only because it keeps the backend-specific code all grouped together within the same namespace, which makes it slightly easier to work with if we ever deprecate/remove an obsolete backend or decide to break things out into separate projects.

PR #2087, opened 2021-08-31
Improve/add tests for search commands.

demiankatz:
I believe that this PR should get us to 100% test coverage for the VuFindSearch\Command namespace (though some of our current coverage is a side effect of other tests, and it may be beneficial to add more unit tests for specific commands). This also makes some existing test code slightly more concise.


demiankatz:
@aleksip and @EreMaijala, I think this should be a fairly uncontroversial PR since it just increases test coverage. However, I thought the two of you might want to have a conversation about it, since it highlights one of the existing commands that makes extra backend calls, which could potentially be problematic. I'm of course happy to join that conversation (either here, on Slack, or via video chat) if it would be useful. My feeling is that it's probably not a dire problem, but it's something we might as well think through while we're doing all this related work.

aleksip:
@demiankatz We discussed this and agreed that I would open a related PR which should not block merging this PR.

demiankatz:
Thanks, @aleksip, I'll merge this one now so you can build upon it when time permits.

PR #2088, opened 2021-08-31
Improve array support in config writer.

demiankatz:
This adds support for associative arrays in the config writer, and fixes a bug related to multi-valued, "normal" arrays.


xmorave2:
Just for information: in PHP 8.1 will be function for checking if array is list (or associative): https://wiki.php.net/rfc/is_list

xmorave2 (approved):
Great work @demiankatz, thanks!

demiankatz:
Thanks, @xmorave2, and good point about `array_is_list` -- I've added a TODO comment as a reminder that we can eventually simplify this code once the minimum PHP version is raised to 8.1.

PR #2089, opened 2021-09-01
Open an external link instead of loading it into the lightbox

hajoseng:
If there is an external Link for example on the help-pages, the lightbox script tries to load it into the lightbox causing a violation against the same-origin-policy and hence an error. Such links shoud be just opened. I'm not very into JavaScript so probably there is a better solution (eg. by unbinding an event). I pullrequest this to raise this issue.

### DIFF
--- a/themes/bootstrap3/js/lightbox.js
+++ b/themes/bootstrap3/js/lightbox.js
@@ -123,7 +123,7 @@ VuFind.register('lightbox', function Lightbox() {
       return;
     }
     // An external link should be just opened
-    if (obj.url.indexOf(window.location.hostname) == -1) {
+    if (obj.url.indexOf(window.location.hostname) === -1) {
       window.open(obj.url, "_blank").focus();
       return true;
     }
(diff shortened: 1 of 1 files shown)


xmorave2:
@hajoseng You can use target='_new'

If we want to support the _blank value, than the change should be done here I think: https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/js/lightbox.js#L238-L251

Maybe @crhallberg can say more about this

demiankatz:
I'm requesting a review from @crhallberg so he can comment on the general problem as well as the specific solution. I'm also not sure what, if anything, should be done about the LGTM comment above. Additionally, it looks like there's an eslint warning that should be fixed (a use of `==` where `===` would be preferable).

crhallberg:
Your implementation would work but Demian is correct: I would expect this to be added as an exception at the top of constrainLink.

demiankatz:
@hajoseng, it looks like a copy of this branch got pushed to the vufind-org repo in addition to the subhh repo. Do you mind if I remove it? I don't want to delete it without your permission in case it contains any changes missing from this version.

hajoseng:
@demiankatz Oh yes, please remove the copy; it got there accidently.

hajoseng:
@xmorave2 @crhallberg It took me a while to understand: Using target="_new" oder no target solves the problem. That's fine for me and if it's ok for you I would close the pullrequest.

demiankatz:
No objections here!

PR #2090, opened 2021-09-01
Refactor WorldCat code to use search service.

demiankatz:
More code broken off from #2036.

TODO


demiankatz:
@aleksip, this follows a similar pattern to other recent PRs and is working fine with my real WorldCat credentials.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat/Command/GetHoldingsCommand.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Execute command on backend.
+     *
+     * @param BackendInterface $backend Backend
+     *
+     * @return CommandInterface Command instance for method chaining
+     */
+    public function execute(BackendInterface $backend): CommandInterface
+    {
+        if (!($backend instanceof \VuFindSearch\Backend\WorldCat\Backend)) {
```
aleksip: I might have missed these in previous reviews, and could be remembering/understanding things incorrectly, but I believe that a key idea behind the event based system and resolving backends by name is that listeners listening to `EVENT_RESOLVE` can even replace the backend. And this would mean that the backend identifier probably should be the only way to identify the backend.

I must admit that I still do not fully understand why connectors are exposed via the backends, if they should not be called directly. Based on my limited knowledge about them, connectors seem more like internal helper classes. So in this case for example it would seem best to add `getHoldings()` as a backend method and use CallMethodCommand.

aleksip: Looking at #2091 and #2092, the above comments would seem to apply to them too.

I think that out of the two things pointed out, the backend identifying issue is something that probably should be changed, but possible connector code refactoring could be left to a later time, since it does not change the way the command is called or used.

demiankatz: @aleksip, the purpose for the type checking here is not to identify that a specific instance of the backend was provided, but to ensure that the provided backend is capable of supporting the subsequent method calls. I would expect that even if the user swapped out a backend using a listener, the replacement backend would have to be the same type as the original backend (or a subclass, which would also work). I think it's appropriate to throw an exception if the command receives a totally unexpected class, regardless of what caused that to happen, since the alternative is going to be a PHP fatal error due to an illegal method call. (This is essentially an interface check, but I felt these were too special-purpose to justify defining a separate interface... I could take the time to do that if you feel it makes things more clear, though).

As for why connectors are exposed, it's likely for convenience, but it seems to encourage bad behavior. Perhaps we should look at making the getConnector methods protected in the future, but I think that's beyond the scope of what I want to attempt right now, with the 8.0 release right around the corner.

I do agree, in any case, that moving some of this logic to backend methods may clean things up a bit. I don't have much time to work on this today, but I'll see if I can do a bit more polishing tomorrow!

aleksip: As a related thought: I wonder if we should change `CommandInterface::getTargetBackendName()` to `getTargetBackendIdentifier()` to match `BackendInterface::getIdentifier()`?

That is, if I have understood everything correctly and these indeed are the same thing. Looks like it would be a small change and this would be the time to do it.

Edit: #2096

demiankatz: Yes, that definitely makes sense.

aleksip: > the purpose for the type checking here is not to identify that a specific instance of the backend was provided, but to ensure that the provided backend is capable of supporting the subsequent method calls

Ah yes, sorry, indeed this does need to be done too! However, I guess the identifier should still be checked, in case there are several instances? So both checks should be done, just like in `CallMethodCommand`?

Maybe this is nitpicking, if two instances of a WorldCat backend make no sense at all (I have no idea). 😄 But I do see some value in consistency, as developers new to the codebase can find different ways of doing things confusing.

demiankatz: I see what you're saying -- I think this is just another argument for creating backend methods instead of accessing the connector in the command, and then using CallMethodCommand, since then we get all the necessary checking "for free" and everything looks more consistent. I'll refactor as soon as time permits.

demiankatz:
@aleksip, I've converted this as we discussed; mind giving it one last look?

aleksip (approved):
Looks good to me!

PR #2091, opened 2021-09-01
Refactor BrowZine code to use search service.

demiankatz:
More code split off from #2036.

TODO


demiankatz:
@aleksip, more of the same here, and once again, I have tested that the refactored code works as expected with the real BrowZine service.

demiankatz:
@aleksip, I've updated this one to match #2090, so I think it's ready to go as well, with your approval! I'll do some more work on #2092 and #2093 tomorrow.

aleksip (commented):
Looks good to me, with the constructor returns removed.

demiankatz:
Oops, that was silly of me -- thanks for the catch, @aleksip!

PR #2092, opened 2021-09-01
Refactor MapSelection to use search service.

demiankatz:
More code split off from #2036.

TODO
- [x] Refactor command to contain less logic
- [x] Add test coverage
- [x] Open ticket about adding cursorMark support (for future improvement) (see [VUFIND-1501](https://vufind.org/jira/browse/VUFIND-1501))


demiankatz:
@aleksip, this is my first attempt at improving the refactoring of the MapSelection module. Rather than creating a single overly-specific command, I instead created a generic command for calling the connector's search() method and returning a raw JSON response instead of doing the usual record collection processing. This also bypasses "normal" search events, which I think is potentially desirable as well. (These types of special-case lookups probably need to be handled differently by custom listeners).

It's very possible there's still room for improvement, though, so I'm awaiting your feedback before adding tests and other finishing touches -- just want to see if you like this direction or have different ideas.

demiankatz:
I suspect that the pattern (and command) used here will also help with #2093, but I'm going to wait on the review here before spending more time on refactoring there.

demiankatz:
(By the way, I see that I assigned this to @aleksip yesterday -- that was an accident, I just meant to request a review. Sorry for any confusion!)

demiankatz:
@aleksip, I've put some more work into this one -- since it is a fairly common need to perform searches with custom parameters and then deal directly with the JSON response instead of using the processed collection object, I refactored the `search` method so that it utilizes a `rawJsonSearch` method, then created a new command that allows users to short-circuit the process and get directly to the raw JSON. If you think this approach seems reasonable, I will add test coverage here, merge this, and then apply the same pattern to #2093. If you have other thoughts, let me know!

aleksip:
This looks good to me, with the general exception of internal backend calls made to methods that are also exposed as commands.

I continue to wonder about these calls, since having them does seem to mean that event listeners will not be able to catch all searches for example, at least not without full knowledge of internal backend implementations. So if for example a listener wanted to change a parameter for all search operations, this would seem impossible to do for internal search calls?

However, it looks like not allowing internal calls to methods exposed as commands would mean moving lot of logic away from the backend. So I wonder if there should be another class of backend events, which would not be command specific but function specific? So in addition to `SearchCommand` and `RawJsonSearchCommand` events triggered by the `Search` class there would be a general `SEARCH` event of some kind, triggered by the backend implementation classes. There would be a general event like this for every internally called method also exposed as a command.

In any case I don't think there is anything stopping us from going ahead with the current implementation as it is and making improvements later on.

aleksip (approved):

EreMaijala:
A question: How much do we expect listeners to be able to do? I mean, if something happens inside a backend, how complicated can it get to support hooking into that? I'm just wondering where the line should be drawn since some things might be better done with overriding actual code than hooking to events.

aleksip:
That is a very good question. I don't really know enough about how all the backends are used to be able to give an answer.

In general there is no point in adding features if there is no use for them, or if using them would always be more complicated than another approach.

However, if we don't have a feature available, the challenge is to remember the possibility of implementing and using that feature at some later point.

demiankatz:
@aleksip / @EreMaijala, I think this is going to be complex and challenging now matter how we approach it. Ultimately, I think the question is, what is the target events are hooked to? Is it a particular backend method, or is it a particular command? I think in the past, the target was "a particular search service method," but we are now changing it to "a particular command." In either situation, there is a close but inexact correlation with backend methods, which creates edge cases where things are potentially ambiguous or confusing.

I think, as long as we agree that this is the convention, we can live with internal backend calls, because they don't violate the convention. So, for example, to use the search() vs. rawJsonSearch() example: as written here, the search() method calls the rawJsonSearch() method. The idea of this is that search() is used for "normal" VuFind searches where we fully process the results into a collection, and rawJsonSearch() is used for special cases where we want to pull some details out of Solr in a non-standard format. If we write an event to hook rawJsonSearch(), it will miss all the calls coming from search(). But we can very easily revise such a listener to look for either of those two possible scenarios, and then we're fine. A lot of the listeners that hook search are about adding features to standard search results (like spell checking, etc.). These things are generally just a waste of cycles if applied to the low-level calls, so I think it's useful to be able to hook only the "regular" search() calls independently. The problem, obviously, is that you have to have a pretty deep understanding of the code in order to make the judgement calls necessary to make effective use of the events... but I don't know how we avoid that. It's inherently complicated. For me, the bottom line is that using events is generally a last resort, and there is often an easier and clearer way to solve most problems... but they are useful for some things. The best we can do is try our best not to make it even more complicated than it has to be.

I'm not sure if that helps at all, but that's sort of where my head is. And, of course, keep in mind that I didn't design the event system in the first place, so my opinions may not be the best ones. ;-)

aleksip:
The `context` of the events has not changed with the introduction of commands: it is still set to be name of the backend method by default. The `context` is still used in code like

```php
if ($context === 'search')
```

This could of course be changed to

```php
if ($command instanceof SearchCommand)
```

but since we already have and use the more flexible `context`, and can even combine it with command `instanceof` checks, I see no benefit in removing it.

In the documentation of Laminas EventManager `target` is described as the object that triggered the event. "This gives event listeners access to the calling object, which can often be useful."

While it can be said that it is the command that triggered the event, the actual _object_ triggering the event is always the search `Service` instance. 

We already have the command also available as an event parameter, and all listener code is accessing the command via this parameter. Accessing the command was done like this for backwards compatibility reasons, but would now make it easy to set the search `Service` instance as the event target instead. This could indeed often be useful, since in many cases it could mean that the search `Service` does not have to be passed in the listener's constructor. Just a thought! 😄 

The publish-subscribe pattern is especially useful in frameworks and platforms that can be assembled from modules coming from multiple different sources. In the case of VuFind, we don't really have a contributed module ecosystem, and in the case of the search backend for example it can just be easier to create a backend subclass and add the functionality there. However, functionality added or changed this way is much harder to share or combine.

Anyway, I think we are fine with the way the event system works now. Events from low level calls might be useful in implementing things like access control, but adding low level events can wait until there is an actual use case for them.

demiankatz:
Thanks, @aleksip, that all makes sense -- and I really need to be careful about using words like "context" and "target" in the context of events, because they have specific meanings that I think I easily fumble. ;-)

It does seem like it would be worthwhile to make the search service the target of the event, especially if that can be done without introducing backward compatibility issues. Do you think it's worth making a PR for that in advance of 8.0, or should we deal with that in a future release? I'd like to try to get all changes stabilized by the end of this week if at all possible, so no pressure if it's safer to wait... but if it's truly simple, it might be worth squeezing in!

In any case, for today, my priority is going to be finishing up test coverage for this PR and then applying equivalent changes to the other outstanding work so we can wrap up the bulk of the search refactoring. More later, after I've had time to do all that!

demiankatz:
Okay, I've added test coverage for the new command, as well as doing a bare minimum amount to establish a unit test for MapSelection (though a lot more work is needed to add meaningful tests -- I don't have time right now but wanted to at least get the ball rolling). I also did a bit more code simplification to MapSelection. I'm going to merge now so I can move on to updating the remaining search-related PRs.

PR #2093, opened 2021-09-01
Refactor Solr TreeDataSource to use search service.

demiankatz:

TODO
- [x] Revise to match strategy used in #2092 (after that PR is merged)


demiankatz:
@aleksip, I've refactored this to use the new command from #2092, and I think it's ready to go now. Do you mind double-checking?

aleksip (approved):
LGTM!

PR #2095, opened 2021-09-02
Make handling of missing locale settings more robust.

EreMaijala:
Throws a proper exception if language in [Site] is not defined or [Languages] section is missing.


EreMaijala:
👍

demiankatz:
Thanks, @EreMaijala! I added c8bc010 to avoid a notice if the setting is totally missing; other than that, looks good. Merging as soon as CI passes.

PR #2096, opened 2021-09-02
Rename CommandInterface method

aleksip:
To better match the method names in `BackendInterface`.

In addition to being more consistent, and hopefully making code easier to understand, makes the command method name less backend specific, which can be nice if we want to use the command pattern in other places too.


demiankatz:
Thanks!

PR #2098, opened 2021-09-03
Remove unused/broken multi-breadcrumb functionality.

demiankatz:
I recently discussed with @EreMaijala the fact that the multi-breadcrumb functionality in VuFind is never used and has some problems (missing translations, improper escaping, etc.).

This code existed as part of the public VuDL implementation which was subsequently removed due to lack of interest/support. It is extremely unlikely that anyone is using it in production.

I have recently been working on upgrading Villanova's local instance of VuDL + VuFind, and in the new version of the code, I reimplemented our multi-breadcrumbs by creating a local template to render them, and then simply assigning the output of rendering that template to `$this->layout()->breadcrumbs`. This is much more self-contained, and less potentially confusing, than adding a whole bunch of additional layout variables, complex logic in the layout template, and the possibility of breadcrumbs being either an array or a string. If we ever need something like functionality this in core VuFind in the future, I would favor using that new strategy instead of the old one being removed here. Another argument in favor of eliminating this obsolete code.


demiankatz:
@crhallberg, any objections?

EreMaijala (approved):
Nice!

crhallberg:
No objections!

demiankatz:
Thanks, both of you. Fastest turn-around on a PR review ever?

EreMaijala:
Could be :)

PR #2099, opened 2021-09-03
MARC 21 XML type

mtrojan-ub:


module/VuFind/src/VuFind/RecordDriver/Feature/MarcAdvancedTrait.php
```diff
@@ -889,7 +889,7 @@ public function getXML($format, $baseUrl = null, $linker = null)
                 'http://www.loc.gov/standards/marcxml/schema/MARC21slim.xsd',
                 'http://www.w3.org/2001/XMLSchema-instance'
             );
-            $xml->record->addAttribute('type', 'Bibliographic');
+            $xml->record->addAttribute('type', $this->xmlType);
```
demiankatz, line 899: I like the overall idea here, but I wonder if it would be more clear to define the property in the trait, and then set it in the constructors of the classes using the trait. Alternatively, at a bare minimum, it may be a good idea to add `?? 'Bibliographic'` here, to account for the possibility of the trait being used in a class that does not define the property at all.

demiankatz (changes requested):
Thanks for this! Just one thought...

demiankatz:
(Based on the CI failure, it looks like phpstan shares my concerns :-) ).

module/VuFind/src/VuFind/RecordDriver/Feature/MarcAdvancedTrait.php
mtrojan-ub: You're right! I'm gonna change the logic as you proposed.

demiankatz:
Thanks, @mtrojan-ub, I think this looks great!

PR #2100, opened 2021-09-03
[VUFIND-1461] Add change tracking to XSLT examples.

demiankatz:
TODO
- [x] Implement for all examples
- [x] Add a note to the [relevant video](https://vufind.org/wiki/videos:xml_change_tracking_and_email_alerts) about the new feature
- [x] Resolve [VUFIND-1461](https://vufind.org/jira/browse/VUFIND-1461) when merging


demiankatz:
@EreMaijala, the need for XML change tracking came up during the recording of [one of our Community Call videos](https://vufind.org/wiki/videos:xml_change_tracking_and_email_alerts). I've just implemented this for the OJS example; if you approve of this implementation, I'll go ahead and copy-and-paste the logic into all of the other examples. Figured it would make sense to get your review BEFORE I duplicate it many times, though.

EreMaijala (commented):
Looks good to me. One potential change: would it make sense to make the identifier field a variable as well, so that it can be defined with the other params? Then you could copy the actual selector block as is to other xslt files and just update the param.. Further on, if this param was present, it should probably control what gets indexed in Solr's id field as well.

demiankatz:
Thanks, @EreMaijala, that's a very good idea. I've gone ahead and revised the OJS example to use a variable for the identifier field -- seems like we might as well make the whole thing configurable all at once because the changes are so closely related.

My only question is whether the property names should be something different; maybe "selector" is implying the wrong thing and "tag_name" would be a more accurate suffix. I'm not sure whether more clever things are feasible in XSLT or not.

Please let me know what you think, and once we've settled on a pattern here, I'll copy and paste it to the other eligible configs.

EreMaijala:
@demiankatz This revision looks good to me!

demiankatz:
@EreMaijala, I believe this is done now -- I updated all the eligible examples and tested that they all still work. Do you mind giving it one last glance?

EreMaijala (approved):
Looks good to me!

PR #2102, opened 2021-09-05
Add ability to filter sitemap results.

demiankatz:
I just realized that for some reason, Villanova's Digital Library VuFind instance has custom filtering functionality built into its sitemap generator which was never contributed upstream. This PR addresses that shortcoming so that the feature will be available in future. While it might be possible to do more or less the same thing by event-hooking the getIds backend method, I think there's value in being able to specifically target sitemap generation with a configuration.

@EreMaijala, do you mind taking a look? I'd like to include this in 8.0 if possible.

### DIFF
--- a/module/VuFind/src/VuFind/Sitemap/Plugin/Index/CursorMarkIdFetcher.php
+++ b/module/VuFind/src/VuFind/Sitemap/Plugin/Index/CursorMarkIdFetcher.php
@@ -103,7 +103,7 @@ class CursorMarkIdFetcher extends AbstractIdFetcher
         string $backend,
         string $cursorMark,
         int $countPerPage,
-        array $filters = []
+        array $filters
     ): array {
         // If the previous cursor mark matches the current one, we're finished!
         if ($cursorMark === $this->prevCursorMark) {
--- a/module/VuFind/src/VuFind/Sitemap/Plugin/Index/CursorMarkIdFetcher.php
+++ b/module/VuFind/src/VuFind/Sitemap/Plugin/Index/CursorMarkIdFetcher.php
@@ -27,7 +27,6 @@
  */
 namespace VuFind\Sitemap\Plugin\Index;
 
-use VuFindSearch\Backend\Solr\Backend;
 use VuFindSearch\Backend\Solr\Response\Json\RecordCollectionFactory;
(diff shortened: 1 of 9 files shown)


module/VuFind/src/VuFind/Sitemap/Plugin/Index/CursorMarkIdFetcher.php
```diff
@@ -96,13 +95,15 @@ public function setupBackend(string $backend): void
...
+     * @param array  $filters      Filters to apply to the search
      *
      * @return array
      */
     public function getIdsFromBackend(
         string $backend,
         string $cursorMark,
-        int $countPerPage
+        int $countPerPage,
+        array $filters = []
```
EreMaijala: Does this need the default? Since the whole plugin was just added, I think changing the signature would be quite alright at this point.

module/VuFind/src/VuFind/Sitemap/Plugin/Index/TermsIdFetcher.php
```diff
@@ -73,14 +73,19 @@ public function setupBackend(string $backend): void
...
+     * @param array  $filters      Filters to apply to the search
      *
      * @return array
      */
     public function getIdsFromBackend(
         string $backend,
         string $lastTerm,
-        int $countPerPage
+        int $countPerPage,
+        array $filters = []
```
EreMaijala: ditto

EreMaijala (approved):
Looks good, just the one question.

Additionally, I think at some point the question whether it makes sense to continue to support the terms method makes sense. To me it feels like the terms method has too many caveats, but if you think it still makes sense to maintain it, I'm ok with that.

demiankatz:
Thanks, @EreMaijala -- I gave the parameter a default value mainly to keep the tests passing, since I was adding this as a quick proof-of-concept to start with. Since you approve, I've gone ahead and made the parameter required, and updated the test suite. As long as CI passes, I'm going to merge it now.

Regarding support for the terms option, I agree that its value is diminishing, and if it becomes a burden to maintain in the future, it might be a candidate for removal... but for the moment, since the work is already invested, it does remain a potentially highly performant option for the simple cases where it can be used, so it doesn't seem to hurt to keep it around.

PR #2103, opened 2021-09-05
Make OAI-PMH cursormark configurable.

demiankatz:
There are some situations where using a cursormark for the OAI-PMH server is not desirable (because it is incompatible with other Solr features). This PR is designed to make the behavior configurable.

TODO
- [x] Finish implementing (add to config file, etc.)
- [x] Test thoroughly


demiankatz:
@EreMaijala, we discussed this a few weeks ago, and I finally had time to implement it -- this appears to work, though it's not very elegant at all. If you have any better ideas, I'd be happy to hear them. I'm working through the weekend to make sure VuFind 8.0 stuff is ready on time, so at this point I'm a bit tired and not at my most creative. ;-)

config/vufind/config.ini
```diff
@@ -1406,6 +1406,11 @@ url             = https://www.myendnoteweb.com/EndNoteWeb.html
 ; Omit this setting to return all deleted records. This can be useful for long-lived
 ; systems with many deleted records, to prevent full harvests from becoming unwieldy.
 ;
+; use_cursormark controls whether or not to use Solr's cursormark functionality for
```
EreMaijala: *Pedant mode:* I'd prefer `use_cursor` since "using cursors" is the pagination method and cursorMark (with capital M) is the parameter name. useCursorMark is fine internally.

EreMaijala (changes requested):
Looks ok to me. I'd add to the comments an example (result grouping) as a reason to not use a cursor so that it's not necessary to dig for the reason in the future.

demiankatz:
Thanks, @EreMaijala; I've adjusted accordingly!

PR #2104, opened 2021-09-06
Add support for conditional extra parameters in searchspecs.yaml.

EreMaijala:
This makes it possible to add e.g. boosts by defining them once for all relevant search types. Any parameters are only added at most once regardless of the number of search terms in the query.

Includes tests, but they certainly don't cover all possible permutations of the rules.


config/vufind/searchspecs.yaml
```diff
@@ -70,6 +70,61 @@
 #
 # ...etc.
 #
+# ExtraParams allows extra parameters to be (conditionally) added to the Solr search
```
demiankatz: Might it make sense to use the word Global in the name somewhere, and/or make the comments more explicit, to make it entirely clear that this is a top-level section that is not attached to a handler configuration?

config/vufind/searchspecs.yaml
```diff
@@ -70,6 +70,61 @@
...
+#         - Title
+#         - Author
+#       - SearchTypeNotIn:
+#         - Series
+#
+# Value can be a single string or a list of strings to repeat the parameter.
+#
+# Available conditions:
+#
+# - NoBoost: None of the search types specifies a boost of their own in DismaxParams
```
demiankatz: Would it be more generic/flexible to use NoMatchingDismaxParams, with a value indicating which parameters to skip, rather than defining separate conditions for specific parameters? Or would it make sense to add NoMatchingDismaxParams and then make NoBoost / NoBoostQuery / NoBoostFunctions "semantic sugar" on top of it for commonly-used configurations?

module/VuFindSearch/src/VuFindSearch/Backend/Solr/QueryBuilder.php
```diff
@@ -154,10 +161,12 @@ public function build(AbstractQuery $query)
                     $oldString = $string;
                     $string = $handler->createBoostQueryString($string);
 
-                    // If a boost was added, we don't want to highlight based on
-                    // the boost query, so we should use the non-boosted version:
-                    if ($highlight && $oldString != $string) {
-                        $params->set('hl.q', $oldString);
+                    if ($string !== $oldString) {
```
demiankatz: I'm not sure what the purpose of this refactoring is (other than making the comparison more strict). Did you refactor this as part of work that then got rolled back, or is there a particular reason to use nested `if` statements instead of an `&&` condition?

module/VuFindSearch/src/VuFindSearch/Backend/Solr/QueryBuilder.php
```diff
@@ -180,9 +189,136 @@ public function build(AbstractQuery $query)
         }
         $params->set('q', $string);
 
+        // Handle any extra parameters:
+        foreach ($this->extraParams as $extraParam) {
+            if (empty($extraParam['param']) || empty($extraParam['value'])) {
+                continue;
+            }
+            if (!$this->checkParamConditions($query, $extraParam['conditions'] ?? [])
```
demiankatz, line 195: Any reason to have two separate `if` statements here rather than combining all the OR'ed conditions into one? I think it would be equally readable if each `||` had its own line...

module/VuFindSearch/src/VuFindSearch/Backend/Solr/QueryBuilder.php
```diff
@@ -180,9 +189,136 @@ public function build(AbstractQuery $query)
...
+        }
+        $searchTypes = $this->getSearchTypes($query);
+        $searchTypeOk = null;
+        foreach ($conditions as $condition) {
+            if (is_array($condition)) {
+                $values = reset($condition);
+                $condition = key($condition);
+                switch ($condition) {
+                case 'SearchTypeIn':
+                    if (null === $searchTypeOk) {
```
demiankatz: I'm a little confused about the interaction between SearchTypeIn and SearchTypeNotIn. I guess my head wants them to be either AND'ed together or OR'ed together, and I think what's actually happening is not quite either of those scenarios... is there a particular use case that led you to this implementation?

demiankatz (commented):
Thanks, @EreMaijala, I think this is looking pretty good, but I had a few comments and questions!

config/vufind/searchspecs.yaml
EreMaijala: Renamed to GlobalExtraParams.

config/vufind/searchspecs.yaml
EreMaijala: Yes, done. Named the key NoDismaxParams.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/QueryBuilder.php
EreMaijala: Oops, that was left over from a previous version.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/QueryBuilder.php
EreMaijala: No other than that the latter one doesn't fit on a single line inside the first if statement. I can revise, but this seemed cleaner.

demiankatz: I don't have strong feelings about it; if you prefer it separate, that's fine with me.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/QueryBuilder.php
EreMaijala: It is a bit complicated because a single search may contain multiple search types, and I wanted to be able to do something like "Don't boost most recent material for PublishDate search unless there are AllFields or Title fields in the search as well." But I see it doesn't quite make sense. I've now simplified the rules and added AllSearchTypesIn as a condition that requires all search types to be in the given list. I think this would cover the use cases while being simple enough even for me to understand later on. :) But if you see more room for improvement, please let me know!

config/vufind/searchspecs.yaml
```diff
@@ -70,6 +70,63 @@
...
+#
+# Available conditions:
+#
+# - NoDismaxParams: None of the search types specifies any of the listed parameters
+#   in DismaxParams
+# - SearchTypeIn      At least one of the search types must be in the given list
+# - AllSearchTypesIn  All of the search types must be in the given list
+# - SearchTypeNotIn   None of the search types must be in the given list
+#
+# Note that conditions are processed in the order of occurrence. SearchTypeNotIn is
```
demiankatz: This note is no longer true and should be removed, correct?

config/vufind/searchspecs.yaml
```diff
@@ -70,6 +70,63 @@
 #
 # ...etc.
 #
+# GlobalExtraParams allows extra parameters to be (conditionally) added to the Solr
+# search request:
+#
+# GlobalExtraParams:
```
demiankatz, line 78: You have examples below with explanations, but you also have this unspecified example up here. Should this get a label too, or should it be made more generic so that it demonstrates all available syntax rather than looking like a real-world example?

module/VuFindSearch/src/VuFindSearch/Backend/Solr/QueryBuilder.php
```diff
@@ -180,9 +187,119 @@ public function build(AbstractQuery $query)
...
+    /**
+     * Get an array of search types used in the given search
+     *
+     * @param AbstractQuery $query Query
+     *
+     * @return array
+     */
+    protected function getSearchTypes(AbstractQuery $query): array
+    {
+        $types = [];
```
demiankatz: Nitpick, but I'd move this inside the `if` since it doesn't apply outside.

demiankatz: Or alternatively could we eliminate the initialization by using an array_reduce(), like:

<pre>
$callback = function ($carry, $item) {
    return array_merge($carry, $this->getSearchTypes($item));
};
return array_unique(array_reduce($query->getQueries(), $callback, []));
</pre>

Apologies if I'm misremembering how array_reduce works, but this seems like it might be a perfect use case for it.

demiankatz (commented):
A few minor thoughts; I haven't taken a close look at the test suite yet, but I'll do so after you've responded to these comments.

config/vufind/searchspecs.yaml
EreMaijala: Yes, removed.

config/vufind/searchspecs.yaml
EreMaijala: Done.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/QueryBuilder.php
EreMaijala: Thanks, that's an excellent idea. I added your example, and it works great!

demiankatz:
@EreMaijala, I refactored your test case into two tests using data providers, because I think that makes the testing a little more granular and improves error reporting. Data providers are a pretty new thing for me, so this is the first time I've tried this... but I've been meaning to learn about them, and this seemed like the perfect use case to try. I'm glad I made the effort, because I think it has revealed a flaw in the tests: in both of your "Single value, no extra params" cases, no actual assertions are made, so the tests aren't actually testing anything. I think perhaps these cases need to be moved to separate tests that make different types of assertions, but I wasn't sure exactly what the best approach might be. Do you have any ideas?

EreMaijala:
@demiankatz I wasn't even aware of data providers.. Your solution is pretty elegant. I fixed the expectations so that all tests have assertions.

demiankatz:
Thanks, @EreMaijala -- not sure how I managed to attach the same data provider to both tests (I actually thought I was being careful NOT to do that), but glad you caught that!

Also glad to have demonstrated data providers, as I'm sure we'll both find them useful in future... and there are probably a number of existing tests that would benefit from conversion. I'll try to remember to fix them up as I notice them!

PR #2039, opened 2021-07-30
Accessibility: improved login forms

al-pu:
small changes to fulfill accessibility standards 

1. commit: connecting labels to inputs correctly
2. commit: add password autocomplete hint for login
3. commit: add username / email autocomplete hint for login
4. commit: add password autocomplete hint for reset password

Please have a look, if forms still behave as expected.

PR #2105, opened 2021-09-06
Fix catalog login with MultiBackend.

EreMaijala:
PR #2039 caused cataloglogin.phtml to display two username fields with the MultiBackend driver. This fixes it and also adds resetting of the password field when selecting an ILS with username+password login after having one with email login selected.


demiankatz:
Thanks!

demiankatz:
It would probably be a good idea to get some Mink test coverage on this functionality when time permits... but at the moment, it does not. :-)

PR #2106, opened 2021-09-07
Refactor SwitchQuery module to use search service.

demiankatz:

TODO
- [x] Adjust based on recent discussion
- [x] Add tests
- [x] Update changelog

### DIFF
--- a/module/VuFind/src/VuFind/Recommend/SwitchQuery.php
+++ b/module/VuFind/src/VuFind/Recommend/SwitchQuery.php
@@ -28,7 +28,8 @@
  */
 namespace VuFind\Recommend;
 
-use VuFind\Search\BackendManager;
+use VuFindSearch\Backend\Solr\Command\GetLuceneHelperCommand;
+use VuFindSearch\Service;
 
 /**
  * SwitchQuery Recommendations Module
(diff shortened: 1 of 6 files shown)


demiankatz:
@aleksip and @hajoseng, based on discussion on #2036, I was seriously considering refactoring this code to have a couple of separate commands for performing the actions of the Lucene syntax helper individually... but between the fact that time is running out and that there is some shared functionality with the Summon backend which I am no longer able to test, it seemed safer for now to just keep this as-is, to reduce the chances of disruption. If either of you have strong feelings (or the willingness to improve the implementation this week), I'm certainly open to a different approach... but otherwise, I think it may be best to stick with what we have here, and perhaps consider a more significant overhaul of the Lucene syntax helper in a future release.

hajoseng:
@demiankatz I agree to leave it just as it is. In my opinion we should rather consider an overhaul of the LuceneSyntaxHelper what shouldn't be done in a hurry.

aleksip (approved):
I agree that we can keep this as is now, and continue related work after the 8.0 release. This also seems very much related to #1886.

demiankatz:
Thanks, @aleksip, merging now. Speaking of #1886, are you planning on finishing that up before 8.0, or are you letting it wait until later? (I'm fine either way -- just want to know what to expect).

aleksip:
@demiankatz After seeing this PR and considering that @hajoseng is also proposing a related overhaul, I think #1886 can wait until later. It introduces a `QueryAnalysisInterface`, and looking at `LuceneSyntaxHelper` I'm thinking we could have an interface, perhaps with a different name, that combines both use cases.

demiankatz:
Thanks, @aleksip! Do you think we should open a JIRA ticket to capture this conversation and discuss further, or should we just repurpose the existing PR #1886 as a place to work through this project?

PR #2107, opened 2021-09-07
Aleph: parse position in queue from item status

xrosecky:


xrosecky:
I also removed if condition that was always true.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -1284,52 +1284,62 @@ public function getMyHolds($user)
...
-                    'create' => $this->parseDate($create),
-                    'status' => $status,
-                    'position' => ltrim($seq, '0')
-                ];
+            $type = "hold";
+            $seq = null;
+            $item_status = preg_replace(
+                "/\s[\s]+/",
+                " ",
+                (string)$item->{'status'}
```
demiankatz: I think you can just use `$item->status` here; the `{'...'}` syntax is only necessary when the property name contains a `-` character.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -1284,52 +1284,62 @@ public function getMyHolds($user)
...
-                    'id' => $this->barcodeToID($barcode),
-                    'expire' => $this->parseDate($expire),
-                    'holddate' => $holddate,
-                    'delete' => $delete,
-                    'create' => $this->parseDate($create),
-                    'status' => $status,
-                    'position' => ltrim($seq, '0')
-                ];
+            $type = "hold";
+            $seq = null;
```
demiankatz: Was the old `z37-sequence` logic completely wrong, or would it be useful to maintain it as a fallback/default value?

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -1284,52 +1284,62 @@ public function getMyHolds($user)
...
+            $item_status = preg_replace(
+                "/\s[\s]+/",
+                " ",
+                (string)$item->{'status'}
+            );
+            if (preg_match(
+                "/Waiting in position ([0-9]+) in queue; "
```
demiankatz: Is this guaranteed to be consistent, or might it vary based on i18n or Aleph settings?

demiankatz (commented):
Thanks for this; just a few minor comments/questions.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
xrosecky: It is wrong, because it is sequence number used to differentiate holds for the same item.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
xrosecky: It may depend on language in Aleph and even Aleph administrators can change the text in translation files. Maybe we should make it configurable?

xrosecky: I made it configurable.

demiankatz:
Thanks, @xrosecky -- it looks like a long line in the file is now causing the build to fail for style checks, but other than that, I think this looks good. I'll give @welblaud a little bit of time to comment in case he has anything to add, but I suspect this is uncontroversial and low-risk.

demiankatz:
This is also approved. Thanks again, merging now!

PR #2108, opened 2021-09-07
Refactor backend event targets and Solr error listeners

aleksip:
Here is a PR changing all backend event targets to the `Service` instance, as discussed in #2092. 

I changed the target of error events as well for consistency. This is a new breaking change if someone has subclassed one of the Solr error listener classes. I considered it to be relatively small breakage, but I can also refactor the PR to keep the error event target unchanged.

The refactoring of Solr error listeners and related tests highlighted that there is still refactoring to do on the other listeners to replace actual backend instances with backend identifiers. But I believe that this can be done after the 8.0 release.

Interestingly, it looks like backend identifiers are only set in `BackendManager::get()` and are `null` in freshly created backend instances. I might be missing something important, but it seems odd and even undesired that a backend instance can be created without an identifier, especially when it would be easy to set in the constructor? So this could be another refactoring task for the future.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
+++ b/module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
@@ -153,7 +153,7 @@ abstract class AbstractSolrBackendFactory implements FactoryInterface
      * Create service
      *
      * @param ContainerInterface $sm      Service manager
-     * @param string             $name    Requested service name (unused)
+     * @param string             $name    Requested service name
      * @param array              $options Extra options (unused)
      *
      * @return Backend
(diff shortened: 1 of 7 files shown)


demiankatz:
Thanks for this; it seems fine to me, and all tests are passing. I'll add more notes to the changelog.

Regarding the setting of backend identifiers, I suspect the current implementation is related to a limitation of Zend Framework factories back when this was initially designed, but a great deal has changed since then. Since we now use the backend identifier as the service name, and the service name is always passed to the factory, I don't see a reason why we can't adjust all the factories as you have done here to immediately initialize the identifier. The only caveat is that the current pattern we use for backends differs from almost everything else in VuFind, in that we normally use class names as service names, with shorter names as aliases... except here we do not use class names. This is because multiple services share the same class but simply use different factories to configure instances differently. I'm not sure if this is something we want to try to change in the future for improved consistency, or if we want to fully embrace it since it is convenient for solving this problem. I personally don't mind keeping it the way it is and taking advantage of it, but the inconsistency of it all does trouble me sometimes. In any case, I hope that provides a little more context for you!

PR #2109, opened 2021-09-08
Make EDS connector not override HTTP client's adapter.

EreMaijala:
Currently the connector forces curl adapter and sets curl-specific options, but there appears to be nothing actually requiring curl. This change switches to standard options. The Laminas HTTP client handles location redirects as well, so there should be no need for that. 

I've tested this with the default adapter and curl adapter, and everything seems to still work.

### DIFF
--- a/module/VuFindSearch/src/VuFindSearch/Backend/EDS/Connector.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/EDS/Connector.php
@@ -79,7 +79,6 @@ class Connector extends Base implements LoggerAwareInterface
             [
                 'timeout' => $settings['timeout'] ?? 120,
                 'sslverifypeer' => false,
-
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks, @EreMaijala -- if I had to guess, the original developers of the EDS code were having trouble getting SSL to work, and they added this code as a brute force method of evading those problems... and I never took the time to question why it was there. I'll do a bit of testing on this end when time permits, but I don't anticipate problems; I'll merge it in time for 8.0 unless I encounter an unexpected complication.

demiankatz:
This is working fine for me, so I'm going to go ahead and merge it. I suspect a strong argument could be made for entirely removing the `sslverifypeer` option, but in the interest of minimal change close to a release, I'm not going to worry about it at this time.

PR #2111, opened 2021-09-08
Consistent variable names and code style for backend commands

aleksip:
This PR attempts to make all backend instances `$backend` and all backend identifiers `$backendId`.

This touches a lot of files, but I think consistent variable names would improve the developer experience.

Additionally there are some other small improvements like added parameter types.

### DIFF
--- a/module/VuFindSearch/src/VuFindSearch/Service.php
+++ b/module/VuFindSearch/src/VuFindSearch/Service.php
@@ -85,7 +85,7 @@ class Service
     /**
      * Constructor.
      *
-     * @param EventManagerInterface $events Event manager (optional)
+     * @param ?EventManagerInterface $events Event manager (optional)
      *
      * @return void
      */
@@ -110,11 +110,11 @@ class Service
         // All other legacy event parameters are accessible via the command object.
         $args = ['command' => $command];
 
-        $backendInstance = $this->resolve($command->getTargetIdentifier(), $args);
+        $backend = $this->resolve($command->getTargetIdentifier(), $args);
 
         $this->triggerPre($this, $args);
         try {
-            $command->execute($backendInstance);
+            $command->execute($backend);
         } catch (BackendException $e) {
             $args['error'] = $e;
             $this->triggerError($this, $args);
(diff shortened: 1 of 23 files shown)


module/VuFindSearch/src/VuFindSearch/Service.php
```diff
@@ -182,92 +182,105 @@ public function getIds(
...
      *
-     * @param string   $backend  Search backend identifier
-     * @param string   $id       Id of record to compare with
-     * @param array    $workKeys Work identification keys (optional; retrieved from
-     * the record to compare with if not specified)
-     * @param ParamBag $params   Search backend parameters
+     * @param string    $backendId Search backend identifier
+     * @param string    $id        Id of record to compare with
+     * @param ?array    $workKeys  Work identification keys (optional; retrieved
+     *                             from the record to compare with if not specified)
```
demiankatz, line 270: You don't need to align wrapped comments; they can go all the way left. It doesn't really hurt here, but going to the left margin as a general rule prevents things from getting too cramped when the types and parameter names are very long.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/SearchServiceTest.php
```diff
@@ -781,7 +781,7 @@ public function __construct($backendMock)
      *
      * @return Service
      */
-    protected function resolve($backend, $args)
+    protected function resolve($backendId, $args)
```
demiankatz, line 784: I'm not sure why this SearchServiceMock class even exists -- it would be better addressed with "normal" PHPUnit mocks. I've revised this test to eliminate it, but unfortunately I don't have permission to push my changes back to your repo. Right now, I'm running the full test suite; if everything passes, I'll just go ahead and merge everything to dev manually and then close this PR.

demiankatz:
Okay, tests passing, so I've gone ahead and merged this manually as ba9aabc5eb0bb478cb340589e558ea769eb2272d (with my minor revisions incorporated).

module/VuFindSearch/src/VuFindSearch/Service.php
aleksip: Ah, I see `PEAR.Commenting.FunctionComment.ParamCommentAlignment` has been excluded. I have to fix it so that my IDE picks that up.

PR #2113, opened 2021-09-09
Improvements to feedback forms

EreMaijala:
Note: Three commits that build on each other.

The first commit is some cleanup with no functional changes. The second one makes it possible to change e.g. the name of the email from field for a specific form e.g. for routing purposes. The third one adds date field support.


demiankatz:
This looks fine to me; I'll just add a note to the changelog about the changes to the \VuFind\Form\Form constructor.

PR #2114, opened 2021-09-09
Refactor CallMethodCommand and add backend command argument accessors

aleksip:
I was refactoring Finna's listener code and realized that the new style events did not provide all the event parameters that the old ones did (e.g. `query`). So I am afraid that one more command refactoring was required... 😅 

However, I am quite happy with how this turned out. Now the concrete command classes have complete control over the arguments, and `$addParamsToArgs` is no longer required. Accessing an argument array by index is no longer required either, and everything seems much clearer, including the role of command classes.

The `getArguments()` construct allows for event listeners to change non-object arguments as well, as long as the command provides a setter method. I have only added getters for now, and setters can be added later if needed.

### DIFF
--- a/module/VuFind/tests/unit-tests/src/VuFindTest/Sitemap/Plugin/Index/TermsIdFetcherTest.php
+++ b/module/VuFind/tests/unit-tests/src/VuFindTest/Sitemap/Plugin/Index/TermsIdFetcherTest.php
@@ -92,7 +92,7 @@ class TermsIdFetcherTest extends \PHPUnit\Framework\TestCase
         return function ($command) use ($expectedCursorMark) {
             $this->assertEquals(
                 [$this->uniqueKey, $expectedCursorMark, $this->countPerPage],
-                $command->getArguments()
+                array_slice($command->getArguments(), 0, 3)
             );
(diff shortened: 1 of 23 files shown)


demiankatz:
@aleksip, this looks very good to me --in retrospect, I think we should probably have taken the `$addParamsToArgs` constructor argument as a sign that we had a design flaw... that annoyed me every time I used it, but not quite enough to make me stop and think about a better solution. In any case, this is a very nice fix to that problem. More verbose, obviously, but also more clear.

One suggestion, just for organizational purposes: can we put the traits in a VuFindSearch\Command\Feature namespace, just to keep them separate from the commands themselves? I've been trying to use that convention for traits elsewhere, and I think it helps keep our directories uncluttered while making it easier to quickly survey available traits. I'm happy to make the change if you agree in principle but don't have time!

aleksip:
@demiankatz Yes, `$addParamsToArgs` annoyed me a lot too! 😄 

I have now moved the traits as you requested.

demiankatz:
Thanks, @aleksip -- all tests are passing and I can't see any problems, so I'm merging now!

PR #2115, opened 2021-09-09
Simplify unnecessary array_push call.

demiankatz:
This is a truly trivial change, but it bothered me that we were updating the module array using two different methods in adjacent lines of code. There is no reason for this, and the simple `$modules[] =` assignment is slightly more efficient.


demiankatz:
This is such a trivial change that I almost committed it directly to dev, but in the interest of following protocols, I'm asking for a review. Do you mind double-checking me here, @aleksip?

aleksip (approved):
Looks good to me!

demiankatz:
Thanks!

PR #2116, opened 2021-09-09
Add support for including browser's user agent in the feedback email.

EreMaijala:
This includes a translatable string 'User Agent', but I've left out the translations as I'm not sure if it's worth translating, and without a translation this could be included in 8.0. :) If you feel that the translation is needed, this can wait until after 8.0.


demiankatz:
Thanks, @EreMaijala, I agree that we can safely merge this into 8.0 without the translation, since it's an internal/administrative string that is unlikely to cause any harm. However, do you mind opening a new PR that just adds the translation string, so we don't forget about it during the next translation cycle? I'll merge that separately after 8.0 is released.

PR #2118, opened 2021-09-10
Validate submitted form option values

aleksip:
Adds validation for submitted form option values (checkbox, radio and select elements), so that only the values specified by the form are accepted.

### DIFF
--- a/module/VuFind/src/VuFind/Form/Form.php
+++ b/module/VuFind/src/VuFind/Form/Form.php
@@ -415,6 +415,7 @@ class Form extends \Laminas\Form\Form implements
             ]
         ];
 
+        $elementObjects = parent::getElements();
         foreach ($this->getElements() as $el) {
             $isCheckbox = $el['type'] === 'checkbox';
             $requireOne = $isCheckbox && ($el['requireOne'] ?? false);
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/Form/Form.php
```diff
@@ -415,6 +415,7 @@ public function getInputFilter(): InputFilterInterface
             ]
         ];
 
+        $elementObjects = parent::getElements();
```
demiankatz, line 421: Is this a sign that we need to rename our getElements() method to avoid a naming conflict, or are we intentionally overriding the existing one? It's a bit unclear why we would do things this way -- if it really is intentional, perhaps a comment to explain why would help prevent future confusion.

demiankatz:
Thanks, @aleksip -- one question/suggestion before I merge this!

module/VuFind/src/VuFind/Form/Form.php
EreMaijala: I'd support renaming to avoid total confusion. At least I've already stumbled on this a couple of times.

demiankatz:
Thanks, @aleksip, I'll merge this now, but would welcome a refactoring PR to rename the method, if time permits. I'm hoping to freeze code for 8.0 at end of day today, but I can let a few things slide into Monday if necessary.

module/VuFind/src/VuFind/Form/Form.php
aleksip: I agree that this is very confusing and should be changed. We have some plans to use the form classes in a way that probably would require refactoring the form system anyway, so the question is should this wait until then, or would it be ok to commit this with the clarifying TODO comment?

aleksip:
@demiankatz Ah, thanks, that was quick! 😄

module/VuFind/src/VuFind/Form/Form.php
demiankatz: I don't think that the changes here make the existing situation significantly worse -- they just make it more obvious that a problem exists. Thus, I'm comfortable with merging them and addressing the rest separately. The question is whether it's worth doing a quick fix for 8.0 now, or waiting longer and doing a bigger refactor later.

PR #2120, opened 2021-09-10
Rename Form class method

aleksip:
I don't know why the parent class method was originally overridden, perhaps not even intentionally, but it looks like renaming the method did not require much changes in the codebase.

It is a breaking change though, so in that sense it is good to get it in now.


demiankatz:
Looks straightforward; merging now, and I'll add a note to the changelog.

PR #2121, opened 2021-09-10
Create new truncation mechanism; apply to existing truncation functionality.

demiankatz:
This PR splits off some work from #1791 for possible inclusion in VuFind 8.0, leaving behind work that is still in progress or requires additional translations.

TODO
- [x] Run full test suite
- [x] Migrate WorldCatIdentities functionality
- [x] Fix outstanding problems in TopFacets
- [x] Update changelog when merging (removal of old show/hide mechanism; removal of top_cols setting for TopFacets)


demiankatz:
I've asked for reviews from @aleksip, @crhallberg and @EreMaijala -- curious if there are opinions on whether or not it's worth trying to merge this into VuFind 8.0. Obviously this is unlikely to get done before Monday, but I'm willing to squeeze it in before end of day Monday if there is significant interest (and no concern about the minor changes to look and feel that it introduces).

If we merge this, I also want to fix the TopFacets and WorldCatIdentities templates if possible... I'm willing to let that slide to closer to the last minute if necessary, since they're (hopefully) less impactful changes.

EreMaijala:
I'll let @aleksip say what he thinks about the inclusion to 8.0.

aleksip:
@demiankatz This looks otherwise fine to me, but I am unable to say anything about the extent or impact of the TopFacets and WorldCatIdentities changes without putting some time into studying them.

We do have interest in getting this merged, but for us it does not have to be in 8.0.

demiankatz:
Thanks, @aleksip -- the WorldCatIdentities template is of very little importance (since it's a proof-of-concept demo that it is likely nobody actually uses). It may be a good example of nested collapsible areas, though, and as such may be worth adapting to be sure the library is flexible enough.

TopFacets is a bit strange, and I think the way that the collapsing works there currently is suboptimal. There's a TODO at the top of the template saying that it needs some attention. I think it would be a good idea to spend a little time making that better. In theory, it would be great to do that before the 8.0 release, but in practice, I'm not sure that we have sufficient time remaining.

I plan to talk about this with the Project Management Committee today to see what others think in terms of priorities... my feeling is that it would be nice to merge the improved facet functionality IF we are confident that the truncation code is flexible enough... but it's possible that these other two template conversions could reveal shortcomings and require changes... and we probably don't have time to get through them quickly enough. I'll keep you posted on outcomes!

crhallberg (approved):

demiankatz:
I was able to get WorldCatIdentities ported over, which revealed a bug in the truncation logic (now fixed)... but TopFacets still needs work.

demiankatz:
@aleksip / @crhallberg, I made some progress on implementing TopFacets, but things aren't quite right. Please take a look at 972513e -- I had to switch from "height" to "max-height" and add an "overflow: hidden" to get things to work at all using the pixel-based mechanism (which seems most appropriate for this use case). I'm not sure why height had no effect, but it simply didn't work. I'm not sure if fixing this case is breaking a different one. Also, the height calculations are totally wrong. I think it's calculating based on line height, but I think that's overlooking margins and padding and such, so things are wildly incorrect. Any suggestions on the right way to fix that?

If we can get these issues sorted out, I think we might be able to merge this in time for 8.0... but I'm not sure how complicated the last mile is going to be.

I'm also a bit concerned about the complexity of truncate.js, which is doing two very different things in one long block of code. I wonder if it might actually be easier to understand if we totally forked the logic into the pixel-based and container-element-based paths. That might be slightly more redundant, but I think it might be significantly more readable.

aleksip:
@demiankatz Yes, unfortunately the CSS changes caused problems with my test cases. I made a [PR](https://github.com/demiankatz/vufind/pull/2) to your repo containing alternative partial fixes.

The height-based truncation first tries to use the height of the first child element, if one exists. I have improved the original to allow for any `block` or `inline-block` elements (the original did not pick up the `span`s used here) and also to use the outer height, including paddings and borders. Margins are problematic though, since they can extend beyond the container component. I would rather not add any more complexity to the JS code, but rather refactor the CSS (and possibly the template) so that there are no margins. This remains to be done, as I thought this approach should be discussed first.

Although the script could add `overflow: none`, I would suggest adding it in CSS files (as done in the PR), and only to use JS for setting dynamically calculated properties.

I have also added a separate inner `div` for the truncate function, because having both collapse and truncate functions working on the same `div` seemed to cause problems.

demiankatz:
Thanks for all the work on this, @aleksip, I think it is looking really good, and I'd be comfortable merging this soon. (When the time comes, I will try to break this into multiple commits, since we now have a lot of unrelated things mixed together here... as usual, what seemed like a simple job turned out to be a bit more complex).

I did discover one new bug, though I'm not sure how significant it is; certainly worth fixing if the fix is easy: the new mechanism doesn't quite play well with the responsive design features. Here are steps to reproduce the problem:

1.) Shrink your browser window so that the top facets collapse into an expandable area.
2.) Expand the collapsed area
3.) Hit the "more" button to show all the facets
4.) Collapse the collapsible area
5.) Maximize your browser window -- now you have a garbled display.

aleksip:
@demiankatz Should be fixed now!

demiankatz:
Wonderful, thank you! I believe this is now ready to merge, but I'm interested to hear if @crhallberg or @EreMaijala have any last-minute thoughts.

aleksip:
@demiankatz I'm afraid I found one more problem, for some reason the 0 rows feature is not working any more. Will look into it now.

aleksip:
@demiankatz False alarm, sorry! Everything working fine after all. 😅

demiankatz:
@aleksip and @EreMaijala: I have just force-pushed this branch to create a slightly more helpful commit history; I'll rebase this onto dev first thing in the morning unless the test suite fails (I'm running it now!) or one of you objects before then. Please let me know if you have any concerns!

demiankatz:
All tests are passing!

EreMaijala (approved):
Looks good to me. :)

demiankatz:
Merged. Thanks again!

PR #2123, opened 2021-09-13
Pass all form element data to the email template.

EreMaijala:
This makes it possible for the Email/form.phtml template to check field settings etc.

@demiankatz formatEmailMessage was filtering out all but the elements used by the default Email/form.phtml template. This is fine until you need to know more about the field in question, e.g. its name. Id' consider this a fix that I'd like to include in 8.0, but I'll leave the final decision to you.

### DIFF
--- a/module/VuFind/src/VuFind/Form/Form.php
+++ b/module/VuFind/src/VuFind/Form/Form.php
@@ -358,11 +358,10 @@ class Form extends \Laminas\Form\Form implements
         $params = [];
         foreach ($this->getFormElementConfig() as $el) {
             $type = $el['type'];
-            $name = $el['name'];
             if ($type === 'submit') {
                 continue;
             }
-            $value = $requestParams[$name] ?? null;
+            $value = $requestParams[$el['name']] ?? null;
 
             if (in_array($type, ['radio', 'select'])) {
                 $value = $this->translate($value);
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/Form/Form.php
```diff
@@ -377,9 +376,10 @@ public function formatEmailMessage(array $requestParams = [])
                     ?? $this->vufindConfig['Site']['displayDateFormat'] ?? 'Y-m-d';
                 $value = date($format, strtotime($value));
             }
-
-            $label = isset($el['label']) ? $this->translate($el['label']) : null;
-            $params[] = ['type' => $type, 'value' => $value, 'label' => $label];
+            $el['value'] = $value;
+            $el['label'] = isset($el['label']) ? $this->translate($el['label'])
+                : null;
+            $params[] = $el;
```
demiankatz: Would it be cleaner to implement this as below?

<pre>
            $label = isset($el['label']) ? $this->translate($el['label'])
                : null;
            $params[] = $el + compact('value', 'label');
</pre>

I realize it probably doesn't hurt anything, but it seems wrong to manipulate the actual element configuration in order to pass it to the template... feels more appropriate to merge it with additional fields. Though feel free to disagree if I'm missing a reason for doing it as-is.

demiankatz:
Thanks, @EreMaijala -- I'm comfortable with merging this for 8.0.

PR #2124, opened 2021-09-13
Catch exceptions when loading records from the cache.

EreMaijala:
If there's a cached record that cannot be loaded (for any reason), it would have caused an uncaught exception. With this change any exception is logged but it won't stop execution.

While I encountered this with my test database, I'm not sure how the record got broken (might have been a manual tweak), so to test this I think it's necessary to break a serialized record in the database.


demiankatz:
This makes sense to me. Strange that the record got corrupted, but regardless, it never hurts to have proper error-checking in place.

PR #2126, opened 2021-09-14
Aleph: paging support in my transactions

xrosecky:


demiankatz:
@xrosecky, I didn't have a chance to review #2122 until today, and I see it has now been superseded here. You can always force-push to an existing branch rather than opening a new PR if you need to replace content. In any case, is this ready for review now, or should I wait until more work is completed?

xrosecky:
It is now ready for review, I didn't realize that I need also modify the getConfig method in Aleph to disable sorting and make it compliant with the driver specification.

demiankatz:
@xrosecky, I took a closer look at this; is it intentional that this seems to break/remove sorting capabilities?

xrosecky:
@demiankatz Yes, it is intentional. Aleph API unfortunately does not support sorting so to order the items reliably you need to fetch them all.

demiankatz:
Thanks! I'll just make a note in the changelog.

PR #2128, opened 2021-09-14
VuFind\Captcha\ImageFactory: Solve cache path problem

mtrojan-ub:
Since the 'home' URL will usually only contain a `/`, adding another one at the beginning would result in `//` being generated, which would cause problems later.

### DIFF
--- a/module/VuFind/src/VuFind/Captcha/ImageFactory.php
+++ b/module/VuFind/src/VuFind/Captcha/ImageFactory.php
@@ -98,8 +98,8 @@ class ImageFactory implements FactoryInterface
 
         return new $requestedName(
             new \Laminas\Captcha\Image($imageOptions),
-            ($container->get('ViewHelperManager')->get('url'))('home')
-                . 'cache/'
+            rtrim(($container->get('ViewHelperManager')->get('url'))('home'), '/')
+                . '/cache/'
         );
     }
 }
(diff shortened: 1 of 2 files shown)


demiankatz:
@mtrojan-ub, this is causing a unit test to fail. While that may be a sign that the test itself is flawed, I'd suggest that a safer approach would be:

<pre>
rtrim(($container->get('ViewHelperManager')->get('url'))('home'), '/') . '/cache/'
</pre>

This also avoids the double slash problem, but is less likely to cause a missing slash problem.

mtrojan-ub:
I wonder why the test case fails. Does `$container->get('ViewHelperManager')->get('url')('home')` return an empty string in command line mode? (In this case, the test case should be changed.)

edit: Sorry, didn't see you answered already! Thanks for having a look, i'll change it!

demiankatz:
Thanks, @mtrojan-ub -- it's possible that the test case is just using a mock object which returns null for everything, which is not a realistic representation of the helper, but would explain what we were seeing. In any case, with the changes you've made, it now works either way. (It might be worth adding a test case where the route helper returns "/" instead of "" to confirm that everything works... if you'd like to do that but don't know where to start, I can refresh my memory and try to point you in the right direction).

mtrojan-ub:
Now i added another test case and tried to make the biggest part of the first test case re-usable. Unfortunately i am not very experienced regarding PHPUnit so far... it seems like willReturn() doesn't accept a string, so i added a workaround with a getter function. Also, i changed the once() calls to atLeast(1) since the objects seem to be cached. 

Do you know if there is any better solution?

demiankatz:
Thanks, @mtrojan-ub! I revised this significantly; I think what we really needed to do here was make the mock URL helper return a custom value. I set this up by adding some expectations on the mock object, and I used a data provider to pass multiple scenarios to the test case. I think this is ready to merge now. I hope my changes to the test are informative for future reference, but if you have questions about anything, let me know -- happy to help increase your knowledge of the tool if that would be useful. :-)

mtrojan-ub:
Thanks! And now i also understand what a `@dataProvider` is and how it works :+1:

PR #1962, opened 2021-05-19
Icon helper improvements

crhallberg:
## Needed:
- [x]  Wait for #1926.
    - [x] Merge dev
    - [x] Re-search for icons
    - [x] Compile LESS
    - [x] lessToSass
- [x] Replace Bootstrap template icons.
- [x] Set up Fugue icons in Bootprint
- [x] Test functionality and style with non-FA icon set
- [x] Replace style-inserted icons.
- [x] Check local and other themes for icons
- [x] Replace JS icons (loading FA to be compatible)
- [x] Translation style passing of icons to JS
- [x] Create Font Awesome 5 mixin
- [x] Fix double-height myresearch icon bug ([see comments below](https://github.com/vufind-org/vufind/pull/1962#issuecomment-919112474))
- [x] Before merging, double-check for any reintroduced explicit FontAwesome references (e.g. #2190).

## Problems to fix:
- [x] Javascript auto-replaces all request icons (storage/ILL/hold/recall) to the same flag icon when AJAX validity checking is used
- [x] Tag display is misaligned in bootprint3
- [x] Similar items carousel looks wrong in bootprint3
- [x] The 'more' icon as a long arrow looks strange in quite a few places; maybe worth revisiting (it was previously either a short arrow or a greater-than sign, depending on context)
- [x] Fix underlined whitespace between icon and text in links (#2374)
- [x] Look into contrast in bootprint3 on ui-remove button and green background; also lack of hover interaction.
- [x] Status icons for item availability are inconsistently and confusingly named: ‘ui-success’ (available), ‘status-indicator’ (unkown) and ‘ui-failure’ (not available). Need consistent, unique and meaningful identifiers.
- [x] themes/bootstrap3/templates/eds/advanced.phtml has a recently added "fa fa-times" icon for removing a term.
- [x] Should `ui-remove` be `search-delete` in themes/bootstrap3/templates/search/controls/results-tools.phtml?
- [x] Underlined icon in "Add Tag" control in bootstrap3/sandal themes

## Mixins to Be Made later:
- [Bootstrap Icons](https://icons.getbootstrap.com/)
- [Remix Icons](https://remixicon.com/)
- [Heroicons](https://heroicons.com/)
- [Feather Icons](https://feathericons.com/)

## TODO
- [ ] Update changelog when merging (change to Icon helper constructor signature)

PR #2129, opened 2021-09-14
Use icon helper for spinners in templates.

demiankatz:
This PR extracts spinner icon support from #1962 in an effort to introduce spinners as a first phase of icon helper utilization.


demiankatz:
@crhallberg, I've started extracting some of the most basic spinner logic to this PR -- it's helpful to be able to focus on reviewing just one thing at a time. Testing this turned up a major issue: none of the spinners spin anymore in #1962! I think the fix is 
0bf3668, but I wasn't sure if you had a more elegant approach to propose.

Another thing to consider: right now, this PR will load FontAwesome twice, because it's still included in compiled.css in addition to being conditionally loaded by the icon system. Should we disable the configuration to load the CSS here, and only enable it in #1962 when we remove the icons from the compiled bundle?

demiankatz:
@crhallberg, I also found a problem with the hierarchy search (the icon was not initially hidden). It was easily fixed with 35e3c641b7d25addffeda487181723bbad8a7e46.

I've copied both this fix and the previous spinning spinners fix over to #1962 to prevent things from getting out of sync.

demiankatz:
@crhallberg, now I have a mystery I can't explain: I thought that a good solution to the double-FontAwesome-loading problem would be to simply change the src in theme.config.php to compiled.css, since FontAwesome is in that bundle... then we don't double-load anything in this PR, but we can still switch it to load conditionally (and be excluded from the bundle) later, when we're ready.

However, it doesn't work! If I stop loading vendor/fontawesome.min.css, the spinners stop spinning! I don't understand why... is the LESS compiler breaking the spinning? Is it an order of loading issue, and something is breaking/overriding the spin behavior?

demiankatz:
(I had a brainstorm and thought that maybe the non-spinning had something to do with an override in the bootprint3 theme, which I was using for testing... but no, the problem is everywhere if I load only from compiled.css. This is very weird, because the icons themselves load correctly, it's just the animation that doesn't occur... and if I inspect the element, Firefox even seems to think it IS animating).

demiankatz:
On a positive note, it looks like all three of the Summon Deferred templates were missing `fa-spin` from the classes and so never spun to begin with... so if we get this sorted out, we'll be fixing some bugs.

crhallberg:
The only change I'd recommend is changing from setting `fa-spin` in theme.config.php to setting the icons like so:

```php
<?=$this->icon('spinner', 'icon--spin') ?>
```

This would ensure that the icon always spins, even if an image or svg icon is used (where a spin class cannot be added in theme.config.php). Demian points out that would mess up someone who uses an animated icon but I think that case is very unlikely compared to the existence of a static icon in bootprint.

cc @EreMaijala @demiankatz

demiankatz:
@crhallberg, I've put the `icon--spin` class on all the spinners in this PR (and cherry-picked the same changes to #1962 to keep things in sync). I still think I would prefer a more automatic mechanism where the spinning behavior could be added as part of the icon configuration, because otherwise you have to remember two things instead of one thing to create a spinner. So, for example, the icons could just be defined as `set:icon name:extra classes` and then you could apply (or not apply) spin and other features to any icon as part of the configuration. But maybe that makes things too complicated. I can live with it as-is, but would be interested to hear @EreMaijala's thoughts.

In any case, all that aside, I STILL have the problem where spinning only works when I load fontawesome twice, which has me totally baffled.

demiankatz:
Okay, thanks to @crhallberg's fix for the "no spinning without double-loading FA" problem (4c6d3bd), I believe this is now in a potentially mergeable state. I'm still interested in @EreMaijala's opinion on whether it's better to add the `icon--spin` class everywhere explicitly or add a mechanism to embed spinning in the icon configuration. I may also try to get some of the "spinners in Javascript" code merged here as well if time permits before the end of the day.

demiankatz:
@EreMaijala, never mind on the "icon--spin" point -- @crhallberg and I reached a consensus on this, and it seems to be much simpler to encapsulate the extra classes in the icon configuration. Those changes have now been made here (and everywhere). This will make some of the Javascript code in #1962 less complicated as well... once that gets simplified, I'll merge more of it back here if time permits.

crhallberg:
JS simplified: https://github.com/vufind-org/vufind/pull/1962/commits/91b76225b938da58e1b40d98c8a11d9baf0e5cdc

themes/bootstrap3/templates/Recommend/SideFacetsDeferred.phtml
```diff
@@ -75,7 +75,7 @@
       <div id="side-collapse-<?=$this->escapeHtmlAttr($field) ?>" class="collapse<?php if (!in_array($field, $collapsedFacets)): ?> in<?php endif ?>" data-facet="<?=$this->escapeHtmlAttr($field) ?>"<?php if (in_array($field, $forceUncollapsedFacets)): ?> data-force-in="1"<?php endif ?>>
         <span class="facet-load-indicator hidden">
           <span class="text">
-            <i class="fa fa-spinner fa-spin"></i> <?=$this->transEsc('Loading')?>...
+            <?=$this->transEsc('Loading')?>... <?=$this->icon('spinner') ?>
```
EreMaijala: The spinner is moved after the text. Is that intentional? I see Summon templates have it this way, but then cover and relais the other way. Whatever the way, we need to make this consistent.

EreMaijala (commented):
This looks mostly good to me now. :) I'd just like to make the spinner placement consistent across templates. Now it's sometimes on the right, like in side facets:
spinner-right

Personally I find this better-looking and more consistent with the use of icons in other places:
spinner-left

If it was on the right side, I'd perhaps get rid of the three dots, then it would be fine:
spinner-right-no-dots

It's not a major issue, but we should make it consistent across the templates.

themes/bootstrap3/templates/Recommend/SideFacetsDeferred.phtml
crhallberg: I agree that the spinner looks better on the left. This was not an intentional inconsistency, it has been addressed.

EreMaijala (approved):
Now this looks great to me. :)

demiankatz:
Thanks, @EreMaijala -- I've merged this as incremental progress. I'll now merge the results back into #1962 and create another PR for the Javascript-based spinners.

PR #2130, opened 2021-09-14
Fix Hierarchy Highlighting

crhallberg:
Aslo make jstree icons images/svgs, removing another point of icon failure/dependency.

Addresses a few test issues with #1962.


demiankatz:
@EreMaijala, do you have a moment to look at this one? It should take care of an existing bug with tree items highlighting incorrectly. (If you don't use this aspect of the tree functionality, don't worry about it -- I can test tomorrow morning when I have a bit more time).

EreMaijala:
Highlighting is one thing, but moving from icons to images feels like moving in the wrong direction. Why is this necessary?

demiankatz:
@EreMaijala, the issue here is that jsTree only accepts either an image link or a set of CSS classes for icons. When we move to the icon helper, it is going to generate arbitrary HTML, which won't work with jsTree. So our options are either to use hard-coded icon classes (which is inconsistent with the philosophy of the icon helper, since we can't override or replace them easily) or use images (which is also inconsistent with the icon helper, but is a bit more self-contained).

EreMaijala:
@demiankatz Ok, thanks for the explanation.

EreMaijala (approved):
This looks good to me. I suppose the arrows should be changed similarly as well?

EreMaijala:
The arrows apply to hierarchical facets as well. For both the hierarchy tree and facets we have dependency on FontAwesome in js-tree.less:
```
.jstree-ocl:before {
  float: left;
  width: 10px;
  padding: 0;
  margin-left: -10px;
  font-family: 'FontAwesome';
  font-weight: normal;
  font-style: normal;
  text-decoration:inherit;
  cursor: pointer;
  speak: none;
}
```

demiankatz:
@crhallberg, also note that it seems like the highlighting DOES work in the existing code while on the collection page; it's just broken everywhere else (in lightboxes and on record pages). I'm not sure why that is. In any case, this is what it looks like when it works:

image

By comparison, the new code has contrast problems, especially when a link is in the "just clicked" state:

image

But even when nothing is in a clicked state, the contrast on the icon isn't great. Since this is an image, I'm not sure if there's anything we can do about it... but a possible trade-off of switching approaches.

image

EreMaijala:
@demiankatz Right, I didn't notice the contrast issue.

crhallberg:
I fixed the contrast issues, despite `js-tree.less` being the first LESS file I ever wrote, apparently. Should I reformat it? It's kind of a mess.

demiankatz:
Thanks, @crhallberg -- I'll give this another look as soon as I can. Given time constraints, I think we should save refactoring for another day. Hoping we can get a little further on #2129 before the end of the day. Please check Teams -- I've sent a number of messages there, but perhaps you haven't seen them.

demiankatz:
@crhallberg, the new drop-arrows aren't quite right for me (bootprint3 in Firefox):

image

demiankatz:
@crhallberg, follow-up: the arrows look fine in collection hierarchies, so there must be a style problem specific to hierarchical facets.

demiankatz:
Interestingly, the problem applies to the facets in not just the sidebar but also the home page:
image

PR #2131, opened 2021-09-15
Indexing: Avoid fatal error if configs are missing.

demiankatz:
This fixes a problem that can be caused by failing to update author-classification.ini with settings introduced by #2028.


demiankatz:
This fixes a problem identified by @welblaud. I think the fix is pretty straightforward, but if anyone has a moment to double-check, I'd appreciate a review.

EreMaijala (approved):

PR #2132, opened 2021-09-15
Update logos to indicate trademark status.

demiankatz:


demiankatz:
Thanks again, @lmgonzales, for sharing these images. I've noticed one problem, though -- the new version of the dark image is missing the white text. See screen shots below:

Before:

image

After:

image

PR #2134, opened 2021-09-16
Truncate fixes and improvements

aleksip:


aleksip:
@demiankatz So, not sure about commits 2 and 3, but the first one at least should be ok. (Unless `truncate-done` is renamed of course 😄. In that case feel free to just close this without merging anything.)

aleksip:
@demiankatz I have force pushed a new approach suggested by @EreMaijala which covers both discussed issues. We also agreed that the `numRows` and `rowCount` variable names caused confusion, so I have renamed `rowCount` to `maxRows`.

demiankatz:
This looks good to me now, and fixes my known issues.

PR #2137, opened 2021-09-20
[VUFIND-1505] Avoid cache-related out-of-memory exceptions.

demiankatz:
TODO
- [x] Close [VUFIND-1505](https://vufind.org/jira/browse/VUFIND-1505) when merging.


EreMaijala (approved):
Makes sense to me. I'd still like to have some sort of a limit for how much can be cached, but in this case it's quite alright since there's a limited set of icons.

PR #2138, opened 2021-09-20
Make generator config update more fault-tolerant.

demiankatz:
See also [VUFIND-1506](https://vufind.org/jira/browse/VUFIND-1506).

### DIFF
--- a/module/VuFind/src/VuFind/Config/Upgrade.php
+++ b/module/VuFind/src/VuFind/Config/Upgrade.php
@@ -679,9 +679,9 @@ class Upgrade
         // Eliminate obsolete config override settings:
         unset($newConfig['Extra_Config']);
 
-        // Update generator if it is default value:
+        // Update generator if it contains a version number:
         if (isset($newConfig['Site']['generator'])
-            && $newConfig['Site']['generator'] == 'VuFind ' . $this->from
+            && preg_match('/^VuFind (\d+\.?)+$/', $newConfig['Site']['generator'])
         ) {
             $newConfig['Site']['generator'] = 'VuFind ' . $this->to;
         }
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/Config/Upgrade.php
```diff
@@ -679,9 +679,9 @@ protected function upgradeConfig()
         // Eliminate obsolete config override settings:
         unset($newConfig['Extra_Config']);
 
-        // Update generator if it is default value:
+        // Update generator if it contains a version number:
         if (isset($newConfig['Site']['generator'])
-            && $newConfig['Site']['generator'] == 'VuFind ' . $this->from
+            && preg_match('/^VuFind (\d+\.?)+$/', $newConfig['Site']['generator'])
```
EreMaijala, line 684: Would it be too greedy to preg_replace `VuFind (\d+\.?)+` with `'VuFind ' . $this->to` anywhere in the string? This would update something like 'Finna (VuFind 7.1.0)' as well. Just a thought, please disregard if you'd like to keep it as is.

EreMaijala (approved):

module/VuFind/src/VuFind/Config/Upgrade.php
demiankatz: That's not a bad idea for future consideration, but in the interest of making this a bug fix rather than an enhancement (since I want to include it in 8.0.1), I think I'll keep it as-is for now. Would this be useful for you? If so, I can open a separate PR for 8.1 that broadens the scope and adds more tests.

EreMaijala: We can't use the config generator, so not really useful for us.

demiankatz: (Update: following some discussion on Slack, we decided to leave this as-is, because neither of us actually uses the web-based upgrader, so the benefits of adding this functionality would be limited... if somebody else actually needs this, though, it would be an easy request to fulfill).

PR #2139, opened 2021-09-21
Trigger print dialog only once

aleksip:
It the user closes the print dialog and stays on the page, and the keep-alive feature is enabled, the print dialog keeps popping up. With this change the print dialog is shown only once.


demiankatz:
Thanks! This fix should be able to be included in release 8.0.1.

PR #2140, opened 2021-09-21
Make PHPUnit command a configurable Phing property.

demiankatz:
I'm looking into the possibility of using [PCOV](https://github.com/krakjoe/pcov) instead of Xdebug for code coverage. This led me to realize that it might be helpful to be able to customize the PHPUnit command used by Phing, so this PR refactors the command into a property to make it easier to override.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -20,7 +20,7 @@
   <property name="mysqlrootpass" value="password" />
   <property name="pgsqlhost" value="localhost" />
   <property name="pgsqlrootuser" value="postgres" />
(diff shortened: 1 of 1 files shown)


EreMaijala (approved):
Looks good to me.

We should also check if `-dzend.enable_gc=0` is really needed nowadays. I believe it might be a remnant of a problem with older PHP versions, and with the new setting it doesn't seem nice to have it hardcoded.

demiankatz:
Good point, @EreMaijala -- I've factored it out with a second commit (but left a comment, just in case we need it in future).

EreMaijala (approved):

PR #2141, opened 2021-09-23
Make page mandatory in content page route.

EreMaijala:
Without the change requesting /Content/ causes an invalid parameter to be passed to page locator.


demiankatz:
Thanks!

PR #2142, opened 2021-09-23
XCNCIP: Various holds related fixes

xmorave2:


demiankatz:
This all makes sense to me; I'll merge these fixes now so they can be incorporated into release 8.0.1.

PR #2143, opened 2021-09-23
Move print dialog triggering code to separate file

aleksip:
To make it possible to avoid triggering the dialog by not appending the file in `layout.phtml`.


demiankatz:
@aleksip, if we're going to split this out, is there any reason not to move it to theme.config.php instead of explicitly including it in the layout template?

aleksip:
@demiankatz I guess there isn't, if it is possible to remove it in a subtheme's config?

demiankatz:
@aleksip, without reviewing the code, I'm not sure if there's currently a way to remove a file via a sub-theme... but if there isn't at present, it seems like an obvious need for the future. In the meantime, overriding trigger_print.js with an empty file in a subtheme would be one possible way to disable this, which would probably be easier to maintain than a custom layout.phtml (and which I think would have minimal performance impact with the asset pipeline turned on).

aleksip:
Tested with an empty `trigger_print.js` file in the subtheme and that works.

The ablility to remove the parent theme's files in a subtheme's config does seem like a good feature to have.

aleksip:
I've force pushed new commits, which could be merged separately if so wanted.

The second commit adds a minimal timeout to ensure Ajax loaded content is refreshed before the dialog is displayed. We noticed this being a problem at least on Mac Safari browsers.

demiankatz:
Thanks, @aleksip; I think the next question is where this falls in the release cycle -- do we tag it for 9.0 since it's a refactoring? If that's the case, does it make sense to apply the timing fix to common.js first so we can take advantage of the bug fix sooner?

Regarding the ability to disable Javascript entries, I wonder if a 'disabled' setting that could be set to true would be sufficient, or if a special 'priority' value would make sense. (I'd think a separate setting would be more clear). It appears that the code currently explicitly prevents you from redefining an existing file in a later theme (see https://github.com/vufind-org/vufind/blob/9fae3329a3c7b71da6fd30ec3d8296e6ddbad896/module/VuFindTheme/src/VuFindTheme/ResourceContainer.php#L211) but maybe it's worth the effort to think through all the implications and make this more flexible. Do you think it's worth opening a JIRA ticket?

aleksip:
@demiankatz That is a good question! I suppose it depends on what the backwards compatibility promise regarding themes is. Which is something I have not really thought about before because we live on the bleeding edge here. 😄

Looking at the [related documentation](https://vufind.org/wiki/development:architecture:user_interface), the premise seems to be that user interface customization sould be done in a separate custom theme. If this is the case, and I'm not missing something, the change here, now that it includes no template changes, should not be BC breaking even if bootstrap3 is used as a parent theme. Well, unless a custom theme already has its own `trigger_print.js`.

I agree that a separate `disabled` setting would be more clear. Also, opening a JIRA ticket seems like a good thing to do.

demiankatz:
@aleksip, we've always been a bit looser about the backwards compatibility promise in themes than in core code, but I still try to think it through before making a change. :-) In this instance, the one situation where there will be a backward compatibility break is if the user has overridden common.js in their local theme -- which is something that could happen, though perhaps not very often. In that scenario, I imagine the print logic might get applied twice. We'll definitely want to note that in the changelog, but I'm still not sure whether we can get away with 8.1 or should wait for 9.0 (though I lean toward 9.0). If you need this sooner rather than later, we can always establish a dev-9.0 branch so we can start merging 9.0-targeted features.

Regarding the JS disable feature, I've opened [VUFIND-1507](https://vufind.org/jira/browse/VUFIND-1507); I've assigned it to myself and targeted release 8.1, but I'm not sure yet when I'll actually get around to working on it. If you have any desire to get there sooner, just let me know.

aleksip:
@demiankatz Replacing `common.js` with an own copy would mean missing new functionality, which can be added in a minor release. And missing new `common.js` functionality could mean that other related new code might break. So in the case of completely replacing a file like `common.js` I would expect the subtheme's author to monitor changes to the file even in minor and patch releases.

I am unaware of a way to extend and modify the functionality of theme JS files like one can do when extending a PHP class. If there is a way, I could close this PR, as we could then disable the triggering of the print dialog in our subtheme.

In any case a `dev-9.0` branch sounds like a good option too!

demiankatz:
@aleksip, you make a good point regarding the need to monitor local JS files. I think it's reasonable to include this in 8.1, with an appropriate changelog note. I'll merge this soon.

Regarding extending JS files, there is not currently a mechanism to support that; it would be nice, but I'm not sure that there's a feasible technical approach that would allow it. (At least, not without introducing some kind of a build system for JS, which we've been resisting in the interest of simplicity).

themes/bootstrap3/js/common.js
```diff
@@ -489,19 +489,6 @@ $(document).ready(function commonDocReady() {
   var url = window.location.href;
   if (url.indexOf('?print=') !== -1 || url.indexOf('&print=') !== -1) {
```
demiankatz, line 523: Is there a compelling reason to leave some print-related functionality behind in common.js?

aleksip: Without adding a specific setting for it, triggering the print dialog is the only functionality that should be in `trigger_print.js`.

In our use case we still want to show what the printed version of the page looks like, which is what the code left in `common.js` does.

I originally thought of adding a setting, but we already have so many settings, and for many JS functionalities this kind of opt out of/opt in to adding JS files could be a simpler alternative, especially when [VUFIND-1507](https://vufind.org/jira/browse/VUFIND-1507) lands.

demiankatz: That makes sense. I'm going to leave this open until @crhallberg has had a chance to weigh in, but I think I'm comfortable merging this if he is.

crhallberg:
Just to make sure I'm understanding the original issue: moving this printing based on URL functionality to a separate file allows customizers to disable it? Is that correct?

demiankatz:
@crhallberg, yes, that is correct.

demiankatz:
(@aleksip may be able to say more about exactly why that is needed/desirable).

EreMaijala:
This would make it possible to add a print button instead of automatically triggering the print dialog.

aleksip:
For even more background information: full holdings status information in Finna can be really long (several printed pages for just one record), and while our UX team has not decided on the final implementation, we probably want to allow the user to select what (if any) holdings information they want to print for each record.

Loading the holdings information using Ajax can also cause problems with the browser not updating the loaded content when the print dialog opens automatically. This is what ed7a87c attempts to fix.

We also discovered a related bug: [VUFIND-1509](https://vufind.org/jira/browse/VUFIND-1509).

aleksip:
@demiankatz @crhallberg I've force-pushed new commits, adding the timeout in the first commit. So if moving the functionality requires more thought, changes or a different approach, the first commit could still be cherry-picked separately.

demiankatz:
Thanks, @aleksip, I have cherry-picked the commit back to the release-8.0 branch so the fix can be included in 8.0.2. I've also merged it through dev and back to here, so this PR now really does only what the description says it does. @crhallberg is currently on vacation for a couple of days, but maybe when he returns, we can pick up the conversation about the overall approach.

themes/bootstrap3/js/common.js
crhallberg: Just to make sure I'm understanding: you always want the page to preview the print styles but not always trigger the print, so that can be configured.

crhallberg (approved):

themes/bootstrap3/js/common.js
aleksip: @crhallberg Yes, we want to add a print button on the print preview page, and likely also to allow the user to somehow first select which holdings status information they wish to print.

crhallberg: Makes sense to me!

themes/bootstrap3/js/trigger_print.js
```diff
@@ -0,0 +1,23 @@
+/* global VuFind */
+
+$(document).ready(function triggerPrint() {
```
demiankatz, line 3: Is there a reason this changed from `$(document).one('ajaxStop' ...` to `$(document).ready`? If memory serves, the ajaxStop event was hooked to ensure that ajax data was fully loaded before printing. Was that not working correctly?

themes/bootstrap3/js/trigger_print.js
```diff
@@ -0,0 +1,23 @@
...
+  var url = window.location.href;
+  if (url.indexOf('?print=') !== -1 || url.indexOf('&print=') !== -1) {
+    $(document).one('ajaxStop', function doTriggerPrint() {
```
aleksip, line 6: @demiankatz You were probably thinking of this line, which changed?

demiankatz: Ahh, yes, sorry, I had missed that! Since we have @crhallberg's approval, I'll do a little more testing of this when time permits and then get it merged.

themes/bootstrap3/js/trigger_print.js
EreMaijala: @demiankatz ajaxStop fires for any completed AJAX request. It can be item status, versions, OpenURL resolution or session keep-alive. The last one is funny in that it makes the dialog appear periodically again and again..

aleksip: There have been three changes to this code recently, for slightly different reasons! 😅 

- A `one()` was added, to stop keep-alive from re-triggering a closed dialog. 
- A `setTimeout()` was added, to allow some browsers more time to refresh the page with Ajax-fetched data.
- And the change in this PR, which moves most of the code to a separate file.

demiankatz: I see -- so we can't just use one() because if there are multiple updates, then only the first one will be captured in what gets printed... so we use a setTimeout to work around that, which probably means there's still a potential to print a partially-filled page (potentially contributing to, but not being the sole cause of, [VUFIND-1509](https://vufind.org/jira/browse/VUFIND-1509)).

In any case, I'm hoping to find time to test and merge soon -- I think these bigger problems are independent of the work done in this PR and can be handled separately.

aleksip: I believe the `one` refers to the handler being called only once, so it is not directly related to the number of Ajax calls.

Also, the jQuery documentation for `ajaxStop()` states that 

> Whenever an Ajax request completes, jQuery checks whether there are any other outstanding Ajax requests. If none remain, jQuery triggers the ajaxStop event.

So I believe this does not contribute to the problem in VUFIND-1509, which is due to the hunt JS library being used to make Ajax calls only when the records are in view.

demiankatz: Thanks for the clarification, @aleksip, I agree with your assessment -- I had previously made an incorrect assumption about when 'ajaxStop' would trigger. :-)

demiankatz:
I've tested this and all seems well, so I have merged it! I did run into one odd issue, but it's independent of the changes here and likely a browser bug... I mention it just in case anyone else is interested...

In Firefox 93 under Ubuntu, the "afterprint" event doesn't fire correctly, so if I print or cancel print, I don't get redirected back to the previous page like I should. Everything is working correctly in Firefox 78esr under Windows, and in Chrome under Ubuntu.

PR #2144, opened 2021-09-23
Add proxy settings for Phing

literarymachine:
We had trouble installing from behind a proxy. This fixed things.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -32,6 +32,7 @@
   <property name="solr_version" value="8.9.0" />
   <property name="skip_phpdoc" value="false" />
   <property name="phpdoc_version" value="3.0.0" />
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks for sharing, @literarymachine. Out of curiosity, does adding `-Dphing.http.proxy=$http_proxy` on the command line have the same effect? Is using `http_proxy` as an environment variable a widely-used standard or a local convention? I just want to be sure we do this in the most universally helpful and least intrusive way. :-)

demiankatz:
@literarymachine, this PR has been open for quite some time without activity. Did you see my questions above? I'm certainly open to continuing this work if it is needed, but if it is not, we should close this PR.

demiankatz:
I'm closing this one due to inactivity, but I'm happy to discuss further if there is still a need.

PR #2148, opened 2021-09-24
Improve Document title with libraryname

gawol:
The goal is to improve the head title of current website page. The head title of the website are not descriptively named, as they only describe the individual website, but do not allow any conclusions to be drawn about the website. This can pose difficulties for users navigating browser-tabs with a screenreader and should be changed to improve accessibility.
The Solution is to adds the library name to the document titles.


demiankatz:
@gawol, thanks for sharing this. I think perhaps it would be better to implement this through modifications to layout.phtml, rather than replicating the change to every individual template. This would make it easier for users to globally customize it without having to edit many files. For example, perhaps there should be a new setting in config.ini which controls whether or not a library name is appended or prepended to all page titles. What do you think?

gawol:
@demiankatz, Thanks for your advice. I have reverted the commits and set a postfix tot the headtitle in the layout template.

demiankatz:
Thanks, @gawol, this definitely seems significantly simpler! :-) Is there a reason you're using the hard-coded LibraryName string instead of pulling the [Site] title configuration (https://github.com/vufind-org/vufind/blob/dev/config/vufind/config.ini#L31) which seems like it would perhaps be the most logical value to use here? And what do you think about having a config setting like "title_in_header" which could be set to "prepend," "append," or false to disable? If you like this approach but need help with implementation details, I'm happy to assist.

gawol:
I've now implemented it in a more configurable way. If there is more information to be displayed in the header, the information can be displayed as a prefix, postfix, or both.
@demiankatz What do you think?

demiankatz:
Thanks, @gawol, I think this is evolving nicely, but now I think it could actually be simplified to use fewer settings. What if we add just three new config.ini settings: headTitlePrefix, headTitleSuffix and headTitleSeparator? They can default to empty string, but if set, they are applied to the headTitle. I don't think we need a setting to enable or disable the functionality, since behavior can be based on whether the prefix/suffix settings are set. We might be able to drop the separator setting too, but then we would need to embed the separator in the translation strings, which might or might not be desirable. (I suppose a different alternate implementation could have just one config.ini setting: headTitleFormat, which would support pre/post/both, and the layout template could use hard-coded translation strings, so configuration of text would be handled via language files instead of config.ini).

What do you think? And apologies for making this such a drawn-out process, but I want to be sure we reach the most convenient and flexible solution before we merge it. :-)

gawol:
@demiankatz I have extract the functionality in a new View Helper. I think it make sense to reduce the configuration to a minimum with the values separator, prefix and postfix. Maybe we can change some names.
What do you think about it?

config/vufind/config.ini
```diff
@@ -165,6 +165,17 @@ allowSavedSearches = true
 nonJavascriptSupportEnabled = false
 ; Generator value to display in an HTML header <meta> tag:
 generator = "VuFind 8.0.1"
+; headTitle version
+; pre = set the information as prefix before header
+; pos = set the information as postfix after header
```
demiankatz: I would prefer "post" to "pos" here -- I think "pos" could be misunderstood as "position" but "post" is more clear.

config/vufind/config.ini
```diff
@@ -165,6 +165,17 @@ allowSavedSearches = true
...
+; both = set both (prefix and postfix)
+headTitle_ver = pos
+; set separator for title in header
+headTitle_sep = " | "
+; label for prefix value
+headTitle_pre = LibraryName
+; label for postfix value
+headTitle_pos = LibraryName
```
demiankatz: As above, I'd use headTitle_post rather than headTitle_pos.

config/vufind/config.ini
```diff
@@ -165,6 +165,17 @@ allowSavedSearches = true
 nonJavascriptSupportEnabled = false
 ; Generator value to display in an HTML header <meta> tag:
 generator = "VuFind 8.0.1"
+; headTitle version
+; pre = set the information as prefix before header
+; pos = set the information as postfix after header
+; both = set both (prefix and postfix)
+headTitle_ver = pos
```
demiankatz: I wonder if headTitle_style or headTitle_addons might be a better name. I don't have particularly strong feelings about this one, though.

module/VuFindTheme/Module.php
```diff
@@ -150,6 +151,7 @@ public function getViewHelperConfig()
                     View\Helper\HeadScript::class,
                 \Laminas\View\Helper\InlineScript::class =>
                     View\Helper\InlineScript::class,
+                'headtitlewithinfo' => View\Helper\HeadTitlewithInformation::class,
```
demiankatz: I would make the alias match the class name exactly... so either put "information" in the alias or abbreviate the class name to "Info."

module/VuFindTheme/src/VuFindTheme/View/Helper/HeadTitlewithInformation.php
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @return string
+     */
+    public function getHeadTitlewithInformation()
+    {
+        $configHelper = $this->getView()->plugin('config');
+        $headTitleHelper = $this->getView()->plugin('headTitle');
+        $translateHelper = $this->getView()->plugin('translate');
+        $config = $configHelper->get('config');
+        $ver = isset($config->Site->headTitle_ver)
```
demiankatz: You can simplify all this isset code with null coalescing... e.g.:

`$ver = $config->Site->headTitle_ver ?? '';`

module/VuFindTheme/src/VuFindTheme/View/Helper/HeadTitlewithInformation.php
```diff
@@ -0,0 +1,88 @@
...
+{
+
+    /**
+     * Retrieve the Helper
+     *
+     * @return HeadTitlewithInformation
+     */
+    public function __invoke()
+    {
+      return $this;
```
demiankatz: Since this helper only does one thing, it may make sense to put all the logic in __invoke rather than calling a separate public method. That would make the call to the helper more concise.

module/VuFindTheme/src/VuFindTheme/View/Helper/HeadTitlewithInformation.php
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Head Title with additional Information View Helper
+ *
+ * @category VuFind
+ * @package  View_Helpers
+ * @author   Gregor Gawol <gawol@ub.uni-leipzig.de>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+class HeadTitlewithInformation extends \Laminas\View\Helper\AbstractHelper
```
demiankatz: I wonder if AugmentedHeadTitle or ExpandedHeadTitle might be a more concise name (would also potentially clear up the alias inconsistency I noted above).

module/VuFindTheme/src/VuFindTheme/View/Helper/HeadTitlewithInformation.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        $configHelper = $this->getView()->plugin('config');
+        $headTitleHelper = $this->getView()->plugin('headTitle');
+        $translateHelper = $this->getView()->plugin('translate');
+        $config = $configHelper->get('config');
+        $ver = isset($config->Site->headTitle_ver)
+          ? $config->Site->headTitle_ver : '';
+        $sep = isset($config->Site->headTitle_sep)
+          ? $config->Site->headTitle_sep : '';
+        $pre = isset($config->Site->headTitle_pre)
```
demiankatz: I think it would be useful if the pre/post options defaulted to the site title if they're not explicitly overridden in the configuration... for example:

`$pre = $config->Site->headTitle_pre ?? $config->Site->title ?? '';`

This way, there's no need to enter the title multiple times if you want it to be universal, but you can override it as needed.

If you agree, the config.ini comments should be updated to reflect this behavior.

module/VuFindTheme/src/VuFindTheme/View/Helper/HeadTitlewithInformation.php
```diff
@@ -0,0 +1,88 @@
...
+          ? $config->Site->headTitle_pre : '';
+        $pos = isset($config->Site->headTitle_pos)
+          ? $config->Site->headTitle_pos : '';
+
+        switch ($ver) {
+            case "pre":
+                $headTitle = $headTitleHelper->setPrefix($sep . $translateHelper($pre));
```
demiankatz: You could return directly out of the switch in all of these cases, rather than setting $headTitle and breaking. It would make the code more concise.

demiankatz (changes requested):
Thanks for the progress -- this just keeps improving. I've left a few comments, mostly about naming and possible minor enhancements.

I also noticed that this PR was showing some inappropriate diffs for some reason. I merged dev back into your branch and that sorted it out.

gawol:
Thanks @demiankatz  for your advices. I have improve the implementation. The goal is to expanded the document title not the title itself.

demiankatz:
Thanks, @gawol, this is looking good! I've made a few very minor adjustments (updated comments, style adjustments to get continuous integration to pass). Beyond that, I think there are only two things to discuss:

1.) I still think it would be useful to use the existing Site->title setting as a fallback for the pre/post settings. This way, if the user wants to use their site's existing title as a prefix or suffix, they only need to change one setting (expandedHeadTitle_style) to make it appear. I think this is probably the most common situation, but expandedHeadTitle_pre/post allow other scenarios to be achieved when needed.

2.) I think by default, we should either set expandedHeadTitle_style to "none" or false by default, or else we should comment out expandedHeadTitle_pre/post by default in combination with my previous suggestion, so that upgrading VuFind doesn't force users to take action to avoid seeing the default "LibraryName" unexpectedly appearing in their titles.

gawol:
> Thanks, @gawol, this is looking good! I've made a few very minor adjustments (updated comments, style adjustments to get continuous integration to pass). Beyond that, I think there are only two things to discuss:
> 
> 1.) I still think it would be useful to use the existing Site->title setting as a fallback for the pre/post settings. This way, if the user wants to use their site's existing title as a prefix or suffix, they only need to change one setting (expandedHeadTitle_style) to make it appear. I think this is probably the most common situation, but expandedHeadTitle_pre/post allow other scenarios to be achieved when needed.
> 
I think the best solution is to comment out pre/post by default. The most pages will be use the normal head title i think.

> 2.) I think by default, we should either set expandedHeadTitle_style to "none" or false by default, or else we should comment out expandedHeadTitle_pre/post by default in combination with my previous suggestion, so that upgrading VuFind doesn't force users to take action to avoid seeing the default "LibraryName" unexpectedly appearing in their titles.

If we set expandedHeadTitle by default as false or as empty string i think that the best solution.

demiankatz:
Sounds good to me, @gawol. I think we can merge this once the final adjustments have been made (and I can help with them if you prefer).

gawol:
Thanks @demiankatz I have set the default values in the last commit.

demiankatz:
Thanks, @gawol; do you object if I add the title fallback discussed above as well? With that detail in place, I'd be happy to merge this, though I think I'll also ask @EreMaijala for input in case he has any additional suggestions.

gawol:
@demiankatz Feel free to set an fallback value. If @EreMaijala has any additional suggestions we can discuss this.

gawol:
But the idea is to expanded the headTitle not to replace the HeadTitle.

config/vufind/config.ini
```diff
@@ -165,6 +165,17 @@ allowSavedSearches = true
 nonJavascriptSupportEnabled = false
 ; Generator value to display in an HTML header <meta> tag:
 generator = "VuFind 8.0.1"
+; ExpandedHeadTitle style version
+; pre = set the information as prefix before header
+; post = set the information as postfix after header
+; both = set both (prefix and postfix)
+expandedHeadTitle_style = false
```
EreMaijala, line 172: Why camelCase and underscores in the same setting? I'd prefer using one for all settings. Seems to me that there are enough settings that they could all be in their own section such as `[HeadTitle]`, then the actual settings could be simple like `prefix`, `postfix`, `separator`.

demiankatz:
Agreed! I just talked to @EreMaijala and he has some good ideas that will further simplify this; I think he'll be commenting soon.

config/vufind/config.ini
```diff
@@ -165,6 +165,17 @@ allowSavedSearches = true
...
+; set separator for title in header
+;expandedHeadTitle_sep = " | "
```
EreMaijala, line 174: I'd add a note that the prefix gets prepended or appended directly to the postfix or prefix and should include any spaces needed around it.

config/vufind/config.ini
EreMaijala: I think the style setting could be eliminated. Instead the code could check what's defined and use it.

EreMaijala (changes requested):
There's an issue here that if someone is using the prefix or postfix e.g. on a single page, this overrides it at the end. I'd like it better if the config settings provided defaults that could then be overridden if necessary. Since we already have a [custom factory for HeadTitle](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/View/Helper/Root/HeadTitleFactory.php), I think it would be possible to apply all settings there. That would eliminate the need for the ExpandedHeadTitle altogether.

Finally, it seems to me that it might be easier to just add a title_wrapper translation that would get the actual title as parameter `%%title%%`. By default it would be simply:

`title_wrapper = "%%title%%"`

but you could add anything around the %%title%% in translation files.

demiankatz:
@EreMaijala, I do agree that the title_wrapper approach does have the potential to be simpler and more flexible... If we do want to consider that approach, I'd suggest that perhaps the translation should accept both `%%pageTitle%%` and `%%siteTitle%%` parameters, which would enable us to easily make use of the $config->Site->title setting if we want to, with something like `%%siteTitle%% | %%pageTitle%%`. I think the translation-based approach offers maximum flexibility, so it's really just a question of whether it's preferable/more visible to have something in config.ini... but it would certainly be possible to include a note about the title_wrapper translation string in config.ini...

EreMaijala:
@demiankatz Good point, I agree that %%siteTitle%% is needed as well.

gawol:
An additional SiteTitle is a good idea.

demiankatz:
Thanks, @gawol -- would you like to work on further revisions, or would it be helpful if I opened a separate PR implementing the translation-based approach so that we can compare side-by-side?

gawol:
In my opinion the solution is okay. Maybe for a Refactoring we can use a new PR.

demiankatz:
@gawol, please see #2157 for comparison purposes.

gawol:
@demiankatz I close the Pullrequest.

demiankatz:
Thanks, @gawol -- I appreciate the work you put into this, which was helpful in leading us to our chosen solution even though we didn't use it in its original form. :-)

PR #2149, opened 2021-09-24
Add GetItemStatuses::renderFullStatus()

aleksip:
Moves the template render call to a separate overridable method, so that a subclass can process and add more values for the template (or even use a different template).


demiankatz:
Makes sense to me -- thanks, @aleksip!

PR #2150, opened 2021-09-27
[VUFIND-1508] Improve version number processing during upgrade.

demiankatz:
TODO
- [x] Resolve JIRA ticket when merging

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         "cap60552/php-sip2": "1.0.0",
         "colinmollenhour/credis": "1.12.1",
         "composer/package-versions-deprecated": "1.11.99.2",
+        "composer/semver": "3.2.5",
         "endroid/qr-code": "4.2.2",
         "jasig/phpcas": "1.4.0",
         "laminas/laminas-cache": "2.13.0",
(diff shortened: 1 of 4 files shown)


demiankatz:
When testing 8.0.1 I found that there are still problems with the naive version-to-float conversion being done in the upgrade controller. This PR is an attempt at making everything smarter and more consistent. What do you think, @EreMaijala?

EreMaijala (commented):
How about switching to https://packagist.org/packages/composer/semver? I understand that we currently only have it as a dev dependency via php-cs-fixer, but it's a small library and it can handle any semantic version properly.

demiankatz:
Good idea, @EreMaijala; I've just revised this accordingly.

EreMaijala (approved):
While I can't really test this, the code looks good to me.

demiankatz:
Thanks, @EreMaijala! The Config\Upgrade code has pretty good unit test coverage (I confirmed that intentionally implementing the Comparator code incorrectly causes broken tests), and I also manually tested the web-based upgrade to make sure it blocks invalid version numbers appropriately and allows expected legal ones. I think we're ready to merge!

PR #2151, opened 2021-09-27
Add RTL support for jstree icon

crhallberg:
- [x] Could use another eye on the look in the side facets

Addresses JIRA issue [VUFIND-1504](https://vufind.org/jira/browse/VUFIND-1504).


demiankatz:
Thanks, @crhallberg -- I just rebased this against release-8.0 so we can merge it as a bug-fix. I also ran lessToSass to get the build to pass. Testing now!

demiankatz:
@crhallberg, it's looking better, but I think there are some minor problems. Here's hierarchical facets in Arabic:

image

And here they are in English:

image

As you can see, the RTL version now looks better than the LTR version, which seems to have some extra indentation on the right side.

I haven't tested anything else yet, but I'm happy to take a look (and run the full test suite) once you've had a chance to address this part.

demiankatz:
Thanks for the progress, @crhallberg. I notice that some stray commits from the dev branch somehow ended up in here, so I reverted them to keep this compatible with the release-8.0 branch. I'm going to do a bit more testing now...

demiankatz:
@crhallberg, the hierarchical facets look good now, but the collections seem to have gotten broken. Here's the "View Context" dialog box in the dev branch for the standard collection test data:

image

...and here's the same thing in your branch:

image

As you can see, there's some out-of-control indentation going on. Any idea what broke this?

demiankatz:
Oh, and for future reference: I think the reason we ended up with stray dev commits here is because I rebased this against release-8.0, but when you worked on it some more, you pulled and merged with your local branch instead of getting the force-pushed version I had created. In any case, if you pull now, everything should get back in sync.

crhallberg:
> @crhallberg, the hierarchical facets look good now, but the collections seem to have gotten broken. Here's the "View Context" dialog box in the dev branch for the standard collection test data:
> 
> image
> 
> ...and here's the same thing in your branch:
> 
> image
> 
> As you can see, there's some out-of-control indentation going on. Any idea what broke this?

My guess is that some of the styles I rolled back (that weren’t necessary for the facets) we’re keeping the context tree in line. Why they’re behaving differently, I’m not sure. I’ll get content working on my server again and work it out.

demiankatz:
Thanks, @crhallberg!

demiankatz:
@crhallberg, I'm seeing one minor weird problem in Firefox, using the Sandal theme, with RTL turned on -- the left-pointing triangles are squished. I'm not seeing this problem in other themes, and the right-pointing triangles in LTR mode are fine. This affects both hierarchical facets and collection hierarchy contexts. Any idea what's up?

image

demiankatz (approved):
Looking good to me now! Just going to ask @EreMaijala for a double-check...

EreMaijala (changes requested):
Looks good apart from vertical alignment of arrows in hierarchical facets. The arrows seem to be a bit high.
dev branch:
dev
This PR: 
rtl

demiankatz:
Thanks, @EreMaijala, I hadn't noticed that before, but I see what you mean. Any ideas, @crhallberg?

crhallberg:
Better?

image

demiankatz:
Looks fine to me in Firefox and Chrome -- if @EreMaijala has no further concerns, I think this can be merged now. (I confess that I'm not really sure why a line height of 1.75 is necessary to fix this, and I wish the solution felt less magical, but I'm prepared to take your word for it ;-) ).

EreMaijala (commented):
Looks good, but I agree that the line-height feels arbitrary. How about using the same padding for the icon as div.facet? Seems to me that the `padding: 2px 0;` currently applied to .jstree-facet .jstree-node .facet would do it for the icon as well.

crhallberg:
Done! `line-height` is just a method that worked reliably for me in the past, padding might be the more explicit option.

demiankatz:
Sorry to make this the issue that never ends, @crhallberg, but is it just me, or does this look off in RTL mode?

image

Here's the LTR equivalent:

image

Seems like things are getting pushed down a couple extra pixels in RTL.

EreMaijala:
Oh no. It seems that there's something a bit fragile. If you add too much top padding or increase the line height too much, this happens. Easy to reproduce also with LTR with this rule:

```
.jstree-icon {
    line-height: 24px;
}

```

crhallberg:
Looks like float was being applied to all trees instead of just in facets where it was needed.

demiankatz:
I think we've finally got it! Thanks, everyone -- just merged this.

PR #2152, opened 2021-09-28
CursorMarkIdFetcher: move params to property for easier extension.

demiankatz:


demiankatz:
@EreMaijala, I have a use case where I need to pass an extra parameter to Solr in the ParamBag when generating sitemaps (in order to disable field grouping during sitemap generation). As written, the CursorMarkIdFetcher makes it hard to do this customization in a clean way. This refactoring makes it a single-line constructor tweak. Any objections?

EreMaijala (approved):
No objections, makes sense to me!

PR #2153, opened 2021-09-29
Add support for CSP without 'strict-dynamic' script-src.

EreMaijala:
This makes it possible to use the CSP with e.g. Safari, which doesn't support 'strict-dynamic'. All content loaded dynamically is updated with the correct nonce, and rules have been updated to include worker-src for jsTree.

I'm not saying this is beautiful in any way, but I couldn't find a better way to handle the issue that we can't have the same nonce for the main content and any ajax-loaded content.


EreMaijala:
I'm now running the test suite, and I also plan to run it again with the CSP enforced and 'strict-dynamic' disabled.

EreMaijala:
Mink tests are passing, also when the content security policy is enabled without 'strict-dynamic'. I also verified that they would fail without the changes in the first commit they.

build.xml
```diff
@@ -377,6 +377,16 @@
     <exec command="touch ${srcdir}/vufind-exception.log" />
     <exec command="chmod a+w ${srcdir}/vufind-exception.log" />
 
+    <!-- Update contentsecuritypolicy.ini to enforce the CSP and disable strict-dynamic that's not supported on all platforms. -->
+    <copy file="${srcdir}/config/vufind/contentsecuritypolicy.ini" tofile="${srcdir}/local/config/vufind/contentsecuritypolicy.ini">
```
demiankatz: Do we always want to do this, or should there be a property that controls whether or not we run tests with CSP active? It might be useful to be able to turn it on or off if we want to verify whether a test failure is CSP-related.

themes/bootstrap3/js/common.js
```diff
@@ -139,13 +165,17 @@ var VuFind = (function VuFind() {
     addTranslations: addTranslations,
     init: init,
     emit: emit,
+    getCspNonce: getCspNonce,
```
demiankatz, line 168: Is there a compelling reason to have a public getter? Does making the nonce available in this way increase security risks? Do these JS tools make it potentially easier for untrusted code to get executed? (I obviously can't think of any reason why somebody would bother to write a specialized, VuFind-targeted vector to inject malicious code... but regardless of that, I'm just trying to think through all the possibilities).

demiankatz (commented):
This largely makes sense to me, though I want to be sure we don't accidentally undermine the benefits of the CSP by working around a problem. My expertise is limited in this area, so apologies if I speak out of ignorance. Anyway, see some comments below.

The other thing to think about is where this belongs in the release cycle. Is this critical enough to include in 8.0.2, even though it has some significant additions, or should it wait until 8.1? I'm inclined to wait for 8.1 due to the scale of the changes, but if this is a serious enough problem, I could be persuaded to move faster (though in that case, we would need to rebase this branch against release-8.0).

themes/bootstrap3/js/common.js
EreMaijala: I wanted to make it available for any situation where the basic updateCspNonce doesn't do the job. While I'm not currently aware of such case, this should pose no security issue. If you can inject js to call getCspNonce, it's already game over, and you could just as well read it from a script tag in the DOM.

build.xml
EreMaijala: I think we always want to do this, but I'll add an option to turn it off for easier testing.

EreMaijala:
@demiankatz I think targeting 8.1 makes sense especially since this isn't a new issue in 8.0.

demiankatz:
Thanks, @EreMaijala, this looks pretty good to me... though I'm a little confused about exactly what is failing in Safari, since according to [the documentation](https://content-security-policy.com/strict-dynamic/), there's a standard way to work around lack of strict-dynamic support in browsers, and the default contentsecuritypolicy.ini already seems to be doing something similar to the recommendation. Is it possible that a tweak to the .ini file might be enough to sort this out? I suspect that what you've done may still be needed, but I'm not entirely sure it's only needed because of strict-dynamic.

I'm also not sure if it's the best thing to change the default contentsecuritypolicy.ini settings in Phing startup (other than enable, of course, because we want the test to actually tell us something). I think it makes sense for the default settings to be as broadly compatible as possible, so if the workaround isn't working, maybe we need to consider changing the defaults. If we test a non-standard configuration, we may miss problems that average users are more likely to encounter. If we want to test the case where strict-dynamic is disabled, maybe we would be better off building a specific Mink test for that scenario.

EreMaijala:
@demiankatz The workaround would work, but it would effectively disable most of the CSP. Since enabling nonce disabled 'unsafe-inline', nonce would need to be disabled as well, and that's not good.

I kept the defaults with 'strict-dynamic' because it allows some external content integration without changes for supported browsers, and depending on the case that might be good enough. But if we want to ensure that it is a more conscious decision not to support Safari etc. with this content, I'd be happy to change the default and add a comment about it.

demiankatz:
@EreMaijala, I see, that explains it. But if that's the case, I wonder if having the workaround-related settings in contentsecuritypolicy.ini really makes any sense. If these changes make things work without unsafe-inline everywhere, and if unsafe-inline is getting disabled by nonce anyway, maybe we should at least comment out or remove the unsafe-inline default setting.

EreMaijala:
@demiankatz I think it may still make sense for browsers that don't support even nonce. I think Internet Explorer is notable in this regard since it only has partial support for CSP level 1, and nonce support comes at level 2. If we can forget IE, then I think it would make sense to remove 'unsafe-inline'.

demiankatz:
@EreMaijala, I think it's safe to drop IE support at this point. We state that VuFind 8 requires ES6-capable browsers, and IE11 does not support ES6.

EreMaijala:
Ok, I decided to comment out strict-dynamic and add a comment about it.

demiankatz:
Thanks, @EreMaijala -- at this point I'll wait to see what @xmorave2 has to say before proceeding further. I'm satisfied, based on my limited knowledge. :-)

xmorave2 (approved):
Looks good to me. Thanks @EreMaijala and @demiankatz for taking care of it.

demiankatz (approved):

PR #2154, opened 2021-09-29
Refactor MultiBackend driver to avoid repeating similar code.

EreMaijala:
Uses the __call magic method to serve most functions, and makes it possible to introduce new driver methods with no or minor changes.

Some methods remain because the driver interface requires them.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
@@ -131,27 +131,6 @@ class MultiBackend extends AbstractBase implements \Laminas\Log\LoggerAwareInter
     ];
 
     /**
-     * Methods that are supported unconditionally if patron functionality is enabled.
-     *
-     * If a patron can log in or add a library card (i.e. the source is enabled for
-     * login), these methods are used without capability checks.
-     *
-     * @var array
-     */
-    protected $alwaysSupportedMethods = [
-        'cancelHolds',
-        'getCancelHoldDetails',
-        'getHoldLink',
-        'getMyFines',
-        'getMyProfile',
-        'getMyTransactionHistory',
-        'getMyTransactions',
-        'getRenewDetails',
-        'patronLogin',
-        'renewMyItems',
-    ];
-
-    /**
      * Constructor
      *
      * @param \VuFind\Config\PluginManager  $configLoader Configuration loader
@@ -1449,9 +1428,7 @@ class MultiBackend extends AbstractBase implements \Laminas\Log\LoggerAwareInter
                 }
                 unset($param);
             }
-            if (in_array($method, $this->alwaysSupportedMethods)
-                || $this->driverSupportsMethod($driver, $method, $params)
-            ) {
+            if ($this->driverSupportsMethod($driver, $method, $params)) {
                 $result = call_user_func_array([$driver, $method], $params);
                 if ($addPrefixes) {
                     $result = $this->addIdPrefixes($result, $source);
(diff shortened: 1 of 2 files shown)


demiankatz:
Thanks, @EreMaijala -- at first glance, this looks great to me, but it's a huge amount of code to review, so I can't offer 100% confidence based on an eyeball review alone. I'm encouraged by the fact that our test coverage is high and things are still passing, though. A couple of suggestions to consider, though no pressure if you don't think they're good ideas:

1.) It may be worth looking at the [code coverage report](https://vufind.org/jenkins/job/VuFind/clover/VuFind/src/VuFind/ILS/Driver/MultiBackend.php.html) and adding a few more test cases to cover the bits that are missing. If we can get to 100% coverage in this PR, I'd feel even better about it.

2.) A large number of changes here aren't really related to the refactoring -- like repeatedly moving the `$driver = ...` assignment inside the `if ($driver)` check to save a line of code, and some corrections of double quotes to single quotes for efficiency. I wonder if it's worth submitting a separate cleanup PR to get those changes merged first, so that what's left behind is a little easier to review and more focused on the actual logic changes.

EreMaijala:
@demiankatz 
1.) I'll check the coverage.
2.) I tried, but it results in a wildly inconsistent commit since in many cases the method would be removed in refactoring so not worth touching. I can still drop the non-functional changes from this PR and try to do the later, if that's better.

demiankatz:
1.) Thanks!

2.) Thanks for trying -- it's probably not worth the trouble at this point; it sounds like the work it would take to separate things out probably outweighs the benefits of an easier code review on my part.

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
```diff
@@ -109,6 +109,40 @@ class MultiBackend extends AbstractBase implements \Laminas\Log\LoggerAwareInter
...
+        'getCancelILLRequestDetails' => 'cat_username',
+        'getCancelStorageRetrievalRequestDetails' => 'cat_username',
+        'getMyFines' => 'cat_username',
+        'getMyProfile' => 'cat_username',
+        'getMyTransactionHistory' => 'cat_username',
+        'getMyTransactions' => 'cat_username',
+        'renewMyItems' => 'cat_username',
+    ];
+
+    protected $alwaysSupportedMethods = [
```
demiankatz: This property should have a `@var` comment above it. I'm also not sure I understand the values here -- not all of these methods are guaranteed to exist in all drivers (i.e. they're not part of AbstractBase or DriverInterface) so I'm not sure why they're listed as "always supported." Perhaps I misunderstand the context!

EreMaijala:
@demiankatz Added a comment. I hope the comment clarifies the meaning. In any case, these functions are assumed to be present if patron login is available. So, if a patron can log in with proper credentials, we don't do capability checks to see if e.g. getMyHolds is available. For getCancelHoldDetails and getRenewDetails we have indirect checks in that they're only called if canceling holds or renewing loans is supported, but even then the drivers are not checked if the methods are available.

demiankatz:
Thanks, @EreMaijala, that makes sense -- but I think perhaps the property needs a clearer name, since the current name implies something other than its actual meaning. What about `preCheckedPatronMethods`, or something like that? Because it seems to me that it's not so much that they're always supported so much as that VuFind calls appropriate capability checks before calling these methods, ensuring that they're not going to be called inappropriately if unsupported.

EreMaijala:
@demiankatz The name stemmed from the fact that having something in "always supported methods" avoided checking it with `supportsMethod()`. However, I just realized that the whole check is unnecessary. All drivers that implement supportsMethod respond can answer properly to any of the listed methods (they mostly just check if the method is callable). I've now removed the check.

demiankatz:
Even better, @EreMaijala! I can't find further fault here, so I'm going to go ahead and merge. :-)

PR #2155, opened 2021-09-29
Remove reference to custom.css in icon configuration.

EreMaijala:
It seems that FontAwesome is included in bootstrap.less, and custom.css doesn't exist.

### DIFF
--- a/themes/bootstrap3/theme.config.php
+++ b/themes/bootstrap3/theme.config.php
@@ -81,11 +81,11 @@ return [
             'FontAwesome' => [
                 'template' => 'font',
                 'prefix' => 'fa fa-',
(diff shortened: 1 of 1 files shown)


demiankatz:
I think this was a typo on my part and was supposed to be compiled.css! Not sure how that slipped through the cracks. Do you mind changing custom to compiled in the text and comment and see if that also sorts everything out?

demiankatz:
(I think `src` is mandatory, so deleting it may cause other issues).

EreMaijala:
@demiankatz Right, ok, updated. I didn't encounter any issues without src, but then I didn't do any extensive testing either.

PR #2157, opened 2021-10-01
Use translation to simplify page title customization.

demiankatz:
This PR provides a translation-based mechanism for customizing the formatting of page titles on all pages of VuFind, based on @EreMaijala's suggestions on #2148. It is proposed as a possible alternative to that pull request; we should decide which approach is most preferred.


demiankatz:
@gawol, while there are a lot of changed files here, most of the changes are single-line additions to the language files to create the string for formatting titles (and it was possible to create all of these changes using the language/addusingtemplate command-line tool, so it required no significant effort). The only code change to make this work is in layout.phtml. I added a comment to config.ini to document it. I think the advantage to this solution is that I believe it addresses all of the same use cases as #2148, while also offering slightly more flexibility (e.g. the potential to use different separators for different parts of the string). The disadvantage is that it's less structured, and requires language file customization instead of config.ini changes.

What do you think? I'm happy to throw this away if the #2148 solution is preferred by others, but I do find the overall simplicity of this version attractive.

lahmann:
@gawol, @ckaz and me discussed this pull request today and we think that it is most important to decide whether extending the page title by the `siteTitle` should be default or not.
As part of our ongoing accessibility project we had the finc catalogues being tested by an agency specialised on accessibility and they suggested to add the library name after the page title in order to improve browser tab navigation with screen readers (until now VuFind does not mention the configured `title` string in the page title, therefore e.g. it would be difficult to differentiate one VuFind catalogue tab of library A from another VuFind catalogue tab of library B). Following this advise we have started extending the page titles accordingly in our finc catalogues.
In this regard we think that in terms of improving VuFind catalogues's accessibility the `siteTitle` should be added by default - so to say as best practise.
If we agree upon adding the `siteTitle` by default there is no real need to keep the `pageTitle` `siteTitle` composition configurable, therefore this changeset would be favourable and #2148 could be closed. In the unlikely case that someone wants to put the `siteTitle` first (resulting in each browser tab title beginning with the same string) this would be still possible by altering the language files in use.
So ultimately from our perspective it comes down to the question if we want to add the `siteTitle` by default to improve accessibility.

What do you think?

EreMaijala:
@lahmann I'd be happy with having `%%siteTitle%%` included, but that means we have to agree on the default separator as well. In Finna we use `" | "`.

demiankatz:
I have no objection to changing this default; as you say, it's easy to reverse if people care, but it's more likely to help than to hurt. I would also be fine with a " | " separator, though we could also make the separator configurable via config.ini and pass it through as another placeholder in the translation string. I think that's probably excessively complicated, though, since there are other ways to change it. If everyone is happy with " | ", I can revise this PR accordingly.

lahmann:
I think it would be good to check which separator screen readers are handling best. @ckaz do you have any suggestions for screen reader friendly separators?

ckaz:
>@ckaz do you have any suggestions for screen reader friendly separators?
Screen readers won't read : | etc. so if we'd want to have the reader announce the separator we'd go for minus, which many readers read as "dash" -- alternatively, good screen readers will make a pause when encountering periods, semicolons or commas, so we could also go for one of those.

demiankatz:
@ckaz, if the pipe character is not a good choice, I would probably pick dash as my next preference, since I feel that it provides a more clear division than punctuation like period or comma.

EreMaijala:
Screen reader support is a valid point, and I'd be fine with dash as the default.

EreMaijala:
However, now I remember why dash wasn't the first choice for us: https://www.finna.fi/Record/fikka.1505956 . Pipe is a much less likely character to occur in a page title than dash, and having dash as the separator would make the example more confusing. Same applies to slash, colon etc. that often occur in a title.

demiankatz:
Would an em dash be preferable?

EreMaijala:
@demiankatz As to how it looks, yes, I'd prefer em dash over dash, but I've no idea what screen readers do with it.

demiankatz:
I would expect screen readers to handle em dashes, but perhaps @ckaz has more insight. I'd be interested in knowing whether screen readers distinguish between different types of dashes or treat them all the same. (I imagine this may vary).

ckaz:
> I would expect screen readers to handle em dashes, but perhaps @ckaz has more insight. I'd be interested in knowing whether screen readers distinguish between different types of dashes or treat them all the same. (I imagine this may vary).

Hi, I just tested this and screen reader support **varies indeed**. 

- The **Windows10 screen reader** (which comes with any Windows) ignores all dashes, minus, pipe but pauses at comma, semicolon, colon as expected. 
- **NVDA** ignores all dashes and minus but pauses at comma, semicolon, colon as expected. It doesn't read out neither pipe nor semicolon, nor colon.
- **Jaws**, which is promoted as the most popular screen reader (but really only supports IE/Edge browsers), will read the following (default settings):
-- minus: "dash"
-- ndash: "ndash"
-- mdash: "mdash"
-- dash: <pause>
-- comma, <pause>
-- semicolon: "semicolon"
-- colon: "colon"
-- pipe: "vertical bar".

So, Jaws has the best support of separators but whatever we choose, there will be screen readers that announce nothing at all.

You may copy this simple test code into an html file for your own testing:
```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>Screen reader support test</title>
</head>
<body>
<h1>Screen reader support test</h1>
<h2>English examples</h2>
<ol>
  <li>This is a test title - with a minus.</li>
  <li>This is a test title &ndash; with a ndash.</li>
  <li>This is a test title &mdash; with a mdash.</li>
  <li>This is a test title &dash; with a dash.</li>
  <li>This is a test title, with a comma.</li>
  <li>This is a test title; with a semicolon.</li>
  <li>This is a test title: with a colon.</li>
  <li>This is a test title | with a pipe.</li>
</ol>

<h2>Deutsche Beispiele</h2>
<ol>
  <li>Das ist ein Test-Titel - mit einem Minus.</li>
  <li>Das ist ein Test-Titel &ndash; mit ndash.</li>
  <li>Das ist ein Test-Titel &mdash; mit mdash.</li>
  <li>Das ist ein Test-Titel &dash; mit dash.</li>
  <li>Das ist ein Test-Titel, mit Komma.</li>
  <li>Das ist ein Test-Titel; mit Semicolon.</li>
  <li>Das ist ein Test-Titel: mit Doppelpunkt.</li>
  <li>Das ist ein Test-Titel | mit <span lang="en">pipe</span>.</li>
</ol>
</body>
</html>
```

demiankatz:
Thanks, @ckaz. This does make me think of another separator I have seen on some occasions: `::` -- it sounds like this might actually cause some response from all of the screen readers you list. Is that worth considering?

lahmann:
As it looks like finding the best separator is almost impossible how about making this one configurable as well? We could go by default with dash and if this happens to be not very suiteable due to lots of subsidiary titles within the catalogue records one can change it.
So it would be `%%pageTitle%% %%titleSeparator%% %%siteTitle%%` in all language files and we could add a paragraph to the config.ini explaining the difficulty in setting this separator.

ckaz:
> 
> 
> Thanks, @ckaz. This does make me think of another separator I have seen on some occasions: `::` -- it sounds like this might actually cause some response from all of the screen readers you list. Is that worth considering?

Hi Demian, well, yes and no. It would indeed get some kind of response but could also cause confusion since many titles already contain colons. IMO, it looks like a matter of devil and the deep blue sea ... How about using your :: suggestion as a default and making it configurable as André suggests?

demiankatz:
Thanks, @ckaz, that makes sense to me. I'll update this PR in the near future, as soon as time permits.

demiankatz:
Okay, @ckaz (and @lahmann, @crhallberg and @EreMaijala), I have made the suggested revisions. Any further thoughts?

EreMaijala (approved):
Looks good to me!

demiankatz:
Thanks, @EreMaijala! Hopefully I can merge this soon, but I'll wait a little longer in case others have feedback. If the team working on #2148 is happy with this and wants to close that other PR, I can take that as a signal that this one is safe to merge.

config/vufind/config.ini
```diff
@@ -28,7 +28,12 @@ url             = http://library.myuniversity.edu/vufind
 ; make sure your reverse proxy sets the necessary headers.
 ;reverse_proxy    = true
 email           = support@myuniversity.edu
+; The title of your site, used in some messages and (by default) page <title> tags.
+; If you wish to remove this from titles, you can customize the title_wrapper
+; translation in your language files to remove the %%siteTitle%% token.
```
lahmann: It might be more clear why this changed by adding an explanation what the site title is used for in the title tags.

e.g.:
```
; This was added in order to improve accessibility for browser tab navigation with screen readers.
```

lahmann:
Thanks @demiankatz - this looks very good to me. The only suggestion I have is to add a line to the comments in order to explain what the title is used for (cf. my comment in the code). I think this might be useful to clarify why this should be adapted e.g. in customized language files.

Update: in any case #2148 can be closed.

demiankatz:
Thanks, @lahmann, that is a good suggestion. I have adjusted the comment wording to incorporate this. I will merge now!

lahmann:
Great! Thanks!

ckaz:
Great! Many thanks!

PR #2160, opened 2021-10-06
Fix CSP handling in combined results.

EreMaijala:
Also makes VuFind.loadHtml parameters compatible with jQuery.load to support data and success callback.

I had missed this one in #2153.


demiankatz:
Looks good to me -- thanks!

PR #2161, opened 2021-10-06
Fix a PHP notice when a driver does not return callnumber_prefix.

EreMaijala:
This usually remains hidden as it happens during an AJAX call and just produces an additional field in the JSON response.


demiankatz:
Since this is a simple bug fix, I've backported it to the release-8.0 branch and merged it to dev from there. Thanks!

PR #2162, opened 2021-10-07
Add a Csp view helper that allows disabling Content Security Policy.

EreMaijala:
This makes it possible to disable the Content Security Policy for a single page that embeds content that is incompatible with the CSP, typically a third-party service of some kind.


demiankatz:
Thanks, @EreMaijala. This looks good to me, though one naming question: I don't mind keeping disableCsp, but it does seem a little redundant to say $this->csp()->disableCsp(). I wonder if simply `disable`, or, if you prefer greater specificity, `disableHeaders`, might be a little nicer.

The other question is whether to merge the functionality of the CspNonce helper into this one as we recently discussed; maybe it's worth opening a JIRA ticket about that (targeted to release 9.0) after merging this. If you agree, we can put a TODO checkbox on this PR so we don't forget.

EreMaijala:
@demiankatz I've added https://vufind.org/jira/browse/VUFIND-1510. I renamed the method to `disablePolicy`. I'd like to keep it specific so that if we ever add something like `disableReporting`, it won't clash with the existing method. Also, I'd avoid using the word "Headers" since that's more of a detail of how CSP instructions are transmitted, and if we ever need meta tags for that, "Headers" wouldn't quite fit.

demiankatz:
Makes sense to me -- thanks!

PR #2164, opened 2021-10-11
Make MarkdownFactory easier to extend

aleksip:
We have use cases where we need a differently configured Markdown environment. These changes should make the base service class easier to extend with a different environment, configuration or extensions.


demiankatz:
Thanks, @aleksip, this makes sense to me; just giving @xmorave2 an opportunity to comment as well, since he developed the original factory and might have additional input.

xmorave2 (approved):
Thanks @aleksip, this looks well structured and is definitely improvement of current code.

demiankatz:
Thanks for the double-check, @xmorave2. Merging now!

aleksip:
@xmorave2 @demiankatz Thanks!

PR #2165, opened 2021-10-12
Fix password reset form to display the password policy.

EreMaijala:
Targeting release-8.0 branch.

The password form displayed after clicking a recovery link in email was missing the password policy hints (Database authentication).


demiankatz:
Thanks, @EreMaijala!

PR #2169, opened 2021-10-13
Add missing closing tag

aleksip:


demiankatz:
@aleksip, since it is legal to omit the closing PHP tag at the end of the file, it is possible this was actually intentional to avoid extra whitespace being added after the rendered Markdown. However, I do see some potential advantages to being consistent in our template formatting, and I think this may be the only .phtml file that omits the final PHP close tag. @xmorave2, do you have any strong feelings about this one way or the other?

xmorave2 (approved):
I have no objections  against this patch. And consistency is always good.

demiankatz:
Thanks, @xmorave2 -- merging now.

PR #2170, opened 2021-10-14
JS helper for VuFind

LuomaJuha:
- [ ] Tests (What tests should I create?)

Decided now to open a pr of this in here, so the creation of custom JS elements is a little bit easier without jQuery.


demiankatz:
@LuomaJuha, a few questions/comments:

1.) Can you investigate the "useless conditional" feedback above?

2.) Might it make sense to reformat the code as a plug-in for the VuFind namespace, to reduce globals? I think this is just a matter of providing an `init()` function to initialize your logic and then reformatting as a callback passed to `VuFind.register`. If you search for `VuFind.register` in the code you'll find several examples.

3.) Regarding testing, there is not currently a JS unit testing framework incorporated into VuFind, so unless we want to figure out a strategy for that, the best option may be some form of integration testing... Is there existing code that you want to refactor to use this helper? If so, it may make the most sense to start by doing that refactoring and then making sure that the related functionality is covered by integration tests.

4.) Also semi-related to points 2 and 3: if you think this might have general purpose use beyond VuFind itself, we could also consider publishing it as a separate project similar to [autocomplete.js](https://github.com/vufind-org/autocomplete.js) and then including it here in a vendor directory. Obviously, if we made it more stand-alone, that would complicate the idea of incorporating it into the VuFind module, but it would also offer an opportunity for setting up unit testing. That's all a lot of work, of course, and I'm not proposing it's worth the effort -- just mentioning the possibility. This also raises the question of whether there's an existing third-party library that does the same thing, in which case perhaps we could use that instead (I didn't take the time to do any research, but it seems like a theoretical possibility).

I hope this is helpful!

LuomaJuha:
@demiankatz 1. Sure, bots are fancy! 2. Yeah! 3. Oh yeah, well, have to check out about that is there some smooth way. 4. Sure could sound plausible. I will check if someone has had the same idea and has published it. :)

demiankatz:
Thanks, @LuomaJuha, let me know if you need any assistance with anything. :-)

LuomaJuha:
@demiankatz I was looking for a github repository which would do the same, but mostly the documentation was not in English or the functions were meant for something more unique. I think it would be nice to have our own version where we could easily make functions to work in our project.

demiankatz:
Thanks for taking the time to do the research, @LuomaJuha. I think in terms of next steps, I would be interested to hear what @crhallberg thinks of this... and I think it would be helpful to find some places to apply it in the existing code so that we can integrate and test it, rather than having it as a free-floating, unused helper.

crhallberg:
4. Instead of `vendor/`, making a `lib/` folder for standalone components (like the lightbox) might be a better way to organize the code.

demiankatz:
@crhallberg, I agree. Vendor is for things that come from a third party... if it only exists in VuFind, it doesn't belong there.

LuomaJuha:
Well, sounds plausible to me so I will get back to this after I get my workqueue a littlebit shorter!

demiankatz:
Thanks, @LuomaJuha -- happy to take another look whenever you've had time to do some more work and feel ready for review; please let me know whenever that is.

LuomaJuha:
Closing this one for now.

demiankatz:
Thanks, @LuomaJuha -- did you determine that this was unnecessary, or do you plan to reopen in the future? (Just curious what the long-term plans are, and whether you need any assistance).

LuomaJuha:
@demiankatz Yeah, it seems like javascript these days can handle things pretty well like closest / parentNode etc. and adding unnecessary wrappers can make things too quirky. Maybe if something appears in future which would be nice to add as a helper function, then it would be better to add it in the VuFind module.

PR #2171, opened 2021-10-15
Aleph: Fix for pagination of historic loans

welblaud:
According to the [docs](https://developers.exlibrisgroup.com/aleph/apis/aleph-restful-apis/loans/), the parameter for getting just a chunk of historic loans should be **no_loans** instead of **noItems**, which is used for regular loans. In my case, it really works this way. It is a rather annoying inconsistency.


demiankatz:
Thanks, @welblaud! I'm requesting a review from @xmorave2 (or perhaps @xrosecky would like to comment) to make sure he is seeing the same thing. With confirmation from another Aleph user (which seems likely!) I'll merge this.

xrosecky:
Yes, it is correct, you can merge it.

xmorave2 (approved):

demiankatz:
Merged -- thanks again, everyone!

PR #2172, opened 2021-10-15
[VUFIND-1495] Improve addition of non-tab record routes in RouteGenerator.

mtrojan-ub:
https://vufind.org/jira/browse/VUFIND-1495

Just a first idea how to solve this (not tested yet). What do you think?


module/VuFind/src/VuFind/Route/RouteGenerator.php
```diff
@@ -113,6 +104,39 @@ public function addDynamicRoutes(& $config, $routes)
...
+    /**
+     * Add non tab record action & re-register all record routes to support it.
+     *
+     * @param array  $config     Configuration array to update
+     * @param string $action     Action to add
+     */
+    public function addNonTabRecordAction(& $config, $action)
+    {
+        // Explicitly use RouteGenerator:: here, so even children will
+        // use the parent's variable.
```
demiankatz: I think you could use `self` instead of `RouteGenerator` and it would have the same effect...

demiankatz (commented):
Thanks, @mtrojan-ub, this makes sense to me! Might make sense to add some unit test support to confirm desired behavior. I suspect that you could create an anonymous subclass within the test class to test extensibility.

mtrojan-ub:
Thanks for having a first look! Not sure which existing test class you mean, though. Do you mean `VuFindTest\Record\RouterTest`?

demiankatz:
@mtrojan-ub, I wasn't referring to an existing test class -- I think we need to create a new one for this class. I just meant that when we do create a test class, we could use an anonymous subclass of the RouteGenerator in one of the tests to confirm that extensibility works as expected. If it would be helpful for me to create a baseline RouteGeneratorTest so that you have something to build upon, I might be able to find time for that sometime this week.

mtrojan-ub:
That would be very helpful indeed!

demiankatz:
@mtrojan-ub, as requested, I've created a baseline route generator test in #2185. I've requested your review, since I expect you'll be taking a look at it anyway. If you're satisfied with what I've done, feel free to merge the PR yourself so that you can build upon the test case here.

mtrojan-ub:
Thanks, @demiankatz! I already merged it and will implement the additional test case when time permits.

mtrojan-ub:
@demiankatz, seems like it works like this. Could you please have another look?

demiankatz:
Thanks, @mtrojan-ub, I think this looks good! Just two comments:

1.) I've set this with a 9.0 milestone -- the way the default generation of non-tab routes has changed could cause some backward compatibility breakage, so I think it's better to make in a major version. If you need access to this in merged form in the near future, I'm happy to establish a dev-9.0 branch so we can begin merging 9.0 changes.

2.) In the test suite, it might be better to add a new test rather than modifying the existing one. I think there's some advantage to having a basic test that just tests the core functionality, and then a testAddRecordRoutesWithSubclass test that does the more advanced subclassing and adding behavior. I think this is easy to do with a little bit of copy and paste, and while it's mostly redundant, I think it might help us pinpoint a problem if there is a failure specifically related to the new logic.

mtrojan-ub:
Good idea, @demiankatz! I restored the test case and inserted a new one. Also, there is absolutely no urgency, so i'm fine waiting for the 9.0 for this to be merged.

demiankatz:
Thanks, @mtrojan-ub -- I've established a dev-9.0 branch so that this can be merged ahead of the 8.1 release. For the next few months, I'll merge 8.1-related work into dev, and then merge that forward into dev-9.0, so we have a place to merge forward-looking 9.0 work and prevent pull requests from piling up here.

PR #2173, opened 2021-10-18
Fix loginIsHidden call in MultiBackend ILS driver

xmorave2:
loginIsHidden has no parameters and in MultiBackend driver is called without source parameter (in __call method) which ends in 'No suitable backend driver found' ILS exception

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
@@ -1417,6 +1417,9 @@ class MultiBackend extends AbstractBase implements \Laminas\Log\LoggerAwareInter
         bool $stripPrefixes = true,
         bool $addPrefixes = true
     ) {
+        if (empty($params) && null === $source) {
+            return null;
+        }
         if (null === $source) {
             $source = $this->getSourceForMethod($method, $params);
         }


module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
```diff
@@ -1417,6 +1417,9 @@ protected function callMethodIfSupported(
         bool $stripPrefixes = true,
         bool $addPrefixes = true
     ) {
+        if (empty($params) && null === $source) {
```
EreMaijala, line 1420: I think with MultiBackend we could have a slightly different logic: return true if there are no configured login targets ( = getLoginDrivers returns an empty array). MultiBackend could also check all configured login targets for loginIsHidden support and verify that at least one of the configured login targets allows login, but I wouldn't go that far without a use case.

In a related note, I'm not quite sure if loginIsHidden works with ChoiceAuth at all, but that's a different issue.

EreMaijala (changes requested):

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
xmorave2: I don't think it's OK just return true if there are login drivers - we are using MultiBackend with Shibboleth login - so we still need to show login option to users...

Maybe in case of getLoginDrivers not empty AND Authentication->method === 'MultiILS'

demiankatz: The purpose of loginIsHidden() is to disable the login option when the ILS is offline, so it is only used when a particular NoILS.ini setting is turned on and the NoILS driver is active. I don't think it matters what authentication method is used (ChoiceAuth, etc.), since this function simply controls whether or not the login control is displayed at all -- it will disable login before any authentication logic can be triggered.

I think the open question is whether we disable login if ANY driver reports that we should, or we only disable login when ALL drivers report that we should. Seems like there could be some tricky edge cases here. However, I'd probably err on the side of permissiveness and only hide login when we're very sure we shouldn't be displaying it... better to let the user try and potentially see an error message than to inadvertently prevent them from doing something they should be able to do.

EreMaijala: @xmorave2 @demiankatz It's not quite clear to me when the ILS should be considered unavailable in this case.

I don't think the mechanism works properly in @xmorave2's case, or in any other case where method in [Authentication] is something else than ILS. With ChoiceAuth it seems that LoginIsHidden should only have the following effect:
1.) If choice_order contains ILS, remove the choice if loginIsHidden() returns true.
2.) If choice_order contains MultiILS, remove any driver from the selection list if its loginIsHidden() returns true, and remove the whole choice if no choices remains.

EreMaijala: I think a more proper quick fix would be to make MultiBackend::supportsMethod return false for loginIsHidden, but that doesn't fix the underlying issues.

demiankatz: I think an important clarification may be in order: the loginIsHidden setting is not meant to be strictly about the actual act of authentication. It really doesn't matter what authentication method is selected from the perspective of the original intention of this setting. It is meant to be used in two scenarios:

1.) When there is no need for login at all, and a VuFind admin wants to totally disable the login mechanism.

2.) When LoadNoILSOnFailure is active, and we want to lock users out of their accounts when the ILS is offline.

Regardless of what might happen after the user clicks "login," in these scenarios, we don't want the user to even have the option of trying.

EreMaijala: @demiankatz I understand that if ILS is used as the login method, but I can't see the point with ChoiceAuth, Shibboleth, Database or any other one. Why shouldn't the user be able to e.g. log in with Shibboleth to access electronic material when the ILS is down?

demiankatz: @EreMaijala, this is why it is an optional setting with a default value of false. It is designed for the use case where the only reason to log in is to access ILS functionality, where you want to disable login if the ILS is unavailable. For any other scenario, you should not use this setting.

EreMaijala: @demiankatz But you can't actually disable the logic that calls loginIsHidden in Auth Manager's factory, can you?

demiankatz: @EreMaijala, that's true. The mechanism is awfully heavy-weight for something that is almost never used... but I'm just describing the only known use case for its existence.

EreMaijala: @demiankatz So the problem is that the new MultiBackend driver tries to handle it but fails since it can't determine the backend at least unless you define a default_driver in MultiBackend.ini.

EreMaijala: I'll post a separate PR to fix that case.

demiankatz: Perhaps the best solution for this particular functionality is to default to false if no backend can be determined; I can't imagine there's a particularly good reason to dynamically hide login in a multi-backend scenario.

EreMaijala: Indeed. And the same applies to a couple of other methods as well that should be disabled when default_driver is not set.

EreMaijala: I've now opened #2174 that should cover all the methods that don't have source-specific parameters.

xmorave2:
I am closing this in favor of #2174

PR #2174, opened 2021-10-18
MultiBackend: Handle methods with no source specific params properly.

EreMaijala:
Continued from #2173.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
@@ -131,6 +131,23 @@ class MultiBackend extends AbstractBase implements \Laminas\Log\LoggerAwareInter
     ];
 
     /**
+     * Methods that don't have parameters that allow the correct source to be
+     * determined. These methods are only supported for the default driver.
+     */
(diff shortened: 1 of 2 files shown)


demiankatz:
Thanks, @EreMaijala. Two questions:

1.) Is it worth adding anything to the test suite related to this functionality?

2.) Does this supersede #2173 or simply complement it? Wondering what should happen to that PR after this one is merged.

xmorave2 (approved):

xmorave2:
> 2.) Does this supersede #2173 or simply complement it? Wondering what should happen to that PR after this one is merged.

For our use case it's enough to push one of these PRs, but I am not sure if it will work for everybody...

I do like @EreMaijala solution - it is clean from the code what is going on

EreMaijala:
@demiankatz 

1.) Good point, I've now added a simple test to verify that supportsMethod works properly with loginIsHidden.

2.) I'd prefer this to supercede #2173. The problem with 2173 is that it has the potential to mask problems. The exception is there for a reason, just like in this case where it uncovered an issue when default driver is not set.

demiankatz:
Thanks, @EreMaijala and @xmorave2 -- merging now!

PR #2175, opened 2021-10-18
Do not move MARC files to processed if ingest fails.

demiankatz:
TODO
- [x] Equivalent Windows support


demiankatz:
@EreMaijala, I discovered that the batch MARC ingest routine moves harvested MARC files to the "processed" directory whether or not the ingest succeeds. This PR addresses that bug. I'm stuck on the Windows piece, though -- I spent about half an hour trying to figure it out, but the language is so clumsy I'm not sure there's a sane way to do it. I might settle on opening a JIRA ticket, because I somewhat doubt that anyone actually cares, and I don't want to waste too much time just on the principle of cross-platform compatibility. :-)

EreMaijala:
@demiankatz I admit that I'm out of my depth with Windows these days, but can't resist trying. Could you do something like this in batch-import-marc.bat:

```
rem Process all the files in the target directory:
for %%a in (%BASEPATH%\*.xml %BASEPATH%\*.mrc) do (
  if "%LOGGING%"=="0" (
    call %VUFIND_HOME%\import-marc.bat %%a
  )
  rem Capture solrmarc output to log
  if "%LOGGING%"=="1" (
    call %VUFIND_HOME%\import-marc.bat %%a 2> %BASEPATH%\log\%%~nxa.log
  )
  if ERRORLEVEL 1 goto skipmove
  if "%MOVE_DATA%"=="1" (
    move %%a %BASEPATH%\processed\ > nul
  )
  :skipmove
)
```

Then in import-marc.bat, instead of `goto end` do `exit /b 1` if there's an error. And perhaps `exit /b %ERRORLEVEL%` in the end (not sure about needing that)?

demiankatz:
@EreMaijala, your proposed solution looks very similar to one that I attempted. I think the issue I ran into is that you can't use `goto` inside a `for` loop (or at least, if you can, I didn't find the right syntax to make it work). I also tried using a flag variable to skip out of the loop, but I didn't manage to get that to work either (though by that point, I confess that I was losing patience and running out of time). I agree that making the import-marc.bat return an appropriate errorlevel is a good first step, so maybe I'll do that in a separate PR and then circle back to this one for another attempt.

EreMaijala:
@demiankatz Maybe a nested if with `IF %ERRORLEVEL% EQU 0` (taken from https://ss64.com/nt/errorlevel.html)?

demiankatz:
@EreMaijala, I think I've finally gotten it! I refactored the code to use a subroutine since that made it a bit easier to follow (though it's still batch file spaghetti). I think my problem before is that I had forgotten that the `if errorlevel` construct is actually a `>=` check rather than an `==` check, so `if errorlevel 0` does not do what you might reasonably expect it to do! In any case, this is all tested as working on my Windows 10 box, but if you see anything concerning, please let me know. Otherwise, I think it's ready to merge.

EreMaijala (approved):
Looks good to me. I'd add a newline at the end of both bat files, but those were missing before as well.

demiankatz:
Thanks, @EreMaijala, I've added the missing line breaks, and I'll merge this once the build passes.

Strangely, one of the builds failed yesterday because the `npm install` command for eslint ran for six hours without completing. Hopefully that was an isolated incident, but something to watch out for in the future! If it happens again, maybe we should investigate whether we can cache globally installed npm modules.

PR #2177, opened 2021-10-19
Add a QA script for composer

LuomaJuha:
Easier to remember to use all the tools provided. Merge-plugin allows for custom edits.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -111,6 +111,7 @@
         "post-install-cmd": "@phing-install-dependencies",
         "post-update-cmd": "@phing-install-dependencies",
         "qa" : [
+            "phing eslint",
             "phing phpunitfast",
             "phing phpcs-console",
             "phing php-cs-fixer-dryrun"
--- a/composer.json
+++ b/composer.json
@@ -110,15 +110,6 @@
         "phing-install-dependencies": ["phing patch-dependencies", "phing installsolr installswaggerui"],
         "post-install-cmd": "@phing-install-dependencies",
         "post-update-cmd": "@phing-install-dependencies",
-<<<<<<< Updated upstream
-        "qa" : [
-            "phing eslint",
-            "phing phpunitfast",
-            "phing phpcs-console",
-            "phing php-cs-fixer-dryrun"
-        ]
-=======
         "qa": "phing eslint phpunitfast phpcs-console php-cs-fixer-dryrun"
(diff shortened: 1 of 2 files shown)


demiankatz:
Thanks, @LuomaJuha! Two thoughts:

1.) Is it worth considering creating a QA task in build.xml to group these things together at a lower level?

2.) If not, I think you can do this as a one-liner: phing will accept a series of tasks as a single command line, so I don't think it's necessary to run Phing once for each task.

LuomaJuha:
never really looked at build.xml but can take a look @demiankatz. Will try to merge them into a one-liner if it seems like too much!

demiankatz:
@LuomaJuha, if you look at the ci-tasks task, you'll see how you can easily group together a bunch of tasks in build.xml:

https://github.com/vufind-org/vufind/blob/dev/build.xml#L68-L86

I think this might be worth considering, because then we can offer the same functionality to people who prefer Phing and people who prefer Composer (with Composer just proxying the Phing task). But if it turns out to be a lot of work, it might not be worth the effort -- just an option to consider, which I don't think will turn out to be too hard.

demiankatz:
Also, if we go the phing route, maybe we should call the new task "qa-console" in Phing to differentiate from the report-generating tasks in ci-tasks... but from Composer it could just be qa because the "console" part is assumed. Again, just brainstorming -- feel free to use or ignore suggestions as you see fit!

LuomaJuha:
@demiankatz took your advice and added a quality assurance task to build.xml and changed the composer command to a one-liner.

composer.json
```diff
@@ -109,6 +109,7 @@
     "scripts": {
         "phing-install-dependencies": ["phing patch-dependencies", "phing installsolr installswaggerui"],
         "post-install-cmd": "@phing-install-dependencies",
-        "post-update-cmd": "@phing-install-dependencies"
+        "post-update-cmd": "@phing-install-dependencies",
+        "qa": "phing eslint phpunitfast phpcs-console php-cs-fixer-dryrun"
```
demiankatz: For consistency, I would make this `phing qa-console` so that if we add more tasks, we only need to put them in one place.

LuomaJuha: Yeah, forgot that one and just changed it!

demiankatz:
Thanks again, @LuomaJuha!

PR #2178, opened 2021-10-19
Fix Csp helper's disablePolicy when CSP mode is "report_only".

EreMaijala:
Apparently the header may be castable as an array which breaks the foreach loop.

This seems to happen only when CSP is set to "report_only" mode, though I'm not quite sure why. In any case, this adds a couple more tests to make sure disablePolicy works in all modes.

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/Csp.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/Csp.php
@@ -72,12 +72,12 @@ class Csp extends \Laminas\View\Helper\AbstractHelper
             as $field
         ) {
             if ($cspHeaders = $headers->get($field)) {
-                if ($cspHeaders instanceof \ArrayIterator) {
-                    foreach ($cspHeaders as $header) {
-                        $headers->removeHeader($header);
-                    }
-                } else {
-                    $headers->removeHeader($cspHeaders);
+                // Make sure the result is iterable (an array cast doesn't work here
+                // as a single header may be castable as an array):
+                $headerArray = $cspHeaders instanceof \ArrayIterator
+                    ? $cspHeaders : [$cspHeaders];
+                foreach ($headerArray as $header) {
+                    $headers->removeHeader($header);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/View/Helper/Root/Csp.php
```diff
@@ -72,8 +72,12 @@ public function disablePolicy()
             as $field
         ) {
             if ($cspHeaders = $headers->get($field)) {
-                foreach ((array)$cspHeaders as $header) {
-                    $headers->removeHeader($header);
+                if ($cspHeaders instanceof \ArrayIterator) {
```
demiankatz: Would `is_iterable()` be a more generic way of doing this?

demiankatz:
Thanks, @EreMaijala. I'm also asking for input from @xmorave2 in case he can add any additional context/insight.

xmorave2 (approved):

module/VuFind/src/VuFind/View/Helper/Root/Csp.php
EreMaijala: Perhaps, but I'm now spooked about the possibility of a single header to be somehow iterable, so I wanted to be explicit about it.

EreMaijala: If you think the risk is minimal, I'm happy to revise.

demiankatz: I see from the laminas-http documentation that `get()` is always supposed to return an `ArrayIterator` for multiple headers, so I think your check does make sense. I wonder if it would be more concise to do something like this, though:

<pre>
$headerArray = $cspHeaders instanceof \ArrayIterator
    ? $cspHeaders : [$cspHeaders];
foreach ($headerArray as $header) {
    $headers->removeHeader($header);
}
</pre>

This reduces the amount of indentation and eliminates some minor duplication of logic.

EreMaijala: Good idea, done. I also added a comment to clarify.

demiankatz: Thanks! Looks good to merge now.

PR #2182, opened 2021-10-21
Add option to set accepted HTTP codes (other than 2xx) for NCIP driver

xmorave2:
Some NCIP responders could send other than 2xx codes when there is an issue, but response with 'Problem' element is still valid NCIP response and we may need to parse it.


demiankatz:
Looks good to me -- thanks!

PR #2184, opened 2021-10-21
Unify qa tasks

LuomaJuha:
Hmm, is this the proper way or is it getting too complicated? any thoughts?

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -88,12 +88,12 @@
   <!-- Quality Assurance Tasks -->
   <target name="qa-console" description="quality assurance tasks">
     <!-- Call standard tasks -->
-    <phingcall target="qa-javascript"/>
+    <phingcall target="qa-js-and-less"/>
     <phingcall target="qa-php"/>
   </target>
 
   <!-- Quality Assurance Javascript Tasks -->
-  <target name="qa-javascript" description="quality assurance php tasks">
+  <target name="qa-js-and-less" description="quality assurance js and less tasks">
     <phingcall target="eslint"/>
     <phingcall target="jshint"/>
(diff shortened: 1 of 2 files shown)


build.xml
```diff
@@ -88,10 +88,23 @@
   <!-- Quality Assurance Tasks -->
   <target name="qa-console" description="quality assurance tasks">
     <!-- Call standard tasks -->
+    <phingcall target="qa-javascript"/>
+    <phingcall target="qa-php"/>
+  </target>
+
+  <!-- Quality Assurance Javascript Tasks -->
+  <target name="qa-javascript" description="quality assurance php tasks">
```
demiankatz: The description here says "php", which I assume is a copy-and-paste error. I also wonder if we need a different name for this task, since "checkLessToSass" is not a Javascript check, though I do agree that these checks all belong together. Maybe "check-theme-assets" or "check-js-and-less"?

LuomaJuha: @demiankatz  qa-front-end and qa-back-end? And yeah, thats a copypaste error from me...

demiankatz: I did think of qa-front-end and qa-back-end as a possibility, but some of the current qa-php tasks (like php-cs-fixer) touch both front and back-end, so it wouldn't be strictly true.

demiankatz: I suppose `qa-non-php` would be another option, though I'm not thrilled with it. That approach does help to clarify what's going on in the GitHub actions, though...

LuomaJuha: Yeah, maybe check js-and-less would be the most describing, will change

demiankatz:
Looks great -- thanks again, @LuomaJuha!

PR #2186, opened 2021-10-25
Fix CSS property name

aleksip:
I believe this is the correct name for the property?

### DIFF
--- a/themes/root/css/icon-helper.css
+++ b/themes/root/css/icon-helper.css
@@ -73,10 +73,10 @@
 @keyframes bounce {
   0%, 100% {
     transform: translateY(-25%);
-    animationTimingFunction: cubic-bezier(0.8, 0, 1, 1);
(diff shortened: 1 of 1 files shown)


crhallberg (approved):

demiankatz:
@crhallberg, passing this along for your consideration. I believe that @aleksip is correct but wanted to see if you had any concerns about the implications of the fix. If this is indeed a needed correction, we should cherry-pick this back to the release-8.0 branch rather than merging it directly to dev.

crhallberg:
Yes this is correct. This is the JavaScript name slipping into the CSS. Thank you, @aleksip !

demiankatz:
Thanks, both of you! I cherry-picked this as 7129ca2d781c2b933dc8a168986bee8b623c9762 and merged it into dev, so I can now close the PR.

PR #2188, opened 2021-10-26
Show last issue date for patron

xmorave2:
This PR is intended to add ability of showing the information about the fact that logged in patron have had some document checked out in the history. I don't think this is final to be merged, but could be start of a discussion if we wan't such a feature and how it should work.


demiankatz:
Thanks, @xmorave2 -- just to be sure I understand, the idea here is to let people know which books they have previously borrowed as part of the status display, correct? If that's the case, a few thoughts:

1.) This should definitely be an optional feature if we go forward with it, since some systems won't support it (no retained history) and some institutions won't want it (privacy concerns).

2.) I imagine we would want to display this consistently between the results list and the holdings tab of the record view. We might also want to reflect it in some way when "full status" is not enabled for results (maybe an icon or status indicator in addition to availability status?).

3.) We should think about the most appropriate language -- I don't think most users will understand what "last issue" means. Terminology varies a lot from library to library and system to system -- "charged," "checked out," etc. I wonder if "last borrowed" would be the most universally understood language.

Definitely interested to hear what others think of this, and happy to mention it on a future Community Call if that would be helpful.

LuomaJuha:
I'd suggest something like previous_checked_out?

demiankatz:
@xmorave2, I'm just reviewing old pull requests and see that this hasn't moved in more than a year. Is this something you are still interested in pursuing?

demiankatz (changes requested):
@xmorave2, while reviewing old PRs in need of reviews, I noticed that this one hasn't moved in quite a long time. Is this worth bringing up to date so we can finish it?

PR #2189, opened 2021-10-26
Return appropriate errorlevels from import .bat files.

demiankatz:
This is an incremental step toward implementing Windows support in #2175 -- it makes sure that the import .bat files return appropriate error levels after running SolrMarc.


EreMaijala (approved):
Looks good to me, though I can't verify in action.

demiankatz:
Thanks, @EreMaijala -- I did verify in action, so I'm comfortable merging!

PR #2190, opened 2021-11-03
Add option for icon placement in truncate. Use in facets.

EreMaijala:
We need to be able to change the icon placement for truncate buttons. I also think the icons would look better before the label in facets where there's also the ' ...' after the label.

### DIFF
--- a/themes/bootstrap3/js/truncate.js
+++ b/themes/bootstrap3/js/truncate.js
@@ -5,7 +5,9 @@ VuFind.register('truncate', function Truncate() {
     'btn-class': '',
     'in-place-toggle': false,
     'label': null,
+    'less-icon': '<i class="fa fa-arrow-up" aria-hidden="true"></i>',
     'less-label': VuFind.translate('show_less'),
+    'more-icon': '<i class="fa fa-arrow-down" aria-hidden="true"></i>',
     'more-label': VuFind.translate('show_more'),
     'rows': 3,
     'top-toggle': Infinity,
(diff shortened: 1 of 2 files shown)


aleksip (approved):
Looks good to me!

Maybe we could also wrap the label in a `<span>` just to make styling easier, since overriding JS is something one would rather avoid if possible?

demiankatz:
Thanks, @EreMaijala and @aleksip -- I'll see if @crhallberg has anything to add, and I'll wait for resolution of the span conversation before merging anything.

aleksip:
The span just a thought, so no need to add it here, as we can do it when a real world requirement actually appears.

EreMaijala:
Well, I have no problem adding the span. Just a sec!

EreMaijala:
Err... the default translations seem to be broken. Fixing, will take a bit longer.

demiankatz:
@EreMaijala, are you saying you've found a translation-related bug? If so, should that be handled separately so we can backport it to release-8.0, or am I misunderstanding the situation?

EreMaijala:
@demiankatz Yes. I'll post a separate PR based on release-8.0.

EreMaijala:
This one is ready.

themes/bootstrap3/js/truncate.js
```diff
@@ -98,8 +99,16 @@ VuFind.register('truncate', function Truncate() {
       }
 
       if (shouldTruncate) {
-        var btnMore = '<button type="button" class="btn more-btn' + btnClass + '">' + moreLabel + ' <i class="fa fa-arrow-down" aria-hidden="true"></i></button>';
-        var btnLess = '<button type="button" class="btn less-btn' + btnClass + '">' + lessLabel + ' <i class="fa fa-arrow-up" aria-hidden="true"></i></button>';
+        var btnMore = '<button type="button" class="btn more-btn' + btnClass + '">'
+          + (settings['label-icons'] === 'before' ? '<i class="fa fa-arrow-down" aria-hidden="true"></i> ' : '')
```
demiankatz: Should we define upIcon and downIcon variables so we don't have to repeat the icon definitions? In addition to reducing redundancy, I think that will make it easier to update this code in future when we roll out the icon helper everywhere.

EreMaijala:
The additional fix: #2192.

themes/bootstrap3/js/truncate.js
EreMaijala: Makes sense, I'll revise.

EreMaijala: I've now made the icons part of the settings array.

crhallberg (approved):
We'll have to merge this into dev before #1962 but that shouldn't be a problem.

demiankatz:
> 
> 
> We'll have to merge this into dev before #1962 but that shouldn't be a problem.

Good point, @crhallberg. I've added a new TODO checkbox on #1962 to re-check the code for any new explicit FontAwesome references before merging, since it's possible some things will slip through the cracks into dev while we're finishing up our discussions there.

themes/bootstrap3/templates/Recommend/SideFacets/cluster-list.phtml
```diff
@@ -14,6 +14,7 @@
     'less-label' => $this->transEsc('less') . ' ...',
     'wrapper-class' => false,
     'wrapper-tagname' => 'div',
+    'label-icons' => 'before'
```
demiankatz: Would it make more sense to change the default to `before` instead of just explicitly changing the behavior in this one place? Right now, top-facets also uses the truncate mechanism, and it looks strange having some facets with the icon on one side, and other facets with the icon on the other. If there is a preference for "icon on left," then I think perhaps we should just do that everywhere, rather than risk visual inconsistencies.

This also makes me wonder if we need to account for this in right-to-left mode. Should we detect if we're using a RTL language and flip-flop the icon setting in that situation?

EreMaijala: I think both would make sense. I'll revise.

EreMaijala: Default changed to 'before'. Looks like no changes are necessary for RTL mode work correctly, unless I'm missing something.

demiankatz:
Thanks, @EreMaijala, this looks good to me, and you're right that RTL works as expected... I'm not sure why I thought that was a problem yesterday. :-)

PR #2192, opened 2021-11-03
Fix truncate default labels.

EreMaijala:
Initialize the default settings in the initTruncate method to make sure that translations have been added to the VuFind object. Use translations that are actually available.


demiankatz:
Cherry-picked to release-8.0 and merged forward from there; thanks!

PR #2194, opened 2021-11-04
Fix handling response cache in XCNCIP2 ILS driver

xmorave2:


module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
```diff
@@ -1998,6 +1999,7 @@ public function renewMyItems($renewDetails)
     {
         $details = [];
         foreach ($renewDetails['details'] as $detail) {
+            $username = $renewDetails['patron']['cat_username'];
```
demiankatz: This value doesn't appear to change, so would it be better to set it above the foreach loop?

demiankatz (commented):
Thanks, @xmorave2 -- just one very minor suggestion.

Also, a question: when was this bug introduced? Should I backport this to the release-8.0 branch, or does this only apply to the dev branch?

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
xmorave2: You are right, it is fixed now

xmorave2:
@demiankatz: This bug was introduced before 8.0 was released, so should be backported. Thanks

demiankatz:
Thanks, @xmorave2. Unfortunately, it doesn't look like I can do a clean cherry-pick from here to release-8.0, so I'm just going to merge as-is to dev. I suspect XCNCIP2 adoption is low enough that it is okay for this to wait until release 8.1, but if anyone feels greater urgency, I imagine that manually backporting this would not take too much effort. :-)

PR #2195, opened 2021-11-04
Add place request message translations for NCIP driver

xmorave2:
@demiankatz Feel free to change the English translation to whatever you feel is better


demiankatz:
@xmorave2, there are already `hold_success`, `hold_error_fail`, `storage_retrieval_request_place_success` and `storage_retrieval_request_error_fail` messages in the main language files. Can we use these existing translations instead of creating new ones?

I realize that this opens up a conversation about where the strings should live; maybe we need to refactor everything to move these (and other request-related) strings into the ILSMessages text domain. One possible compromise could be to copy the strings for now (using the command line language tools -- no need to manually copy and paste anything) and then open a JIRA ticket to refactor these strings in future. I'd be interested to hear what @EreMaijala thinks about this, too.

EreMaijala (commented):
I think there are two separate issues here:
1.) I don't think `sysMessage` is needed or appropriate here when it doesn't convey any additional information and doesn't actually come from the backend system. If additional status would be needed, we should add support for the `status` field like in `cancelHolds`. In other words, simply returning `['success' => true]` should be enough so that VuFind knows to display the normal success message to the user.
2.) If the message was indeed needed, ILSMessages would be the appropriate place. But e.g. for the `hold_success` message that doesn't originate from the ILS driver, I'm not so sure. This is of course debatable, but I'd say that only messages coming from the ILS drivers (or relayed by them from the backend systems) should live in ILSMessages.

xmorave2:
Thanks @EreMaijala. This does make sense to me. I refactored the code int that way and added detailed sysMessage parsed from NCIP response when place a request fails.

demiankatz:
Thanks, @xmorave2 and @EreMaijala. This seems much more self-contained now, so I am going to merge it.

PR #2196, opened 2021-11-04
Translate message for user in MultiBackend driver

xmorave2:


demiankatz:
@xmorave2, I notice that we seem to have some inconsistency in our templates about how sysMessage values from ILS drivers are handled. It seems that we sometimes translate them and other times simply escape them. It seems to me that we should make this more consistent, and it seems it might be simplest if we always translate them, so we don't need to do additional translation work inside the drivers. This once again raises the question of whether strings might need to be moved into the ILSMessages text domain, however... I'm copying @EreMaijala into this thread since he may have opinions, especially since he is probably the one who created that untranslated sysMessage string in the MultiBackend driver, and perhaps it is being handled differently in Finna than in VuFind core.

EreMaijala (changes requested):
I think translating all sysMessages (in templates where they're displayed) would make sense. Making MultiBackend translate the message would only fix that single case. Putting the messages into the ILSMessages domain would make sense, though it would be a backward incompatible change unless we fallback to default domain.

demiankatz:
Thanks, @EreMaijala. As a small step forward, would it make sense to change the MultiBackend driver to explicitly return the string with the "ILSMessages::" prefix and then adjust all sysMessages displays in templates to be sure they include translation? That would allow us to add this missing translation for VuFind 8.1 without any backward compatibility issues (unless there is something in Finna relying on this string). We could then undertake a separate project targeted against 9.0 to do a bigger reorganization of ILS-related strings. If @xmorave2 thinks there is value in keeping some of the refactoring work done here around translation, we could potentially repurpose this current PR for that while moving the other details into a new one. What do you think?

EreMaijala:
@demiankatz Sounds good to me.

demiankatz:
@EreMaijala, one other question: do you have local translations for hold_wrong_user_institution that are worth sharing, or is this something that was defined but never translated?

EreMaijala:
@demiankatz I don't have translations. We always use "driver" as holds mode so this doesn't actually happen unless checkRequestIsValid somehow returns true. Note that the same message is also used in placeStorageRetrievalRequest, so both should be updated.

PR #2197, opened 2021-11-04
Make lightbox loading spinner nicer.

EreMaijala:
This makes the lightbox loading spinner an overlay when the lightbox has content. This makes the spinner more visible and less likely to mess with the contents or shift it around.

Also includes a fix to avoid double spinner when loading the initial content.


EreMaijala:
While working on merging latest VuFind changes to Finna I realized that I couldn't make the spinner with the "Loading" message look good without making some more changes. I also found that not all content shown in lightbox has a h2 element with lightbox-header class, which made the spinner placement erratic. This avoids it all.

demiankatz:
Thanks, @EreMaijala. I'll wait for @crhallberg's review on this and then can do some additional hands-on testing as needed. (At a glance, it looks fine to me, but the lightbox can hold surprises sometimes).

themes/bootstrap3/js/lightbox.js
```diff
@@ -474,7 +474,7 @@ VuFind.register('lightbox', function Lightbox() {
   // Element which to focus after modal is closed
   var _beforeOpenElement = null;
   function reset() {
-    _html(VuFind.loading());
+    _html('');
```
crhallberg, line 477: Do we want to reset to the loading overlay?

themes/bootstrap3/js/lightbox.js
```diff
@@ -318,9 +318,9 @@ VuFind.register('lightbox', function Lightbox() {
...
     // Loading
-    var lightboxHeader = $(_modalBody).find('.lightbox-header');
-    if (lightboxHeader.length > 0) {
-      lightboxHeader.after(VuFind.loading());
+    $(_modalBody).find('.modal-loading-overlay,.loading-spinner').remove();
+    if (_modalBody.children().length > 0) {
+      _modalBody.prepend('<div class="modal-loading-overlay">' + VuFind.loading() + '</div>');
     } else {
       _modalBody.prepend(VuFind.loading());
     }
```
crhallberg: This isn't a fault with your improvement but I think we should move this to the ajax function so that links and forms both show the loading overlay. This is working beautifully for forms but not links thought this seems to be an existing issue.

EreMaijala: True, changed.

themes/bootstrap3/js/lightbox.js
EreMaijala: No. I originally removed the default spinner to avoid double spinners when the lightbox is opened from a form submit button. Now that the ajax method adds the spinner in any case, it makes sense to let it handle it unconditionally. And making the modal empty makes the check for loading message style easier; if lightbox is empty, only a single line "Loading..." message is added since the overlay would look silly without underlying content.

crhallberg:
Excellent! I think this looks really nice! Thank you very much!

demiankatz:
Thanks -- sounds like this is now ready for merging. Next question: should we target this for 8.1 or 9.0? It seems like a somewhat significant change, so maybe merging to dev-9.0 is preferable... but let me know if you feel differently. Either way, I'll run the full test suite to be sure everything is still passing.

themes/bootstrap3/js/lightbox.js
```diff
@@ -318,9 +318,9 @@ VuFind.register('lightbox', function Lightbox() {
       }, false);
     }
     // Loading
-    var lightboxHeader = $(_modalBody).find('.lightbox-header');
-    if (lightboxHeader.length > 0) {
-      lightboxHeader.after(VuFind.loading());
+    $(_modalBody).find('.modal-loading-overlay,.loading-spinner').remove();
```
LuomaJuha: Just a small notice, but the conversion of _modalBody to jQuery object might be not necessary?

demiankatz:
For what it's worth, all tests still pass. I'll wait to see if anyone has a response to @LuomaJuha's latest comment before taking further action on my end.

themes/bootstrap3/js/lightbox.js
EreMaijala: Thanks, fixed!

EreMaijala:
@demiankatz @LuomaJuha was right. Now fixed.

EreMaijala:
I'm fine with dev-9.0, though the current way the "loading" message is added to the lightbox in 8.x may prove problematic to others as well. When it was only a spinner it wasn't that prominent.

demiankatz:
You're right, @EreMaijala, this does kind of live on the threshold between bug fix and outright change. Perhaps the most conservative solution is to merge it to dev-9.0, and we can always cherry-pick it back to 8.1 if we decide that we need it there. Or we could go directly to 8.1 now, if people prefer. @crhallberg, any opinion?

EreMaijala:
LGTM is being weird, by the way. It complains about not being able to merge the latest commit with something from 2016.

demiankatz:
@EreMaijala, that's probably my fault. I accidentally clicked the wrong button and changed the base branch to something really old. I changed it back, but probably not before LGTM tried to interact with it.

crhallberg:
I don't have a strong opinion but since it's an improvement, I think merging it earlier (into 8.1) is better.

demiankatz:
Yeah, you've persuaded me that we probably prevent more problems than we cause if we just merge to dev for 8.1. I'll do that and add a changelog note.

PR #2198, opened 2021-11-05
[VUFIND-1364] Add support for storing extra data in search objects.

EreMaijala:


module/VuFind/src/VuFind/Search/Base/Results.php
```diff
@@ -554,6 +554,35 @@ public function getRawSuggestions()
...
+     *
+     * @param array $data Extra data
+     *
+     * @return void
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     */
+    public function setExtraData(array $data): void
+    {
+        // Not implemented in the base class
```
demiankatz, line 583: Should we throw an exception or log a warning if non-empty data is sent here?

module/VuFind/src/VuFind/Search/Minified.php
```diff
@@ -74,20 +74,54 @@ class Minified
     public $hf = [];
 
     /**
-     * ID, start tIme, query Speed, Result total, search TYpe, search CLass id
+     * Search ID, start tIme, query Speed, Result total, search TYpe, search CLass id
```
demiankatz: This comment can be trimmed down to only reference the id.

module/VuFind/src/VuFind/Search/Base/Results.php
EreMaijala: I'd rather not. It would cause vanilla VuFind to error with a search saved with Finna, and I'd rather let it silently ignore the data, even if it would mean it wouldn't represent the original search exactly.

module/VuFind/src/VuFind/Search/Minified.php
EreMaijala: Fixed.

module/VuFind/src/VuFind/Search/Base/Results.php
demiankatz: That's a good reason to avoid an exception, but what about a call to `error_log`? My concern is just avoiding a situation where parameters disappear inexplicably without any warning. If you think even that is too noisy, I can live without it, but just wanted to fully advocate for the possibility. :-)

EreMaijala: I suppose a simple error_log message is fine, though I'm not sure if that's necessary either.

EreMaijala: Log message added.

demiankatz: Yeah, I really don't expect that this is ever likely to happen... but I just don't like having a setter that does absolutely nothing, because it seems like an opening for unexpected behavior. :-)

PR #2200, opened 2021-11-09
Revert encodeURIComponent from hierarchical facets

LuomaJuha:
It broke the exclude button

TODO:
- [x] backport to release-8.0 branch if confirmed to be needed

### DIFF
--- a/themes/bootstrap3/js/facets.js
+++ b/themes/bootstrap3/js/facets.js
@@ -47,7 +47,7 @@ function buildFacetNodes(data, currentPath, allowExclude, excludeTitle, counts)
       badge.appendChild(document.createTextNode(facet.count.toString().replace(/\B(?=(\d{3})+\b)/g, separator)));
       html.appendChild(badge);
       if (allowExclude) {
-        var excludeUrl = encodeURIComponent(currentPath + facet.exclude);
+        var excludeUrl = currentPath + facet.exclude;
         var a = document.createElement('a');
         a.className = 'exclude';
         a.setAttribute('href', excludeUrl);


demiankatz:
@LuomaJuha, can you elaborate on when/how this breaks the exclude button? When this change was introduced, I tested that exclude buttons worked as expected, and the automated test suite also passed. Is there an edge case that triggers this problem? If the problem is broader, I want to be sure there isn't a flaw in my testing process.

demiankatz:
I've also added a note that once we finish this discussion, if we don't find an alternate solution, we need to backport this fix to release-8.0, since I already backported the commit which introduced the change there.

demiankatz:
Now that I read more carefully, perhaps the problem is that this is specific to hierarchical facets, and tests may only cover "normal" facets. I'll double-check that now....

LuomaJuha:
Yeah, I tested it also on "normal" facet so I didn't see that it didn't work..

LuomaJuha:
And about the backporting, is there something I should do differently on this pr so it goes directly to the correct place?

demiankatz:
@LuomaJuha, I have found that there was indeed a gap in the test suite, so I'm working on adding a test to cover this case. You don't need to do anything special to support the backporting. I'm just going to cherry-pick the commit. Should all be done soon!

LuomaJuha:
@demiankatz, thanks!

demiankatz:
Thanks again, @LuomaJuha -- backported as 9882b7e52e9680c7e6114f37e98b36b5d67a1861 (and merged through dev and dev-9.0 from there). Test added at 344e2602245f76363047fcd69fede37fae1b6b56.

PR #2201, opened 2021-11-09
Make Csrf an interface for better customizability

xrosecky:


demiankatz:
@xrosecky, I think this would be easier to understand if the interfaces were named with Interface at the end, which is the convention throughout VuFind. I'm also not sure it's desirable to embed Laminas in a class name when that's more of an implementation detail than anything else. So perhaps rather than having `LaminasTokenCountingCsrf implements TokenCountingCsrf` it would be better to have `TokenCountingCsrf implements TokenCountingCsrfInterface`.

I also assume the intent is to target release 9.0 here, since this changes quite a few things in a backward-incompatible way. (I don't think this will cause people many problems, but it's a bigger change than I think we could justify for a feature release).

demiankatz:
I'm also interested in whether @EreMaijala has any thoughts on this before you spend a lot of time on revisions. :-)

xrosecky:
Renaming to TokenCountingCsrfInterface is good idea. Should I rename also the Csrf interface to CsrfInterface?

xrosecky:
@demiankatz Is there any reason why is not CSRF used for validation in VuFind\Controller\Plugin\Holds instead of storing IDs in user's session?

demiankatz:
@xrosecky, yes, I would recommend renaming Csrf to CsrfInterface as well -- it's helpful to be able to identify interfaces from the class/file names.

Regarding the holds validation, that logic was written before Csrf was introduced to VuFind. Its intent is not just to prevent session hijacking but also to prevent users from trying to force holds on items that do not support them. (Not sure how likely that is, but some ILS APIs don't validate as well as they should -- or didn't ten years ago, anyway -- so we added this as an extra layer of protection).

xrosecky:
@demiankatz Done, I renamed the classes.

module/VuFind/src/VuFind/Auth/EmailAuthenticator.php
```diff
@@ -58,7 +59,7 @@ class EmailAuthenticator implements \VuFind\I18n\Translator\TranslatorAwareInter
     /**
      * CSRF Validator
      *
-     * @var \VuFind\Validator\Csrf $csrf CSRF validator
+     * @var \VuFind\Validator\TokenCountingCsrfInterface $csrf CSRF validator
```
demiankatz: Since you have a use statement for this above, you don't have to use the fully qualified class name here.

module/VuFind/src/VuFind/Auth/EmailAuthenticator.php
```diff
@@ -117,7 +118,7 @@ class EmailAuthenticator implements \VuFind\I18n\Translator\TranslatorAwareInter
      */
     public function __construct(
         \Laminas\Session\SessionManager $session,
-        \VuFind\Validator\Csrf $csrf,
+        \VuFind\Validator\TokenCountingCsrfInterface $csrf,
```
demiankatz: You can also remove namespace here due to use statement above.

demiankatz (commented):
Just some minor aesthetic comments for now; I'll wait to see if @EreMaijala has any more detailed suggestions.

Also, one other thought: is the TokenCountingCsrfInterface required everywhere that it is used here, or might there be places where a less specific interface would be more appropriate?

xrosecky:
> Also, one other thought: is the TokenCountingCsrfInterface required everywhere that it is used here, or might there be places where a less specific interface would be more appropriate?

@demiankatz Fixed in the latest commit.

module/VuFind/tests/unit-tests/src/VuFindTest/Validator/LaminasTokenCountingCsrfTest.php
```diff
@@ -39,7 +39,7 @@
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
  */
-class CsrfTest extends \PHPUnit\Framework\TestCase
+class LaminasTokenCountingCsrfTest extends \PHPUnit\Framework\TestCase
```
demiankatz: This test class should be renamed to match the renaming of the class under test.

demiankatz (changes requested):
Thanks for the progress, @xrosecky -- just one more observation!

demiankatz:
...once that is fixed, I'll try running the full test suite on this to be sure none of the integration tests are affected (unless you have already done this).

xrosecky:
OK, fixed. Please run the integration tests.

demiankatz:
@xrosecky, all tests are passing! I'll just wait for @EreMaijala's feedback (probably post-Summit next week) to determine next steps.

EreMaijala (approved):
I'm not sure it's that important to emphasize the token counting nature of the CSRF validator that it needs to be in the interface name. I'd rather have a single interface that includes all the methods, but that's probably because I can't think of a situation where the methods couldn't or shouldn't be supported. That said, I can live with the current interface name.

demiankatz:
Thanks, @EreMaijala. I agree that I can live with the name, but if there is not a use case for an alternate strategy/implementation, it might not be necessary. @xrosecky, did you name it this way because you envision a different strategy or implementation of CSRF that does not use token counting?

xrosecky:
@demiankatz Our use case is that we need implementation of CSRF that does not depend on user session - for example by storing CSRF token in database - to avoid overwriting of session in concurrent AJAX requests.

demiankatz:
@xrosecky, would it make sense in that case to have multiple CSRF implementations in core VuFind, with a configuration setting to determine which one is used? And if we go that route, do the current interface names make sense, or would we need to use something more general to make it easier to swap in different implementations?

xrosecky:
@demiankatz Yes, it makes sense to have multiple implementation of CSRF in core VuFind. There can also be implementation of CSRF that does not use token counting feature at all - for example job that deletes all expired tokens periodically in database. In that case the token counting methods can be dummy.

I have no preference regarding the interface names.

demiankatz:
@xrosecky, on deeper review of the code, I definitely agree that we should simplify the interface naming. It appears that getTokenCount() is only used in a unit test, not in any VuFind code... so the only method that we really need in the interface from the token-counting version is trimTokenList(), and I think this could easily be implemented in different ways in different contexts. Would you mind consolidating the interfaces to simplify this?

xrosecky:
@demiankatz Do I understand correctly?:

- remove the interface TokenCountingCsrf and move method trimTokenList() to CsrfInterface
- rename the TokenCountingCsrf class to SessionCsrf or do you have better name?

demiankatz:
@xrosecky, yes, that sounds like a good solution to me!

xrosecky:
@demiankatz OK, I will consolidate the interfaces.

xrosecky:
@demiankatz Done.

demiankatz:
Thanks, @xrosecky, this looks really good to me now. Just one more small suggestion: should we add an alias from `VuFind\Validator\Csrf` to `VuFind\Validator\CsrfInterface` to improve backward compatibility if somebody is fetching the old service name somewhere? (We could flag it with a `// deprecated` comment if we want to eliminate it in a future release).

If @EreMaijala has no further suggestions, I think we can merge this into dev-9.0 very soon. I'll just wait to hear your thoughts on the alias issue, and then I can run the test suite one last time to be safe before merging.

module/VuFind/config/module.config.php
```diff
@@ -520,7 +520,8 @@
             'VuFind\Translator' => 'Laminas\Mvc\I18n\Translator',
             'VuFind\WorldCatUtils' => 'VuFind\Connection\WorldCatUtils',
             'VuFind\YamlReader' => 'VuFind\Config\YamlReader',
-            'Laminas\Validator\Csrf' => 'VuFind\Validator\Csrf',
+            'Laminas\Validator\Csrf' => 'VuFind\Validator\CsrfInterface',
```
EreMaijala: Interface is what the service must provide, but since there could be several classes providing the interface, ot seems a bit backwards to me to handle services by the interface. I'd rather map `'Laminas\Validator\Csrf' => 'VuFind\Validator\Csrf'` and `'VuFind\Validator\Csrf' => 'VuFind\Validator\SessionCsrf'`.

EreMaijala (changes requested):
I might be missing something, but I don't think we should be requesting services by the interface. So all the factories should ask for `\VuFind\Validator\Csrf:class` and that should be mapped to the actual class to be used. Otherwise we end up in a really weird situation if we want to have more than one validator active at a time for different purposes, but still use the same interface for them.

module/VuFind/config/module.config.php
xrosecky: Should I use string `'VuFind\Validator\Csrf'` or non-existing class `\VuFind\Validator\Csrf:class`  in factories?

EreMaijala: Good question, I'm not sure about that. Maybe @demiankatz has a further idea. Since the plain Csrf class doesn't exist it might be better to use `\Laminas\Validator\Csrf::class => \VuFind\Validator\SessionCsrf::class` and `'\VuFind\Validator\Csrf' => \VuFind\Validator\SessionCsrf::class` instead.

demiankatz: I think the only place we should use the string 'VuFind\Validator\Csrf' is in the module configuration to define an alias for backward compatibility. In any code that requests a service, we should pull either \VuFind\Validator\CsrfInterface::class (if we need a service that satisfies this interface, but we don't care which one), or \VuFind\Validator\SessionCsrf::class (if there is a particular reason why we would want specifically this implementation). I don't think it's a problem that there might be multiple implementations of the interface in the code; the idea of requesting an interface from the service manager is to find out which implementation of the class is configured to be preferred. I suspect in the future if we add multiple CSRF implementations, we'll change our factories anyway to use some kind of plugin manager and new configuration.

I realize this requires @xrosecky to undo some work from the most recent commit. I apologize for not getting this response written a little faster!

EreMaijala: Ok. Sorry for making a mess out of this, @xrosecky.

xrosecky: @demiankatz OK, I added VuFind\Validator\Csrf alias.

demiankatz:
Thanks again, @xrosecky. This is looking good now. However, when I test-merged this with dev-9.0 and ran the test suite, I encountered a problem with the entire test process hanging. I don't think it is related to this PR directly, but I want to get to the bottom of it before I merge this, just to be sure. Hopefully I can figure it out and then get this merged later this week!

demiankatz:
As I suspected, the test problems were unrelated to the changes here (see #2216) and everything is passing. I'm going to merge this to dev-9.0 now!

PR #2202, opened 2021-11-09
Fix balance in Aleph driver

xmorave2:


module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -1447,7 +1447,7 @@ public function getMyFines($user)
             $checkout = $finesListSort[$key]["checkout"];
             $transactiondate = $finesListSort[$key]["transactiondate"];
             $transactiontype = $finesListSort[$key]["transactiontype"];
-            $balance += $finesListSort[$key]["amount"];
+            $balance = $finesListSort[$key]["amount"];
```
demiankatz: $balance is initialized to 0 above in line 1428 and then assigned in line 1437. If we're no longer manipulating that value, is there a reason to use a variable at all? It might make the overall code less confusing to make some changes there.

xmorave2: You are right, I will simplify it a bit.

demiankatz:
Thanks for the progress on this, @xmorave2 -- and adding test coverage is a great idea! I see this is currently a draft, so please let me know when you're ready for further review.

xmorave2:
@demiankatz It is ready for a review now

demiankatz:
Thanks, @xmorave2. My main question is whether it's ever necessary to account for credits, because it seems like the old code tried to, but the new code simply filters them out. Without a detailed knowledge of Aleph, I'm not sure whether or not this might be a problem....

xmorave2:
@demiankatz I asked my colleagues about Aleph (as don't know it much) and you are right. So I changed it back, just fix handling the credit/debit distinction.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -1419,51 +1417,28 @@ public function getMyFines($user)
             $barcode = (string)$z30->{'z30-barcode'};
             $checkout = (string)$z31->{'z31-date'};
             $id = $this->barcodeToID($barcode);
-            $mult = ($transactiontype == "Credit") ? 100 : -100;
+            $mult = (string)($item->attributes()->{'type'} ?? '') == 'debit'
```
demiankatz: I notice that the old code checked for "Credit" (with a capital C) while the new code checks for "debit" (all lowercase). Does this indicate that there may be inconsistent casing between Aleph installations? If so, would it be worth putting in a strtolower here just to be on the safe side?

demiankatz (commented):
Thanks -- this makes sense to me; just one small question.

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
xmorave2: Before, the text value of element was checked, which is in Aleph literally a free text, fully configurable on Aleph's site and even translatable. That's the reason why it mostly didn't work. Now I check the xml attribute, which should be always lower cased... But to be absolutly sure, we could lowercased it. I added a followup.

demiankatz: Yes, that makes sense. Thank you! Just one last very minor detail -- you're using `->{'type'}` here, but you can simplify this to `->type`. You only need to use the "braces-and-string" syntax when a key contains characters which confuse the PHP parser... so we need it for the properties with dashes in them, but not for a simple type key.

xmorave2: Should be fixed now. Thank for all your feedback @demiankatz

demiankatz:
Thanks again, @xmorave2 -- I just merged this, and have also asked for feedback from @welblaud in case he has anything to add (though I felt safe enough merging this given the relatively straightforward nature of the changes, and the increase in test coverage).

PR #2203, opened 2021-11-11
[VUFIND-1476] Improvements to the Evergreen ILS driver

gmcharlt:
This patch fixes and enhances the current Evergreen ILS driver by:

- updating authentication to use bcrypt password hashing, which was
  introduced in Evergreen 2.10. This breaks compatibility with
  versions of Evergreen older than 2.10, but those are years past
  Evergreen community support
- fix various places where an Evergreen item record ID was returned
  when a bib ID was wanted.
- Various fixes to retrieving current checkouts:
    - use the new return style for getMyTransactions
    - ensure that titles are displayed
    - return the due time for hourly loans
    - return owning and borrowing locations
    - return information about available renewals
- Various fixes to retrieving current fines:
    - excluded transactions for which nothing has been billed
    - link to the relevant title if the billing is for a loan
    - return the due date, fine creation date, and original loan
      date where applicable and properly distinguish between them
- Various fixes to retrieving hold requests:
    - ensure that all active holds are retrieved
    - excluded canceled holds
    - indicate when a hold is available, on the pickup shelf, frozen,
      or has an item in transit to fill it
- Various fixes to item and bib retrieval
    - for now, limit the new item pull to 50 copies and only available items
- Various fixes to the user profile retrieval:
    - include country, city, and expiration date
- Various modernizations:
    - Use VuFind's date and time formatter

Signed-off-by: Galen Charlton <gmc@equinoxOLI.org>

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Evergreen.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Evergreen.php
@@ -894,7 +894,9 @@ HERE;
      */
     private function _formatDate($date)
     {
+        if (!$date) {
+            return '';
+        }
         return $this->dateConverter->convertToDisplayDate('Y-m-d', $date);
-        //return $date ? (new \DateTime($date))->format('Y-m-d') : '';
     }
 }
--- a/module/VuFind/src/VuFind/ILS/Driver/Evergreen.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Evergreen.php
@@ -22,6 +22,7 @@
  * @category VuFind
  * @package  ILS_Drivers
  * @author   Warren Layton, NRCan Library <warren.layton@gmail.com>
+ * @author   Galen Charlton, Equinox Open Library Initiative <gmcharlt@equinoxOLI.org>
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org/wiki/development:plugins:ils_drivers Wiki
  */
@@ -43,8 +44,10 @@ use VuFind\Exception\ILS as ILSException;
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org/wiki/development:plugins:ils_drivers Wiki
  */
-class Evergreen extends AbstractBase
+class Evergreen extends AbstractBase implements \Laminas\Log\LoggerAwareInterface
 {
+    use \VuFind\Log\LoggerAwareTrait;
+
     /**
      * Database connection
      *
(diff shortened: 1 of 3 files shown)


module/VuFind/src/VuFind/ILS/Driver/Evergreen.php
```diff
@@ -814,4 +881,22 @@ public function findReserves($course, $inst, $dept)
...
+     * Format date
+     *
+     * This formats a date coming from Evergreen for display
+     *
+     * @param string $date The date string to format; may be null
+     *
+     * @throws ILSException
+     * @return string The formatted date
+     */
+    private function _formatDate($date)
```
demiankatz: As a matter of style, VuFind generally favors protected over private to ensure that everything is overrideable in custom subclasses. You might want to adjust this method accordingly.

demiankatz (changes requested):
Thanks, @gmcharlt, this is great!

Regarding your question on the JIRA ticket about testing, I'm willing to take your word on these changes in the short term, given that the existing code probably won't work, there don't seem to be active VuFind/Evergreen users who will be impacted by this, and getting these changes in place will ensure that new users have a solid foundation to work from. 

In the long term, I think the best thing to do would be to get better unit test coverage in place. That's difficult to do meaningfully with database-driven classes like this one, but if you're planning on replacing this with an API-driven version, it would be comparatively easy to create fixtures with real Evergreen API responses that could be used to build tests. If you build that new driver in a test-driven way, then it should be pretty reliable, and we can deprecate this one.

What do you think? Let me know if I can be of any help with the test development, or anything else...

module/VuFind/src/VuFind/ILS/Driver/Evergreen.php
gmcharlt: Thanks for the feedback; I've made that method protected rather than private.

gmcharlt:
> Regarding your question on the JIRA ticket about testing, I'm willing to take your word on these changes in the short term, given that the existing code probably won't work, there don't seem to be active VuFind/Evergreen users who will be impacted by this, and getting these changes in place will ensure that new users have a solid foundation to work from.

I'm aware of at least one production Evergreen+VuFind site out there, but I can give them a heads-up. Also, I don't think that they're currently using any of the patron account features at the moment, just getStatus.

I'm satisfied with my own testing, but also, I'm not going anywhere and will keep an eye on bug reports.

> In the long term, I think the best thing to do would be to get better unit test coverage in place. That's difficult to do meaningfully with database-driven classes like this one, but if you're planning on replacing this with an API-driven version, it would be comparatively easy to create fixtures with real Evergreen API responses that could be used to build tests. If you build that new driver in a test-driven way, then it should be pretty reliable, and we can deprecate this one.

Agreed, I can readily write unit tests when I develop the forthcoming API-based driver.

module/VuFind/src/VuFind/ILS/Driver/Evergreen.php
demiankatz: Thanks!

demiankatz:
Thanks, @gmcharlt, that all sounds good, so I'm going to merge this and resolve your JIRA ticket. Feel free to open a new ticket for the API-driven driver if you like (or you can just open an in-progress PR whenever you're ready).

PR #2204, opened 2021-11-15
add subject from OJS to topic and topic-facet fields

horaciod:
Topics and Topic_facet fields are missing in imported records from OJS


horaciod:
add subject to ojs xls importer.

demiankatz:
Thanks, @horaciod!

PR #2205, opened 2021-11-16
Update en.ini

tigran54:


horaciod:
Are you modifying the translations in the correct file?
I think that you should create a new file with the language to translate (example: in.ini)

demiankatz:
Merged into #2166.

PR #2206, opened 2021-11-16
Method validateIds in AbstractRequestBase

xrosecky:
We need some refactoring to disable validation of IDs against session in our code and this PR makes it possible.

### DIFF
--- a/module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
+++ b/module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
@@ -136,13 +136,16 @@ abstract class AbstractRequestBase extends AbstractPlugin
     }
 
     /**
-     * Get remembered valid IDs
+     * Validate supplied IDs against remembered IDs. Returns true if all supplied
+     * IDs are remembered, otherwise returns false.
      *
-     * @return array
+     * @param array $ids IDs to validate
+     *
+     * @return bool
      */
-    public function getValidIds(): array
+    public function validateIds($ids): bool
     {
-        return $this->getSession()->validIds ?? [];
+        return (bool)array_diff($ids, $this->getValidIds());
     }
 
     /**
@@ -353,5 +356,15 @@ abstract class AbstractRequestBase extends AbstractPlugin
             date('d') + $d,
             date('Y') + $y
         );
+    }
+
+    /**
+     * Get remembered valid IDs
+     *
+     * @return array
+     */
+    protected function getValidIds(): array
+    {
+        return $this->getSession()->validIds ?? [];
     }
 }
--- a/module/VuFind/src/VuFind/Controller/HoldsController.php
+++ b/module/VuFind/src/VuFind/Controller/HoldsController.php
@@ -208,7 +208,7 @@ class HoldsController extends AbstractBase
         }
         // If the user input contains a value not found in the session
         // legal list, something has been tampered with -- abort the process.
-        if (array_diff($selectedIds, $this->holds()->getValidIds())) {
+        if ($this->holds()->validateIds($selectedIds)) {
             $this->flashMessenger()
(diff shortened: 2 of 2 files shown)


demiankatz:
This seems reasonable to me as a non-backward-breaking improvement. I'm a little surprised that there's not parallel logic to this for the ILL and storage retrieval request modules, but I couldn't find any. Is that cause for concern, or are holds just more complicated than those other types of requests?

xrosecky:
Yes, I was also surprised that the code from AbstractRequestBase is used only in HoldsController.

demiankatz:
Any thoughts, @EreMaijala? (I can dig into this in more detail after the Summit if research is needed).

EreMaijala:
Validation is probably one of things that should have been added to ILL and storage retrieval request modules, and that's why the methods are in the base class. I probably just forgot to do that..

demiankatz:
There is also outstanding work to refactor more logic out of the MyResearchController. Maybe we can find time to do these things as part of the same project. Perhaps at least worth a JIRA ticket, and maybe some discussion during roadmapping tomorrow. (Getting rid of MyResearch is still a "nice-to-have" on my list).

EreMaijala (approved):
If we do these changes in dev-9.0, I'm fine with this. However, I'm not sure moving `getValidIds` from public to protected in dev is a good idea.

EreMaijala:
@demiankatz Now that I think of it, the idea when I was working on ILL and SR requests could have been that the driver would bear more responsibility on doing the right thing, but validating IDs still wouldn't hurt. I created https://vufind.org/jira/browse/VUFIND-1521 as a placeholder. It's true that refactoring of MyResearch would be nice, but I don't think that's necessary for adding the validation.

module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
```diff
@@ -136,13 +136,16 @@ public function rememberValidId($id)
...
-     * @return array
+     * @param array $ids IDs to validate
+     *
+     * @return bool
      */
-    public function getValidIds(): array
+    public function validateIds($ids): bool
     {
-        return $this->getSession()->validIds ?? [];
+        return array_diff($ids, $this->getValidIds());
```
demiankatz: Should we typecast this to (bool) or use some other check that creates a bool in order to be a little more strict about our typing?

module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
```diff
@@ -354,4 +357,14 @@ protected function getDateFromArray($dateArray)
             date('Y') + $y
         );
     }
+
+    /**
+     * Get remembered valid IDs
+     *
+     * @return array
+     */
+    protected function getValidIds(): array
```
demiankatz, line 366: To @EreMaijala's point, if we keep this public here, we avoid backward compatibility breaks. Would you like to change it back to public so we can merge to dev, and then I can make it protected in dev-9.0, or do you prefer to simply merge this change directly to dev-9.0?

demiankatz (commented):
Thanks for the input, @EreMaijala. In light of that, I took another look, which also led me to notice one other very minor thing. See comments below.

Also, regarding the MyResearch refactor, I agree that it's certainly not a prerequisite for adding more validation... just thought that one project might help to justify the completion of the other in the same chunk of work. :-)

module/VuFind/src/VuFind/Controller/Plugin/AbstractRequestBase.php
xrosecky: @demiankatz You can merge this change only to dev-9.0.

demiankatz:
Thanks, @xrosecky; I have changed the base and will merge this to dev-9.0 now.

PR #2210, opened 2021-11-16
Update advsearch.phtml

tigran54:


horaciod:
Are you modifying the files in the correct folder?
You should create a new folder with the language to translate (example: in)

demiankatz:
Merged into #2166.

PR #2211, opened 2021-11-16
Update geosearch.phtml

tigran54:


EreMaijala:
@tigran54 I see that you have placed several pull request updating language files. However, the changes affect the en (English) versions of these files. I believe that your intention would be to provide these files in another language, so e.g. this one should be `themes/root/templates/HelpTranslations/LNG/geosearch.phtml` where LNG is the proper language code.

demiankatz:
Thanks for noticing that, @EreMaijala. @tigran54, if you're not sure how to fix these, let me know and I can help with the renaming.

demiankatz:
Merged into #2166.

PR #2216, opened 2021-11-16
Fix test hang related to static RouteGenerator logic.

demiankatz:


demiankatz:
For reasons that I admit that I don't fully understand, the introduction of static RouteGenerator properties in #2172 causes our PHPUnit integration tests to hang. Resetting the static properties every time we include module.config.php resolves the problem. This PR does that, and moves configuration loading to a separate test trait to centralize this shared behavior, since we may want to revisit/revise it in future.

I found this problem very difficult to troubleshoot because there's not an obvious spot in the code where things are hanging when the problem occurs. Commenting out the static logic in the RouteGenerator makes the tests run, and leaving it in causes hard-to-trace bad behavior. It's possible this is related to a bug in PHP or PHPUnit that will go away in the future... or maybe I'm just overlooking some obvious detail. In any case, I understand in theory why statics can be problematic in tests (because the static values persist across multiple tests) but I can't figure out in practice what the problem is here -- I couldn't find any evidence of values accumulating to an unreasonable degree, nor do I understand why having a bunch of extra route data would break everything. All I know is that this fix seems to solve things, and I was able to implement it in a way that doesn't really intrude on the core code outside of the test suite, so I can live with it for now.

I welcome feedback/ideas from @mtrojan-ub and others, though, if they have anything to add!

EreMaijala:
@mtrojan-ub @demiankatz `addRecordRoute` is adding each routeBase+controller pair into `self::$recordRoutes` unconditionally. Since addNonTabRecordAction iterates `self::$recordRoutes`, it blows up the array. Something like storing the routes by key in the array could help:

```
        // Store the added route in case we need to add
        // more nonTabRecordActions later
        self::$recordRoutes["$routeBase::$controller"] = [
            'routeBase' => $routeBase,
            'controller' => $controller,
        ];

```

mtrojan-ub:
@EreMaijala, that's a good point! @demiankatz, can you please try it out?

demiankatz:
Sure enough, @EreMaijala's proposed solution of storing the routes by key solves my problem (and is, generally speaking, a good idea). I still don't understand exactly why PHPUnit was hanging the way it was, though, or why it was so difficult to trace the failure, but at this point I don't think I need to lose sleep over that. Please let me know if you have any objections to merging this; with approval from either one of you, I think I'm comfortable moving forward with this solution.

EreMaijala (approved):
I've also tested this, and it works for me whereas without the change the integration tests would exceed the memory limit.

demiankatz:
Great, thanks again! Merging now.

PR #2217, opened 2021-11-17
Move currency formatting to a separate class from SafeMoneyFormat.

EreMaijala:
This makes currency formatting independent of theme and view, and avoids an issue where the KohaRest driver could be initialized before theme initialization when Bootstrapper::initUserLanguage would call the auth manager. This also avoids a potential extraneous escaping of the formatted value in KohaRest and GetUserFines.

This is targeted at dev branch, but if all the changes are too much for dev, I can pick the addition of CurrencyConverter separately. Can't really avoid the constructor change in KohaRest driver in any case, so I think we actually may need all this in dev. And this is pretty important fix. Without the fix with KohaRest driver, a logged-in user just gets an error if their fine limit is exceeded, because KohaRest won't have the helper in place.

### DIFF
--- a/module/VuFind/src/VuFind/Service/CurrencyFormatter.php
+++ b/module/VuFind/src/VuFind/Service/CurrencyFormatter.php
@@ -1,0 +1,106 @@
+<?php
+/**
+ * Currency formatter
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2010.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Service
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+namespace VuFind\Service;
+
+use NumberFormatter;
+
+/**
+ * Currency formatter
+ *
+ * @category VuFind
+ * @package  Service
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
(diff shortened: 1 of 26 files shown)


module/VuFind/src/VuFind/Service/CurrencyFormatter.php
```diff
@@ -0,0 +1,106 @@
+<?php
+/**
+ * Currency formatter
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2010.
```
demiankatz: Should probably make this 2021.

module/VuFind/src/VuFind/Service/CurrencyFormatter.php
```diff
@@ -0,0 +1,106 @@
...
+     * @return string
+     */
+    public function convertToDisplayFormat($number, $currency = null)
+    {
+        if (null === $currency) {
+            $currency = $this->defaultCurrency;
+        }
+        // Workaround for a problem in ICU library < 4.9 causing formatCurrency to
+        // fail if locale has comma as a decimal separator.
+        // (see https://bugs.php.net/bug.php?id=54538)
```
demiankatz: bugs.php.net is not responding right now, so I can't look at the details of this... but I wonder if this is now so old that it is no longer an issue and this code can be simplified.

demiankatz (commented):
Thanks, @EreMaijala, this looks good -- just a couple of small comments.

As far as where to merge this, I wonder if the conservative approach is to add the new service to 8.1 and only change KohaRest, to minimize backward-breaking changes. We can leave the existing view helper as-is to keep things stable (and simply live with the redundancy caused by that), cleaning things up fully in 9.0. If you're okay with that strategy, I'm happy to do the necessary work to split the diffs between the dev and dev-9.0 branches when this is all finished.

module/VuFind/src/VuFind/Service/CurrencyFormatter.php
EreMaijala: Well, it's mostly copy-paste from SafeMoneyFormat, so I thought it would be most appropriate to keep the old year.

module/VuFind/src/VuFind/Service/CurrencyFormatter.php
EreMaijala: Removed the code. Seems to be a fairly old ICU version that had it, and e.g. Ubuntu 18.04 LTS has a much newer one.

EreMaijala:
@demiankatz Your proposed approach sounds good to me.

module/VuFind/src/VuFind/Service/CurrencyFormatter.php
demiankatz: I usually update dates when I create new files, but I don't have strong feelings about this. No objections to leaving as-is. :-)

module/VuFind/src/VuFind/Service/CurrencyFormatter.php
demiankatz: Looks like the underlying ICU bug was fixed in 2018 in v49, so I agree that this should be safe to remove. I checked my PHP 7.3 installation on RHEL7 (which is about the oldest realistic platform that anyone would be trying to run VuFind 8.1 on), and it's using ICU v50.

demiankatz:
Thanks, @EreMaijala -- I have done what I suggested, with commit 8d70097 putting minimum changes into the dev branch; I will now merge the rest to dev-9.0.

PR #2218, opened 2021-11-17
Add ThemeConfig view helper.

crhallberg:
A view helper to pull data from theme.config.php via ThemeInfo's new mergedConfig.

This is the first step towards putting more appearance configuration in theme.config.php. For example, we could use this in the new theme to organize items in the header and nav.

- [ ] Update to match laminas-cache v3 standards (#2299) after merging dev to dev-9.0.


module/VuFind/src/VuFind/View/Helper/Root/ThemeConfig.php
```diff
@@ -0,0 +1,105 @@
...
+     * - ['less'] => same as above
+     * - ['less', 'active'] => would return LESS active status
+     *
+     * @param string|string[] $path Path to return from theme.config.php
+     *
+     * @return mixed|null
+     */
+    public function __invoke($path = [])
+    {
+        if (!is_array($path)) {
```
demiankatz: You can typecast `$path = (array)$path;` to do this in a single line.

module/VuFind/src/VuFind/View/Helper/Root/ThemeConfig.php
```diff
@@ -0,0 +1,105 @@
...
+use Laminas\Cache\Storage\StorageInterface;
+use Laminas\View\Helper\AbstractHelper;
+use VuFindTheme\ThemeInfo;
+
+/**
+ * Theme config view helper
+ *
+ * @category VuFind
+ * @package  View_Helpers
+ * @author   Demian Katz <demian.katz@villanova.edu>
```
demiankatz, line 39: Don't forget to give yourself credit. :-)

module/VuFind/src/VuFind/View/Helper/Root/ThemeConfigFactory.php
```diff
@@ -0,0 +1,83 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  View_Helpers
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Mario Trojan <mario.trojan@uni-tuebingen.de>
```
demiankatz, line 25: More author tags that you might want to change.

demiankatz (commented):
This makes sense to me. Just a few minor comments. We should also add a unit test; let me know if you need any help with that!

aleksip:
Appearance configuration is something many non-technical end users will want to do. Putting it in a PHP file, which also contains other technical configuration, might not be ideal from this perspective?

module/VuFind/src/VuFind/View/Helper/Root/ThemeConfig.php
crhallberg: Oh, awesome!

crhallberg:
I agree, it's not ideal but I don't think it would be a huge issue if we keep it well-organized and well-documented. Switching to another text config format would probably not improve things much.

The most ideal solution I can think of at the moment is a DevTools visual customization tool. It might look something similar [Bootstrap's old online customization tool](https://getbootstrap.com/docs/3.4/customize/) which is easier to use than a PHP file but is daunting and less than ideal in many ways (probably why they stopped using it).

module/VuFind/src/VuFind/View/Helper/Root/ThemeConfig.php
demiankatz: I see that you resolved the comment but didn't change the code. Is there something that needs to get pushed up?

demiankatz:
Regarding the conversation about putting configurations in PHP files, a few thoughts/comments:

1.) I agree that it's not ideal, but also that proliferation of additional config files just to avoid it is probably no better. It's currently necessary to create a theme.config.php to set up a theme, so in theory, users should already be familiar with this process. Adding more settings in there could potentially make it harder to navigate, but that's really a problem with all formats.

2.) There's nothing stopping anyone from calling `parse_ini_file` from appropriate parts of their local theme.config.php file if they really want to manage settings through an .ini file instead -- perhaps a useful strategy if some users maintain the theme and other users maintain the settings. Not sure that we'd want to make this a standard practice, but it might be a recommendation we could make for some situations.

3.) Are there some settings that might have defaults in config.ini that could be overridden through theme configuration? That might be a "best of both worlds" approach, though it could also be confusing/complicated to maintain.

I think perhaps our next step should be to start collecting a list of settings that might benefit from being moved into theme config; having some use cases might make it a little easier to continue this conversation.

crhallberg:
We should definitely draw attention to `parse_ini_file` in theme.config.php.

I still think it's the better place for strictly visual configurations.

demiankatz:
@crhallberg, do you think perhaps the local theme example template would be the best place to put that comment? Maybe we should add a TODO comment here so we don't forget this detail.

module/VuFind/src/VuFind/View/Helper/Root/ThemeConfig.php
```diff
@@ -0,0 +1,104 @@
...
+        $cacheKey = $this->themeInfo->getTheme() . $key;
+        $cached = $this->cache->getItem($cacheKey);
+
+        if ($cached == null) {
+            $cached = $this->themeInfo->getMergedConfig($key, true);
+            $this->cache->setItem($cacheKey, $cached);
+        }
+
+        foreach ($path as $p) {
+            $cached = $cached[$p] ?? null;
```
LuomaJuha: What if it would be done like this:

if ($cached = $cached[$p] ?? false) {
    return $cached[$p]
}

and in the end we could return return []; to always assure that it is an array. Then we could declare type return for the __invoke() function :array. And casting to array could be removed if the parameter $path = [] had array type declared?

demiankatz: I think the critical difference between your proposal and the existing code is that existing code will return the LAST match it finds, whereas yours would return the FIRST match. Of course, if that matters, we could do an array_reverse in combination with your proposal to retain the existing behavior. I haven't taken the time to analyze what's going on here, so I'm not sure if it really makes a difference... but I think that's the next detail to think about at any rate.

LuomaJuha: Yea, it can be crucial difference if it matters :) Just a thought tho!

crhallberg: What @LuomaJuha proposed would work if we were to flatten the hierarchy of the config when we load it. That might speed things up.

Currently:
```php
$cached = [
    'a' => [
        'b' => 'magic',
    ],
];
```

Flattened
```php
$cached = [
    'a' => ['b' => 'magic'],
    'b' => 'magic',
];
```

demiankatz: Ahh, I understand now -- this isn't about searching a series of paths for possible settings, it's about descending down a tree according to a specific set of branches. I don't think flattening things would make that better, honestly.

demiankatz: I wonder if this would be more clear if at some point we replaced the `$cached` variable with `$nextNode` or something like that -- because we first retrieve our tree from the cache, and then we navigate it... and once we start navigating it, a different variable name might make it a little more clear what we are doing.

module/VuFind/src/VuFind/View/Helper/Root/ThemeConfig.php
```diff
@@ -0,0 +1,106 @@
...
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+class ThemeConfig extends AbstractHelper
+{
+    /**
+     * Cache for merged configs
+     *
+     * @var StorageInterface
+     */
+    protected $cache = [];
```
demiankatz: We shouldn't initialize this to an array since that's the wrong type and it gets set in the constructor anyway.

module/VuFind/src/VuFind/View/Helper/Root/ThemeConfigFactory.php
```diff
@@ -0,0 +1,83 @@
...
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options sent to factory.');
+        }
+
+        $themeInfo = $container->get(\VuFindTheme\ThemeInfo::class);
+
+        // As of release 1.1.0, the memory storage adapter has a flaw which can cause
+        // unnecessary out of memory exceptions when a memory limit is enabled; we
+        // can disable these problematic checks by setting memory_limit to -1.
+        $cacheConfig = ['name' => 'memory', 'options' => ['memory_limit' => -1]];
```
demiankatz, line 77: We'll need to remember to update this to match laminas-cache v3 standards (see #2299) after merging dev to dev-9.0. Might be worth adding a TODO checkbox on the PR.

demiankatz (commented):
Thanks for the progress, @crhallberg -- I think this is nearly ready to merge... just some minor nitpicks/notes.

demiankatz (approved):

PR #2219, opened 2021-11-18
[VUFIND-1365] Add admin API with a cache flush action.

EreMaijala:
Includes changes to make it possible to include definitions and other shared elements in the base Swagger spec.

I've tried to design the admin API controller so that additional admin functionality can be added in future with granular permissions. Admin API's Swagger spec fragment is created in code as opposed to a template to allow for this granularity (endpoints are not advertised unless the user has access).

Note that this does not currently address the use case of removing a specific image from the cover cache. I think that should be a separate API method.

### DIFF
--- a/module/VuFindApi/config/module.config.php
+++ b/module/VuFindApi/config/module.config.php
@@ -4,12 +4,14 @@ namespace VuFindApi\Module\Configuration;
 $config = [
     'controllers' => [
         'factories' => [
+            'VuFindApi\Controller\AdminApiController' => 'VuFindApi\Controller\AdminApiControllerFactory',
             'VuFindApi\Controller\ApiController' => 'VuFindApi\Controller\ApiControllerFactory',
             'VuFindApi\Controller\SearchApiController' => 'VuFindApi\Controller\SearchApiControllerFactory',
             'VuFindApi\Controller\Search2ApiController' => 'VuFindApi\Controller\Search2ApiControllerFactory',
             'VuFindApi\Controller\WebApiController' => 'VuFindApi\Controller\WebApiControllerFactory',
         ],
         'aliases' => [
+            'AdminApi' => 'VuFindApi\Controller\AdminApiController',
             'Api' => 'VuFindApi\Controller\ApiController',
             'SearchApi' => 'VuFindApi\Controller\SearchApiController',
             'Search2Api' => 'VuFindApi\Controller\Search2ApiController',
@@ -18,14 +20,34 @@ $config = [
     ],
     'service_manager' => [
         'factories' => [
+            'VuFindApi\ApiPluginManager' => 'VuFind\ServiceManager\AbstractPluginManagerFactory',
             'VuFindApi\Formatter\FacetFormatter' => 'Laminas\ServiceManager\Factory\InvokableFactory',
             'VuFindApi\Formatter\RecordFormatter' => 'VuFindApi\Formatter\RecordFormatterFactory',
             'VuFindApi\Formatter\Search2RecordFormatter' => 'VuFindApi\Formatter\Search2RecordFormatterFactory',
             'VuFindApi\Formatter\WebRecordFormatter' => 'VuFindApi\Formatter\WebRecordFormatterFactory',
         ],
     ],
+    'vufind_api' => [
+        'register_controllers' => [
+            \VuFindApi\Controller\AdminApiController::class,
+            \VuFindApi\Controller\SearchApiController::class,
+            \VuFindApi\Controller\Search2ApiController::class,
+            \VuFindApi\Controller\WebApiController::class,
+        ]
+    ],
     'router' => [
         'routes' => [
+            'adminClearCacheApiv1' => [
+                'type' => 'Laminas\Router\Http\Literal',
+                'verb' => 'delete,options',
+                'options' => [
+                    'route'    => '/api/v1/admin/cache',
+                    'defaults' => [
+                        'controller' => 'AdminApi',
+                        'action'     => 'clearCache',
+                    ]
+                ]
+            ],
             'apiHome' => [
                 'type' => 'Laminas\Router\Http\Segment',
                 'verb' => 'get,post,options',
--- a/module/VuFindApi/config/module.config.php
+++ b/module/VuFindApi/config/module.config.php
@@ -20,7 +20,6 @@ $config = [
     ],
     'service_manager' => [
         'factories' => [
-            'VuFindApi\ApiPluginManager' => 'VuFind\ServiceManager\AbstractPluginManagerFactory',
             'VuFindApi\Formatter\FacetFormatter' => 'Laminas\ServiceManager\Factory\InvokableFactory',
             'VuFindApi\Formatter\RecordFormatter' => 'VuFindApi\Formatter\RecordFormatterFactory',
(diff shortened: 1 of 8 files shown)


module/VuFindApi/src/VuFindApi/Controller/AdminApiController.php
```diff
@@ -0,0 +1,173 @@
...
+            return $result;
+        }
+
+        try {
+            $cacheList = $this->getRequest()->getQuery()->get('id')
+                ?: $this->getDefaultCachesToClear();
+            foreach ((array)$cacheList as $id) {
+                $cache = $this->cacheManager->getCache($id);
+                if (!($cache instanceof \Laminas\Cache\Storage\FlushableInterface)) {
+                    throw new \Exception("Cache $id is not flushable");
```
demiankatz: Do we want a single unflushable cache to kill the whole process? Couldn't this cause some caches to get flushed and others to get skipped, if the problem happens in the middle of the sequence? Would it be better to collect an array of cache-specific messages and then report them all at the end of the process instead of letting the first problem terminate the process? I'm thinking we could have a structure like:

<pre>
{
    'errorCount': 1,
    'details': [
        { 'name': 'cache1', 'error': false, 'message': 'ok' },
        { 'name': 'cache2', 'error': true, 'message': 'Cache cache2 is not flushable' },
        { 'name': 'cache3', 'error': false, 'message': 'ok' },
    ]
}
</pre>

Maybe this is overkill, but it might allow finer-grained control. If you don't think it's worth the effort now, we could always open it as a possible future enhancement ticket.

module/VuFindApi/src/VuFindApi/Controller/AdminApiController.php
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * Get an array of caches to clear by default
+     *
+     * @return array
+     */
+    protected function getDefaultCachesToClear(): array
+    {
+        return array_diff(
```
demiankatz: Could we filter the unflushable caches at this layer to reduce the odds of errors?

module/VuFindApi/src/VuFindApi/Controller/AdminApiController.php
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Constructor
+     *
+     * @param ServiceLocatorInterface $sm Service locator
+     * @param CacheManager            $cm Cache manager
+     */
+    public function __construct(ServiceLocatorInterface $sm, CacheManager $cm)
+    {
+        $this->serviceLocator = $sm;
```
demiankatz: Would it be better to call parent::__construct() rather than just assigning a property here that is not actually used by this subclass? I still hope to someday refactor controllers so they don't all depend on the top-level service manager...

module/VuFindApi/src/VuFindApi/Controller/ApiControllerFactory.php
```diff
@@ -44,6 +44,18 @@
  */
 class ApiControllerFactory implements FactoryInterface
 {
+    /**
+     * API controllers to register
+     *
+     * @var array
+     */
+    protected $apiControllers = [
```
demiankatz: Is this property unused?

module/VuFindApi/src/VuFindApi/Controller/ApiController.php
```diff
@@ -124,7 +133,6 @@ protected function getApiSpecs()
             }
         }
 
-        // Return the fragment without the enclosing curly brackets
-        return substr(trim(json_encode($results, JSON_PRETTY_PRINT)), 1, -1);
+        return $results;
```
demiankatz, line 136: I'm having a little trouble following why we were able to remove the brace-stripping logic in the refactoring, but I assume it works, so I'm probably just missing something.

demiankatz (commented):
Thanks, @EreMaijala -- haven't done hands-on testing yet, but a few thoughts from my initial read-through. I'm not sure how much we need to actually worry about unflushable caches, so feel free to ignore or defer suggestions on that topic as appropriate.

module/VuFindApi/src/VuFindApi/Controller/AdminApiController.php
EreMaijala: Good idea, done.

module/VuFindApi/src/VuFindApi/Controller/AdminApiController.php
EreMaijala: Now that the caches are filtered, I think this shouldn't be necessary. If you feel it'd still be needed, I can add it, but I wouldn't expect issues normally.

module/VuFindApi/src/VuFindApi/Controller/AdminApiController.php
EreMaijala: Mistake, now fixed.

module/VuFindApi/src/VuFindApi/Controller/ApiControllerFactory.php
EreMaijala: Yes, it was left over as a part of previous iteration.

module/VuFindApi/src/VuFindApi/Controller/ApiController.php
EreMaijala: That was because previously the returned string was added in the main JSON response as a string, so it couldn't have had the surrounding braces. Now that it's handled as an array, that hackish approach is no longer necessary.

EreMaijala:
@demiankatz Thanks for the comments. Should be addressed now.

module/VuFindApi/src/VuFindApi/Controller/ApiController.php
demiankatz: Makes sense, thanks for clarifying!

module/VuFindApi/src/VuFindApi/Controller/AdminApiController.php
demiankatz: We can always add it in the future if we need it; as you say, probably not a requirement at this point.

module/VuFindApi/config/module.config.php
```diff
@@ -18,14 +20,34 @@
     ],
     'service_manager' => [
         'factories' => [
+            'VuFindApi\ApiPluginManager' => 'VuFind\ServiceManager\AbstractPluginManagerFactory',
```
demiankatz: Is this unused? I don't see a definition for this class.

demiankatz (changes requested):
Thanks, @EreMaijala -- this looks good and tests check out. I just happened to notice what I think may be a stray line of unneeded configuration during a final review. Mind giving that one more look and let me know if I'm mistaken?

demiankatz:
Also, please let me know if you have time to document this in the wiki, or if you'd like me to do that -- I assume we should add a mention of it somewhere.

module/VuFindApi/config/module.config.php
EreMaijala: Unused, remnants of a bad idea. Now removed.

EreMaijala:
@demiankatz Good point, I can add some documentation.

EreMaijala:
See https://vufind.org/wiki/development:apis:admin

demiankatz:
Thanks, @EreMaijala -- merging now!

PR #2220, opened 2021-11-26
XCNCIP2: Request all pages of LookupItemSet

xmorave2:
Some responders return paginated results even if we don't want them to paginate


demiankatz:
Thanks, @xmorave2 -- this looks good to me. Before I merge, though, one question: do we need to impose any limits on this? What happens if there are hundreds or thousands of pages? (Probably not likely, but just thinking about edge cases).

xmorave2:
@demiankatz I added a configuration option, which add possibility to set limit for problematic responders if needed. This is disabled by default.

demiankatz:
Thanks, @xmorave2! I just added a note to the comment in the .ini file about the default value/behavior, then I merged this.

PR #2221, opened 2021-11-26
Add user object as usable parameter when building notification emails

tmikkonen:
When sending due date reminder and scheduled alert emails this addition makes it possible to include the user's first name e.g. in the greeting to present a bit more personal feel to the recipient.


demiankatz:
Thanks, @tmikkonen. I see that this change is causing a test failure; please let me know if you need help figuring that out!

EreMaijala:
@tmikkonen  I think it would make sense to include the whole user object instead of just the firstname. This would allow more flexibility in the template to e.g. do something with the last name if first name is empty.

demiankatz:
Thanks, @EreMaijala and @tmikkonen -- this does seem improved... but it seems that tests still need to be adjusted to recognize this change.

tmikkonen:
Seems to have done it.
And I learned something about running tests and only one test locally, so... :)

demiankatz:
Looks good! Thanks, @tmikkonen!

PR #2222, opened 2021-12-01
Add minimal timeout before returning to previous page after printing

aleksip:
We noticed that on Firefox the browser does not return to the previous page after printing. This is apparently because the event is fired a bit too early, when the print dialog is still open. See [here](https://stackoverflow.com/questions/68154482/afterprint-event-on-firefox-89) for more information.

Funnily enough, the same timeout trick that works for updating Ajax results on Safari works for this as well!


demiankatz:
I was seeing this problem too, but only under Linux, not in Windows, so I thought maybe it was a quirk of my setup. Apparently it's a broader issue. Your solution seems reasonable to me, though -- thanks for the fix!

PR #2223, opened 2021-12-01
Move record collection factory creation to a separate method.

EreMaijala:
This in addition to the record collection class being defined in a property allows use of a different record collection class and/or factory without having to override the whole createBackend method.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Factory/SolrDefaultBackendFactory.php
+++ b/module/VuFind/src/VuFind/Search/Factory/SolrDefaultBackendFactory.php
@@ -30,7 +30,9 @@ namespace VuFind\Search\Factory;
 
 use VuFindSearch\Backend\Solr\Backend;
 use VuFindSearch\Backend\Solr\Connector;
+use VuFindSearch\Backend\Solr\Response\Json\RecordCollection;
 use VuFindSearch\Backend\Solr\Response\Json\RecordCollectionFactory;
+use VuFindSearch\Response\RecordCollectionFactoryInterface;
 
 /**
  * Factory for the default SOLR backend.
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/Search/Factory/SolrDefaultBackendFactory.php
```diff
@@ -82,11 +91,23 @@ protected function getSolrCore()
...
+    protected function createRecordCollectionFactory()
+        : RecordCollectionFactoryInterface
+    {
         $manager = $this->serviceLocator
             ->get(\VuFind\RecordDriver\PluginManager::class);
-        $factory
-            = new RecordCollectionFactory([$manager, $this->createRecordMethod]);
-        $backend->setRecordCollectionFactory($factory);
-        return $backend;
+        return new RecordCollectionFactory(
```
demiankatz, line 108: Should we also make RecordCollectionFactory a class property, so you can swap out the factory class without overriding the method?

demiankatz (commented):
Thanks, @EreMaijala, this makes sense to me -- just one small comment. Also, this counts as progress on [VUFIND-1446](https://vufind.org/jira/browse/VUFIND-1446). Do you think there is further work to be done along similar lines, or should we close that ticket?

module/VuFind/src/VuFind/Search/Factory/SolrDefaultBackendFactory.php
EreMaijala: I didn't do it yet because createRecordCollectionFactory is pretty simple and the likelihood of needing to override the factory is small, but I'll be happy to make it that way.

EreMaijala: Now that I look at it more broadly, I think I'll do a bit more refactoring...

EreMaijala:
Continued in #2225.

PR #2224, opened 2021-12-01
Make form sender fields overridable.

EreMaijala:
This makes it possible to e.g. define the field positions on the form and add additional fields like a phone number in the correct position. These and other reserved fields are now documented in FeedbackForms.yaml.

The email field now accepts the maximum length that an email address can be.

### DIFF
--- a/module/VuFind/src/VuFind/Form/Form.php
+++ b/module/VuFind/src/VuFind/Form/Form.php
@@ -548,14 +548,24 @@ class Form extends \Laminas\Form\Form implements
         $elements = [];
         $configuredElements = $this->getFormElements($config);
 
-        // Add sender contact name & email fields
+        // Defaults for sender contact name & email fields:
         $senderName = [
-            'name' => 'name', 'type' => 'text', 'label' => 'feedback_name',
-            'group' => '__sender__'
+            'name' => 'name',
+            'type' => 'text',
+            'label' => $this->translate('feedback_name'),
+            'group' => '__sender__',
+            'settings' => [
+                'size' => 50
+            ]
         ];
         $senderEmail = [
-            'name' => 'email', 'type' => 'email', 'label' => 'feedback_email',
-            'group' => '__sender__'
+            'name' => 'email',
+            'type' => 'email',
+            'label' => $this->translate('feedback_email'),
+            'group' => '__sender__',
+            'settings' => [
+                'size' => 254
+            ]
         ];
         if ($formConfig['senderInfoRequired'] ?? false) {
             $senderEmail['required'] = $senderEmail['aria-required']
@@ -567,8 +577,6 @@ class Form extends \Laminas\Form\Form implements
         if ($formConfig['senderEmailRequired'] ?? false) {
             $senderEmail['required'] = $senderEmail['aria-required'] = true;
         }
-        $configuredElements[] = $senderName;
-        $configuredElements[] = $senderEmail;
 
         foreach ($configuredElements as $el) {
             $element = [];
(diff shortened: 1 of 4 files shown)


module/VuFind/src/VuFind/Form/Form.php
```diff
@@ -548,14 +548,24 @@ protected function parseConfig($formId, $config, $params)
...
         $configuredElements = $this->getFormElements($config);
 
-        // Add sender contact name & email fields
+        // Defaults for sender contact name & email fields:
         $senderName = [
-            'name' => 'name', 'type' => 'text', 'label' => 'feedback_name',
-            'group' => '__sender__'
+            'name' => 'name',
+            'type' => 'text',
+            'label' => $this->translate('feedback_name'),
```
demiankatz, line 555: Was it a bug that this was not translated before, or are we now double-translating it?

module/VuFind/src/VuFind/Form/Form.php
```diff
@@ -673,9 +681,27 @@ protected function parseConfig($formId, $config, $params)
                     $element['settings']['rows'] = 8;
                 }
             }
+
+            // Merge sender fields with any existing field definitions:
+            if ('name' === $element['name']) {
+                $element = array_merge($senderName, $element);
+                $senderName = null;
+            } elseif ('email' === $element['name']) {
+                $element = array_merge($senderEmail, $element);
```
demiankatz: Should we add a test case to prove that this new logic works as expected?

demiankatz (commented):
Makes sense to me... just one question and one comment.

module/VuFind/src/VuFind/Form/Form.php
EreMaijala: Previously it was translated in the loop below, but now it resembles the result of changes made in the loop. In other words, previously it was a configuration entry, while now it's an internal field definition. So, neither.

module/VuFind/src/VuFind/Form/Form.php
EreMaijala: Yes. I'll add the tests.

module/VuFind/src/VuFind/Form/Form.php
demiankatz: Ahh, of course, makes sense!

module/VuFind/src/VuFind/Form/Form.php
demiankatz: Thanks! I expect I'll be able to merge this whenever you have a chance to complete that.

EreMaijala: Added tests and fixed an issue with default field size handling being done too early and overriding the sender fields. Thanks for suggesting the tests! :)

demiankatz: Great! Writing tests sometimes feels like a burden, but it pays off very frequently -- glad this was one of those cases.

One suggestion for an additional test case: what happens if you provide a custom email element with a settings array that does not include a size? Does the default size get merged in correctly, or do all the settings get overwritten? I _think_ the current code will work as desired, but the behavior of `array_merge` on nested arrays is not especially well-documented, so it might be good to capture our desired behavior in a test just to be sure things are doing what they're supposed to.

EreMaijala: Good point. It didn't work as it overrides the content of top-level elements. I switched to array_replace_recursive, added tests and also fixed the code to not override aria-label if it's set in the settings of a field.

PR #2225, opened 2021-12-01
[VUFIND-1446] Refactor Solr backend factories to simplify overriding of record coll…

EreMaijala:
…ection factory.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
+++ b/module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
@@ -448,18 +448,22 @@ abstract class AbstractSolrBackendFactory implements FactoryInterface
         $builder = new QueryBuilder($specs, $defaultDismax);
 
         // Configure builder:
-        $search = $this->config->get($this->searchConfig);
-        $caseSensitiveBooleans
-            = $search->General->case_sensitive_bools ?? true;
-        $caseSensitiveRanges
-            = $search->General->case_sensitive_ranges ?? true;
-        $helper = new LuceneSyntaxHelper(
-            $caseSensitiveBooleans,
-            $caseSensitiveRanges
-        );
-        $builder->setLuceneHelper($helper);
+        $builder->setLuceneHelper($this->createLuceneSyntaxHelper());
 
         return $builder;
+    }
+
+    /**
+     * Create Lucene syntax helper.
+     *
+     * @return LuceneSyntaxHelper
+     */
+    protected function createLuceneSyntaxHelper()
+    {
+        $search = $this->config->get($this->searchConfig);
+        $caseSensitiveBooleans = $search->General->case_sensitive_bools ?? true;
+        $caseSensitiveRanges = $search->General->case_sensitive_ranges ?? true;
+        return new LuceneSyntaxHelper($caseSensitiveBooleans, $caseSensitiveRanges);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
```diff
@@ -142,6 +142,27 @@ abstract class AbstractSolrBackendFactory implements FactoryInterface
...
+     * @var string
+     */
+    protected $recordCollectionFactoryClass = RecordCollectionFactory::class;
+
+    /**
+     * Method for creating a record driver.
+     *
+     * @var string
+     */
+    protected $createRecordMethod = 'getSolrRecord';
```
demiankatz: I'd suggest, instead of using a property, you add a `getCreateRecordCallback` method. This has a couple of advantages:

1.) It allows you to return null here in the abstract factory, since otherwise the code changes here constitute a behavior change (causing subclasses to use getSolrRecord instead of no custom callback by default).

2.) It allows you to override less code in SolrReservesBackendFactory and SolrWebBackendFactory below, since you can put the custom callback logic in its own method.

demiankatz (changes requested):
This makes things much better, but I have one suggestion that I think will further improve it.

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
EreMaijala: Right. Now implemented.

module/VuFind/src/VuFind/Search/Factory/SolrReservesBackendFactory.php
```diff
@@ -55,24 +55,20 @@ public function __construct()
...
-     * Create the SOLR backend.
+     * Create the record collection factory.
      *
-     * @param Connector $connector Connector
-     *
-     * @return \VuFindSearch\Backend\Solr\Backend
+     * @return RecordCollectionFactoryInterface
      */
-    protected function createBackend(Connector $connector)
+    protected function createRecordCollectionFactory()
```
demiankatz: Is there a reason you didn't use `getCreateRecordCallback()` instead of the whole `createRecordCollectionFactory()` method here?

module/VuFind/src/VuFind/Search/Factory/SolrWebBackendFactory.php
```diff
@@ -81,24 +81,20 @@ protected function getSolrUrl($config = null)
...
-     * Create the SOLR backend.
+     * Create the record collection factory.
      *
-     * @param Connector $connector Connector
-     *
-     * @return \VuFindSearch\Backend\Solr\Backend
+     * @return RecordCollectionFactoryInterface
      */
-    protected function createBackend(Connector $connector)
+    protected function createRecordCollectionFactory()
```
demiankatz: Same question here...

demiankatz (commented):
I think we're nearly done, but one last question....

module/VuFind/src/VuFind/Search/Factory/SolrReservesBackendFactory.php
EreMaijala: No, I just forgot to check the factories after adding getCreateRecordCallback.

module/VuFind/src/VuFind/Search/Factory/SolrWebBackendFactory.php
EreMaijala: Same here. Now fixed.

demiankatz:
Looks good to me now, and all tests are passing!

PR #2226, opened 2021-12-01
[VUFIND-1446] Primo: Make backend and connector classes easily overridable.

EreMaijala:


module/VuFind/src/VuFind/Search/Factory/PrimoBackendFactory.php
```diff
@@ -154,7 +171,7 @@ protected function createConnector()
         $timeout = $this->primoConfig->General->timeout ?? 30;
         $client->setOptions(['timeout' => $timeout]);
 
-        $connector = new Connector(
+        $connector = new $this->connectorClass(
```
demiankatz, line 174: Would it be worthwhile to refactor HTTP client creation to a separate method, since that's something else that people may wish to customize?

demiankatz (commented):
Happy to merge this as-is, but one question in the interest of thoroughness.

module/VuFind/src/VuFind/Search/Factory/PrimoBackendFactory.php
EreMaijala: Yes, but this follows a pattern that all backend factories use. I'll post a separate PR to do that for all backend factories.

EreMaijala: Separation/standardization of HTTP client creation is in  #2229.

PR #2227, opened 2021-12-01
Replace class name strings with static class names.

demiankatz:
While reviewing #2223 I noticed that we had missed some of these. This PR fixes them.


EreMaijala (approved):
Looks good to me.

PR #2229, opened 2021-12-03
[VUFIND-1446] Unify HTTP client creation across backends.

EreMaijala:
Adds a common base class and modifies the Solr connector a bit to avoid storing timeout there.

Potentially controversial:
- Makes HTTP Client a mandatory constructor parameter in all connector classes.
- As a result of above change, shuffles constructor parameters so that optional ones follow the required ones.
- Turns SSL peer verification on in EDS connector by default but allows turning it off in settings (I've tested that at least the default server has a valid certificate).
- Modifies Solr Connector to avoid storing HTTP client's timeout in it.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Factory/AbstractBackendFactory.php
+++ b/module/VuFind/src/VuFind/Search/Factory/AbstractBackendFactory.php
@@ -57,24 +57,15 @@ abstract class AbstractBackendFactory implements FactoryInterface
     }
 
     /**
-     * Create service
+     * Initialize the factory
      *
-     * @param ContainerInterface $sm      Service manager
-     * @param string             $name    Requested service name
-     * @param array              $options Extra options (unused)
+     * @param ContainerInterface $sm Service manager
      *
-     * @return Backend
-     *
-     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     * @return void
      */
-    public function __invoke(
-        ContainerInterface $sm,
-        $name,
-        array $options = null
-    ) {
+    public function setup(ContainerInterface $sm)
+    {
         $this->serviceLocator = $sm;
-
-        return null;
     }
 
     /**
--- a/module/VuFind/src/VuFind/Search/Factory/AbstractBackendFactory.php
+++ b/module/VuFind/src/VuFind/Search/Factory/AbstractBackendFactory.php
@@ -1,0 +1,101 @@
+<?php
+/**
+ * Abstract factory for backends.
+ *
+ * PHP version 7
+ *
+ * Copyright (C) The National Library of Finland 2021.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Search
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Site
+ */
+namespace VuFind\Search\Factory;
+
+use Interop\Container\ContainerInterface;
+
+use Laminas\ServiceManager\Factory\FactoryInterface;
+
+/**
+ * Abstract factory for backends.
+ *
+ * @category VuFind
+ * @package  Search
+ * @author   David Maus <maus@hab.de>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Site
+ */
+abstract class AbstractBackendFactory implements FactoryInterface
+{
+    /**
+     * Superior service manager.
+     *
+     * @var ContainerInterface
+     */
+    protected $serviceLocator;
+
+    /**
+     * Constructor
+     */
+    public function __construct()
+    {
+    }
+
+    /**
+     * Create service
+     *
+     * @param ContainerInterface $sm      Service manager
+     * @param string             $name    Requested service name
+     * @param array              $options Extra options (unused)
+     *
+     * @return Backend
+     *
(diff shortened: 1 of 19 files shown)


demiankatz:
@EreMaijala, I introduced a conflict here by merging #2226; can you resolve when time permits?

EreMaijala:
@demiankatz Merged.

module/VuFind/src/VuFind/Search/Factory/AbstractBackendFactory.php
```diff
@@ -0,0 +1,101 @@
...
+     *
+     * @param ContainerInterface $sm      Service manager
+     * @param string             $name    Requested service name
+     * @param array              $options Extra options (unused)
+     *
+     * @return Backend
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     */
+    public function __invoke(
```
demiankatz: Does it really make sense to call this `__invoke`, since it doesn't actually build anything? It's really more of a setup method for use by child classes. Would it be better to call this `initialize` or `setup` and leave `__invoke` undefined in the abstract base class?

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
```diff
@@ -53,10 +55,13 @@ class WriterTest extends \PHPUnit\Framework\TestCase
      */
     public function testCommit()
     {
-        $bm = $this->getBackendManagerWithMockSolr();
+        $client = $this->getMockBuilder(\Laminas\Http\Client::class)
```
demiankatz: Since you do exactly the same thing (apart from the expected timeout value) in two places, maybe it's worth a `getMockHttpClientExpectingTimeout` method. Or maybe not. :-)

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
```diff
@@ -163,24 +171,26 @@ public function testDeleteRecords()
...
+    protected function getBackendManagerWithMockSolr($client = null)
     {
         $sm = new \Laminas\ServiceManager\ServiceManager();
         $pm = new BackendManager($sm);
         $mockBackend = $this->getMockBuilder(\VuFindSearch\Backend\Solr\Backend::class)
             ->disableOriginalConstructor()
             ->onlyMethods(['getConnector', 'getIdentifier'])
             ->getMock();
+        $handlerMap = new HandlerMap();
+        $client = $client ?? new HttpClient();
```
demiankatz: Would it be better to always use a mock client here (and in other tests) to be sure we never actually make HTTP requests?

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Backend.php
```diff
@@ -440,20 +440,17 @@ public function writeDocument(
...
-            $connector->setTimeout($timeout);
-        }
-
         // Write!
-        $connector->write($doc, $handler ?? 'update', $params);
-
-        // Restore previous timeout value, if necessary:
-        if (null !== $oldTimeout) {
-            $connector->setTimeout($oldTimeout);
+        if (is_int($timeout ?? null)) {
```
demiankatz: Would it be preferable to always use `callWithHttpOptions` here, and adjust that method so that it only does the client override when the options array is non-empty? That way, we don't have to worry about maintaining parallel argument lists in this code, which seems like it could potentially get out of sync if we refactor in the future.

I don't have strong feelings about this, and don't mind if you prefer to keep it as-is; just a thought.

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
demiankatz: ...or, given that this logic is used in multiple tests, maybe it could be part of a trait for testing with HTTP clients. But again, maybe that's overkill.

demiankatz (commented):
Thanks for this -- overall, a lot cleaner than before! Just had a few (mostly minor) comments.

I also wonder whether it's worth trying to avoid changing the constructor signature on the connector; is there a strong reason to avoid creating a default client if not provided? But if you feel it's best this way, it's easy enough to leave a changelog note.

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
EreMaijala: I considered a generic getter, but I think it would be a too far-fetched expectation that setOptions would always be called with just the timeout and just once, and if you start adding more options, the helper method easily becomes harder to use and understand than just wiring things up case by case.

I managed to convince myself to add `getMockHttpClientExpectingOnlySetTimeout` locally, however. Coming up.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Backend.php
EreMaijala: Makes sense, done.

EreMaijala:
@demiankatz I believe I've addressed your comments in the latest commit.

module/VuFind/src/VuFind/Search/Factory/AbstractBackendFactory.php
EreMaijala: Agreed. I decided to call it `setup`.

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
demiankatz: Makes sense!

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
```diff
@@ -332,6 +297,33 @@ public function query($handler, ParamBag $params, bool $cacheable = false)
...
+    /**
+     * Call a method with provided options for the HTTP client
+     *
+     * @param array  $options HTTP client options
+     * @param string $method  Method to call
+     * @param array  ...$args Method parameters
+     *
+     * @return mixed
+     */
+    public function callWithHttpOptions(
```
demiankatz, line 310: Thanks for all the progress here; I think this is coming along nicely. Just one last question: do we really want to allow arbitrary methods to be called via this public method? As written, you could easily call protected or private methods from the outside if you wanted to by using this proxy method... which I don't think is the intended effect. Do we either want to use reflection to only permit calls to public methods, or to create a fixed list of legal methods, just to be on the safe side? Or do we want to make this method protected, and then wrap it with a public `writeWithHttpOptions` method?

EreMaijala: Good point. I added a reflection-based check and tests.

demiankatz:
All tests are passing (including manual tests of all the third-party backends).

PR #2230, opened 2021-12-03
Form: Add support for labels in email subject, fix form validation etc.

EreMaijala:
This started from two requirements: 

1.) Need to be able to use select field labels in email subject
2.) Need to hide select values from the form

The other refactoring and fixes revolve around these and are so intertwined that I couldn't split this into more sensible pieces. Some notes:

1.) Avoid translating stuff early. Leave it more to the UI layer. This makes the separation more well-defined and would make it possible to use the data in other ways than just an email message (like storing feedback in database).
2.) formatEmailMessage is deprecated. I don't think the form should return the template name to use.
3.) aria attributes were removed and changed to fix HTML validation.
4.) Element id's were modified to make it less likely for them to collide with other UI elements. Also, this would make it possible to display multiple forms on a page without duplicate id's.

### DIFF
--- a/themes/root/templates/Email/form.phtml
+++ b/themes/root/templates/Email/form.phtml
@@ -6,7 +6,9 @@
 <?=$data['value']?>
 
 -----
+<?php elseif (in_array($data['type'], ['checkbox', 'radio', 'select'])): ?>
+<?=($label ? "$label: " : '') . implode(', ', array_map([$this, 'translate'], (array)$data['value'])) . PHP_EOL ?>
 <?php else: ?>
-<?= ($label ? "$label: " : '') . implode(', ', array_map([$this, 'translate'], (array)$data['value'])) . PHP_EOL ?>
+<?=($label ? "$label: " : '') . $data['value'] . PHP_EOL ?>
 <?php endif ?>
 <?php endforeach ?>
--- a/themes/root/templates/Email/form.phtml
+++ b/themes/root/templates/Email/form.phtml
@@ -1,8 +1,9 @@
 <?php foreach($this->fields as $data): ?>
 <?php $label = $this->translate($data['label'] ?? '') ?>
 <?php if ($data['type'] === 'textarea'): ?>
------
 <?=$label ? "$label: " : ''?>
+
+-----
 <?=$data['value']?>
 
 -----
--- a/themes/root/templates/Email/form.phtml
+++ b/themes/root/templates/Email/form.phtml
@@ -1,13 +1,12 @@
 <?php foreach($this->fields as $data): ?>
-<?php $label = $data['label'] ?? null ?>
-<?php $isTextarea = $data['type'] === 'textarea'; ?>
-<?php if ($isTextarea): ?>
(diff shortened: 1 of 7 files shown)


demiankatz:
@EreMaijala, I just merged #2224, which introduced some conflicts here. Can you please resolve them when time permits?

EreMaijala:
@demiankatz That was to be expected. Now resolved.

EreMaijala:
I also added a new page to be displayed when the form has been sent. This avoids displaying the form when not in lightbox.

themes/root/templates/Email/form.phtml
```diff
@@ -1,13 +1,12 @@
...
+<?php $label = $this->translate($data['label'] ?? '') ?>
+<?php if ($data['type'] === 'textarea'): ?>
 -----
 <?=$label ? "$label: " : ''?>
 <?=$data['value']?>
 
 -----
 <?php else: ?>
-<?= ($label ? "$label: " : '') . $data['value'] . PHP_EOL ?>
+<?= ($label ? "$label: " : '') . implode(', ', array_map([$this, 'translate'], (array)$data['value'])) . PHP_EOL ?>
```
demiankatz: Is there ever a value that you DON'T want to translate? Could this cause weird things to happen to user input, or is the idea that we're always going to use textareas for free-form user input?

demiankatz (commented):
It's a bit hard to review this many changes at once, but I did have one question (see below). Additionally, a few other things:

1.) The Mink FeedbackTest is failing... I'm guessing this is because you changed some IDs, so it can't select the elements it's expecting. Let me know if you need help reproducing/fixing it.

2.) I spotted a couple of small changes here that look like bug fixes -- things like adding parentheses around a `!$x = $this->function()` condition in an if statement. Is it worth pulling some of those back to dev so we can benefit from them sooner, or are they not significant enough?

Bottom line: once tests are passing, I'm reasonably comfortable with merging this... but let me know if there are any specific areas you'd like me to review with particular care.

themes/root/templates/Email/form.phtml
EreMaijala: Good point. Indeed there is, and this didn't make sense at all. I changed it to translate only values of checkbox, radio and select.

EreMaijala:
@demiankatz Sorry about making this so big. Most of the changes alone don't make that much sense, which caused things to pile up..

1.) Fixed.

2.) I don't think the fixes are significant enough. The particular case you mentioned seems to work without the parenthesis as well, but I'd rather have them in to make the intention it more obvious and readable.

I think the most important aspect to test is that translations still work properly and the resulting email looks ok and contains the right stuff. Also the different placeholders in email subject since that was one of the main drivers for these changes.

The code could use some more tests. Mink tests could include a more complex form, and unit tests should e.g. verify the contents of the resulting email message. But those have to be left for another task. I recorded them in https://vufind.org/jira/browse/VUFIND-1524.

demiankatz:
I can't seem to break it, so I have merged it. Hopefully there will be time for improved testing soon, in case we missed anything.

PR #2231, opened 2021-12-06
Added Ukrainian translation

bankovska:
Localization - Language files for Ukrainian


demiankatz:
Thanks, @bankovska, this looks great. The only details currently missing are the addition of the new language in config.ini, and the native.ini line mapping the English label Ukrainian to the name of the language _in_ Ukrainian. I'm happy to help with these details if you like!

demiankatz:
@bankovska, it also looks like a bit of stray whitespace in one of the files is causing a validation error. You can fix this by running `php $VUFIND_HOME/public/index language normalize languages` and then committing the resulting corrections. (Again, I'm happy to help with this if it would be easier).

demiankatz:
Thanks for the ongoing work, @bankovska -- since you seem to still be actively making changes right now, please let me know when you are ready for me to do a final review and (hopefully) merge. :-)

bankovska:
Yes, I updated accordingly config.ini and native.ini files. And I tried to fix the problem with files geosearch.phtml, tag.phtml and visualization.phtml. I saw that these files also had problems with encoding, which is a common thing for Cyrillic. Could you check my work - if something wrong - I'm ready to fix it. Maybe I missed something. Thank you!

demiankatz:
@bankovska, the help files look good to me. However, I believe that the encoding on the ILSMessages/uk.ini file may be incorrect -- it seems to be causing some problems. Could you make sure that's UTF-8?

demiankatz:
@bankovska, it also looks like the CallNumberFirst, Exception and HoldingStatus text domains still need to be translated (if you think it's worth doing all of them).

bankovska:
I updated ILSMessages/uk.ini and added CallNumberFirst, Exception and HoldingStatus text domains in Ukrainian. I hope it all will work now. :)

demiankatz:
Thanks, @bankovska! This is looking good now -- fully up to date with the translations released in 8.0. I've run a couple of tools (the language normalizer and php-cs-fixer) to correct some minor formatting inconsistencies, and I added one recently-added string that was missing (it's a layout-oriented string with no actual words in it, so it's the same for every language). That leaves just two recently-added lines that should eventually be translated. However, since this is complete as far as the most recent release is concerned, I'm going to merge this now. I will reach out to you in a few months when we're preparing the 8.1 release, and we can bring everything up to date at that point.

bankovska:
Thank you very much! I look forward to continuing this work.

PR #2233, opened 2021-12-07
Convert from Laminas\Db to Doctrine

demiankatz:
This is work in progress on replacing VuFind's database abstraction layer.

TODO
- [x] Handle minimum PHP version increase separately
- [x] Handle Laminas dependency upgrades separately
- [x] Evaluate whether we need BOTH `doctrine/doctrine-orm-module` and `roave/psr-container-doctrine` in composer.json, or if we can migrate fully from the psr-container-doctrine to the doctrine-orm-module.
- [x] Finish testing/developing updates in commit ec5d995
- [x] Consider handling public/index.php refactoring separately
- [x] Fully reflect all existing configuration settings
- [x] Finish designing entity and service interfaces for all tables
  - [x] access_token
    - [x] port code to use Doctrine
    - [x] remove legacy Laminas code
  - [x] auth_hash
    - [x] finish #3712
    - [x] finish #3714
    - [x] port code to use Doctrine
    - [x] remove legacy Laminas code
  - [x] change_tracker
    - [x] finish #3646
    - [x] finish #3706
    - [x] finish #3776 
    - [x] review Doctrine code and make sure time zone handling is appropriate; simplify if possible
  - [x] comments
    - [x] finish #3666
    - [x] finish reconciling CommentsService with dev version (use CommentsServiceInterface to retrieve services)
  - [x] external_session
    - [x] finish #3715
    - [x] finish #3732 
    - [x] convert expiration command to use database service
    - [x] port code to use Doctrine
    - [x] remove legacy Laminas code
  - [x] feedback
    - [x] finish #3650
    - [x] finish #3709 
    - [x] finish #3731 
  - [x] login_token
    - [x] finish #3667
    - [x] create Doctrine entity class
    - [x] port database service methods to Doctrine; remove Laminas classes
  - [x] oai_resumption
  - [x] ratings
    - [x] finish #3739
    - [x] finish #3727 
    - [x] finish #3603
  - [x] record
    - [x] finish #3716
    - [x] finish #3735
  - [x] resource
    - [x] finish #3708
    - [x] finish #3727 
    - [x] finish #3783
    - [x] finish backporting service refactoring to dev branch
    - [x] investigate whether getOrderByClause() can/should be revised (having a static method seems suboptimal)
  - [x] resource_tags
    - [x] finish #3741
    - [x] finish #3774
    - [x] finish developing ResourceTagsService
  - [x] search
    - [x] finish #3747
    - [x] finish #3775
    - [x] convert expiration command to use database service
    - [x] port database service methods to Doctrine; remove Laminas classes
  - [x] session
    - [x] finish #3653
    - [x] convert expiration command to use database service
    - [x] remove legacy Laminas code
  - [x] shortlinks
    - [x] finish #3737
    - [x] finish #3749
    - [x] backport remaining logic
  - [x] tags
    - [x] finish #3703
    - [x] finish #3771
    - [x] finish #3772
    - [x] finish #3788 
    - [x] finish #3789 
    - [x] should getResourceIDsForTag move into ResourceService, possibly as a protected method?
    - [x] refactor fixDuplicateTags logic from TagService into \VuFind\Tags\TagsService (it belongs at a higher level)
  - [x] user
    - [x] finish #3700
    - [x] finish #3702
    - [x] finish #3625
    - [x] finish #3774
    - [x] finish refactoring core code to use interfaces
    - [x] port code to use Doctrine
    - [x] remove legacy Laminas code
  - [x] user_card
    - [x] finish #3629
    - [x] finish #3679
    - [x] backport UserCardService::getAllRowsWithUsernames method and SwitchDbHashCommand changes
  - [x] user_list
    - [x] finish #3660
    - [x] finish #3701
    - [x] finish #3782
    - [x] finish backporting UserListService to dev
  - [x] user_resource
    - [x] finish #3743
    - [x] finish https://github.com/demiankatz/vufind/pull/28
    - [x] finish #3784
    - [x] investigate whether getOrderByClause() can/should be revised (having a static method seems suboptimal)
- [x] Replace low-level Laminas\Db functionality with Doctrine
  - [x] InstallController
  - [x] UpgradeController
- [x] Investigate creating an Entity plugin manager, and test overriding/extending entities
- [x] Replace or remove database logging functionality (see LoggerFactory::addDbWriters)
- [x] Test all platforms (MySQL / MariaDB / PostgreSQL / etc.)
  - [x] Be sure that code leveraging `createNativeQuery` works correctly across platforms (e.g. make sure tag deduplication works in PostgreSQL)
- [x] Make sure Doctrine proxy cache is being generated in appropriate place (i.e. reflecting VUFIND_LOCAL_DIR)
- [ ] Resolve [VUFIND-1385](https://openlibraryfoundation.atlassian.net/issues/VUFIND-1385) and [VUFIND-1480](https://vufind.org/jira/browse/VUFIND-1480) when merging.
- [x] Investigate supporting inheritance of annotations using [kenny1911/doctrine-inherit-annotations](https://packagist.org/packages/kenny1911/doctrine-inherit-annotations) or similar. [not worth it: annotation driver has been deprecated]
- [x] Convert from annotation driver to attribute driver on entity classes.
- [x] Figure out how to reconcile dependency conflict between php-cs-fixer and Doctrine module (see https://github.com/doctrine/DoctrineModule/issues/800)
- [x] Remove folder_id from search table entity after merging https://github.com/vufind-org/vufind/pull/2699
- [x] Clean up the way the case_sensitive property is set up in various service classes; this should be centralized in a cleaner way instead of by sharing the TagServiceFactory.
- [x] Adjust getEntityClass() calls to use interfaces instead of direct class names.
- [x] Investigate upgrading to Doctrine 3 [not currently feasible due to https://github.com/doctrine/DoctrineORMModule being a bit behind the times due to Laminas churn]
- [x] If appropriate, update code to use Doctrine's UUID generator (see #2999) [not appropriate -- this tool has been deprecated; see https://github.com/doctrine/orm/issues/7312]


demiankatz:
I spent some more time on proof-of-concept work today. I decided to try revising some of the code in the admin module's tag controller, since that does complex queries and seemed like a potentially interesting starting point. So far, results are promising -- not only is the DQL query language more readable than the Laminas\Db object-oriented query builder, but I also managed to find and fix a bug along the way.

Some notes on today's progress:

1.) We need to have a place to put database-related logic. In the previous code, this was in the row and table classes. In the new code, I propose a set of VuFind\Db\Service classes designed to deal with various types of entities. I've started work on a TagService here, and I've set up the necessary plugin mechanism for interacting with services. This may need more granularity, but we can decide that as we gain more experience working with Doctrine.

2.) For extensibility purposes, we also need a way to abstract entity classes so that they can be overridden. I have created a VuFind\Db\Entity\PluginManager for this purpose, and I defined a marker interface for entities so that the plugin manager can validate what it is creating. Note that overriding entities appears to have one major limitation: annotations are not inherited from parent classes, so any local custom entity needs to fully duplicate its parent's properties and annotations. Perhaps there is a way around this, but given how infrequently I expect people to need to extend/override entities, it may also be a limitation that we can live with.

There is still a massive amount of work that needs to be done here, but I'm encouraged by this start. I will continue to chip away as time permits, and I welcome input or offers of collaboration. Eventually, it may be best to "divide and conquer" on all the conversion work, since there is so much of it.

demiankatz:
Just a quick update on progress, because I believe this branch is currently in a somewhat broken state:

In order to incorporate Doctrine queries with the Laminas Paginator, I had to add the `doctrine/doctrine-orm-module` dependency to get an appropriate paginator adapter. In order to install this, I had to upgrade several other dependencies to newer versions, and the upgrade to the latest laminas-cache seems to be causing problems. I worked around this while developing by hacking out the cache logic that was failing, but obviously that needs a proper solution, which I will address in a separate Laminas upgrade PR.

The `doctrine/doctrine-orm-module` dependency also raises the question of whether we actually need `roave/psr-container-doctrine`, which may be less desirable if the official Doctrine piece meets our needs. I'll investigate that further when time permits.

Anyway, I've created TODO checkboxes to capture all these balls that are in the air, and I apologize for leaving this in a somewhat broken state, but I didn't want to lose my work in progress, and I'm nearly out of time for today. I'll pick this up again as soon as practical.

aleksip:
Hi @demiankatz, this change is very exciting because our Admin Interface has been using `doctrine/doctrine-orm-module` for a while and it also reads the VuFind database.

I wonder whether it would make sense to develop the new Doctrine based VuFind database layer as a separate module/package, in a similar way that has been discussed in #2272 and #2283 (which I hope to continue working on as soon as time allows)?

In any case it would be great to have this developed in a way that would allow for code reuse, even if only by copying the classes over!

demiankatz:
Thanks, @aleksip, I'm glad that you're happy with this approach. I'm open to organizing the code any way that is helpful; I think it's just a question of what works best. Perhaps one approach could be to create a VuFindDb module to contain the Entity classes, ConnectionFactory, etc. I'm not sure if the actual Service classes would fit there as well, or if they'd be better off in the main VuFind module due to their potential entanglement with other dependencies. I think I'd have to get deeper into the project to have a stronger opinion about that, but I'm open to input!

In any case, right now, my priority is finishing up all the outstanding work on VuFind 9, so this is more or less on hold, but I'm hoping this will be one of the priority changes for VuFind 10, and I expect I'll be spending a lot of time on it in 2023. I'm definitely happy to discuss/plan in the meantime so we have a clear path forward when it comes time to execute the work. I'm also very open to brainstorming strategies for collaboration if you or anyone on your team might be able to help with the updates, as this is a very big project. I'm hoping that, in addition to updating everything, we can make strides toward introducing better test coverage for core code, since if we do this right, it should make things easier to test than the old database library did.

aleksip:
@demiankatz Sounds good! I am currently experimenting with potential `vufind-db`, `vufind-view` and `vufind-config` packages in our Admin Interface, will report on how it goes. 🙂

demiankatz:
Thanks, @aleksip, that's great! Certainly if you have classes from your admin code that it would make sense to move over here for use in the core, that might save us some time and trouble. I'm not sure if the existing auto-generated Entity classes are actually the best they could be, as most of them have not yet been tested.

module/VuFind/config/module.config.php
```diff
@@ -378,7 +379,10 @@
             'VuFind\Crypt\HMAC' => 'VuFind\Crypt\HMACFactory',
             'VuFind\Date\Converter' => 'VuFind\Service\DateConverterFactory',
             'VuFind\Db\AdapterFactory' => 'VuFind\Service\ServiceWithConfigIniFactory',
+            'VuFind\Db\Connection' => 'VuFind\Db\ConnectionFactory',
```
demiankatz, line 468: @aleksip, if there's not an actual VuFind\Db\Connection class, shouldn't we use the real class name here?

module/VuFind/config/module.config.php
```diff
@@ -338,6 +338,7 @@
     'service_manager' => [
         'allow_override' => true,
         'factories' => [
+            'Doctrine\ORM\Mapping\Driver\AnnotationDriver' => 'VuFind\Db\AnnotationDriverFactory',
```
demiankatz: @aleksip, I see that you deleted this factory, but the configuration is still here. We should clean something up.

demiankatz (commented):
@aleksip, thanks for the progress on configuration here -- just leaving a couple of comments/questions related to the latest changes.

aleksip:
@demiankatz I have accidentally pushed my changes to your branch, sorry about that! I have been refactoring our Admin Interface database implementation to be based on this PR, and those were the changes I needed to make. They were needed to support two entity managers with connections to different databases.

I don't claim to fully understand how Doctrine's configuration works, but the `orm_default` naming convention seems a bit problematic in integrations. It would seem better to define an application specific name, and if necessary create an `orm_default` alias for that.

I also wondered about the need for `AnnotationDriverFactory`, as it seems that the same result can be achieved with just configuration.

module/VuFind/config/module.config.php
aleksip: If the real class name is used, other connections will use the factory too. We could always create an empty `VuFind\Db\Connection` marker class inheriting `Doctrine\DBAL\Connection`?

module/VuFind/config/module.config.php
aleksip: Yes, forgot to remove that line. Should I push the change to your branch or what would be the best way to work on this together?

module/VuFind/config/module.config.php
demiankatz: The general pattern is that the name of the service should match the resource created for the service. Can we use a base class or interface for Doctrine drivers, or are you saying that doing so would override core Doctrine configuration and cause problems?

module/VuFind/config/module.config.php
demiankatz: Please feel free to push directly here, as long as the code still works. I'm not going to be working on this much until next year, so I welcome any progress you can make in the meantime, and I don't anticipate that we'll "step on each other," so to speak.

demiankatz:
@aleksip, I am certainly not an expert on Doctrine's configuration either, and I've been learning as I go... so if the annotation driver can be replaced by pure configuration, I'm entirely in favor of that, and if a different naming convention would be more robust, I also have no objection to changes.

module/VuFind/config/module.config.php
demiankatz: Also, if at any point it would be helpful to get a second opinion or anything, or if you would like me to run any tests, just let me know and I'll be happy to help!

aleksip: @demiankatz I meant to push the changes to my repository and then point you there. Not sure how multi-user PRs should work... Should I push future suggested changes to my clone so you can then decide whether to merge to your branch from there?

demiankatz: @aleksip, it is possible to make a pull request against a pull request -- you can just target a PR against this branch. That might be the easiest workflow going forward, especially when multi-user collaboration picks up... but since things are quiet right now, feel free to work directly in this branch, at least until you get things into a correct/stable state again. (I think it's easier to just move forward and fix things than to roll back changes and make a separate PR, at least under the present circumstances).

aleksip: @demiankatz Seems like I can create a PR for the branch in your repository! Should I do that? Edit: missed your above reply.

demiankatz: If you want to try it as a proof-of-concept, feel free... but otherwise, as I said, you can just go ahead and push. :-)

module/VuFind/config/module.config.php
aleksip: Yes, using `'Doctrine\DBAL\Connection' => 'VuFind\Db\ConnectionFactory'` causes the VuFind specific factory to be used for all Doctrine configuration based connections created by `AbstractDoctrineServiceFactory`. Other connections can be specifically configured to avoid this, but since VuFind has a connection created by a custom factory class maybe it should have a custom connection class as well.

demiankatz: True, that does make sense.

aleksip: Will have to stop now for today but will push the new connection class tomorrow!

demiankatz:
I spent some time today working on PostgreSQL compatibility. The test suite is now in parity with the dev branch when running against PostgreSQL (there is one known bug causing a single test to fail in both places).

module/VuFind/src/VuFind/Db/Service/UserListService.php
```diff
@@ -205,12 +245,34 @@ public function getUserListsByTagAndId(
...
+    }
+
+    /**
+     * Retrieve a batch of list objects corresponding to the provided IDs
+     *
+     * @param int[] $ids List ids.
+     *
+     * @return array
+     */
+    protected function getUserListsById(array $ids): array
```
demiankatz, line 248: Right now this is a support method, but we might want to consider promoting it to a public method and adding it to the interface. We might also be able to eliminate the need for it by refactoring the query in getUserListsByTagAndId, so I'm leaving it protected for the moment to keep the interface as small as possible.

config/vufind/config.ini
```diff
@@ -830,6 +830,7 @@ database          = mysql://root@localhost/vufind
 ;database_name = "vufind"
 
 ; Should SSL be enabled on connections? (Currently only supported for MySQL).
+; Also requires apppropriate extra_options settings below.
```
ThoWagen: Typo

```suggestion
; Also requires appropriate extra_options settings below.
```

config/cli-config.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+$app = include __DIR__ . '/application.php';
+return \Doctrine\ORM\Tools\Console\ConsoleRunner::createHelperSet(
```
ThoWagen, line 4: Are you aware that the `createHelperSet` method is marked as deprecated in 2.10.2.

module/VuFind/src/VuFind/Auth/Database.php
```diff
@@ -184,7 +184,7 @@ public function create($request)
         $user = $this->createUserFromParams($params, $userService);
         try {
             $userService->persistEntity($user);
-        } catch (\Laminas\Db\Adapter\Exception\RuntimeException $e) {
+        } catch (\Doctrine\DBAL\Exception\UniqueConstraintViolationException $e) {
```
ThoWagen: Should we create own db exceptions to reduce the framework specific dependencies?

ThoWagen (changes requested):
I took a superficial look at it and only have some small comments. Everything seems to work in my environment and the tests are also passing.

I will look into it again when I find the time.

Thanks for all the work!

config/cli-config.php
demiankatz: This configuration file is needed to make the `vendor/bin/doctrine` command run. That command currently outputs a warning indicating that it shouldn't be used, and it's better to set up a [project-specific bin/doctrine command](https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/tools.html#doctrine-console).

I discovered that the configuration was not actually correct, and I have pushed up a fix to make to work -- but that does not address the deprecation.

I think we have two options:

1.) Leave this as-is for now, and worry about it when we are ready/able to upgrade to Doctrine 3.

2.) Remove this configuration file and create a VuFind-specific bin/doctrine command -- but if we go this route, we'll have to decide where to put it and how it should interact with our existing framework of command line tools.

What do you think?

module/VuFind/src/VuFind/Auth/Database.php
demiankatz: Good idea. Is 65239612aa04b23a87ff9b5f00687508339be4db more or less what you had in mind?

config/cli-config.php
ThoWagen: Since this already a huge PR I think leaving it as-is for now is ok.

module/VuFind/src/VuFind/Auth/Database.php
ThoWagen: Yes, that looks good to me!

module/VuFind/src/VuFind/Db/Entity/Resource.php
```diff
@@ -0,0 +1,233 @@
...
+     */
+    #[ORM\Column(name: 'extra_metadata', type: 'text', length: 16777215, nullable: true)]
+    protected $extraMetadata;
+
+    /**
+     * Id getter
+     *
+     * @return int
+     */
+    public function getId(): int
```
EreMaijala: Should this be `?int`?

EreMaijala (changes requested):
A couple of notes that apparently don't affect functionality but I'd like to get sorted:

- Should PHPDoc type for the `id` column be `?int`? Now it seems to be `int` for all I checked, while the getId() method often (but not always) has return type `?int`.
- Since the `JoinColumn` annotation has nullable true by default (the opposite for `Column` where nullable is false by default), perhaps it would be a good idea to explicitly state it. Also, the PHPDoc type annotations for `JoinColumn`s need to be checked.

module/VuFind/src/VuFind/Db/Entity/ExchangeArrayInterface.php
```diff
@@ -21,39 +21,38 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  *
  * @category VuFind
- * @package  Db_Table
+ * @package  Db_Interface
```
EreMaijala:
```suggestion
 * @package  Database
```

module/VuFind/src/VuFind/Db/Entity/ExchangeArrayInterface.php
```diff
@@ -21,39 +21,38 @@
...
-namespace VuFind\Db\Table;
+namespace VuFind\Db\Entity;
 
 /**
- * Marker interface for classes that depend on the \VuFind\Db\Table\PluginManager
+ * Interface for translating an entity to and from array format.
  *
  * @category VuFind
- * @package  Db_Table
+ * @package  Db_Interface
```
EreMaijala:
```suggestion
 * @package  Database
```

EreMaijala (changes requested):

module/VuFind/src/VuFind/Db/Entity/AccessToken.php
```diff
@@ -0,0 +1,250 @@
...
+    #[ORM\GeneratedValue(strategy: 'NONE')]
+    protected $type;
+
+    /**
+     * User.
+     *
+     * @var UserEntityInterface
+     */
+    #[ORM\JoinColumn(name: 'user_id', referencedColumnName: 'id')]
+    #[ORM\ManyToOne(targetEntity: \VuFind\Db\Entity\User::class)]
```
EreMaijala: Should the `targetEntity` attribute refer to the interface instead of the actual class (here and elsewhere)?

demiankatz: I'm not 100% sure, but I suspect it needs to refer to the actual Doctrine class for things to work correctly... could be wrong, though!

demiankatz: (Let me know if I should do deeper research into this -- I'm just working off of autogenerated code and assumptions at the moment).

module/VuFind/src/VuFind/Db/Entity/Resource.php
demiankatz: The property is `int` and the field is not nullable. However, since an uninitialized object could still have a null value, maybe the type should change -- but if so, the property type would also need to change, and we would want to check all entities for consistency, since this issue likely applies to every entity with an autoincrementing ID.

module/VuFind/src/VuFind/Db/Entity/AccessToken.php
EreMaijala: If that's the case, then let's just leave it there. I've no idea how Doctrine instantiates the class, that's why it was just a question.

module/VuFind/src/VuFind/Db/Entity/Resource.php
EreMaijala: Yes, that's why I mentioned it in the comment above. This doesn't seem to affect the currently used functionality, but we should be consistent. At the moment some of the `getId()` methods return `int` and some return `?int`.

**If** someone would call Resource's getId() for a non-saved resource, the result would currently be
`TypeError: VuFind\Db\Entity\Resource::getId(): Return value must be of type int, null returned` (I verified this).

demiankatz: Seems like the best course of action may be to adjust all the getId() types (and associated property types) to `?int` and make sure the comments include the note that a null value indicates an uninitialized object.

I guess an alternative would be to keep the type as `int` but throw an exception when uninitialized. I guess it comes down to whether we want to guard against null IDs in lots of code, or use try..catch blocks when appropriate.

Any thoughts/preferences?

module/VuFind/src/VuFind/Db/Service/ShortlinksService.php
```diff
@@ -30,10 +30,10 @@
 
 namespace VuFind\Db\Service;
 
+use DateTime;
 use Exception;
+use VuFind\Db\Entity\Shortlinks;
```
aleksip: This does not seem to be used.

module/VuFind/src/VuFind/Db/Service/ShortlinksService.php
```diff
@@ -104,10 +102,12 @@ public function createEntity(): ShortlinksEntityInterface
...
     public function createAndPersistEntityForPath(string $path): ShortlinksEntityInterface
     {
-        $table = $this->getDbTable('shortlinks');
-        $table->insert(['path' => $path]);
-        $id = $table->getLastInsertValue();
-        return $table->select(['id' => $id])->current();
+        $shortlink = $this->createEntity()
+            ->setPath($path)
+            ->setCreated(new DateTime());
+        $this->entityManager->persist($shortlink);
```
aleksip: This seems to be the only place (apart from a few tests) that calls `entityManager->persist()` directly instead of calling `AbstractDbService::persistEntity()`. If there is a reason for that a comment would be helpful.

aleksip (commented):
@demiankatz I am currently testing this against our admin interface code. Looks good so far but most likely I am not able to continue before Monday.

aleksip:
@demiankatz Made a PR [here](https://github.com/demiankatz/vufind/pull/43). Looks like I am not able to make a review request so just letting you know here!

module/VuFind/src/VuFind/Db/Service/ShortlinksService.php
demiankatz: Agreed -- the automatic tools for cleaning up imports are conservative about phrases that appear in comments, so sometimes things don't get removed when they should. Fixed in 448563070aaa952e0f45ba2308d9dec69e63bc84.

module/VuFind/src/VuFind/Db/Service/ShortlinksService.php
demiankatz: No reason, just inconsistent code from multiple developers. I've fixed it in 448563070aaa952e0f45ba2308d9dec69e63bc84, and all tests still pass. Thanks!

module/VuFind/src/VuFind/Db/Entity/Resource.php
ThoWagen: I don't have strong feelings here, but would prefer to keep `int` and throw an exception when uninitialized. With that we always make sure that we catch calls of `getId` on uninitialized objects directly and don't get harder to debug errors.

module/VuFind/config/module.config.php
```diff
@@ -636,6 +634,44 @@
             'VuFind\Http\CachingDownloader' => false,
         ],
     ],
+    'caches' => [
+        'doctrinemodule.cache.filesystem' => [
+            'options' => [
+                'cache_dir' => LOCAL_CACHE_DIR . (PHP_SAPI == 'cli' ? '/cli' : '') . '/objects',
```
ThoWagen, line 647: Could we just append `/cli` to LOCAL_CACHE_DIR once after it was declared in `constants.config.php` instead?

module/VuFind/src/VuFind/Db/Feature/DateTimeTrait.php
```diff
@@ -0,0 +1,82 @@
...
+trait DateTimeTrait
+{
+    /**
+     * Get a date or null from non-nullable date with a default value.
+     *
+     * @param DateTime $date Date
+     *
+     * @return ?DateTime
+     */
+    protected function getNullableDateTimeFromNonNullable(DateTime $date): ?DateTime
```
ThoWagen, line 53: Are this and `getNonNullableDateTimeFromNullable` used anywhere? I did not find any reference.

aleksip:
@demiankatz I have a new PR [here](https://github.com/demiankatz/vufind/pull/44). It changes concrete class references to interface references and adds types to database entity classes. It is possible to use interfaces in Doctrine attributes by adding a [ResolveTargetEntityListener](https://www.doctrine-project.org/projects/doctrine-orm/en/2.20/cookbook/resolve-target-entity-listener.html) in a custom `EntityManagerFactory`.

Unfortunately it seems that `AbstractDbService::getEntityClass()` is still needed in DQL queries although the Doctrine class metadata factory has the correct event manager with the `ResolveTargetEntityListener` in place. This seems odd (even a Doctrine bug perhaps) but at least the interface to concrete class mapping is restricted to that method.

The PR also includes some small changes to the code to fix minor issues uncovered by the added strict typing.

EDIT: @EreMaijala discovered the reason why the `ResolveTargetEntityListener` was not called, and an update to the PR is in the works!

module/VuFind/src/VuFind/Db/Entity/Resource.php
demiankatz: I'm inclined to agree that @ThoWagen's proposal seems simpler and more straightforward. Interested in what @aleksip and @EreMaijala think, too!

EreMaijala: That'd make sense if calling `getId()` on a new object would be illegal, but why would it be, and would it always be?

module/VuFind/src/VuFind/Db/Entity/AccessToken.php
EreMaijala: As @aleksip has been working on updating these, I'll mark this resolved. :)

module/VuFind/src/VuFind/Db/Entity/Resource.php
demiankatz: In any situation where the ID is an auto-incrementing database field, I think the ID should be considered to be in an undefined state until the object is persisted, and it really doesn't make any sense to try to get it first. If we want a test to find out whether an object has been persisted, adding a separate method for that might make more sense than checking for a non-null ID value.

I don't think there's a perfect solution here, but if we use a return type of `?int` then any code that expects an integer ID (which I would argue is MOST code using this method) will need to add checks or else static analysis will get upset. This seems like a lot of unnecessary extra complexity, whereas the exception approach allows us to use a try..catch if we really need to test something, and lets us just make a safe assumption the rest of the time.

module/VuFind/tests/unit-tests/src/VuFindTest/OAuth2/Repository/AbstractTokenRepositoryTestCase.php
```diff
@@ -30,18 +30,20 @@
 namespace VuFindTest\OAuth2\Repository;
 
 use PHPUnit\Framework\MockObject\MockObject;
-use VuFind\Db\Row\AccessToken as AccessTokenRow;
-use VuFind\Db\Row\User as UserRow;
+use VuFind\Db\Entity\AccessToken;
+use VuFind\Db\Entity\AccessTokenEntityInterface;
+use VuFind\Db\Entity\User;
```
ThoWagen: `use VuFind\Db\Entity\User;` is not used.

```suggestion
```

module/VuFind/tests/unit-tests/src/VuFindTest/Db/Service/SessionServiceTest.php
```diff
@@ -0,0 +1,348 @@
...
+ * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
+ */
+class SessionServiceTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * Mock entity plugin manager.
+     *
+     * @param bool $setExpectation Flag to set the method expectations.
+     *
+     * @return MockObject
```
ThoWagen: `MockObject` has not been imported.

Also would it make sense to add type declarations to all of the new and maybe even old method signatures?

module/VuFind/tests/unit-tests/src/VuFindTest/Db/Service/OaiResumptionServiceTest.php
```diff
@@ -22,19 +22,20 @@
...
+namespace VuFindTest\Db\Service;
 
 use Exception;
 use Generator;
-use Laminas\Db\ResultSet\AbstractResultSet;
+use VuFind\Db\Entity\OaiResumption;
 use VuFind\Db\Entity\OaiResumptionEntityInterface;
+use VuFind\Db\PersistenceManager;
 use VuFind\Db\Service\OaiResumptionService;
-use VuFindTest\Container\MockContainer;
```
ThoWagen, line 37: Also `MockObject` is used in this file. Is there a reason it gets removed here?

module/VuFind/tests/unit-tests/src/VuFindTest/Db/Service/OaiResumptionServiceTest.php
```diff
@@ -44,27 +45,158 @@
...
+
     /**
-     * Mock container
+     * Test retrieving a row from the database based on primary key.
      *
-     * @var MockContainer
+     * @return void
      */
-    protected MockContainer $container;
+    public function testfindToken(): void
```
ThoWagen:
```suggestion
    public function testFindToken(): void
```

module/VuFind/tests/unit-tests/src/VuFindTest/Db/Service/FeedbackServiceTest.php
```diff
@@ -0,0 +1,161 @@
...
+    {
+        $mocks = $this->getConfiguredFeedbackService();
+        $entityManager = $mocks['entityManager'];
+        $feedbackService = $mocks['feedbackService'];
+        $queryStmt = "SELECT f.id, f.status FROM VuFind\Db\Entity\Feedback f "
+            . 'ORDER BY f.status';
+        $query = $this->getMockBuilder(\Doctrine\ORM\AbstractQuery::class)
+            ->disableOriginalConstructor()
+            ->onlyMethods(['getResult'])
+            ->getMockForAbstractClass();
```
ThoWagen, line 82: `getMockForAbstractClass` is deprecated. But since we also use it in other cases it probably can be addressed in a separate PR.

module/VuFind/src/VuFind/Db/Entity/Resource.php
EreMaijala: In the end I'm fine either way, just want it to be implemented consistently.

module/VuFind/tests/integration-tests/src/VuFindTest/Db/Service/ChangeTrackerServiceTest.php
```diff
@@ -69,62 +69,71 @@ public function setUp(): void
...
+        $row = $tracker->getChangeTrackerEntity($core, 'test1');
         $this->assertIsObject($row);
-        $this->assertEmpty($row->deleted);
-        $this->assertEquals($row->first_indexed, $row->last_indexed);
-        $this->assertEquals($row->last_record_change, '2012-01-17 20:46:10');
+        $this->assertEmpty($row->getDeleted());
+        $this->assertEquals($row->getFirstIndexed(), $row->getLastIndexed());
+        $this->assertEquals(
+            $row->getLastRecordChange(),
+            \DateTime::createFromFormat('Y-m-d H:i:s', '2012-01-17 20:46:10')
```
ThoWagen: I wonder if it would be beneficial to add a helper class that contains the format 'Y-m-d H:i:s' and some helper functions (e.g. a wrapper for `createFromFormat`) in a followup PR.

module/VuFind/src/VuFind/Form/Handler/Database.php
```diff
@@ -82,7 +86,7 @@ public function handle(
         unset($formData['message']);
         $now = new \DateTime();
         $data = $this->feedbackService->createEntity()
-            ->setUser($user)
+            ->setUser($user ? $this->userService->getDoctrineReference(User::class, $user) : null)
```
ThoWagen: Is this not introducing Doctrine specific code and should it not be hidden somewhere behind interfaces? E.g. should there be a method in the Feedback service that uses `getDoctrineReference` to set the user for an entity?

module/VuFind/src/VuFind/Favorites/FavoritesService.php
```diff
@@ -114,7 +115,8 @@ public function createListForUser(?UserEntityInterface $user): UserListEntityInt
 
         return $this->userListService->createEntity()
             ->setCreated(new DateTime())
-            ->setUser($user);
+            // Stopgap until we've fully converted to Doctrine:
+            ->setUser($this->userListService->getDoctrineReference(User::class, $user));
```
ThoWagen: Same as above.

module/VuFind/src/VuFindTest/Feature/LiveDatabaseTrait.php
```diff
@@ -69,39 +67,141 @@ trait LiveDatabaseTrait
...
+        $options['cache_dir']
+            = LOCAL_CACHE_DIR . '/' . $options['cache_dir'] . '_testmode';
+        if (!is_dir($options['cache_dir'])) {
+            mkdir($options['cache_dir'], 0o777, true);
+        }
+        $cacheAdapter = new \Laminas\Cache\Storage\Adapter\Filesystem($options);
+        $cacheAdapter->addPlugin(new \Laminas\Cache\Storage\Plugin\Serializer());
+        $container->set(
+            'doctrine.cache.filesystem',
+            new \DoctrineModule\Cache\LaminasStorageCache($cacheAdapter)
```
ThoWagen, line 132: This is also deprecated. Just wanted to mention it but there probably is no need to change that right now as well.

ThoWagen (changes requested):
Here is another round of review.

module/VuFind/config/module.config.php
demiankatz: Some caches are adjusted for console mode and others are not; I would have to go through the code in more detail to see exactly how this is used, but I don't think we can just globally assign the predefined /cli suffix without breaking some things. We could, of course, just assign a variable at the top of module.config.php to avoid the repeated logic; would that be an improvement?

module/VuFind/src/VuFind/Db/Feature/DateTimeTrait.php
demiankatz: @EreMaijala added these because they are going to be used in his upcoming payment system PR.

module/VuFind/tests/unit-tests/src/VuFindTest/OAuth2/Repository/AbstractTokenRepositoryTestCase.php
demiankatz: Fixed.

module/VuFind/tests/unit-tests/src/VuFindTest/Db/Service/SessionServiceTest.php
demiankatz: Addressed by 4ef6fbff50d1400b8186739436a1155bd5b4442b

module/VuFind/tests/unit-tests/src/VuFindTest/Db/Service/OaiResumptionServiceTest.php
demiankatz: Addressed by ae329ea98b2cf4bdf0c93ca6eb12f260d1f9a0b0

module/VuFind/tests/unit-tests/src/VuFindTest/Db/Service/FeedbackServiceTest.php
demiankatz: Yes, unfortunately it looks like significant work to eliminate this.

module/VuFind/tests/integration-tests/src/VuFindTest/Db/Service/ChangeTrackerServiceTest.php
demiankatz: Might this be something that should be added to the https://github.com/vufind-org/vufinddate library?

module/VuFind/src/VuFind/Form/Handler/Database.php
demiankatz: I don't think we need this at all now that we are done transitioning from Laminas. We can probably go back to assigning `$user` directly. I don't have time to test right now, but I'll try it later if nobody else gets there first.

module/VuFind/src/VuFind/Favorites/FavoritesService.php
demiankatz: As noted in the comment, it's a stopgap that we should now be able to remove. I'll remove it soon and test if nobody else gets there first.

module/VuFind/src/VuFindTest/Feature/LiveDatabaseTrait.php
demiankatz: If we can just use the laminas-cache PSR-6 wrapper to solve the problem, we might as well, but again, I don't currently have time to try it and see if it works. Maybe later today!

module/VuFind/tests/integration-tests/src/VuFindTest/Db/Service/ChangeTrackerServiceTest.php
EreMaijala: When we drop PHP 8.1 support we can start adding consts to traits... just saying. But since PHP already provides consts for some date formats, how about just adding a const globally in constants.config.php (or is it not used in tests)?

module/VuFind/src/VuFind/Form/Handler/Database.php
demiankatz: Okay, addressed in e01c5603041db6873631aa06f34463774c649e2b.

module/VuFind/src/VuFind/Favorites/FavoritesService.php
demiankatz: Also addressed in e01c5603041db6873631aa06f34463774c649e2b.

module/VuFind/src/VuFindTest/Feature/LiveDatabaseTrait.php
demiankatz: At a glance, it looks like this behavior is derived from the Doctrine Laminas module, which I expect will have to be significantly rewritten or removed in order to facilitate upgrade to Doctrine 3. Probably best to leave this alone until that project.

module/VuFind/tests/integration-tests/src/VuFindTest/Db/Service/ChangeTrackerServiceTest.php
demiankatz: I guess the question is, what do we call it, regardless of how we implement it? Is this something like `VUFIND_DEFAULT_DATABASE_DATE_FORMAT`?

demiankatz: I'm also not sure that the constant is really helpful in cases like this where we're creating a date from a particular format. We could really use any arbitrary format here and it wouldn't make a difference as long as it resulted in the same DateTime object being created. I can see how a constant might be more useful for situations where we're using it as an output format rather than an input format... but in cases where we're loading hard-coded strings, having the format parallel with the string might actually increase readability, since all the meaning is in the same place this way.

EreMaijala: Good point. I wouldn't change these tests at all.

ThoWagen: > but in cases where we're loading hard-coded strings, having the format parallel with the string might actually increase readability, since all the meaning is in the same place this way.

Yes, good point. I also agree.

module/VuFind/config/module.config.php
ThoWagen: Besides this "LOCAL_CACHE_DIR" is only used in the Cache/Manager.php and AssetPipeline.php. I would not expect anything to break, only some duplicated caching and in other directories than before. 

But I guess unless we use the logic in even more places in the future it is ok like this or with a variable at the top of module.config.php.

config/cli-config.php
```diff
@@ -0,0 +1,6 @@
...
+    $app->getServiceManager()->get(\Doctrine\ORM\EntityManager::class)
```
aleksip: Shouldn't we get `'doctrine.entitymanager.orm_vufind'` here?

module/VuFind/config/module.config.php
demiankatz: I believe at least one thing would break: we have a "create hierarchies" command line tool to pre-populate the hierarchy cache. I think that's the reason that we sometimes differentiate between cli/non-cli, and other times do not. If we always forced the cli prefix, then the console tool would populate the wrong cache, and the performance improvement would not work.

module/VuFind/tests/integration-tests/src/VuFindTest/Db/Service/ChangeTrackerServiceTest.php
demiankatz: Okay, since we all agree about not changing the tests, do we still need to take action on use of this string in other places? If so, my earlier questions about naming, etc., would still benefit from some input.

EreMaijala: For a const I'd like `VUFIND_DATABASE_DATETIME_FORMAT`.

ThoWagen: I like the idea of adding it to constants.config.php and both of your name suggestions. Maybe a trait that adds some helper methods would still be use full. But that is not in conflict with this solution.

demiankatz: @ThoWagen, would you like to open a separate PR against dev to introduce this constant, and then we can merge it into #2233? Seems like it might be a good idea to approach this as a separate task so we don't add extra weight to this already-huge PR. Or, alternatively, we could do it as a follow-up step.

ThoWagen: I would prefer to do this as a follow-up.

ThoWagen: Here is a ticket https://openlibraryfoundation.atlassian.net/browse/VUFIND-1780

module/VuFind/config/module.config.php
ThoWagen: Ok, that makes sense. Then I don't see a reason to change anything right now.

module/VuFind/src/VuFind/Db/Entity/Resource.php
aleksip: I experimented with making all return types `int` only and found out that it would result in all kinds of interesting problems that can be avoided with `?int`. The property types can still stay `int` only as the contract is with the interface and the getters can `return $this->id ?? null`. So I'm voting for consistently returning `?int` IDs while keeping the actual properties `int` only.

aleksip:
I tried running `doctrine orm:validate-schema -v` and it detected 88 schema diffs. Most of these are constraints and indexes.  Since constraints include e.g. `on delete cascade`s this seems to mean that Doctrine will not stay in sync with the database state in some situations, which could result in unwanted consequences. We probably should try to get the schema validation to pass before version 11 is released and also add a new test that validates the schema?

config/cli-config.php
demiankatz: Yes, I have fixed it. I don't think I've actually used this since auto-generating the entity classes four years ago, and clearly things have evolved in the interim. As discussed elsewhere, this whole approach is deprecated now and will need to be revisited... but in the interest of getting the job done, I'm hoping to wait until after the initial merge to address it.

demiankatz:
> I tried running `doctrine orm:validate-schema -v` and it detected 88 schema diffs. Most of these are constraints and indexes. Since constraints include e.g. `on delete cascade`s this seems to mean that Doctrine will not stay in sync with the database state in some situations, which could result in unwanted consequences. We probably should try to get the schema validation to pass before version 11 is released and also add a new test that validates the schema?

Yes, that is a good idea. I imagine having a test to detect schema anomalies will require improving the integration with the Doctrine tools to use a non-deprecated method. Maybe those jobs could be done together. In the meantime, do you have any time to work on adding missing index annotations, or do you need me to figure that out on my end? (My first order of business is to finish testing and reviewing your latest PR -- hopefully I can get that merged today if I don't find further issues).

aleksip:
> Yes, that is a good idea. I imagine having a test to detect schema anomalies will require improving the integration with the Doctrine tools to use a non-deprecated method. Maybe those jobs could be done together. In the meantime, do you have any time to work on adding missing index annotations, or do you need me to figure that out on my end?

It might be possible to write a regular test and use the [SchemaValidator class](https://www.doctrine-project.org/projects/doctrine-orm/en/2.20/reference/tools.html#runtime-vs-development-mapping-validation). I'm not sure if I will have time to work on this in the near future, unfortunately.

demiankatz:
> It might be possible to write a regular test and use the [SchemaValidator class](https://www.doctrine-project.org/projects/doctrine-orm/en/2.20/reference/tools.html#runtime-vs-development-mapping-validation). I'm not sure if I will have time to work on this in the near future, unfortunately.

Thanks for that pointer -- when time permits, I'll see if I can do something with that! If anyone else has time in the meantime, please just let me know so we can coordinate our work.

module/VuFind/src/VuFind/Db/Service/UserListService.php
```diff
@@ -106,21 +109,24 @@ public function getUserListById(int $id): UserListEntityInterface
...
+        $dql = 'SELECT ul FROM ' . UserListEntityInterface::class . ' ul ';
+
+        $parameters = [];
+        $where = ["ul.public = '1'"];
+        if (!empty($includeFilter)) {
+            $where[] = 'ul.id IN (:includeFilter)';
+            $parameters['includeFilter'] = $includeFilter;
+        }
+        if (!empty($excludeFilter)) {
+            $where[] = 'ul NOT IN (:excludeFilter)';
```
LuomaJuha, line 121: Might have noticed a small error here, but not totally sure. Here the list is being compared and in include filter its comparing the id.

module/VuFind/src/VuFind/Db/Service/UserResourceService.php
```diff
@@ -107,35 +142,27 @@ public function createOrUpdateLink(
...
                 ->setUserList($list);
         }
         // Update the notes:
         $result->setNotes($notes);
-        $this->persistEntity($result);
+        try {
+            $this->persistEntity($result);
+        } catch (\Exception $e) {
+            $this->logError('Could not save user resource: ' . $e->getMessage());
+            return false;
```
LuomaJuha: Also here is a false value returned, but in function, the return type is expecting : UserResourceEntityInterface

module/VuFind/src/VuFind/Db/Service/UserListService.php
demiankatz: Yes, this definitely merits deeper testing. I'll try to find time to take a closer look soon!

module/VuFind/src/VuFind/Db/Service/UserResourceService.php
demiankatz: Good catch -- we really don't need that exception handling at all. I have removed it in 9e9aba1a74fbeee77a07f64549341fe0d32b6534.

demiankatz:
> It might be possible to write a regular test and use the [SchemaValidator class](https://www.doctrine-project.org/projects/doctrine-orm/en/2.20/reference/tools.html#runtime-vs-development-mapping-validation). I'm not sure if I will have time to work on this in the near future, unfortunately.

It's definitely possible -- I've created a proof-of-concept [DoctrineSchemaValidationTest.php](https://github.com/demiankatz/vufind/blob/287a8aeb0e7a02d8bb440a9e2441d2fd7fdaa27e/module/VuFind/tests/integration-tests/src/VuFindTest/Db/Service/DoctrineSchemaValidationTest.php) in another branch. The problem is that I don't really understand how to fix the errors it is revealing. It seems that Doctrine wants to drop and recreate many of the indexes; I'm not sure why it doesn't recognize the existing ones as being valid and correct. I imagine some of this may be related to naming issues, but I don't think that's the whole story. I also strongly suspect there will be additional challenges getting this test to pass consistently in both MySQL  and PostgreSQL. Any ideas?

module/VuFind/src/VuFind/Db/Service/UserListService.php
demiankatz: I did some experimentation and found that the syntax is equivalent -- whether you use `ul` or `ul.id`, the query works correctly when you pass in an array of integers or an array of entity objects. But to make this more consistent, I've pushed a commit to use `ul` in both places so it's less confusing!

demiankatz:
Okay, thanks to @sambhavp96's excellent work, several schema inconsistencies have been sorted out, and the Doctrine schema validation test I developed is now passing. See https://github.com/vufind-org/vufind/pull/2233/commits/f8526d2227312f1bc2586aed09387daa1b171ad9 for details.

EreMaijala (approved):

ThoWagen (approved):

module/VuFind/src/VuFind/Db/Service/SearchService.php
```diff
@@ -118,7 +119,7 @@ public function destroySession(string $sessionId, UserEntityInterface|int|null $
      */
     public function getSearchById(int $id): ?SearchEntityInterface
     {
-        return $this->getDbTable('search')->select(['id' => $id])->current();
+        return $this->entityManager->find(SearchEntityInterface::class, $id);
```
stweil, line 122: @demiankatz, this line raises a runtime exception in my local test installation:

`Ausnahme: Doctrine\Persistence\Mapping\MappingException
Class 'VuFind\Db\Entity\SearchEntityInterface' does not exist`

Should it be `Search::class` or `\VuFind\Db\Entity\Search::class` here (and in the rest of the file)?

demiankatz: The SearchEntityInterface is defined in https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/Db/Entity/SearchEntityInterface.php, and [SearchEntityInterface is set up as an alias in the PluginManager](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/Db/Entity/PluginManager.php#L66), so I'm not sure why you are encountering a problem. I would expect the code to work as written (and indeed, if it didn't, I would expect some of our integration tests -- particularly [SavedSearchesTest](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SavedSearchesTest.php) -- to fail).

I'll try to find time to double-check this more carefully in my own test environment when I'm back in the office tomorrow. In the meantime, is there anything unusual about your environment? Can you specify what steps you're taking to raise the exception? Are you able to run integration tests?

stweil: I get more similar exceptions ("Class 'VuFind\Db\Entity\ResourceEntityInterface' does not exist" and others).

demiankatz: Does the whole system fail to work, or are you just seeing these types of errors occasionally? What PHP version are you using? Is there any possibility that your Composer dependencies are out of date or that there's something odd in your cache?

stweil: I don't think that macOS is a very common environment for VuFind tests, so yes, that's unusual. The integration tests (`vendor/bin/phing qa-php`) show lots of messages like `Unable to connect to localhost:80 (Connection refused)`, so it is also unusual that the Homebrew Apache listens on port 8080.

demiankatz: I believe that @EreMaijala has had success getting things working in the environment, so he might have some advice. Have you seen the notes on the [testing wiki page](https://vufind.org/wiki/development:testing:unit_tests)?

EreMaijala: That exception happens at runtime if you catch all exceptions in a debugger, but does not affect functionality. I suppose the mechanism is trying to look for a corresponding class but doesn't find it for an interface.

stweil: I catch "everything" (which includes all exceptions) in my debugger because it typically indicates errors or at least weaknesses in the code. Here an AI chatbot also says that looking for a class with the name of the interface is wrong, and the name of the class which is derived from the interface should be passed instead.

demiankatz: @stweil, it is conventional when using Laminas service managers to name services for interfaces, to make it easier to install local custom versions without having to alias concrete classes in potentially confusing ways. That is the pattern that has been followed here, and it is set up in our custom [EntityManagerFactory](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/Db/EntityManagerFactory.php). This may be a conflict between Laminas practices and Doctrine practices, but it was an intentional choice based on Laminas practice.

I don't really have strong feelings about this one way or the other; I think the likelihood of locally overriding entities is relatively small. However, it is a fairly large change to make this close to a major release, so I would be hesitant to change it further unless there's a very strong reason to do so. Maybe the thing worth investigating further is why exceptions are being thrown and caught when aliases are properly set up -- but I imagine that would require a deep dive into the Doctrine code.

PR #2234, opened 2021-12-08
HTML validation fixes

EreMaijala:
This is a set of changes that allows VuFind's bootstrap3 and sandal themes to produce valid HTML without warnings.

Theme config now includes doctype so that it can be set early before processing layout.phtml. This allows e.g. Laminas Form and HeadScript helper to do the right thing, and makes it possible to leave out the type attribute from script tags.


themes/root/templates/layout/help.phtml
```diff
@@ -1,4 +1,4 @@
-<?=$this->doctype('XHTML1_TRANSITIONAL')?>
+<?=$this->doctype()?>
 <html xmlns="http://www.w3.org/1999/xhtml" lang="<?=$this->layout()->userLang ?>" xml:lang="en">
```
demiankatz: Does it make sense to use the XHTML namespace without an XHTML doctype?

module/VuFindTheme/src/VuFindTheme/View/Helper/HeadScript.php
```diff
@@ -114,6 +114,9 @@ public function itemToString($item, $indent, $escapeStart, $escapeEnd)
         }
 
         $this->addNonce($item);
+        if ($this->view && $this->view->plugin('doctype')->isHtml5()) {
+            $item->type = '';
```
demiankatz: Just out of curiosity, what is the purpose of this change? Perhaps a comment would be worthwhile here.

demiankatz (commented):
A couple of questions; I'm also requesting a review from @crhallberg since he may be more familiar with some of this territory than I am.

themes/root/templates/layout/help.phtml
EreMaijala: While it didn't cause any trouble, I don't think we need it. Changed. While at it I noticed that there were a lot of bad anchors in search help pages and replaced those as well.

module/VuFindTheme/src/VuFindTheme/View/Helper/HeadScript.php
EreMaijala: It was a misguided attempt at getting rid of the `type="text/javascript"` attribute, which the Laminas helper actually takes care of already. Now removed. Thanks for pointing it out!

demiankatz:
Thanks, @EreMaijala. I've gone ahead and cherry-picked the help translation fixes to dev since they added a lot of noise to this PR and make perfect sense (I also applied them to the new Ukrainian translation that was introduced after you started this PR).

Everything else looks fine to me, but I'm going to leave this open until @crhallberg has a chance to comment, in case he has any other thoughts.

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -1,11 +1,12 @@
-<?=$this->doctype('HTML5')?>
+<?=$this->doctype()?>
+
 <html lang="<?=$this->layout()->userLang?>"<?php if ($this->layout()->rtl): ?> dir="rtl"<?php endif; ?>>
   <head>
     <?php $this->setupThemeResources(); ?>
-    <meta charset="utf-8">
```
demiankatz, line 5: Oh, one last question: what's the motivation in removing this?

EreMaijala: Validator complained that you shouldn't have it when you already have `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">`. And that's what headMeta adds.

demiankatz: Okay, that makes perfect sense!

crhallberg (approved):
All these changes make sense to me!

demiankatz:
Thanks, @crhallberg and @EreMaijala -- just merged this!

PR #2235, opened 2021-12-08
Handle collections in merge-marc utility.

demiankatz:
This adds support for `<collection>`s as well as single records in the merge-marc utility; useful when OAI-PMH harvesting MARC records in batches.

TODO
- [x] Thoroughly test (and ensure unit test coverage of new case)

### DIFF
--- a/module/VuFindConsole/src/VuFindConsole/Command/Harvest/MergeMarcCommand.php
+++ b/module/VuFindConsole/src/VuFindConsole/Command/Harvest/MergeMarcCommand.php
@@ -87,17 +87,13 @@ class MergeMarcCommand extends RelativeFileAwareCommand
      */
     protected function recordXmlToString(SimpleXMLElement $record): string
     {
-        // Add missing namespace declarations:
-        $xml = $record->asXml();
-        foreach ($record->getNamespaces() as $prefix => $uri) {
-            // Not an ideal way to check if the namespace is already defined, but
-            // SimpleXML doesn't seem to provide a more reliable option.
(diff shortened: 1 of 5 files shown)


demiankatz:
@EreMaijala, interested in your thoughts on this (though I don't want to take up too much of your time, since I don't think this is of much direct value to you). This utility for grouping together harvested MARCXML files was very limited, and I needed to add some smarts so that it could handle collections as well as individual records. Because of the oddities and quirks of SimpleXML, I was unable to do so without multiple ugly hacks. I'm sure there are better ways, and there are probably some edge cases that won't get handled correctly, but this was good enough for my needs and was the best I could do in the available time. Interested if you have any suggestions for improvement, though!

EreMaijala:
@demiankatz As far as I can see (and admittedly I didn't spend too much time on this), it would make things more clear if you define the namespaces and their prefixes at the document level and then just prefix the tags on collection and record level. That would allow you to remove the check for existing namespace declarations and make sure that the collection element belongs to the correct namespace.

demiankatz:
Thanks, @EreMaijala; I've made some revisions to try to normalize things into using a marc prefix (and found and fixed a bug along the way), and I think this is better... though still far from ideal, because SimpleXML, as the cliche goes, is actually not simple, and doesn't really allow straightforward namespace handling. I wish I could normalize namespaces and prefixes via SimpleXML itself rather than via hacky string replacement, but I don't think there's a great way to do that.

EreMaijala (approved):
Thanks, this now looks good enough to me. :)

demiankatz:
Thanks, @EreMaijala!

PR #2236, opened 2021-12-09
FOLIO: include course number in course name

demiankatz:
It was requested that we incorporate the course code into the course name for display purposes in our course reserves search; this PR adds a more flexible mechanism for retrieving course reserves data from the API, and uses it to create "code: name" strings to describe courses.

It might make sense to set this up as a configuration option in Folio.ini -- I'm interested in feedback.

TODO
- [x] Make configurable?


demiankatz:
I believe that this is now ready to merge -- just looking for a review from a FOLIO user. @bbusenius or @todolson, any chance one of you can take a look?

todolson:
From what I can see, this seems fine. However, we do not use VuFind for Course Reserves, and I'm not really a PHP coder. So I don't think I'm the right person to give feedback.

todolson (commented):
This is just to reiterate that while this looks like fine to me, I don't think I'm the right person to really give a constructive review.

That is, parameterizing the display of course numbers seems like the right approach. But I don't think I can speak to the particulars.

bbusenius:
Like @todolson said, we don't use this but it looks good to me. There's nothing in the PHP offensive to my sensibilities. I like to see the `trim`s done in the `getCourses` callback.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1284,7 +1290,17 @@ public function getInstructors()
...
-        return $this->getCourseResourceList('courses');
+        $showCodes = $this->config['CourseReserves']['displayCourseCodes'] ?? false;
+        $courses = $this->getCourseResourceList(
+            'courses',
+            null,
+            $showCodes ? ['courseNumber', 'name'] : ['name'],
+            $showCodes ? '%s: %s' : '%s'
+        );
+        $callback = function ($course) {
+            return trim(ltrim($course, ':'));
```
bbusenius, line 1301: This is nice. I feel like when we forget to do this, there can be problems.

bbusenius (approved):

demiankatz:
Thanks, @todolson and @bbusenius. I think this is probably safe to merge (since it doesn't change default behavior and just adds a new option). However, I'll leave it open another few days in case somebody with a stronger opinion has anything to say. :-)

module/VuFind/src/VuFind/ILS/Driver/Folio.php
demiankatz: The `ltrim` also takes care of the case where the course number field is blank. :-)

demiankatz:
There have been no objections, so I'm going ahead and merging this one.

PR #2237, opened 2021-12-09
Add email message as a view param to feedback response page.

EreMaijala:
Also simplifies response handling by eliminating the showResponse method.


demiankatz:
Looks good to me!

PR #2238, opened 2021-12-13
Avoid notice in EIT code when results are missing.

demiankatz:
While testing #2229, I encountered some notices when EIT was not fully configured. This PR eliminates the notices to avoid noise in logs; better overall validation might also be in order, but I didn't want to invest a lot of time in this rarely-used backend.


EreMaijala (approved):
Looks good to me!

PR #2239, opened 2021-12-13
Avoid notice in SearchMemory helper.

demiankatz:
I ran into a notice caused by the fact that `parse_url` can return null, and `\Laminas\Stdlib\Parameters::fromString()` requires a string. This PR fixes the problem and adds a test case to cover the situation.


EreMaijala (approved):

demiankatz:
Thanks, @EreMaijala!

PR #2240, opened 2021-12-14
XCNCIP2: Fix get bibs for bad responses

xmorave2:
Some responders return empty `NextItemToken` element.


demiankatz:
Makes sense to me! Thanks, @xmorave2.

PR #2241, opened 2021-12-14
Result counts for inactive tabs

hajoseng:
addresses jira ticket 1488 (https://vufind.org/jira/browse/VUFIND-1488)


module/VuFind/src/VuFind/AjaxHandler/GetResultCountFactory.php
```diff
@@ -0,0 +1,72 @@
...
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options passed to factory.');
+        }
+        return new $requestedName(
+            $container->get('VuFind\Search\Results\PluginManager')
```
demiankatz: Please use `\VuFind\Search\Results\PluginManager::class` here instead of the string; this makes static analysis work better.

themes/bootstrap3/js/resultcount.js
```diff
@@ -0,0 +1,24 @@
+/* global VuFind */
+
+VuFind.register('resultcount', function resultCount() {
+  function init() {
+    $('ul.nav-tabs li a').each(function queryResultCount(){
+      var $this = $(this);
+      if ($this.attr('href') !== undefined) {
+        var queryString = $this.attr('href').replace(/^.*\?/, '');
+        var source = $this.data('source');
+        jQuery.ajax({
```
demiankatz: Any particular reason for using `jQuery` instead of `$` here? I think we usually use `$`, so it might be more stylistically consistent.

module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @param Params $params Parameter helper from controller
+     *
+     * @return array [response data, HTTP status code]
+     */
+    public function handleRequest(Params $params)
+    {
+        $queryString = $params->fromQuery('querystring');
+        $queryString = urldecode(
+            str_replace(
```
demiankatz: Why is this str_replace necessary? Is the Javascript code sending things entity-encoded? If so, is it possible that you can fix that on the JS side using the `text()` method or something similar?

module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
```diff
@@ -0,0 +1,105 @@
...
+                '&amp;',
+                '&',
+                trim($queryString)
+            )
+        );
+
+        $queryArray = explode('&', $queryString);
```
demiankatz: Did you consider using PHP's built-in `parse_url` function? I imagine that might simplify some of this code, unless there's a specific reason why it won't work.

demiankatz (commented):
Thanks, @hajoseng. I haven't tried this yet, but it looks like a great start. I had a few suggestions for improving style and simplifying code (see below). I also wonder if/how we should make this configurable? As currently implemented, it can be fairly easily disabled by overriding the JS piece with an empty file... but I wonder if something more explicit might be preferable. What do you think?

themes/bootstrap3/js/resultcount.js
hajoseng: No, we can use $ here.

module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
hajoseng: This is a leftover from an earlier stage when I struggled with encoding issues. Going to fix it.

module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
hajoseng: parse_url has a quite weird behaviour in dealing with querystring keys ending with [] - at least when I use it. Either I haven't understood the function or it's behaviour in handling []-keys really weird. Do you have an idea?

module/VuFind/src/VuFind/AjaxHandler/GetResultCountFactory.php
hajoseng: ok

hajoseng:
I like the idea of making it configurable. Maybe it takes some days until I proceed with this pull request but I will

module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
demiankatz: I did some quick experimentation -- I was able to use `parse_url` to extract query strings from URLs, and then `parse_str` to convert the resulting query strings into PHP arrays. It seemed to handle brackets correctly.

demiankatz: It's a bit weird that `parse_str` sets the PHP array by reference instead of returning a value, but that was the biggest confusion I ran into.

demiankatz:
> I like the idea of making it configurable. Maybe it takes some days until I proceed with this pull request but I will

Thanks for the progress today! Let me know if I can be of any further assistance (brainstorming, etc.); otherwise, I'll wait to hear from you and can review again whenever you're ready.

hajoseng:
Now this feature is configurable. @demiankatz: Do you want to have a look at it and see if there's a need of some more changes?

config/vufind/config.ini
```diff
@@ -2110,6 +2110,10 @@ treeSearchLimit = 100
 ;EIT = access.EITModule
 ;Primo = access.PrimoModule
 
+
+[SearchTabsDetails]
```
demiankatz: Would SearchTabsSettings be more open-ended if we want to add more in the future? Interested in whether @EreMaijala has any thoughts/preferences. Also, before we merge this, we should be sure to add some comments describing the purpose of the new section and setting.

module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
```diff
@@ -0,0 +1,89 @@
+<?php
+/**
+ * Ajax Controller for Libraries Extension
```
demiankatz: I think it would probably make the most sense to change this to match the class comment below (`"Get Result Counts" AJAX Handler`).

module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Handle a request.
+     *
+     * @param Params $params Parameter helper from controller
+     *
+     * @return array [response data, HTTP status code]
+     */
+    public function handleRequest(Params $params)
+    {
+        $queryString = urldecode(
```
demiankatz: I'm still not clear on why this is being urldecoded here when it doesn't seem to be getting urlencoded on the other side.

module/VuFind/src/VuFind/Search/SearchTabsHelper.php
```diff
@@ -72,6 +72,13 @@ class SearchTabsHelper extends \Laminas\View\Helper\AbstractHelper
      */
     protected $permissionConfig;
 
+    /**
+     * Tab detailed configuration
+     *
+     * @var array
+     */
+    protected $detailConfig;
```
demiankatz: If we decide to change the name of the .ini setting, we might also want to review the naming of this property and related code. Maybe a simple `getOptions` or `getSettings` would be more concise than `getTabDetailConfig`. At very least, I don't think we need Tab in the method name since the method belongs to a helper with "Tabs" in the name.

themes/bootstrap3/templates/search/searchTabs.phtml
```diff
@@ -2,7 +2,7 @@
   <ul class="nav nav-tabs">
     <?php foreach ($searchTabs as $tab): ?>
         <?php if ($this->permission()->allowDisplay($tab['permission'])): ?>
-          <li<?=$tab['selected'] ? ' class="active"' : ''?>>
+          <li<?=$tab['selected'] ? ' class="active"' : ($showCounts ? ' class="show-counts"' : '')?>>
```
demiankatz: Would it be better to use a `data-show-counts` attribute instead of a class, since this is not really a style-related issue?

demiankatz (commented):
Thanks, @hajoseng, this is nearly done. I still have one encoding-related question, though, and some minor suggestions about naming and implementation details.

config/vufind/config.ini
EreMaijala: I agree that SearchTabsSettings would make sense. I'd also make the bool setting `... = true` instead of `... = 1`.

I think we're slowly approaching the point where moving search tab configuration to a separate config file would make sense, but this change won't mandate a switch.

hajoseng: Ok; I changed it to SearchTabsSettings.

module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
hajoseng: Oh sorry; a mistake due to my habit to paste and copy comments.

module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
hajoseng: I easily get stuck in encoding issues, but I can see that this isn't neccessary anymore. I removed it.

module/VuFind/src/VuFind/Search/SearchTabsHelper.php
hajoseng: I opted for getSettings.

themes/bootstrap3/templates/search/searchTabs.phtml
hajoseng: Yes.

hajoseng:
Thank you for your comments. There's another issue left: After an advanced search the result count for the other tab shows the result count of an empty search. This is a proper result because the inactive tabs are related to an empty search in this case but these numbers are meaningless, because changing the tab leads to another advanced search. I think in this case the result counts shouldn't be shown; what do you think?

demiankatz:
@hajoseng, yes, I agree that it probably doesn't make sense to show the result count under those circumstances. This inconsistency might be mildly confusing to users, but no numbers are probably less likely to mislead than unrepresentative numbers.

themes/bootstrap3/theme.config.php
```diff
@@ -45,6 +45,7 @@
         ['file' => 'lightbox.js', 'priority' => 320],
         ['file' => 'truncate.js', 'priority' => 330],
         ['file' => 'trigger_print.js', 'priority' => 340],
+        ['file' => 'resultcount.js', 'priority' => 350],
```
demiankatz: Now that showCounts is configurable, would it make sense to conditionally load this via searchTabs.phtml when the feature is enabled, rather than always loading the Javascript?

demiankatz (changes requested):
@hajoseng, I just tested this using a plain Solr tab, a filtered Solr tab, and an EDS tab. The EDS tab consistently displayed the count for Solr results, not EDS results. I suspect that this means that the "source" value is not getting set or passed correctly. Can you investigate, or would it be helpful for me to dig deeper? I don't have a lot of time today but can look into it further tomorrow if that would help.

Also see below for a minor, unrelated question.

hajoseng:
@demiankatz Thank you for raising this issue. I'm going to investigate deeper on monday and come back with either results or further questions.

themes/bootstrap3/theme.config.php
hajoseng: Sounds reasonable; I'm going to change this on monday.

hajoseng:
@demiankatz Indeed the source value was missing in the template; I added it. I don't have the opportunity to test it with EDS but my tests with Search2 passed.

demiankatz (approved):
Thanks, @hajoseng -- I added a missing `$this->escapeHtmlAttr()` call around the newly-added data value for source, and now with testing, I believe that all is well. Thanks again for taking the time to develop this feature; I'll merge as soon as CI passes!

PR #2242, opened 2021-12-15
Improve advanced search "clear" behavior for radio buttons/checkboxes

xrosecky:
We noticed that advanced search in EDS contains checkboxes that are not cleared when clear button is used.

### DIFF
--- a/themes/bootstrap3/js/advanced_search.js
+++ b/themes/bootstrap3/js/advanced_search.js
@@ -137,7 +137,11 @@ deleteGroup = function _deleteGroup(group) {
 $(document).ready(function advSearchReady() {
   $('.clear-btn').click(function clearBtnClick() {
     $('input[type="text"]').val('');
-    $('input[type="checkbox"]').prop("checked", false);
+    $('input[type="checkbox"]').each(function onEachCheckbox() {
+      var checked = $(this).data('reset-checked');
+      checked = (checked == null) ? false : checked;
+      $(this).prop("checked", checked);
+    });
     $("option:selected").prop("selected", false);
     $("#illustrated_-1").click();
   });
--- a/themes/bootstrap3/js/advanced_search.js
+++ b/themes/bootstrap3/js/advanced_search.js
@@ -137,12 +137,11 @@ deleteGroup = function _deleteGroup(group) {
 $(document).ready(function advSearchReady() {
   $('.clear-btn').click(function clearBtnClick() {
     $('input[type="text"]').val('');
-    $('input[type="checkbox"]').each(function onEachCheckbox() {
+    $('input[type="checkbox"],input[type="radio"]').each(function onEachCheckbox() {
       var checked = $(this).data('reset-checked');
       checked = (checked == null) ? false : checked;
       $(this).prop("checked", checked);
     });
     $("option:selected").prop("selected", false);
-    $("#illustrated_-1").click();
   });
 });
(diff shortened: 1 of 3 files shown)


demiankatz:
Thanks, @xrosecky! When time permits (either later this week or early next week) I'm going to see if I can write a Mink integration test to cover this case so that we can catch future regressions. I also wonder if the very specific code to click on a particular "illustrated" option can be made more generic (it would be better to reset all radio buttons rather than just this specific one, to support future growth/customization).

demiankatz:
...at a glance, it actually looks like repeating your newly-added line with a type of "radio" in place of the illustrated click logic might do exactly what we need. I'll try it when I have time, unless you think it's worth doing and get to it first.

themes/bootstrap3/js/advanced_search.js
```diff
@@ -137,6 +137,7 @@ deleteGroup = function _deleteGroup(group) {
 $(document).ready(function advSearchReady() {
   $('.clear-btn').click(function clearBtnClick() {
     $('input[type="text"]').val('');
+    $('input[type="checkbox"]').prop("checked", false);
```
EreMaijala: EDS can have checkboxes that are checked by default. I don't think it's appropriate to clear them. All checkboxes should rather be reset to the initial state.

EreMaijala (changes requested):

themes/bootstrap3/js/advanced_search.js
xrosecky: So should I store the initial state as data attribute and use it for reseting the checkbox to the initial state?

EreMaijala: Yes, I believe that's the clearest way to do it.

demiankatz: Thanks for adding this, @xrosecky -- so I think now the next step is adding the "reset-checked" attribute where it is needed. I'll look into this next week when I have a bit more time, unless somebody else gets there first. :-)

demiankatz: Okay, please take a look at ed5453e8b093d5a1a18348ee5bcacf577201acf6 -- this uses the new reset-checked attribute for both EDS checkboxes and Solr radio buttons. I think this is probably safe to merge now; does anyone have any objections or further suggestions?

EreMaijala (changes requested):
Looks good to me otherwise, but now that I look at it, it occurs to me that something like 'data-initially-checked' might be more intuitive, as 'reset-checked' might be misunderstood for "checked needs to be reset [to false]".

demiankatz:
Thanks, @EreMaijala. `data-checked-by-default` might also be a potential name for this. @xrosecky, do you have any preferences? I'm happy to make the change once we agree on the best name.

themes/bootstrap3/js/advanced_search.js
xrosecky: It is OK for me.

demiankatz: Thanks, @xrosecky. Do you have any opinions about the discussion below about renaming `reset-checked` to `initially-checked` or `checked-by-default` for clarity?

xrosecky: Data attribute `checked-by-default ` seems better to me.

demiankatz: Thanks, @xrosecky, I have made the change and will merge this as soon as CI passes.

demiankatz (approved):

PR #2243, opened 2021-12-16
Alma: Encode all url's properly.

EreMaijala:
At least patron id can include characters, such as a question mark, that would otherwise break url parsing.


EreMaijala:
Not sure what's up with CI, the error is unrelated.

demiankatz:
There was an accidental commit to dev which broke the build; I've reverted it and merged the fix here, so I think all is well! Merging now.

demiankatz:
I'd be interested to hear if @betullam has any feedback on this, but it seems uncontroversial enough to merge right away. :-)

betullam:
@demiankatz That change is totally fine for me ... it works perfectly with our patron IDs that could also contain special characters

PR #2247, opened 2021-12-21
[dev-9.0] Add support for overriding local dir in phing, use for composer qa

EreMaijala:
First commit makes it possible to override VuFind local dir for phing tasks. The second one uses it for `composer qa`. This allows one to run the QA tasks with customized local config that would otherwise cause trouble (particulary FeedbackForms.yaml and site name in config.ini are culprits).


EreMaijala:
I'm currently running ci-tasks to verify I didn't break them.

EreMaijala:
phing ci-tasks completed successfully.

build.xml
```diff
@@ -604,12 +605,12 @@
     <!-- create custom import configurations to load the MARC filename into the building facet to help
           test cases to limit searches to within specific collections of test records.
           Note: ${srcdir}/local/import/import.properties is created in install.php. -->
-    <copy file="${srcdir}/import/marc_local.properties" tofile="${srcdir}/local/import/marc-${BASENAME}.properties" overwrite="true" />
-    <append destFile="${srcdir}/local/import/marc-${BASENAME}.properties" text="building=&quot;${BASENAME}&quot;${line.separator}" />
+    <copy file="${srcdir}/import/marc_local.properties" tofile="${localdir}/import/marc-${BASENAME}.properties" overwrite="true" />
```
demiankatz, line 621: Is it also worth replacing the `${srcdir}/local` reference in the comment above this for consistency?

composer.json
```diff
@@ -122,6 +122,6 @@
         "phing-install-dependencies": ["phing patch-dependencies", "phing installsolr installswaggerui"],
         "post-install-cmd": "@phing-install-dependencies",
         "post-update-cmd": "@phing-install-dependencies",
-        "qa": "phing qa-console"
+        "qa": "phing qa-console -Dlocaldir=$(pwd)/qa_local"
```
demiankatz: A few questions:

- I'm having trouble finding documentation about `$(pwd)` -- is this reliable cross-platform, or if you run composer outside of the working directory (which is possible with the `--working-dir` switch)? I can test if you like; just curious if you've already done that investigation.
- Is this going to leave behind a qa_local directory with junk in it? Do we want to clean up after ourselves, or put this in the /tmp directory (if there's a platform-agnostic way to do that), or at least .gitignore it?

demiankatz (commented):
Thanks, @EreMaijala, this is a great idea. Just a few questions about the details....

composer.json
EreMaijala: Now I'm not sure. I just found it in some examples, but you're right, it might not work correctly for all cases. 

Yes, it's going to leave junk just like before (e.g. an extra ini file). Maybe using tmp and cleaning up would work. I'll see what I can do.

build.xml
EreMaijala: Yes, done.

composer.json
EreMaijala: Okay, looks like there's not much that can be done in composer.json, so I resorted to adding support for a new defaultconfigs=true property that sets localdir appropriately. I also fixed the test that left a config file behind, so now all that results is an empty directory under tmp.

demiankatz:
Thanks for the progress on this! I've cherry-picked your test fix back to the dev branch, since I think that improvement is needed immediately.

A couple more suggestions for you to consider before I merge this:

1.) It probably wouldn't hurt to put a comment in build.xml above the defaultconfigs check, explaning the motivation and purpose of that property, in case we forget in future.

2.) One disadvantage to this approach is that it always overrides the localdir property, so you can't set a custom localdir anymore. Might it be better if the defaultconfigs setting defined a "localdirparent" property, and then "localdir" was defined as "${localdirparent}/local" but with override="false"? That way, the "defaultconfigs" setting will work as intended, but if somebody wants to use a totally different localdir for some reason, we don't stand in the way of that. (Obviously this would mean that localdir and defaultconfigs couldn't both be set at the same time... but I don't think that's necessarily a problem).

What do you think?

EreMaijala:
@demiankatz Thanks!

1.) Comment added.

2.) I changed the override for normal localdir to "false". However, it seems that it doesn't affect properties set on the command line; they're never overridden. Phing docs are really vague, but it looks like the override attribute only affects properties defined in the build file, and anything from command line parameters always overrides `<Property>` in build.xml. This means that you can actually define both `defaultconfigs=true` and `localdir=something`, though it doesn't make sense.

demiankatz:
Thanks for the confirmation and finishing touches!

PR #2248, opened 2021-12-21
Add as-needed option to ESLint func-names rule

aleksip:
We have a large existing JavaScript codebase with which we would like to start using VuFind's coding standards. This small change would save us from a lot of refactoring and should not affect existing VuFind code.

https://eslint.org/docs/rules/func-names#as-needed

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,7 +8,7 @@ module.exports = {
   rules: {
     // errors
     "block-scoped-var": "error",
-    "func-names": "error",
+    "func-names": ["error", "as-needed"],
     "no-loop-func": "error",
     "no-param-reassign": "error",
     "no-shadow": "error",


demiankatz:
Thanks, @aleksip! This looks fine to me, but I'd also like to hear @crhallberg's opinion in case he has any concerns about using a non-default ESlint behavior (though it seems to me that, now that we're relying on ES6, this change could actually allow us to improve code readability in some places by eliminating redundant names).

aleksip:
I very well could be missing or misunderstanding something, but it seems to me that `func-names` is not in the default `eslint:recommended` rules, so this would be changing something that is already VuFind specific.

https://github.com/eslint/eslint/blob/main/conf/eslint-recommended.js

demiankatz:
@aleksip, I didn't mean to imply that `func-names` itself was turned on by default. However, the documentation for `func-names` says that if you do not provide a string option to the setting, it defaults to `always` rather than `as-needed`. That's the non-default option that I was referring to. As I said earlier, I think your suggestion makes sense and I expect we'll merge this; I just wanted to make sure @crhallberg agrees.

crhallberg (approved):
This seems to be in line with why I added this rule in the first place: easier debugging. I don’t see any reason why this should cause any issues.

demiankatz:
Thanks, @aleksip and @crhallberg -- I'm merging this now.

PR #2249, opened 2021-12-22
Fix notices in tag admin code.

demiankatz:
This PR fixes some notices discovered while working on PR #2233. There were some incorrect assumptions about operator precedence and types which are corrected here.

### DIFF
--- a/module/VuFindAdmin/src/VuFindAdmin/Controller/TagsController.php
+++ b/module/VuFindAdmin/src/VuFindAdmin/Controller/TagsController.php
@@ -198,18 +198,18 @@ class TagsController extends AbstractAdmin
         $user = $this->getTable('user')
             ->select(['id' => $this->getParam('user_id')])
             ->current();
-        $userMsg = (false !== $user)
+        $userMsg = is_object($user)
             ? $user->username . " (" . $user->id . ")" : "All";
 
         $tag = $this->getTable('tags')
             ->select(['id' => $this->getParam('tag_id')])
             ->current();
-        $tagMsg = (false !== $tag) ? $tag->tag . " (" . $tag->id . ")" : " All";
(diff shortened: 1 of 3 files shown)


module/VuFindAdmin/src/VuFindAdmin/Controller/TagsController.php
```diff
@@ -198,18 +198,18 @@ protected function getConfirmDeleteMessages($count)
         $user = $this->getTable('user')
```
EreMaijala, line 198: I'd like an explicit null check for getParam return value for each type. Currently, in the unlikely case that the object selected gets deleted before the form is submitted, the current code results in "All", which is not right. I think the code should first check getParam(), and if it's set, either fetch the data or set the message to '<unknown>' or something.

Later on, when adding the 'tag_delete_filter' message, I'd rather check for message strings other than 'All' rather than comparing them with false.

module/VuFindAdmin/src/VuFindAdmin/Controller/TagsController.php
```diff
@@ -198,18 +198,18 @@ protected function getConfirmDeleteMessages($count)
...
             ->select(['id' => $this->getParam('user_id')])
             ->current();
-        $userMsg = (false !== $user)
+        $userMsg = is_object($user)
             ? $user->username . " (" . $user->id . ")" : "All";
```
EreMaijala, line 202: "All" here doesn't have a leading space while the others do. Any reason for that?

EreMaijala (changes requested):
While the fixes seem to work, I'd like a couple of tweaks to the current code. Can I ask them here while we're at it? :)

demiankatz:
Thanks, @EreMaijala -- I have merged the fixes as-is and will open a separate PR to address your suggestions (since they are not directly related to the work here).

PR #2250, opened 2021-12-28
Add a simple interval captcha that prevents too frequent actions.

EreMaijala:
This adds a simple interval captcha that only allows protected actions every X seconds from previous action or session start. The most significant benefit of this captcha is that it does not have a user interface, so it avoids any accessibility issues. The downside is that it can be annoying in case there's a legitimate need to perform an action more frequently. We have not had any real issues with the interval set to 60 seconds, however.


module/VuFind/src/VuFind/Controller/Plugin/Captcha.php
```diff
@@ -128,11 +128,12 @@ public function verify(): bool
         }
 
         if (!$captchaPassed && $this->errorMode != 'none') {
+            $message = $captcha->getErrorMessage() ?: 'captcha_not_passed';
```
demiankatz: Would it make more sense to put the `captcha_not_passed` message in the abstract base class rather than using an empty string there and defaulting it here?

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/FeedbackTest.php
```diff
@@ -160,4 +160,43 @@ public function testFeedbackFormWithCaptcha()
...
+     * @return void
+     */
+    public function testIntervalCaptcha()
+    {
+        $page = $this->setupPage(
+            [
+                'Captcha' => [
+                    'types' => ['interval'],
+                    'forms' => 'feedback',
+                    'action_interval' => 8
```
demiankatz: How did you come upon this number? Just wondering if we can get away with less, since adding 8 seconds to the test suite isn't ideal.

config/vufind/config.ini
```diff
@@ -2154,6 +2156,11 @@ treeSearchLimit = 100
 ;image_dotNoiseLevel = 100
 ;image_lineNoiseLevel = 5
 
+; Interval options:
+; Minimum interval between actions, or time from session start (seconds, default is
+; 60):
+;action_interval = 60
```
demiankatz, line 2175: Might there be a need to have separate values for "time from start" vs. interval? e.g. what if we want to allow somebody to do something once per hour -- then we don't want to make them wait an hour to do it the first time.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/FeedbackTest.php
demiankatz: Perhaps a simple unit test would suffice... and in any case, a unit test would be a nice complement to this. (I don't mind writing one if that would help).

demiankatz (commented):
Thanks, @EreMaijala -- a few questions/comments.

module/VuFind/src/VuFind/Controller/Plugin/Captcha.php
EreMaijala: Fixed.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/FeedbackTest.php
EreMaijala: I updated the test to avoid the delay. What do you think? That doesn't exclude adding a unit test. I'll see what I can do.

config/vufind/config.ini
EreMaijala: Added!

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/FeedbackTest.php
EreMaijala: Unit tests added. Please check whether they align with your expectations!

demiankatz (approved):

demiankatz:
Looks good to me -- merging now!

PR #2252, opened 2021-12-30
Replace File_MARC with MarcReader and friends

EreMaijala:
This is a set of patches that removes the dependency on pear/file_marc and pear/validate_ispn packages.

TODO (at least):

- [x] Test Voyager
- [x] Test Unicorn
- [x] Test everything else
- [x] Test samples in MarcAdvancedTrait
- [x] Run CI tests
- [x] Check if the default record factory in module/VuFindSearch/src/VuFindSearch/Backend/WorldCat/Response/XML/RecordCollectionFactory.php is needed (result: no, but let it be for now)
- [x] Check if the include path mangling is still necessary in public/index.php and module/VuFind/tests/bootstrap.php (result: no, PEAR modules only used in current version of phing)
- [x] Change log / release notes


module/VuFind/src/VuFind/Marc/Serialization/MarcXml.php
```diff
@@ -48,29 +48,52 @@ class MarcXml implements SerializationInterface
...
+     *
+     * @throws \Exception
+     * @return array
+     */
+    public static function collectionFromString(string $collection): array
+    {
+        $xml = static::loadXML(trim($collection));
+        $results = [];
+        foreach ($xml->record as $record) {
+            $results[] = $record->saveXML();
```
demiankatz: Any strong reason to use `saveXML` here instead of `asXML`? I prefer `asXML` (though the two are synonymous). It looks like both forms are currently used in the VuFind codebase, but there are more `asXML`'s than `saveXML`'s. Perhaps I should open a PR to normalize them.

module/VuFind/src/VuFind/Marc/MarcCollection.php
```diff
@@ -0,0 +1,158 @@
...
+     * Iteration position
+     *
+     * @var int
+     */
+    protected $position = 0;
+
+    /**
+     * Constructor
+     *
+     * @param string $data MARC record collection in MARCXML or ISO2709 format
```
demiankatz, line 68: While I think this approach should work for most of VuFind's purposes, I wonder if an exclusively string-based solution is too limiting. For example, what if I want to pass a 4GB MARC file to the "deletes" script? As implemented, we would have to load the whole thing into memory, and it would likely fail. It might be nice to have support for streaming data instead (which I believe could be accomplished for both binary and XML MARC, though XML handling would probably require use of the [XMLReader class](https://www.php.net/manual/en/book.xmlreader.php)).

If I'm reading the code correctly, these problems already exist in File_MARC, so I don't think we urgently need to address them in order to move forward with this work... but I'm wondering if we should at least lay some groundwork now so we could add these as enhancements without having to break existing interfaces.

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcLintTest.php
```diff
@@ -60,5 +61,90 @@ public function testMarcLint()
         $marc = $this->getFixture('marc/marclint_bad2.xml');
         $reader = new \VuFind\Marc\MarcReader($marc);
         $this->assertEquals(38, count($lint->checkRecord($reader)));
+
+        // Test checkArticle with field 130 and an invalid field:
+        $lint = new \VuFind\Marc\MarcLint();
+        $this->CallMethod(
```
demiankatz: Is that capital C on CallMethod a typo? It's repeated a few times but seems odd.

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcLintTest.php
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * MarcLint Test Class
+ *
+ * @category VuFind
+ * @package  Tests
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
+ */
+class MarcLintTest extends \PHPUnit\Framework\TestCase
```
demiankatz, line 43: Not sure if you already looked at them, but there are a bunch of test classes (marc_lint_001.phpt through marc_lint_005.phpt) in the [original project](https://github.com/pear/File_MARC/tree/master/tests). Might it be worth porting them over? (I didn't closely review past or current tests, so maybe you've already done this analysis).

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat/Response/XML/RecordCollectionFactory.php
```diff
@@ -67,11 +66,7 @@ class RecordCollectionFactory implements RecordCollectionFactoryInterface
      */
     public function __construct($recordFactory = null, $collectionClass = null)
     {
-        if (null === $recordFactory) {
-            $recordFactory = function ($i) {
```
demiankatz, line 71: I'm not sure I understand why this needed to be removed. I think the intention here was to have a simple container class for fetching WorldCat records when a full record driver was not needed. It's probably unused, but I'm not sure there's a compelling reason to get rid of it. If we do want to get rid of it, it might also be better to actually delete the Record class rather than leaving it here unused... but I'd be inclined to keep things as they are just to minimize impact at this time (unless I'm overlooking a strong reason that the removal is needed).

demiankatz (commented):
I've just done a first review, focusing primarily on the commits that build up the Marc library, since I'm hoping that we can cherry-pick those first to get the ball rolling here. There are a few things that might be worth fixing or discussing further before taking that step, however... see below!

module/VuFind/src/VuFind/Marc/Serialization/MarcXml.php
EreMaijala: Not really (apart from the fact that DOMDocument only has `saveXML`, but there's no need for unity here). I'll change it.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat/Response/XML/RecordCollectionFactory.php
EreMaijala: It didn't need to be removed, but since it was unused and needed to be changed I thought it could just as well be dropped, but I'll drop the changes from here. In the long run we may want to revisit at least EDS, EIT and LibGuides as well.

demiankatz: Would it be worth deprecating the unused classes in 8.1 and removing them in 9.0, perhaps? I doubt anyone is going to care about them, but that at least gives some fair warning.

EreMaijala: That would make sense, though if there's really a use case for them, I have no problem keeping them around.

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcLintTest.php
EreMaijala: I did actually overlook them. I've now ported most of the tests over and expanded on that to bring the coverage to 100%. There are minor differences due to how things are handled in MarcReader, but the ported tests are now pretty nice as they're more granular than what I had added, and I also made the result checks more specific than the original ones.

module/VuFind/src/VuFind/Marc/MarcCollection.php
EreMaijala: I changed the MarcCollection class definition to account for the possibility of streaming load, most importantly removing the `\Countable` support. File_MARC can do streaming for ISO2709 records but not MARCXML.

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcLintTest.php
EreMaijala: Typo, fixed.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat/Response/XML/RecordCollectionFactory.php
demiankatz: Perhaps not a particularly compelling one, but they're probably not hurting anything for now. :-)

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcCollectionFileTest.php
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Return collection fixtures for the tests
+     *
+     * @return array
+     */
+    public function collectionFixtures(): array
+    {
+        return [
+            ['marc/marc_collection.xml'],
```
demiankatz, line 53: Might it be worth adding a third fixture containing non-namespaced XML? For example, if somebody harvests MARCXML using VuFind's OAI-PMH tool, it's possible they could end up with namespaced `<record>` elements contained within a non-namespaced `<collection>` container (if they use the default behavior for grouping batches of harvested records). Might be good to ensure that this works appropriately. I'm happy to create the fixture and add the test, but I thought I'd see if you found the idea problematic before trying it.

EreMaijala: I added a third fixture that contains a non-namespaced collection element and also a record from another namespace and fixed the collection reader to actually look for a correct namespace. It's still a bit lenient as it also allows records without a namespace, but at least it will now skip records with a bad one.

module/VuFind/src/VuFind/Marc/Serialization/MarcXml.php
```diff
@@ -48,29 +69,72 @@ class MarcXml implements SerializationInterface
...
+     * @return bool
+     */
+    public static function canParseCollectionFile(string $file): bool
+    {
+        if (false === ($f = @fopen($file, 'r'))) {
+            throw new \Exception("Cannot open file '$file' for reading");
+        }
+        do {
+            $s = ltrim(fgets($f, 10));
+        } while (!$s);
```
demiankatz: What happens if you feed this a file containing only whitespace? Do we need extra logic to break out of the loop at EOF?

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcCollectionFileTest.php
demiankatz: I think that makes sense. Only one more question: should we emit a notice or warning if we encounter a bad namespace? I'm imagining a scenario where somebody builds XML by hand and makes a typo in the namespace URI, then can't figure out why the record won't load. Another scenario: since the namespace URI contains http:// and there's increasing pressure to move to https://, I could imagine a global search-and-replace going too far and confusing things. :-)

EreMaijala: I'd be inclined to say "no" at least without a verbose flag, since it would be analogous to encountering e.g. an `<item>` element instead of `<record>`. If you believe such flag would be useful, I guess it would be ok to do something like call `trigger_error()` with E_WARNING level if a verbose flag is set. What do you think?

WRT http vs. https, as far as I know XML namespace URLs should be considered opaque identifiers, and using https would mean that it's not the same schema as with http. Thus I wouldn't accept https in this case.

demiankatz: A verbose flag might be helpful, but if it's a nuisance to implement, I don't think it's essential on day one. (Though if you do want to implement it, I'm also thinking it might make sense to switch it on when calling this from the linter code, which would be a good way to demonstrate/test the functionality).

I absolutely agree about XML namespace URLs -- changing http to https would NOT be correct. It just seems like a naive mistake that somebody might make. :-)

EreMaijala: I don't particularly want to implement it, so I'd be happy to defer it to a later date. :)

I can actually imagine a situation where someone comes up with the great idea that they can interleave normal MARC records and "enhanced" format by using https for the enhanced format's schema. Or maybe it's just MARCXML 2.0. In any case, I still don't think it's a good idea to support something that's not done right.

module/VuFind/src/VuFind/Marc/Serialization/MarcXml.php
EreMaijala: Oh, bad things probably. I'll need to fix that.

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcCollectionFileTest.php
demiankatz: No, I wouldn't want to support an incorrect namespace -- but if somebody made that mistake, they might spend a lot of time wondering why it wasn't working, and a warning about a bad namespace could be very valuable in that situation. In any case, maybe it's worth a JIRA ticket so we don't forget about it (or maybe if we move the MARC code to a separate repo, it could be an issue there... though the current monorepo conversation also expands all the possibilities we can imagine :-) ).

module/VuFind/src/VuFind/Marc/Serialization/MarcXml.php
demiankatz: Probably worth another test fixture! :-) In any case, I think once you get through that, I can get down to the remaining testing work.

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcCollectionFileTest.php
EreMaijala: I think I'll add the verbose support after all since it has a minor effect on the constructor.

EreMaijala: Done as a separate commit. I decided to implement it as a callback so that anyone using the class can display the messages in a way they want, and did that for the deletes command.

module/VuFind/src/VuFind/Marc/Serialization/MarcXml.php
EreMaijala: This is now fixed, so no endless loop. However, I'm undecided on whether an empty file should be considered unknown format or ISO2709. Currently it is considered unknown, but it could be argued that an empty file is a valid one, just without any records.

module/VuFind/src/VuFind/Marc/Serialization/SerializationFileAbstractBase.php
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Abstract base class for serialization format support classes.
+ *
+ * @category VuFind
+ * @package  MARC
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:plugins:record_drivers Wiki
+ */
+abstract class SerializationFileAbstractBase implements SerializationFileInterface
```
demiankatz: Would AbstractSerializationFile be a better name for this?

module/VuFind/src/VuFind/Marc/Serialization/SerializationFileAbstractBase.php
```diff
@@ -0,0 +1,107 @@
...
+    protected $messageCallback = null;
+
+    /**
+     * Check if the serialization class can parse the given MARC collection file
+     *
+     * @param string $file File name
+     *
+     * @return bool
+     */
+    abstract public static function canParseCollectionFile(string $file): bool;
```
demiankatz: I might be mistaken, but I don't think you need to add abstract function declarations for interface methods in an abstract class unless you really want to.

demiankatz (commented):
Thanks for taking the time to add the message feature -- just a couple of minor comments, which you can use or ignore as you see fit.

module/VuFind/src/VuFind/Marc/Serialization/MarcXml.php
demiankatz: I have no problem with leaving it "Unknown" for now -- I don't think there are many scenarios where somebody would intentionally pass in an empty file, so treating that as an error rather than a valid but empty collection seems more appropriate. I suppose it might be annoying to get intermittent errors if there's an automated process that sometimes generates records and sometimes generates empty files... but if a process operating on an empty file fails and does nothing, the net result as the same as if it succeeds and does nothing, so I don't think it really matters.

module/VuFind/src/VuFind/Marc/Serialization/SerializationFileAbstractBase.php
EreMaijala: Yes, definitely. Changed.

module/VuFind/src/VuFind/Marc/Serialization/SerializationFileAbstractBase.php
EreMaijala: You're not mistaken, and I removed the declarations. Thanks, so much better. :)

EreMaijala:
@demiankatz Thanks for adding the tests for Unicorn and Voyager. I've merged them and fixed issues they uncovered. In Unicorn, there was a `$field->getTag()` that should have been `$field['tag']` and in Voyager there was an extraneous call for `next()` to get the record. With those fixed the tests are passing. :)

module/VuFind/src/VuFind/ILS/Driver/Voyager.php
```diff
@@ -891,16 +891,16 @@ protected function getMFHDData($record, $fieldSpecs)
             $subfieldCodes = substr($fieldSpec, 3);
             if ($fields = $record->getFields($fieldCode)) {
                 foreach ($fields as $field) {
-                    if ($subfields = $field->getSubfields()) {
+                    if ($subfields = $field['subfields'] ?? []) {
                         $line = '';
-                        foreach ($subfields as $code => $subfield) {
-                            if (false === strpos($subfieldCodes, (string)$code)) {
+                        foreach ($subfields as $subfield) {
+                            if (!strstr($subfieldCodes, $subfield['code'])) {
```
demiankatz: Is there a strong reason to switch to using strstr here? Isn't strpos more efficient and less prone to error (e.g. if the subfield being evaluated somehow happens to be a "falsy" string)?

module/VuFind/src/VuFind/Marc/MarcReader.php
```diff
@@ -84,13 +85,19 @@ public function __construct($data)
...
+     * associative array with 'leader' and 'fields' in the internal format
      *
      * @throws Exception
      * @return void
      */
-    public function setData(string $data): void
+    public function setData($data): void
     {
+        if (is_array($data)) {
+            $this->leader = $data['leader'];
```
demiankatz, line 102: Should we validate that the array comes in the expected format instead of assuming that leader and fields will be set?

module/VuFind/src/VuFind/Marc/Serialization/Iso2709.php
```diff
@@ -176,4 +243,47 @@ public static function toString(string $leader, array $fields): string
...
+    }
+
+    /**
+     * Rewind the collection file
+     *
+     * @return void;
+     */
+    public function rewind(): void
+    {
+        rewind($this->file);
```
demiankatz, line 271: Should we add some validation here and in getNextRecord to throw an exception if these get called when no file has been opened yet?

module/VuFind/src/VuFind/Marc/Serialization/MarcXml.php
```diff
@@ -188,4 +271,73 @@ protected static function loadXML(string $xml): \SimpleXMLElement
...
+    }
+
+    /**
+     * Get next record from the file or an empty string on EOF
+     *
+     * @return string
+     */
+    public function getNextRecord(): string
+    {
+        while ($this->xml->read()) {
```
demiankatz, line 316: As above, should we validate that we have opened an XML file before assuming we can read it?

module/VuFind/src/VuFind/Marc/Serialization/SerializationFileInterface.php
```diff
@@ -0,0 +1,81 @@
...
+    public static function canParseCollectionFile(string $file): bool;
+
+    /**
+     * Set message callback
+     *
+     * @param callable $callback Message callback
+     *
+     * @return void
+     */
+    public function setMessageCallback(?callable $callback): void;
```
demiankatz, line 57: I don't mind leaving this alone if it's easier, but does the message callback logic really belong in the serialization file interface? Mightn't messaging be useful for other kinds of serialization objects? Is this better as a separate interface or something lower-level?

module/VuFind/src/VuFind/RecordDriver/Feature/MarcAdvancedTrait.php
```diff
@@ -890,12 +945,6 @@ public function getXML($format, $baseUrl = null, $linker = null)
 
             // Set up proper namespacing and extract just the <record> tag:
             $xml->record->addAttribute('xmlns', "http://www.loc.gov/MARC21/slim");
-            $xml->record->addAttribute(
```
demiankatz, line 893: Why is this being removed? I don't doubt that there's a good reason for it, but it seems not to be directly related to the rest of this PR (unless I'm missing something).

module/VuFind/src/VuFind/RecordDriver/WorldCat.php
```diff
@@ -64,11 +64,6 @@ public function setRawData($data)
             $data = $data['fullrecord'];
         }
 
-        // Make sure the XML has an appropriate header:
-        if (strlen($data) > 2 && substr($data, 0, 2) != '<?') {
```
demiankatz, line 68: Is this being removed as unnecessary, or was it actively breaking something?

module/VuFind/src/VuFindTest/Feature/FixtureTrait.php
```diff
@@ -64,6 +86,7 @@ protected function getFixtureDir($module = 'VuFind')
      */
     protected function getFixture($filename, $module = 'VuFind')
     {
+        $realFilename = $this->getFixturePath($filename, $module);
         $realFilename = realpath($this->getFixtureDir($module) . $filename);
```
demiankatz: Looks like you added a line here but didn't delete the old line, so there's double work being done.

module/VuFindConsole/src/VuFindConsole/Command/Util/DeletesCommand.php
```diff
@@ -72,10 +71,9 @@ protected function configure()
                 "the format of the file -- it may be one of the following:\n"
                 . "flat - flat text format "
                 . "(deletes all IDs in newline-delimited file)\n"
-                . "marc - binary MARC format (delete all record IDs from 001 "
-                . "fields)\n"
-                . "marcxml - MARC-XML format (delete all record IDs from 001 "
-                . "fields)\n",
+                . "marc - MARC record in binary or MARCXML format (deletes all "
+                . "record IDs from 001 fields)\n"
+                . "marcxml - same as marc\n",
```
demiankatz: Should we deprecate `marcxml` since `marc` now autodetects format?

module/VuFindConsole/src/VuFindConsole/Command/Util/LintMarcCommand.php
```diff
@@ -74,14 +76,12 @@ protected function configure()
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $filename = $input->getArgument('filename');
-        $marc = substr($filename, -3) !== 'xml'
-            ? new \File_MARC($filename) : new \File_MARCXML($filename);
-        $linter = new \File_MARC_Lint();
+        $collection = new MarcCollection(file_get_contents($filename));
```
demiankatz: Should we update this to use MarcCollectionFile too?

demiankatz (commented):
@EreMaijala, I've just opened #2286 with a little more test coverage (though I don't expect it will change anything here). Additionally, I took a comprehensive look through all the changes here and added a handful of new comments. Let me know what you think about them, and then I'll do a final round of hands-on testing when everything else has been resolved.

module/VuFind/src/VuFind/RecordDriver/WorldCat.php
EreMaijala: MarcXml serialization class already takes care of it so there is no need to do it here.

module/VuFind/src/VuFind/ILS/Driver/Voyager.php
EreMaijala: That was copied by mistake from some local code. Reverted.

module/VuFind/src/VuFind/Marc/MarcReader.php
EreMaijala: I added checks for the basic elements, but I don't think it's worth it checking the full structure.

module/VuFind/src/VuFind/Marc/Serialization/Iso2709.php
EreMaijala: I first decided against it as the usage is protected by MarcCollectionFile, but now I realized that there's nothing stopping someone from instantiating the class standalone, so I added some checks.

module/VuFind/src/VuFind/Marc/Serialization/MarcXml.php
EreMaijala: Done.

module/VuFind/src/VuFind/Marc/Serialization/SerializationFileInterface.php
EreMaijala: I thought I would keep things simple and be able to expect the functionality from anything that implements SerializationFileInterface, but you're right in that it could prove useful in other cases too, so I moved the callback method to its own interface and added a check for it.

module/VuFind/src/VuFind/RecordDriver/Feature/MarcAdvancedTrait.php
EreMaijala: MarcXml serialization already added that. It wasn't quite right, though, since it's better to have it at the collection level. I changed it and reverted the change here. An additional problem was that this referred the xsi namespace without declaring it. I've now added definition for that, but it's a bit ugly due to a SimpleXML quirk.

module/VuFind/src/VuFind/Marc/MarcReader.php
demiankatz: Thanks!

module/VuFind/src/VuFind/Marc/Serialization/Iso2709.php
demiankatz: Thanks!

module/VuFind/src/VuFind/RecordDriver/Feature/MarcAdvancedTrait.php
demiankatz: Thanks! And yes, I've run into that SimpleXML quirk too -- quite annoying.

module/VuFind/src/VuFindTest/Feature/FixtureTrait.php
```diff
@@ -64,7 +86,7 @@ protected function getFixtureDir($module = 'VuFind')
      */
     protected function getFixture($filename, $module = 'VuFind')
     {
-        $realFilename = realpath($this->getFixtureDir($module) . $filename);
+        $realFilename = $this->getFixturePath($filename, $module);
         if (!$realFilename || !file_exists($realFilename)
```
demiankatz: This logic appears to be redundant since it has been refactored into getFixturePath above.

EreMaijala: I should have paid more attention. Fixed again. Third time lucky?

demiankatz (approved):

demiankatz:
Thanks, @EreMaijala, this is looking good to me, and I couldn't find a way to break anything. I've merged it and updated the [changelog](https://vufind.org/wiki/changelog#release_90_-_currently_in_development).

PR #2253, opened 2021-12-31
Update Croatian translation

milotype:
- one minor correction
- added and translated two missing strings


demiankatz:
Thanks for the updates!

PR #2255, opened 2022-01-04
Fix Variable in isset() always exists and is not nullable from Phpstan

xmorave2:
Fixes some issues found by PHPStan


demiankatz:
@xmorave2, as with #2256, there are some function signature changes here that should probably wait until a major version release. Do you mind if I bump this to the dev-9.0 branch, or would you like to try to separate the signature-changing changes from the lighter-weight fixes?

xmorave2:
Thanks @demiankatz. I have no objections to move this to dev-9.0. I rebased and change base of this PR too

demiankatz:
Thanks, @xmorave2 -- I'll try to finish reviewing (and hopefully merging) this and #2257 when I'm back in the office tomorrow.

module/VuFindSearch/src/VuFindSearch/Backend/Primo/Connector.php
```diff
@@ -603,44 +603,40 @@ protected function process($data, $params = [])
...
+     * @param bool    $onCampus  Whether the user is on campus
      *
      * @throws \Exception
      * @return string    The requested resource
      */
-    public function getRecord($recordId, $inst_code = null, $onCampus = false)
+    public function getRecord(string $recordId, $inst_code = null, $onCampus = false)
     {
         // Query String Parameters
-        if (isset($recordId)) {
```
demiankatz, line 616: I'm curious if @EreMaijala has any thoughts/concerns regarding this change. Would it be better to change it to a `strlen` check so we don't call the API if we're passed an empty string, or is there really no reason to have that empty response case in the code?

demiankatz (commented):
Thanks, @xmorave2, this all looks good to me, though I had one Primo-related question that I'm hoping @EreMaijala can weigh in on.

module/VuFindSearch/src/VuFindSearch/Backend/Primo/Connector.php
EreMaijala: Sorry for missing the question! I'm not sure why we would ever try to retrieve a record without an ID, but to be safe I'd just set the parameter type to string and check `if ('' !== $recordId) {`.

demiankatz: Looks to me like putting this at the top of the function:

<pre>
if ('' !== $recordId) {
    return self::$emptyQueryResponse;
}
</pre>

would cover the case while making things more readable -- better to have an early return than huge amounts of logic inside an if statement. @xmorave2, what do you think?

xmorave2: I do agree it is more readable and cleaner solution. I added it.

demiankatz: Thanks!

demiankatz (approved):

PR #2256, opened 2022-01-04
Fix Anonymous function has an unused use

xmorave2:
Found by PHPStan

### DIFF
--- a/module/VuFindConsole/src/VuFindConsole/Command/Language/AddUsingTemplateCommand.php
+++ b/module/VuFindConsole/src/VuFindConsole/Command/Language/AddUsingTemplateCommand.php
@@ -109,7 +109,7 @@ class AddUsingTemplateCommand extends AbstractCommand
             if (!$sourceDir) {
                 return 1;
             }
-            $sourceCallback = function ($full) use ($domain, & $tokens) {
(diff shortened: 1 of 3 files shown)


demiankatz:
Thanks, @xmorave2 -- glad to see you're doing some more work with PHPStan!

Since your change to the GeneratorTools changes a method signature in addition to fixing the issue detected by PHPStan, I would suggest one of two approaches:

1.) Leave the method signature unchanged, but add an unused parameter annotation; this way, if somebody wants to build a subclass that relies on the class name, the data is still available to the method.

2.) Change the target of this PR from dev to dev-9.0, so we can introduce a breaking change.

I personally think option 2 is the easiest approach, since I doubt anyone is subclassing this particular class, so the break shouldn't matter. But in the interest of proper semantic versioning, if we do go with that change, I think it's best to wait.

If you prefer option 1, please revise the PR. If you're okay with option 2, just let me know and I can take care of merging appropriately.

Thanks again!

xmorave2:
Thanks for a review @demiankatz. I rebased the branch and changed base of PR to dev-9.0. I do agree it is the better approach here.

demiankatz:
Thanks, @xmorave2!

PR #2258, opened 2022-01-04
[dev] Matomo fixes

EreMaijala:
This is a set of commits that fix several issues with Matomo tracking.

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/Matomo.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/Matomo.php
@@ -370,7 +370,8 @@ class Matomo extends \Laminas\View\Helper\AbstractHelper
             'Sort' => $params->getSort(),
             'Page' => $params->getPage(),
             'Limit' => $params->getLimit(),
-            'View' => $params->getView()
+            'View' => $params->getView(),
+            'Context' => $this->context ?: 'page'
         ];
     }
 
@@ -405,6 +406,7 @@ class Matomo extends \Laminas\View\Helper\AbstractHelper
         $institutions = $institutions;
 
         return [
+            'Context' => $this->context ?: 'page',
             'RecordFormat' => $formats,
             'RecordData' => "$id|$author|$title",
             'RecordInstitution' => $institutions
@@ -418,7 +420,9 @@ class Matomo extends \Laminas\View\Helper\AbstractHelper
      */
     protected function getLightboxCustomData(): array
     {
-        return [];
+        return [
+            'Context' => $this->context ?: 'page'
+        ];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


EreMaijala:
Note that while the addition of Context to custom data fields is not techically a fix, without the data it's not possible to segment the different kinds of loads in Matomo to e.g. filter out tab loads.

module/VuFind/src/VuFind/View/Helper/Root/Matomo.php
```diff
@@ -489,14 +489,11 @@ protected function getPageUrl(): string
         $path = $this->request->getUri()->toString();
         $routeMatch = $this->router->match($this->request);
         if ($routeMatch
-            && $routeMatch->getMatchedRouteName() == 'vufindrecord-ajaxtab'
+            && substr($routeMatch->getMatchedRouteName(), -8) === '-ajaxtab'
+            && null !== ($pos = strrpos($path, '/AjaxTab'))
+            && ($tab = $this->request->getPost('tab'))
         ) {
-            // Replace 'AjaxTab' with tab name in record page URLs
```
demiankatz, line 490: Should this comment be retained to clarify what is happening here? It's a bit dense to work through otherwise.

module/VuFind/src/VuFind/View/Helper/Root/Matomo.php
```diff
@@ -418,7 +420,9 @@ protected function getRecordPageCustomData(RecordDriverBase $recordDriver): arra
      */
     protected function getLightboxCustomData(): array
     {
-        return [];
+        return [
+            'Context' => $this->context
```
demiankatz: Is there a reason this is defaulted to page everywhere but here? (I assume there is, but maybe it's worth adding a comment since this is an exception to a pattern seen elsewhere in the code).

demiankatz (commented):
Thanks, @EreMaijala -- a couple of very minor suggestions.

module/VuFind/src/VuFind/View/Helper/Root/Matomo.php
EreMaijala: getLightboxCustomData should never be called in normal page context. There's no harm in making it identical to the others, though, so I'll just change it.

module/VuFind/src/VuFind/View/Helper/Root/Matomo.php
EreMaijala: Yes. I've now put it back but above the whole block.

module/VuFind/src/VuFind/View/Helper/Root/Matomo.php
EreMaijala: Changed.

PR #2259, opened 2022-01-05
Make getFacetLabel work for checkbox filters.

EreMaijala:
Since Params::getFilterList can return checkbox filters, it would make sense for it to return the labels as well.

This allows Piwik and Matomo to record custom variables or dimensions properly for checkbox facets instead of `unrecognized_facet_label`.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Base/Params.php
+++ b/module/VuFind/src/VuFind/Search/Base/Params.php
@@ -970,8 +970,6 @@ class Params
      * @param string $default Default field name (null for default behavior).
      *
      * @return string         Human-readable description of field.
-     *
-     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
     public function getFacetLabel($field, $value = null, $default = null)
     {
--- a/module/VuFind/src/VuFind/Search/Base/Params.php
+++ b/module/VuFind/src/VuFind/Search/Base/Params.php
@@ -973,20 +973,18 @@ class Params
      */
     public function getFacetLabel($field, $value = null, $default = null)
     {
-        $checkboxFacet = $this->checkboxFacets[$field]["$field:$value"] ?? null;
         if (!isset($this->facetConfig[$field])
             && !isset($this->extraFacetLabels[$field])
-            && null === $checkboxFacet
             && isset($this->facetAliases[$field])
         ) {
             $field = $this->facetAliases[$field];
-            $checkboxFacet = $this->checkboxFacets[$field]["$field:$value"] ?? null;
+        }
+        $checkboxFacet = $this->checkboxFacets[$field]["$field:$value"] ?? null;
+        if (null !== $checkboxFacet) {
+            return $checkboxFacet['desc'];
         }
         if (isset($this->facetConfig[$field])) {
             return $this->facetConfig[$field];
-        }
-        if (null !== $checkboxFacet) {
-            return $checkboxFacet['desc'];
         }
         return $this->extraFacetLabels[$field]
             ?? ($default ?: 'unrecognized_facet_label');
--- a/module/VuFind/src/VuFind/Search/Base/Params.php
+++ b/module/VuFind/src/VuFind/Search/Base/Params.php
@@ -975,14 +975,20 @@ class Params
      */
     public function getFacetLabel($field, $value = null, $default = null)
     {
+        $checkboxFacet = $this->checkboxFacets[$field]["$field:$value"] ?? null;
         if (!isset($this->facetConfig[$field])
             && !isset($this->extraFacetLabels[$field])
+            && null === $checkboxFacet
             && isset($this->facetAliases[$field])
         ) {
             $field = $this->facetAliases[$field];
+            $checkboxFacet = $this->checkboxFacets[$field]["$field:$value"] ?? null;
         }
         if (isset($this->facetConfig[$field])) {
             return $this->facetConfig[$field];
+        }
+        if (null !== $checkboxFacet) {
+            return $checkboxFacet['desc'];
         }
         return $this->extraFacetLabels[$field]
             ?? ($default ?: 'unrecognized_facet_label');
--- a/module/VuFind/tests/unit-tests/src/VuFindTest/Search/Base/ParamsTest.php
+++ b/module/VuFind/tests/unit-tests/src/VuFindTest/Search/Base/ParamsTest.php
@@ -31,6 +31,7 @@ namespace VuFindTest\Search\Base;
 use VuFind\Config\PluginManager;
 use VuFind\Search\Base\Options;
 use VuFind\Search\Base\Params;
+use VuFindTest\Feature\ReflectionTrait;
 
 /**
  * Base Search Object Parameters Test
@@ -44,6 +45,8 @@ use VuFind\Search\Base\Params;
  */
 class ParamsTest extends \PHPUnit\Framework\TestCase
 {
+    use ReflectionTrait;
+
     /**
      * Get mock configuration plugin manager
      *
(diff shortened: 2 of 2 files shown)


module/VuFind/src/VuFind/Search/Base/Params.php
```diff
@@ -975,15 +975,21 @@ public function addCheckboxFacet($filter, $desc)
      */
     public function getFacetLabel($field, $value = null, $default = null)
     {
+        $checkboxFacet = $this->checkboxFacets[$field]["$field:$value"] ?? null;
```
demiankatz: You can remove the UnusedFormalParameter annotation above now that we're using `$value`.

demiankatz (commented):
Thanks for this -- see below for one minor comment.

Beyond that, I think it would be helpful to add some test coverage for this change. To that end, I reviewed the existing Base\Params test and found it rather lacking, so I revised it in #2261. If you're happy with my work there, would you mind merging that and then revising this PR to add coverage for the new case? (Doing so may require setting up a mock plugin manager to return the desired configuration; let me know if you need any help with that).

module/VuFind/src/VuFind/Search/Base/Params.php
EreMaijala: True, now removed.

module/VuFind/src/VuFind/Search/Base/Params.php
```diff
@@ -970,20 +970,24 @@ public function addCheckboxFacet($filter, $desc)
      * @param string $default Default field name (null for default behavior).
      *
      * @return string         Human-readable description of field.
-     *
-     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
     public function getFacetLabel($field, $value = null, $default = null)
     {
+        $checkboxFacet = $this->checkboxFacets[$field]["$field:$value"] ?? null;
```
demiankatz: It seems like an oversight that facet aliases are not applied for checkbox facets. Perhaps that's better addressed in a separate PR, but I wonder if we should resolve the alias first, and then check for checkboxes (we'd have to change the behavior of `addCheckboxFilter()` as well to account for aliasing and keep things consistent -- obviously it would be wise to add coverage for weird edge cases to the test suite).

module/VuFind/src/VuFind/Search/Base/Params.php
```diff
@@ -970,20 +970,24 @@ public function addCheckboxFacet($filter, $desc)
...
+            && null === $checkboxFacet
             && isset($this->facetAliases[$field])
         ) {
             $field = $this->facetAliases[$field];
+            $checkboxFacet = $this->checkboxFacets[$field]["$field:$value"] ?? null;
         }
         if (isset($this->facetConfig[$field])) {
             return $this->facetConfig[$field];
         }
+        if (null !== $checkboxFacet) {
```
demiankatz, line 983: I think there's a question about precedence here. If we have a field active, plus a checkbox facet related to that field, which label should we use? Right now, the field is taking precedence over the checkbox, but I wonder if favoring the most specific option would be preferable. In any case, I've added a test case to #2261 to cover this situation, and it currently matches the behavior here (since that's the only way it will pass on the dev branch before this is merged) -- see commit 6585e4d37fc54a8b6547a3a4850858539a232819. If we change this, we'll need to revise that test... and in any case, we'll need to add more tests to cover situations where regular facet fields and checkbox facet fields do not overlap. Obviously that will have to wait until we can merge either this PR or #2261.

demiankatz (commented):
@EreMaijala, since #2261 cannot be merged yet, I decided to see if I could write some tests for the changes here, merge this one, and then commit those tests to #2261 until your return. However, I discovered that my test work just raised some questions that need to be answered before we can proceed. See review below. In the meantime, it also inspired me to add more tests to #2261, so it was not wasted time!

module/VuFind/src/VuFind/Search/Base/Params.php
EreMaijala: Makes sense. I changed the order so that aliases are checked first and checkbox facets before normal facets. Fortunately checkbox facets are added with `addFilter()` (there's no `addCheckboxFilter()`, right?), so that should already work.

module/VuFind/src/VuFind/Search/Base/Params.php
EreMaijala: Yes, now updated. Test revised and another added to confirm aliases work.

EreMaijala:
@demiankatz I believe I've managed to update this to include your suggestions.

module/VuFind/src/VuFind/Search/Base/Params.php
demiankatz: You're right, `addCheckboxFilter` was a typo -- sorry about that! I think I meant `addCheckboxFacet`. Did you check whether that one might need any adjustments?

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Base/ParamsTest.php
```diff
@@ -149,13 +152,42 @@ public function testGetFacetLabel(): void
...
 
+    /**
+     * Test that getFacetLabel works as expected with aliases.
+     *
+     * @return void
+     */
+    public function testGetFacetLabelWithAliases(): void
+    {
+        $params = $this->getMockParams();
+        $this->setProperty(
```
demiankatz, line 170: I wonder if this is a sign that we need an addFacetAlias() method, so it's not necessary to rely on reflection here... but that's probably a matter for a separate PR.

module/VuFind/src/VuFind/Search/Base/Params.php
demiankatz: ...and yes, on reviewing the code, I think it would be a good idea to make `addCheckboxFacet` alias-aware to ensure consistent behavior. Would you like to handle that here, or should I address it in a separate PR?

demiankatz (approved):

demiankatz (changes requested):
I approved this because I thought it was done, but on closer review, I think there are still some open questions. Did this just fall off the radar, @EreMaijala?

EreMaijala:
@demiankatz Indeed it did. I'll see to it.

module/VuFind/src/VuFind/Search/Base/Params.php
EreMaijala: Now I must admit that I'm not sure what it means to make `addCheckboxFacet` alias-aware. Since it's called with values from configuration, what does it need to handle?

demiankatz: The configuration could still use an alias as the field name. I think we need to map those to "true" names in order for things to work consistently. We could also potentially log a warning if we want to discourage that behavior.

EreMaijala: Still at loss.. Why would one define mappings in LegacyFacets but keep old field names for the checkbox facets? Or rather, why would we want to support that? I've been under the impression that `[LegacyFields]` exists only to map values from pre-existing searches (saved, bookmarked etc.), not for mapping current configuration.

demiankatz: You wouldn't do it intentionally, but it could happen by accident and cause subtle problems... that's why I suggest making it work but logging a warning.

EreMaijala: Ok. How about addFacet then? Shouldn't it have the same mapping logic then?

demiankatz: Yes, I'd think they should be consistent. I probably assumed it already did. :-)

EreMaijala: Nope. :) And then there's at least `getFacetOperator` as well. Sorry, I'm still not convinced this is the right path. If I miss any of the places where the mapping should have been added as well, the result is much worse than what we have now.

demiankatz: That's fine, it's a narrow edge case and at this point should be addressed separately if at all since the scope is larger than I thought.

EreMaijala: Ok. Was there something else that needs to be addressed here in this PR?

demiankatz: No, I don't think so, unless you think it's worth adding an addFacetAlias method (which I doubt -- as noted here, that might be better done as a subsequent PR, if at all).

EreMaijala: A subsequent PR, if anything, is fine for me. :)

demiankatz (approved):

PR #2261, opened 2022-01-05
Improve tests for Base Params and Solr Options; fix related issues.

demiankatz:


demiankatz:
@EreMaijala, this PR revises the VuFind\Search\Base\Params test to actually test the Base class instead of using Solr. Doing so required some special logic to deal with mock class names when determining the search class ID. I then added a VuFind\Search\Solr\Options test to ensure that the search class ID determination still works correctly for normal objects, which revealed a situation where a null value could get passed to a method expecting a Config object, so I addressed that as well.

I'm not exactly sure why the view initialization logic was factored to a trait when the actual properties and methods related to views are part of the base class, but I decided not to pull on that thread right now, as I already have enough distractions. Something that may benefit from a future review, though!

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -969,7 +969,19 @@ public function getRecommendationSettings($handler = null)
...
     {
         // Parse identifier out of class name of format VuFind\Search\[id]\Options:
-        $class = explode('\\', get_class($this));
+        $className = get_class($this);
+        $class = explode('\\', $className);
+
+        // Special case: if there's only one part, we may be testing with a mock
+        // object; if so, that's okay, but anything else is unexpected.
+        if (count($class) < 3) {
+            $parts = explode('_', $class[0]);
```
EreMaijala: Since we don't need the parts, I think it would be cleaner to just check with `strncmp('Mock_', $className, 5)`.

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -969,7 +969,19 @@ public function getRecommendationSettings($handler = null)
     public function getSearchClassId()
     {
         // Parse identifier out of class name of format VuFind\Search\[id]\Options:
-        $class = explode('\\', get_class($this));
+        $className = get_class($this);
+        $class = explode('\\', $className);
+
+        // Special case: if there's only one part, we may be testing with a mock
+        // object; if so, that's okay, but anything else is unexpected.
+        if (count($class) < 3) {
```
EreMaijala: Why `< 3` and not `< 4` as the comment at the top suggests?

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -969,7 +969,19 @@ public function getRecommendationSettings($handler = null)
     public function getSearchClassId()
     {
         // Parse identifier out of class name of format VuFind\Search\[id]\Options:
-        $class = explode('\\', get_class($this));
+        $className = get_class($this);
+        $class = explode('\\', $className);
+
+        // Special case: if there's only one part, we may be testing with a mock
```
EreMaijala: The "one part" doesn't match the actual `if` check. Maybe revise?

EreMaijala (changes requested):

module/VuFind/src/VuFind/Search/Base/Options.php
demiankatz: Good point -- I tend to "do what I know" rather than refresh my memory on all the string comparison functions, but this is much more elegant; I'll fix it next week.

module/VuFind/src/VuFind/Search/Base/Options.php
demiankatz: Good point -- I'll change it when I have time to make revisions.

EreMaijala:
@demiankatz With those fixes done, consider this approved.

demiankatz:
Thanks, @EreMaijala -- I believe I've addressed your review comments, but I've also enabled branch protection, so now I need your formal approval to merge this. I have a lot on my plate this week, so I don't mind waiting until you get back. :-)

EreMaijala (approved):

PR #2262, opened 2022-01-05
Add constant definitions for PHPStan

xmorave2:
This add basic definition for constants usually coming from environment - these are not discoverable by PHPStan.

This change is part of my effort to move PHPStan analysis to higher level.


demiankatz:
Thanks, @xmorave2, this is great! I think in the long term, it may be worthwhile to create a class to wrap around these environmental constants for ease of testing and reduction of global data, but I'm happy to see there is a much simpler solution to get PHPstan to work right.

One suggestion: how about renaming the file to phpstan-constants.php, so it's clear what context this is being used in if somebody is just browsing the file system?

xmorave2:
Thanks @demiankatz. I renamed the constants.php file as you suggested.

demiankatz:
Thanks, @xmorave2!

PR #2264, opened 2022-01-05
FOLIO: Add config setting to allow debug of GET requests.

demiankatz:
While troubleshooting #2236, I found this feature helpful, so I propose making it a config setting. Would appreciate a review from a FOLIO user if time permits!


todolson (approved):
This seems like a really good idea, good to be able to scope the debugging messages

demiankatz:
Thanks, @todolson!

bbusenius (commented):
This looks good to me.

PR #2266, opened 2022-01-10
Make ID used for ILS health check configurable.

he247:
SetUp an override variable in config.ini: healthCheckId

### DIFF
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -281,9 +281,12 @@ driver          = Sample
 ; loadNoILSOnFailure - Whether or not to load the NoILS driver if the main driver fails
 loadNoILSOnFailure = false
 
-; healthCheckId - Set a valid mmsid here to override the default ID of 1 in
-; module/VuFind/src/VuFind/ILS/Connection.php#L870
-; healthCheckId = 12345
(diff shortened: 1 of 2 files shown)


demiankatz:
Thanks, @he247, this is great! I'm having a busy week, but as soon as time permits, I'll test it (and possibly revise the .ini comment slightly to make it more generic -- bibliographic ID instead of mmsId, etc.). But this is a huge help, and I don't think it will take much to get it merged.

demiankatz:
Also, @he247, for future reference, it probably is a good idea to create a separate branch for each pull request you open. I see that you made these changes directly to your dev branch, which in the long term could become confusing for you when it gets out of sync with the upstream version. It doesn't cause any problems from the project's end, though, as long as you don't commit unrelated changes to this branch before the PR gets merged.

demiankatz (approved):

demiankatz:
Thanks again, @he247 -- I've revised the config.ini documentation and tested that this works as expected. I'll merge as soon as continuous integration passes.

he247:
Thank you @demiankatz for the fast integration and your helpful feedback! 😊

PR #2267, opened 2022-01-11
Refactor sorting logic to allow locale-specific behavior

xmorave2:
This added just to start discussion about sorting in VuFind.

When we sort using strcasecmp, the locale is not taken into account which leads to bad alphabetical order in some languages (like Czech for example).

Do we want to change all string comparison and sorting functions into variants respecting locales? Or do we even want to make it configurable?


demiankatz:
Thanks for raising this issue, @xmorave2. I think the biggest trade-off between `strcasecmp` and `strcoll` is that (at least according to the [PHP manual](https://www.php.net/manual/en/function.strcoll.php)) `strcoll` is case-sensitive, so if our intention is to do a case-insensitive comparison, it may cause problems. Another possible consideration is binary safety (`strcoll` is not binary safe, while `strcasecmp` is), though I suspect that is not a significant problem for most parts of VuFind.

It does seem that different institutions may have different needs and preferences with regard to sorting, so maybe something configurable is desirable... but at the same time, making sorting configurable may be complex. Do we need a `VuFind\String` service that contains sorting and comparison methods that can be controlled by configuration? Is that overkill? Does PHP even contain the functions we need for our purposes, or might we need to investigate libraries that do things more flexibly or consistently?

Obviously, as you can see, I still have more questions than answers at this point... but I think this is a useful conversation to have. I'll also be interested to hear what @EreMaijala thinks, since he has more experience working in mixed-language data sets. However, he's on holiday this week, so we won't hear from him for a few days.

xmorave2:
@demiankatz Instead of strcoll, we could you Collator from intl extension - it is binary safe. 

I don't know any library solving this problem, but we could simply encapsulate the sorting into own class which internally could choose between collator->compare and strcasecmp (and probably some more options) based on configuration. This would be the cleanest solution I think...

demiankatz:
@xmorave2, the `Collator` class does look promising, and relying on the `intl` extension shouldn't be a problem since we already use it for many other things.

I agree that it seems to make the most sense to encapsulate this functionality in a class. That provides a useful extension point for customization, too, since users can override a service to customize sorting behavior globally.

EreMaijala:
We already use Collator locally, so thumbs up for it.

xmorave2:
I started some work on this - the main idea was to identify places where comparing/sorting is used for strings for display to user and use the new Sorter class to sort.

I added the Sorter class into Service namespace, but maybe better to use I18n namespace here, what do you think?

demiankatz:
Thanks, @xmorave2, this looks like a good start (though I confess that I haven't done a deep review yet). A few initial comments:

1.) I'm inclined to agree that the I18n namespace is better than Service.
2.) I'm not especially enthusiastic about adding a new "Aware" interface/trait, as I've been trying to move the code in a more explicit direction to make it easier to analyze. I think it's a great approach for quickly developing and testing the approach, but once we settle on other details, it may be worth revising some of these classes to accept an explicit dependency. At very least, maybe the trait should have a getSorter() method that constructs a default object if none is injected, so that we don't add a requirement of calling a setter in order to get working objects.

demiankatz:
@xmorave2, what is the status of this PR? It looks like you added some commits without further comment, and if you're waiting for me to review again, I may have missed the notifications -- if so, I apologize! I also notice that there are some conflicts which have accumulated since we last discussed this. When you have a chance, can you resolve the conflicts and let me know if you're ready for further review?

Thanks!

xmorave2:
@demiankatz 

I merged current `dev-9.0` branch and remove "Aware" magic. Could you review it now? Thanks

module/VuFind/src/VuFind/I18n/SorterTrait.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Get the sorter
+     *
+     * @return Sorter
+     */
+    public function getSorter(): Sorter
+    {
+        return $this->sorter ?? new Sorter();
```
demiankatz: Might it be better to throw an exception or log a warning if there is no sorter defined? I'm concerned about a scenario where forgetting to inject a dependency causes inconsistent sorting behavior which leads to confusion. Of course, maybe allowing this to work seamlessly is a more important priority... I'm certainly open to discussion on this!

module/VuFind/src/VuFind/Hierarchy/TreeDataFormatter/AbstractBase.php
```diff
@@ -36,8 +36,10 @@
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org/wiki/development:plugins:hierarchy_components Wiki
  */
-abstract class AbstractBase
+abstract class AbstractBase implements \VuFind\I18n\SorterInterface
 {
+    use \VuFind\I18n\SorterTrait;
```
demiankatz: Do we need to adjust corresponding factories to inject the sorter?

module/VuFind/src/VuFind/I18n/Sorter.php
```diff
@@ -0,0 +1,114 @@
...
+     * Compare function
+     *
+     * @param string $string1 First string to compare
+     * @param string $string2 Second string to compare
+     *
+     * @return int
+     */
+    public function compare(string $string1, string $string2): int
+    {
+        if ($this->respectLocale) {
```
demiankatz: Purely a stylistic suggestion which you are free to ignore, but all of these methods could be made slightly more concise with a ternary operator... e.g. `return $this->respectLocale ? $this->collator->compare($string1, $string2) : strcasecmp($string1, $string2);`

module/VuFind/src/VuFind/I18n/SorterInterface.php
```diff
@@ -0,0 +1,51 @@
...
+interface SorterInterface
+{
+    /**
+     * Set the sorter
+     *
+     * @param Sorter $sorter Sorter service
+     *
+     * @return void
+     */
+    public function setSorter(Sorter $sorter): void;
```
demiankatz: Would it make sense for SorterInterface to also include getSorter(), if that's the standard mechanism we plan to use for obtaining the sorter?

module/VuFind/src/VuFind/ILS/Driver/Unicorn.php
```diff
@@ -45,9 +45,11 @@
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     http://code.google.com/p/vufind-unicorn/ vufind-unicorn project
  **/
-class Unicorn extends AbstractBase implements \VuFindHttp\HttpServiceAwareInterface
+class Unicorn extends AbstractBase implements \VuFindHttp\HttpServiceAwareInterface,
+    \VuFind\I18n\SorterInterface
 {
     use \VuFindHttp\HttpServiceAwareTrait;
+    use \VuFind\I18n\SorterTrait;
```
demiankatz: Are we missing a factory change for this one?

module/VuFind/src/VuFind/Search/Solr/HierarchicalFacetHelper.php
```diff
@@ -43,9 +45,10 @@
  * @link     https://vufind.org Main Site
  */
 class HierarchicalFacetHelper implements HierarchicalFacetHelperInterface,
-    TranslatorAwareInterface
+    TranslatorAwareInterface, SorterInterface
 {
     use TranslatorAwareTrait;
+    use SorterTrait;
```
demiankatz: Missing corresponding factory update?

module/VuFind/src/VuFind/View/Helper/Root/SortFacetList.php
```diff
@@ -64,7 +67,7 @@ public function __invoke($results, $field, $list, $searchRoute)
                 ->addFacet($field, $value['value'])->getParams();
             $facets[$url] = $value['displayText'];
         }
-        natcasesort($facets);
+        $this->getSorter()->asort($facets);
```
demiankatz: Is the loss of natural case sort potentially a problem here?

module/VuFind/src/VuFind/View/Helper/Root/Sorter.php
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Compare function
+     *
+     * @param string $string1 First string to compare
+     * @param string $string2 Second string to compare
+     *
+     * @return int
+     */
+    public function compare(string $string1, string $string2): int
```
demiankatz: Would it be simpler to just return the sorter object via an __invoke method rather than duplicating all of the method signatures?

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/SorterTest.php
```diff
@@ -0,0 +1,212 @@
...
+{
+    /**
+     * Test compare function
+     *
+     * @return void
+     */
+    public function testCompare()
+    {
+        $testCases = [
+            [
```
demiankatz, line 53: No pressure if you don't want to spend the time, but some of the tests in here might benefit from using a `@dataProvider`.

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/SorterTest.php
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * Test asort function
+     *
+     * @return void
+     */
+    public function testAsort()
+    {
+        $testCases = [
+            [
+                'input' => ['a' => 'a', 'c' => 'c', 'b' => 'b'],
```
demiankatz: With regard to my natcasesort question above, it might be useful to have a test with 1, 2, 3, 10, 100 in case we need a method that will reliably do natural sort and not put 10 and 100 between 1 and 2.

themes/bootstrap3/templates/search/advanced/solr.phtml
```diff
@@ -35,7 +35,7 @@
                   }
                 }
               }
-              natcasesort($sorted);
+              $this->sorter()->sort($sorted);
```
demiankatz: Another instance of natcasesort being changed, though into a different method in this case than in the other facet scenario.

themes/bootstrap3/templates/search/advanced/summon.phtml
```diff
@@ -16,7 +16,7 @@
             foreach ($list['list'] as $i => $value) {
               $sorted[$i] = $value['displayText'];
             }
-            natcasesort($sorted);
+            $this->sorter()->sort($sorted);
```
demiankatz: If we change solr.phtml above, we should remember to make this match.

demiankatz (changes requested):
Thanks for the progress on this, @xmorave2 -- some questions, comments and suggestions below!

demiankatz:
@xmorave2, I also renamed this PR to reflect its broader scope since its original creation.

module/VuFind/src/VuFind/I18n/SorterTrait.php
xmorave2: I think it is not so important. It throws an exception now.

module/VuFind/src/VuFind/Hierarchy/TreeDataFormatter/AbstractBase.php
xmorave2: I've done it

module/VuFind/src/VuFind/I18n/Sorter.php
xmorave2: I agree it is more readable, so I've changed it.

module/VuFind/src/VuFind/I18n/SorterInterface.php
xmorave2: I make a bit more refactoring here:
* renamed SorterInterface to SortingInterface
* add SorterInterface a and implement it in Sorter class
* getSorter and setSorter uses SorterInterface as type instead of Sorter class

module/VuFind/src/VuFind/ILS/Driver/Unicorn.php
xmorave2: Fixed

module/VuFind/src/VuFind/View/Helper/Root/SortFacetList.php
xmorave2: I added new method 'natsort' to sorter and use it here.

module/VuFind/src/VuFind/View/Helper/Root/Sorter.php
xmorave2: That's much cleaner, I've changed it.

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/SorterTest.php
xmorave2: dataProvider looks like it is great for this type of test. Thanks for the suggestion. I've used it.

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/SorterTest.php
xmorave2: I added some more test altogether with new `natsort` method

themes/bootstrap3/templates/search/advanced/solr.phtml
xmorave2: changed to natsort

themes/bootstrap3/templates/search/advanced/summon.phtml
xmorave2: changed to natsort

xmorave2:
@demiankatz Thanks for the review. I've tried to address all your comments, so I think it is ready for another round.

module/VuFind/src/VuFind/I18n/SorterInterface.php
```diff
@@ -0,0 +1,70 @@
...
+    public function sort(array &$array): bool;
+
+    /**
+     * Sort array by values and maintain index association
+     *
+     * @param array $array Array to sort
+     *
+     * @return bool
+     */
+    public function asort(array &$array): bool;
```
demiankatz, line 69: Should you also include natsort in the interface?

module/VuFind/src/VuFind/Search/Solr/HierarchicalFacetHelperFactory.php
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @throws ServiceNotFoundException if unable to resolve the service.
+     * @throws ServiceNotCreatedException if an exception is raised when
+     * creating a service.
+     * @throws ContainerException&\Throwable if any other error occurs
+     */
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
```
demiankatz, line 66: Might be worth doing something with $options here to avoid an unused parameter warning (even if it's just the usual check that throws an exception if it is unexpectedly non-null).

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/SorterTest.php
```diff
@@ -0,0 +1,353 @@
...
+
+    /**
+     * Data Provider for sort tests
+     *
+     * @return array
+     */
+    public static function sortProvider(): array
+    {
+        return [
+            [
```
demiankatz, line 172: You can name the keys in the array returned by a data provider, which is sometimes useful if you want to quickly identify exactly which test case failed, or if you want to briefly describe the reason for a particular set of test data. I certainly don't insist on you adding/changing anything further, but just mentioning it for future reference in case you were not already aware.

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/SorterTest.php
```diff
@@ -0,0 +1,353 @@
...
+                ],
+                ['a', 'c', 'č', 'd'],
+            ],
+            [
+                [
+                    'input' => ['100', '3', '10', '2', '1'],
+                    'locale' => 'cs',
+                    'respectLocale' => true,
+                ],
+                ['1', '2', '3', '10', '100'],
```
demiankatz, line 210: I'm a little surprised that we're getting a natural sort here -- I would have expected 10 and 100 to come between 1 and 2. Is this because the strings are getting treated as integers behind the scenes? If so, is that a bug or a feature? Either way, maybe it would be good to also have a test case similar to the one in your natcase tests below (e.g. a1, a2, a10), just to show differentiated behavior.

module/VuFind/src/VuFind/I18n/SorterInterface.php
demiankatz: Would it be less confusing if we renamed SortingInterface to HasSorterInterface? This way, SorterInterface defines how we sort, and HasSorterInterface means we have access  to a Sorter. I think that might be a clearer distinction than Sorter vs. Sorting.

demiankatz (changes requested):
Thanks, @xmorave2, this is greatly improved! I have a few more small comments/suggestions.

module/VuFind/src/VuFind/I18n/SorterInterface.php
xmorave2: Yes, you are right, I've added it.

module/VuFind/src/VuFind/Search/Solr/HierarchicalFacetHelperFactory.php
xmorave2: I've added suggested check and throw exception

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/SorterTest.php
xmorave2: I am aware of ability to name test using array key, but it was enough for me the default behaviour which prints the whole data item in case the test fails ;)

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/SorterTest.php
xmorave2: It is sorted like numbers. It is feature of PHP itself when numeric string are treated as numbers and compared like numbers then. I any case, I've just added the case you suggested to show the sorting is not "natural" on strings

xmorave2:
Thank for quick review @demiankatz. I made some more adjustments as you suggest and also renamed the SortingInterface to HasSorterInterface. What about SorterTrait? Should we rename it too somehow?

demiankatz:
Thanks for the quick updates, @xmorave2! I would rename the trait to HasSorterTrait so that it matches the interface that it implements.

xmorave2:
@demiankatz I renamed the trait too.

demiankatz (approved):
Thanks, @xmorave2, this looks good to me now, and the full test suite is passing.

However, since this touches a lot of files, I'm going to also request a review from @EreMaijala before merging, just in case I've overlooked anything!

config/vufind/config.ini
```diff
@@ -2246,3 +2246,10 @@ show_solr_options_in_tag_search = false
 description = "The REST API provides access to search functions and records contained in the search index."
 ; URL pointing to a Terms of Service page (optional, default is none):
 ;termsOfServiceUrl = "https://something"
+
+[Sorting]
+; Sorting in VuFind is case insensitive and does not respect locales. By setting
+; this option to true, the sorting will be respecting locales but also case
```
EreMaijala: Is there a reason for case-sensitivity? Could we rather set the strength by default to `Collator::SECONDARY` or make it configurable?

module/VuFind/src/VuFind/I18n/Sorter.php
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Constructor
+     *
+     * @param string $locale        Current user locale
+     * @param bool   $respectLocale Do respect current locale?
+     */
+    public function __construct(string $locale = 'en', bool $respectLocale = false)
+    {
+        $this->collator = new \Collator($locale);
```
EreMaijala: I think Sorter should take Collator as a constructor parameter instead of instantiating it. This would allow setting further options for the Collator in the factory instead of overriding Sorter class.

EreMaijala (changes requested):

module/VuFind/src/VuFind/I18n/Sorter.php
demiankatz: I agree, this would be more flexible. If you want to have an easy way to build a Sorter without having to also construct a Collator, you could create a static method like `buildSorterForLocale` that could be used as an alternate constructor (e.g. for more concise Sorter construction in unit tests).

config/vufind/config.ini
xmorave2: I've set the collator to be case insensitive for more backwards compatibility

module/VuFind/src/VuFind/I18n/Sorter.php
xmorave2: I've changed it as you suggest

xmorave2:
Thanks @EreMaijala for review, I tried to address your comments.

config/vufind/config.ini
```diff
@@ -2246,3 +2246,8 @@ show_solr_options_in_tag_search = false
 description = "The REST API provides access to search functions and records contained in the search index."
 ; URL pointing to a Terms of Service page (optional, default is none):
 ;termsOfServiceUrl = "https://something"
+
+[Sorting]
+; Default sorting in VuFind is case insensitive and does not respect locales. If
```
EreMaijala: I think you can remove the "is case insensitive and" part.

demiankatz:
Thanks, @xmorave2 -- I made some minor adjustments to the .ini comment. I think this is ready to merge with @EreMaijala's final approval. Please let me know if you'd like me to re-run the full integration test suite one more time.

EreMaijala (approved):
Fix the comment in config.ini and I'll be happy. I'll just approve now anyway since I'll be away for a few days. :)

config/vufind/config.ini
demiankatz: Good point -- I totally rewrote the comment.

xmorave2: Thanks @demiankatz

xmorave2:
As this has two approves and is passing CI I've merged this into dev-9.0

demiankatz:
Thanks, @xmorave2!

PR #2268, opened 2022-01-11
Web-based installer: add installation report link; format success message as info box.

demiankatz:
This PR adds a link to the new installation report form to encourage users to share their VuFind instances with the community. It also reformats the existing message about disabling auto-configuration as an info box for a more consistent display.

### DIFF
--- a/themes/bootstrap3/templates/install/home.phtml
+++ b/themes/bootstrap3/templates/install/home.phtml
@@ -13,5 +13,7 @@
 <?php endforeach; ?>
 
 <?php if ($errors == 0): ?>
-  <p>No problems were found.  You may wish to <a href="<?=$this->url('install-done')?>">Disable Auto Configuration</a> at this time.</p>
+  <div class="alert alert-info">No problems were found.  You may wish to <a href="<?=$this->url('install-done')?>">Disable Auto Configuration</a> at this time.</div>
 <?php endif; ?>
+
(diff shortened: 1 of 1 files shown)


demiankatz:
There are some questions we should discuss:

1.) Is the wording of the new message acceptable or can it be improved?

2.) Do we want to display the info message(s) below the status notes (as they currently are) or move them to the top of the page for greater visibility?

Screen shots of the new/changed messages in action:

Before all problems have been fixed:

image

After all problems have been fixed:

image

lahmann (approved):
Thanks Demian, this looks like a reasonable spot for the message in the installation process, I don't see why this shouldn't get merged right away.

hajoseng:
I agree with @lahmann.

xmorave2:
Looking fine. Great idea with form for reporting new instances @demiankatz. Would like to have it translated in other languages too? For now, as some string are same as in base VuFind, the form is partially translated in some languages. If you think it could be heplful to have it translated I could send you Czech translation for the needed strings.

demiankatz:
Thanks, everyone. I'll leave this open until next week to give others a chance to offer feedback, but I'll merge it then if no changes are suggested.

@xmorave2, if you would like to share Czech translations for the form fields that are not part of standard VuFind language files, I'm happy to add them to the demo instance. No pressure if you have more urgent matters to attend to, though. :-)

demiankatz:
Thanks again, everyone -- since there has been no further feedback, I have gone ahead and merged this.

PR #2271, opened 2022-01-14
Ajax facets - fetch only requested facets from Solr

xrosecky:
We found that getSideFacets AJAX call requests all facets, not only from the _enabledFacets_ parameter.


module/VuFind/src/VuFind/Search/SearchRunner.php
```diff
@@ -140,6 +140,10 @@ public function run(
             compact('params', 'request', 'runningSearchId')
         );
 
+        if (isset($rawRequest['enabledFacets'])) {
+            $params->retainFacets($rawRequest['enabledFacets']);
```
demiankatz: While I don't see any major harm in globally enabling an enabledFacets query parameter, perhaps it would a more targeted fix to put this logic in the setup callback created by GetSideFacets::getFacetResults() instead of the SearchRunner itself.

demiankatz (commented):
Thanks for this! I'm interested in hearing @EreMaijala's opinion of this when he's back in the office next week. I think it's workable as-is, but have an alternative suggestion below. Additionally, it might be nice to add some unit test coverage for the new functionality. Perhaps we should put this on hold until #2261 gets merged, and then we can add some additional test cases to the improved Base Params test I've created there.

module/VuFind/src/VuFind/Search/SearchRunner.php
xrosecky: I tried to put this logic to setup callback at first, but it seems to be overridden later in the code, so I moved it to SearchRunner.

xrosecky: Or maybe I am doing something wrong.

demiankatz: I think what's happening is that the setup function is setting up the RecommendListener, which is then setting up the Facets recommendation module when the "setup complete" event triggers. So you're clearing out the facets first, then the recommendation module is putting them back in. I wonder if a better thing to do would be to manipulate the settings here:

https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/AjaxHandler/GetSideFacets.php#L199-L203

Right now, there's already code that converts SideFacetsDeferred into SideFacets. Maybe we also need to adjust other configuration settings.

I admit that I can't remember all the details of this, so maybe that's not practical -- I realize that the recommendation module points to a configuration file section, so maybe that's too indirect to usefully manipulate. But I'm not sure I understand why SideFacetsDeferred would get set up in the configuration file with one set of facets, and then we would need to load a different set when the AJAX call occurs.

I wonder if @EreMaijala can shed any light on this upon his return.... I do think the existing solution here will probably work, but I wonder if we're just masking a problem that could be solved in a better way.

EreMaijala: One problem with the current patch, as far as I can see, is that it will prevent displaying counts for checkbox facets (we have a local implementation for that). So at least those should be taken into account.

In the bigger picture I think the problem is that side facets are initialized by RecommendListener and it all happens quite far away from the context of the search request, so passing information from the context is not so straight-forward. One option would be to bypass RecommendListener in GetSideFacets and just activate the required facets by providing a filter to the init method. I don't think it's feasible to modify the configuration beforehand since it's quite complex with some logic built into the SideFacets recommendation module. This of course breaks the isolation, but would also avoid enabling any extra recommenders when we just need results for side facets.

demiankatz: I didn't review the code extremely closely, so maybe I'm missing something, but if the only reason we're using RecommendListener and the SideFacets module is to activate facets in the Params object, it does seem a lot more straightforward and efficient to just activate the facets we want directly in the setup callback instead of manipulating configurations and wiring up listeners. Then we can just turn on what we need, instead of turning everything on and then turning some things back off again.

It looks to me like there may be some cases where the AJAX routine is passed a list of facet fields (which is easy to deal with) and others where it's not. I didn't trace things back to see what all the circumstances might be. But one option would be to use the listener in cases where we don't know what we want, and to bypass it when we have a known list of facet fields.

EreMaijala: We still need to know something about the facet configuration, such as the number of facets to be displayed, and handling all that separately from SideFacets seems error-prone.

demiankatz: Aren't all of those configurations already being processed in SideFacetsDeferred? Would it be cleaner to pass the AJAX routine the information it needs to fetch the relevant data, rather than recalculating all the settings every time? I realize that might be a significant change/refactor, but it seems like what we currently have is pretty opaque, and maybe something simpler and more explicit would be more efficient and less potentially error-prone.

EreMaijala: Perhaps, yes. I'm a bit hesitant on making that modifiable by an AJAX request, though, since it would make it possible manually craft a request to e.g. ask for a million facets, and that might have serious performance implications.

demiankatz: Unless I'm missing something, facet limits are initialized by the Params object, not by the recommendation module (https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/Search/Params/FacetLimitTrait.php). I don't think SideFacets has anything to do with that.

demiankatz: I think the piece that _is_ handled by SideFacets is figuring out which facets are OR facets and which are not... but I think there's less danger of abuse if those kinds of settings are passed in from the outside.

EreMaijala: Oh, sorry about the confusion. That wouldn't be scary at all then.

demiankatz: Just reviving this thread to try to figure out where we left this -- I'm not sure who is waiting for whom at this point. I see that #2261 (discussed above) has now been merged, so maybe one possible next step is to add some tests to try to cover the changed functionality. @xrosecky, do you have any sense of what next steps should be? Do you need any specific guidance from @EreMaijala or me? (I think I'll have to take a bit of time to refresh my memory of this whole conversation before I can make a useful contribution... but I can at least tell at a glance that we never fully resolved it).

demiankatz:
@xrosecky, it's been a few months without activity here; just checking again on the status of this work. In any case, I've gone ahead and merged the latest dev branch into the PR to see if that will get us a green checkmark again.

demiankatz:
@xrosecky, I'm reviewing older, less active pull requests and see that this still hasn't moved in quite some time. Is this something you are still interested in working on?

demiankatz:
We discussed this on today's Community Call, and I've set a 9.1 milestone since this seems like a relatively small piece of work that we can complete fairly quickly (though we can also close it out if @xrosecky feels it is no longer needed/valuable).

Assuming that we want to move forward, based on a review of outstanding discussions here, I think these are the remaining tasks/decisions:

1.) We need to determine whether there is a better place than SearchRunner to apply the enabledFacets setting -- I suggested that this might be possible in the GetSideFacets AjaxHandler instead, and I believe that @EreMaijala offered some possible alternatives as well.

2.) We need to account for checkbox facet counts (as highlighted by @EreMaijala).

3.) It would be helpful to have code somewhere that actually _uses_ the new setting, to provide context. I assume if we're adding the back-end feature, there is some front-end place where we should be using it.

Let me know what you think! If we're on the same page about what needs to happen, we can set up some TODO checkboxes at the top of the PR if that would be helpful.

demiankatz:
I've just merged the latest dev branch here to get our green checkmark back (since this hasn't run since we added PHP 8.2 testing). Still waiting to see if @xrosecky is interested in moving this forward.

xrosecky:
OK, I will look into it.

xrosecky:
It seems I don't have rights to update this pull request, so I created another one: https://github.com/vufind-org/vufind/pull/2974.

demiankatz:
That's strange, @xrosecky, I'm not sure why you would be unable to update your own pull request... but GitHub has sometimes done strange things over the past few months. In any case, thanks for opening the new one!

PR #2272, opened 2022-01-18
Implement monorepo development model

aleksip:
We have developed an interface for administering multiple VuFind views. It is currently closed source, but hopefully we could make it at least partially open source at some point. We are currently refactoring the codebase to adhere to VuFind's coding standards and to use the same dependencies as VuFind.

There are also several opportunities to use functionality from VuFind itself. Some examples we have thought of are the asset pipeline, and form and configuration file handling. While we can always copy the code over, this is of course much less convenient than adding it as a dependency.

Since splitting VuFind into multiple packages in their own repositories is also not very convenient, one possible way to have our cake and eat it too is to use the monorepo approach.

This PR contains an initial proof of concept implementation of splitting VuFind's asset pipeline code into a separate vufindview package, while still maintaining the monolithic development and release model.

I have tested the related GitHub action, updating a read-only vufindview package repository and using it in our administration interface as a Composer dependency. While there are still some things to solve and decide, it works!

The monorepo split workflow is currently set up to work with my test repositories, and needs to be changed for testing under the vufind-org GitHub organisation.

One benefit of developing packages like this is that it should produce better code. For example I found a couple of bugs in the pipeline code when using it in our administration interface.

I have not moved related tests over yet, but the PR contains many required changes for moving tests to the package.

Related links:

- https://github.com/symplify/monorepo-builder
- https://github.com/symplify/monorepo-split-github-action
- https://blog.logrocket.com/hosting-all-your-php-packages-together-in-a-monorepo/


demiankatz:
@aleksip, thanks for sharing this; it's an interesting idea. I think there's quite a lot to discuss!

First of all, as you may know, we already have a number of VuFind components that have been broken off to their own projects. I think it's probably preferable to stick with one approach: either continue to add separate repositories, or else move more things into the monorepo. Personally, I think sticking with separate repositories may actually be preferable, but I haven't done a deep study of this option, so I may be under-informed. Let me raise what I see as the potential disadvantages of this approach, and you can tell me whether I've persuaded you, or I'm mistaken. ;-)

1.) I think in the ideal situation, the split-off packages will have a different release cadence than VuFind itself. Some may change very infrequently; others may benefit from having fixes and new features released quickly/frequently. Having to release them in lock-step with the larger package seems disadvantageous, as does tying together version numbers, if that's a potential side effect.

2.) VuFind tends to pin very specific dependency versions, because as an application, I like to have fine-grained control over when things change. Smaller components should have looser dependencies so that they can be more easily incorporated into applications using a variety of versions of shared dependencies. I don't think anything about mono-repo prevents this (just a matter of editing composer.json files differently), but this somehow makes me feel better about keeping things separated.

3.) While there is a setup overhead, and convenience does count for something, I like having separate continuous integration set up for each of these projects so that they can be tested and measured independently in Jenkins. I suspect that's easier to manage with separate repositories.

I'm not sure whether or not that's very persuasive, but it's a start of talking points. :-) One other general issue to think about:

- I wonder whether the VuFindView vs. VuFind\View namespacing is going to be confusing. Do we want a single package with all of VuFind's view helpers, or do we want separate packages for specific groups of functionality? Just wondering about what granularity makes the most sense....

Anyway, thanks again for getting this ball rolling. I definitely agree that it's worth splitting this functionality out for the reasons you've already cited, I think it's just a matter of deciding on the specifics of naming and approach. If it turns out that a new project is a better approach, I'm happy to establish new projects in the vufind-org organization as needed.

aleksip:
@demiankatz Thanks, and indeed, this was just meant to be a starting point for discussion. There are several main approaches, and within them several implementation details to decide on.

I was aware of the separate VuFind components, and continuing with that approach is one option. However, it does come with the overhead of more maintenance and makes development somewhat harder, especially if we were to end up with a lot of packages. And because of this it will most likely raise the bar when deciding whether to make something a separate package. Which can be considered a good thing too, of course.

It should be relatively easy to move packages in to and out of a monorepo. So the existing separate packages could be moved to the monorepo while maintaining the existing repositories as read-only. And the monorepo approach can be abandoned by either moving development to the read-only repositories or by not publishing to them any more.

1) I don't think there is anything in the general idea of monorepos that forces all packages to use the same release cycle or versioning. The monorepo-builder tool I have used defaults to simultaneous releases with identical versioning, but I believe that the tool can be customized, or a different tool can be used. Or everything could just be done manually as well. I think this is an implementation detail we can freely decide on.

2) In the current PR the vufindview package dependency versions are fixed because the monorepo-builder tool by default writes them as-is to the main composer.json, and at this stage I did not want to change them. I agree that a general use package should have looser dependencies, and this is one of those things we would have to solve.

3) I'm not sure if I know everything that this would involve, but since the packages are published to separate repositories, they could still have their own CI workflows.

Ah, naming things... 😅 I have no strong thoughts on this, but the logic in the current PR is based on package names and namespaces from the existing VuFind components, e.g. vufindhttp and VuFindHttp.

My initial thought would be to create packages corresponding with related Laminas packages when applicable. So vufindview for everything based on laminas-view, vufindform for everything based on laminas-form, vufindconfig for everything based on laminas-config etc. This would mean eventually moving some other helpers to vufindview as well.

EreMaijala:
Maybe vufind-view and vufind-config then?

demiankatz:
Thanks, @aleksip, that's all helpful. A few follow-on thoughts:

1.) In terms of versioning, I'm accustomed to using Git tags to manage versions... and if we have multiple packages with independent versioning in the same repository, it seems like tagging is no longer a viable option for version management (unless we prefix tags, etc.).

2.) It's a good point that we can move things in and out of monorepo if we want to, so there is some freedom to experiment and change our minds. But I think if we're going to contemplate a mix of approaches, we perhaps need to come up with a test to determine which approach is appropriate for any given potential component. (For example, something that is extremely likely to be reused outside of VuFind itself might merit its own separate repo for ease of forking/reuse/lightweight cloning... but something that's more special purpose might "incubate" as a monorepo package and then move out in the future).

3.) On the subject of naming and namespaces, the reason we have namespaces like `VuFindCode` and `VuFindHttp` in some of our offshoot repositories is that originally, we were using Zend Framework modules for everything, and they imposed restrictions on our naming. As autoloading became smarter and we moved away from encapsulating stand-alone utility code in framework-specific modules, this requirement loosened -- so, for example, https://github.com/vufind-org/vufinddate just uses a plain `VuFind\Date` namespace. I think I generally prefer not creating extra top-level namespaces if we don't have to, but there may be reason to do so in some situations.

4.) In general, I agree that having some parallel between vufind and laminas packages could make sense (and I agree with @EreMaijala that I prefer hyphenated names if we go that route). However, I could envision situations where we might want to be more granular, particularly with things like view helpers, where we cover an awful lot of ground within a single namespace. In particular, I've long thought about trying to break out the VuFindTheme code into a separate package, since I think our theme system could have value for other applications... but I know there are currently some interdependencies that would need to be broken out to make that work, and I've never had time to put in the effort. Anyway, my point is that if we wanted to package up our theme code, making that its own package rather than mixing in a lot of general purpose and/or VuFind-specific view helpers might be preferable for clarity.

5.) In terms of continuous integration, you make a good point that there's nothing stopping us from creating a number of CI jobs that all clone the same monorepo but run different tasks on it, so we could continue to have separate CI for each sub-project. This does have some disadvantages, though, like greater disk utilization (a duplicate copy of the same large Git repo for each workspace) and muddled commit history (since CI will show all commits for all parts of the mono-repo mixed together) and lots of duplicated work (since I don't think it would be easy to trigger a build only for commits that change particular parts of the file tree, so we'd end up running CI on everything any time an individual component changed). There may be workarounds for some or all of these concerns, but not without at least requiring some time investment and possibly increasing complexity of setup.

6.) At some point, we're obviously going to need to figure out a mechanism for making a decision about this. Perhaps an item for a future Community Call and/or PMC agenda.

aleksip:
@demiankatz 

1. Indeed, separate versioning would probably have to use tag prefixes or some other solution. A common monorepo convention is to use the same versioning for all packages and always release everything, even if nothing has changed in a specific package. While this can be considered odd, it does it make it easier to know which versions of different packages work together.

2. Using a tool like monorepo-builder makes it easy to have a separate read-only repository for everything, even if all development is done in the monorepo.

3. I agree that it would be nice to have everything under the `VuFind` namespace, including all the modules.

4. I also prefer hyphenated names. Regarding the theme system, I think it would be great to have separate theme package. My initial thoughts would be to have separate vufind-view and vufind-theme packages, with vufind-view being a dependency of vufind-theme. In this PR I have separated the theme specific `HeadLink` and `HeadScript` code from the asset pipeline code, so that the pipeline can be used separately as a part of vufind-view.

5. The monorepo-split-github-action I have used in this PR actually already does the hard part! The monorepo can have its own workflows and the packages can have their own, package specific workflows. I have now added a demo workflow to the vufindview package. I have tested the workflows in a separate [vufind-monorepo](https://github.com/aleksip/vufind-monorepo) clone publishing to a [vufindview](https://github.com/aleksip/vufindview) repo. The package repo only contains package specific commits (and thus should be very small), and the package specific workflow runs when monorepo-split-github-action pushes a new version.

6. Sounds good!

We are actually already using the asset pipeline in the development version of our administration interface, and it seems to be working very well! As mentioned above, I had to make some changes to the current classes to split out the theme specific code. Regardless of what the decision regarding new packages is, it would be nice to have the separated functionality committed to the current codebase, so that even without separate packages, the classes could be copied over as-is.

If that makes sense, I could change the namespace of my "new" asset pipeline classes to `VuFind\View\Helper`, place them in the main VuFind module, and create a separate PR for that.

demiankatz:
@aleksip, some responses on your points above:

1/2. Regarding versioning, it does feel to me like doing development in a single shared repository and then publishing to separate read-only repositories might offer a best-of-all-worlds solution. This lets us manage all the code in a single place, but distribute it in a lighter-weight way... and I'm assuming we could maintain version tags for components in their own repositories, rather than making a mess in the monorepo. I feel strongly that we should not have shared version numbers -- while I understand the convenience of it in this scenario, it really messes up semantic versioning, which I think is very useful to maintain.

3. If we do decide at some point to move some of the currently-external projects back into the core as a monorepo, that might be an opportunity to mint major new versions and change namespacing for clarity. I feel like it might make sense to leave VuFindHarvest with its own top-level namespace since it's essentially its own stand-alone application (which might also be a reason _not_ to monorepo that one), but most of the smaller pieces like VuFindCode and VuFindHttp might benefit from re-namespacing. (Of course, VuFindHttp might become entirely obsolete at some point if we replace Laminas\Http with a more modern, PSR-compliant client... but that's another discussion for another day).

4. This sounds reasonable to me.

5. I think you're focused on GitHub Actions CI while I'm focused on Jenkins CI, which is why some of my previously-cited concerns might not have made sense. In any case, I think the "publish to separate read-only repository" strategy solves my Jenkins problems, if we decide to go that route.

And yes, I agree that it makes sense to separate the refactoring work you've done from the monorepo work. This discussion is probably still going to take a while to resolve, and we shouldn't hold up what you actually need in the short term with that longer-term planning. Ideally, for readability and clarity, the monorepo PR should just move files around without a lot of changes (apart from the monorepo-specific configuration, of course).

And one last question is whether it makes sense to target this work against the dev branch for the 8.1 release, or the dev-9.0 branch for the 9.0 release. Depending on how potentially disruptive the refactoring is, it could really go either way. If it doesn't change anything for upstream subclasses, then it's probably fine for 8.1... but if there are any signature or name changes, 9.0 would be more appropriate.

aleksip:
@demiankatz Very good point re messing up semantic versioning, totally agree about its usefulness! I think your suggestion of using repository specific tags should work. monorepo-builder also seems very flexible, so if we decide to go that route we could even write our own custom release workers for it (or contribute upstream).

I will open another PR with the short term changes and continue discussion about the target branch there!

demiankatz:
Thanks, @aleksip, glad that we seem to be getting on the same page. :-) We can focus on the refactoring as the next step, but once we're ready to return here, maybe the next step is to do a proof of concept to see if we really can publish the new code to a read-only repo and tag things there... unless you've already done that experimentation in your own workspace.

packages/vufindview/src/Helper/HeadScript.php
```diff
@@ -0,0 +1,207 @@
+<?php
+/**
+ * Head script view helper (extended for VuFind's theme system)
+ *
+ * PHP version 7
```
demiankatz, line 5: I'm currently going through old PRs and making sure that none of them introduce outdated "PHP version" comments into the dev branch. If we want to move forward with this work, we need to remember to update this and any other comments listing version 7.

demiankatz (changes requested):
@aleksip, this PR has been in a holding pattern for a long time. Is there any value in moving it forward, or should we close it for now?

As previously discussed, I'm certainly not opposed to investigating the monorepo model, but if neither of us has time to move this forward at the moment, maybe we should consider closing this for the moment.

If we do want to move forward, I've just put an important reminder below.

aleksip:
@demiankatz Closing this as we have abandoned the shared code approach and are now building on top of VuFind instead. Should be much easier and hopefully means even more contributions to VuFind in the future. 🙂

demiankatz:
Thanks, @aleksip!

PR #2273, opened 2022-01-18
Show cover for non-existent records

xmorave2:
When loading cover for records that are not indexed in solr (usually in user account, where data comes from ILS driver and could not be in index), the cover is never load and spinner stay visible.

I've changed the missing record handling to show format-other fallback icon.

### DIFF
--- a/module/VuFind/src/VuFind/RecordDriver/Missing.php
+++ b/module/VuFind/src/VuFind/RecordDriver/Missing.php
@@ -103,4 +103,14 @@ class Missing extends DefaultRecord
         $title = parent::getShortTitle();
         return empty($title) ? $this->determineMissingTitle() : $title;
     }
+
+    /**
+     * Get an array of all the formats associated with the record.
+     *
+     * @return array
+     */
+    public function getFormats()
+    {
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/RecordDriver/Missing.php
```diff
@@ -103,4 +103,14 @@ public function getTitle()
...
     }
+
+    /**
+     * Get an array of all the formats associated with the record.
+     *
+     * @return array
+     */
+    public function getFormats()
+    {
+        return ['other'];
```
demiankatz: As a side effect, I think this is going to cause an untranslated "other" text to appear next to placeholder entries for missing records in lists. Would it be better to use 'Unknown' here to leverage existing translations, or does that not work in other ways?

xmorave2: Does make sense, I've changed it to "Unknown"

demiankatz (approved):
Looks great -- I've confirmed that this works as expected in my test environment. Thanks!

PR #2274, opened 2022-01-18
FIX: vestigial fa-spinners

crhallberg:


themes/bootstrap3/templates/cart/cart.phtml
```diff
@@ -69,7 +69,7 @@
       submitFormWithButton(this, 'empty');
   });
   $("#cart-refuse-empty").click(function($e) {
-      $('.fa.fa-spinner').remove();
+      $('.loading-spinner').remove();
```
demiankatz: I couldn't figure out what circumstances required this code to exist, so I wasn't sure how to test it. Do you remember?

themes/bootstrap3/templates/Helpers/email-form-fields.phtml
```diff
@@ -40,7 +40,7 @@
 </div>
 <?php $emailFormScript = <<<'JS'
     $('#email_to, #email_from, #email_subject').on('invalid', function($e) {
-        $('#modal .fa-spinner').remove();
+        $('.modal-loading-overlay,.loading-spinner', '#modal').remove();
```
demiankatz: I'm also not sure where this applies -- at what point is the spinner added that this is removing?

demiankatz (commented):
Thanks, @crhallberg! The one piece of this that I could test looks good... but I couldn't figure out how to test the other two. Wondering if you had better luck. Pretty sure it's all fine, but I'm wondering if there's some useless code here that is no longer needed.

themes/bootstrap3/templates/cart/cart.phtml
crhallberg: It looks like rejecting the confirmation to empty the cart.

themes/bootstrap3/templates/Helpers/email-form-fields.phtml
crhallberg: When you submit a form in the lightbox, it switches to loading. If the form is invalid, this ensures that the loading indicators are cleared.

demiankatz: My question, though, is if the form is invalid, how does it ever get submitted in the first place? This event seems to trigger when you enter an invalid email address in the form, which prevents the form from submitting... so there's never a loading icon.

themes/bootstrap3/templates/cart/cart.phtml
demiankatz: That's correct... but bringing up the confirmation doesn't add a loading icon, so I'm not sure what its purpose is. (I did confirm that the code triggers, I just couldn't figure out when it would trigger with a loading icon on the screen).

themes/bootstrap3/templates/Helpers/email-form-fields.phtml
crhallberg: Yeah, you know I looked closer at both of these cases and they don't apply. I think it might be a mix of out of date and referencing a customization that wasn't contributed.

demiankatz:
Thanks, @crhallberg, I think it is safe to remove that code, but for a double-check, I'm asking for a second opinion from @xmorave2, who most recently touched some of this code when refactoring Javascript code out of event attributes a couple of years ago.

xmorave2 (approved):
I did test it, and could confirm, that there are no spinners to hide in that two occurrences so they can be removed.

demiankatz:
Thanks, @crhallberg and @xmorave2!

PR #2275, opened 2022-01-19
Add UUID as new identifier used in covers

xmorave2:
TODO
- [ ] Document new Solr field in wiki when merging.

### DIFF
--- a/module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
+++ b/module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
@@ -1316,7 +1316,9 @@ class DefaultRecord extends AbstractBase
     }
 
     /**
-     * Get UUIDs (Universally unique identifier).
+     * Get UUIDs (Universally unique identifier). These are commonly used in, for
+     * example, digital library or repository systems and can be a useful match
+     * point with third party systems.
      *
      * @return array
      */
--- a/module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
+++ b/module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
@@ -1320,7 +1320,7 @@ class DefaultRecord extends AbstractBase
      *
      * @return array
      */
-    public function getUuid()
+    public function getUuids()
     {
         return (array)($this->fields['uuid_str_mv'] ?? []);
     }
@@ -1333,8 +1333,8 @@ class DefaultRecord extends AbstractBase
      */
     public function getCleanUuid()
     {
-        $uuid = $this->getUuid();
-        return empty($uuid) ? false : $uuid[0];
+        $uuids = $this->getUuids();
+        return empty($uuids) ? false : $uuids[0];
     }
 
     /**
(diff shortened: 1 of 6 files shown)


module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
```diff
@@ -1311,6 +1315,28 @@ public function getUPC()
         return (array)($this->fields['upc_str_mv'] ?? []);
     }
 
+    /**
+     * Get UUIDs (Universally unique identifier).
+     *
+     * @return array
+     */
+    public function getUuid()
```
demiankatz: Should this be getUuids since it may return more than one?

demiankatz (commented):
Thanks, @xmorave2 -- I just have one small comment below.

Beyond that, a bigger question: are the UUIDs referenced here something created specifically for ObalkyKnih, or do they come from some other source? I'm wondering if just calling them UUIDs is too broad, if they're UUIDs that were generated for a particular purpose or by a particular standards body. What happens if you have multiple cover providers that accept UUIDs, but each one expects a different UUID for any given record? Bottom line: I'm wondering if there is a naming issue here, but maybe a better understanding of the use case will reduce my concerns.

xmorave2:
The UUIDs use in ObalkyKnih do come from external systems - usually digital libraries. As far as I know, they should be 'real' UUIDs, not one-purpose identifiers. In fact UUID should be only one for the record. I used multi valued field only for one reason - multiple uuids could be added to record by accident/mistake. 

Also the problem you've described could happen with ISBNs too - for example some books have more ISBN (published from more than one publishers, bad isbn later fixed...). Then we can't know which use which ISBN and we just send the first one.

module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
xmorave2: I changed it

demiankatz:
Thanks, @xmorave2, that makes sense. Maybe it's worth adding a little more to the comment on the getUuids() function to clarify this. For example, something like: `Get UUIDs (Universally unique identifier). These are commonly used in, for example, digital library or repository systems and can be a useful match point with third party systems.` Maybe that's not the best place for that comment, but if we're going to add a new data field, it would be helpful to have an explanation somewhere. If you don't like the record driver method as the place for this type of comment, maybe a better option would be to add a commented-out example in marc_local.properties showing an example of how to index UUIDs and an explanation of why you might want to. Also, after merging this, I should add a note to the Solr schema changelog in the wiki -- I'll put a TODO checkbox in the top comment to remind myself.

In any case, apart from figuring out this documentation issue, I think we're all set now. Thanks!

xmorave2:
Thanks @demiankatz. I've added a comment to record driver and also an example into marc_local.properties.

demiankatz:
Thanks, @xmorave2! I commented out the new setting in marc_local.properties, since anything in there is intended as an example and should not be active unless uncommented by the end user. If you think it might be valuable to have this on by default, though, we could discuss moving it to marc.properties as an active default setting -- I don't think it adds too much overhead. Otherwise, I'll merge this as-is!

xmorave2:
thanks @demiankatz I intended to comment it out, but forgot at the end. So I am ok with it as it  is now.

demiankatz (approved):

demiankatz:
Thanks again!

PR #2278, opened 2022-01-20
Add allow-plugins to composer.json for new composer versions.

EreMaijala:
See https://getcomposer.org/allow-plugins for details on the configuration.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2282, opened 2022-01-21
log exceptions in VuFind\AuthManager to VuFind logger

xrosecky:
Log exceptions in VuFind\AuthManager to VuFind logger instead to PHP error_log + exception chaining for AuthException

### DIFF
--- a/module/VuFind/src/VuFind/Auth/Manager.php
+++ b/module/VuFind/src/VuFind/Auth/Manager.php
@@ -44,8 +44,11 @@ use VuFind\Validator\CsrfInterface;
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org Main Page
  */
-class Manager implements \LmcRbacMvc\Identity\IdentityProviderInterface
+class Manager implements \LmcRbacMvc\Identity\IdentityProviderInterface,
+    \Laminas\Log\LoggerAwareInterface
 {
+    use \VuFind\Log\LoggerAwareTrait;
+
     /**
      * Authentication modules
      *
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -649,6 +652,10 @@ public function login($request)
         ) {
             if (!$this->csrf->isValid($request->getPost()->get('csrf'))) {
                 $this->getAuth()->resetState();
+                $this->logError(
```
demiankatz: I wonder if this should be logged at a different level than "error" -- to use an HTTP analogy, this is a 400 error, while the other exceptions are 500 errors -- one is related to misconfiguration or system failure, which requires urgent attention. Bad CSRF tokens are probably the result of external user actions. It might not be desirable to flood the error logs if, for example, somebody does some sort of bot-net attack sending invalid submissions to VuFind forms. What do you think?

xrosecky: Yes, it makes sense, I changed it to logWarning.

EreMaijala:
@xrosecky I believe you can remove at least the `get_class()` part from the log message since the logger already adds the class name. So something like 'Invalid CSRF token passed to login' should work.

module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -668,11 +675,12 @@ public function login($request)
...
             // difficulties
-            error_log(
-                "Exception in " . get_class($this) . "::login: " . $e->getMessage()
+            $this->logError(
+                "Exception in " . get_class($this) . "::login: "
+                . $e->getMessage()
             );
-            error_log($e);
-            throw new AuthException('authentication_error_technical');
+            $this->logError($e);
```
demiankatz: Another observation: `$e` is an exception object in this instance, and `logError` expects a string. This appears to be a pre-existing problem, and maybe automatic typecasting makes it work right, but I think it might be better to explicitly say `$e->getMessage()` here.

xrosecky: Isn't better `(string) $e` because it also logs the stack trace? The error message is logged in the line above `$this->logError($e)`.

demiankatz: Yes, that is much better, thank you!

demiankatz (approved):

PR #2283, opened 2022-01-21
Split asset pipeline functionality to separate classes

aleksip:
Moved to this PR from #2272.

This can be considered a breaking change if someone is using `\VuFindTheme\View\Helper\ConcatTrait` in a way that expects it to provide the methods moved to `\VuFind\View\Helper\ConcatTrait`. Not sure how probable this actually is though. In any case, happy to target 9.0.

Includes a minor refactoring (adds a `getResourceFileFilesystemPath()`), and fixes two minor bugs (now works correctly with a `true` config value and does not add empty nonce attributes).

I guess the nonce functionality should be split too, so it could be used separately. Can have a go here or in a follow-up PR.

I wondered about moving these to `\VuFind\View\Helper\Root\`, but if we are to place these in a separate package, that namespace seemed implementation specific.


demiankatz:
Thanks, @aleksip -- I haven't done an exhaustive review yet to double-check all the details, but first, a few general thoughts and replies to your comments:

1.) It probably is best to target 9.0 because of the trait issue you mention. As you say, it's not especially likely, but it does seem like a backward-breaking change, so we should probably treat it as such. I have to say that it annoys me that you can't extend or nest traits -- it would make things like this a lot cleaner and easier to deal with.

2.) I agree that it makes sense NOT to put these in the `VuFind\View\Helper\Root` namespace; honestly, in retrospect, I'm not especially happy with the theme-specific namespacing strategy I chose a decade ago. If we're creating a package of general-purpose helpers, moving them out of a theme-specific namespace seems like a step in the right direction.

3.) Would it be feasible/practical/helpful to add some unit tests? If we're going to split this to another package, I'd like to have decent test coverage on it, so maybe now is the time to work on that. I also wonder if designing tests answers your question about whether to split the nonce logic now or later -- would splitting now make testing easier or more logical?

I'll take a closer look once we've finished this conversation. Let me know if I can help with anything else in the meantime!

aleksip:
@demiankatz My work on this has stalled a bit due to a hard deadline with other related work, but this is still very much in the pipeline! (We are actually going to deploy copies of these classes to production this week.)

I don't think adding tests is ever a bad idea 🙂 so yes, we should add them, especially as this will need a different setup from the monolithic tests (there are some test setup related changes in #2272 already).

In case you already had specific things for tests in mind, listing them would be a big help, but I can definitely come up with something on my own too.

demiankatz:
Thanks, @aleksip, I didn't look deeply enough to have specific ideas for tests, but if it would be helpful, I can review in more depth next week when I'm back from my vacation and brainstorm some suggestions.

module/VuFind/src/VuFind/View/Helper/PipelineInjectorFactory.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options sent to factory.');
+        }
+        if (!defined('APPLICATION_ENV')) {
```
demiankatz, line 63: If we're looking to make this more generic/reusable, should we loosen this constant requirement, and just assume we're not in development mode if the constant is unset?

module/VuFindTheme/src/VuFindTheme/View/Helper/ConcatTrait.php
```diff
@@ -46,191 +46,20 @@
...
-     * Returns true if file should not be included in the compressed concat file
+     * Get the filesystem path for the provided resource file path
      *
-     * @param stdClass $item Element object
+     * @param string $path Path
      *
-     * @return bool
+     * @return ?string
      */
-    abstract protected function isExcludedFromConcat($item);
```
demiankatz, line 57: Since we're deleting so much code from here, I think we should probably update the file/trait-level comments above to indicate that this trait now contains overrides specific to the VuFindTheme module and assumes that it will be mixed into classes extending more generic versions of the helpers.

module/VuFind/src/VuFind/View/Helper/ConcatTrait.php
```diff
@@ -0,0 +1,467 @@
...
+    /**
+     * Get the filesystem path for the provided resource file path
+     *
+     * @param string $path Path
+     *
+     * @return ?string
+     */
+    protected function getResourceFileFilesystemPath(string $path): ?string
+    {
+        if (!defined('APPLICATION_PATH')) {
```
demiankatz, line 170: Should we think about a way to make this more generic? Maybe add a getApplicationPath method with this as a default implementation?

module/VuFind/src/VuFind/View/Helper/ConcatTrait.php
```diff
@@ -0,0 +1,467 @@
...
+
+    /**
+     * Get the path to the directory where we can cache files generated by
+     * this trait.
+     *
+     * @return string
+     */
+    protected function getResourceCacheDir()
+    {
+        if (!defined('LOCAL_CACHE_DIR')) {
```
demiankatz, line 243: Another constant that perhaps should be made more generic.

demiankatz (commented):
Sorry for the slow response, @aleksip, but I finally had time to take a closer look at this, and a few minor comments came to mind.

As far as testing goes, I have no specific suggestions beyond trying to systematically test all the helpers. I do wonder if this might be easier to work with if we could break it into smaller parts -- for example, I wonder if any of the trait logic might be refactored to support classes -- but I didn't spend enough time studying the code to know whether or not that's a good idea. Just occurred to me as a possibility to think about.

demiankatz:
@aleksip, just checking on the status of this work -- is this still on your list somewhere? Do you have time to resolve the conflicts that have been introduced here? Do you need my help with anything?

demiankatz:
@aleksip, while we're touching base on statuses, where does this one stand? As always, happy to leave it open if you're going to need it / get back to it, but if it's abandoned, maybe we should close it.

module/VuFind/src/VuFind/View/Helper/HeadLink.php
```diff
@@ -0,0 +1,204 @@
+<?php
+
+/**
+ * Head link view helper (extended for VuFind's theme system)
+ *
+ * PHP version 7
```
demiankatz, line 6: This is another PR that needs PHP version updates if we're going to move forward with it.

demiankatz (changes requested):
@aleksip, here's another old PR that we should probably either update or close, depending on whether further progress is likely to be made.

aleksip:
Seems best to close this due to [VUFIND-1677](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1677). Will try experimenting with the monorepo model using some other classes (there are plenty of options to choose from), and hopefully the new asset pipeline view helpers can be implemented in a decoupled way.

PR #2284, opened 2022-01-21
Add test for MARC processing in Unicorn driver.

demiankatz:
This is a quick-and-dirty effort to add some test coverage for MARC processing in the Unicorn driver, so we can validate the refactoring in #2252 without the need for a running Unicorn system.

I'm not convinced that there's need to maintain the Unicorn driver -- it hasn't been touched in a decade, and I suspect no one is using it in production -- so while this test is pretty crude and minimal, I think it's probably an adequate good faith effort.


EreMaijala (approved):
Looks good to me. Thanks!

PR #2285, opened 2022-01-21
Add test for Voyager MARC processing.

demiankatz:
This is further work in support of #2252 -- this uses real MARC data pulled from a running instance of the Voyager driver as the basis for a new test. I believe this is sufficient to cover the refactored code.


EreMaijala (approved):
Thanks! Looks good to me.

PR #2286, opened 2022-01-24
Improve Unicorn MARC test.

demiankatz:
A quick look at the code coverage in Jenkins revealed a couple of details that I wasn't testing before, so I've added a bit more data to cover more edges. I don't think this will get us to 100%, but it at least should hit all the remaining File_MARC logic.


EreMaijala (approved):
Looks good and works for me.

PR #2289, opened 2022-01-25
Add ability to configure problem information showing to user when request fails

xmorave2:


module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
```diff
@@ -339,6 +348,13 @@ public function init()
                 );
         }
         $this->maxNumberOfPages = $this->config['Catalog']['maxNumberOfPages'] ?? 0;
+        if (isset($this->config['Catalog']['holdProblemsDisplay'])) {
+            $this->holdProblemsDisplay
+                = explode(
+                    ',',
+                    $this->config['Catalog']['holdProblemsDisplay']
```
demiankatz, line 359: Is it worth wrapping an `array_map('trim', ...)` around this in case the user puts spaces in the comma-separated list?

demiankatz (commented):
Makes sense to me -- just one small suggestion.

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
xmorave2: That's true, I've added it

demiankatz (approved):

demiankatz:
Thanks, @xmorave2!

PR #2290, opened 2022-01-26
Add ability to check ObalkyKnih servers are alive

xmorave2:


module/VuFind/src/VuFind/Content/ObalkyKnihService.php
```diff
@@ -198,4 +226,41 @@ protected function createLocalIdentifier(string $recordid): ?string
...
+     * Get currently available base URL
+     *
+     * @return string
+     */
+    protected function getBaseUrl(): string
+    {
+        if ($this->checkServersAvailability) {
+            return $this->getAliveUrl();
+        }
+        return '';
```
demiankatz: Maybe I'm misreading something, but shouldn't this return the first base URL from the configuration by default, rather than failing if checkServersAvailability is disabled?

Also, you might consider making this a ternary statement for brevity (though that's just a matter of opinion :-) ).

demiankatz (commented):
Thanks for this! One question about what looks like it might be a bug...

module/VuFind/src/VuFind/Content/ObalkyKnihService.php
xmorave2: Thanks for catching this @demiankatz, that is definitely bug. I'll fix it.

demiankatz (approved):

module/VuFind/src/VuFind/Content/ObalkyKnihService.php
demiankatz: Thanks for the fix -- just merged!

PR #2291, opened 2022-01-26
Update Laminas dependencies (mostly for PHP 8.1 compatibility).

demiankatz:
This pull request updates nearly all of the Laminas packages used by VuFind to latest versions.

A few components cannot be upgraded to the latest without dropping compatibility with PHP 7.3. That will be addressed separately in the dev-9.0 branch once we raise the minimum PHP requirement.

There is a new major version of laminas-cache which contains breaking changes; that should also be addressed as a separate dev-9.0 PR.

TODO
- [x] Review all component changelogs for potential concerns
- [x] Run full test suite


demiankatz:
All tests are passing, so I think this is safe to merge. @EreMaijala, mind giving it a double-check?

EreMaijala (approved):
Thanks for working on this, all looks good to me!

PR #2293, opened 2022-01-28
Adding hold history config example to KohaRest.ini

JacobSandin:
Just adding config settings to the KohaRest.ini to make it easier to find how to enable holds history.


demiankatz:
@JacobSandin, it appears that your BokInfo cover image code ended up in this pull request by accident. I recommend working on separate features in separate branches to avoid this kind of confusion. Let me know if you need any help sorting it out! :-)

PR #2294, opened 2022-01-28
Adding Bokinfo as a source for coverimages.

JacobSandin:
Bokinfo is a website providing book information, such as cover images.

edit 1: Working on fixing the errors

### DIFF
--- a/config/vufind/XCNCIP2.ini
+++ b/config/vufind/XCNCIP2.ini
@@ -74,3 +74,8 @@ consortium  = false
 ; paginate even if we want all data at one time. We then ask for all pages, but it
 ; could possibly lead to long response times. (Default = 0, meaning no limit).
 ;maxNumberOfPages = 10
+
+; Which subelements of Problem element show to user when placing hold fails.
+; Possible values are: 'ProblemType', 'ProblemDetail', 'ProblemElement',
+; 'ProblemValue'
+;holdProblemsDisplay = 'ProblemType,ProblemDetail';
--- a/module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
@@ -153,6 +153,15 @@ class XCNCIP2 extends AbstractBase implements \VuFindHttp\HttpServiceAwareInterf
     protected $disableRenewals = false;
 
     /**
+     * Which subelements of Problem element show to user when placing hold fails.
+     * Possible values are: 'ProblemType', 'ProblemDetail', 'ProblemElement',
+     * 'ProblemValue'
+     *
+     * @var string[]
+     */
+    protected $holdProblemsDisplay = ['ProblemType'];
+
+    /**
      * Schemes preset for certain elements. See implementation profile:
      * http://www.ncip.info/uploads/7/1/4/6/7146749/z39-83-2-2012_ncip.pdf
      *
@@ -332,13 +341,25 @@ class XCNCIP2 extends AbstractBase implements \VuFindHttp\HttpServiceAwareInterf
         }
 
         if (isset($this->config['Catalog']['otherAcceptedHttpStatusCodes'])) {
-            $this->otherAcceptedHttpStatusCodes
+            $otherAcceptedHttpStatusCodes
                 = explode(
                     ',',
                     $this->config['Catalog']['otherAcceptedHttpStatusCodes']
                 );
+            $this->otherAcceptedHttpStatusCodes = array_map(
+                'trim',
+                $otherAcceptedHttpStatusCodes
+            );
         }
         $this->maxNumberOfPages = $this->config['Catalog']['maxNumberOfPages'] ?? 0;
+        if (isset($this->config['Catalog']['holdProblemsDisplay'])) {
+            $holdProblemsDisplay
+                = explode(
+                    ',',
+                    $this->config['Catalog']['holdProblemsDisplay']
+                );
+            $this->holdProblemsDisplay = array_map('trim', $holdProblemsDisplay);
+        }
     }
 
     /**
@@ -1814,7 +1835,7 @@ class XCNCIP2 extends AbstractBase implements \VuFindHttp\HttpServiceAwareInterf
             $failureReturn = ['success' => false];
             $problemDescription = $this->getProblemDescription(
                 $response,
-                ['ProblemType'],
+                $this->holdProblemsDisplay,
                 false
             );
             if (!empty($problemDescription)) {


demiankatz:
Looks like this has a few style issues that need sorting out. Some can be auto-fixed with phpcbf (i.e. run `$VUFIND_HOME/vendor/bin/phing phpcbf`) and others may require a little manual intervention. If you have difficulty sorting any of them out, please let me know and I'll be happy to help!

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,204 @@
...
+     *
+     * @param string $url API Url
+     *
+     * @return HttpClient Http Client
+     */
+    protected function createHttpClient($url)
+    {
+        $client = $this->httpService->createClient($url);
+
+        if (isset($this->config['Http']['ssl_verify_peer_name'])
```
demiankatz: Where is `$this->config` supposed to be coming from? It doesn't seem to be defined or set anywhere, which is why PHPStan is complaining. Let me know if you need help with configuration injection.

Is there a strong reason to put all this configuration logic here? If the ssl_verify_peer_name option is set, does it need to be handled differently by Bokinfo than by everything else?

JacobSandin: I changed so that it just verify and set timeout to 30. It was copied code from KohaRest I think. The debugger did not complain so I thought nothing of it until now when tests failed.

demiankatz: That's the beauty of static analysis! :-) The code worked because it was checking if the value was set before doing anything, so it wasn't doing anything... but PHPStan was smart enough to figure out that the value could NEVER be set, so it gave us a warning.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,200 @@
...
+{
+    use \VuFind\Log\LoggerAwareTrait;
+    use \VuFindHttp\HttpServiceAwareTrait;
+
+    /**
+     * Constructor
+     */
+    public function __construct()
+    {
+        $this->supportsIsbn = $this->cacheAllowed = true;
```
demiankatz: Just to be safe: have you confirmed that the Bokinfo terms of service allow caching?

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,200 @@
...
+
+            $client->getRequest()->getHeaders()
+                  ->addHeaderLine("Ocp-Apim-Subscription-Key", $key);
+
+            try {
+                $resp= $client->send();
+                $body = $resp->getBody();
+                $url = $this->getImageUrl($body);
+            } catch (\Exception $e) {
+            }
```
demiankatz: Do you really want to ignore exceptions completely? Should this be a return false? Or do you not need this try..catch at all because it's already inside a different try..catch?

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,200 @@
...
+
+            if ($this->testUrlFunction($url)) {
+                return "$url";
+            }
+        } catch (\Throwable $ex) {
+            return false;
+        } catch (\Exception $ex) {
+            return false;
+        } catch (\RuntimeException $ex) {
```
demiankatz: I don't think you need all three of these catches -- all exceptions are \Throwable, so that may be enough to cover all situations if the same behavior is expected in all cases.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,200 @@
...
+                'ssl',
+                'verify_peer_name',
+                false
+            );
+            if (!$res) {
+                throw new \Exception('Unable to set sslverifypeername option');
+            }
+        } elseif ($adapter instanceof \Laminas\Http\Client\Adapter\Curl) {
+            $adapter->setCurlOption(CURLOPT_SSL_VERIFYHOST, false);
+        }
```
demiankatz: Why is it necessary to turn off hostname verification? Is there a known problem with this service? I just want to be sure we're not creating global code to work around a local problem.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,200 @@
...
+
+        // Set timeout value
+        $timeout = 30;
```
demiankatz: It might be useful to make the timeout value a class property rather than a local variable -- then it's easier to customize or make configurable in future if a need arises.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,200 @@
...
+     *
+     * @param string $url image Url
+     *
+     * @return bool Http Client
+     */
+    protected function testUrlFunction($url)
+    {
+
+        // Use get_headers() function
+        $headers = @get_headers($url);
```
demiankatz: It would probably be better to do this using the HTTP client object -- since get_headers is using a low-level PHP function, it means you're bypassing all the configuration encapsulated in the HTTP client. This could cause the function to break when running behind a proxy server (to offer just one example).

demiankatz (changes requested):
Thanks for working through all the tests! Here are a few other comments that I hope will be helpful. :-)

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
JacobSandin: No I have not, but I will send a question right away. They have been eager to get plugins for both koha and vufind thoug, so I think so.

demiankatz: Thank you! It's always good to check this point because different vendors have different opinions. Some prefer us to cache locally so we reduce load on their servers; others don't allow caching because of copyright concerns.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
JacobSandin: Check, removing.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
JacobSandin: Check

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
JacobSandin: I did not think it costed that much extra to check the cert. But I will remove it. Should I also remove it from the context above?

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
JacobSandin: Actually, it might not be needed If I leave it out it chould take the \VuFindHttp\HttpServiceAwareInterface default right?

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
JacobSandin: I used get_headers because it is less load than geting the whole body. Is there such a function on the HTTP client that I missed?

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
demiankatz: If there is not a specific reason to turn off hostname verification, I would remove the whole block that's turning it off in different ways depending on adapter settings. I don't think any of that should be strictly necessary.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
demiankatz: Even if you take this out, you'll still need `HttpServiceAwareInterface` -- that interface is a marker that tells the framework to inject the service that creates HTTP clients, and you need that for the `createHttpClient` method to work. Of course, if you prefer not to rely on this kind of "magic," you could also set up a custom factory for the plug-in and inject the client via the factory.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
demiankatz: I think if you get the request object from the client, you can call `setMethod` on the request to do a HEAD request instead of GET or POST.

JacobSandin: It had actually, so forget my question.

JacobSandin:
I hope all seem ok now. Thank you so much for all your comments Demian.

PR #2295, opened 2022-01-28
Override relevance for ID-based searches.

demiankatz:
It was recently brought to my attention that having relevance-based sorting for ID-based searches is meaningless, just as it is for empty searches. This makes the results in course reserves and new items searches potentially confusing. This PR extends the empty search relevance override to also impact ID-based searches.

### DIFF
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -12,7 +12,7 @@ default_sort         = relevance
 ; searches when relevance sort is selected. Since relevance doesn't have a
 ; meaningful function without query terms, this can be set to e.g. "title" or
 ; "year desc".
-empty_search_relevance_override = year desc
+;empty_search_relevance_override = title
 
 ; This setting controls the default view for search results; the selected option
 ; should be one of the options present in the [Views] section below.
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -8,10 +8,11 @@ default_handler      = AllFields    ; Search handler to use if none is specified
 ; below.
 default_sort         = relevance
 
-; This setting controls the sort order to be used for empty search when relevance
-; sort is selected. Since relevance doesn't have a meaningful function with an empty
-; search, this can be set to e.g. "title".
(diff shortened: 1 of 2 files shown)


config/vufind/searches.ini
```diff
@@ -8,10 +8,11 @@ default_handler      = AllFields    ; Search handler to use if none is specified
...
 
-; This setting controls the sort order to be used for empty search when relevance
-; sort is selected. Since relevance doesn't have a meaningful function with an empty
-; search, this can be set to e.g. "title".
-;empty_search_relevance_override = title
+; This setting controls the sort order to be used for empty (or purely ID-based)
+; searches when relevance sort is selected. Since relevance doesn't have a
+; meaningful function without query terms, this can be set to e.g. "title" or
+; "year desc".
+empty_search_relevance_override = year desc
```
EreMaijala: I think this (actually, other sorts as well) should contain `, id desc` as a tie-breaker. Otherwise the order of records that have the same year is undetermined.

EreMaijala (changes requested):

config/vufind/searches.ini
demiankatz: I didn't intend to commit the change from "title" to "year desc" as the default value here -- that was an accident. I've reverted it.

We could certainly change the other example in the comment to "year desc,id desc" if you think it important... though maybe "year desc,title asc" would be more meaningful. I'd think that sorting by ID is just about as arbitrary as sorting by nothing in particular (except that it might survive reindexing with greater consistency).

EreMaijala: Ok. Let's discuss any changes to the sort fields separately as it's not directly related to the matter at hand.

EreMaijala (approved):

config/vufind/searches.ini
demiankatz: I've opened [VUFIND-1529](https://vufind.org/jira/browse/VUFIND-1529) to capture the results of this conversation beyond what got merged here.

PR #2296, opened 2022-01-28
Demo driver: make new/recent IDs configurable.

demiankatz:
While working on #2295, I needed an easy mechanism to test with a specific set of new records, so I created these new settings.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Demo.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Demo.php
@@ -1549,7 +1549,7 @@ class Demo extends AbstractBase
      *
      * @return string[]
      */
-    protected function getRandomBiBIds($limit): array
+    protected function getRandomBibIds($limit): array
     {
         $count = rand(0, $limit > 30 ? 30 : $limit);
         $results = [];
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -1542,6 +1542,28 @@ public function getCourses()
...
     }
 
+    /**
+     * Get a set of random bib IDs
+     *
+     * @param int $limit Maximum number of IDs to return (max 30)
+     *
+     * @return string[]
+     */
+    protected function getRandomBiBIds($limit): array
```
EreMaijala: I suppose the capital B in `BiB` is a typo, right?

EreMaijala (changes requested):

module/VuFind/src/VuFind/ILS/Driver/Demo.php
demiankatz: Ha, yes! Case insensitive function names in PHP can be a nuisance sometimes. Fixed now. :-)

EreMaijala (approved):

PR #2297, opened 2022-01-29
Adding config for history to KohaRest-ini

JacobSandin:
Lets try and get this right. This is just configuration settings for KohaRest so that it is easier to find settings to enable history.

### DIFF
--- a/config/vufind/KohaRest.ini
+++ b/config/vufind/KohaRest.ini
@@ -152,7 +152,7 @@ extraFields = item-issue:acceptTerms:pickUpLocation
 ;pattern = "numeric"
 ;hint = "Your optional custom hint can go here."
 
-; Show holds transaction history, default is set to false.
 [TransactionHistory]
-enabled = false
+; By default the loan history is disabled. Uncomment the following line to enable it.
+;enabled = true
 
--- a/config/vufind/KohaRest.ini
+++ b/config/vufind/KohaRest.ini
@@ -151,3 +151,8 @@ extraFields = item-issue:acceptTerms:pickUpLocation
 ; config.ini defaults when Voyager is used for authentication.
 ;pattern = "numeric"
 ;hint = "Your optional custom hint can go here."
+
+; Show holds transaction history, default is set to false.
+[TransactionHistory]
+enabled = false
+


demiankatz:
Thanks, @JacobSandin -- asking for @EreMaijala's review on this since he's more likely to have an opinion than I am. :-)

config/vufind/KohaRest.ini
```diff
@@ -151,3 +151,8 @@ extraFields = item-issue:acceptTerms:pickUpLocation
 ; config.ini defaults when Voyager is used for authentication.
 ;pattern = "numeric"
 ;hint = "Your optional custom hint can go here."
+
+; Show holds transaction history, default is set to false.
+[TransactionHistory]
+enabled = false
```
EreMaijala, line 157: This is otherwise fine, but should be commented-out by default so that enabling the history is a conscious decision.

EreMaijala (changes requested):

config/vufind/KohaRest.ini
EreMaijala: Err.. oops. Misread it.

EreMaijala (approved):
Looks good to me!

config/vufind/KohaRest.ini
```diff
@@ -151,3 +151,8 @@ extraFields = item-issue:acceptTerms:pickUpLocation
 ; config.ini defaults when Voyager is used for authentication.
 ;pattern = "numeric"
 ;hint = "Your optional custom hint can go here."
+
+; Show holds transaction history, default is set to false.
```
EreMaijala: Sorry, changing approval. The comment should be revised. The comment above the section should be about the full section which might contain other settings as well. Then right above the `enabled = true` you'd have something like `; Whether the transaction history functioanlity is enabled. Default is false.`. This does not involve holds.

EreMaijala (changes requested):

config/vufind/KohaRest.ini
JacobSandin: I have copied the section from demo.ini so I hope it is ok.

EreMaijala (approved):

PR #2298, opened 2022-01-29
Adding Bokinfo coverimages support. (Reworked)

JacobSandin:
Support for using https://www.bokinfo.se as a source for coverimages.


module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,173 @@
+<?php
+
+/**
+ * VuFind Driver for Koha, using REST API
```
demiankatz: This comment isn't accurate.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * PHP version 7
+ *
+ * Copyright (C) The National Library of Finland 2016-2020.
```
demiankatz: These copyright notices probably aren't right either.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,173 @@
...
+ * @author   Jacob Sandin <jacob@imcode.com>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+namespace VuFind\Content\Covers;
+
+use SimpleXMLElement;
+
+/**
+ * Summon cover content loader.
```
demiankatz: This comment should also be adjusted to match whatever text you choose above.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * @category VuFind
+ * @package  Content
+ * @author   Demian Katz <demian.katz@villanova.edu>
```
demiankatz: Feel free to give yourself author credit.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,173 @@
...
+{
+    use \VuFind\Log\LoggerAwareTrait;
+    use \VuFindHttp\HttpServiceAwareTrait;
+
+    /**
+     * Constructor
+     */
+    public function __construct()
+    {
+        //I will leave this untill I know if they allow it.
```
demiankatz: You could reformat this to:

`// TODO: confirm whether Bokinfo allows caching before uncommenting`

(Which makes it a little more clear what the next step is here).

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,173 @@
...
+     * @return bool Http Client
+     */
+    protected function testUrlFunction($url)
+    {
+        try {
+            $client = $this->createHttpClient(
+                "$url"
+            );
+
+            $resp= $client->send();
```
demiankatz: Were you going to set the method to HEAD here, or did you find that it is not necessary/doesn't work?

module/VuFind/src/VuFind/Content/Covers/PluginManager.php
```diff
@@ -61,6 +61,7 @@ class PluginManager extends \VuFind\ServiceManager\AbstractPluginManager
         'orb' => Orb::class,
         'summon' => Summon::class,
         'syndetics' => Syndetics::class,
+        'bokinfo' => Bokinfo::class,
```
demiankatz: I try to keep the configuration lists in alphabetical order; do you mind moving the factory and alias lines up?

demiankatz (changes requested):
Thanks for all the progress -- just a few more questions and comments on the latest revision!

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
JacobSandin: It actually worked fine without, so therefore I removed it.

module/VuFind/src/VuFind/Content/Covers/PluginManager.php
JacobSandin: Do not mind at all.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
JacobSandin: They confirmed that it was ok.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,172 @@
...
+ * @author   Jacob Sandin <jacob@imcode.com>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+namespace VuFind\Content\Covers;
+
+use SimpleXMLElement;
+
+/**
+ * Summon cover Bokinfo content loader.
```
demiankatz: This mentions Summon, which is not correct; I think you should just copy the "Plugin for Bokinfo coverimages" line from above for consistency.

(I would have done this myself to save you the trouble, but it looks like I don't have write access to this fork).

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,172 @@
...
+     * pointing to an ISBN object and 'issn' pointing to a string)
+     *
+     * @return string|bool
+     */
+    public function getUrl($key, $size, $ids)
+    {
+        if (!isset($ids['isbn'])) {
+            return false;
+        }
+        if (!isset($key)) {
```
demiankatz: Maybe this should be `if (empty($key))` rather than `if (!isset($key))`. By virtue of being a required function parameter, `$key` is always going to be set to something... but if it's an empty string, that's going to be a problem. :-)

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
```diff
@@ -0,0 +1,172 @@
...
+            );
+
+            $client->getRequest()->getHeaders()
+                  ->addHeaderLine("Ocp-Apim-Subscription-Key", $key);
+
+            $resp = $client->send();
+            $body = $resp->getBody();
+            $url = $this->getImageUrl($body);
+            if ($this->testUrlFunction($url)) {
+                return "$url";
```
demiankatz: You don't need the double quotes here if you're just returning the variable unmodified -- they serve no purpose in this context.

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
demiankatz: I'm curious, though, if this is causing VuFind to effectively download every image twice -- once to check if it's valid, and again to cache/display it. Is there a reason that you have to do this? Are you unable to simply trust the results of the API call?

demiankatz (commented):
Thanks -- we're getting close to finishing this (though if you want to be really thorough, a unit test would be great... but that can wait if it's an unreasonable hurdle at this stage). Just a few more minor comments below...

module/VuFind/src/VuFind/Content/Covers/Bokinfo.php
JacobSandin: I do not mind learning to get things right. However, how would I give that access? I guess it gets removed on  a fork?

demiankatz: I received your invitation, so I see you have figured this out. :-) I believe there is also a checkbox when you create a pull request that lets you give permission to project owners to modify your PR branch without giving access to your entire repository, though there may be repository-level settings that impact whether or not the option appears. In any case, I think we're in good shape now; thanks!

demiankatz:
Thanks, @JacobSandin! I've added a comment to the class indicating that we checked about cache permissions (just in case we forget in the future, I wanted to capture the fact that we had done this).

I think this can be merged, though if you'd like me to add a unit test to show you how that's done, I'd be happy to help if you can provide an example XML response from the service to use as a test fixture.

demiankatz:
Thanks for sharing test data, @JacobSandin! I've added a test and made a few very small edits. Please review my work, and let me know if you have any questions. If you are happy with this, I believe it is ready to be merged!

JacobSandin:
I is all good for me @demiankatz, will take me a bit longer to understand the testing and mocking. However I will be happy to get the code added.

demiankatz (approved):

demiankatz:
Thanks, @JacobSandin -- merging now!

PR #2299, opened 2022-01-31
Upgrade to Laminas\Cache v3.

demiankatz:
This PR upgrades to the new major version of Laminas\Cache and adds/modernizes necessary configurations to make it work.

TODO
- [x] Run full test suite


EreMaijala (approved):
I find it a bit odd that an Adapter is a complete module, but works for me.

demiankatz:
Thanks, @EreMaijala. I agree that "one module per adapter" seems a bit strange, but given the ecosystem, I can understand why it's necessary. It makes sense to separate the adapters from the core project, because it reduces/distributes maintenance burden and makes it easier to deprecate obsolete drivers... but if you split up the projects, you need a way to inject the adapters into the adapter plugin manager... and that's what the modules are doing: each module adds another delegator factory to the adapter plugin manager factory, adding configuration for the relevant adapter.

PR #2300, opened 2022-02-01
Translate subject email with translate function

LuomaJuha:
If @EreMaijala  wants to say something about this.

### DIFF
--- a/module/VuFind/src/VuFind/Form/Form.php
+++ b/module/VuFind/src/VuFind/Form/Form.php
@@ -303,11 +303,7 @@ class Form extends \Laminas\Form\Form implements
             $translated["%%{$key}%%"] = implode(', ', $translatedVals);
         }
 
-        return str_replace(
-            array_keys($translated),
-            array_values($translated),
-            $subject
-        );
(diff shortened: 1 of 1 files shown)


EreMaijala (changes requested):
This breaks all placeholders.

demiankatz:
Doesn't passing the tokens to the translate function address that? Or am I missing something?

Sounds like test coverage would help clarify matters, if we don't already have it. (But if we do have it and the tests are passing, that would seem to me like a good sign).

EreMaijala (approved):
Err.. yes. I don't know how I misread this. Sorry!

demiankatz (approved):

demiankatz:
I'm going to go ahead and merge this now -- it actually seems like a bug fix since the comments in the .yaml file indicate that email subjects are translated, but the actual behavior was not doing that previously.

It does seem like it might be valuable to add a test case that covers the placeholder functionality of email subjects, since that doesn't seem to be addressed in the current test suite... but I won't hold this up on that point alone. Merging now!

PR #2301, opened 2022-02-01
Refactor constants/Laminas app setup to separate files.

demiankatz:
TODO
- [x] Confirm that profiling still works
- [x] Run full test suite

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -1,0 +1,16 @@
+<?php
+require __DIR__ . '/constants.config.php';
+
+chdir(APPLICATION_PATH);
+
+// Composer autoloading
+if (file_exists('vendor/autoload.php')) {
+    $loader = include 'vendor/autoload.php';
+}
+
+if (!class_exists('Laminas\Loader\AutoloaderFactory')) {
+    throw new RuntimeException('Unable to load Laminas autoloader.');
+}
+
+// Return the application!
+return Laminas\Mvc\Application::init(require 'config/application.config.php');
--- a/config/constants.config.php
+++ b/config/constants.config.php
@@ -34,6 +34,5 @@ defined('LOCAL_CACHE_DIR')
             : (strlen(LOCAL_OVERRIDE_DIR) > 0 ? LOCAL_OVERRIDE_DIR . '/cache' : ''))
     );
 
-// Save original working directory in case we need to remember our context, then
-// switch to the application directory for convenience:
+// Save original working directory in case we need to remember our context:
 define('ORIGINAL_WORKING_DIRECTORY', getcwd());
--- a/config/constants.config.php
+++ b/config/constants.config.php
@@ -1,0 +1,39 @@
+<?php
+
+// Define path to application directory
+defined('APPLICATION_PATH')
+    || define(
+        'APPLICATION_PATH',
+        (getenv('VUFIND_APPLICATION_PATH') ? getenv('VUFIND_APPLICATION_PATH')
+            : dirname(__DIR__))
+    );
+
+// Define application environment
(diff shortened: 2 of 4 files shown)


demiankatz:
@EreMaijala, this PR is an attempt to extract and refine the refactoring of public/index.php from #2233, so we can merge it independently of the larger Doctrine project. Please let me know what you think!

aleksip:
@demiankatz Just out of curiosity: why the separate `config/application.php`, if it is only included in `public/index.php`? Are there plans to include it from elsewhere too?

demiankatz:
> @demiankatz Just out of curiosity: why the separate `config/application.php`, if it is only included in `public/index.php`? Are there plans to include it from elsewhere too?

Yes, when we add Doctrine, we need to access the Application from the Doctrine console configuration file.

config/application.php
```diff
@@ -0,0 +1,16 @@
+<?php
+require __DIR__ . '/constants.config.php';
+
+chdir(APPLICATION_PATH);
```
EreMaijala, line 4: I don't really like this. I wonder if we could move away from, or if it's absolutely necessary, keep it in index.php?

demiankatz: Is it the chdir you don't like? I don't like it either, but I didn't want to make significant changes while refactoring. I'm not convinced that it's actually necessary -- we may be able to just add `__DIR__` in a place or two to avoid relative paths and then have no problems. Maybe that's better handled as a separate PR, though.

EreMaijala: Yes, sorry for being too vague. I'm ok with keeping it, but `config/application.php` feels like wrong place for it. I'd rather keep it in index.php.

demiankatz: I think it all comes down to what purpose it serves in the first place -- and honestly, I'm not sure; I think it may come from original Zend Framework boilerplate, though I could be mistaken.

If we move it to index.php, we're changing the order of operations, and if the Application initialization depends on it in some way, then we're performing the action too late.

I suppose I could just delete it and run tests and see what happens -- but I'm just concerned there may be some subtle or obscure reason for it that will bite us later.

config/constants.config.php
```diff
@@ -0,0 +1,39 @@
...
+defined('LOCAL_CACHE_DIR')
+    || define(
+        'LOCAL_CACHE_DIR',
+        (getenv('VUFIND_CACHE_DIR')
+            ? getenv('VUFIND_CACHE_DIR')
+            : (strlen(LOCAL_OVERRIDE_DIR) > 0 ? LOCAL_OVERRIDE_DIR . '/cache' : ''))
+    );
+
+// Save original working directory in case we need to remember our context, then
+// switch to the application directory for convenience:
```
EreMaijala: This comment shouldn't include the switching part anymore.

config/application.php
EreMaijala: Okay, let's keep it for now and revisit in another PR or something. My worry is that this may bite us at some point e.g. if a utility is started from another directory with a relative path to a file as parameter.

EreMaijala (changes requested):

config/constants.config.php
demiankatz: Fixed!

config/application.php
demiankatz: Sounds good. As soon as this PR gets merged, I'll open a new one to remove the chdir entirely, run tests, and report on the results. If we find a compelling reason to keep it, we can move it to public/index.php if feasible. But hopefully it can just go away and fix the relative file problem you mention.

EreMaijala (approved):

PR #2302, opened 2022-02-02
Add test for email subject

LuomaJuha:
This is my first unit test provided here so any tips would be appreciated. It just tests if the subject email is properly formed. One subject email is with placeholders and the second is without.

### DIFF
--- a/module/VuFind/tests/unit-tests/src/VuFindTest/Form/FormTest.php
+++ b/module/VuFind/tests/unit-tests/src/VuFindTest/Form/FormTest.php
@@ -22,6 +22,7 @@
  * @category VuFind
  * @package  Tests
  * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Juha Luoma <juha.luoma@helsinki.fi>
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
  */
@@ -37,6 +38,7 @@ use VuFind\Form\Form;
  * @category VuFind
  * @package  Tests
  * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Juha Luoma <juha.luoma@helsinki.fi>
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
  */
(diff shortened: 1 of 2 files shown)


LuomaJuha:
Seems like i have to create these for the version 9.0 somehow?

module/VuFind/tests/unit-tests/src/VuFindTest/Form/FormTest.php
```diff
@@ -564,4 +566,36 @@ public function testCheckboxRequiredValidators()
             $this->assertFalse($form->isValid());
         }
     }
+
+    /**
+     * Test email subjects.
+     *
+     * @return void
+     */
+    public function testEmailSubjects()
```
demiankatz: This would be a really good place to use a `@dataProvider` annotation -- instead of hard-coding $ids and $results into the body of the function, you can create a separate function that returns an array of test data (keyed by descriptive text), and it will get sent to the test method one set at a time. It makes everything a bit cleaner and easier to extend.  It would look something like:

<pre>
public function getEmailSubjectsData()
{
    return [
         'with placeholders' => ['TestSubjectEmailWithPlaceHolders', 'Subject One Two option-1'],
         'without placeholders' => ['TestSubjectEmailWithoutPlaceholders', 'Subject without placeholders'],
    ];
}

/**
 * @dataProvider getEmailSubjectsData
 */
public function testEmailSubjects($formToTest, $expectedSubject)
{
   ...
</pre>

... and then you just use the function parameters instead of having to do a for loop and index through the results. Does that make sense?

LuomaJuha: Oh, that makes actually sense, will try to adjust it tomorrow!

LuomaJuha: @demiankatz I've now updated pr with the dataprovider call

demiankatz (approved):

module/VuFind/tests/unit-tests/src/VuFindTest/Form/FormTest.php
demiankatz: Looks great -- thanks! I only learned about dataProviders recently, but they're really useful! :-)

demiankatz (changes requested):
Just realized that there's a minor problem in comments that is making the build fail -- can you run `vendor/bin/phing php-cs-fixer` and commit the results so we can merge this? Thanks! (I tried to do it myself, but I don't have write access to your branch -- sorry for the inconvenience).

LuomaJuha:
@demiankatz Seems like I totally skipped that part. Sorry about that!

demiankatz (approved):

demiankatz:
No worries. Thanks for the contribution! Merging now...

PR #2303, opened 2022-02-02
Add GeniePlus ILS driver.

demiankatz:
TODO
- [x] Access token caching
- [x] Smarter access token renewal logic
- [x] Support paginated transaction lists
- [x] User functionality
- [x] Tests


config/vufind/GeniePlus.ini
```diff
@@ -0,0 +1,77 @@
+[API]
+base_url = "https://your-api-base-url"
```
EreMaijala, line 5: Some comments for these would be nice, unless they're totally obvious for any GeniePlus user.

module/VuFind/src/VuFind/ILS/Driver/GeniePlus.php
```diff
@@ -0,0 +1,584 @@
...
+            'username' => $this->config['API']['username'],
+            'password' => $this->config['API']['password'],
+        ];
+        $headers = [
+            'Accept: application/json',
+        ];
+        $response = $this->makeRequest('POST', '/_oauth/token', $params, $headers);
+        $result = json_decode($response->getBody());
+        if ($response->getStatusCode() >= 400) {
+            throw new ILSException($response->getBody());
```
EreMaijala: I think it would be better to log any errors and throw a generic "Problem with GeniePlus API" exception. This would avoid accidentally revealing too much.

module/VuFind/src/VuFind/ILS/Driver/GeniePlus.php
```diff
@@ -0,0 +1,584 @@
...
+     *
+     * @return mixed     On success, an associative array with the following keys:
+     * id, availability (boolean), status, location, reserve, callnumber.
+     */
+    public function getStatus($id)
+    {
+        $template = $this->config['API']['catalog_template'];
+        $path = $this->getTemplateQueryPath($template);
+        $idField = $this->config['Item']['field']['id'];
+        $safeId = str_replace("'", '', $id); // don't allow quotes in IDs
```
EreMaijala: A helper method that returns a safe id or a safe search string would be nice, since this is done in more than one place.

EreMaijala (changes requested):
There are a lot of places where the code trusts a setting to be found in the ini file (e.g. `$template = $this->config['API']['patron_template']`). I think it would be good to ensure that required settings exist in `init()`.

config/vufind/GeniePlus.ini
demiankatz: Comments expanded.

module/VuFind/src/VuFind/ILS/Driver/GeniePlus.php
demiankatz: Good idea; done!

module/VuFind/src/VuFind/ILS/Driver/GeniePlus.php
demiankatz: I went back and forth about this, because I was worried that the current method is only suitable for single-quoted strings, and a shared sanitization method might create incorrect assumptions about the context in which it can be safely used... but I think the benefits outweigh the risks. I have also reached out to the vendor to see if there's a more reliable way to ensure safety of parameters without potentially garbling data, and I will update this method in the future if I receive helpful information.

demiankatz:
Thanks for the review, @EreMaijala -- I believe that I have now addressed all of your comments.

EreMaijala (approved):
Looks good to me now!

PR #2306, opened 2022-02-03
Aleph: Use default values for each Holds config key separately

xmorave2:


demiankatz (approved):

demiankatz:
Makes sense to me -- thanks!

PR #2307, opened 2022-02-03
Include Memcached adapter by default.

EreMaijala:
Since we claim to support Memcached for SearchCache, we should actually include the adapter. Otherwise setting adapter to Memcached in searches.ini will cause a failure.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
         "laminas/laminas-cache": "3.1.2",
         "laminas/laminas-cache-storage-adapter-blackhole": "^2.0",
         "laminas/laminas-cache-storage-adapter-filesystem": "^2.0",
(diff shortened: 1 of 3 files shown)


demiankatz (approved):

demiankatz:
Good catch -- thanks! I forgot about the search cache. Hopefully there aren't any other adapters expected anywhere else in the code. I know we used to have APC cache support, but when that extension was discontinued in favor of APCu, we just removed the support since nobody seemed to be using it.

Presumably it's easy enough to add additional adapters via composer.local.json as needed, though this reminds me that we need to get rid of the replace settings that were needed for transitioning from laminas-cache v2 to v3. I'll open a separate PR to address that.

PR #2310, opened 2022-02-07
Make session file writer handle simultaneous requests properly.

EreMaijala:
- Require exclusive write lock before writing.
- Truncate the file to the correct length to avoid any leftover content.


demiankatz (approved):

demiankatz:
I have backported this to the release-8.0 branch since it is a fairly high-priority bug fix.

PR #2311, opened 2022-02-07
Avoid session manager's writeClose in disableWrite.

EreMaijala:
If the session is already active (and it probably is since bootstrapper will have it initialized), writeClose will write the session data, which needs to be avoided.


demiankatz:
I have backported this to the release-8.0 branch since it is a fairly high-priority bug fix.

demiankatz (approved):

PR #2313, opened 2022-02-09
Build OpenURL without http_build_query to avoid PHP array syntax in URL

rajaro:
PHP array syntax in url parameters does not work with OpenURL

### DIFF
--- a/module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
+++ b/module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
@@ -886,7 +886,14 @@ class DefaultRecord extends AbstractBase
         }
 
         // Assemble the URL:
-        return http_build_query($params);
+        $query = [];
+        foreach ($params as $key => $value) {
+            $value = (array)$value;
+            foreach ($value as $sub) {
+                $query[] = $key . '=' . urlencode($sub);
+            }
+        }
+        return implode("&", $query);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


demiankatz:
Thanks, @rajaro! Two questions:

1.) Could you add a test to [DefaultRecordTest](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/tests/unit-tests/src/VuFindTest/RecordDriver/DefaultRecordTest.php) to exercise/demonstrate the change? Let me know if I can be of assistance with that.

2.) Do we need to URL-encode the keys as well as the values just to be on the safe side?

rajaro:
@demiankatz Hi,

1.) Sure. The issue is basically the parameter `rft_id`, which can have multiple values. How should I write a test for it? I don't think the default record driver adds multiple values to any key right now.

2.) Better to be on the safe side! I added URL-encoding for the keys as well

demiankatz:
Thanks, @rajaro!

Regarding testing, adding a `testGetOpenUrl` method to DefaultRecordTest would be a good starting point. In terms of testing repeating parameters, if there's no way to make that happen in default VuFind, we could simulate it with a mock method. I'm on vacation this week, but I may be able to help with test development after my return if that would be useful!

demiankatz:
@rajaro, apologies for losing track of this PR. Would you like me to work on adding the test I suggested, or do you have any bandwidth to do that?

rajaro:
@demiankatz I also had to focus on other things, but will continue with this next week. 

E: ..And I was doing other things. NEXT week I'll finish this one.

demiankatz:
@rajaro, just checking in on this one -- you planned to finish this in April, but it seems you didn't get back to it. Is there anything I can do to help?

rajaro:
@demiankatz Hi, I wrote up a test that simulates a situation where `DefaultRecord::getBookOpenUrlParams` returns one of the parameters as an array with multiple values. This should demonstrate the change, as the URL returned by `DefaultRecord::getOpenUrl` does not contain braces as it would if http_build_query was used.

demiankatz (approved):
Thanks, @rajaro!

PR #2314, opened 2022-02-09
Fix bugs in updateRecordId() method.

demiankatz:
- We should not create resources when checking if they already exist!
- We should not use the record ID as a function name!


demiankatz:
@EreMaijala, I'm working on creating a fallback record loader for Solr, and in testing it, I discovered some major bugs in the logic supporting fallback loading functionality. Not sure how these slipped through the cracks!

EreMaijala (approved):
That sure was broken. The fixes look good to me.

demiankatz:
Thanks, @EreMaijala -- pretty sure those were all my fault somehow. Good thing we now have a review-based workflow, because I'm pretty sure those types of mistakes would not have gotten in if we had started doing this sooner. :-)

PR #2315, opened 2022-02-09
Solr fallback record loader

demiankatz:
TODO:
- [x] Add configuration option to set the fallback field or disable fallback loading (default should probably be off)
- [x] Merge #2314 first -- this depends on those bug fixes
- [x] Decide whether this is too backward-breaking for inclusion in 8.1; if so, rebase to dev-9.0

### DIFF
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -137,7 +137,7 @@ retain_filters_by_default = true
 ; If you set this setting, VuFind will use a secondary Solr field to try to find
 ; previous record IDs when primary ID lookups fail. If you leave the setting
 ; commented out, fallback ID lookups will be disabled.
-;legacy_id_field = previous_id_str_mv
+;fallback_id_field = previous_id_str_mv
 
 [Cache]
 ; This controls whether the parsed searchspecs.yaml file will be stored to
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -134,6 +134,11 @@ retain_filters_by_default = true
 ;default_parameters[*] = "echoParams=none&debugQuery=false"
 ;default_parameters[search] = "sow=false"
 
+; If you set this setting, VuFind will use a secondary Solr field to try to find
+; previous record IDs when primary ID lookups fail. If you leave the setting
+; commented out, fallback ID lookups will be disabled.
+;legacy_id_field = previous_id_str_mv
+
 [Cache]
 ; This controls whether the parsed searchspecs.yaml file will be stored to
 ; improve search performance; legal options File (store on disk) or false (do not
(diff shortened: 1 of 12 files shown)


demiankatz:
@EreMaijala, this creates a mechanism to prevent links from breaking when IDs change (e.g. when you migrate from one ILS to another, your IDs change, but you still have access to your old IDs). It uses infrastructure originally built for Summon ID changes, but I think it should work well as a general-purpose tool.

EreMaijala:
@demiankatz This looks like a solid start. I think it would be useful to provide the possibility of using a rule to pick something from the old ID. In my experience the old ID and/or new ID might have a prefix that needs to be taken into account. Also, this would probably not work well with MultiBackend and prefixed IDs. For reference, this is what we've done locally: 

https://github.com/NatLibFi/NDL-VuFind2/blob/dev/module/Finna/src/Finna/Record/Loader.php#L218
https://github.com/NatLibFi/NDL-VuFind2/blob/dev/local/config/finna/config.ini.sample#L1402

It's not architected nearly as well as your proposal, but you get the idea. So I think it would be useful to be able to provide rules that could be applied to the requested ID. If such rules are configured, any that match the given ID could be looked up.

demiankatz:
Thanks, @EreMaijala, that's a good point. I guess the question is where that functionality belongs, both in terms of code and PR logistics.

The code in this PR uses the existing record fallback mechanism and meets my local use case, so an argument could be made that it's worth merging as-is and then addressing ID manipulation in a separate PR. The main reason I would instead opt to wait would be if the changes you propose would significantly reshape the work in progress here.

And that leads to the question of "where in the code?" -- do we want some kind of central mechanism for ID remapping where the logic resides in the Record Loader, or do we want to handle ID remapping on a case by case basis in different fallback loader plug-ins? I think tying it to the fallback loader probably makes the most sense, since each search backend may have very different use cases.

You you have different feelings about any of this, let me know, but if you generally agree, I propose this: we merge what we have, then you try to incorporate your local fix into the core code, and open a new PR to add the ID-related functionality. What do you think?

EreMaijala:
@demiankatz For me the question is if the settings need changes or it's difficult to support the existing setting if this is expanded. E.g. if the setting name would be `fallback_id_field`, it would fit better with future settings and could be used as a fallback when the id doesn't match any other rules. With that I'd be happy to merge as is.

demiankatz:
Thanks, @EreMaijala, I'm happy to rename the setting to `fallback_id_field` and will do so when time permits (hopefully later today). Do you have any thoughts on whether this is better for 8.1 or 9.0? On the one hand, it's a fairly significant refactoring; on the other, it doesn't change public interfaces, and it's a feature that is only currently used as part of a rarely-used backend. On balance, I lean toward 8.1 but am open to acting differently. :-)

EreMaijala:
@demiankatz I think this would be fine for 8.1, though I don't have a strong opinion on it.

EreMaijala (approved):

PR #2316, opened 2022-02-10
Prevent adding facet count if it is undefined

LuomaJuha:
Sometimes this will break fetching of facets if there is some oddities in the index.

### DIFF
--- a/themes/bootstrap3/js/facets.js
+++ b/themes/bootstrap3/js/facets.js
@@ -42,10 +42,12 @@ function buildFacetNodes(data, currentPath, allowExclude, excludeTitle, counts)
     html.appendChild(item);
 
     if (!facet.isApplied && counts) {
-      var badge = document.createElement('span');
-      badge.className = 'badge';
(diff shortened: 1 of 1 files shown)


demiankatz (approved):

demiankatz:
Thanks, @LuomaJuha. This seems like a reasonable extra safety check which I am happy to merge, but a related question: should there be matching code on the back end somewhere that prevents invalid count values from getting passed forward to the Javascript?

LuomaJuha:
@demiankatz I actually have no idea where my error is coming from but if I find the reason, I will try and see if there is anything to be done. At the moment only thing reading in my Organization facet is: null and after that facets are normal. That was the reason my facets did not work.

demiankatz:
I wonder if @EreMaijala might have any ideas...

EreMaijala:
@demiankatz I'll take a look.

EreMaijala:
It's a local indexing issue where a hierarchical facet value contains more levels than it should, so the index entries are not correct. I don't think there should be further need to guard against the Solr index being invalid.

demiankatz:
Thanks for the investigation, @EreMaijala!

PR #2322, opened 2022-02-14
Properly escape quotes in GeniePlus queries.

demiankatz:
I've received more detailed documentation on the GeniePlus query language and was thus able to improve escaping. I have tested that this really works as it should.


EreMaijala (approved):
Nice!

PR #2323, opened 2022-02-14
GeniePlus: Add custom holdings sorting option.

demiankatz:
The order of holdings/items coming back from the GeniePlus API can be quite arbitrary; this adds a mechanism for controlling the display.


EreMaijala (approved):
I suppose the comment in the ini file could allow for other fields as well, but this looks like it would work fine as it is.

demiankatz:
Thanks, @EreMaijala -- I made the comment more accurate and then merged this.

PR #2324, opened 2022-02-14
EDS backend: support additional "missing record" cases

he247:
To prevent errors of the backend to interrupt the batch process of retrieve the general BackendException had to be caught as well.


module/VuFindSearch/src/VuFindSearch/Backend/EDS/Backend.php
```diff
@@ -342,6 +342,8 @@ public function retrieve($id, ParamBag $params = null)
             default:
                 throw $e;
             }
+        } catch(BackendException $e) {
```
demiankatz: What exception is actually getting thrown here? Is there some more specific metric we can use to determine when it's appropriate to return an empty response (because, for example, the record really doesn't exist) and when something is more seriously wrong with the system?

demiankatz: Also note that you're missing a space after `catch` if you want to conform to style guidelines, which is at least in part why the build is failing here. You should be able to auto-fix it with `vendor/bin/phing php-cs-fixer`.

he247: The Exception is thrown in line 293, so I think we would have to define the metric by ourself if I'm not mistaken.

he247:
Could we check for Error-Code 0? Or set a specific number in line 293?

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Backend.php
demiankatz: I thought the exception thrown in line 293 was related to input validation -- that's saying that the provided ID is not in the expected format before even making an API call at all. How would invalid IDs end up on your lists to begin with? Or am I misunderstanding something? I thought the problem you were trying to solve here had to do with requesting IDs for records that were previously available but no longer exist.

he247: Ah, sorry then it's my fault. I tried to reproduce an error reported by a user but it seems I just didn't create a proper test case. :/ 
I will search for the id and check again what the exact exception will be.

he247: The actual error was much simpler. Just an APIException 132  which said that the record does not exist. Sorry for the confusion.

demiankatz: Thanks, @he247, that's great! My only remaining suggestion would be to put the `case 132` above the `case 135` to keep things in numeric order.

I am curious why the existing code used 135 for this scenario and you're getting 132. I wonder if something changed, or if there are potentially other 13x codes with the same basic meaning. Are error codes documented somewhere, or is it worth reaching out to EBSCO for clarity? Perhaps @cwolfebsco can shed some light.

cwolfebsco: @demiankatz I don't have that insight, but I will happily reach out to API product management, if you wanted me to. The code however goes back a few years, so I am not sure we will be lucky.

From my reading both 132 and 135 are likely cases. 132 would suggest a record deletion, 135 would suggest customer removed a db from the API profile.

**_132_** will be thrown, if you request an AN for an active DB, but the AN is not found. This case should be very rare, as content removals don't happen that often (i.e. those records rarely get deleted).

**_135_** will be thrown if you request a db that is not within your EDS API profile (even if the record theoretically exists), this would happen if the API profile was "cleaned up"

Under very awkward conditions a 133 might be encountered, but that's only likely if a client had a database with simultaneous user limits on their profile (I even have a hard to time to come up with an example right now)

For reference, here is a full list of Error Codes, Source: https://connect.ebsco.com/s/article/EBSCO-Discovery-Service-API-Reference-Guide-Error-Codes (retrieved 15 Feb 2022)


Error Code | Description
-- | --
100 | Unknown Parameter
101 | Incorrect Parameter Format
102 | Invalid Parameter Index
103 | Missing Parameter
104 | Auth Token Invalid
105 | Incorrect Number of Arguments
106 | Unknown Error
107 | Authentication Token Missing
108 | Session Token Missing
109 | Session Token Invalid
111 | Unknown Action
112 | Invalid Argument Value
113 | Error encountered while creating a session.
114 | Required data missing.
116 | Duplicate query string parameter found where only one should be specified.
119 | Invalid Page Size.
120 | Error saving session
121 | Error ending session
123 | Invalid Expander
124 | Invalid Search Mode
125 | Invalid Limiter
126 | Invalid Limiter Value
127 | Unsupported Profile
128 | Profile not supported for Discovery Service
129 | Invalid Content Provider
130 | Invalid Source Type
131 | XSLT Error
132 | Record not found
133 | Simultaneous User Limit Reached
134 | No Guest Access
135 | DbId Not In Profile
136 | Invalid Search View
137 | Error Retrieving Full Text
139 | Invalid IP Address
140 | Missing IP Address
141 | Unknown IP Authentication Error
142 | Invalid Preferred eBook Format
143 | Invalid Search Request - Probable Invalid Query Term
144 | Profile ID is not assocated with caller's credentials.
145 | Profile not configured for Publication features
148 | Guest session access restricted
149 | Operation unsupported. Please contact EBSCO
150 | Error occurred when calling Citation Adapter Service
151 | Citation Styles are not available for this record
152 | Export format not available for this record

he247: Numeric order corrected. :)

Should we add the 133 as well? Even if it is rare, I have the philosophy that the overall system should not be blocked by exceptions (especially not if it is coming from outside like an API) but rather handle them in a proper way, if possible.

I mean, if there is an error on any API I want the user to be able to continue working as normal as possible and maybe only to receive a message that something went wrong in a specific part of the process. But if we have information locally I want the system to show what we have and not let an API exception block VuFind in a broader range than it should.
For example in the user lists, I'm showing the user that the record is not available in the database any more and give all the information stored in the DB, with a link to a new search, so that the user could get alternative sources if available.

demiankatz: Thanks, both of you! I don't think @cwolfebsco needs to bother anyone else -- the reference he shared is extremely helpful and I think answers the necessary questions. @he247, I agree that adding 133 is probably worthwhile, even if it's rare. It might also be worth using comments to annotate the codes in the switch statement with the descriptions shared above and/or including a link to the referenced documentation, just so it's more clear exactly what everything means if we review this code again in future.

he247: Great I'll add the comments and 133. Then I'll open the review again.

demiankatz: Thanks, @he247 -- I'm leaving on vacation right now, but I'll merge this as soon as I can. :-)

he247: @demiankatz enjoy your holiday! :) Thank you for your help, as always!

demiankatz (approved):

PR #2325, opened 2022-02-16
SierraRest: Rework canceling and updating of holds.

EreMaijala:
- Decouples them from each other.
- Makes canceling of available/in transit holds configurable.
- Uses Sierra's canFreeze field to determine if the hold can be frozen (this is only checked when making the changes for performance reasons).
- Allows partial update even if not everything can be changed.


module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -914,32 +935,70 @@ public function updateHolds(
...
+                );
+
+                if (!empty($result['code'])) {
+                    $results[$requestId] = [
+                        'success' => false,
+                        'status' => $this->formatErrorMessage(
+                            $result['description'] ?? $result['name']
+                        )
+                    ];
+                } else {
```
demiankatz, line 1012: Is there a compelling reason to nest here -- would this be a little more readable as a single `if .. elseif .. else` clause?

EreMaijala: No reason, fixed!

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -821,7 +835,14 @@ public function getMyHolds($patron)
                 ) : '';
             $inTransit = $entry['status']['code'] === 't';
             $requestId = $this->extractId($entry['id']);
-            $updateDetails = ($available || $inTransit) ? '' : $requestId;
+            // Allow the user to attempt update if freezing is enabled or the hold
+            // is not available or in transit. Checking if the hold can be freezed
+            // up front is slow, so defer it to when update is requested.
+            if ($freezeEnabled || (!$available && !$inTransit)) {
+                $updateDetails = $requestId;
```
demiankatz: I think you need to initialize $updateDetails before this, or have an else clause, to prevent it from being potentially undefined when used below.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -914,32 +935,68 @@ public function updateHolds(
...
+                . (isset($fields['frozen']) ? ',canFreeze' : '');
+            $hold = $this->makeRequest(
+                [$this->apiBase, 'patrons', 'holds', $requestId],
+                [
+                    'fields' => $reqFields
+                ],
+                'GET',
+                $patron
+            );
+            $available = in_array($hold['status']['code'], ['b', 'j', 'i']);
```
demiankatz: Should the available codes be defined in a property for easier overriding/future configuration support?

demiankatz (commented):
I gave this a slightly more careful second look and came up with a couple more comments...

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
demiankatz: (Not sure why PHPstan didn't complain about this -- maybe I'm misreading something).

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Those are pretty static, but since they're used in a couple of places, that makes sense. Now done.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Oops. And since it's used in a loop, the previous value might have been used. Now fixed.

demiankatz (approved):

demiankatz:
Thanks, @EreMaijala -- this looks good to me. Unfortunately, I can't actually test this functionality, but I've done the best I can in terms of review. In the long term, some better test coverage of this driver would be good as well, but I'm not going to insist on that right now. :-)

PR #2326, opened 2022-02-16
Add safeguards to language normalization.

EreMaijala:
The command can be quite destructive if run on non-language files. These changes prevent normalization if the file doesn't look like a language file, and make the normalization of a directory only process file names matching the normal language file name pattern.


module/VuFind/tests/unit-tests/src/VuFindTest/I18n/ExtendedIniNormalizerTest.php
```diff
@@ -55,6 +57,23 @@ public function testLanguageFileIntegrity()
...
+     * Test bypassing of non-language-files.
+     *
+     * @return void
+     */
+    public function testLanguageFileCheck()
+    {
+        $file = $this->getFixtureDir() . 'language/base/non-language.ini';
+        $normalizer = new ExtendedIniNormalizer();
+
+        $this->expectExceptionMessage(
```
demiankatz, line 70: Should we have a second fixture to test the missing equal sign check for completeness?

demiankatz (changes requested):
Thanks; this looks good to me -- just one minor comment (which I can take care of myself after vacation if you prefer). :-)

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/ExtendedIniNormalizerTest.php
EreMaijala: I should have added it when I briefly thought about it, but then forgot. Now added.

demiankatz (approved):

PR #2327, opened 2022-02-17
Clean up layout.phtml a bit

LuomaJuha:
Altered the comment of Cart::getCookieSameSite as the constructor of CookieManager offers 'Lax' as the default value.

### DIFF
--- a/themes/bootstrap3/templates/layout/layout.phtml
+++ b/themes/bootstrap3/templates/layout/layout.phtml
@@ -94,9 +94,9 @@ JS;
           $this->headScript()->appendFile("covers.js");
       }
     ?>
-    <?=$this->headScript()?>
+    <?=$this->headScript() ?>
   </head>
-  <body class="template-dir-<?=$this->templateDir?> template-name-<?=$this->templateName?> <?=$this->layoutClass('offcanvas-row')?><?php if ($this->layout()->rtl): ?> rtl<?php endif; ?>">
+  <body class="template-dir-<?=$this->templateDir ?> template-name-<?=$this->templateName?> <?=$this->layoutClass('offcanvas-row')?><?php if ($this->layout()->rtl): ?> rtl<?php endif; ?>">
     <?php // Set up the search box -- there are three possible cases:
       // 1. No search box was set; we should default to the normal box
       // 2. It was set to false; we should display nothing
@@ -132,16 +132,16 @@ JS;
       <div class="modal-dialog">
         <div class="modal-content">
           <button type="button" class="close" data-dismiss="modal">
-            <svg aria-label="<?=$this->transEscAttr('Close')?>" viewBox="0 0 24 24"><path d="M6 18L18 6M6 6l12 12"></path></svg>
+            <svg aria-label="<?=$this->transEscAttr('Close') ?>" viewBox="0 0 24 24"><path d="M6 18L18 6M6 6l12 12"></path></svg>
           </button>
           <div class="modal-body"></div>
         </div>
       </div>
     </div>
     <div class="offcanvas-overlay" data-toggle="offcanvas"></div>
-    <?=$this->googleanalytics()?>
-    <?=$this->piwik()?>
-    <?=$this->matomo()?>
+    <?=$this->googleanalytics() ?>
+    <?=$this->piwik() ?>
+    <?=$this->matomo() ?>
     <?php foreach ($this->captcha()->js() as $jsInclude):?>
       <?=$this->inlineScript(\Laminas\View\Helper\HeadScript::FILE, $jsInclude, 'SET')?>
     <?php endforeach; ?>
--- a/themes/bootstrap3/templates/layout/layout.phtml
+++ b/themes/bootstrap3/templates/layout/layout.phtml
@@ -96,7 +96,7 @@ JS;
     ?>
     <?=$this->headScript() ?>
   </head>
-  <body class="template-dir-<?=$this->templateDir ?> template-name-<?=$this->templateName?> <?=$this->layoutClass('offcanvas-row')?><?php if ($this->layout()->rtl): ?> rtl<?php endif; ?>">
+  <body class="template-dir-<?=$this->templateDir?> template-name-<?=$this->templateName?> <?=$this->layoutClass('offcanvas-row')?><?php if ($this->layout()->rtl): ?> rtl<?php endif; ?>">
     <?php // Set up the search box -- there are three possible cases:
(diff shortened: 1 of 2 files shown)


themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -119,17 +116,14 @@ JS;
     </header>
     <nav class="breadcrumbs" aria-label="<?=$this->transEscAttr('Breadcrumbs') ?>">
       <div class="container">
-        <?php if((!isset($this->layout()->showBreadcrumbs) || $this->layout()->showBreadcrumbs == true)
-          && !empty($this->layout()->breadcrumbs)
-          && $this->layout()->breadcrumbs !== false
-        ): ?>
+        <?php if (!empty($this->layout()->showBreadcrumbs) && !empty($this->layout()->breadcrumbs)): ?>
```
demiankatz: I don't think this refactoring is correct, because the old code defaults showBreadcrumbs to true and the new code defaults showBreadcrumbs to false. I think you probably want `!empty($this->layout()->showBreadcrumbs ?? true)` to avoid changing behavior.

LuomaJuha: Oh, took a while to notice but it seems like I understood that part totally wrong. Gonna tune it!

demiankatz:
Thanks, @LuomaJuha, this looks good to me now, but I'm asking for a double-check from @crhallberg just in case I'm missing anything. (Since this is such a central file, I want to be sure we're careful about changing it).

Would you object if I merge this into the dev-9.0 branch instead of the dev branch, just to move the template changes into a major release (to reduce the amount of reconciliation work needed in local templates when upgrading to the eventual minor 8.1 release)?

LuomaJuha:
@demiankatz I don't mind at all as it does not change the functionality of the code in any way. It is just a small tuning pr.

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -138,7 +132,7 @@ JS;
       <div class="modal-dialog">
         <div class="modal-content">
           <button type="button" class="close" data-dismiss="modal">
-            <svg aria-label="<?=$this->transEscAttr('Close') ?>" viewBox="0 0 24 24"><path d="M6 18L18 6M6 6l12 12"></path></svg>
+            <svg aria-label="<?=$this->transEscAttr('Close')?>" viewBox="0 0 24 24"><path d="M6 18L18 6M6 6l12 12"></path></svg>
```
crhallberg: I don't necessarily disagree with these white-space changes, but I am curious if they came from a style guide or a personal preference. As the progressive theme is built, I want to make sure the formatting is most agreeable.

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -50,16 +50,13 @@
...
             $appendScripts[] = 'VuFind.cart.setDomain("' . $domain . '");';
           }
-          $cookiePath = $cart->getCookiePath();
-          if (!empty($cookiePath)) {
+          if ($cookiePath = $cart->getCookiePath()) {
             $appendScripts[] = 'VuFind.cart.setCookiePath("' . $cookiePath . '");';
           }
-          $cookieSameSite = $cart->getCookieSameSite();
-          if (null !== $cookieSameSite) {
+          if ($cookieSameSite = $cart->getCookieSameSite()) {
```
crhallberg, line 59: This is a drop in specificity from `!== null`. I don't know if it's a meaningful in the practice of our configs. I guess I should trust the tests.

themes/bootstrap3/templates/layout/layout.phtml
demiankatz: It would be nice to see if we could find a php-cs-fixer rule to apply consistent formatting, if we have a preference one way or the other.

demiankatz:
Thanks again, @LuomaJuha and @crhallberg. I've re-targeted this PR to dev-9.0 based on earlier conversation, and I will wait to see how the current discussions play out before merging.

themes/bootstrap3/templates/layout/layout.phtml
crhallberg: I'm not seeing anything in the obvious places about this. The examples tend to have a space but there's inconsistency with semicolons: `); ?>` vs. `) ?>`.

Nothing in php-cs-fixer, as far as I can tell.

crhallberg: One final reference: Laminas documentation seems to use no-semicolon, space at both ends: `<?= $this->htmlTag(['lang' => 'en'])->openTag() ?>`.

Not a fan of the space at the front, but this puts the majority of large projects in the space-please category. https://docs.laminas.dev/laminas-view/helpers/html-tag/.

Sorry to make this about the color of the bikeshed.

demiankatz: Yeah, I'm also not a fan of the leading space, but I could live with it if we had tools to apply it -- consistency would be nice. However, it seems you are correct that php-cs-fixer is not going to help us at present; see: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5138

LuomaJuha: @crhallberg They were actually out of the habit. Can tune them back into what they were.

themes/bootstrap3/templates/layout/layout.phtml
LuomaJuha: The default value returned is Lax as if someone tries to put an empty value there, it would just be odd.

themes/bootstrap3/templates/layout/layout.phtml
demiankatz: @LuomaJuha, that's probably worth doing -- reducing whitespace-based churn will make the commit history easier to read. I'm definitely open to revisiting how we format our short-tags in templates, but I think that's best done globally with an automated tool rather than piecemeal by hand. :-)

LuomaJuha: Yea, I've restored the whitespaces into their original places. : )

demiankatz (approved):
This looks good to me now, but I'll leave it open a little longer in case @crhallberg has any remaining concerns about the relaxed cookie checks that he wants to discuss.

crhallberg:
My concerns have all been addressed.

demiankatz (approved):

demiankatz:
Thanks, @crhallberg and @LuomaJuha -- merging now!

PR #2328, opened 2022-02-18
Relax MarcReader's format checking.

EreMaijala:
Invalid ISO2709 records no longer cause exceptions. Any errors are stored as warnings.

Resolves [VUFIND-1534](https://vufind.org/jira/browse/VUFIND-1534).


demiankatz:
@EreMaijala, as we discussed earlier, I have added a long record as a test fixture in this PR. If you have time to write the test, please do -- otherwise, let me know and I'll work on it when time permits.

EreMaijala:
@demiankatz I added a basic test. I tried to check the results, but then I realized that the directory also overflows so after data offset 99999 all the fields are borked as well.

demiankatz:
@EreMaijala, I'm not at all familiar with the specifics of the MARC format so I could be totally wrong about this, but based on vague recollection of past conversations about this issue, I seem to recall hearing that the length value in the leader isn't really necessary because you can still parse the file based on the end-of-field markers. I think this is how marc4j's tolerant mode works. Is it possible we could ignore the length value and apply a different strategy when length is 99999 (or maybe all the time, if it makes no difference), or am I off-base?

EreMaijala:
@demiankatz That's right, but each directory entry also has a 5 digit offset indicating the offset of the field in the record. There's no way around that without breaking parsing the directory as it should be done.

EreMaijala:
@demiankatz As far as I can see, this applies to File_MARC's parsing as well.

demiankatz:
Ahh, I see. I thought there was a way to completely ignore the directory and parse the record without it... but it seems that I was mistaken.

I see that your test is currently counting the 852's. Are you saying that the actual content of some of those fields is garbled even though the count is correct? Or are we only getting a subset back? Is it worth adding a check to validate the correct 245 and 100, at least, to verify that early fields in the record are interpreted correctly?

I also wonder if we should go ahead and backport this to release-8.0 so we can get a release out where long MARC isn't completely breaking everything, even if we can't handle it perfectly. As long as the behavior is no worse than File_MARC, we're at least maintaining the status quo.

EreMaijala:
@demiankatz Yes, all fields in directory are present, but when the offset overflows, they just repeat parts of subfields starting from the same offset, so yes, they're garbled. I'll add checks for proper content in early fields.

EreMaijala:
@demiankatz Since this is based on dev, it should be easy to backport, and it would make sense to do, I think.

EreMaijala:
More tests added.

demiankatz (approved):

demiankatz:
Thanks again, @EreMaijala! I have merged this and backported it to release-8.0.

PR #2330, opened 2022-02-23
Fix EDS advanced search icons

crhallberg:
Replace Unicode character in remove button with icon.


demiankatz (approved):

demiankatz:
Thanks, @crhallberg! I've merged this all the way to dev-9.0. Do you want to open a new separate PR to use the icon helper in 9.0? There should now be enough logic already merged there to support doing it directly, rather than adding on to #1962.

PR #2331, opened 2022-02-23
Theme Config Helper Edge fixes

crhallberg:
- Fixes a bug when string values were being merged (still returns array)
- Adds ability to return entire config.

- [x] Add caching? Move caching from ThemeConfig helper up to ThemeInfo?


module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
```diff
@@ -310,6 +310,30 @@ public function testGetMergedConfigMixinWithFlattening()
         );
     }
 
+    /**
+     * Test getMergedConfig() with string return
```
demiankatz: Just curious about name/comment on this test -- not sure what you mean by "with string return" since this is returning an array. Do you mean with string key?

demiankatz:
Thanks, @crhallberg -- apart from the minor question above, this makes sense to me. I do agree that it might make more sense to move the caching into ThemeInfo instead of the helper, since then you'll be caching data at the source.... If you were to do that, though, it would be a good idea to make the cache an optional dependency, and only cache if it's injected, so we don't introduce a backward-breaking change in a minor release.

demiankatz:
Thanks for the comment cleanup, @crhallberg. Do you want me to merge this now, or are you still contemplating the cache changes?

module/VuFind/src/VuFind/View/Helper/Root/ThemeConfigFactory.php
```diff
@@ -69,15 +69,16 @@ public function __invoke(
...
         // As of release 1.1.0, the memory storage adapter has a flaw which can cause
         // unnecessary out of memory exceptions when a memory limit is enabled; we
         // can disable these problematic checks by setting memory_limit to -1.
         $cacheConfig = ['name' => 'memory', 'options' => ['memory_limit' => -1]];
         $cache = $container->get(\Laminas\Cache\Service\StorageAdapterFactory::class)
             ->createFromArrayConfiguration($cacheConfig);
 
-        return new $requestedName($themeInfo, $cache);
+        $themeInfo = $container->get(\VuFindTheme\ThemeInfo::class);
+        $themeInfo->setCache($cache);
```
demiankatz: I think it would make much more sense to set the cache in the factory for ThemeInfo rather than here. We don't want the factory for one service to modify a different service as a side effect.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -176,7 +197,7 @@ protected function loadThemeConfig($theme)
     public function getThemeInfo()
     {
         // Fill in the theme info cache if it is not already populated:
-        if (null === $this->allThemeInfo) {
+        if ($this->allThemeInfo === null) {
```
demiankatz, line 200: Any particular reason you reversed the comparison here? Seems like a potentially unnecessary diff.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -82,6 +91,18 @@ public function __construct($baseDir, $safeTheme)
...
     }
 
+    /**
+     * Provide cache and activate info caching
+     *
+     * @param StorageInterface $cache cache object
+     *
+     * @return void
+     */
+    public function setCache(StorageInterface $cache)
```
demiankatz, line 101: In order to keep our coverage up, I think we should add some new tests to [ThemeInfoTest](https://github.com/vufind-org/vufind/blob/dev/module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php) to inject mock cache objects to confirm that values coming from the cache get returned and that the cache gets updated when it returns null.

demiankatz (changes requested):
Thanks for the progress, @crhallberg -- I think we need just a little more work done here.

Also, do you have any idea why the Snyk service is failing the build? I think that's something you set up. Has it changed or broken? Do we need to remove it, or can it be fixed?

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
crhallberg: Just a nitpick.

demiankatz: I thought the reversed form was preferred since it prevents accidental assignments, but it's definitely not a hill I'm willing to die on. ;-)

demiankatz (approved):
I made a couple of trivial tweaks (added a test assertion and updated a comment). I think this is ready to go now. Thanks!

PR #2334, opened 2022-02-25
Add MARC-in-JSON support.

EreMaijala:
Initial MARC-in-JSON support implementation. 

Notes:

- MarcReader lists JSON before MARCXML with the anticipation that JSON would become more common than MARCXML.
- MarcCollection and MarcCollectionFile, on the contrary, list it last, since MARC-in-JSON collection is not expected to be as common.
- Tests include schema validation. Schema is based on https://web.archive.org/web/20151112001548/http://dilettantes.code4lib.org/blog/2010/09/a-proposal-to-serialize-marc-in-json/ with minor updates to


module/VuFind/src/VuFind/Marc/Serialization/MarcInJson.php
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * Convert record to ISO2709 string
+     *
+     * @param array $record Record data
+     *
+     * @return string
+     */
+    public static function toString(array $record): string
+    {
+        // We need to cast any subfield with '0' as key to an object to, otherwise
```
demiankatz: I think there's an extra "to" in this sentence.

module/VuFind/src/VuFind/Marc/Serialization/MarcInJson.php
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Get next record from the file or an empty string on EOF
+     *
+     * @return string
+     */
+    public function getNextRecord(): string
+    {
+        ++$this->position;
```
demiankatz: This does not appear to be streaming support. Should we add a TODO or open a JIRA ticket? I realize that's probably a bigger project and not necessarily a priority right now... but we might want it in the future, so we shouldn't forget about it.

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcReaderTest.php
```diff
@@ -54,7 +54,19 @@ public function testMarcReader()
...
         // Test round-trips
         $reader = new \VuFind\Marc\MarcReader($reader->toFormat('MARCXML'));
+        $this->assertXmlStringEqualsXmlString($marc, $reader->toFormat('MARCXML'));
         $reader = new \VuFind\Marc\MarcReader($reader->toFormat('ISO2709'));
+        $this->assertXmlStringEqualsXmlString($marc, $reader->toFormat('MARCXML'));
+        $reader = new \VuFind\Marc\MarcReader($reader->toFormat('JSON'));
+        $this->assertXmlStringEqualsXmlString($marc, $reader->toFormat('MARCXML'));
+
+        // Verify JSON schema
+        $validator = new \Opis\JsonSchema\Validator();
```
demiankatz, line 64: If the validator is only used in tests, should it be a dev dependency? Or do you plan to use it elsewhere (or is it already in use somewhere that I overlooked?)?

demiankatz (commented):
Thanks for this! It's the end of the day at the end of the week, so I don't claim I did my best review work, but I did have a few small questions. :-)

module/VuFind/src/VuFind/Marc/Serialization/MarcInJson.php
EreMaijala: Fixed!

module/VuFind/src/VuFind/Marc/Serialization/MarcInJson.php
EreMaijala: I added proper streaming support. :)

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcReaderTest.php
EreMaijala: It *is* a dev dependency (in `"require-dev"` of composer.json), or am I missing something?

he247:
Dear @EreMaijala and @demiankatz do you plan to merge this branch later into Release 8 as well, or are there too many other changes that would make it impossible?

EreMaijala:
@he247 It's possible to backport, but there are at least the following reasons I didn't do that yet:

1. This changes the internal format MarcReader uses, and this is a compatibility-breaking change for anyone extending MarcReader. I'm not sure how much of an issue it is, and the change could be considered a bug fix as well, since it makes MarcReader able to maintain order of non-adjacent MARC fields.
2. Tests depend on a recent version of [opis/json-schema](https://packagist.org/packages/opis/json-schema) that requires PHP 7.4. This conflicts with VuFind 8's minimum of PHP 7.3.

he247:
@EreMaijala I see, thank you for the clarification.

module/VuFind/tests/unit-tests/src/VuFindTest/Marc/MarcReaderTest.php
demiankatz: Oops, I just misread the file. My apologies!

demiankatz:
Thanks for all the progress on this, @EreMaijala. I've tested that this works as expected in combination with the `fullrecord = FullRecordAsJSON2` setting in SolrMarc. Do you think we should introduce that as a commented-out example in marc_local.properties as a starting point here, or do you want to take more aggressive action toward using MARC-in-JSON?

Regarding backporting this to VuFind 8, I agree that it's probably not too hard to cherry-pick a few commits to get this in a local instance, but trying to backport it into a formal release would be a bit complex due to backward compatibility / semantic versioning issues. @he247, are you comfortable dealing with this locally, or do you think it is a high priority to try to backport it officially?

EreMaijala:
@demiankatz I believe we will be switching to JSON locally as soon as possible, but that doesn't need to be the default for everyone, and I don't have a strong opinion. That said, using JSON would alleviate the problem with long broken records, which are obviously met in the wild, so based on that I'd be tempted to turn MARC-in-JSON on by default.

he247:
@demiankatz, if I understood correctly, the solr data of oversize records is stored corrupt in the index and was already like this in former versions, right? At least it wouldn't break something by not backporting it, as it wasn't working properly anyway. On the other hand this could also be an argument for backporting it.
Personally, I will either backport it locally, if not done officially, or try to change to Release 9 asap.

Maybe an official patch for version 8 could be an option, this would have the advantage to not add compatibility-breaking changes directly to the release-branch but providing a more streamlined approach to fix the bug, if one decides to do it. What do you think? (I didn't do such thing yet)

@EreMaijala and @demiankatz, I would support turning MARC-in-JSON on by default.

demiankatz:
@EreMaijala, I've added commit 1304122 to change the default, but give an easy way to roll back via marc_local.properties. What do you think?

demiankatz:
@he247, the fix to tolerate broken binary MARC records has been backported to VuFind 8 (and was released as part of 8.0.4 today). This isn't perfect, but it restores VuFind to the behavior it had in releases 7.x and before. The part that is more complex to backport is the addition of JSON support, since that changes some internals of classes in a backward-compatibility-breaking way.

EreMaijala:
@demiankatz The patch looks good, but maybe a short comment about the length limit with FullRecordAsMarc might be useful?

demiankatz:
Okay, done in dc0c3da. If that looks good to you, I think we can merge this. I'll add a note to the changelog about the change, since indexing records with this new setting will break compatibility between the Solr index and earlier versions of VuFind.

demiankatz (approved):

he247:
@demiankatz, yes I already merged the changes into our system. :)

What I meant with an "official patch" was a separate, extracted branch. But at second thought: it might not be that practical.

EreMaijala:
@demiankatz Looks good to me!

PR #2335, opened 2022-03-01
Update API spec to OpenAPI 3.0.3, update Swagger UI

EreMaijala:
I propose we switch from the pretty old Swagger spec to OpenAPI 3. What do you think?


EreMaijala:
Note: While OpenAPI 3.1.0 has been published, it's not yet recognized by Swagger UI, so I believe it's better to keep with the older and more supported 3.0.3.

demiankatz:
Thanks, @EreMaijala -- I'll review this in more detail when I have a bit more time (busy day today!). One quick question in the meantime, though: does this impact the API in any way for end users, or is this just an internal rearrangement? I'm just wondering if there are any documentation or backward-compatibility implications we need to worry about.

EreMaijala:
@demiankatz This does not affect the functionality. Only the spec loaded with /api?swagger or now /api?openapi will return a newer version. It could affect someone using a separate very old version of Swagger UI or an old code generator. However, these are typically used once when starting to use an API, not later.

demiankatz (approved):
Looks good to me, and works fine in my test environment.

PR #2340, opened 2022-03-02
Add custom filter listener.

demiankatz:
This PR makes it possible to represent complex filters using simpler placeholders in the URL. It also creates the ability to have "inverted" filters, where filters are applied to Solr UNLESS a specific filter value is applied in the URL. Both of these features are particularly useful in combination with checkbox facets.

It's entirely possible that there is room for improvement in the configuration layout and comments; I'm interested in feedback on these (and anything else).

TODO:
- [x] Write tests
- [x] Discuss configuration format


module/VuFind/src/VuFind/Search/Solr/CustomFilterListener.php
```diff
@@ -0,0 +1,155 @@
...
+     * @var array
+     */
+    protected $invertedFilters;
+
+    /**
+     * Name of parameter used to store filters
+     *
+     * @var string
+     */
+    protected $filterParam = 'fq';
```
demiankatz, line 77: I put this into a property because I think we could theoretically move most of the logic to an abstract base class and then have separate implementations for other search backends like Summon.

demiankatz:
@EreMaijala, I believe this is ready for review now. There are a lot of different ways we could structure the configuration, and I'm open to alternative ideas if you don't like what I came up with here. I tried to separate the behavior of the listener from the structure of the configuration to keep things flexible.

config/vufind/facets.ini
```diff
@@ -304,3 +304,31 @@ visual_facets = "callnumber-first,topic_facet"
...
+; in combination with custom_filter_field in VuFind URLs, and the values are
+; the actual filters applied to Solr.
+;translated_filters[book_or_journal] = 'format:("Book" OR "Journal")'
+
+; The inverted_filters setting is formatted like translated_filters, but the
+; filtering behavior is inverted. These filters will ALWAYS be applied to Solr
+; UNLESS the custom filters are applied by the user. This is useful in
+; combination with checkbox facets if you want to exclude values by default
+; and provide a checkbox that allows them to be included.
+;inverted_filters[include_books] = '-format:"Book"'
```
EreMaijala: Missing newline at EOF

config/vufind/facets.ini
```diff
@@ -304,3 +304,31 @@ visual_facets = "callnumber-first,topic_facet"
...
+; This can be useful in combination with checkbox filters and hidden filters
+; if you wish to define filters that are too complex to easily represent in
+; .ini file syntax, and it also makes URLs more readable and easier to share.
+[CustomFilters]
+; This is the name of the virtual Solr field used when applying custom
+; filters. For example, if you define an inverted filter named include_books
+; below, and your custom filter field was the default of __custom__, you
+; would add &filter[]=__custom__:include_books to the URL in order to apply
+; the inverted filter.
+custom_filter_field = "__custom__"
```
EreMaijala: Would it be possible to use "vufind" as the default custom filter field? It would look nicer. Also, it would be nice to include an example of how to set up a checkbox filter to use the field, and in its configuration 'vufind' would look nicer and be easier to get right than double underscore.

module/VuFind/src/VuFind/Search/Solr/CustomFilterListener.php
EreMaijala: Ok. This reminds me.. we really should try to move more of the backend-specific stuff to VuFindSearch, I think.

EreMaijala (changes requested):
Looks good to me, just some minor niggles.

module/VuFind/src/VuFind/Search/Solr/CustomFilterListener.php
demiankatz: I think the original intention of putting the listeners in VuFind instead of VuFindSearch was that the listeners were implementing functionality specific to VuFind rather than generic to backends... but that may no longer be true of all of the listeners, or we may no longer care about that distinction. Certainly happy to revisit. :-)

config/vufind/facets.ini
demiankatz: Fixed.

config/vufind/facets.ini
demiankatz: Agreed. I have made the change (and added the example... though the comment on the example is perhaps too short because I didn't want to go multi-line; open to improvements if you have any suggestions).

demiankatz: ...actually, the comment was bothering me, so I revised it. Hopefully better now!

demiankatz:
Thanks, @EreMaijala -- I've addressed your review comments and also fixed a small bug that was turning up when running the full test suite (missing configuration was causing a notice when building the SolrAuth backend).

module/VuFind/src/VuFind/Search/Solr/CustomFilterListener.php
EreMaijala: What I meant is that e.g. 'fq' is pretty specific to how the url query syntax with Solr currently works. There's been at least some level of agreement that the legacy facets api that we use should be deprecated with Solr 9, which probably means that in the long run we need to switch to the JSON APIs, for all queries. Then the 'fq' in VuFind classes becomes a burden. So, while I think it's fine that a listener is specific to a backend, I'd like the interface between VuFind and VuFindSearch to be more backend-agnostic. It's for sure not this pull request that needs to address it. :)

EreMaijala (approved):
Looks good to me!

module/VuFind/src/VuFind/Search/Solr/CustomFilterListener.php
demiankatz: That's a good point, though I don't think there's an easy solution to this problem. The original design decision was to have the contents of the ParamBag passed to the search backend be backend-specific, and to have the Params classes take care of the primary mapping of VuFind parameters to backend parameters. Of course, custom listeners live somewhat outside of this piece, but in the end, I think the problem here is deciding what needs to be given a VuFind-level abstraction, and where we put the mapping logic. Maybe it would be helpful to make facets more abstract so that more code can be shared across backends, but given the inconsistencies in implementations on different platforms (particular with regard to things like AND/OR/NOT facets), that may not be practical. And even if we do decide to make facets uniform, there are going to be other backend-specific features that perhaps do not make sense to develop abstractions for, because the use cases are too narrow. Anyway, you rightly point out that we need to work on the JSON facets implementation, and perhaps that work will inspire our next step forward -- that certainly seems like a good context in which to continue this conversation.

PR #2342, opened 2022-03-02
Add CSRF check to SMS functionality to prevent abuse.

demiankatz:
There have been reports of abuse of the SMS functionality. In addition to the existing possibility of turning on CAPTCHA, I thought a built-in CSRF check would also be helpful.

TODO
- [x] Run full test suite


crhallberg (approved):

demiankatz:
Thanks, @crhallberg!

EreMaijala (commented):
For whetever it's worth, this looks good to me too. Not sure if it would deter a more determined spammer, but at least it will prevent simple requests.

PR #2343, opened 2022-03-03
Add message if the browser doesn't support es6.

LuomaJuha:
Added few common things, which are new in es6 to be checked. Had no idea if that is the proper place to put the less style, but was closest which I could find.


languages/en.ini
```diff
@@ -893,6 +893,7 @@ Other Sources = "Other Sources"
 other_versions_link = "Show other versions (%%count%%)"
 other_versions_search_link = "Show all versions (%%count%%)"
 other_versions_title = "Other Versions (%%count%%)"
+outdated_browser_message_html = "<h2>Your are using outdated and unsupported browser.</h2><p>To ensure information pricacy and full functionality of site, we recommend you update your browser.</p>"
```
demiankatz: Typo: "Your are" instead of "You are". Might also be good to add "the" in the phrase "full functionality of site".

demiankatz (changes requested):
Thanks, @LuomaJuha. A couple of small notes on the language text. I see also that LGTM had some comments about the Javascript which should probably be addressed. I'll ask @crhallberg to weigh in about the LESS (and anything else he wishes to review).

languages/en.ini
LuomaJuha: Oh, thanks!

LuomaJuha:
@demiankatz Oh, its the bot telling about unused variables. Gonna address that issue.

languages/en.ini
LuomaJuha: This one has been now addressed

themes/bootstrap3/less/vendor/bootstrap/alerts.less
```diff
@@ -34,6 +34,19 @@
...
+  background-color: @state-warning-bg;
+  padding-top: 0.6em;
+  padding-bottom: 0.6em;
+  text-align: center;
+  z-index: 1;
+  @media (max-width: @grid-float-breakpoint) {
+    position: relative;
+    top: @navbar-height;
+  }
+}
```
crhallberg, line 48: The mobile styling overlaps the content due to the use of `top: 50px;`. Would it make things simpler and clearer if we use the stylings from [`.alert.alert-warning`](https://getbootstrap.com/docs/3.4/components/#alerts-examples)?

image

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -116,6 +116,9 @@ JS;
       <?php endif; ?>
       <a class="sr-only" href="#content"><?=$this->transEsc('Skip to content') ?></a>
       <?=$this->render('header.phtml')?>
+      <div id="browser-outdated" class="container-fluid system-message hidden">
+        <?=$this->translate('outdated_browser_message_html');?>
+      </div>
```
crhallberg, line 121: It's concerning to me that this message will be on every page without an option to dismiss. Also, moving or removing this message will require changing `layout.phtml`. Can we add a configuration option?

themes/bootstrap3/js/common.js
```diff
@@ -479,7 +479,43 @@ function setupQRCodeLinks(_container) {
...
+    if (outdated) {
+      outdated.classList.remove('hidden');
+    }
+  }
+  /* eslint-enable */
+  /* jshint ignore:end */
+}
+
 $(document).ready(function commonDocReady() {
+  checkForBrowserSupport();
```
crhallberg, line 518: I tried this out in [Firefox 43](https://filehippo.com/download_mozilla-firefox-64/43.0.4/) ([two versions before ES6 support](https://caniuse.com/es6-class)) and this check didn't parse, which broke all of `common.js` and led to the VuFind object not existing.

image

crhallberg: I actually had to change a few `let variable`s to `var variable`s to get this far.

crhallberg (changes requested):

themes/bootstrap3/js/common.js
LuomaJuha: Going to test it with the same browser. But thats odd.

themes/bootstrap3/less/vendor/bootstrap/alerts.less
LuomaJuha: I can try that one too!

LuomaJuha:
This seems to be too tricky to implement reasonably.

demiankatz:
Thanks for making the attempt, @LuomaJuha -- it's a good idea, but perhaps not worth the effort. Let's hope that non-ES6 browsers finish disappearing before too much longer. :-)

he247:
Dear @demiankatz, we added such a warning in our system based on https://github.com/exiguus/js.device.detector.git. If there is still some interest, we basically add the error message via javascript:

```js
$.ulbfind.detectBrowserSupport = function () {
  /* detect Browser and show message if not fully supported */
  // setup local storage to track time of life for Message suppression - currently hardcoded 10 days
  var today = Date.now();
  var timeOfLife = 10;
  if (localStorage.getItem('showBrowserUnsupportedDate') === null)
  {
    localStorage.setItem('showBrowserUnsupportedDate',today);
  }
  if((localStorage.getItem('showBrowserUnsupported') !== 'false')
    || localStorage.getItem('showBrowserUnsupported') === null
    || (today - localStorage.getItem('showBrowserUnsupportedDate')) / (1000 * 60 * 60 * 24) >= timeOfLife)
  {
    localStorage.setItem('showBrowserUnsupported', 'true');
  }
  // init device.detector
  // custom browser matrix - Browsers that fully support ES6
  $.fn.deviceDetector(
    {
      'supports': {
        'msie': {'id': 'msie', 'version': '-'},
        'edge': {'id': 'edge', 'version': '15'},
        'chrome': {'id': 'chrome', 'version': '51'},
        'firefox': {'id': 'firefox', 'version': '54'},
        'safari': {'id': 'safari', 'version': '10'},
        'opera': {'id': 'opera', 'version': '38'},
      },
    }
  );
  var d = $.fn.deviceDetector;
  // if browser not supported, show alert
  if(!d.isSupported()){
    if(localStorage.getItem('showBrowserUnsupported') === 'true') {
      $("#content").prepend('<div id="browser-alert">' +
          '<div class="alert alert-danger alert-dismissible">' +
            VuFind.translate('limited_browser_support') +
            '<button type="button" class="close" data-dismiss="alert" aria-label="Close">' +
              '<span aria-hidden="true">×</span>' +
            '</button>' +
          '</div>' +
        '</div>');
      $('#browser-alert > .alert').on('closed.bs.alert', function () {
        localStorage.setItem('showBrowserUnsupported', 'false');
        localStorage.setItem('showBrowserUnsupportedDate',today);
      });
    }
  }
};
```

demiankatz:
Thanks for sharing this, @he247 -- if you, @LuomaJuha or @crhallberg are interested in expanding this comment into a new PR, I'm certainly happy to take another look at it. I think this feature is a good idea if we can make it work consistently, but it is likely to require ongoing maintenance to keep accurate.

he247:
@demiankatz If it won't be done before I'll create a new PR as soon as I have time to adopt our changes for deployment.

PS: In this case I'm not sure how to do the automated testing, do you have some standards for javascript testing similar to the php unit tests?

demiankatz:
@he247, thank you!

Regarding automated testing, we currently only exercise our Javascript code through our Mink integration tests. Since those are always run through Chrome at the moment, I don't think there's any obvious way to test the browser warning (except possibly to create a test to confirm that it does NOT appear). I think it could be useful to add Javascript unit tests using something like [jest](https://jestjs.io/), which I have very much enjoyed using in other projects; perhaps that is something that @crhallberg can consider as part of theme redesign/retooling work.

PR #2344, opened 2022-03-03
Index entire title from 246, not just short/proper part.

demiankatz:
I recently discovered that our default SolrMarc settings were only importing 246a (the "title proper/short title") without 246b (the "remainder of title") into the title_alt field. It seems useful to be able to search for the subtitle from alternate title fields, so I propose that we expand the definition.


demiankatz:
This is a pretty trivial (and I suspect harmless/uncontroversial) change, but I need somebody to approve it. I'm wondering if @todolson or someone on his team might have an opinion. I know that @EreMaijala doesn't use SolrMarc, but I'm interested in whether or not RecordManager is already doing this as another point of comparison.

todolson (approved):
I think this is perfectly fine, and likely of general benefit.

In our local configuration, we actually grab all (alphabetic) subfields for fields in `title_alt`:
```title_alt = custom, getAllSubfields(210:242:246:247:740:780:785:787, " ")```
and then in our `title_browse` we use `246abnp` (plus other fields).
I'm not recommending this as the project default, just illustrating that we've found it useful to include more than 246‡a.

EreMaijala (approved):
This makes sense to me. I'd consider adding subfields n and p as well since those can be quite important for parts.

EreMaijala:
@demiankatz RecordManager base version seems to have had this wrong for the past nine years since I broke it in a refactoring somehow to just take subfield g, which makes no sense. I hadn't noticed since our local version takes a+b+n+p. Will fix, of course. :)

demiankatz:
Thanks, @EreMaijala, I've revised to include n+p as well. As you say, they may also include useful searchable information, and that is also compatible with what @todolson is doing locally. I think this is a useful but unintrusive change, and of course I'd be open to discussing adding even more in future if there's strong interest.

PR #2350, opened 2022-03-15
Fix bug with Solr NOT handling.

demiankatz:
TODO
- [x] Add Mink test for advanced search with NOT
- [x] Test thoroughly
- [x] Resolve [VUFIND-1537](https://vufind.org/jira/browse/VUFIND-1537) when merging

### DIFF
--- a/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AdvancedSearchTest.php
+++ b/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AdvancedSearchTest.php
@@ -191,11 +191,12 @@ class AdvancedSearchTest extends \VuFindTest\Integration\MinkTestCase
     }
 
     /**
-     * Test that the advanced search form works correctly with a NOT group
+     * Test that the advanced search form works correctly with a NOT group combined
+     * with another group.
      *
      * @return void
      */
-    public function testAdvancedSearchWithNotOperator()
+    public function testAdvancedMultiGroupSearchWithNotOperator()
     {
         $session = $this->getMinkSession();
         $page = $this->goToAdvancedSearch($session);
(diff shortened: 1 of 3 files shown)


module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AdvancedSearchTest.php
```diff
@@ -190,6 +190,40 @@ public function testAdvancedSearch()
...
+     *
+     * @return void
+     */
+    public function testAdvancedSearchWithNotOperator()
+    {
+        $session = $this->getMinkSession();
+        $page = $this->goToAdvancedSearch($session);
+
+        // Add a group
+        $session->executeScript("addGroup()");
```
EreMaijala, line 205: I think it would be useful to also test a case where we only have a single NOT group.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/QueryBuilder.php
```diff
@@ -460,18 +460,17 @@ protected function reduceQueryGroupComponents(AbstractQuery $component)
...
                     return '' !== $s;
                 }
             );
-            if ($reduced) {
-                $searchString .= sprintf(
-                    '(%s)',
-                    implode(" {$component->getOperator()} ", $reduced)
-                );
+            $searchString = $reduced
+                ? sprintf('(%s)', implode(" {$component->getOperator()} ", $reduced))
```
EreMaijala: Since we usually just do concatenation instead of sprintf, I'd favor it here as well. Needs extra parenthesis, but still.

EreMaijala (changes requested):

demiankatz:
Thanks for the review, @EreMaijala -- see [affc09a](https://github.com/vufind-org/vufind/pull/2350/commits/affc09a7f424972a360848984b37ef7613e57fda); I think this should address your comments. I'm currently running the test suite and do not expect it to fail... but if any surprise problems pop up, I'll fix them shortly.

EreMaijala (approved):

PR #2352, opened 2022-03-16
Add mechanism for combining raw and translated facet values.

mtrojan-ub:


demiankatz:
Thanks, @mtrojan-ub! A few questions/comments:

1.) The continuous integration check is failing -- it looks like perhaps there are some minor errors in some of the files (maybe duplicate/incorrect lines?).

2.) We should probably include an example configuration to leverage these in facets.ini. We should also make sure that the indexing configuration aligns with the language files.

3.) Should the translations themselves include the numbers? Mapping the number to the description without actually showing the number may be confusing.

mtrojan-ub:
1.) This is fixed now, there have been 2 duplicate keys which have been corrected.

2.)+3) An easy proof-of-concept would be to add the following lines to facets.ini:
```
[Results]
dewey-raw = "Call Number"

[Advanced_Settings]
translated_facets[] = dewey-raw:DDC23
```

However, if somebody wants to use dewey-hundreds only for a translated facet, he would have to change the indexing to omit the ddc23_map.properties in the corresponding solr field. Thereby the plain text information would be lost, which might come in handy to search a category via full text. On the other hand, the current solution only provides the english translation in the solr index... so if a text search functionality would be needed, the other languages would need to be indexed too anyway... So the question is which approach we prefer? (a) only index the number or b) index all languages for search)

For the display in the facets, to me it sounds like adding the key to the displayed text is more of a formatting issue. Another option to solve this would be to change the syntax of the translated_facets INI entry to allow some kind of string formatting that might also include placeholders for the initial value and the translated value, here's just a first example:

translated_facets[] = "dewey-raw:DDC23/%untranslated%: %translated%"

What do you think?

demiankatz:
Thanks for fixing the duplicate keys!

Regarding changing the indexing to only include numbers rather than text, while you're correct that this theoretically limits the searchability of the Dewey fields, I think in practice it makes no real difference, because the dewey fields in the Solr index are string fields that aren't keyword searchable, and none of that text is copied into allfields or other text-searchable fields... If we wanted to make that text searchable, we would have to make changes to the current approach anyway, and there's nothing stopping us from using the map to do that in the future if we wished.

Another question is whether the descriptions/translations of the numbers need to change based on the granularity of the indexing -- i.e. if we're only indexing the hundreds, do we use a broader description for an x00 call number than we would for that same number when indexing down to the ones level? (I spot checked a few, and I couldn't find any differences, so hopefully this is a non-issue... but it occurred to me that there might be a reason we have separate ones/tens/hundreds maps in our translation map at present).

Regarding your suggestion of adding a rule for combining raw and translated values in the facet display, that does seem like it would be useful in this particular situation (and perhaps some other ones, like if we standardized indexing of languages to use ISO codes instead of names). The next question then becomes: do we need to work on that as a separate PR before moving this forward, or do we want to reformat the language files for now and address that in the future? I think the answer entirely depends on community interest and bandwidth. :-)

mtrojan-ub:
Thanks for your feedback!

I added a change to the translated_facets ini setting. Do you think we also need to add a test case for this one?
Also, i don't think it's necessary to add translations based on the granularity.
So, does that mean we could also remove the ddc23_map.properties entirely (also from marc.properties) and switch over entirely to the new mechanism?

demiankatz:
Thanks, @mtrojan-ub! It never hurts to add test coverage, but if there is not already well-established test infrastructure for the affected code, let me know and maybe I can help set something up that we can build upon here.

As for removing ddc23_map.properties, maybe thing to do is to introduce this as an option in release 8.1 and then change the defaults in 9.0. I don't want to introduce a backward-breaking change in a minor release, but I think in the long term this is a better approach.

mtrojan-ub:
Sounds good! So i couldn't find any test that's related to the translated_facets directly so far. Could you please help to set something up?

And to wait until 9.0 with the ddc_map.properties is absolutely fine!

module/VuFind/src/VuFind/Search/Solr/Results.php
```diff
@@ -349,6 +352,28 @@ public function getFacetList($filter = null)
...
+                    // Apply a format to the translation (if available)
+                    if ($translateFormat) {
+                        $translatedValue = $translateFormat;
+                        $translatedValue = str_replace(
+                            '%raw%',
+                            $displayText,
+                            $translatedValue
+                        );
+                        $translatedValue = str_replace(
+                            '%translated%',
```
demiankatz: You can do all the replacements in a single step if you pass arrays to the first two parameters of `str_replace` -- i.e. `str_replace(['%raw%', '%translated%'], [$displayText, $translated], $translatedValue);`

module/VuFind/src/VuFind/Search/Solr/Results.php
```diff
@@ -349,6 +352,28 @@ public function getFacetList($filter = null)
...
                     ? $this->translate([$translateTextDomain, $displayText])
                     : $displayText;
+                $currentSettings['displayText'] = $displayText;
+                if ($translate) {
+                    $translated = $this->translate(
+                        [$translateTextDomain, $displayText]
+                    );
+                    // Apply a format to the translation (if available)
+                    if ($translateFormat) {
+                        $translatedValue = $translateFormat;
```
demiankatz: I don't think this line serves any purpose, since it is immediately overwritten.

demiankatz:
Thanks, @mtrojan-ub, it does look like we need to establish a unit test for \VuFind\Search\Solr\Results. I'm having a busy week (... month ... year) but will see if I can find some time to do that in the coming days. In the meantime, I did a closer review of your code and I think there's some opportunity for simplification; I left a couple of comments.

mtrojan-ub:
Thanks for creating the basic ResultsTest! I'm trying to apply that onto the facet translation changes in this PR, however i'm not sure how to create a dummy result for which the translation will be applied and that actually contains translated facets that i could test against. Could you please have a look?

mtrojan-ub:
Thanks for creating the basic ResultsTest! I'm trying to apply that onto the facet translation changes in this PR, however i'm not sure how to create a dummy result for which the translation will be applied and that actually contains translated facets that i could test against. Could you please have a look?

module/VuFind/src/VuFind/Search/Solr/Results.php
```diff
@@ -349,6 +352,21 @@ public function getFacetList($filter = null)
                 $currentSettings['displayText'] = $translate
                     ? $this->translate([$translateTextDomain, $displayText])
                     : $displayText;
+                $currentSettings['displayText'] = $displayText;
```
demiankatz: You are overwriting `$currentSettings['displayText']` here -- I believe that the above three lines need to be removed to avoid performing duplicate work.

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
```diff
@@ -59,6 +60,57 @@ public function testCursorMark(): void
...
 
+    /**
+     * Test facet translation functionality.
+     *
+     * @return void
+     */
+    public function testFacetTranslations(): void
+    {
+        $mockTranslator = $this->createMock(TranslatorInterface::class);
+        $mockTranslator->expects($this->exactly(2))
```
demiankatz, line 78: `$this->exactly(2)` should be changed to `$this->once()` when the bug I just noted is fixed... I had to put it in this way to make the test pass for now.

demiankatz (changes requested):
@mtrojan-ub, thanks for getting things started. I believe I've filled in the missing details. In the process, I found a bug -- see notes below.

module/VuFind/src/VuFind/Search/Solr/Results.php
mtrojan-ub: Good catch! And many thanks for all the help with the test cases!

demiankatz: Happy to help!

module/VuFind/src/VuFind/Search/Solr/Results.php
```diff
@@ -346,9 +349,21 @@ public function getFacetList($filter = null)
...
-                    ? $this->translate([$translateTextDomain, $displayText])
-                    : $displayText;
+                $currentSettings['displayText'] = $displayText;
+                if ($translate) {
+                    $translated = $this->translate(
+                        [$translateTextDomain, $displayText]
+                    );
+                    // Apply a format to the translation (if available)
+                    if ($translateFormat) {
+                        $translated = str_replace(
```
demiankatz: What if, instead of doing a `str_replace` here, we use this instead to take advantage of our standard translation token system?

```
$translated = $this->translate(
    $translateFormat,
    ['%%raw%%' => $displayText, '%%translated%%" => $translated]
);
```

Now, the format descriptor can be a language file token which incorporates %%raw%% and/or %%translated%%, and this enables us to use different formats for different languages if necessary. I'm pretty sure that including the full raw format string in the .ini file instead of putting it in the language files would also still work, though I'd want to double check that to be sure I'm not forgetting some important detail.

(Note that I switched to double-percent-signs on the tokens for consistency with other tokenized language strings).

config/vufind/facets.ini
```diff
@@ -230,9 +230,13 @@ special_facets   = "illustrated,daterange"
...
 ; name of a text domain after the field name to specify translations in a specific
 ; text domain (subdirectory of the languages folder). If you comment out the lines
-; below, no facets will be translated.
+; below, no facets will be translated. If you add a second colon, you may add a
+; format descriptor containing the placeholders %raw% and %translated%. This can
+; be useful if the facet is related to a classification like e.g. dewey and
+; you would like to show the raw value together with the translation.
 ;translated_facets[] = institution
 ;translated_facets[] = building
+;translated_facets[] = dewey-raw:DDC23:%raw% - %translated%
```
demiankatz: I don't think `dewey-raw` actually makes sense as the example here, since that contains the full call number, not just the first three digits. Maybe it would be better to set up a slightly larger example with a prefixing comment, like:

```
; If you change the default Dewey indexing to omit translation mapping at index time,
; you can uncomment the below configuration to take advantage of on-the-fly
; translation into multiple languages.
;translated_facets[] = dewey-ones:DDC23:%raw% - %translated%
;translated_facets[] = dewey-tens:DDC23:%raw% - %translated%
;translated_facets[] = dewey-hundreds:DDC23:%raw% - %translated%
```

demiankatz (commented):
Thanks, @mtrojan-ub! I've left a couple more comments, but I think this is getting close to being finished. It might also be worth requesting a review from @EreMaijala before merging this in case he has additional thoughts, since he's been working on code in some of the same areas.

demiankatz:
@mtrojan-ub, I've also renamed this PR, because at this point I think the new mechanism is actually a bigger change than the inclusion of a DDC23 example. :-)

module/VuFind/src/VuFind/Search/Solr/Results.php
```diff
@@ -346,9 +349,21 @@ public function getFacetList($filter = null)
                     $displayText = $this->hierarchicalFacetHelper
                         ->formatDisplayText($displayText);
                 }
-                $currentSettings['displayText'] = $translate
-                    ? $this->translate([$translateTextDomain, $displayText])
-                    : $displayText;
+                $currentSettings['displayText'] = $displayText;
```
EreMaijala: I think this would be better as an `else` branch to the `if ($translate)` branch below.

EreMaijala (dismissed):
All in all this looks good to me. Is there ever a case where would want this for hierarchical facets? That'd open a new can of worms, but I can't think of a proper use case.

This said, I'm not sure why this is really necessary since you could just include the codes in the translation strings. Is it that you want to keep the translations clean for other uses? You could still have a separate facet-specific translation file. I'm asking because I feel that the facet translation is getting a bit complex with a stack of translations.

demiankatz:
@EreMaijala, I think the argument for this is the idea that some users might want to include the code in the facet display, and others might not. If we hard-code it into every line of the language files, it's a lot more work to change/customize than if we format it after the fact with a single translation string. But I don't have particularly strong feelings about how likely/common that scenario is, or how to balance it against the increase in code complexity.

EreMaijala:
@demiankatz If it happens with a lot of varying facets, I get it, but for something relatively static I'd find it easier to tell people to use one of two text domain (for example DDC23 and DDC23WithCodes). The latter could be created with a regex replace from the former. But I have no strong opinion on this, either way is fine. The thing I'd want changed is the potentially duplicate assignment to `$currentSettings['displayText']`.

demiankatz:
@EreMaijala, I could imagine this happening anywhere that there are codes used -- so some other situations might be languages (if we eventually switch to indexing by ISO code instead of English text -- I could envision some wanting to use a format like "English (en)") and relator codes (if we ever wanted to facet by relator... which may be a fairly far-fetched idea). So I'm still on the side of seeing some potential use cases, but not finding them incredibly strong/compelling. I guess we can wait and see how strongly @mtrojan-ub feels about this!

module/VuFind/src/VuFind/Search/Solr/Results.php
mtrojan-ub: Done!

mtrojan-ub:
Thanks for your feedback!

I think there are multiple reasons for "not" including codes in the translations:
- Have flexibility for the admin to allow stripping of the code with configuration (without writing any code) => we're planning to use this for multiple kinds of customized facets as well.
- If you ever want to change the style of the delimiter, you can do that without searching/replacing all over your INI files.
- Reduce data overhead (just a basic DB design principle - it's always good to store information as granular as possible, you never know whether you can use it in a different way tomorrow that you can't think of today. And then it's always better to concatenate granular information than splitting/parsing existing strings with e.g. regexes, because that's not only annoying but also usually costs more performance than concatenating.)
- We have a translation workflow attached to some of the facets and for some of them, the IDs will consist of letters. The less information a translator can edit, the less can go wrong. And actually, cost calculation for translators is often done by counting words, so you also might save some money this way.

Maybe i think too much as a DB designer here :sweat_smile: but i'd really prefer to have the codes out of the strings.

demiankatz:
This all makes sense to me, @mtrojan-ub. I'll wait to see if @EreMaijala has anything to add, but if he has no major objections, I'll do a final round of testing and get this merged. I do feel that from a maintenance perspective, these few lines of code are probably less of a burden then maintaining multiple text domains to offer alternatives... :-)

config/vufind/facets.ini
```diff
@@ -230,11 +230,20 @@ special_facets   = "illustrated,daterange"
 ; language files found in the languages directory. You may add a colon and the
 ; name of a text domain after the field name to specify translations in a specific
 ; text domain (subdirectory of the languages folder). If you comment out the lines
-; below, no facets will be translated.
+; below, no facets will be translated. If you add a second colon, you may add a
+; format descriptor containing the placeholders %%raw%% and %%translated%%. This can
```
demiankatz: I've tested that it now works to either directly place the format descriptor here, or to use a translation string. We should probably adjust this documentation accordingly. Maybe something like:

If you add a second colon, you may add a translation string containing the placeholders %%raw%% and %%translated%%. You can enter the translation string directly (if the formatting is uniform for all languages), or you can use a key pointing to values found in the language files (if you need different formatting for different languages).

(...and then the text can continue as it presently exists).

What do you think?

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
```diff
@@ -59,6 +60,60 @@ public function testCursorMark(): void
...
```
demiankatz, line 78: @mtrojan-ub, I just made a couple of commits to make this part of the test more specific and more realistic; the mock was prematurely producing the final expected value in a previous revision, but now it should be working more like it's supposed to. :-)

EreMaijala:
@demiankatz @mtrojan-ub I'll accept the reasons for this, no further questions. :)

demiankatz:
Thanks, @EreMaijala. I think at this point, once @mtrojan-ub has a chance to review my latest comments (regarding config comment and unit test) we should be in a good position to finalize and merge this.

config/vufind/facets.ini
mtrojan-ub: Sounds good to me!

mtrojan-ub:
Thanks, everything looks good to me! Feel free to merge :blush:

config/vufind/facets.ini
demiankatz: Thanks! I've incorporated the proposed text and done a bit of reformatting to improve the readability of the documentation comment.

demiankatz (approved):

PR #2341, opened 2022-03-02
Add Blender for searching two or more backends and merging results

EreMaijala:
Blender can display results from multiple backends in a single list. It can also merge facets and map active filters for each backend. 

Note that I've long been trying to avoid doing this, but a blended search has been so often requested that I've had to yield. There's no real magic with the results. It's mostly round-robin with the possibility to boost some records on the initial results list. Actually, the merged facets are kinda magical. :)

TODO: 

- [x] Document (mappings) configuration better. Initial documentation now https://vufind.org/wiki/configuration:blender
- [x] Add an example for hierarchical format facet.
- [x] Verify that labels display properly with EDS
- [x] Testing
- [x] Unit tests
- [x] Integration tests
- [x] Maybe extract some of the RecordInterface/RecordTrait refactoring to a separate commit?
- [x] Add caching support to at least Primo and EDS backends (now in #2406)

PR #2356, opened 2022-03-18
Fix Primo Options constructor to set ini files first.

EreMaijala:
This makes the base class constructor load correct settings.


EreMaijala:
Noticed when working on #2341.

demiankatz (approved):

PR #2357, opened 2022-03-18
Add support for translating tokens in flash messages.

EreMaijala:
Spun off from #2341.


demiankatz (changes requested):
This is complicated enough that we probably ought to add test coverage to be sure all the permutations are covered. I realize we don't have a FlashmessagesTest yet, but maybe this is the time to establish one. If you don't have time, I might be able to squeeze it in next week.

demiankatz:
Also, do you think it might be worth rebasing this against dev instead of dev-9.0? This seems like a new feature rather than a BC break, so maybe it would be better to have it available sooner (especially if we're adding test coverage as part of the package).

EreMaijala:
True. I'll see what I can do.

EreMaijala:
Sure, dev is fine as well.

EreMaijala:
Continued in #2359.

PR #2358, opened 2022-03-18
Fix duplicate checkbox display when using multiple SideFacets instances.

demiankatz:
If you use multiple instances of the SideFacets recommendation module, and some have checkbox filters configured while others do not, duplicate checkboxes will appear. This PR fixes this problem, only displaying checkboxes when they are configured to appear.

TODO
- [x] Add test coverage


demiankatz:
I've added test coverage for this situation, and while I was at it, I did a bit of formatting and modernization work on the existing test class for SideFacets, as it was a bit messy.

EreMaijala (approved):
Looks good to me and tests are passing. Thanks for cleaning up the tests as well!

PR #2359, opened 2022-03-18
Add token translation support and tests to Flashmessages

EreMaijala:
Supercedes #2357. Spun off from #2341.

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/Flashmessages.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/Flashmessages.php
@@ -102,30 +102,38 @@ class Flashmessages extends AbstractHelper
                 $html .= '>';
                 // Advanced form:
                 if (is_array($msg)) {
-                    // Use a different translate helper depending on whether
-                    // or not we're in HTML mode.
-                    if (!isset($msg['translate']) || $msg['translate']) {
-                        $helper = (isset($msg['html']) && $msg['html'])
-                            ? 'translate' : 'transEsc';
-                    } else {
-                        $helper = (isset($msg['html']) && $msg['html'])
-                            ? false : 'escapeHtml';
+                    $msgHtml = $msg['html'] ?? false;
+                    $message = $msg['msg'];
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/View/Helper/Root/Flashmessages.php
```diff
@@ -111,9 +111,18 @@ public function __invoke()
...
                             ? false : 'escapeHtml';
                     }
-                    $helper = $helper
-                        ? $this->getView()->plugin($helper) : false;
+                    $helper = $helper ? $this->getView()->plugin($helper) : false;
                     $tokens = $msg['tokens'] ?? [];
+                    if ($tokens && $mode = ($msg['translateTokens'] ?? false)) {
+                        // Escape translated tokens unless html is requested or the
+                        // main message is translated:
+                        $translator = 'html' === $mode || empty($msg['html'])
```
demiankatz: This line of code has me really confused, though if tests are passing, maybe I'm just totally misunderstanding what I'm looking at... but why are we comparing `$mode` (which appears to be a boolean) with the string `'html'`? Why is `$msg['html']` being empty (as opposed to non-empty) causing us to go into non-escaping translation mode? Am I misreading the operator associativity or something, or is something wrong here?

demiankatz: Also, if we're going to use `empty($msg['html'])` in one place, and `(isset($msg['html']) && $msg['html'])` in others, maybe this is a sign that we should capture this value in a variable for consistency and improved readability.

EreMaijala: It was a bit messy, and coupling translation and escaping of tokens perhaps wasn't the best idea ever. I've now decoupled them and rewritten the part. Hope it makes more sense now!

demiankatz (approved):

demiankatz:
Thanks, @EreMaijala, this is much more clear now!

PR #2360, opened 2022-03-22
Fix minor problems in example theme.

demiankatz:
- Missing escaping of one URL
- Fatal error if DevTools module is disabled (e.g. in production mode)


demiankatz:
@crhallberg, while testing #2331 by looking at the local_theme_example theme, I ran into a very puzzling problem: I got a fatal error on the home page, but when I turned on development mode to investigate, everything was fine. A bit of trial and error revealed the problem: the example configuration refers to DevTools routes, and those routes don't exist in production mode, because the module is only included conditionally. This PR makes the template smarter about missing routes, and just hides things that are not available, thus avoiding the fatal error problem.

crhallberg (approved):

PR #2361, opened 2022-03-22
Separate record source identifier and search backend identifier.

EreMaijala:
This is a pre-requisite for blended search where the search backend returns records for multiple real backends.

### DIFF
--- a/module/VuFind/src/VuFind/RecordDriver/AbstractBase.php
+++ b/module/VuFind/src/VuFind/RecordDriver/AbstractBase.php
@@ -46,13 +46,7 @@ abstract class AbstractBase implements \VuFind\Db\Table\DbTableAwareInterface,
 {
     use \VuFind\Db\Table\DbTableAwareTrait;
     use \VuFind\I18n\Translator\TranslatorAwareTrait;
-
-    /**
-     * Used for identifying search backends
-     *
-     * @var string
-     */
-    protected $sourceIdentifier = 'Solr';
+    use \VuFindSearch\Response\RecordTrait;
 
     /**
      * For storing extra data with record
@@ -94,6 +88,7 @@ abstract class AbstractBase implements \VuFind\Db\Table\DbTableAwareInterface,
     {
         $this->mainConfig = $mainConfig;
         $this->recordConfig = $recordConfig ?? $mainConfig;
+        $this->setSourceIdentifiers('Solr');
     }
 
     /**
@@ -275,28 +270,6 @@ abstract class AbstractBase implements \VuFind\Db\Table\DbTableAwareInterface,
             $this->getSourceIdentifier(),
             $user_id
         );
-    }
-
-    /**
-     * Set the source backend identifier.
-     *
-     * @param string $identifier Backend identifier
-     *
-     * @return void
-     */
-    public function setSourceIdentifier($identifier)
-    {
-        $this->sourceIdentifier = $identifier;
-    }
-
-    /**
-     * Return the source backend identifier.
-     *
-     * @return string
-     */
-    public function getSourceIdentifier()
-    {
-        return $this->sourceIdentifier;
     }
 
     /**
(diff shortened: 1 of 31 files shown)


EreMaijala:
Spun off from #2341.

module/VuFind/src/VuFind/RecordDriver/AbstractBase.php
```diff
@@ -94,6 +88,7 @@ public function __construct($mainConfig = null, $recordConfig = null)
     {
         $this->mainConfig = $mainConfig;
         $this->recordConfig = $recordConfig ?? $mainConfig;
+        $this->setSourceIdentifiers('Solr');
```
demiankatz: Would it be better to use the `DEFAULT_SEARCH_BACKEND` constant here instead?

themes/bootstrap3/templates/Recommend/SideFacets/cluster-list.phtml
```diff
@@ -20,6 +20,13 @@
   <?php foreach ($this->cluster['list'] as $i => $thisFacet): ?>
     <?php if ($i == $this->facets_before_more): ?>
       <?php
+        $queryParams = [
+            'facet' => $this->title,
+            'facetop' => $thisFacet['operator'],
+            'facetexclude' => $this->allowExclude ? 1 : 0,
+            'searchClassId' => $this->searchClassId
+        ];
+        $facetLightboxParams = http_build_query($queryParams, '', '&amp;');
```
demiankatz: This seems like a partially implemented change. Do we need to merge #2362 and update this before it is ready to merge?

demiankatz (changes requested):
Thanks, @EreMaijala -- just a couple of small comments. Also, I have started drafting changelog notes for when this is merged. How does this sound to you?

> To increase flexibility, record drivers can now differentiate between a "source identifier" (the backend where the record resides) and a "search backend identifier" (the backend used to find the record). This distinction makes it possible to have "virtual" search backends which pull data from multiple sources. Supporting this required several changes which may need to be reflected in local custom code:
>   * The setSourceIdentifier record driver and record collection methods have been deprecated in favor of setSourceIdentifiers, which allows both IDs to be set. The legacy method will set both IDs to the same value for backward compatibility. These changes affect any classes implementing VuFindSearch\Response\RecordCollectionInterface or VuFindSearch\Response\RecordInterface. The changes have been implemented in VuFindSearch\response\AbstractRecordCollection and VuFindSearch\Response\RecordTrait for easy reuse.
>   * Record controllers (i.e. VuFind\Controller\AbstractRecord and its children) have replaced the $searchClassId property with $sourceId.

themes/bootstrap3/templates/Recommend/SideFacets/cluster-list.phtml
demiankatz: ...and #2362 has now been merged, so you should be able to merge dev-9.0 back into this and resolve the issue.

EreMaijala: This was accidentally included here, not related to the source id change.

module/VuFind/src/VuFind/RecordDriver/AbstractBase.php
EreMaijala: Yes. Done. Another possibility would be make the default empty or something so that it's not accidentally left as is.

demiankatz: If we default the properties to empty strings in the trait, can we drop this set call entirely? It does seem like maybe it would be better to fail due to an empty string than to accidentally refer to the wrong backend if something is overlooked.

EreMaijala: Yes, and now done. Mink tests are still passing. :)

EreMaijala:
@demiankatz Thanks, the release notes are great!

demiankatz (approved):

PR #2362, opened 2022-03-23
Use http_build_query for facet list parameters.

EreMaijala:
Makes sure the parameters are properly encoded and collected just once.


demiankatz (approved):
Looks good to me!

PR #2363, opened 2022-03-23
Make it possible to set pcAvailability to a falsy value. 

EreMaijala:
This makes it essentially a no-op, but is needed for default filter value override in upcoming blended search. Extracted from #2341, also adds unit test for different values.

### DIFF
--- a/module/VuFindSearch/src/VuFindSearch/Backend/Primo/Backend.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/Primo/Backend.php
@@ -246,10 +246,12 @@ class Backend extends AbstractBackend
             $options[$key] = in_array($key, $arraySettings) ? $param : $param[0];
         }
 
-        // Use special facet pcAvailabilty if it has been set
-        if (isset($params['filterList']['pcAvailability'])) {
+        // Use special pcAvailability filter if it has been set:
+        if ($values = $params['filterList']['pcAvailability']['values'] ?? []) {
+            $value = $values[0];
+            $options['pcAvailability']
+                = !in_array($value, [false, 0, '0', 'false'], true);
             unset($options['filterList']['pcAvailability']);
-            $options['pcAvailability'] = true;
         }
 
         return $options;
--- a/module/VuFindSearch/src/VuFindSearch/Backend/Primo/Backend.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/Primo/Backend.php
@@ -249,6 +249,7 @@ class Backend extends AbstractBackend
         // Use special pcAvailability filter if it has been set:
         if ($values = $params['filterList']['pcAvailability']['values'] ?? []) {
             $value = $values[0];
+            // Note that '' is treated as true for the simple case with no value
             $options['pcAvailability']
                 = !in_array($value, [false, 0, '0', 'false'], true);
             unset($options['filterList']['pcAvailability']);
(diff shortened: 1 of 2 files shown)


module/VuFindSearch/src/VuFindSearch/Backend/Primo/Backend.php
```diff
@@ -246,10 +246,13 @@ protected function paramBagToPrimoQuery(ParamBag $params)
             $options[$key] = in_array($key, $arraySettings) ? $param : $param[0];
         }
 
-        // Use special facet pcAvailabilty if it has been set
-        if (isset($params['filterList']['pcAvailability'])) {
+        // Use special pcAvailability filter if it has been set:
+        if ($values = $params['filterList']['pcAvailability']['values'] ?? []) {
+            $value = $values[0];
```
demiankatz: If `$params['filterList']['pcAvailability']['values']` is unset and you default to `[]`, then I think that `$values[0]` will be unset and cause a notice. I guess the unset case is not covered by the tests below!

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Primo/BackendTest.php
```diff
@@ -182,6 +182,83 @@ public function testMergedParamBag()
...
+                'filterList' => [
+                    'pcAvailability' => [
+                        'values' => [
+                            $value
+                        ]
+                    ]
+                ]
+            ]
+        );
+        $expectedParams = ['limit' => 10, 'pageNumber' => 1, 'filterList' => [], 'pcAvailability' => $expected, 'query' => [['index' => null, 'lookfor' => 'foo']]];
```
demiankatz: While we don't enforce long line checking in tests, It might be worth wrapping this and line 256 below for readability.

demiankatz (changes requested):
Thanks, @EreMaijala -- I think I may have found a small problem here.

module/VuFindSearch/src/VuFindSearch/Backend/Primo/Backend.php
EreMaijala: It won't go into the `if` branch since `[]` is falsy, right? It's still possible that there are values but keyed somehow weird, but I'm not sure if that's something we need to guard against.

EreMaijala: Anyway, I replaced `$values[0]` with `reset($values)` for extra safety.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Primo/BackendTest.php
EreMaijala: True. I went throught the whole file and wrapped long lines while at it.

module/VuFindSearch/src/VuFindSearch/Backend/Primo/Backend.php
demiankatz: You are, of course, correct -- I wasn't thinking it all the way through. In any case, thanks for the adjustment!

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Primo/BackendTest.php
demiankatz: Thanks!

demiankatz (approved):

PR #2365, opened 2022-03-25
Always show inverted Solr checkbox filters, even if no results.

demiankatz:
This is a follow-up to #2340, which ensures that inverted checkbox filters are always visible, even if there are no search results, since they may be needed to expand the result set.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Solr/Params.php
+++ b/module/VuFind/src/VuFind/Search/Solr/Params.php
@@ -668,4 +668,36 @@ class Params extends \VuFind\Search\Base\Params
 
         return $filter;
     }
+
+    /**
+     * Get information on the current state of the boolean checkbox facets.
+     *
+     * @param array $include List of checkbox filters to return (null for all)
+     *
+     * @return array
+     */
+    public function getCheckboxFacets(array $include = null)
+    {
+        // Grab checkbox facet details using the standard method:
+        $facets = parent::getCheckboxFacets($include);
+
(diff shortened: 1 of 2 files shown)


EreMaijala (changes requested):
This works but I'd like to refine the looks a bit. The filter doesn't have a heading or anything, so it looks out of place, particularly with narrow displays. Something like "Expand Search" as a heading would help, I think.

EreMaijala:
@demiankatz On a related note, I only now realized that inverted filters make the "Narrow Search" heading in results a bit misleading. Perhaps we should consider something like "Refine Search" for both cases?

demiankatz:
@EreMaijala, I agree with your suggestions, and indeed, Villanova changed "Narrow Search" to "Refine Search" in its instance many years ago as a local customization. I think the only reason I never changed it in the core is the inconvenience of changing such an important translation string in all the language files... but maybe we could add "Refine Search" as a new string and set up the templates so that they'll use "Narrow Search" as a fallback if the "Refine Search" translation is not yet defined.

However, I'd suggest that these modifications might be better handled in a separate PR -- this one is wiring up Solr to an existing mechanism (that's been in use for EDS for several years). I think improving that mechanism is a separate task that can be addressed independently of what I've done here.

EreMaijala (approved):
Right, I didn't realize that the lone checkbox facets are already displayed with EDS. Approved.

demiankatz:
I've opened [VUFIND-1539](https://vufind.org/jira/browse/VUFIND-1539) to track the UI improvements discussed above.

PR #2366, opened 2022-03-28
Fix filtering of dynamic checkbox facets.

demiankatz:
I just discovered that #2358 has the unfortunate side effect of breaking checkbox facets in EDS, because the EDS checkboxes are dynamically-generated, rather than tied directly to SideFacets configuration.

This PR attempts to sort that out, by adding a new flag to identify dynamic vs. user-configured checkboxes, and a new setting for filtering these out (or not) from SideFacets. Some aspects of this could be viewed as backward-breaking (we changed the way the first parameter to getCheckboxFacets works -- by making it actually match the documented behavior more strictly -- and we added an optional second parameter to that function), but I think they are justified in being necessary to effect a fix.

I think there is room to have more test coverage (at very least, I'd like to get coverage on the EDS and Summon classes impacted by the change), but I'm going to wait for feedback on the general approach before I invest more time in test coverage.

TODO
- [x] Add more tests

### DIFF
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -301,13 +301,18 @@ WorkKeys = year
 ;       displaying in a small result set. (default = 0)
 ;       [facet-n] A facet to apply to the random selection
 ;       [facetvalue-n] The facet value to apply to the random selection
-; SideFacets:[regular facet section]:[checkbox facet section]:[ini name]
+; SideFacets:[regular facet section]:[checkbox facet section]:[ini name
+;       :[include dynamic checkboxes]
 ;       Display the specified facets, where [ini name] is the name of an ini file
 ;       in your config directory (defaults to "facets" if not supplied),
 ;       [regular facet section] is the name of a section of the ini file containing
 ;       standard facet settings (defaults to "Results" if not specified),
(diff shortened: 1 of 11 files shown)


EreMaijala:
@demiankatz Is there a benefit with making the configuration manual, or could we just deduplicate the filters automatically?

demiankatz:
> @demiankatz Is there a benefit with making the configuration manual, or could we just deduplicate the filters automatically?

The primary benefit of making it manual is that it allows the end user to control where the checkboxes appear, since one of the use cases of creating multiple instances of SideFacets is to move the checkboxes lower in the sidebar. We need a manual setting to allow that. I suppose we could complement the manual setting with an automatic feature to avoid duplication, though -- so the dynamic checkboxes only display a maximum of one time. I'm not sure if it's worth it, though -- we would have to do something tricky with statics to make it work, and that might cause as many problems as it solves.

EreMaijala:
@demiankatz Ok, thanks. In that case I think this approach is sound.

EreMaijala (approved):

demiankatz:
Thanks, @EreMaijala -- I have now added test coverage for everything, so I am merging this.

PR #2367, opened 2022-03-29
Include all primary authors in export formats.

EreMaijala:
Most export formats included all secondary authors but only the first primary author. This adds support for all primary authors and simplifies the code for secondary authors a bit while at it.

### DIFF
--- a/themes/root/templates/RecordDriver/AbstractBase/export-bibtex.phtml
+++ b/themes/root/templates/RecordDriver/AbstractBase/export-bibtex.phtml
@@ -60,12 +60,24 @@ if (is_array($series)) {
     }
 }
 
-foreach ($this->driver->tryMethod('getPrimaryAuthors', [], []) as $current) {
-    echo "author = {{$current}},\n";
+$mergeAuthors = function (array $a): string {
+    return implode(
+        ' and ',
+        array_map(
+            function ($s) {
+                return str_replace(' and ', ' {and} ', $s);
+            },
+            $a
+        )
+    );
+};
+
+if ($authors = $mergeAuthors($this->driver->tryMethod('getPrimaryAuthors', [], []))) {
+    echo "author = {{$authors}},\n";
 }
 
-foreach ($this->driver->tryMethod('getSecondaryAuthors', [], []) as $current) {
-    echo "editor = {{$current}},\n";
+if ($authors = $mergeAuthors($this->driver->tryMethod('getSecondaryAuthors', [], []))) {
+    echo "editor = {{$authors}},\n";
 }
 
 $pubPlaces = $this->driver->tryMethod('getPlacesOfPublication');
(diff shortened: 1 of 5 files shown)


EreMaijala:
The BibTeX handling was apparently a bit broken for repeated editors as well, should be fixed now. I used the following resources for reference:
https://nwalsh.com/tex/texhelp/bibtx-23.html
https://verbosus.com/bibtex-style-examples.html
https://bibtex.online/

themes/root/templates/RecordDriver/AbstractBase/export-refworks.phtml
```diff
@@ -40,19 +40,15 @@ if (!empty($journalTitle)) {
...
-if (is_array($secondaryAuthors)) {
-    foreach ($secondaryAuthors as $current) {
-        echo "A2 $current\n";
-    }
+foreach ($this->driver->tryMethod('getSecondaryAuthors', [], []) as $current) {
+    echo "A2 $current\n";
 }
 
-$languages = $this->driver->tryMethod('getLanguages');
+$languages = $this->driver->tryMethod('getLanguages', [], []);
```
demiankatz: Looks like you started to simplify this but didn't remove the is_array check below...

themes/root/templates/RecordDriver/AbstractBase/export-bibtex.phtml
```diff
@@ -60,16 +60,24 @@ if (is_array($series)) {
...
+if ($authors = $mergeAuthors($this->driver->tryMethod('getPrimaryAuthors', [], []))) {
+    echo "author = {{$authors}},\n";
 }
 
-$secondaryAuthors = $this->driver->tryMethod('getSecondaryAuthors');
-if (is_array($secondaryAuthors)) {
-    foreach ($secondaryAuthors as $current) {
-        echo "editor = {{$current}},\n";
-    }
+if ($authors = $mergeAuthors($this->driver->tryMethod('getSecondaryAuthors', [], []))) {
```
demiankatz: Would it be better to call this variable `$editors` to avoid reusing the `$authors` name? Not a big difference, but just feels slightly more clear to me.

demiankatz (changes requested):
Thanks, @EreMaijala! A couple of small comments. I also wonder whether it's worth putting in some effort to add tests for these (not necessarily right now, but at some point... this is the kind of code that could benefit from test coverage).

themes/root/templates/RecordDriver/AbstractBase/export-refworks.phtml
EreMaijala: Oops. I started it but meant to revert to avoid scope creep.

themes/root/templates/RecordDriver/AbstractBase/export-bibtex.phtml
EreMaijala: Changed. Not sure I like it, though, since secondary authors can be more than editors. It's just the only field we have for them.

demiankatz: Agreed, it's not ideal, but it seems to be what we're stuck with.

demiankatz (approved):

PR #2368, opened 2022-03-30
Fix facet lightbox params in cluster-list.phtml.

EreMaijala:
I made a mistake in collecting the facet lightbox params too early when `$thisFacet` is not yet initialized.

### DIFF
--- a/themes/bootstrap3/templates/Recommend/SideFacets/cluster-list.phtml
+++ b/themes/bootstrap3/templates/Recommend/SideFacets/cluster-list.phtml
@@ -15,20 +15,18 @@
     'wrapper-class' => false,
     'wrapper-tagname' => 'div',
   ];
+  $facetLightboxParams = http_build_query(
+      [
+          'facet' => $this->title,
+          'facetop' => $this->cluster['list'][0]['operator'] ?? 'AND',
+          'facetexclude' => $this->allowExclude ? 1 : 0,
+      ],
+      '',
+      '&amp;'
+  );
 ?>
 <div<?php if ($truncateSettings): ?> class="truncate-facets" data-truncate="<?=$this->escapeHtml(json_encode($truncateSettings))?>"<?php endif; ?>>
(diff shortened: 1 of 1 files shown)


EreMaijala:
Now I'm not sure why the facetop parameter exists, since we should be able to get that piece of information from configuration, but I wasn't able to dig deeper into it at this point.

demiankatz:
@EreMaijala, thanks for the fix. Some questions and comments:

1.) I may be mistaken, but I think the reason for the facetop parameter is so that we could theoretically open a facet lightbox in a non-default mode -- e.g. if a facet is configured for AND by default but we want to do an OR operation, we could potentially create a mechanism to support that. No such mechanism currently exists, but it seems like something that might be useful for future growth.

2.) Should we create a Mink test that opens the facet lightbox and tests OR facets, since presumably if we had had such a thing in the first place, we would have caught this bug immediately? (I can help with it if you don't have time).

3.) Is every entry in the list guaranteed to have the same operator? I wonder if we should define $facetLightboxParams inside the loop to be more consistent with past behavior. (I admit I didn't review the context of this very closely, so maybe this is an obvious optimization... but some of the scoping seems a little questionable (for another example: why do we define $moreUrl on every iteration of the foreach starting in line 29 when we only use it inside an if condition?).

EreMaijala:
@demiankatz 

1.) Makes sense.

2.) Yes, and help appreciated. I'm a bit shocked it didn't cause a Whoops exception either, or I failed completely to test it.

3.) I kept it outside the loop since it's also used _after_ the loop, and defining things in a loop to use after is not one my favorite practices. And it doesn't really make a difference if there are entries with different operators when the link isn't specific to one of them.

demiankatz (approved):

demiankatz:
Thanks, @EreMaijala, you've persuaded me that the fix is correct. I see a couple of optimizations that can be made, but in the interest of getting the bug fixed, I'll merge this as-is and work on optimizing and testing later when time permits.

PR #2369, opened 2022-03-30
Refactor cluster list to fix bug and streamline code.

demiankatz:
This is a follow-up to #2368, intended to streamline code, fix another bug, and add test coverage.

TODO
- [x] Add Mink test for OR facets
- [x] Double-check correctness of baseUriExtra handling


demiankatz:
@EreMaijala, this is some follow-up work from last month that I didn't get around to until today. Note that the added test and template fix/refactoring are not directly related to one another -- just two facet-related improvements that we talked about doing around the same time.

EreMaijala (approved):
Looks good to me!

PR #2372, opened 2022-03-31
Standardize getFacets methods of different RecordCollection classes.

EreMaijala:
This makes all record collections' getFacets method return the facet array as defined in RecordCollectionInterface. It also consolidates different Results::getFacetList methods into a common helper method in the base Results class. And fixes a problem that caused Primo searches to display "reference_entrys" in the active filter list.

A couple of interface files that were never used are also removed.

Inspired by #2341 and would eliminate the need to convert facet formats of different backends.

Mink tests are passing.

 TODO:
 - [x] Finish #2394 first.
 - [x] Release notes
 - [x] Resolve https://vufind.org/jira/browse/VUFIND-1540 when merged
 - [x] After merging, create follow-up PR (or JIRA ticket) to wire up HideFacetValueListener to other search backends


demiankatz:
@EreMaijala, there's a lot to absorb here, and I wonder if the best approach to a review is through improving test coverage. Would it make sense to put more effort into unit testing the various Search\xxx\Results classes to make sure all facet-related edge cases are covered, and then make sure the tests still pass here? Or do you think that the refactoring here is so dramatic that it would require tests to be revised? The work here also seems very likely to conflict with #2352, and I wonder if we should get that one merged before we finish dealing with this one... but curious to hear your thoughts on the best path forward.

I'm certainly happy to help with test development, time permitting -- or I can just keep reading this code. I just feel like it's going to take a lot of time to mentally process this, and maybe some of that time would be better invested in writing tests (which would then refresh my memory of the code and probably make the read of this a little quicker).

EreMaijala:
@demiankatz I'd be fine with postponing this until #2352 is merged. In fact, I'd prefer getting #2341 in before this as well, but since this would simplify #2341 a bit, either way is justifiable.

EreMaijala:
@demiankatz The diffs look worse than it is, really, because I unified the different `getFacetList` methods to a common `buildFacetList` method in the Base\Params class. `getFacetList` methods are now mostly just wrappers around `buildFacetList`. Other than that it's mostly:

1.) Undecoration of Solr facets from array objects.
2.) Splitting the Solr facet retrieval methods by facet type (query facets and pivot facets are now separate).
3.) Tweaking HideFacetValueListener to work without the assumption that it handles references to the original list.
4.) Making EDS return the same kind of array as the others.

Making sure everything is covered by the tests makes sense. I'll work on that a bit.

demiankatz:
Thanks, @EreMaijala! I think #2352 is nearly complete (and might benefit from your input, if you have any strong feelings about what we're doing there). I'll continue chipping away at the review of this and #2341 as time permits. If you're able to add tests here, that might help me get through this faster, but in any case, we can decide on the merge order based on whichever one gets fully reviewed/revised/tested first!

EreMaijala:
Last commits and #2387 add some tests to cover Solr aspect of this. Backend side of EDS and Primo are already covered by the tests here. The Results classes of EDS and Primo are not covered, but now when they're mostly just wrappers for the shared functionality, #2387 will already cover most of them. I'll do some coverage analysis when #2387 is merged here and add any missing pieces if this theory doesn't hold. :)

demiankatz:
Thanks, @EreMaijala, that's great! I've left some comments on #2387; please remind me to review this again once that part is done.

EreMaijala:
@demiankatz That part is done. :)

demiankatz:
@EreMaijala, now that #2352 is merged, there is a conflict here that needs to be resolved. Please let me know if you need any help sorting it out (and sorry for the inconvenience!).

EreMaijala:
@demiankatz I think I managed to merge the changes and incorporate them into the new method in the base class.

module/VuFind/src/VuFind/Search/Base/Params.php
```diff
@@ -1064,23 +1064,24 @@ public function getFilterList($excludeCheckboxFilters = false)
...
      *
-     * @param string $field       The facet
-     * @param string $displayText The facet value
+     * @param string $field Facet field
+     * @param string $value Facet value
      *
      * @return string
      */
-    public function checkForDelimitedFacetDisplayText($field, $displayText)
+    public function getFacetValueDisplayText(string $field, string $value): string
```
demiankatz: I've extracted this change to #2394 since I think it's a separate issue from the rest of the work here, and we should try to resolve in in the dev branch to avoid introducing an extra backward compatibility break between newly-added 8.1 functionality and subsequent 9.0 code. We should resolve that PR before merging this one, so I'll add a TODO checkbox at the top here.

module/VuFind/src/VuFind/Search/Base/Results.php
```diff
@@ -30,6 +30,7 @@
 use Laminas\Paginator\Paginator;
 use VuFind\Record\Loader;
 use VuFind\Search\Factory\UrlQueryHelperFactory;
+use VuFind\Search\Solr\HierarchicalFacetHelper;
```
demiankatz: I'm not sure I like having the base class dependent on a Solr-specific helper. Should we define an interface to make this more generic/flexible?

module/VuFind/src/VuFind/Search/Base/Results.php
```diff
@@ -735,4 +755,99 @@ public function getFullFieldFacets(
...
+        if (null === $filter) {
+            $filter = $this->getParams()->getFacetConfig();
+        }
+
+        // Start building the facet list:
+        $result = [];
+
+        // Loop through every field returned by the result set
+        $translatedFacets = $this->getOptions()->getTranslatedFacets();
+        $hierarchicalFacets
```
demiankatz, line 781: If we're moving hierarchical facet processing into the base class, should we just move getHierarchicalFacets up into the base class as well so we can assume it's defined?

demiankatz: ...and of course this makes me wonder if we should have a separate PR for hierarchical facet adjustments as a prerequisite for this one, to keep all the changes separate.

module/VuFind/src/VuFind/Search/Primo/Params.php
```diff
@@ -113,9 +107,15 @@ protected function formatFilterListEntry($field, $value, $operator, $translate)
      */
     public function fixPrimoFacetValue($str)
     {
-        // Special case: odd spelling error in Primo results:
-        if ($str == 'reference_entrys') {
-            return 'Reference Entries';
+        // Special case: odd spelling errors and other special cases in Primo
+        // results (some at least in CDI):
+        $replacements = [
```
demiankatz: Should this enhancement be merged separately? Also, do we need better Primo test coverage? Might it make sense to introduce those things together in a separate PR?

module/VuFind/src/VuFind/Search/Solr/HideFacetValueListener.php
```diff
@@ -131,20 +132,29 @@ public function onSearchPost(EventInterface $event)
     protected function processHideFacetValue($event)
     {
         $result = $event->getParam('command')->getResult();
-        $facets = $result->getFacets()->getFieldFacets();
+        $facets = $result->getFacets();
```
demiankatz, line 135: Is this listener no longer Solr-specific thanks to this standardization? Should we consider renaming it and wiring it up to other search backends? (Happy to do that as a follow-up PR if necessary).

demiankatz (commented):
This is looking pretty good, but I think there may be a few bits worth separating and resolving prior to this PR.

demiankatz:
One other question: do we need to worry about standardizing the Summon code as well? I notice that it is not covered here. Unfortunately, I no longer have a means of testing it. Do we need to see if we can find a current Summon/VuFind user to help us with this?

module/VuFind/src/VuFind/Search/Base/Results.php
EreMaijala: Yes, good idea. Now done.

module/VuFind/src/VuFind/Search/Solr/HideFacetValueListener.php
EreMaijala: I think so, yes. That would make sense.

EreMaijala:
@demiankatz I wouldn't spend too much with Summon. It should still work the way it does, it's just not compatible with Blender.

module/VuFind/src/VuFind/Search/Primo/Params.php
EreMaijala: Now in #2396. Test coverage added for these cases, not much more.

module/VuFind/src/VuFind/Search/Base/Results.php
EreMaijala: Now in #2397.

module/VuFind/src/VuFind/Search/Solr/HideFacetValueListener.php
demiankatz: I've put a TODO checkbox at the top to remind myself to do this as a follow-up after we merge this PR.

demiankatz (approved):

PR #2373, opened 2022-03-31
Improve test coverage for VuFindSearch\Command namespace.

EreMaijala:
These tests cover the newly added setter methods and more, and bring unit test coverage to 100% for the VuFindSearch\Command namespace.


demiankatz (approved):

demiankatz:
Thanks, @EreMaijala -- not entirely sure why we didn't have a SearchCommandTest already, but it must have gotten overlooked because the SearchCommand was fully covered due to its use in other tests.

PR #2374, opened 2022-03-31
Add a post-processor to support <Consume-Whitespace /> tag.

EreMaijala:
The tag is used to remove all whitespace around it (and the tag itself).

So the template code can be something like this:

```
<a href="#">
  <i class="something">
  <Consume-Whitespace />
  Label Text
</a>

```
And the resulting HTML is:

<a href="#">
  <i class="something">Label Text
</a>

This could help with the underline issue in #1962.

### DIFF
--- a/module/VuFind/src/VuFind/Mvc/PhpPostProcess.php
+++ b/module/VuFind/src/VuFind/Mvc/PhpPostProcess.php
@@ -98,6 +98,30 @@ class PhpPostProcess extends AbstractListenerAggregate
      */
     protected function postProcessHtml(string $content): string
     {
-        return preg_replace('/\s+<Consume-Whitespace\s*\/>\s+/', '', $content);
+        // Avoid doing any further processing if there isn't a <Consume-Whitespace>
+        // tag in any form in the response:
+        if (strpos($content, '<Consume-Whitespace') === false) {
+            return $content;
+        }
+        // Remove spaces around a self-closing Consume-Whitespace tag:
+        $content = preg_replace('/\s+<Consume-Whitespace\s*\/>\s+/', '', $content);
+        // Remove spaces between html elements inside a Consume-Whitespace tag:
+        $content = preg_replace_callback(
+            '/(<Consume-Whitespace\s*>.*?<\/Consume-Whitespace>)/s',
+            function (array $matches): string {
+                // First handle the spaces between tags, including the
+                // Consume-Whitespace tag:
+                $s = preg_replace('/>\s+</', '><', $matches[1]);
+                // Remove Consume-Whitespace tag:
+                $s = preg_replace(
+                    '/<Consume-Whitespace\s*>(.*)<\/Consume-Whitespace>/',
+                    "$1",
+                    $s
+                );
+                return $s;
+            },
+            $content
+        );
+        return $content;
     }
 }
--- a/module/VuFind/src/VuFind/Mvc/PhpPostProcess.php
+++ b/module/VuFind/src/VuFind/Mvc/PhpPostProcess.php
@@ -1,0 +1,102 @@
+<?php
+/**
+ * PHP-rendered page post-processor.
+ *
+ * PHP version 7
+ *
+ * Copyright (C) The National Library of Finland 2022.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Service
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+namespace VuFind\Mvc;
+
+use Laminas\EventManager\AbstractListenerAggregate;
+use Laminas\EventManager\EventManagerInterface;
+use Laminas\Http\PhpEnvironment\Response;
+use Laminas\Mvc\MvcEvent;
+
+/**
+ * PHP-rendered page post-processor.
+ *
+ * @category VuFind
+ * @package  Service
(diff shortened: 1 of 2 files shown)


EreMaijala:
The docs implied that you could hook into the SendResponse event, but it seems that class that triggers it (using MvcEvent::EVENT_FINISH) uses its own event manager. One option would be to extend SendResponseListener, but it's just another way to do almost the same thing.

module/VuFind/src/VuFind/Mvc/PhpPostProcess.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public function handlePostProcess(MvcEvent $e)
+    {
+        $response = $e->getResponse();
+        if (!($response instanceof Response)) {
+            return;
+        }
+        $contentTypeHeader = $response->getHeaders()->get('Content-Type');
+        $contentType = $contentTypeHeader ? $contentTypeHeader->getFieldValue() : '';
+        if (!$contentType || 'text/html' === $contentType) {
```
demiankatz: What does it mean if there is no Content-Type header? Is it safe to assume that implies HTML?

demiankatz (commented):
Thanks, @EreMaijala, this looks good to me -- I just left one question (which is based entirely on my ignorance of framework behavior). Whether or not we merge this will probably depend on the outcome of discussion on #1962, though this may be valuable as a general-purpose tool regardless.

crhallberg:
I like this approach, but I'm wary of adding a custom element in so many places in the code.

1. Could we add the `<Consume-Whitespace />` element/marker into the icon templates? That way every icon automatically adds this tag, minimizing wider template changes.

1. Could we change the regex to target the icons on their own without adding a new element/markup?

demiankatz:
@crhallberg, my vision was definitely to include this marker in the icon templates, rather than everywhere that icons are used. I think that makes it reasonably self-contained. I like using a generic marker rather than targeting the icons directly, because there may be other situations where consuming whitespace will solve a problem, and then this gives us a tool to solve those issues.

module/VuFind/src/VuFind/Mvc/PhpPostProcess.php
EreMaijala: Yes. In that case we haven't overridden the default, which we assume to be text/html (could be considered a bold assumption, but that's another discussion). You can see the effect by adding `ini_set('default_mimetype', 'text/plain');` at the beginning of index.php. Without an explicitly set Content-Type header, the default is added by PHP when the response is echo'd in HttpResponseSender.

EreMaijala:
@crhallberg It would also be possible to add something like an attribute-based whitespace filtering: `<i class="icon ... consume-whitespace-after"></i>` or `<i class="icon ..." data-consume-whitespace-after></i>`. This would look nicer in the code, but might be just a bit confusing to the uninitiated.

aleksip:
If we decide to go this route support for opening and closing tags would be even more useful. Similar to [how Twig does this](https://twig.symfony.com/doc/3.x/filters/spaceless.html) with the apply tag, for example.

The cautions and warnings on that page are not very reassuring, though! 😅

module/VuFind/src/VuFind/Mvc/PhpPostProcess.php
demiankatz: If that's the case, could we make this safer by using an `ini_get` to retrieve the default if no override is set?

EreMaijala: Sure, but I don't expect many things in VuFind to work properly if the default is something else than `text/html`...

demiankatz:
It does sound to me like part of the decision to be made here is between flexibility/complexity and performance/simplicity. I kind of like the approach of just removing one very specific tag, because the RegEx is simple and it's very clear what is going on. Something that requires a more nuanced understanding of the content is going to require more processing and can go wrong in more ways. Right now, I can't think of a strong need for all the flexibility that Twig offers, but maybe I'm being short-sighted there. At the moment, I'm kind of looking for a workable last resort solution that we can use until we come up with something better, rather than trying to build a robust framework feature. I still wish we could find a solution that didn't require any of this!

module/VuFind/src/VuFind/Mvc/PhpPostProcess.php
demiankatz: That's fair... but it might make the intention of the code more explicit as well.

EreMaijala: Done.

EreMaijala:
I think adding support for opening and closing tags would add flexibility without making this much more complex. I'll see if I can do it without creating a mess...

EreMaijala:
Done, in a somewhat naïve way. The limitation is that you can't put an opening+closing tag inside another. 

Since the opening-closing version doesn't trim whitespace in text nodes, both can be combined as needed (or the text done enclosed in a `<span>`):

```
            <Consume-Whitespace>
            <li>
              <a id="feedbackLink" data-lightbox href="<?=$this->url('feedback-home') ?>">
                <?=$this->icon('feedback') ?>
                <Consume-Whitespace/><?=$this->transEsc("Feedback")?>
              </a>
            </li>
            </Consume-Whitespace>

```

I haven't added any tests yet since I considered this more of a proof of concept, but I think that'd be the natural next step.

demiankatz:
@EreMaijala, I wouldn't invest too much time in writing tests or further developing this until we see how the conversation on #1962 plays out -- I remain torn between the fact that I think this could be useful and the fact that it's adding extra processing to every page in order to create a potentially unintuitive mechanism.

If we do decide to go forward with this, I also wonder if the new opening-closing version you've created should actually have a different name, because it feels to me like a somewhat different feature than the self-closing version. Maybe calling it `<Consume-Whitespace-Between-Tags>` or `<Consume-Contained-Whitespace>` would be more clear.

But don't get too caught up in that potential debate just yet -- I'm still having trouble getting past the "should we do this?" question to be deeply committed to the "exactly how should we do this?" discussion. :-)

EreMaijala:
@demiankatz No problem. :) While I did this as a solution to a problem, creating this was more of "can I make this work?" rather than "we should do this". Thus I have no problem abandoning this.

demiankatz:
Yes, I understand completely -- it reminds me of the fun I had with [this bit of madness](https://blog.library.villanova.edu/2011/10/05/separating-local-code-customizations-in-php/) ages ago -- and I think it's good work, it's just a question of whether it's the right decision. We should definitely leave this open while the conversations continue...

demiankatz:
@EreMaijala, I think the work on #2400 has rendered this unnecessary for the moment. I think it's an interesting solution that people might find useful, but there are also good reasons not to add it to the core code (as discussed above). How would you feel about closing this PR but creating a `feature/consume-whitespace` branch in the main VuFind repo so we can find this again in future if we need it?

EreMaijala:
@demiankatz Makes sense to me.

demiankatz:
Thanks, @EreMaijala! The feature branch has been created and documented on the [git branches wiki page](https://vufind.org/wiki/development:git-branches).

PR #2375, opened 2022-04-01
Add ability to use custom commonmark extensions

xmorave2:


module/VuFind/src/VuFind/Service/MarkdownFactory.php
```diff
@@ -135,7 +143,12 @@ protected function getEnvironment(): EnvironmentBuilderInterface
         $environment->addExtension(new CommonMarkCoreExtension());
         foreach ($this->extensions as $extension) {
             $extensionClass = $this->getExtensionClass($extension);
-            $environment->addExtension(new $extensionClass());
+            // For case, somebody needs to create extension using custom factory, we
+            // try to get the object from DI container if possible
+            $extensionObject = $this->container->has($extensionClass)
+                ? $this->container->get($extensionClass)
```
demiankatz, line 149: Does it make sense to put extensions in the top-level container, or should we create a specialized container for them? (I'm not necessarily saying we should change anything -- just wondering if you've considered both possibilities).

module/VuFind/src/VuFind/Service/MarkdownFactory.php
```diff
@@ -172,11 +185,13 @@ protected function getBaseConfig(): array
      */
     protected function getExtensionClass(string $extension): string
     {
-        $extensionClass = sprintf(
-            'League\CommonMark\Extension\%s\%sExtension',
-            $extension,
-            $extension
-        );
+        $extensionClass = (strpos($extension, '\\') !== false)
```
demiankatz, line 188: Would it make sense to use a `class_exists` here, or do you want to support service names that don't necessarily match existing class names?

module/VuFind/tests/unit-tests/src/VuFindTest/Service/MarkdownFactoryTest.php
```diff
@@ -281,8 +300,24 @@ protected function getMarkdownEnvironmentExtensions(array $config): array
...
+            'League\CommonMark\Extension\Table\TableExtension',
+            'League\CommonMark\Extension\TaskList\TaskListExtension',
+        ];
         $config = new Config($config);
         $container = new \VuFindTest\Container\MockContainer($this);
+        foreach ($disabledServices as $service) {
+            $container->disable($service);
+        }
+        $container->set(
+            'VuFindTest\Markdown\ExampleExtension',
```
demiankatz: Might be better to use `::class` syntax here rather than a string.

module/VuFind/tests/unit-tests/src/VuFindTest/Service/MarkdownFactoryTest.php
```diff
@@ -281,8 +300,24 @@ protected function getMarkdownEnvironmentExtensions(array $config): array
      */
     protected function getMarkdownConverter(array $config): ConverterInterface
     {
+        $disabledServices = [
+            'League\CommonMark\Extension\Autolink\AutolinkExtension',
```
demiankatz: Can we use `::class` syntax for these to improve static analysis?

demiankatz (changes requested):
Thanks for this! A few comments -- but please consider most of these to be suggestions rather than requests. :-)

module/VuFind/src/VuFind/Service/MarkdownFactory.php
xmorave2: The `class_exists` is used few lines later for checking both - custom and commonmark internal extensions.

I did not want to support non-class service names - but it could be done in future if needed.

module/VuFind/src/VuFind/Service/MarkdownFactory.php
xmorave2: I don't expect too many custom extension to be added in typical instalation (if even any) so I think it would be just unnecessary overhead.

module/VuFind/tests/unit-tests/src/VuFindTest/Service/MarkdownFactoryTest.php
xmorave2: Yes, I will change it

module/VuFind/src/VuFind/Service/MarkdownFactory.php
demiankatz: Makes sense!

module/VuFind/src/VuFind/Service/MarkdownFactory.php
demiankatz: Okay, makes sense! I agree that there is probably not a strong reason to support non-class service names.

demiankatz (approved):
Thanks for the updates!

PR #2376, opened 2022-04-01
Fix config sanitization threading bug

mtrojan-ub:
If you have a regex in the INI, e.g. `[\\s/:;,=(\\[]*$`
getSanitizedConfigSection() is causing problems and returns `[\s/:`
so it should be better to use getConfigSection() instead, which returns `[\s/:;,=(\[]*$`

TODO
- [x] Close [VUFIND-1546](https://vufind.org/jira/browse/VUFIND-1546) when merging

### DIFF
--- a/import/index_java/src/org/vufind/index/ConfigManager.java
+++ b/import/index_java/src/org/vufind/index/ConfigManager.java
@@ -172,9 +172,9 @@ public class ConfigManager
      * @param filename configuration file name
      * @param section section name within the file
      */
-    public Map<String, String> getSanitizedConfigSection(String filename, String section)
+    public Map<String, String> getConfigSection(String filename, String section)
     {
-        Map<String, String> retVal = getConfigSection(filename, section);
+        Map<String, String> retVal = getRawConfigSection(filename, section);
         if (retVal == null) {
             return new ConcurrentHashMap<String, String>();
         }
@@ -195,7 +195,7 @@ public class ConfigManager
      * @param filename configuration file name
      * @param section section name within the file
      */
-    public Map<String, String> getConfigSection(String filename, String section)
+    public Map<String, String> getRawConfigSection(String filename, String section)
     {
         // Grab the ini file.
         Ini ini = loadConfigFile(filename);
@@ -229,14 +229,22 @@ public class ConfigManager
     }
 
     /**
+     * @deprecated
+     */
+    public Map<String, String> getSanitizedConfigSection(String filename, String section)
+    {
+        return getConfigSection(filename, section);
+    }
+
+    /**
      * Get a setting from a VuFind configuration file and sanitize the value.
      * @param filename configuration file name
      * @param section section name within the file
      * @param setting setting name within the section
      */
-    public String getSanitizedConfigSetting(String filename, String section, String setting)
+    public String getConfigSetting(String filename, String section, String setting)
     {
-        String retVal = getConfigSetting(filename, section, setting);
+        String retVal = getRawConfigSetting(filename, section, setting);
         return retVal == null ? retVal : sanitizeConfigSetting(retVal);
     }
 
@@ -246,7 +254,7 @@ public class ConfigManager
      * @param section section name within the file
      * @param setting setting name within the section
      */
-    public String getConfigSetting(String filename, String section, String setting)
+    public String getRawConfigSetting(String filename, String section, String setting)
     {
         Map<String, String> sectionMap = getConfigSection(filename, section);
         if (sectionMap == null) {
@@ -258,6 +266,14 @@ public class ConfigManager
             logger.warn(section + "." + setting + " setting missing from " + filename);
         }
         return retVal;
+    }
+
+    /**
+     * @deprecated
+     */
+    public String getSanitizedConfigSetting(String filename, String section, String setting)
+    {
+        return getConfigSetting(filename, section, setting);
     }
 
     /**
--- a/import/index_java/src/org/vufind/index/ConfigManager.java
+++ b/import/index_java/src/org/vufind/index/ConfigManager.java
@@ -170,7 +170,6 @@ public class ConfigManager
     {
         Map<String, String> retVal = getConfigSection(filename, section);
         if (retVal == null) {
-            logger.warn(section + " section missing from " + filename);
             return new ConcurrentHashMap<String, String>();
         }
 
@@ -217,6 +216,9 @@ public class ConfigManager
             }
         }
 
+        if (retVal == null) {
+            logger.warn(section + " section missing from " + filename);
+        }
         return retVal;
     }
 
@@ -230,7 +232,6 @@ public class ConfigManager
     {
         String retVal = getConfigSetting(filename, section, setting);
         if (retVal == null) {
-            logger.warn(section + "." + setting + " setting missing from " + filename);
             return retVal;
         }
 
@@ -251,7 +252,15 @@ public class ConfigManager
     public String getConfigSetting(String filename, String section, String setting)
     {
         Map<String, String> sectionMap = getConfigSection(filename, section);
-        return sectionMap.get(setting);
+        if (sectionMap == null) {
+            return null;
+        }
+
+        String retVal = sectionMap.get(setting);
+        if (retVal == null) {
+            logger.warn(section + "." + setting + " setting missing from " + filename);
+        }
+        return retVal;
     }
 
     /**
--- a/import/index_java/src/org/vufind/index/ConfigManager.java
+++ b/import/index_java/src/org/vufind/index/ConfigManager.java
@@ -109,21 +109,27 @@ public class ConfigManager
      */
     private String sanitizeConfigSetting(String str)
     {
+        // Work on a copy of the string.
+        // We do not want the sanitizer to update the cache, because it might
+        // cause problems when executing them multiple times, like
+        // e.g. in multithreaded scenarios.
+        String retVal = new String(str);
+
         // Drop comments if necessary; if the semi-colon is inside quotes, leave
         // it alone. TODO: handle complex cases with comment AND quoted semi-colon
-        int pos = str.indexOf(';');
-        if (pos >= 0 && !str.matches("\"[^\"]*;[^\"]*\"")) {
-            str = str.substring(0, pos).trim();
+        int pos = retVal.indexOf(';');
+        if (pos >= 0 && !retVal.matches("\"[^\"]*;[^\"]*\"")) {
+            retVal = retVal.substring(0, pos).trim();
         }
 
         // Strip wrapping quotes if necessary (the ini reader won't do this for us):
-        if (str.startsWith("\"")) {
-            str = str.substring(1, str.length());
+        if (retVal.startsWith("\"")) {
+            retVal = retVal.substring(1, retVal.length());
         }
-        if (str.endsWith("\"")) {
-            str = str.substring(0, str.length() - 1);
+        if (retVal.endsWith("\"")) {
+            retVal = retVal.substring(0, retVal.length() - 1);
         }
-        return str;
+        return retVal;
     }
 
     /**
@@ -179,7 +185,7 @@ public class ConfigManager
         // e.g. in multithreaded scenarios.
         Map<String, String> retValCopy = new ConcurrentHashMap<>();
         for (Map.Entry<String, String> entry : retVal.entrySet()) {
-            retValCopy.put(entry.getKey(), sanitizeConfigSetting(new String(entry.getValue())));
+            retValCopy.put(entry.getKey(), sanitizeConfigSetting(entry.getValue()));
         }
         return retValCopy;
     }
@@ -231,16 +237,7 @@ public class ConfigManager
     public String getSanitizedConfigSetting(String filename, String section, String setting)
     {
         String retVal = getConfigSetting(filename, section, setting);
-        if (retVal == null) {
-            return retVal;
-        }
-
-        // Return a copy of the string.
-        // We do not want the sanitizer to update the cache, because it might
-        // cause problems when executing them multiple times, like
-        // e.g. in multithreaded scenarios.
-        String retValCopy = new String(retVal);
-        return sanitizeConfigSetting(retValCopy);
+        return retVal == null ? retVal : sanitizeConfigSetting(retVal);
     }
 
     /**
--- a/import/index_java/src/org/vufind/index/ConfigManager.java
+++ b/import/index_java/src/org/vufind/index/ConfigManager.java
@@ -174,8 +174,8 @@ public class ConfigManager
      */
     public Map<String, String> getConfigSection(String filename, String section)
     {
-        Map<String, String> retVal = getRawConfigSection(filename, section);
-        if (retVal == null) {
+        Map<String, String> rawSection = getRawConfigSection(filename, section);
+        if (rawSection == null) {
             return new ConcurrentHashMap<String, String>();
         }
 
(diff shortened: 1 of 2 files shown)


demiankatz:
@mtrojan-ub, I am not able to reproduce the problem you are describing, and I do not believe that the proposed fix is correct. The purpose of getSanitizedConfigSettings is to strip wrapping quotation marks off of values in the .ini file, since the Java-based .ini reader does not do this. The regular expression will not work correctly if it includes literal wrapping quotes that are not intended to be there.

The problem you are seeing appears to be that the rule for stripping comments is corrupting one of the values... but there is a safety mechanism to prevent this problem, as semi-colons inside double quotes should NOT get stripped (see https://github.com/vufind-org/vufind/blob/dev/import/index_java/src/org/vufind/index/ConfigManager.java#L112-L117). When I add a `System.out.println` to dump out the value of `all` in the PunctuationContainer, I get `{0=[[\s/:;,=(\[]*$], 1=[\. -\.?$]}` from the current dev branch code, which is exactly what I expect to see.

Is it somehow possible that the code you are running is out of sync with the dev branch? It might be worth diffing the Java code to be sure there isn't a missing change somewhere. If that doesn't turn up the cause of your problem, maybe it's some environmental factor like Java version. For what it's worth, here's the environment I just tested with:

openjdk 11.0.14.1 2022-02-08
OpenJDK Runtime Environment (build 11.0.14.1+1-Ubuntu-0ubuntu1.20.04)
OpenJDK 64-Bit Server VM (build 11.0.14.1+1-Ubuntu-0ubuntu1.20.04, mixed mode, sharing)

Quite mysterious! In any case, thanks for reporting this, and hopefully we can get to the bottom of it.

mtrojan-ub:
@demiankatz, thanks for your response! This is a tricky one, but i think i know why it can't be reproduced with the original code.

(Neither are there any changes on our side regarding the ConfigManager or PunctuationContainer, nor switching our java version from 8 to 11 made any difference.)

After some more testing i found out that on my local machine, the config gets loaded multiple times. The first time the value seems to be OK (even when sanitizing is active), the second time the value gets broken. I guess this is a multithreading problem since we use `-Dsolrmarc.indexer.threadcount=4`. If we use 1 instead, this problem doesn't seem to occur anymore.

I will investigate further when time permits and see how thread-safety can be improved at that point.

demiankatz:
Thanks, @mtrojan-ub, that makes perfect sense! While the shared instance of the stored configuration is using a thread-safe property, there must be some code somewhere that needs locking added to prevent undesirable manipulation of the stored results. At a glance, I can't figure out what that might be, though! It seems to me that the more likely bug is that the configuration gets loaded multiple times and you end up with extra copies of the regular expressions... 

Maybe the problem is in the ConfigManager; I think the code is written on the assumption that it is returning copies of things, but if it's actually returning pointers/references, maybe it's doing in-place modification.

Let me know if I can be of any help with cleaning this up! My Java is pretty rusty (and my threaded Java even more so), but I'm happy to help if needed.

mtrojan-ub:
The problem seems to be that due to pointer logic in Java, the result from sanitizeConfigSetting will be applied onto the configCache, which will lead to sanitizeConfigSetting being executed another time when the next thread loads and sanitizes the same section over and over again. 

I tried to solve this problem by only returning a copy of the cache's entry. Or is there any better solution?

import/index_java/src/org/vufind/index/ConfigManager.java
```diff
@@ -210,7 +210,17 @@ public Ini loadConfigFile(String filename)
...
             }
         }
-        return retVal;
+
+        // Return a copy of the section.
+        // If there are other operations performed later (e.g. sanitize)
+        // we do not want them to update the cache, because they might cause
+        // problems when executing them multiple times, like
+        // e.g. in multithreaded scenarios.
+        Map<String, String> retValCopy = new ConcurrentHashMap<>();
```
demiankatz: I guess the question is one of efficiency vs. safety -- this solution seems fine to me, but maybe it would be more efficient (though also more likely to permit different errors) if we instead did the copy operation inside getSanitizedConfigSetting. Perhaps the ideal situation would be to make the ini object return immutable maps to prevent errors rather than making copies as a precaution (though I'm not sure if there's a way to do that -- quite possibly there is not an easy option).

If we don't move the logic, I do think we at least need to add a case to check if retVal is null, because otherwise there may be an attempt to perform illegal operations if we request a missing section. (I suspect there may also be some existing bugs in the other logic above if we request a missing section in combination with parent/override configuration files).

demiankatz (commented):
Thanks, @mtrojan-ub! I could certainly live with this approach, but I left some thoughts below for your consideration.

import/index_java/src/org/vufind/index/ConfigManager.java
mtrojan-ub: It might even get more complicated.

For sections, there are 2 public methods: getConfigSection and getSanitizedConfigSection.
For settings, there is only 1 public method: getConfigSetting.
=> If we have 2 versions of getConfigSection, shouldn't we have a getSanitizedConfigSetting as well and sanitize only if this one is called?
=> Since getConfigSetting does not use getConfigSection, we need to fix the logic in both code paths. (maybe the "sanitized" versions only?)

demiankatz: Good point! I wish we could figure out a good way to write unit tests for this Java code, because these are all the kinds of things that would be more easily addressed with test coverage. But that's a whole other rabbit hole...

It does seem to make sense to add a getSanitizedConfigSetting method... and possibly to refactor the existing getConfigSetting to leverage getConfigSection so we don't have to implement the same logic in multiple places.

import/index_java/src/org/vufind/index/ConfigManager.java
```diff
@@ -210,55 +216,42 @@ public Ini loadConfigFile(String filename)
...
-        String override = ini.get("Extra_Config", "local_overrides");
-        if (override != null) {
-            Ini overrideIni = loadConfigFile(override);
-            retVal = overrideIni.get(section, setting);
-            if (retVal != null) {
-                return sanitizeConfigSetting(retVal);
-            }
+        String retVal = getConfigSetting(filename, section, setting);
+        if (retVal == null) {
+            logger.warn(section + "." + setting + " setting missing from " + filename);
```
demiankatz: I'm not sure if a warning is necessary -- at least, I'm not sure why we would want to warn when requesting a sanitized setting, but not when requesting an unsanitized one.

import/index_java/src/org/vufind/index/ConfigManager.java
```diff
@@ -210,55 +216,42 @@ public Ini loadConfigFile(String filename)
...
-            } else {
-                break;
-            }
-        }
+        // Return a copy of the string.
+        // We do not want the sanitizer to update the cache, because it might
+        // cause problems when executing them multiple times, like
+        // e.g. in multithreaded scenarios.
+        String retValCopy = new String(retVal);
+        return sanitizeConfigSetting(retValCopy);
```
demiankatz: Would it be better to move the string copying logic inside sanitizeConfigSetting so we don't need to do it in two places? Since sanitizeConfigSetting returns a value, I don't think it was actually my intention for it to do in-place modification of values -- I just didn't realize that the argument was being treated as a reference.

demiankatz (commented):
Thanks again, @mtrojan-ub -- a couple of additional thoughts.

import/index_java/src/org/vufind/index/ConfigManager.java
mtrojan-ub: I just implemented in in the new function the same way as it was before, but you're right! I'm gonna move it to the non-sanitized version in both cases.

import/index_java/src/org/vufind/index/ConfigManager.java
mtrojan-ub: I had that idea as well and tried it out, but the problem is => due to the logic in e.g. getSanitizedConfigSection, the retVal points to the original cache entry, so even if sanitizeConfigSetting would work on a copy, this would not be enough.

demiankatz: Yeah, I agree that copying strings in sanitizeConfigSetting is not enough by itself to solve the problem; you still need to copy the section in getSanitizedConfigSection(). However, if you copy the string in sanitizeConfigSetting, then you don't have to copy the values in getSanitizedConfigSetting and getSanitizedConfigSection, and you could in fact probably reduce the end of getSanitizedConfigSetting to simply `return retVal == null ? retVal : sanitizeConfigSetting(retVal);`

I don't have really strong feelings about this if you don't want to change it further, but it still feels like it might be a worthwhile simplification.

mtrojan-ub: Done!

mtrojan-ub:
And by the way... since we change the default behaviour of getConfigSetting to non-sanitized:
There seem to be certain calls to getConfigSetting in CreatorTools, FullTextTools and DatabaseManager. Should they be changed to call the new getSanitizedConfigSetting instead?

demiankatz:
Good point, @mtrojan-ub -- I had overlooked the fact that we were changing the behavior of getConfigSetting. I don't think we want to do that, since suddenly switching our standard configuration retrieval mechanism to return unsanitized values could cause nasty side effects in people's local code. I'd suggest two possible approaches: add a "sanitize" parameter to getConfigSetting which defaults to true, or keep the current behavior of getConfigSetting, and switch the names around so that we keep getConfigSetting with its current behavior and add getUnsanitizedConfigSetting or getRawConfigSetting. If you're concerned about consistency of naming, we could also rename getConfigSection to getRawConfigSection and getSanitizedConfigSection to getConfigSection (with a deprecated getSanitizedConfigSection for backward compatibility). I'd feel better about changing the default behavior of getConfigSection to be safer than changing the default behavior of getConfigSetting to be less safe.

mtrojan-ub:
Alright - and i also changed the remaining direct calls of getSanitized... operations to use the getConfig... ones instead.

import/index_java/src/org/vufind/index/ConfigManager.java
```diff
@@ -210,55 +221,59 @@ public Ini loadConfigFile(String filename)
...
-            }
+    /**
+     * Get a setting from a VuFind configuration file.
+     * @param filename configuration file name
+     * @param section section name within the file
+     * @param setting setting name within the section
+     */
+    public String getRawConfigSetting(String filename, String section, String setting)
+    {
+        Map<String, String> sectionMap = getConfigSection(filename, section);
```
demiankatz: Now that getConfigSection returns sanitized values, this is going to prevent getRawConfigSetting from returning raw config settings. I assume you want to use getRawConfigSection instead.

import/index_java/src/org/vufind/index/ConfigManager.java
```diff
@@ -210,55 +221,59 @@ public Ini loadConfigFile(String filename)
...
-                        + " (" + e.getMessage() + ")"
-                    );
-                }
-                retVal = ini.get(section, setting);
-            } else {
-                break;
-            }
+        String retVal = sectionMap.get(setting);
+        if (retVal == null) {
+            logger.warn(section + "." + setting + " setting missing from " + filename);
```
demiankatz: Upon reviewing the way getConfigSetting is used in the rest of the code, I don't think this warning is a good idea. There are many places where we request optional settings and act upon null return values, and all of those scenarios will fill the log with warnings. I think if we have logging at all, it should be "opt in" either through an argument or through a separate method (like getRequiredConfigSetting/getRawRequiredConfigSetting). I'd be inclined to just remove this for simplicity, but if you think it's important, we can discuss further! Maybe that's a project for a separate PR, though, since there's already a lot going on in this one.

import/index_java/src/org/vufind/index/ConfigManager.java
```diff
@@ -166,25 +172,30 @@ public Ini loadConfigFile(String filename)
      * @param filename configuration file name
      * @param section section name within the file
      */
-    public Map<String, String> getSanitizedConfigSection(String filename, String section)
+    public Map<String, String> getConfigSection(String filename, String section)
     {
-        Map<String, String> retVal = getConfigSection(filename, section);
+        Map<String, String> retVal = getRawConfigSection(filename, section);
```
demiankatz: I think this might be more clear if we rename the variables: retVal --> rawSection and retValCopy --> retVal. Only one of these is actually a return value.

import/index_java/src/org/vufind/index/ConfigManager.java
```diff
@@ -210,55 +221,59 @@ public Ini loadConfigFile(String filename)
                 retVal.put(key, overrideSection.get(key));
             }
         }
+
+        if (retVal == null) {
+            logger.warn(section + " section missing from " + filename);
```
demiankatz: Since this code is now getting used more broadly (because of refactoring of the "get setting" methods and changes to default behavior), I think this log message is now going to be problematic for the same reasons as the other log message I mentioned. Perhaps we need to consider the same options I suggested there: adding an argument, adding a method for fetching "required" sections, or a third option I didn't propose before: doing the check and log after calling the method rather than inside the method. What do you think?

demiankatz (changes requested):
Thanks, @mtrojan-ub, this keeps getting better. However, I think I spotted one bug, plus I have concerns about the impact this is going to have on log noisiness. See details below.

import/index_java/src/org/vufind/index/ConfigManager.java
mtrojan-ub: Not sure what should happen in these cases. If a setting or section really is "required", then it feels like logging a warning isn't enough and there should be an error or even exception instead.
And also, I don't think it would be a good idea to have long function names like getRawRequiredSanitizedConfigSection()... another option would be to have a global flag to define whether these warnings should be logged or not, or to set them to debug and control that over the log level. But i'm also not sure what the best solution would be here...

demiankatz: I think my inclination is to drop the logging entirely. If something is required, the calling code should check the return value. If it's not required, having this log message is not helpful. The only place where this message would have been displayed before would have been from the calls you refactored in PunctuationContainer.java... so maybe we should log messages in there if getConfigSection returns empty values. Or maybe we don't even need to do that.

mtrojan-ub: Alright, so i removed the logging from the ConfigManager and added some checks to the PunctuationContainer. Hope it looks better now!

mtrojan-ub:
The more i think of it, the more i'm tempted to add a sanitized cache... if we import millions of records and all settings will be sanitized each time they're queried, we might cause lots of unneccessary string operations too.

import/index_java/src/org/vufind/index/ConfigManager.java
```diff
@@ -210,55 +225,48 @@ public Ini loadConfigFile(String filename)
...
+    public String getRawConfigSetting(String filename, String section, String setting)
+    {
+        Map<String, String> sectionMap = getRawConfigSection(filename, section);
+        return sectionMap == null ? null : sectionMap.get(setting);
+    }
 
-        // Return the processed setting:
-        return retVal == null ? null : sanitizeConfigSetting(retVal);
+    /**
+     * @deprecated
```
demiankatz: Would it be better to use the `@Deprecated` Java annotation outside of the comment? Or do we possibly need both? I'm not sure of the best convention, but while I was testing, I noticed that the compiler was throwing warnings about having deprecated methods without the annotation. Any thoughts?

demiankatz (commented):
Thanks, @mtrojan-ub. I think this is fully functional and just about ready to merge, but I have one last question based on my final round of testing...

demiankatz:
@mtrojan-ub, since this feels like a problem that others may run into, I've also documented it in a [JIRA ticket](https://vufind.org/jira/browse/VUFIND-1546). I've also added a couple of TODOs to the top of this PR to remind myself to update the ticket and changelog when this gets merged.

import/index_java/src/org/vufind/index/ConfigManager.java
mtrojan-ub: I'm no real Java expert so i can only guess, and even SolrMarc doesn't seem to be consistent in this case...

Most IDEs support JavaDoc and e.g. in Netbeans the deprecated functions will be highlighted in a special way. I think the only advantage of using the `@Deprecated` annotation might be to display warnings when compiling or executing so that people would see it more often when migrating to a new version - and that actually could be worth doing it. Some people even recommend to do both, to mark as `@Deprecated` and provide additional information in the JavaDoc as well (in Java 9 `@Deprecated` received additional attributes like "since" and "forRemoval"), but of course many people still rely on 8 and also we could add additional info in the JavaDoc like e.g. information about the successor-function. 

What do you think?

demiankatz (approved):

import/index_java/src/org/vufind/index/ConfigManager.java
demiankatz: I like what you've done; I think this is the most useful approach. I have approved the PR and will merge soon -- just need to find time to write up the changelog notes. :-)

PR #2378, opened 2022-04-05
Make MenuButton Component

crhallberg:
Example for future development.

## Todo

- [x] `->component()` view helper
- [x] Make sure tag options conform to accessibility
- [ ] Compare to finna components
- [x] Ensure replacement of all menubutton/dropdown uses
- [x] Add unit test for new view helper

## Future dev

- [ ] Move JavaScript from Bootstrap

A reference of what I want to keep in mind for component standards: https://github.com/webcomponents/gold-standard/wiki.


themes/bootstrap3/templates/_ui/components/menu-button.phtml
```diff
@@ -0,0 +1,24 @@
...
+  $wrapperClass = $this->wrapperClass ?? 'li';
+
+  $toggleTag = $this->toggleTag ?? 'a';
+  $toggleAttrs = $this->toggleAttrs ?? [];
+?>
+
+<<?=$wrapperTag ?> class="<?=$wrapperClass ?> dropdown">
+  <<?=$toggleTag ?> href="#" class="dropdown-toggle" data-toggle="dropdown">
+    <?=$this->transEsc($this->toggleLabel)?>
+    <b class="caret"></b>
```
demiankatz: Is this an icon? Even if not an icon, would a class name that's more functional (e.g. dropdown-indicator) make sense, in case we choose to style it with something other than a caret?

themes/bootstrap3/templates/_ui/components/menu-button.phtml
```diff
@@ -0,0 +1,24 @@
...
+  </<?=$toggleTag ?>>
+
+  <ul class="dropdown-menu">
+    <?php foreach ($this->menuItems as $current): ?>
+      <li<?php if ($current['selected']): ?> class="active"<?php endif ?>>
+        <a href="<?=$this->escapeHtmlAttr($current['url']) ?>" rel="nofollow">
+          <?=$current['label'] ?>
```
demiankatz: Does this need escaping, or are we assuming pre-escaped labels? Might be worth thinking about which assumption to make, or offering an option to control escaping behavior.

themes/bootstrap3/templates/header.phtml
```diff
@@ -48,33 +48,40 @@
...
-                      <?=$this->transEsc($current['desc']) ?>
-                    </a>
-                  </li>
-                <?php endforeach; ?>
-              </ul>
-            </li>
+            <?php
+              $themeMenuItems = [];
+
+              foreach ($this->layout()->themeOptions as $current) {
```
demiankatz: I wonder if array_map with a callback would be slightly more elegant... but maybe I'm just getting too fancy.

demiankatz (commented):
A few initial thoughts, though I didn't take a super-close look since I'm sure much will change as we work through the TODO checklist.

themes/bootstrap3/templates/_ui/components/menu-button.phtml
crhallberg: This is a Bootstrap component piece, but we can probably replace it.

demiankatz:
Thanks for the latest update, @crhallberg. I like your approach for the view helper -- I think keeping the component names very short and simple makes sense for now. We may eventually want longer or hierarchical component names for organizational reasons, but I'd rather not over-engineer this in advance. We could always set up an alias registry if we later come up with something more complicated and need backward compatibility.

themes/bootstrap3/templates/_ui/components/menu-button.phtml
ckaz: I may misunderstand what you're after, but I'd opt for removing the empty \<b>\-tag. It makes sense but I guess it's more of a gut feeling thing. 
What I'd prefer would be something resulting in an item more in the vein of
\<a 
     class="<?=$dropdown-indicator-class ?> dropdown-toggle" 
     href="#" 
     role="button" 
     id="..."
     data-toggle="dropdown" 
     aria-haspopup="true" 
     aria-expanded="false">\
    Dropdown menu text AND/OR Icon + sr-only or arial-label or arial-labeledby
\</a>

<?=$dropdown-indicator-class ?> we could use if we want it to be a button, say "btn btn-secondary" -- maybe even more specific such as <?=$header-dropdown-indicator-class ?> and <?=$searchbox-dropdown-indicator-class ?> 

An id would be usefull too for using in aria-labeledby for the dropdown-menu item itself.

Should it also be possible to turn this into a button as part of a button group as for bulk actions? So we get something like
\<button 
   type="button" 
   class="\<?=$dropdown-indicator-class ?> dropdown-toggle" 
   data-toggle="dropdown" 
   aria-haspopup="true" 
   aria-expanded="false">
    Dropdown button text AND/OR Icon + sr-only or arial-label or arial-labeledby
\</button>

themes/bootstrap3/templates/header.phtml
crhallberg: I switched both to array_map and while the code is shorter, I'm not sure if this improves readability.

demiankatz: I think it depends on whether or not you like functional programming, which is really a matter of opinion. I personally think about things more procedurally, but I see the elegance of the functional approach and try to apply it when it feels appropriate.

ckaz:
Apologies for chiming in so late -- I've been down with Covid.

In our accessibility test results we were asked to fit toggler/dropdown buttons with an indication whether the element they control is open/expanded (aria-expanded="true" or "false") or not, like so:

`<button ... aria-controls="..." aria-expanded="..." aria-label="...">
  ...
 </button>`
 
 and ideally an aria-label that describes what they do, in the vein of "show hide main navigation" or "show or hide language selection". In some cases an aria-haspopup="..." may be useful too.
 
My best!
Claas

aleksip:
Unfortunately I have been busy with other work and have not had time to follow this and other issues. I would just like to suggest an alternative approach to the view helper: a custom template resolver.

This way everything that supports templates in general would gain support for component templates with no code changes. I am thinking about ViewModels in particular. [Here is an example](https://github.com/NatLibFi/NDL-VuFind2/commit/9f4cb522fd1c8a047e8c732876b6f1addb2f4e2b) from Finna, where we have a resolver that expands shortened versions of component template names.

I have been thinking that we could borrow an idea from Twig: configurable template namespaces. So we could map `@components` to a specific directory, and `$this->render('@components/menu-button', $values)` for example. Or, we could map even more specific namespaces like `@atoms`, `@molecules`, `@organisms` etc.

aleksip:
Just a quick addition to my previous comment: namespaces could also be used for other things than just components. Having developed Twig templates for many years I have found namespaces to be a very useful feature, and it is even a bit surprising that laminas-view does not natively support them.

demiankatz:
@ckaz, sorry to hear about your illness -- welcome back, and I hope you're feeling better! Thanks for the suggestions; I'm sure @crhallberg will incorporate some of them.

Thanks also for the template namespace idea, @aleksip. I agree that this sounds potentially useful and may have applications beyond this case. As is often the case, I think the challenge is figuring out a plan of action to move forward -- that might be a bigger change than we want to introduce in release 8.1 (which I'm hoping to schedule in a couple of months), but maybe if we target 9.0 we'll have time to get it right. Does it make sense to start a separate PR to explore that idea? And if so, what should we do here in the meantime? Do we stick with the view helper on the understanding that it might get deprecated in future? Do we revert to direct render calls on the understanding that the template names might change to a namespace in the future? I don't have strong feelings about any of the implementation details, but I would like to find a way to continue moving forward on this without locking in decisions we're going to regret later....

aleksip:
Comparing just the syntax, the difference would be

`$this->component('menu-button', $values)`

vs

`$this->render('@components/menu-button', $values)`

or, alternatively, e.g.

`$this->render('components::menu-button', $values)`

So not a huge difference, although of course the view helper version does fit in more nicely with

`$this->icon('dropdown-caret')`

and other similar existing or future view helpers.

One thing I think should be considered at this stage is the organization of a large number of components. We probably don't want to put all component templates in one directory, so the currently deceivingly simple and elegant `$this->component('menu-button', $values)` could eventually end up being a not-as-simple `$this->component('much/deeper/folder/structure/menu-button', $values)`. (Edit: just remembered that @demiankatz  had already mentioned this earlier, and suggested an alias registry as a solution.)

I have no strong objections on going the helper route, but I do think that the better framework compatibility, possibility to simplify multiple deep folder structures and other potential applications are definite pluses for the namespace approach.

One way to go forward would be to add a simple custom template resolver hardcoded for the current use case/single namespace, and we could then add the configurability later.

We could also support both approaches, so the helper would just be changed to use the namespace if/when we add the namespace support at a later stage. The helper would then be the preferred choice in templates, but e.g. ViewModels could use namespaces.

demiankatz:
@aleksip, I think I like the approach of using the helper for now, since it doesn't cut us off from the option of using namespaces, and we can always deprecate the helper in future if we opt for a more framework-compatible road (though I think all of these proposed solutions have some degree of VuFind-specificity and extra abstraction; it's hard to entirely avoid). I think the helper approach seems a natural way to set up the alias registry, as discussed above -- having something equivalent to that work by magic within custom template resolver code seems potentially less intuitive, though maybe I'm just not thinking it through all the way. As I said, I certainly welcome a PR to start some kind of namespace experiment, and I think this is a good thing for @crhallberg to keep in mind as he works through theme design issues. Perhaps use of namespaces could even be a way to maintain some degree of compatibility between legacy and new themes as we transition -- i.e. if we use namespaced names to load templates, we could move some things around as long as the namespaces resolve to the right places in the context of the right themes. Though maybe that's not a good idea. Just brainstorming! :-)

aleksip:
@demiankatz No magic needed for the template resolver: the code and alias/namespace configuration for both the view helper and the template resolver would be almost identical. Template resolvers are what Laminas documentation [suggests using](https://docs.mezzio.dev/mezzio/v3/features/template/interface/) to implement namespace support.

I know ViewModels are currently only used in a rather "monolithic" way in VuFind, but the reason I've mentioned them is that they can be used as representations of UI components in backend code. If we are moving towards components in the UI layer, moving towards organizing backend functionality in a similar way could make sense.

I might well be missing something important here, but ViewModels seem like the framework designed link between backend and frontend code, and adding what is basically template resolving to a view helper seems to weaken that link.

Just to show that this is not purely theoretical, we have already implemented some custom elements as [backend components](https://github.com/NatLibFi/NDL-VuFind2/blob/dev/module/Finna/src/Finna/View/CustomElement/AbstractBase.php) with a corresponding component template, and used ViewModels to link them.

demiankatz:
Thanks, @aleksip, that's good to know. I think I need to find some time for more reading/study to get a better handle on how all these pieces fit together. The current way we use view models, helpers and templates is the way it is largely because it worked in 2011 and still works today... but if there are better ways that we can transition to that better leverage the framework and allow for better code organization, I'm all for that... I just want to be sure that we can reach it in as non-disruptive a way possible, and that we don't compromise any of the current customization/overriding behaviors we currently have.

I think the piece that I'm still missing in my mental model of this is how the custom backend component you linked to fits into things. What creates that class and how?

And just for a thought on possible future direction: right now, the Record view helper does a lot of work looking up templates based on record driver classes. Do you think this is something that could be handled similarly by a template resolver through a RecordDriver namespace? I can see how that might be a bit more intuitive and readable... but it's also quite complex (e.g. depending on the $driver variable to figure out which template to load), so it might be more than the template resolver mechanism could handle.

Also, are you proposing that each namespace would have its own custom template resolver class? If so, that would enable us to encapsulate functionality in a tidy way, though it might also get unwieldy and confusing if we have a lot of namespaces that all behave differently. In that regard, using view helpers seems more explicit to me, because you expect them to all be different. If things act wildly differently based only on the template name, that's pretty unintuitive (though conceptually, maybe it's okay).

crhallberg:
Just checking in on this conversation: we want to keep the Component helper or do we want to namespaced render?

demiankatz:
It looks like @aleksip hasn't had a chance to answer my last round of questions yet; I'll still be interested to hear more when time permits. In the meantime, I think it probably makes sense to stick with the view helper solution for now, given that it matches the status quo. If we determine that namespaced templates are a better solution, we could potentially end up replacing/refactoring a number of existing helpers, and that's probably a separate project unto itself.

aleksip:
@demiankatz I'm sorry, I had completely forgotten about the specific questions you made!

The backend components I linked as an example are requested from the service manager by a custom element view helper called by a custom Markdown parser.

More generally the idea is about allowing backend code to directly specify view component templates in a ViewModel. Of course an intermediary template that then calls a component view helper can be specified too.

I had only thought of a single resolver that just maps a name to a specific path without any additional logic. A record driver namespace is an interesting use case, but indeed could get quite confusing. 

Namespaces should probably be something that can always be used in view component templates. What I'm advocating is that view components should be a very specific class of templates, containing presentation logic only. Record driver templates are a good example of templates that I feel are in a "grey area" between backend and frontend, since they have access to the driver class instead of just pure presentation data.

A component view helper would be fine in this class of templates, matches the status quo, and can be called from intermediary templates.

demiankatz:
Thanks, @aleksip, that makes sense. So your proposal is that it might be helpful to add a single custom template resolver to VuFind to add a mechanism that maps specific namespaces to specific file paths. This could be configuration driven (e.g. map of namespace => relative base path) and would result in (at least potentially) some more readable include statements when templates load one another.

When you say that component templates should contain "presentation logic only," how exactly would you draw that line? My first thought was "no reliance on view helpers" but of course that doesn't really work since you need at a minimum other presentation-oriented helpers for things like translation and escaping. Just trying to figure out what rules of thumb need to be established there.

In any case, it sounds to me like there is no present obstacle to @crhallberg moving forward with the current helper-based solution; we may just end up changing some of the internals over time and/or deprecating the helper if it becomes unnecessary in the future. But please correct me if I'm misunderstanding your recommendations!

aleksip:
@demiankatz That was a very good summary of my proposal! :)

In addition to better readability, namespaces enable location independence for view components, so they can more easily be developed in an independent way, even in a different project.

The "presentation logic only" line could be hard to draw in some cases, but I have [sketched some rules](https://github.com/NatLibFi/finna-ui-components/wiki/Template-Files) in our component project wiki. I have managed to refactor most of our administration interface UI as view components using just those helpers.

The new icon and component helpers should be allowed too, for example. But the more view helpers we allow, the harder it will be to decouple view components and develop and [display them independently](https://natlibfi.github.io/finna-ui-components/).

To enable the use of Pattern Lab, I have written a very crude and simple ["engine"](https://github.com/aleksip/engine-phtml/tree/develop/lib) that provides dummy implementations of the allowed helpers. The icon helper is not currently supported, and is slightly more complex to implement, since it uses templates of its own, so it is not just pure PHP.

It might be possible to implement the allowed helpers as a separate module (published from a VuFind monorepo) that could be requred as is by the Pattern Lab engine, but that would need more investigation.

demiankatz:
Thanks, @aleksip!

Do you think it might be feasible to develop / configure some kind of tooling to identify the use of view helpers in templates? It would be interesting if we could restrict which helpers can be used in which directories at the continuous integration layer to ensure that we maintain discipline. A general report on helper usage in the code base might also be valuable for deciding which helpers might be easy candidates for elimination/refactoring if we want to clean things up in the future. This all seems like a solvable problem, but perhaps one that nobody has bothered to solve yet.

Regarding implementation of the icon helper for Pattern Lab, couldn't you just create an implementation that ignores the complexity of the VuFind implementation? I think in terms of the external interface it presents, the icon helper's internal use of templates is not really relevant -- you could easily create something much simpler for demonstration purposes.

themes/bootstrap3/templates/_ui/components/menu-button.phtml
```diff
@@ -0,0 +1,30 @@
+<?php
+  $wrapperTag = $this->wrapperTag ?? 'li';
+  $wrapperClass = $this->wrapperClass ?? 'li';
```
demiankatz: Does it really make sense for the default class to be `li`? For that matter, does it make sense for the default wrapper tag to be `li`? Isn't that kind of specific to our use case; would something like `div` be more generic?

themes/bootstrap3/templates/_ui/components/menu-button.phtml
```diff
@@ -0,0 +1,30 @@
...
+
+  $toggleTag = $this->toggleTag ?? 'button';
+  $toggleAttrs = $this->toggleAttrs ?? [];
+
+  $href = $this->href ?? $toggleTag == 'button' ? null : '#';
+?>
+
+<<?=$wrapperTag ?> class="<?=$wrapperClass ?> dropdown">
```
demiankatz: We should `escapeHtmlAttr` the `$wrapperClass` here.

themes/bootstrap3/templates/_ui/components/menu-button.phtml
```diff
@@ -0,0 +1,30 @@
...
+  <<?=$toggleTag ?>
+    class="dropdown-toggle"
+    data-toggle="dropdown"
+    <?php if ($href): ?>href="<?=$href ?>"<?php endif ?>
```
demiankatz: Do we need to escape `$href` too? In what situation would this be passed in from the outside? (This doesn't seem to be used in existing code here).

demiankatz (commented):
Thanks for continuing to move this forward, @crhallberg -- just a few questions based on a quick review.

themes/bootstrap3/templates/_ui/components/menu-button.phtml
crhallberg: The theme switcher and language dropdowns work off of hrefs with GET parameters.

demiankatz: Isn't it conceivable that the href could contain an `&` character that would need to be HTML escaped? (I realize that is probably not the case in the current applications of the template, but it seems like something that might happen in the future). Adding an escape to be safe seems wise, unless the data is coming in pre-escaped.

themes/bootstrap3/templates/header.phtml
```diff
@@ -48,33 +48,37 @@
...
-                  </li>
-                <?php endforeach; ?>
-              </ul>
-            </li>
+            <?=$this->component('menu-button', [
+              'wrapperClass' => 'language',
+              'toggleLabel' => 'Language',
+              'menuItems' => array_map(
+                  function ($langCode, $langName) {
+                  return [
```
demiankatz: Something weird seems to be going on with the indentation here; is php-cs-fixer messing something up?

themes/bootstrap3/templates/header.phtml
```diff
@@ -48,33 +48,37 @@
...
-                    </a>
-                  </li>
-                <?php endforeach; ?>
-              </ul>
-            </li>
+            <?=$this->component('menu-button', [
+              'toggleLabel' => 'Theme',
+              'menuItems' => array_map(
+                  function ($current) {
+                  $item = $current;
```
demiankatz: More weird indentation issues...

themes/bootstrap3/templates/_ui/components/menu-button.phtml
demiankatz: I see that you fixed the incorrect default wrapper class, but you didn't change the default tag. Is that intentional? (If so, that's fine, just want to be sure this comment wasn't overlooked before I resolve the thread).

demiankatz (changes requested):
Looks like there are some formatting issues here; let me know if you need any help sorting them out.

Also, it might be worth reviewing other open threads on this PR; I'm not sure which issues are still under discussion and which have been resolved. Hopefully we're nearing completion, but I want to be sure we haven't missed anything.

themes/bootstrap3/templates/_ui/components/menu-button.phtml
crhallberg: In truth, it would be a toss-up between `div` and `li` due to the high use of dropdowns inside other menus (ideally `ul`s for accessibility) but `div` is the more neutral of the two so I will change it.

themes/bootstrap3/templates/header.phtml
crhallberg: These were both odd cases. I don't think PCF touches the templates though so this was probably me missing some adjustments.

demiankatz:
@crhallberg, thanks for the progress on this. It looks like the PR now has some merge conflicts with dev; can you resolve those when time permits?

demiankatz:
Thanks for the merge, @crhallberg... now it looks like CI is failing; maybe just a simple php-cs-fixer thing.

crhallberg:
@demiankatz `php-cs-fixer` is suggesting some interesting indentations. Should I follow suit?

demiankatz:
@crhallberg, if php-cs-fixer is doing something goofy, there's probably a way to make it happy without creating illogical indentations. Let me know if you need help working out a solution; I have some experience in this. :-)

module/VuFind/src/VuFind/View/Helper/Bootstrap3/Component.php
```diff
@@ -0,0 +1,55 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  View_Helpers
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+namespace VuFind\View\Helper\Bootstrap3;
```
demiankatz: Is there a strong reason to put this in Bootstrap3 rather than Root? If we plan to use different template storage patterns on a per-theme basis, that would make sense, but if we want to set a global convention, it's probably cleaner to stick to Root so we don't have to move it around later.

themes/bootstrap3/templates/_ui/components/menu-button.phtml
```diff
@@ -0,0 +1,32 @@
...
+  $toggleTag = $this->toggleTag ?? 'button';
+  $toggleAttrs = $this->toggleAttrs ?? [];
+
+  $href = $this->href ?? $toggleTag == 'button' ? null : '#';
+?>
+
+<<?=$wrapperTag ?>
+  class="<?=$wrapperClass ? $this->escapeHtmlAttr($wrapperClass) . ' ' : '' ?>dropdown"
+>
+  <<?=$toggleTag ?>
```
demiankatz: Would this stuff be more readable if we used the makeTag view helper, now that we have it? Or do we want to avoid that in order to keep templates more "pure PHP/HTML"? I know there's been some back-and-forth debate about strategy here. I personally think the helper would actually help here, but maybe I'm wrong. :-)

themes/bootstrap3/templates/_ui/components/menu-button.phtml
```diff
@@ -0,0 +1,32 @@
...
+    data-toggle="dropdown"
+    <?php if ($href): ?>href="<?=$this->escapeHtmlAttr($href) ?>"<?php endif ?>
+  >
+    <?=$this->transEsc($this->toggleLabel) ?>
+    <?=$this->icon('dropdown-caret') ?>
+  </<?=$toggleTag ?>>
+
+  <ul class="dropdown-menu">
+    <?php foreach ($this->menuItems as $current): ?>
+      <li class="dropdown__item <?php if ($current['selected']): ?> active<?php endif ?>">
```
demiankatz: Do we want to be tolerant of missing values in the menuItems array? For example, should we use `$current['selected'] ?? false` to assume that if the attribute is missing, it's not selected? I can see that url and label should probably be required, though...

themes/bootstrap3/templates/header.phtml
```diff
@@ -65,33 +65,39 @@
...
-                  </li>
-                <?php endforeach; ?>
-              </ul>
-            </li>
+              <?=$this->component('menu-button', [
+                  'toggleLabel' => 'Theme',
+                  'wrapperTag' => 'li',
+                  'menuItems' => array_map(
+                      function ($current) {
+                          $item = $current;
```
demiankatz: Do we need this assignment? Can we just call the parameter $item? (I don't think this is getting passed by reference, so I don't think the assignment matters).

crhallberg: Sorry I spent a lot of time writing code where the linter yelled at me if I changed parameters at all. Fixed!

module/VuFind/src/VuFind/View/Helper/Bootstrap3/Component.php
crhallberg: The reason here is that this dropdown still relies on the Bootstrap library JS. A framework agnostic would definitely go in root.

demiankatz: A template dependency on the Bootstrap library would explain why the relevant _ui/component/x.phtml file should go in the bootstrap3 theme... but I think this helper is entirely theme/framework-agnostic and could move to Root. I think it's okay to separate the helper (which defines how component templates are located and loaded) from the templates themselves (which could easily have different implementations in different themes... or even entirely different component sets...). As long as, within any given theme, we only render components defined within that theme, I think we're in good shape, and the component helper doesn't need to care. Am I making any sense? :-)

crhallberg: I misunderstood and now agree.

module/VuFind/src/VuFind/View/Helper/Root/MakeTag.php
```diff
@@ -94,6 +95,7 @@ class MakeTag extends \Laminas\View\Helper\AbstractHelper
         'ins',
         'kbd',
         'label',
+        'li',
```
demiankatz: @crhallberg, I think I see what's going on here -- the list of "valid body tags" you found is literally "tags that are allowed as direct children of `<body>`". But you can't put a `<li>` in there because it needs to be contained by a list. Similarly, I notice that this list is missing things like `<tr>` and `<td>`. I'm sure there are others! It's probably worth opening a separate PR to address this and expand the list. What do you think?

demiankatz:
@crhallberg, in addition to the comment above regarding the MakeTag helper, it looks like merge conflicts need to be resolved so that the CI checks can go through.

module/VuFind/src/VuFind/View/Helper/Root/MakeTag.php
crhallberg: https://github.com/vufind-org/vufind/pull/2560

demiankatz: Thanks! I've merged the new PR and reconciled the conflicts here. Much better now!

crhallberg:
I went through all of the feedback today, trying to implement anything I missed.

The one piece I think needs further discussion is [how I implemented component namespacing](https://github.com/vufind-org/vufind/pull/2378/files#diff-4192787e0b2f09ccf6196e500135f0ea17337fc64d6a0ac2baab41640da3564fR55-R59) in response to @aleksip's comments above. This should at least be configurable/extendable.

aleksip:
Looks like a good minimum namespace implementation that can be extended later.

@crhallberg The current templates use menu-button without a namespace. Do you have any thoughts on when to use and when not to use namespaces? (Sorry if I have missed something in the above conversation.)

demiankatz:
@aleksip, it looks to me like the idea here is that the component helper defaults to the `@component` namespace but supports the use of other namespaces as needed. I think that makes sense in terms of keeping the most common case as concise as possible while allowing more flexibility, but I'm open to further discussion if you feel differently.

@crhallberg, this is looking good -- maybe a finishing touch would be a unit test for the view helper, especially now that it is doing non-trivial work.

demiankatz:
Also, in response to @crhallberg's request for further input on the namespacing implementation, I think this is okay as-is for now. For the moment, the helper takes care of mapping the namespace to the actual template location... but if we follow @aleksip's suggestion of handling namespacing within the view engine itself, then we can just revise the helper to pass along a namespaced name to the view, and then let the view resolve the actual path. I think if we need more detailed configurability, taking that route might make the most sense, and keeping it simple until then saves us from unnecessarily reinventing the wheel. Again, I'm open to more input from @aleksip if I'm not on the right track with these thoughts. :-)

aleksip:
I think moving in small steps is good and this is a fine step forward.

My question above is really about the "as needed" part. Are there thoughts on when the components should be organized in some way? If we start by putting everything in just one big pile, will things be harder to refactor later? If we end up with a decent amount of components, and want to use them in a tool like Pattern Lab or Storybook, organizing them in some way would be beneficial. I have no strong opinions or specific suggestions at this point, just wanted to bring the subject up.

Edit: just remembered that Storybook at least allows for independent organization so maybe this is not something to worry about from that perspective.

demiankatz:
@aleksip, that is a good point regarding organization. I suppose this is also an area where support for aliasing in the view helper could potentially allow us to apply more granular organization in the future while maintaining backward compatibility. (I'm not saying we need to implement aliasing support right now, just pointing out another way in which I think we can safely start with a flat organization while keeping multiple open paths to transitioning to something better-organized when we have a better sense of the scope of the project).

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/ComponentTest.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+/**
+ * Class ComponentTest
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Moravian Library 2020.
```
demiankatz: Should probably make this Villanova University 2022

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/ComponentTest.php
```diff
@@ -0,0 +1,78 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  VuFindTest\View\Helper\Root
+ * @author   Josef Moravec <moravec@mzk.cz>
```
demiankatz: Give yourself some credit! ;-)

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/ComponentTest.php
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @category VuFind
+ * @package  Tests
+ * @author   Josef Moravec <moravec@mzk.cz>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
+ */
+class ComponentTest extends \PHPUnit\Framework\TestCase
+{
+    use \VuFindTest\Feature\ViewTrait;
```
demiankatz: I don't think you're actually using this trait.

themes/bootstrap3/templates/_ui/components/menu-button.phtml
```diff
@@ -0,0 +1,46 @@
...
+      </li>
+    <?php endforeach; ?>
+  </ul>
+
+  <?=$this->makeTag(
+    $toggleTag,
+    $this->transEsc($this->toggleLabel) . $this->icon('dropdown-caret'),
+    $toggleAttrs,
+    ['escapeContent' => false]
+) ?>
```
demiankatz: Should this line be indented to align with 37?

demiankatz (changes requested):
Thanks for the progress, @crhallberg -- a few minor comments on the latest round.

themes/bootstrap3/templates/_ui/components/menu-button.phtml
crhallberg: I agree. `php-cs-fixer` seems to like this because it seems to mix 2 and 4 space indenting.

demiankatz: Yeah, I see that `php-cs-fixer` is still being a pain. Maybe something like this would satisfy it while being more clear?

```
  <?=$this
      ->makeTag(
        $toggleTag,
        $this->transEsc($this->toggleLabel) . $this->icon('dropdown-caret'),
        $toggleAttrs,
        ['escapeContent' => false]
      )
  ?>
```

(I haven't tried it, so maybe that won't work any better... let me know if you need me to put more time into fighting this!)

crhallberg: I'm happy with their counter to this offer:

```php
  <?=$this
      ->makeTag(
          $toggleTag,
          $this->transEsc($this->toggleLabel) . $this->icon('dropdown-caret'),
          $toggleAttrs,
          ['escapeContent' => false]
      )
  ?>
```

demiankatz (changes requested):
@crhallberg, I think this is nearly ready to merge. It looks good to me from a technical perspective, and all tests are passing. However, I think some minor style fixes are required.

In particular, in the bootprint3 theme, the menu button is out of line with other elements in the header:

image

Also, while a much lesser issue, there seems to be a minor height inconsistency in the highlighting of the menu button and the adjacent login link in the sandal theme. I can live with this if necessary, but maybe it's related to the other problem and can be easily fixed:

image

Finally, once we merge this, we'll need to make sure it doesn't need additional reconciliation with #2400.

demiankatz (changes requested):
@crhallberg, thanks for the progress on this, but the latest style changes seem to have introduced a new problem -- in the bootprint3 theme, when you hover over links, the contrast becomes nearly unreadable:

image

Can this be fixed? Do you mind testing all the header functionality in the three main themes before pushing further changes up? I think we're close, but we need to get all the details right!

demiankatz (approved):
All looking good now, and full test suite passing. Thanks for all the work on this, @crhallberg!

demiankatz:
Now the work can continue in #2492!

PR #2380, opened 2022-04-05
Add tests for VuFind\Search\Solr\Results

demiankatz:
As discussed in #2352, we currently lack a test for \VuFind\Search\Solr\Results. It would be useful to have such a test so that we can expand the coverage to address new/changed features. (This may also be useful for #2372).

I'm opening this PR to get the ball rolling. There's not a whole lot here, but this is what I was able to accomplish in 30 minutes, and it's a start! I'd suggest that it may be worthwhile to merge this even though it's incomplete, and then build upon it in the other PR's cited above.


EreMaijala (approved):
Good start for a 30-minute one! +1 for merging this now.

mtrojan-ub (approved):
Thanks, looks good!

demiankatz:
Thanks, @EreMaijala and @mtrojan-ub -- merging now!

PR #2381, opened 2022-04-06
[VUFIND-1503] Make use of ellipsis consistent and customizable in translation

demiankatz:
This makes sure that periods of ellipsis (...) are applied and translated more consistently throughout the system.

TODO
- [x] Fix loading strings
- [x] Fix more/less strings
- [x] Fix other strings
- [x] Test everything
- [x] Add changelog note about new/adjusted translations
- [x] Close [VUFIND-1503](https://vufind.org/jira/browse/VUFIND-1503) when merging


themes/bootstrap3/js/common.js
```diff
@@ -111,8 +111,8 @@ var VuFind = (function VuFind() {
   };
   var loading = function loading(text = null, extraClass = "") {
     let className = ("loading-spinner " + extraClass).trim();
-    let string = translate(text === null ? "loading" : text);
-    return '<span class="' + className + '">' + icon('spinner') + string + '...</span>';
+    let string = translate(text === null ? 'loading_ellipsis' : text);
+    return '<span class="' + className + '">' + icon('spinner') + string + '</span>';
```
demiankatz, line 115: Note that I removed the hard-coded "..." here in addition to changing the default string. The only override used in the code is `relais_loading`, which already includes ellipses -- so this is actually a bug fix, removing redundant punctuation.

demiankatz:
@crhallberg / @EreMaijala, I think we should start to think about issuing a VuFind 8.1 release in the not-too-distant future, since theme work has been pushed forward to 9.0. Thus, I'm starting to work through JIRA tickets currently flagged for 8.1. This was one of them.

My general approach:

   * For the most part, I created new language strings with an "_ellipsis" suffix in place of concatenating existing strings with "..." in templates. I generated the new language strings using the existing eol_ellipsis string when available, and "..." otherwise. This means that in some cases, the translations should be improved, and in the rest they will at least match existing behavior. (I thought it would be desirable to avoid creating a need for additional translation work).
   * In a few cases where strings are used in only one place, I simply inserted the ellipsis into the existing string rather than creating a new suffixed version. I did this for only three cases -- two service-specific recommendation module strings, and one "submitting" message that only flashes on screen for a split second. For these rarely-used/rarely-seen features, it seemed less risky/expensive to change them than to create duplicate strings.
   * The Javascript code should no longer need the non-suffixed versions of the modified strings, but I left the old versions in the js-translations.phtml to avoid backward compatibility breaks in custom code. I will remove these in 9.0 (and have put a note on the existing deprecation ticket in JIRA to remind myself about this).

Other notes:
  * I found a couple of unrelated bugs/errors while testing this -- see #2382 and #2383. I also discovered that the ga.ini translation for eol_ellipsis was wrong, and fixed it as part of this PR (with confirmation from Eoin Kilfeather).
  * I'm a bit suspicious that the eol_ellipsis translation in uk.ini is also incorrect, and I opted to use "..." instead of that value in the translations added to this file, for consistency with other existing translations of sentences ending in periods of ellipsis. Now does not feel like an appropriate time to bother our Ukrainian community members with such a trivial matter, so I'm going to trust that this is good enough for now.

In any case, I've done a fair amount of testing, and all Mink tests are now passing, but given the large number of files touched (and the fact that I didn't have time to comprehensively test every single template), I would appreciate a double-check of my work!

EreMaijala:
@demiankatz Thanks, I'll go through the changes here. One initial note: I believe the ellipsis should be at left of the strings for RTL languages.

demiankatz:
@EreMaijala, I believe that the RTL languages are correct -- I think there may be a GitHub quirk that is reversing the strings in the code view here. When I look at the files in VSCode (or actually view the interface in Arabic or Hebrew) the ellipses show up where I expect to see them... yet in the file diff here on GitHub, everything is inverted. Strange!

EreMaijala:
@demiankatz Sorry, I should have had more faith in you doing the right thing! That's weird indeed. Perhaps GitHub is trying LTR'ize the RTL text or something.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
```diff
@@ -85,7 +85,11 @@
                   echo '<strong>' . $this->transEsc($snippet['caption']) . ':</strong> ';
                 }
                 if (!empty($snippet['snippet'])) {
-                  echo '<span class="quotestart">&#8220;</span>...' . $this->highlight($snippet['snippet']) . '...<span class="quoteend">&#8221;</span><br/>';
+                  echo '<span class="quotestart">&#8220;</span>'
```
EreMaijala: Using *eol*_ellipsis at the middle of the line feels not quite right. Maybe we should have a `highlight_snippet_html` translation that contains all this at least up to the `<br>`? For instance:

`highlight_snippet_html = '<span class="quotestart">&#8220;</span>…%%snippet%%…<span class="quoteend">&#8221;</span>';`

?

EreMaijala:
@demiankatz Looks pretty good to me. The above is the only issue I found, and since the highlight snippet is repeated in a couple of templates, using a common translation for them all would make sense to me.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
EreMaijala: Note: I included the spans in the translation string because I think the quotes should be considered part of it, but if you disagree, the most important part would be `…%%snippet%%…`.

ckaz: Can't check this right now, but does it allow for language differences? In German, we use ellipsis without space before the ellipsis when we truncate a word but there must be a space before the ellipsis if we truncate a sentence or phrase, like so

ellips... 
BUT
ellipses are cool and ...

EreMaijala: Well, at least it wouldn't make the situation worse if we take the default form from eol_ellipsis, though might need some adjustment so that the preceding ellipsis has the right spacing as well, if that's desired. This is assuming all backends cut the snippet by word boundary, otherwise there's no way of doing it perfectly all the time.

demiankatz: Thanks, @EreMaijala, I agree that this is a good idea, and I'll implement it as soon as time permits. Regarding @ckaz's question, @EreMaijala's approach is clearly better than the concatenation version I chose, because as currently implemented here, it will be wrong in German -- `" ...snippet ..."` instead of `"... snippet ..."`. Creating a snippet translation gives us the control we need to do it correctly. This will currently have to be implemented to assume that the snippet ends on word boundaries on both ends, which I suppose may not always be true if different backends implement things differently... but based on the current behavior of the system, we don't have enough information to be more nuanced. Hopefully it's "good enough for now" at least. :-)

demiankatz: Okay, please take a look at [9947180](https://github.com/vufind-org/vufind/pull/2381/commits/9947180855286edf489936726fc065b1435d5375) -- this adds highlight_snippet_html as proposed by @EreMaijala, and it seems to work reasonably well for me. A couple of thoughts, though:

1.) I'm not especially comfortable with including CSS classes in the language files; I prefer to keep things as simple as possible. Given that the quotestart and quoteend classes do not appear to be used in any of our CSS, and they were presumably put there to make it possible to customize the appearance of snippets, how would you feel about removing the spans entirely and perhaps replacing the HTML entities with their Unicode equivalents? I think that might feel like a better fit for a language file, and I don't think it would reduce the ability to customize the appearance. I would still keep the _html suffix on the language string, since we need to be sure not to escape it in order to properly render the highlighting in the injected snippet.

2.) I'm a bit puzzled as to why we have snippet logic in templates related to displaying favorite lists. I don't see how there could ever be active snippets in this context. I'm wondering if this was a long-ago copy-paste error, or if there's some actual intention behind it....

EreMaijala: 1.) That'd make sense. I think that making it possible to style the whole thing easier would be more important for customization than targeting the quotes alone. This could be simply enclosing the whole snippet thing in `<div class="highlight-snippet">...</div>` (which would probably eliminate the need for a trailing `<br>`, but would need to check for preceding stuff). That's something way out of scope for this PR. At least now you can add them to the translation even though that doesn't help with the caption.

2.) I'd bet it was just copied without further thought. That's not the only odd thing, though. It doesn't make sense to me to check for a snippet caption before checking for a snippet, particularly as the caption ends with a colon.

demiankatz: Thanks, @EreMaijala!

1.) I've simplified the language strings with [7182f63](https://github.com/vufind-org/vufind/pull/2381/commits/7182f633283eea8db0144c90a39394f7c5d82f9c). Note that I used "straight" instead of "curly" quotes for the RTL languages to ensure that things don't display strangely (I noticed some inconsistent results when displaying LTR snippets within the RTL context... so using more generic quotes seemed like the easiest solution to avoid a big project). I agree that adding a wrapping `<div>` for styling could be beneficial, but I'm considering that out of scope for this PR.

2.) I'll leave the list-entry.phtml stuff alone for now since it's not hurting anything and changing it seems out of scope for this chunk of work. After merging this, I can open a follow-up PR to remove it, if you think that's worthwhile.

EreMaijala: Thanks, looks good to me now!

2.) I don't think it's worth it alone, but could be done if we decide to do other tweaks as well.

EreMaijala (approved):

demiankatz:
Thanks for the approval, @EreMaijala -- I'll wait a couple of days in case @crhallberg has anything to add, and I might give this one last look-over myself when I have fresh eyes. :-)

languages/ga.ini
```diff
@@ -380,7 +380,7 @@ Empty Book Bag = "Folmhaigh an mála leabhar"
 empty_search_disallowed = "Ní cheadaítear iarratas folamh a dhéanamh i gcás an sprioc-chuardaigh reatha"
 Enable Auto Config = "Cumasaigh an t-uathchumrú"
 End Page = "Deireadh an leathanaigh"
-eol_ellipsis = "Ö"
+eol_ellipsis = "…"
```
crhallberg, line 383: Do we need `eol_ellipsis` in the other language files?

crhallberg (approved):
Seems thorough to me! Just one question about the `eol_ellipsis` string.

languages/ga.ini
demiankatz: It should already be there in most. For those where it is currently undefined, VuFind will fall back to the en.ini version until translation volunteers provide a localized version.

demiankatz: (You're only seeing diffs related to this string in this PR because the existing translation in ga.ini was incorrect and had to be fixed).

crhallberg:
Looks good to me!

PR #2384, opened 2022-04-06
[VUFIND-1544] Eliminate use of deprecated strftime function

demiankatz:
Another discovery made while working on #2381: some of our templates were using a deprecated function. This provides non-deprecated equivalents. Since `strftime` was locale-aware and `date` is not, I had to use the `IntlDateFormatter` class in one place to provide equivalent functionality. (I'm not worried about locales in the admin screen, since that is not fully translated anyway and is rarely used).

I've hand-tested both modified templates and added test coverage to confirm an appropriate fix in the case of the impacted ILS driver.

TODO
- [x] Resolve [VUFIND-1544](https://vufind.org/jira/browse/VUFIND-1544) when merging

### DIFF
--- a/themes/bootstrap3/templates/RecordTab/reviews.phtml
+++ b/themes/bootstrap3/templates/RecordTab/reviews.phtml
@@ -13,7 +13,13 @@
           <?php if (isset($review['Rating'])): ?>
             <img src="<?=$this->imageLink($review['Rating'] . '.gif')?>" alt="<?=$review['Rating']?>/5 Stars"/>
           <?php endif; ?>
-          <strong><?=$review['Summary']?></strong> <?=isset($review['Date']) ? strftime('%B %e, %Y', strtotime($review['Date'])) : ''?>
+          <strong><?=$review['Summary']?></strong>
+          <?php
+            if (isset($review['Date'])) {
+              $formatter = new \IntlDateFormatter(null, \IntlDateFormatter::MEDIUM, \IntlDateFormatter::NONE);
+              echo $formatter->format(strtotime($review['Date']));
+            }
+          ?>
         </p>
       <?php endif; ?>
       <?php if (isset($review['Source'])): ?><strong><?=$this->transEsc('Review by')?> <?=$review['Source']?></strong><?php endif; ?>
--- a/themes/bootstrap3/templates/admin/home.phtml
+++ b/themes/bootstrap3/templates/admin/home.phtml
@@ -33,12 +33,12 @@
         <tr>
           <th><?=$this->transEsc('Start Time')?>: </th>
           <?php $startTime = $core->xpath('//lst[@name="' . $coreName . '"]/date[@name="startTime"]') ?>
-          <td><?=$this->escapeHtml(date("M d, Y g:i:sA", strtotime((string)array_pop($startTime))))?></td>
+          <td><?=$this->escapeHtml($this->dateTime()->convertToDisplayDateAndTime('Y-m-d\TH:i:s.uP', (string)array_pop($startTime)))?></td>
         </tr>
         <tr>
           <th><?=$this->transEsc('Last Modified')?>: </th>
           <?php $lastModified = $core->xpath('//lst[@name="' . $coreName . '"]/lst/date[@name="lastModified"]') ?>
-          <td><?=$this->escapeHtml(date("M d, Y g:i:sA", strtotime((string)array_pop($lastModified))))?></td>
+          <td><?=$this->escapeHtml($this->dateTime()->convertToDisplayDateAndTime('Y-m-d\TH:i:s.uP', (string)array_pop($lastModified)))?></td>
         </tr>
         <tr>
           <th><?=$this->transEsc('Uptime')?>: </th>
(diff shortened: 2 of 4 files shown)


themes/bootstrap3/templates/RecordTab/reviews.phtml
```diff
@@ -13,7 +13,13 @@
           <?php if (isset($review['Rating'])): ?>
             <img src="<?=$this->imageLink($review['Rating'] . '.gif')?>" alt="<?=$review['Rating']?>/5 Stars"/>
           <?php endif; ?>
-          <strong><?=$review['Summary']?></strong> <?=isset($review['Date']) ? strftime('%B %e, %Y', strtotime($review['Date'])) : ''?>
+          <strong><?=$review['Summary']?></strong>
+          <?php
+            if (isset($review['Date'])) {
```
EreMaijala, line 18: Why are we not using the normal display date format (from config.ini) here?

themes/bootstrap3/templates/admin/home.phtml
```diff
@@ -33,12 +33,12 @@
         <tr>
           <th><?=$this->transEsc('Start Time')?>: </th>
           <?php $startTime = $core->xpath('//lst[@name="' . $coreName . '"]/date[@name="startTime"]') ?>
-          <td><?=$this->escapeHtml(strftime("%b %d, %Y %l:%M:%S%p", strtotime((string)array_pop($startTime))))?></td>
+          <td><?=$this->escapeHtml(date("M d, Y g:i:sA", strtotime((string)array_pop($startTime))))?></td>
```
EreMaijala: Could we use the configured date/time format here as well?

EreMaijala (changes requested):

themes/bootstrap3/templates/RecordTab/reviews.phtml
demiankatz: I think the answer to both of your questions is "this code was written before we had a normal display date format," because anything using strftime is very old, coming from pre-1.0 VuFind originally.

In this particular case, I think perhaps it is justified to diverge from the standard date format, since to my mind, the date on a book review designed for long-form reading shouldn't necessarily resemble the date format for due dates -- using a spelled-out, internationalized version makes for more comfortable reading.

That being said, I'm not so strongly attached to the idea that I'm unwilling to standardize this. I'm also open to adding a method to the DateTime view helper to encapsulate the `IntlDateFormatter` logic if you think it's worth keeping but want to make it cleaner/more readable/more reusable.

EreMaijala: I think we need `longDisplayDateFormat` setting or such for this. It's quite possible that the desired format is not the same as the system locale's format. Personally I find long formats harder to understand, but you're right in that in this case it could work.

demiankatz: It's kind of tricky because you have a choice of configuring a date format which is very specific, but which may not work equally well for all locales, or using a locale-friendly date, which only offers you vague options like long/medium/short. Perhaps we should extend the VuFindDate class so that for special values of `$outputFormat`, it uses the `IntlDateFormatter` class instead of `DateTime::format()`. Then, as you say, we could have a separate config.ini setting to control long-form display dates, and everything could be made more uniform and configurable.

That being said, it's an awful lot of work (and requires some significant design decisions) in order to change something that currently affects only one rarely-seen template (now that we've dropped Amazon support, the only thing using this is Guardian reviews, and I doubt many people are turning those on). How would you feel about leaving this as-is for now but opening a JIRA ticket to capture the idea for future reference?

themes/bootstrap3/templates/admin/home.phtml
demiankatz: Adjusted in [d28b38a](https://github.com/vufind-org/vufind/pull/2384/commits/d28b38aafa01887ce3fd40bf80c8a9241b5bff4a). Not sure if there's a better way to do this; I tried to use the ISO-8601 date format constants provided by PHP, but I couldn't get them to work. I think the fractional seconds provided by Solr caused problems.

themes/bootstrap3/templates/RecordTab/reviews.phtml
EreMaijala: Right. Please add a JIRA ticket, and we'll forget about it for now. :)

themes/bootstrap3/templates/admin/home.phtml
EreMaijala: Did you try DATE_RFC3339_EXTENDED as well? Regardless, this is quite alright now.

EreMaijala (approved):

themes/bootstrap3/templates/admin/home.phtml
demiankatz: Ahh, yes, I hadn't found that one. Thanks!

themes/bootstrap3/templates/RecordTab/reviews.phtml
demiankatz: Done: [VUFIND-1545](https://vufind.org/jira/browse/VUFIND-1545).

PR #2385, opened 2022-04-07
[VUFIND-1494] Apply XSLT to RSS feed

demiankatz:
This is work in progress on applying a stylesheet to VuFind's RSS feeds.

TODO
- [x] Create mechanism for applying XSLT to RSS feed
- [x] Create XSLT for RSS feed
- [x] Determine how to expose XSLT in web-accessible way
- [x] Test
- [x] Close [VUFIND-1494](https://vufind.org/jira/browse/VUFIND-1494) when merging


demiankatz:
@EreMaijala, this PR is a response to the fact that modern browsers often lack built-in RSS readers, which results in ugly displays when people click RSS links.

Since this requires a change to the Apache configuration, perhaps it should be re-targeted to 9.0... or maybe it's worth rolling out this enhancement early, since having an assets directory seems generally useful.

I'm very open to feedback on general approach and on the specifics of my stylesheet -- I tried to keep things as language-agnostic as possible, though the atom navigation controls contain some language, so it's not perfect.

I'll work on test coverage once we've settled on approach.

themes/root/assets/xsl/rss.xsl
```diff
@@ -0,0 +1,45 @@
...
+            </xsl:if>
+            <xsl:if test="dc:date">
+              <xsl:value-of select="dc:date" /><br />
+            </xsl:if>
+          </p>
+        </xsl:for-each>
+      </body>
+    </html>
+  </xsl:template>
+</xsl:stylesheet>
```
EreMaijala: Missing line feed

module/VuFind/src/VuFind/Controller/AbstractSearch.php
```diff
@@ -364,8 +364,24 @@ protected function getSearchResultsView($setupCallback = null)
...
             $response->getHeaders()->addHeaderLine('Content-type', 'text/xml');
-            $feed = $this->getViewRenderer()->plugin('resultfeed');
-            $response->setContent($feed($view->results)->export('rss'));
+            $feedHelper = $this->getViewRenderer()->plugin('resultfeed');
+            $feed = $feedHelper($view->results);
+            $writer = new \Laminas\Feed\Writer\Renderer\Feed\Rss($feed);
+            $writer->render();
+            $themeInfo = $this->serviceLocator->get(\VuFindTheme\ThemeInfo::class);
+            $themeHits = $themeInfo->findInThemes('assets/xsl/rss.xsl');
+            if (count($themeHits) > 0) {
```
EreMaijala: Wouldn't `if ($themeHits) {` suffice?

EreMaijala (changes requested):
Makes sense to me. I think early rollout makes sense. Neglecting the Apache change wouldn't make things much worse. I think the browser could display a warning about not being able to load the style sheet, but would display the XML like before. 

I'm not seeing the navigation controls, is there something missing? For language support we could check for a language-specific xsl file (i.e. rss_fi.xsl) first and fall back to the default one if not found.

module/VuFind/src/VuFind/Controller/AbstractSearch.php
demiankatz: Fixed as part of [060f1e9](https://github.com/vufind-org/vufind/pull/2385/commits/060f1e9eb85fed9a782aa22c8e3b3a50938b8c65).

themes/root/assets/xsl/rss.xsl
demiankatz: Fixed.

EreMaijala (approved):

PR #2387, opened 2022-04-08
Add unit tests for getFacetList method of Solr results.

EreMaijala:


module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
```diff
@@ -107,6 +113,230 @@ public function testGetResultTotal(): void
...
     }
 
+    /**
+     * Test retrieving facets.
+     *
+     * @return void
+     */
+    public function testGetFacetList(): void
+    {
+        $config = $this->getMockBuilder(PluginManager::class)
```
demiankatz: Do you think it's worth creating a ConfigPluginManagerTrait that includes a more generalized version of this code for mocking out the config manager?

https://github.com/vufind-org/vufind/blob/dev/module/VuFind/tests/unit-tests/src/VuFindTest/Search/Summon/ParamsTest.php#L53-L68

I'm envisioning a method that accepts an associative array of config filename => config array, and which uses a callback to return an appropriate config object (defaulting to empty for undefined values). This seems generally useful, and it would be more targeted than accepting `$this->any()` here (not to mention making individual tests shorter and more readable).

I'd be willing to do some refactoring as a subsequent pass if you think it's worth doing but don't have the time.

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
```diff
@@ -124,6 +354,26 @@ protected function getResults(
...
+     * @return Results
+     */
+    protected function getResultsFromResponse(
+        array $response,
+        Params $params
+    ): Results {
+        $collection = new RecordCollection($response);
+        $searchService = $this->createMock(SearchService::class);
+        $searchService->expects($this->once())
+            ->method('search')
```
demiankatz, line 371: Is it worth validating the parameters sent to search using `->with()`? I imagine you could pass in an array of expectations if there's value in doing it and if you don't expect the input to be uniform across all tests.

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
```diff
@@ -107,6 +113,230 @@ public function testGetResultTotal(): void
...
+            ],
+            $this->getParams(null, $config)
+        );
+
+        // No facets configured:
+        $facets = $results->getFacetList();
+        $this->assertIsArray($facets);
+        $this->assertEmpty($facets);
+
+        // Invalid facet:
```
demiankatz: Would "Facet not found in search results" be more specific than "Invalid"? "Invalid facet" makes me think that you're going to try to set some kind of really crazy value. ;-)

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
```diff
@@ -107,6 +113,230 @@ public function testGetResultTotal(): void
...
+                            'isApplied' => false,
+                        ],
+                    ],
+                ]
+            ],
+            $facets
+        );
+
+        // Add a filter:
+        $results->getParams()->addFilter('~topic_facet:Research');
```
demiankatz, line 224: Is it worth testing the addition of other types of filters ("plain" and "NOT")? How about testing applied hierarchical filters?

demiankatz (commented):
Thanks, @EreMaijala, this is a great step forward... just a few suggestions for possible improvements (which you're welcome to ignore/defer if you like).

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
EreMaijala: That's a very good idea, now added.

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
EreMaijala: I'd say that not here, it's out of scope. E.g. `testGetResultTotal()` does its own validation, and since there are a lot of parameters, I think doing it with the helper method would become a bit messy.

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
EreMaijala: Yes, changed.

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
EreMaijala: Added a test for plain ("AND") hierarchical facet. "NOT" facets are not reflected here, so not possible to check.

demiankatz (approved):
I hope you don't mind, but I thought it best to rename `getConfigPluginManager` to `getMockConfigPluginManager` for clarity. I went ahead and did it to save you time -- I'm going to work on a follow-up PR now to apply this new trait to other existing tests.

PR #2392, opened 2022-04-11
[VUFIND-1507] Allow sub-themes to disable JS files set by parents

demiankatz:
TODO
- [x] Update documentation once approach is agreed upon
- [x] Resolve [VUFIND-1507](https://vufind.org/jira/browse/VUFIND-1507) when complete

### DIFF
--- a/module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
+++ b/module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
@@ -202,7 +202,8 @@ class ResourceContainer
 
         // If we are disabling the dependency, remove it now.
         if ($jsEntry['disabled'] ?? false) {
-            return $this->removeEntry($jsEntry, $this->js);
+            $this->removeEntry($jsEntry, $this->js);
+            return;
         }
 
         foreach ($this->js as $existingEntry) {
--- a/module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
+++ b/module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
@@ -200,9 +200,13 @@ class ResourceContainer
             );
         }
 
+        // If we are disabling the dependency, remove it now.
+        if ($jsEntry['disabled'] ?? false) {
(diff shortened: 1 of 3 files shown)


demiankatz:
@mtrojan-ub, I'm working on cleaning up outstanding tickets slated for VuFind 8.1, and this was on the list. It seems to be fairly straightforward, but please let me know if I'm overlooking anything.

mtrojan-ub:
@demiankatz, thanks for taking care of this one! It indeed seems to be pretty straightforward and match the new structure. 

An interesting scenario might be => what happens if you disable something which is referenced later via load_after? I guess that, since it won't exist at that point anymore, there will be a "Dependency not found" error. So this seems to be correct as well as consistent to me.

demiankatz:
Thanks, @mtrojan-ub. I have updated theme.config.php to include documentation of the new setting.

Do you think it is also worth adding a "replace" setting which, if found, deletes any previous entry for a file and re-adds it? Seems like it would be very easy to implement using the mechanisms we already have in place. (I realize that we could potentially just replace things without requiring a setting, but to prevent accidental overrides, I think requiring a configuration might be a useful safety mechanism).

In any case, I'm going to be away for the next few days, and @EreMaijala is already on vacation. When the two of us are back, I'll be interested to hear his thoughts and make further progress.

mtrojan-ub:
I think it's more of a nice-to-have scenario. But if you see a benefit in doing it => i guess the most important aspect will be to safely replace an entry which other entries already refer to via `load_after`.

Have a nice vacation!

demiankatz:
Thanks, @mtrojan-ub, you're right that replacements could get a bit complicated. I'm sure we could deal with it by generating a dependency tree, etc., but that doesn't seem worth the effort before we have a real use case. I'm happy to leave this as it is for now and deal with that later.

EreMaijala (approved):
Looks good and useful to me. I agree with @mtrojan-ub that in the perfect world we could allow something to be automatically replaced, but since this is pretty advanced already, I don't think it's unreasonable to require some manual work here.

PR #2388, opened 2022-04-08
[VUFIND-1543] Add missing labels to sliders.

demiankatz:
This PR addresses the problem reported in [VUFIND-1543](https://vufind.org/jira/browse/VUFIND-1543): incomplete aria-label coverage on slider controls. The solution is derived from work that @ckaz shared on the ticket, updated to work with the latest version of the template.

Some notes on changes here:

  * Since it is possible for multiple sliders to be displayed on a single page, I changed some hard-coded IDs in @ckaz's version into dynamic ones, and I made some selectors more specific.
  * I had to pass an extra variable down to the slider template in order to label the slider appropriately; I did some minor cosmetic adjustment of a higher-level template to make the flow more readable while I was in the area.
  * I found the large number of calls to `$this->escapeHtmlAttr($title)` very difficult to read, so I created a `$safeBaseId` variable to simplify things.
  * I added a new English translation string, but I'm open to suggestions on better wording.
  * In @ckaz's version of the template, he changed the top-level `<div>` to an `<li>`. I did not include that change here, as I wanted to focus on fixing labels; I believe that we should address overall facet structure separately.

TODO
- [x] Resolve [VUFIND-1543](https://vufind.org/jira/browse/VUFIND-1543) when merging

PR #2393, opened 2022-04-12
Improve advanced search accessibility.

demiankatz:
This removes a duplicate label and add labels to the slider control (in the same way as #2388 improved SideFacets).

### DIFF
--- a/themes/bootstrap3/templates/search/advanced/layout.phtml
+++ b/themes/bootstrap3/templates/search/advanced/layout.phtml
@@ -103,7 +103,7 @@
             <?php endif; ?>
             <div id="group<?=$group ?>" class="adv-group">
               <div class="adv-group-terms">
-                <label class="adv-group-label" for="search_lookfor<?=$group ?>_0"><?=$this->transEsc("adv_search_label")?>:</label>
+                <label class="adv-group-label" <?=$this->transEsc("adv_search_label")?>:</label>
                 <?php for($search = 0; $search < 3 || (isset($setQueries[$group]) && $search < count($setQueries[$group])); $search++): ?>
                   <?php if($group == 0 && $search == 0): ?>
                     <div id="new_search_template">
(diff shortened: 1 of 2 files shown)


themes/bootstrap3/templates/search/advanced/layout.phtml
```diff
@@ -103,7 +103,7 @@
             <?php endif; ?>
             <div id="group<?=$group ?>" class="adv-group">
               <div class="adv-group-terms">
-                <label class="adv-group-label" for="search_lookfor<?=$group ?>_0"><?=$this->transEsc("adv_search_label")?>:</label>
+                <label class="adv-group-label" <?=$this->transEsc("adv_search_label")?>:</label>
```
demiankatz: search_lookfor_n_0 already has an aria-label directly on it, so associating this label as well causes confusion.

crhallberg: The `for` attribute allows users to click on the label to focus the input. It shouldn't interfere with aria or other screen reader features.

crhallberg: Also, an extra `>` was trimmed here.

crhallberg (changes requested):

themes/bootstrap3/templates/search/advanced/layout.phtml
demiankatz: Thanks for catching the missing `>` -- I have restored it. Regarding the `for` attribute, WAVE reports an error because some elements have multiple labels associated with them. I understand the focus behavior may be useful, but it does seem problematic to me to have two different labels for the same thing -- and this label doesn't really describe the first search box but rather the entire region... so I wonder if we've been abusing the semantics of the tag to get the functionality we desire. Maybe there's a better way?

crhallberg: I think the issue is actually that the template also contains a `for` to the first element of the first group (as a placeholder). This is certainly a solution for bootstrap3. Perhaps a refactor of this template in the new theme can help restore this convenience.

crhallberg (approved):

ckaz (approved):
Looks very good to me!

ckaz (approved):
Looks very good to me!

PR #2394, opened 2022-04-19
Improvements to facet/filter processing logic.

demiankatz:
This is a follow-up to #2352 that renames and slightly revises a method for increased clarity and fixes a problem where filters were not processed the same way as facets. Some of this work is extracted from #2372 -- I think we should get this change merged to the dev branch BEFORE we release VuFind 8.1 so that #2352 doesn't introduce an additional unnecessary BC break between 8.1 and 9.0.

TODO
- [x] Run full test suite

### DIFF
--- a/module/VuFind/src/VuFind/Search/Base/Params.php
+++ b/module/VuFind/src/VuFind/Search/Base/Params.php
@@ -1071,7 +1071,7 @@ class Params
      *
      * @return string
      */
-    public function getFacetValueDisplayText(string $field, string $value): string
+    public function getFacetValueRawDisplayText(string $field, string $value): string
     {
         // Check for delimited facets -- if $field is a delimited facet field,
         // process $displayText accordingly:
@@ -1096,7 +1096,7 @@ class Params
      */
     protected function formatFilterListEntry($field, $value, $operator, $translate)
     {
-        $displayText = $this->getFacetValueDisplayText($field, $value);
+        $displayText = $this->getFacetValueRawDisplayText($field, $value);
 
         if ($translate) {
             $domain = $this->getOptions()->getTextDomainForTranslatedFacet($field);
--- a/module/VuFind/src/VuFind/Search/Base/Params.php
+++ b/module/VuFind/src/VuFind/Search/Base/Params.php
@@ -1064,23 +1064,24 @@ class Params
     }
 
     /**
-     * Check for delimited facets -- if $field is a delimited facet field,
-     * process $displayText accordingly. Return the appropriate display value.
+     * Get a display text for a facet field.
      *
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/Search/Base/Params.php
```diff
@@ -1064,23 +1064,24 @@ public function getFilterList($excludeCheckboxFilters = false)
...
      *
-     * @param string $field       The facet
-     * @param string $displayText The facet value
+     * @param string $field Facet field
+     * @param string $value Facet value
      *
      * @return string
      */
-    public function checkForDelimitedFacetDisplayText($field, $displayText)
+    public function getFacetValueDisplayText(string $field, string $value): string
```
demiankatz: @EreMaijala, since the return value from this method may be subject to additional post-processing (for hierarchical and/or translated facets), should we rename this method to something like `getFacetValueRawDisplayText`?  Or should we pass the `$hierarchical` and `$translated` flags to this method and do additional processing in here to better encapsulate the logic?

demiankatz (commented):
Now that I've pulled the logic into its own PR and taken a closer look at it, I wonder if it needs a little further adjustment....

module/VuFind/src/VuFind/Search/Base/Params.php
EreMaijala: i like `getFacetValueRawDisplayText` better than encapsulating everything in it. The contexts where it's called are a bit different, so the additional processing is also different (e.g. whether to display all hierarchy levels).

demiankatz: Okay, method renamed. I also noticed that formatting is only applied to facets, not to filters; this feels to me like an oversight rather than a decision, so I've created another support method for facet translation (see d0632fb) in order to make it easier to be consistent. This also allowed me to significantly compact some of the code. However, this is breaking a test, and I'm currently having trouble figuring out why... Investigating now.

demiankatz: Broken test was related to a typehint prematurely converting a TranslatableString object into a string. Fixed in 71e9566, and thanks to @EreMaijala for pointing me in the right direction!

demiankatz:
I believe that this is now ready to merge, but I would like a review from @mtrojan-ub before wrapping it up in case he has any different opinions.

EreMaijala (approved):

mtrojan-ub (approved):
In the end it only looks like a minor change so i don't mind, thanks!

demiankatz:
Thanks, @EreMaijala and @mtrojan-ub -- merging now!

PR #2395, opened 2022-04-20
Fix removeAllFilters to remove all types of filters.

EreMaijala:
Previously only the same type of filter (AND, OR or NOT based on the field prefix) was removed.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Base/Params.php
+++ b/module/VuFind/src/VuFind/Search/Base/Params.php
@@ -908,8 +908,13 @@ class Params
     {
         if ($field == null) {
             $this->filterList = [];
-        } elseif (isset($this->filterList[$field])) {
-            unset($this->filterList[$field]);
+        } else {
+            $field = ltrim($field, '-~');
+            foreach (['', '-', '~'] as $prefix) {
+                if (isset($this->filterList[$prefix . $field])) {
+                    unset($this->filterList[$prefix . $field]);
+                }
+            }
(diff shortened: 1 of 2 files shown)


EreMaijala:
Mink tests are passing.

module/VuFind/src/VuFind/Search/Base/Params.php
```diff
@@ -908,8 +908,13 @@ public function removeAllFilters($field = null)
     {
         if ($field == null) {
             $this->filterList = [];
-        } elseif (isset($this->filterList[$field])) {
-            unset($this->filterList[$field]);
+        } else {
+            $field = ltrim($field, '-~');
```
demiankatz: Do we really need/want to do this stripping? Is there any reason a real user would submit a prefixed field? And if for some reason they did, would they really want the prefix stripped? If we're concerned about backward compatibility, not stripping the field would actually maintain existing behavior... and if we're not concerned about backward compatibility, I don't think we want to encourage people to send in prefixed fields. Am I overlooking a strong reason for this?

demiankatz (commented):
Looks good to me -- just one question.

module/VuFind/src/VuFind/Search/Base/Params.php
EreMaijala: No, I just tried.. something. Trimming now removed.

demiankatz (approved):

PR #2397, opened 2022-04-20
Move hierarchical facet options to the base Options class.

EreMaijala:
From #2372.

### DIFF
--- a/module/VuFind/src/VuFind/AjaxHandler/GetSideFacets.php
+++ b/module/VuFind/src/VuFind/AjaxHandler/GetSideFacets.php
@@ -238,8 +238,6 @@ class GetSideFacets extends \VuFind\AjaxHandler\AbstractBase
         Results $results
     ) {
         $response = [];
-        $hierarchicalFacets = [];
-        $hierarchicalFacetSortOptions = [];
         $options = $results->getOptions();
         $hierarchicalFacets = $options->getHierarchicalFacets();
         $hierarchicalFacetSortOptions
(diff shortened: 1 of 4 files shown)


module/VuFind/src/VuFind/AjaxHandler/GetSideFacets.php
```diff
@@ -241,11 +241,9 @@ protected function formatFacets(
         $hierarchicalFacets = [];
         $hierarchicalFacetSortOptions = [];
         $options = $results->getOptions();
-        if (is_callable([$options, 'getHierarchicalFacets'])) {
-            $hierarchicalFacets = $options->getHierarchicalFacets();
-            $hierarchicalFacetSortOptions
-                = $recommend->getHierarchicalFacetSortOptions();
-        }
+        $hierarchicalFacets = $options->getHierarchicalFacets();
```
demiankatz, line 242: You can now eliminate the empty array initializations of these variables above.

demiankatz (changes requested):
One small suggestion...

EreMaijala:
I decided to revert the changes to the constructors. Since almost everything else is also done only in the classes that actually use the functionality, it seems more clear to keep hierarchical facets that way as well (or alternatively move a lot of other settings init also to the base class' constructor).

module/VuFind/src/VuFind/AjaxHandler/GetSideFacets.php
EreMaijala: Oops, done.

demiankatz (approved):

PR #2399, opened 2022-04-20
Add ability to process feedback forms in more ways

xmorave2:
This refactors processing the feedback forms from hardcoded emailing into something a bit  more flexible and configurable.

The email functionality is refactored out of the feedback controller into own plugin class managed by plugin manager. There is also new option to save the form into database.

It is working but i am still feeling it would need some more attention. I am not sure about error handling yet and also the db structure could be more flexible I think... But I would like to have same feedback from the community, so I am sharing it now.

TODO:
- [x] Update requirements to MySQL 5.7.8, MariaDB 10.2.7 respectively due to missing JSON data type support in earlier version (and reflect in changelog; note that earlier versions of these products are at/past end of life, so this should be a safe increase).
- [x] Add wiki page for new plugin type
- [x] Update db schema for feedback table
- [x] Better error handling
- [x] Basic admin page for managing forms saved in db
- [x] Test with PostgreSQL
- [x] Test that PostgreSQL migration works correctly


demiankatz (commented):
This makes sense to me! I'm definitely interested to hear what @EreMaijala thinks, since I think he has more investment in the feedback forms than I do.

EreMaijala:
@xmorave2 Thanks for doing this! We've had some hacks applied locally for database storage and sending to an external API, and this will make things much smoother (and relieves me from doing the hard work of refactoring this!).

I haven't done a thorough review, but I think there are a couple of things to consider regarding the database structure:

1.) Timestamp for creation date and time!

2.) I think it would be useful to have a JSON field or such for additional information so that if we add any new fields similar to e.g. user agent, the database structure doesn't need to be changed to accommodate that. 

3.) In anticipation of some kind of handling of the feedback messages, it would be useful to have a status field (e.g. 'open', 'in progress', 'pending', 'answered', 'closed') alongside an update timestamp.

4.) For multi-instance sites the VuFind instance URL would be important.

5.) Form name should be included so that different kinds of feedback messages can be handled separately.

For reference, see [our current schema](https://github.com/NatLibFi/NDL-VuFind2/blob/dev/module/Finna/sql/mysql.sql#L143). It's far from perfect, but includes the above suggestions.

xmorave2:
@EreMaijala Thanks for your suggestions. I am going to incorporate them. We are going to use external API for one specific form too, so that was my motivation of doing this ;)

EreMaijala:
@xmorave2 We've been through several iterations of external API support, and this could be useful for inspiration:
https://www-kiwi-fi.translate.goog/pages/viewpage.action?pageId=235930674&_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=fi&_x_tr_pto=wapp

The [original page](https://www.kiwi.fi/pages/viewpage.action?pageId=235930674) is only available in Finnish, so the above puts it through Google translation. Sorry about that and roughness of the translation.

EreMaijala:
@xmorave2 Regarding error handling, I think things get complicated if a form can have multiple handlers. A partial failure in such case is difficult to handle sensibly. If it would be possible to manage with a single handler, it would be more straightforward. In that case I find these the most important: 

1.) If there's a failure, the form contents must not disappear. So return back to the form with an error message.

2.) Log the actual issue with enough detail, but show something pretty generic to the user.

xmorave2:
I made some progress on this and I believe it is ready for another round of review/testing.
- I changed the db table structure as @EreMaijala suggested
- Added basic feedback management into admin module
- Added logging to error handling - now user see only general error messages and more detailed messages are logged

xmorave2:
I made some progress on this and I believe it is ready for another round of review/testing.
- I changed the db table structure as @EreMaijala suggested
- Added basic feedback management into admin module
- Added logging to error handling - now user see only general error messages and more detailed messages are logged

xmorave2:
I made some progress on this and I believe it is ready for another round of review/testing.
- I changed the db table structure as @EreMaijala suggested
- Added basic feedback management into admin module
- Added logging to error handling - now user see only general error messages and more detailed messages are logged

xmorave2:
I made some progress on this and I believe it is ready for another round of review/testing.
- I changed the db table structure as @EreMaijala suggested
- Added basic feedback management into admin module
- Added logging to error handling - now user see only general error messages and more detailed messages are logged

demiankatz:
Thanks for moving this forward, @xmorave2! I'm a bit short on time this week, so I'm going to see if @EreMaijala has an opportunity to offer the first round of feedback (since I think he is more directly invested in this feature than I am). However, if my input is needed, just let me know and I'll take a closer look as soon as I can (which probably will be sometime next week).

EreMaijala:
@xmorave2 I mentioned my concern about having multiple handlers for a form. Is there a strong use case for that? I think that one complicates things quite a bit. It also makes it next to impossible for the user to judge whether the form was saved or sent successfully when "something" fails.
 
I understand the need to get email when a form is received even when it's stored to the database, but I'd consider that a secondary task that the user doesn't need to know anything about. I think it would be better to allow an email notification about new feedback, but scrub any details from the message. Also a cron job for nagging about unhandled feedback, that would also make it possible to remind about issues in progress (and would do the job eventually if the instant notification failed). These changes would make it possible to receive feedback including confidential information that must not transmitted via email, while still receiving notifications by email.

xmorave2:
@EreMaijala The use case is not vary stable API of external service we need to implement adding feedback into. So my hope was we can use multiple handlers to overcame it and have the feedback saved somewhere. Because it is nice that failed feedback return user back to filled form, but if the external API fails, there is not much to do with it from user's site...

But we could easily add more detailed log in our custom handler a have some monitoring on it to be sure things are going well.

I could then simplify thing here a bit a make it use only one handler...

Maybe it is good idea to implement it in the manner of "primaryHandler" and "secondaryHandler" - primary fail would be then considered as problem and presented to user with error. Secondary handler would be used only as "notification" and user wouldn't know it exists.

I don't think there are many use cases for more than two handlers - I don't think anybody want to check many systems for feedback ;)

What do you think?

demiankatz:
@xmorave2, this sounds fairly reasonable to me -- but in the interest of flexibility, maybe it's worth considering implementing secondaryHandler as an array, so there's one handler that's mandatory and will cause errors if it fails, but you can potentially send results to more than one backup system. As you say, it seems most likely that no one will want more than two, but if somebody is really worried about fail-safes, I could imagine a scenario where primary is an API, and secondary is both a database and a flat file on disk, with the flat file being the emergency fallback in case all else fails.

xmorave2:
I refactor it in way @demiankatz was suggesting

config/vufind/FeedbackForms.yaml
```diff
@@ -146,6 +156,9 @@ forms:
     #recipient:
     #  name: Your Library
     #  email: feedback@myuniversity.edu
+    handler:
```
EreMaijala: This is no longer valid, right?

languages/en.ini
```diff
@@ -275,6 +278,9 @@ copy_to_clipboard_success_message = "Successfully copied to clipboard"
 Copyright = "Copyright"
 Corporate Author = "Corporate Author"
 Corporate Authors = "Corporate Authors"
+Could not save your Feedback = "Could not save your Feedback"
```
EreMaijala: Lower case "feedback"?

module/VuFind/src/VuFind/Form/Handler/Email.php
```diff
@@ -0,0 +1,209 @@
...
+            $sendSuccess = $sendSuccess && $success;
+            if (!$success) {
+                $errorsDetailed[]  = $errorMsg;
+            }
+        }
+        return [
+            'success' => $success,
+            'errorMessagesDetailed' => $errorsDetailed,
+            'errorMessages'
+                => $success ? [] : ['Could not send e-mail with your feedback.'],
```
EreMaijala: I don't think the user should know what failed and how. It's better to log the details and give the user just a generic "Could not process your feedback. Please try again later." or such.

module/VuFind/src/VuFind/Controller/FeedbackController.php
```diff
@@ -96,49 +98,42 @@ public function formAction()
...
+            $view->setVariable(
+                'successMessage',
+                $result['successMessage'] ?? $form->getSubmitResponse()
             );
+            $view->setTemplate('feedback/response');
+        }
+        foreach ($result['errorMessages'] ?? [] as $error) {
+            $this->flashMessenger()->addErrorMessage($this->translate($error));
+        }
+        foreach ($result['errorMessagesDetailed'] ?? [] as $error) {
```
EreMaijala: Could logging of detailed errors be left to the handler? That would also eliminate the logging loop for the secondary handlers.

module/VuFind/src/VuFind/Controller/FeedbackController.php
```diff
@@ -96,49 +98,42 @@ public function formAction()
...
+        $result = $primaryHandler->handle($form, $params, $user ?: null);
+        $success = $result['success'] ?? false;
+        if ($success) {
+            $view->setVariable(
+                'successMessage',
+                $result['successMessage'] ?? $form->getSubmitResponse()
             );
+            $view->setTemplate('feedback/response');
+        }
+        foreach ($result['errorMessages'] ?? [] as $error) {
```
EreMaijala: I suppose we shouldn't have success and error messages at the same time, so reflect that in code?

EreMaijala (changes requested):

config/vufind/FeedbackForms.yaml
xmorave2: You are right, I've fixed it.

languages/en.ini
xmorave2: Yes, fixed.

module/VuFind/src/VuFind/Form/Handler/Email.php
xmorave2: I've generalized the error message and refactor error handling to make code a bit simpler

xmorave2:
@EreMaijala Thanks for review, I tried to address your comment - especially the error handling, as you suggested

languages/en.ini
```diff
@@ -275,6 +278,8 @@ copy_to_clipboard_success_message = "Successfully copied to clipboard"
 Copyright = "Copyright"
 Corporate Author = "Corporate Author"
 Corporate Authors = "Corporate Authors"
+Could not save your Feedback = "Could not save your feedback"
```
EreMaijala: This is no longer used.

languages/cs.ini
```diff
@@ -273,6 +276,8 @@ copy_to_clipboard_success_message = "Kopírování bylo úspěšné"
 Copyright = "Copyright"
 Corporate Author = "Korporativní autor"
 Corporate Authors = "Korporace"
+Could not save your Feedback = "Vaši zprávu se nepodařilo uložit"
```
EreMaijala: This is no longer used.

themes/bootstrap3/templates/feedback/response.phtml
```diff
@@ -12,7 +12,9 @@
   <<?=$headTag?>><?=$this->escapeHtml($title)?></<?=$headTag?>>
 <?php endif; ?>
 <?=$this->flashmessages()?>
-<div role="status" class="alert alert-success">
-  <?=$this->translate($form->getSubmitResponse()) ?>
-</div>
+<?php if (!empty($this->successMessage)): ?>
```
EreMaijala: We should always have a success message if we get this far, no? Otherwise the user could end up with an empty screen. On error should be directed back to the form.

EreMaijala (approved):
Looking good! This is a approval on the condition that the remaining minor issues are fixed. However, I've not tested this properly, particularly the Admin module part, so perhaps @demiankatz can do the final review.

themes/bootstrap3/templates/feedback/response.phtml
xmorave2: You are right, I've removed the `if`. Thanks.

demiankatz:
Thanks, @EreMaijala. I'm out of time for this week but will try to find some time for review/testing next week!

languages/en.ini
```diff
@@ -229,6 +230,7 @@ citation_volume_abbrev = "Vol."
 Cite this = "Cite this"
 City = "City"
 Clear = "Clear"
+clear_feedback_filter = "Clear filter"
```
demiankatz: Is there a strong reason why this text is "Clear filter" instead of "Clear Filter"? If we don't mind using the capitalized version, we could probably use the language console tools to copy the existing clear_tag_filter translation to clear_feedback_filter, and then we won't have to re-translate that string into all the various languages. I can run this command and commit the changes if you like!

module/VuFind/sql/migrations/pgsql/9.0/002-add-feedback-table.sql
```diff
@@ -0,0 +1,21 @@
...
+message text,
+form_data json DEFAULT '',
+form_name varchar(255) NOT NULL,
+created timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+updated timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+updated_by int DEFAULT NULL,
+status feedback_status NOT NULL DEFAULT 'open',
+site_url varchar(255) NOT NULL,
+PRIMARY KEY (id),
+KEY user_id (user_id),
```
demiankatz: Is there a reason this key exists in the migration but not in the main pgsql.sql file?

module/VuFind/sql/mysql.sql
```diff
@@ -346,3 +346,29 @@ CREATE TABLE `auth_hash` (
...
+CREATE TABLE `feedback` (
+  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+  `user_id` int(11) DEFAULT NULL,
+  `message` longtext,
+  `form_data` json DEFAULT NULL,
+  `form_name` varchar(255) NOT NULL,
+  `created` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `updated` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `updated_by` int(11) DEFAULT NULL,
+  `status` enum('open','in progress','pending','answered','closed') NOT NULL DEFAULT 'open',
```
demiankatz: Have you tested whether VuFind's database upgrade tool can handle enums correctly? I'm not sure if we've encountered this case before. (I haven't taken the time to try it yet -- my guess is that it will probably work when the table is added for the first time, but it might be problematic if we change the table in the future; especially if we change the values in the enum).

module/VuFind/src/VuFind/Controller/FeedbackController.php
```diff
@@ -96,49 +98,20 @@ public function formAction()
...
-                $this->flashMessenger()->addErrorMessage($errorMsg);
-            }
         }
 
-        if ($sendSuccess) {
-            $view->setTemplate('feedback/response');
-            $view->emailMessage = $emailMessage;
+        $handlers = $form->getSecondaryHandlers();
+        foreach ($handlers as $handler) {
+            $handler->handle($form, $params, $user ?: null);
```
demiankatz: Might we want to put the secondary handling in a try..catch so that a fatal error in a secondary handler doesn't break the whole process?

module/VuFind/src/VuFind/Db/Table/Feedback.php
```diff
@@ -0,0 +1,156 @@
...
+        if (null !== $formName) {
+            $select->where->equalTo('form_name', $formName);
+        }
+        if (null !== $siteUrl) {
+            $select->where->equalTo('site_url', $siteUrl);
+        }
+        if (null !== $status) {
+            $select->where->equalTo('status', $status);
+        }
+        $select->join(['u' => 'user'], 'u.id = feedback.user_id', [], 'left')
```
demiankatz: Might be better to use the constant $select::JOIN_LEFT instead of the hard-coded string 'left' ... though it's not a huge matter of importance since hopefully this will all get replaced by Doctrine in the fairly near future.

module/VuFind/src/VuFind/Db/Table/Feedback.php
```diff
@@ -0,0 +1,156 @@
...
+        };
+        return $this->delete($callback);
+    }
+
+    /**
+     * Get available feedback statuses
+     *
+     * @return array
+     */
+    public function getStatuses(): array
```
demiankatz: I wonder if there is a better place to put this; when database code gets refactored, I'm not sure where this will end up. Might it be better as a support method in the controller for now?

module/VuFind/src/VuFind/Form/Form.php
```diff
@@ -1089,4 +1103,40 @@ protected function getElementId(string $id): string
...
+
+    /**
+     * Get secondary form handlers
+     *
+     * @return HandlerInterface[]
+     */
+    public function getSecondaryHandlers(): array
+    {
+        $handlerNames = (array)($this->formConfig['secondaryHandlers'] ?? []);
+        $handlers = [];
```
demiankatz: Could simplify to: `return array_map([$this->handlerManager, 'get'], $handlerNames);`

module/VuFind/src/VuFind/Form/Form.php
```diff
@@ -1089,4 +1103,40 @@ protected function getElementId(string $id): string
...
+        }
+        return $handlers;
+    }
+
+    /**
+     * Get current form id/name
+     *
+     * @return string
+     */
+    public function getFormId(): string
```
demiankatz, line 1134: Would it make sense to use this internally for other references to `$this->formConfig['id']` (e.g. in `getElementId` above)?

module/VuFind/src/VuFind/Form/Handler/HandlerInterface.php
```diff
@@ -0,0 +1,57 @@
...
+ * @category VuFind
+ * @package  Form
+ * @author   Josef Moravec <moravec@mzk.cz>
+ * @license  https://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+interface HandlerInterface
+{
+    /**
+     * Gets data from submitted form and process them.
```
demiankatz: Minor grammar issue: should be either "Gets ... and processes" or "Get ... and process" so that the clauses match. Whichever option you choose should be applied here and also to the two implementing classes.

themes/bootstrap3/templates/admin/home.phtml
```diff
@@ -38,7 +38,8 @@
         <tr>
           <th><?=$this->transEsc('Last Modified')?>: </th>
           <?php $lastModified = $core->xpath('//lst[@name="' . $coreName . '"]/lst/date[@name="lastModified"]') ?>
-          <td><?=$this->escapeHtml($this->dateTime()->convertToDisplayDateAndTime(DATE_RFC3339_EXTENDED, (string)array_pop($lastModified)))?></td>
+          <td><?php //=$this->escapeHtml($this->dateTime()->convertToDisplayDateAndTime(DATE_RFC3339_EXTENDED, (string)array_pop($lastModified)))?></td>
```
demiankatz: Is this meant to be commented out? Is this a fix to an unrelated problem that we should merge separately?

demiankatz (changes requested):
Thanks for this -- I've just done a fairly close review of everything except the admin module pieces (I ran out of time). Hopefully this will help with the next round of incremental improvements, and then I can test and look at admin code next week.

languages/en.ini
xmorave2: No reason, I've changed it copy the translation to all language files

module/VuFind/sql/migrations/pgsql/9.0/002-add-feedback-table.sql
xmorave2: Good catch, there is no reason, I just missed it somehow...

module/VuFind/src/VuFind/Controller/FeedbackController.php
xmorave2: Done

module/VuFind/src/VuFind/Db/Table/Feedback.php
xmorave2: You are right, using constant is better, I've changed it.

module/VuFind/src/VuFind/Db/Table/Feedback.php
xmorave2: I think this information is part of model from my point of view, but I have moved it to not complicate refactoring to Doctrine and we can rethink it when refactoring is done.

module/VuFind/src/VuFind/Form/Form.php
xmorave2: Thanks, I've changed it

module/VuFind/src/VuFind/Form/Form.php
xmorave2: Does make sense, done.

module/VuFind/src/VuFind/Form/Handler/HandlerInterface.php
xmorave2: Fixed now

themes/bootstrap3/templates/admin/home.phtml
xmorave2: This is unrelated to this PR, so I removed it. It was temporary fix, because my index didn't reported last modified date, so admin home page crashed. I don't know if it should be fixed. But probably could be fixed easily by adding a check to date existence.

module/VuFind/src/VuFind/Db/Table/Feedback.php
xmorave2: I think this information is part of model from my point of view, but I have moved it to not complicate refactoring to Doctrine and we can rethink it when refactoring is done.

module/VuFind/src/VuFind/Db/Table/Feedback.php
xmorave2: You are right, using constant is better, I've changed it.

themes/bootstrap3/templates/admin/home.phtml
xmorave2: I added fix for such a situation here: https://github.com/vufind-org/vufind/pull/2485

module/VuFind/sql/mysql.sql
xmorave2: I tested it now and it looks like it is correctly handled. Adding new table was no problem, then I modified the enum values - add some and remove some and it worked well.

languages/en.ini
```diff
@@ -248,6 +250,7 @@ Conference Proceeding = "Conference Proceeding"
 Configuration = "Configuration"
 confirm_delete = "Are you sure you want to delete this?"
 confirm_delete_brief = "Delete Item?"
+confirm_delete_feedback = "Delete Feedback"
```
demiankatz, line 253: Should this have a question mark on it?

module/VuFind/sql/mysql.sql
demiankatz: Thanks for taking the time to confirm!

module/VuFind/src/VuFind/Db/Table/Feedback.php
demiankatz: Thank you, and I agree -- we can improve this after the refactoring!

module/VuFindAdmin/src/VuFindAdmin/Controller/FeedbackController.php
```diff
@@ -0,0 +1,306 @@
...
+    protected $params;
+
+    /**
+     * Get the url parameters
+     *
+     * @param string $param A key to check the url params for
+     *
+     * @return string
+     */
+    protected function getParam($param)
```
demiankatz: I realize that this controller is derived from TagsController and that's why this design is being used... but TagsController is a bit confusing, and maybe we don't want to duplicate all of its flaws. It's confusing to me that we assign values to $this->params in each of the actions just so that we control whether GET or POST is prioritized... and that we do redundant parameter lookups regardless. What if we eliminated the $params property entirely and changed getParam to something like this?

```
protected function getParam($param, $prioritizePost = false)
{
    $primary = $prioritizePost ? 'fromPost' : 'fromQuery';
    $secondary = $prioritizePost ? 'fromQuery' : 'fromPost';
    return $this->params()->$primary($param) ?? $this->params()->$secondary($param);
}
```

Then we could modify `deleteAction()` / `updateStatusAction()` to pass `true` as the second parameter in all the `getParam` calls, and all the assignments to $this->params could be eliminated.

demiankatz: If you agree with this approach, maybe I should open a separate PR to make the same changes in TagsController for consistency.

module/VuFindAdmin/src/VuFindAdmin/Controller/FeedbackController.php
```diff
@@ -0,0 +1,306 @@
...
+        $this->params = $this->params()->fromQuery();
+        $view = $this->createViewModel();
+        $view->setTemplate('admin/feedback/home');
+        $feedbackTable = $this->getFeedbackTable();
+        $feedback = $feedbackTable->getFeedbackByFilter(
+            $this->convertFilter($this->getParam('form_name')),
+            $this->convertFilter($this->getParam('site_url')),
+            $this->convertFilter($this->getParam('status'))
+        );
+        $view->setVariables(
```
demiankatz: What if you just call createViewModel here with this array, instead of creating the view earlier and then setting the variables separately? You would just have to move down the setTemplate call above, and then the code is simplified by a line.

themes/bootstrap3/templates/admin/feedback/home.phtml
```diff
@@ -0,0 +1,153 @@
...
+          <?php endforeach;?>
+        </select>
+      </label>
+      <label for="status">
+        <?=$this->translate('Status')?>
+        <select name="status" id="status" class="form-control">
+          <option value="ALL"><?=$this->translate('All')?></option>
+          <?php foreach ($this->statuses as $status): ?>
+            <option value="<?=$this->escapeHtmlAttr($status)?>"<?php if ($status == ($this->params['status'] ?? null)): ?> selected="selected"<?php endif;?>>
+              <?=$this->transEsc('feedback_status_' . $status)?>
```
demiankatz: Maybe it would make sense to use `<?=$this->transEsc('feedback_status_' . $status, [], $status)?>` to default to the raw status value if the translation string is missing.

themes/bootstrap3/templates/admin/feedback/home.phtml
```diff
@@ -0,0 +1,153 @@
...
+                      </div>
+                    </div>
+                  </div>
+                <?php endif; ?>
+            </td>
+            <td>
+              <select class="form-control status_update" name="status_update">
+                <?php foreach ($this->statuses as $status): ?>
+                  <option value="<?=$this->escapeHtmlAttr($status)?>"<?=$status == $feedbackItem->status ? ' selected="selected"' : ''?>>
+                    <?=$this->transEsc('feedback_status_' . $status)?>
```
demiankatz: Same comment as above (but feel free to ignore if not helpful).

themes/bootstrap3/templates/admin/feedback/home.phtml
```diff
@@ -0,0 +1,153 @@
...
+            $data = json_decode($feedbackItem->form_data, true);
+          ?>
+          <tr>
+            <td>
+              <input id="checkbox_<?=$this->escapeHtmlAttr($feedbackItem->id)?>" type="checkbox" name="ids[]" value="<?=$this->escapeHtmlAttr($feedbackItem->id)?>" class="checkbox_ui"/>
+              <input type="hidden" name="idsAll[]" value="<?=$this->escapeHtmlAttr($feedbackItem->id)?>" />
+              <?=$this->escapeHtml($feedbackItem->form_name)?>
+            </td>
+            <td><?=$this->escapeHtml($feedbackItem->site_url)?></td>
+            <td><?=$this->escapeHtml($feedbackItem->message)?></td>
```
demiankatz: This value can potentially be very long. Would it make sense to limit it in some way to keep the size of the overall table under control?

themes/bootstrap3/templates/admin/feedback/home.phtml
```diff
@@ -0,0 +1,153 @@
...
+                  <option value="<?=$this->escapeHtmlAttr($status)?>"<?=$status == $feedbackItem->status ? ' selected="selected"' : ''?>>
+                    <?=$this->transEsc('feedback_status_' . $status)?>
+                  </option>
+                <?php endforeach; ?>
+              </select>
+            </td>
+          </tr>
+        <?php endforeach; ?>
+      </table>
+      <input type="submit" name="deleteSelected" value="<?=$this->transEscAttr('delete_selected')?>" class="btn btn-default">
```
demiankatz: I realize that this is another detail inherited from the existing Tag functionality, but I wonder if duplicating these buttons at the top would make the form more user-friendly.

themes/bootstrap3/templates/admin/feedback/home.phtml
```diff
@@ -0,0 +1,153 @@
...
+      <input type="submit" name="deletePage" value="<?=$this->transEscAttr('delete_page')?>" class="btn btn-default">
+      <input type="submit" name="deleteFilter" value="<?=$this->transEscAttr('delete_all')?>" class="btn btn-danger">
```
demiankatz: The "delete all" button does not seem to work. But do we even want such a dangerous feature?

demiankatz (changes requested):
Thanks again for all the work on this. I took time to review the remaining code and do some hands-on testing. See comments below.

I also wonder if perhaps we might want to make the feedback statuses configurable in some way (and maybe use a string instead of an enum in the database). While the set you have proposed seems very useful and flexible, this feels like a detail that could be inconvenient to customize should somebody wish to use a slightly different workflow.

module/VuFindAdmin/src/VuFindAdmin/Controller/FeedbackController.php
xmorave2: That does make sense. I have modified the FeedbackController this way.

module/VuFindAdmin/src/VuFindAdmin/Controller/FeedbackController.php
xmorave2: You are right. Done

themes/bootstrap3/templates/admin/feedback/home.phtml
xmorave2: It does not harm. Done

themes/bootstrap3/templates/admin/feedback/home.phtml
xmorave2: Done.

themes/bootstrap3/templates/admin/feedback/home.phtml
xmorave2: I've add truncation to 100 characters and more/less link after the text.

themes/bootstrap3/templates/admin/feedback/home.phtml
xmorave2: It could be useful, I've added the buttons on top of table too.

themes/bootstrap3/templates/admin/feedback/home.phtml
xmorave2: It's too dangerous, I've removed it.

xmorave2:
@demiankatz Thanks for review. I tried to address all your comments. 

Regarding status configuration: I do agree it could be better to have such an feature and I am willing to work on it. But if it could wait for another PR, it would better fits into my work plans.

languages/en.ini
demiankatz: On comparing to the existing `confirm_delete_tags_brief ` string, I think the answer is actually no. Sorry about that! :-)

demiankatz:
Thanks, @xmorave2, I will do some more hands-on testing shortly and will let you know if I find any other problems, but the code now looks good to me! Regarding status configuration, I do not mind merging this as-is and revising it later in a separate PR. My only question is whether it would be worth using a field type other than enum for the statuses now, to make it easier to change in the future -- otherwise, when we increase flexibility, the database migration work may be more complicated (and local customization would require database changes, which might be more difficult to maintain locally).

xmorave2:
@demiankatz Thank you! I've changed the status field from enum to varchar, it will be more flexible for future enhancement and also now are statuses controlled on one place only (in controller) which should more maintanable.

demiankatz:
Thanks again, @xmorave2 -- I've finished testing and haven't been able to break anything significant. I've also modified the tags code (PR #2487) as discussed above. During testing, I did notice a small bug -- if you apply a filter of All/All/All, you get a "clear filter" link even though there really isn't any filter applied. This bug is copied over from the tag manager, so I have fixed it in https://github.com/vufind-org/vufind/pull/2487/commits/a182541e7cbaf392948a630e2d6abd9e8dfcc201 -- you might want to make an equivalent fix here.

Once we have finished our conversation about the database design, I would like to do one more round of testing using PostgreSQL to be sure that works for me too, and then I think this can be merged.

demiankatz:
@xmorave2, sorry, I left that last comment without noticing that you had already addressed the database flexibility issue! I guess it's then just down to the filter fix (if you agree that it is helpful). Once that's done, I'll do a final round of testing and will merge if no problems are found.

xmorave2:
I fixed the 'clear filter' link issue now. It is behaving much better this way I think. Thanks @demiankatz

demiankatz:
@xmorave2, have you successfully tested this with PostgreSQL? I just tried it out in my test environment (using `phing startup -Ddbtype=pgsql` to create a fresh PostgreSQL-based environment), and nothing worked for me at all. In fact, the feedback table wasn't even created in the database due to some syntax issues. I pushed up commit [eab9a7d](https://github.com/vufind-org/vufind/pull/2399/commits/eab9a7dcb6fd11ae5117a7f9f0ad4acdb02d1063) in an effort to fix the problems that were preventing the database from being created, and to reorganize the PostgreSQL database definition so that the new table is added more consistently with other existing tables. I also added a missing configuration line in module.config.php that I think may be important. However, it's still not working for me. When I try to save feedback, I get this error in my log: `VuFind\Form\Handler\Database: Could not save feedback data: While processing primary key data, a known key id was not found in the data array`.

Unfortunately, I'm out of time for this week. I suspect the problem is somehow related to the module.config.php configuration (or maybe some factory isn't building some database-related object correctly for PostgreSQL -- it's been long enough that I can't remember exactly how this works). Do you have any ideas? If not, I'll do some further investigation when I have more time next week. In the meantime, I've put checkboxes on this PR to remind myself that we need to fix this, and also that we should test migrating from the current dev branch to be sure things work correctly after applying the migration.

xmorave2:
@demiankatz I have to admit I don't know postgre sql enough... But I probably found the issue and fixed it hopefully:

It seems that creating empty row, populating data into it and then save does not work for pgsql (at least using PDO) in laminas-db - although it is working form mysql/mariadb.

Laminas is not capable to get last insert id in that case and some code later, if there is no primary key, the exception is thrown.

I used `TableGateway::insert` and it started working.

languages/en.ini
```diff
@@ -433,8 +437,21 @@ fav_list_delete_fail = "Sorry, an error has occurred. Your list was not deleted.
 Favorites = "Saved Items"
 Fee = "Fee"
 Feedback = "Feedback"
+Feedback Management = "Feedback Management"
+feedback_delete_failure = "Feedback deletion failed"
+feedback_delete_filter = "You are using following filter - Form name: %formname%, Site URL: %siteurl%, Status: %status%"
+feedback_delete_success = "Feedback deleted: %count%"
```
demiankatz: I wonder if `"%count% feedback responses deleted."` would sound more natural. Also note that we normally use double-percents for placeholders (e.g. `%%count%%`) but your code seems to be using single percents. I think it's worth adjusting that for consistency.

demiankatz (changes requested):
Thanks for the pgsql fix, @xmorave2 -- this is working for me now, and I think the code is actually slightly more readable this way too.

I'm still in the midst of some hands-on testing and will add comments if I find any other problems, but so far it's looking good. See below for a minor comment about language files and wording.

Also, one functional question: I notice that if I change the status of a record while I have a filter applied, the filter gets cleared. Is this intentional? I imagine there may be good reasons for this behavior, but it also might not be expected. I am okay with either approach but wanted to check on whether you view this as a bug or a feature.

languages/en.ini
xmorave2: I changed the translation and also token identifiers for consistency.

xmorave2:
Thanks @demiankatz. 

Removing filters is bug and should be fixed now.

demiankatz (approved):
Thanks for all the work on this, @xmorave2! I believe it is ready to merge now, once I have finished the remaining documentation tasks (which I will do right now).

PR #2401, opened 2022-04-20
Fix: Lightbox external links

crhallberg:
[VUFIND-1523](https://vufind.org/jira/browse/VUFIND-1523)


themes/bootstrap3/js/lightbox.js
```diff
@@ -250,10 +250,14 @@ VuFind.register('lightbox', function Lightbox() {
    */
   _constrainLink = function constrainLink(event) {
     var $link = $(this);
+    var urlRoot = location.origin + "/" + VuFind.path;
+
     if (typeof $link.data("lightboxIgnore") != "undefined"
       || typeof $link.attr("href") === "undefined"
       || $link.attr("href").charAt(0) === "#"
       || $link.attr("href").match(/^[a-zA-Z]+:[^/]/) // ignore resource identifiers (mailto:, tel:, etc.)
+      || ($link.attr("href").slice(0, 4) == "http" // external links
```
demiankatz: Is there a reason to use `==` instead of `===` here? GitHub Actions is complaining about it.

demiankatz (changes requested):
Thanks, @crhallberg -- one very small question.

themes/bootstrap3/js/lightbox.js
crhallberg: For string comparison there is no difference so == is usually clearer but that's not what our styles enforce - will fix.

crhallberg: Done.

demiankatz:
@crhallberg, I did some hands-on testing, and this wasn't quite working right. I think I've fixed it with 7109b7b -- there was an extra slash in the constructed base URL, and the `indexOf` method wasn't capitalized correctly. I think it's all good now, but do you mind double-checking my work?

demiankatz (approved):

demiankatz:
Thanks for the double-check, @crhallberg -- merging now!

PR #2402, opened 2022-04-20
Update Google Analytics helper for compatibility with GA4

demiankatz:
The Google Analytics code has not been updated since Universal Analytics was new... and now it has reached end of life, being deprecated in favor of GA4 in 2023. This PR removes the legacy code which is no longer functional and also removes references to Universal Analytics, since they will soon be irrelevant. It also updates the Javascript to the latest standard.

TODO
- [x] Confirm that this still works correctly with a GA4 API key
- [x] Update changelog with note about need to migrate to GA4 (and also to change cookieFlags to cookie_flags where applicable).

### DIFF
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -2119,13 +2119,9 @@ treeSearchLimit = 100
 ;[GoogleTagManager]
 ;gtmContainerId = "GTM-1234567"
 
-; Uncomment this section and provide your API key to enable Google Analytics. Be
-; sure to set the "universal" setting to true once your account is upgraded to
-; Universal Analytics; see:
-; https://developers.google.com/analytics/devguides/collection/upgrade/guide
+; Uncomment this section and provide your API key to enable Google Analytics.
 ;[GoogleAnalytics]
 ;apiKey = "mykey"
-;universal = false
 ; Options to pass to the ga() function's create call; if omitted, defaults to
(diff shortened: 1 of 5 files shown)


demiankatz:
We should probably try to get this done for the 9.0 release so we get ahead of the GA4 launch. I currently have this temporarily deployed to the public vufind.org/demo instance for testing purposes. If that works, I think we'll be in good shape to merge it.

EreMaijala (changes requested):
I can't test this, but the changes seem ok and appear to work with the demo instance. However, I'm seeing a content security policy error being reported in browser's console. I think it would be useful to provide example settings that allow GA4 to work with VuFind's CSP enabled.

demiankatz:
Thanks, @EreMaijala, does the change in 3d6d6f6651ea495b34dd7b1136c6aecb277db659 look reasonable to you? I've applied this to the demo instance and it seems to have quieted the CSP warnings. (I was chasing an additional inline script error for a while but then realized it was coming from a local browser extension, not the VuFind demo site :-) ).

EreMaijala (approved):
Yes, that seems good to me if it does the trick. :)

demiankatz:
Thanks again, @EreMaijala. I think this is ready to merge, but we'll let it collect data over the weekend and look at it early next week just to be sure things look normal. If that's all good, then I'll merge it.

PR #2403, opened 2022-04-21
improve format detection for articles encoded as "Monograph Component Parts"

eocarragain:
Hi,
A small tweak to correctly identify journal articles. 

Some context:

- MARC uses LDR/07 to encode "component parts" with "a" for "Monographic Component Parts" and "b" for "Serial Component Parts"
- Despite the name, "Monographic Component Parts" is the correct encoding for "an article in a single issue of a periodical" as well as things like book chapters. "Serial Component Parts" is for things like ongoing columns or features that appear regularly in a parent serial. See https://www.loc.gov/marc/bibliographic/bdleader.html
- Vufind currently maps "Monographic Component Parts" to "book chapter" only, and maps "Serial Component Parts" to article
- In order to distinguish "book chapters" and "articles" that have been encoded as "Monographic Component Parts", the convention seems to be to rely on the corresponding 773 host field: if there is a $g, the item is journal article, otherwise "book chapter". See https://www.oclc.org/bibformats/en/specialcataloging.html#relatedpartsandpublications

Note: this PR simply forces a value of "SerialComponentPart" where a 773$g is detected, which gets mapped on to "Article". Technically it isn't a "Serial Component Part" but since these map to Article anyway, this is probably fine for now.

### DIFF
--- a/import/index_java/src/org/vufind/index/FormatCalculator.java
+++ b/import/index_java/src/org/vufind/index/FormatCalculator.java
@@ -214,7 +214,7 @@ public class FormatCalculator
                 break;
             // Component parts
             case 'a':
-                return (hasSerialHost(record)) ? "SerialComponentPart" : "BookComponentPart";
+                return (hasSerialHost(record)) ? "Article" : "BookComponentPart";
             case 'b':
                 return "SerialComponentPart";
             // Integrating resources (e.g. loose-leaf binders, databases)
--- a/import/index_java/src/org/vufind/index/FormatCalculator.java
+++ b/import/index_java/src/org/vufind/index/FormatCalculator.java
@@ -214,7 +214,7 @@ public class FormatCalculator
                 break;
             // Component parts
             case 'a':
-                return "BookComponentPart";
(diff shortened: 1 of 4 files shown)


demiankatz (changes requested):
Thanks for catching this, @eocarragain! A minor suggestion for adjustment, though: rather than returning "SerialComponentPart" from the new special case, why don't you return "Article" instead? Then you can add `Article = Article` to the format_map.properties file. This way, the net effect is the same, but if somebody wanted to differentiate "true" serial component parts from articles, they would be able to do so by adjusting the translation map. Feels a little bit more flexible as well as more technically correct. What do you think?

eocarragain:
Thanks @demiankatz. I was wary of messing with SerialComponentPart logic/mapping, so this is a great solution. I've pushed an update now including level0 and level1 format maps (I wasn't too sure how these work, but just replicated the existing mappings for SerialComponentPart, so hopefully that's right)

demiankatz (approved):
Thanks again, @eocarragain -- this looks great!

PR #2406, opened 2022-04-21
Add caching support to Primo and EDS connectors.

EreMaijala:
Moves the previously Solr-specific caching routines to a support trait and uses it in Primo and EDS as well.


config/vufind/EDS.ini
```diff
@@ -305,3 +305,19 @@ AU = None
 ;excludeLabel[] = 'Availability'
 ; This repeatable setting can be used to filter out values using group codes:
 ;excludeGroup[] = 'URL'
+
+; This section provides settings for optional caching of search requests. This
+; affects the primary Solr backend only; for caching of other backends, see the
```
demiankatz: This comment mentions Solr and should probably be updated!

config/vufind/Primo.ini
```diff
@@ -259,3 +259,19 @@ onCampusRule['MYINSTITUTION'] = "primoOnCampus.MYINSTITUTION"
 ; If you omit it, only users matching an onCampusRule or institutionCode can use
 ; Primo.
 ;defaultCode = "DEFAULT"
+
+; This section provides settings for optional caching of search requests. This
+; affects the primary Solr backend only; for caching of other backends, see the
```
demiankatz: ...ditto on the Solr reference.

config/vufind/EDS.ini
```diff
@@ -305,3 +305,19 @@ AU = None
...
+; appropriate backend-specific configuration files (Search2.ini, website.ini, etc.).
+[SearchCache]
+; Supported adapters: Memcached, Filesystem. Others may also work, but have not been
+; tested.
+;adapter = Memcached
```
demiankatz, line 313: I wonder if we should update the adapter settings to use the fully qualified class names, for consistency with other code.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
```diff
@@ -407,15 +409,16 @@ protected function call(
...
+            $cacheable
         );
         return $this->process($response);
     }
 
     /**
      * Process EDSAPI response message
      *
-     * @param array $input The raw response from Summon
+     * @param string $input The raw response from Summon
```
demiankatz: This comment mentions Summon -- might as well fix that too while we're in the area.

module/VuFindSearch/src/VuFindSearch/Backend/Feature/ConnectorCacheTrait.php
```diff
@@ -0,0 +1,127 @@
...
+            $this->cache->setItem($key, $response);
+        } catch (\Laminas\Cache\Exception\RuntimeException $ex) {
+            // Try to determine if caching failed due to response size and log the
+            // case accordingly. Unfortunately Laminas Cache does not translate
+            // exceptions to any common error codes, so we must check codes and/or
+            // message for adapter-specific values.
+            // 'ITEM TOO BIG' is the message from the Memcached adapter
+            // and comes directly from libmemcached.
+            if ($ex->getMessage() === 'ITEM TOO BIG') {
+                $this->debug('Cache setItem failed: ' . $ex->getMessage());
```
demiankatz: Is the idea that we only want to use debug level for TOO BIG because it's a common/expected problem, and we want unexpected errors to get higher logging priority? If that's the case, I wonder if we should also consider using a different message for the debug case -- maybe something more specific -- to make this more clear.

demiankatz (changes requested):
This looks good to me, but I have a few minor suggestions (and I realize that some of these refer to existing code that I just happened to notice because you moved things around). :-)

config/vufind/EDS.ini
EreMaijala: Fixed.

config/vufind/Primo.ini
EreMaijala: Fixed.

config/vufind/EDS.ini
EreMaijala: I'm a bit torn on this. While the fully qualified class name is more exact (and perhaps consistent as well), I'd really like to move to the other direction in the general configuration files and hide the implementation details from the user when possible. And I feel that a fully qualified class name is such detail and ties the configuration too tightly to the implementation and complicated things if the implementation needs to be changed later.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
EreMaijala: Heh, managed to miss that somehow! Thanks for catching it, now tweaked.

module/VuFindSearch/src/VuFindSearch/Backend/Feature/ConnectorCacheTrait.php
EreMaijala: Yes, it can be a common occurrence, and should not pollute the logs unless debug level is enabled.

Worth mentioning, since we discussed the error message handling during the [initial implementation](https://github.com/vufind-org/vufind/pull/1854#discussion_r576839049): While the error code from memcached is now available, I didn't feel like it's any better to compare the code with a seemingly magic value of `37` (available as a constant in Memcached extension, but would need to check the presence of the extension first). If you think it would be better, I can change it.

EreMaijala: I updated the debug message test.

config/vufind/EDS.ini
demiankatz: Good point, that level of detail doesn't really belong in the configuration; if we need to use FQCN, we should translate it in the code anyway... So never mind! :-)

module/VuFindSearch/src/VuFindSearch/Backend/Feature/ConnectorCacheTrait.php
demiankatz: I think it's worth changing to use the error code, since that seems less likely to change in the future. Since checking for the constant is extra work, maybe the best thing to do is to compare against the magic number but put a comment indicating that the number is also defined as a constant, but the constant may not be available due to the optional nature of the PHP extension.

EreMaijala: Now changed. And that's a good thing. Now that I verified the functionality, the exception message is a bit confusingly "ITEM TOO BIG; SUCCESS".

module/VuFindSearch/src/VuFindSearch/Backend/Feature/ConnectorCacheTrait.php
```diff
@@ -0,0 +1,133 @@
+<?php
+/**
+ * Caching support trait for connectors.
+ *
+ * Prerequisites:
+ *
+ * - LoggerAwareInterface
```
demiankatz: The methods you depend on here come from LoggerAwareTrait, but are not required by LoggerAwareInterface. Maybe it would be better to just compose the trait here instead of referencing it via a comment.

demiankatz (changes requested):
This looks pretty good to me, though I haven't had a chance to do any hands-on testing yet (if there are specific aspects of EDS you'd like me to try, let me know -- otherwise I'll just do a general check next week). In the meantime, just one observation...

module/VuFindSearch/src/VuFindSearch/Backend/Feature/ConnectorCacheTrait.php
EreMaijala: I changed that to LoggerAwareTrait. I'm not really a fan of using traits in traits since it's confusing when looking at the main class that doesn't directly use a trait but accesses its methods. Also, didn't it cause a problem when a class uses two traits that in turn use the same trait (thinking about the OAuth2 trait with ILS drivers)?

demiankatz: Yes, that makes sense. There's just no good way to safely compose traits, which is frustrating. (Possibly a sign that we're pushing this mechanism too hard... but it is very useful when it works).

A few alternatives that might be worth considering:

Option 1: Revert to a dependency on LoggerAwareInterface, but instead of using the LoggerAwareTrait's convenience methods, instead just call methods directly on `$this->logger` inside an `if ($this->logger ?? false) {` block. That way, failing to compose the dependency isn't a fatal problem. You could even put in an else with an `error_log` message about a missing logger.

Option 2: Keep the dependency on LoggerAwareTrait but add `is_callable` checks.

Option 3: Create a private helper message for logging within the trait, and do Option 1 or Option 2 inside it.

Maybe I'm overthinking all of this, but I just hate the idea that somebody can use a trait and create a situation where a fatal error triggers on the basis of a rare condition, and when it occurs, it would be rather difficult to figure out why.

EreMaijala: No overthinking, it's good to do this properly. How about now?

EreMaijala: I decided to check for logger with the correct interface so that it doesn't need to be any of the "...Aware..." things, but logger itself needs to implement the correct interface.

EreMaijala: Also, I think a shared base class for connectors would help here structure the code better.

demiankatz: Thanks, this looks good to me. I'll do a bit of testing later today (though I can only test EDS) and report back if I encounter any problems.

The idea of a common connector base class is a good idea -- do you think it's worth a JIRA ticket to capture for future reference? I don't think it's our top priority right now, but definitely a "nice to have" for the future. The trick is that not every backend even necessarily has to have a separate connector class, and some backends rely on third-party libraries as connectors, so I don't think we want to enforce any specific interface across all backends... but that doesn't mean we can't have a baseline for internal connectors as an optional best practice.

EreMaijala: Yes. I added [a ticket](https://vufind.org/jira/browse/VUFIND-1549).

demiankatz: Thanks! I added a comment there to summarize this thread (and link back to this PR for context).

demiankatz (approved):
Looks good to me now!

PR #2407, opened 2022-04-21
Fix long cart print not loading (VUFIND-1509)

crhallberg:
I'm concerned that this may be over-engineered (both in the past and for this solution).

- bypass Hunt
- wait for queued status AJAX

TODO
- [x] Resolve [VUFIND-1509](https://vufind.org/jira/browse/VUFIND-1509) when merging

I cannot wait to revisit this in the new theme.

### DIFF
--- a/themes/bootstrap3/js/check_item_statuses.js
+++ b/themes/bootstrap3/js/check_item_statuses.js
@@ -181,8 +181,7 @@ VuFind.register('itemStatuses', function ItemStatuses() {
       return;
     }
     if ($item.find('.hiddenId').length === 0) {
-      VuFind.emit("item-status-done");
-      return false;
+      return;
     }
     var id = $item.find('.hiddenId').val();
     var handlerName = 'ils';
@@ -205,6 +204,8 @@ VuFind.register('itemStatuses', function ItemStatuses() {
     for (var i = 0; i < ajaxItems.length; i++) {
       checkItemStatus($(ajaxItems[i]));
     }
+
+    VuFind.emit("item-status-done");
   }
   function init(_container) {
     if (typeof Hunt === 'undefined' || VuFind.isPrinting()) {
--- a/themes/bootstrap3/js/check_item_statuses.js
+++ b/themes/bootstrap3/js/check_item_statuses.js
@@ -204,8 +204,6 @@ VuFind.register('itemStatuses', function ItemStatuses() {
     for (var i = 0; i < ajaxItems.length; i++) {
       checkItemStatus($(ajaxItems[i]));
     }
-
-    VuFind.emit("item-status-done");
   }
   function init(_container) {
     if (typeof Hunt === 'undefined' || VuFind.isPrinting()) {
--- a/themes/bootstrap3/js/check_item_statuses.js
+++ b/themes/bootstrap3/js/check_item_statuses.js
@@ -157,6 +157,7 @@ VuFind.register('itemStatuses', function ItemStatuses() {
         .fail( this.itemStatusFail)
         .always(function queueAjaxAlways() {
           this.itemStatusRunning = false;
+          VuFind.emit("item-status-done");
         });
     }//end runItemAjaxForQueue
   };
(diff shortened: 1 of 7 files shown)


themes/bootstrap3/js/check_save_statuses.js
```diff
@@ -124,12 +139,10 @@ function checkSaveStatuses(_container) {
         id: idval,
         source: $source.val()
       }, $(ajaxItems[i]));
+    } else {
+      VuFind.emit("save-status-done");
```
demiankatz: Is this right? This else is inside a loop that's checking multiple things. Does this mean that if one entry in the list is missing id/source, it will cause a false "done" status to be emitted? Or am I misunderstanding?

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -180,6 +181,7 @@ VuFind.register('itemStatuses', function ItemStatuses() {
       return;
     }
     if ($item.find('.hiddenId').length === 0) {
+      VuFind.emit("item-status-done");
```
demiankatz: Do we want to mark the whole process done just because a single item was missing an ID?

demiankatz (commented):
Thanks, @crhallberg. I haven't had a chance to do hands-on testing of this yet, but based on a code read, it looks to me like about as elegant a solution as could be found given the current environment. I just had a couple of questions about behavior when hidden id/source information is missing. Maybe I'm misunderstanding, but it seems to me like there might be some potential problems here (though it's been long enough that I'm not entirely sure what circumstances would lead to these hidden elements being missing in the first place).

demiankatz:
@crhallberg, it also looks like there's a small jshint error in need of fixing.

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -203,9 +204,11 @@ VuFind.register('itemStatuses', function ItemStatuses() {
     for (var i = 0; i < ajaxItems.length; i++) {
       checkItemStatus($(ajaxItems[i]));
     }
+
+    VuFind.emit("item-status-done");
```
demiankatz: Is it really right to emit this status here? Isn't checkItemStatus queuing up work? If so, don't we want to emit this when the work is actually done, not when we're done creating the queue?

themes/bootstrap3/js/check_save_statuses.js
```diff
@@ -126,10 +141,8 @@ function checkSaveStatuses(_container) {
       }, $(ajaxItems[i]));
     }
   }
-  // Stop looking for a scroll loader
-  if (saveStatusObserver) {
-    saveStatusObserver.disconnect();
-  }
+
+  VuFind.emit("save-status-done");
```
demiankatz: Same question for this one -- do we need it? Can't we rely on the queue emitting the status when it's done processing?

demiankatz (changes requested):
Thanks for the progress, @crhallberg. I went ahead and fixed the jshint issue to get a green checkmark, but I have a couple of questions -- I'm wondering if some of these event emissions are not actually needed/appropriate.

themes/bootstrap3/js/check_save_statuses.js
crhallberg: I think I spammed a whole bunch all over to get it working. I just need to make sure that a failure also sends out the message so print doesn't wait forever. Maybe I can put a time limit on it? 3 seconds? 6 seconds?

demiankatz: @crhallberg, a time limit seems like a reasonable compromise, though I think it's inevitable that this will result in incomplete results in some situations (e.g. printing a results page with 100 items, using an ILS that runs slowly). Maybe the time limit should be relative to the last activity, rather than to the start of the loading process? And would it be more palatable if we had a "Loading..." message up while things get prepared for printing?

(Note that my goal is to do this as simply as possible, so don't let any of my ideas cause excessive complication... just brainstorming here!)

crhallberg: I believe I have covered all cases but maybe a timeout as a failsafe would be good to add.

PS. I cannot wait to rewrite all of this logic for the new theme.

demiankatz (approved):
This looks good to me now! Even if some odd edge cases remain, it's certainly less broken than what came before.

PR #2408, opened 2022-04-22
improve format detection for electronic integrating resources 

eocarragain:
Another format tweak to allow for more specific forms of "OnlineIntegratingResource", starting with 'Websites'

Context:

- Vufind currently classifies any electronic (007/00=c) integrating resource (ldr/07=i) as "OnlineIntegratingResource" and ultimately mapped to "Database"
- 008/21 can allow us to be more specific here
- e.g. this initial patch handles blogs and websites (an NLI use-case), but more can be added

_Note: this commit also changes the "serial" switch case under getFormatFromBibLevel() to allow for rare edge-cases where the ldr/07=s (serial) but no 008 is present (these will still get mapped to 'Serial')_


import/translation_maps/format_map.properties
```diff
@@ -57,3 +57,4 @@ VideoDisc           = DVD
 VideoOnline         = Video
 VideoReel           = Video
 Video               = Video
+Website             = Website
```
demiankatz: It would probably be good to add an end-of-file line break to all of the properties files just for consistency.

demiankatz (changes requested):
Thanks again for the ongoing improvements here, @eocarragain; the logic looks sound to me! Just two comments: a near-trivial formatting issue (see below) and the fact that you should add `Website = Website` to en.ini (and ga.ini and others, if you're able to provide translations you're confident in), since this adds a new format string, and those need to be translated.

eocarragain:
thanks for reviewing @demiankatz . Those changes are committed now. cheers

demiankatz (approved):

demiankatz:
Thanks, @eocarragain -- looks great. Merging now!

PR #2412, opened 2022-04-26
improve format detection when 007 is missing

eocarragain:
Another proposed format parsing change to use 008 over 007 where possible

Context:

- The MARC 008 field is required, whereas the 007 is optional. See https://www.loc.gov/marc/bibliographic/nlr/nlr00x.html
- Currently Vufind depends a lot on 007 when interpreting the LDR/06 and LDR/07. This causes issues for records without a 007 even though the 008 often contains the relevant information
- This PR tries to use 008 over or in addition to 007 where possible
- It also adds a few suggest changes when parsing 007

@demiankatz would appreciate a review as discussed. Cheers!

TODO:
- [x] Update changelog with backward-breaking changes: method signature changes to getFormatFromBibLevel, getFormatFromRecordType and isElectronic, plus elimination of getTypeOfComputerFile.


import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -261,25 +287,61 @@ protected String getFormatFromBibLevel(Record record, char bibLevel, char format
...
+                return "Slide";
             case 'i':
                 return "SoundRecording";
             case 'j':
                 return "MusicRecording";
             case 'k':
+                if (marc008 != null) {
+                    // Check 008/33 Type of visual material
+                    switch (marc008.getData().toLowerCase().charAt(33)) {
+                        case 'k': // Graphic
```
eocarragain: Having run this over a wider set of records, I think it is safer not to map 'k' ("Graphic") to Drawing as it is too vague and, in our case at least, appears to also have been used for lots of photos.

import/translation_maps/format_map.properties
```diff
@@ -36,6 +36,8 @@ Photonegative       = Photo
 Photo               = Photo
 PhysicalIntegratingResource = Reference Material
 PhysicalObject      = Physical Object
+Poster              = Photo
```
demiankatz: I understand that mapping these to Photo is backward-compatible and a safe path forward... but I wonder if some of these things justify adding some new translations, since both a postcard and a poster are not necessarily a photograph, etc.

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -209,6 +217,18 @@ protected String getFormatFromBibLevel(Record record, char bibLevel, char format
             // Monograph
             case 'm':
                 if (couldBeBook) {
+                    if (marc008 != null) {
+                        // Check 008/23 Form of item
+                        switch (marc008.getData().toLowerCase().charAt(23)) {
```
demiankatz: Should we check for a particular length of 008 in addition to checking that it is not null, so that malformed/incomplete 008's don't cause errors?

demiankatz: (I realize that there are quite a few existing potentially unsafe checks... maybe a helper method for pulling values from 008 would be in order).

import/index_java/src/org/vufind/index/FormatCalculator.java
demiankatz: Would you just drop the case and fail through to the default, then?

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -219,21 +239,27 @@ protected String getFormatFromBibLevel(Record record, char bibLevel, char format
...
-                    // Look in 008 to determine type of electronic IntegratingResource
-                    if (marc008 != null) {
-                        switch (marc008.getData().toLowerCase().charAt(21)) {
-                            case 'h':
-                            case 'w':
-                                return "Website";
-                            default: break;
-                        }
+                // Look in 008 to determine type of electronic IntegratingResource
+                if (marc008 != null) {
```
demiankatz: One thing I'm a bit confused about (and this is probably just because I haven't read the right parts of the spec) is how we can reliably know which of the versions of the 008 we have. In some cases we seem to be assuming based on the bibliographic level, and other times we seem to be assuming based on the type of record. Is this potentially dangerous?

demiankatz (commented):
Thanks, @eocarragain -- I've put a few comments based on a quick review. I also plan to talk to the cataloging team here to see if they have any additional thoughts...

demiankatz:
@eocarragain, some feedback (based on your general approach, not necessarily the specific code) from one of our catalogers:

"I think we went with 007 because it can give more precise information in many instances.  Compare 007 vs. 008 in [Star Wars (#787275)](https://library.villanova.edu/Find/Record/787275#details).  The 008 says it’s a “visual” record—rather vague.  Drilling down into the 008, one can find Video Recording  in Type of material, but that’s as precise as one gets.  But the 007 says Video recording at the get-go, and if one wants to drill down it will clarify as Videodisc and DVD.  Whether we need to go further that the 008 Type of Material code really depends on how the facets are assigned.  Certainly for visual materials, I don’t think we have video recordings aside from DVDs in our collection any more.  If all the facets we assign can be covered by the 008, than that is probably the way to go for us.  Going forward, as more of our records are RDA-ified, refinement beyond the 008 Type of Material can be found in the 337/338 fields."

Thought that might be helpful to pass along.

import/index_java/src/org/vufind/index/FormatCalculator.java
eocarragain: yes, i think so. On the basis of not changing currently behaviour, i.e. it used always return 'Photo' before

eocarragain: yes, i think so. On the basis of not changing currently behaviour, i.e. it used always return 'Photo' before

import/index_java/src/org/vufind/index/FormatCalculator.java
eocarragain: sounds good. I'm not sure if there is a standard length 008, but can check. Are you suggesting checking each time that there are at least the number of chars needed, e.g. 24 in this case but more elsewhere?

import/index_java/src/org/vufind/index/FormatCalculator.java
eocarragain: i wondered that too but I think it is OK. if you look at a given 008 version, they tell you the requirements re LDR/06 and LDR/07, i.e. some say use this only when LDR/06 =x or y. So I think the way we're doing it within specific cases of either LDR/06 and LDR/07 is right.

import/translation_maps/format_map.properties
eocarragain: Sure, I'd take your advice here. I was working on the basis of changing as little as possible for now.

eocarragain:
Thanks @demiankatz & to your cataloguing colleague. Really helpful and makes perfect sense as 007 is definitely more detailed in a lot of cases. Our problem is the majority of our records don't have a 007 at all, and suspect we may not be alone in that. The idea was to allow for a kind of progressive enhancement so minimal records work just with LDR/06, LDR/07 and 008, but more complete records also get the benefit of 007. Where possible, I've tried to ensure that the value is returned from 008 in either getFormatFromRecordType() or getFormatFromBibLevel(), matches what is likely to come back from the getFormatFrom007(). so we don't end up with slightly different facet values for the same thing.

import/index_java/src/org/vufind/index/FormatCalculator.java
demiankatz: Yes, my thought was to have a method like `get008Value(marc008, position)` which returns the requested character from the 008, or null if it is undefined. Then instead of having null checks for marc008 everywhere, we can just fetch values and fail if they can't be found for whatever reason.

import/index_java/src/org/vufind/index/FormatCalculator.java
demiankatz: Thanks, I must have failed to find that exact part of the documentation. :-)

demiankatz:
Thanks, @eocarragain, that approach makes sense -- as long as we're adding specificity and not taking anything away, I think the risk is fairly low... especially since unwanted granularity can be "mapped back down" through the translation maps as needed.

It sounds like you might have an adjustment or two still to do -- please let me know when you feel this is reasonably stable, and then (once time permits) I can index our collection using this code, compare it against the default behavior, and report back if I see anything alarming.

eocarragain:
thanks @demiankatz 

> as long as we're adding specificity and not taking anything away

I think that should be the case

> It sounds like you might have an adjustment or two still to do

I've pushed changes now to address comments above: a) remove "graphic" mapping; b) add translations; c) 008 helper method (feedback on how I handled this, very welcome!)

> then (once time permits) I can index our collection using this code, compare it against the default behavior, and report back if I see anything alarming.

That'd be great. Maybe don't try on your production instance ;)

demiankatz:
@eocarragain, I ran the test, and these are the values which changed (note that I combined your code with our local customizations and mappings, so it may not be perfectly reflective of a "raw" run, but I think it's nonetheless informative).

Book | 2140261 | 2140383
Video | 119138 | 124463
Slide | 5 | 18049
Map | 9151 | 8918
Website | 0 | 7752
Picture | 445 | 444
Film | 3 | 284
Unknown | 204 | 0
Article | 3 | 238
Poster | 0 | 87
Filmstrip | 1 | 3

(First number is "before" count, second number is "after" count).

Note that we map "Atlas" to "Book" which at least partially explains why maps went down and books went up here (though since the numbers don't exactly add up, there have to be other factors in play as well).

I think the item I find most concerning is the huge jump in Slides, which seems to be related to MARC records from streaming video vendors. I wonder if either we need to look more closely at those mappings, or else we have thousands of records with bad coding in them. I don't currently have time to investigate in depth, but I'm happy to do so later if it would be helpful.

I've shared these results with the cataloging team to see if they have specific thoughts or concerns, and I'll share any outcomes from that process... but in any case, I think for the most part, this doesn't look too alarming (except for the slides), and it's certainly good news that "Unknown" has been completely eliminated!

eocarragain:
Hi @demiankatz . Thanks for doing this so quickly. Those numbers are really useful. I may know the issue with 'Slide' as we're seeing a similar thing with our records. I need to dig in a bit more though. Two questions:
- For those 'slide' records are you also seeing a second format assigned from the 007?
- Do you have a few examples from your public catalogue that I can take a look at, e.g. of the the streaming video records etc?

demiankatz:
@eocarragain, the test system I'm working with is indexing all of our records regardless of suppression status, and it turns out that we suppressed the vast majority of these video records a couple of years ago because they were confusing people. If I filter to unsuppressed records, there are only 64 slides in the results! Still, that's enough to turn up a few examples you might find useful:

https://library.villanova.edu/Find/Record/1202097

https://library.villanova.edu/Find/Record/1202005

https://library.villanova.edu/Find/Record/2253033

https://library.villanova.edu/Find/Record/1526467

Most of these look pretty similar to the first three examples above; the fourth one seems to be a weird outlier of some sort.

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -311,7 +316,9 @@ protected String getFormatFromRecordType(Record record, char recordType, Control
                         return "Video";
                     default: break;
                 }
-                return "Slide";
+                // Insufficient info in LDR and 008 to distinguish still from moving images
+                // Leave it to corresponding 007 if it exists, else fall back to "ProjectMedium"  
+                return (formatCodes007.contains('g')) ? "" : "ProjectedMedium";
```
demiankatz: Sorry if I'm confused, but is this logic inverted? If the format code indicates that this is a projected graphic, why are we returning empty string, and returning ProjectedMedium for all other format codes?

eocarragain: Twisted maybe, but I think it is right. Basically if we know there is a 007 for the same type of thing ("g" = project medium), return "" so that the 007 value is the only one used. If there isn't a 007 for the same type of thing, return "ProjectedMedium" as a best guess. Individual instances can map this to something else (includign the old default 'Slide') based on their content. Does that make sense. If you look back at the original, we were actually returning "" for similar reasons

eocarragain: Btw, this currently isn't compiling. I'll take a look later

demiankatz: I see, so if we have an 007 value of 'g' we've already assigned something more specific in `getFormatFrom007`, so adding a generic ProjectedMedium value here would be redundant and unhelpful. Maybe it's worth adjusting the comment to say that a little more explicitly to prevent future confusion. And if nothing else, there's a typo of "ProjectMedium" in the current comment.

If you need help with the compile problem, let me know and I can look again tomorrow. When you're ready, I can reindex again and report back with revised numbers!

eocarragain: Comments improved and compile problem solved

eocarragain:
thanks again @demiankatz. I'm hopeful the latest commits will help with that "Slide" issue. We'll run it against our records again tomorrow. If you can do the same when you get a chance that'd be great.

Sorry for messing up the commit history with garbage. If we think it is working, we might be able to rebase or squash or some other such magic? Happy to do a clean PR too if that's easier

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -199,17 +207,28 @@ protected String getFormatFrom007(char formatCode, String formatString) {
...
+                    // Check 008/23 Form of item
+                    switch (get008Value(marc008, 23)) {
+                        case 'o': // Online
+                        case 'q': // Direct electronic
+                        case 's': // Electronic
+                            return "eBook";
+                        default: break;
+                    }
+                    // Fall-back on 007 if 008 is missing
+                    // Note: relying on 007 here is not ideal as it is repeatable
```
eocarragain: Would appreciate your advice here @demiankatz  (and same below at line 258).

I left this in mostly for backwards compatibility but it doesn't really make sense to me:

- the 007 is repeatable
- Unlike the 008 which tries to encode the main "thing" being described, the 007 can provide additional detail about the main thing being described, or about part of the main thing being described, or about "accompanying material"
- So in this case, if the main thing being described is a good old fashioned printed book, there might be a 007 describing the fact that there is something electronic remote 007cr (e.g. a version of the book online, but presumably also some accompanying material that is online)
- in the latter case, we've now classed the good old fashioned book as an ebook incorrectly
- you can actually see this elsewhere on line 112 where we ignore 007cr entries because they are so vague and might end up with lots of erroneous "Electronic" formats
- Also, from what I can see the old version of this 007 check was somewhat arbitrarily relying on the formatCode from the last occurrence of 007?

As we're now looking in the 008 to determine if the "main thing" is electronic can we just drop this 007 check here (obviously anything returned separately by getFormatFrom007 will still be there). 

If so, we'd just return "Book" at this point rather than looking for 007c

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -261,26 +283,67 @@ protected String getFormatFromBibLevel(Record record, char bibLevel, char format
...
+                        return "MotionPicture";
+                    case 'v': // Videorecording
+                        return "Video";
+                    default: break;
+                }
+                // Insufficient info in LDR and 008 to distinguish still from moving images
+                // If there is a 007 for either "Projected Graphic", "Motion Picture", or "Videorecording" 
+                // that should contain more information, so return nothing here.
+                // If no such 007 exists, fall back to "ProjectedMedium"
+                if (formatCodes007.contains('g') || formatCodes007.contains('m') || formatCodes007.contains('v')) {
```
eocarragain, line 358: need to check for anything that could be a project medium type in 007, so 'g', 'm', 'v'. 

This should cover your first three examples, @demiankatz, i.e. the 008 has been coded as "other" ('z'), so we need to rely on one of the relevant 007s if they exist. Failing that, you'll end up with "ProjectedMedium" which is at least more accurate than "Slide"

demiankatz: Thanks, @eocarragain! I'm going to repeat the indexing with the latest code and will report format counts later today!

Also, to answer your earlier question about commit history, don't worry about it -- I almost always just "squash and merge" to finish up pull requests, unless they are so complex that they justify a more granular commit history. I think this one should be fine to simply squash.

languages/en.ini
```diff
@@ -606,6 +606,8 @@ ils_offline_login_message = "Your account details will be unavailable during thi
 ils_offline_status = "Our Library Management System is currently under maintenance."
 ils_offline_title = "System Under Maintenance"
 ils_transaction_history_disabled = "Loan history is not enabled for the active library card."
+Image = "Image"
+Image or Video = "Image or Video"
```
demiankatz: I don't think I like the addition of an "Image or Video" facet, because that doesn't really make much sense from a faceting perspective. Maybe it's better to just map it to Video, with a comment that it may not be appropriate, and let people address the ambiguities locally. Open to other suggestions -- I just think it's not going to help the user much to have a facet value that could be one of two other categories, but doesn't group with either of them.

eocarragain: Thanks @demiankatz . That's fine and agree this is clunky. I'll map ProjectedMedium to "Slide" (i think this a better default than Video and seems to have been used at some point in the past) in translation_map and leave it out of en.ini/ga.ini? The only alternative would be to have a user-facing string of "Projected Medium" but I don't think that'd make much sense to most users.

import/index_java/src/org/vufind/index/FormatCalculator.java
eocarragain: that's good to know re commit history. Thanks very much for re-indexing

languages/en.ini
demiankatz: Yes, sounds good to me -- mapping to Slide matches past precedent, as you say.

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -261,26 +283,67 @@ protected String getFormatFromBibLevel(Record record, char bibLevel, char format
...
+                        return "Filmstrip";
+                    case 't':
+                        return "Transparency";
+                    case 'm':
+                        return "MotionPicture";
+                    case 'v': // Videorecording
+                        return "Video";
+                    default: break;
+                }
+                // Insufficient info in LDR and 008 to distinguish still from moving images
```
eocarragain, line 354: @demiankatz we could probably try one more thing here:
-- if 008/34 is set to something, it should mean it is a video: "Technique - One-character alphabetic code that indicates the technique used in creating motion in motion pictures or videorecordings."

In your records, it is is set to 'c' ("Animation and live action") even though 008/33 Type of visual material is 'z' ("other"). So on this basis we could set default to "Video" instead of "Slide". What do you think?

demiankatz: That sounds reasonable to me!

demiankatz:
@eocarragain, my latest round of indexing is complete, and now Slide has dropped from 18,049 to 4, Picture has dropped from 444 to 134, and Projected Medium (which is new) has 4 entries. No other counts have changed. I think that's probably a good thing!

import/index_java/src/org/vufind/index/FormatCalculator.java
demiankatz: If you make this last change, I can run one more reindex and see if that cleans up the remaining "Projected Medium" values.

eocarragain:
great @demiankatz . I think we're getting there. I'll make the following changes based on comments above:

- additionally check 008/34 to see if we can further distinguish video from slide in ProjectedMedium category
- Remove "Image or Video" from language ini files and default ProjectedMedium mapping to Slide in translation_map

Any thoughts on my query above at line 229 re whether it is worth trying to maintain the 007 check there for backwards compatibility? I think we should drop it?

Thanks!

import/index_java/src/org/vufind/index/FormatCalculator.java
demiankatz: That sounds reasonable to me. We can make the change and see what it does to our tests, in any case!

demiankatz:
Thanks, @eocarragain, and sorry for missing the line 229 comment before; I think I just overlooked it among the other threads (and I have replied to it above now). I should have time to do more testing next week as needed!

eocarragain:
nice one @demiankatz . I'll make that change too & let's test next week. cheers

import/index_java/src/org/vufind/index/FormatCalculator.java
eocarragain: done

import/index_java/src/org/vufind/index/FormatCalculator.java
eocarragain: this is done now too

languages/en.ini
eocarragain: done in latest commit

eocarragain:
those changes made now, Demian. Thanks

demiankatz:
@eocarragain, thanks for the latest updates. I gave this another try, and none of my counts changed.

I did a bit of digging into some of the changes, though, and I ran into something strange: I now have 50 records with no format at all, which I didn't think was supposed to be possible. I would at least expect to get an "Unknown" value. Are you seeing the same thing with your record set (i.e. if you run a `-format:*` search)? It's possible the problem is somehow related to my local customizations or mappings, but your changes are definitely a factor, since in production, I have only one record without a format (and I'm not sure why that's the case either -- another thing to potentially investigate).

Anyway, [here is the record with no format in production, and which also has no format with your modified code](https://library.villanova.edu/Find/Record/89031#details), and [here is a sample record that has a format in production but lost it when testing with your code](https://library.villanova.edu/Find/Record/1691627#details).

I can email you binary MARC files of these if that would help... and I'm also happy to do some more detailed experiments on my test server if that would help. I'm just out of time for today and wanted to share my findings as soon as possible. Please let me know what my next step(s) should be!

eocarragain:
Thanks for testing more @demiankatz . Hmmm, I would have thought your counts would change especially with additional 008/34 check. 

Re records with no format, I just indexed the two records you linked to with unmodified code from this branch and both have a format assigned -"BookComponentPart" (mapped to Book Chapter) and "Image"

image

eocarragain:
Indexing those records on the master branch produces similar - "Book Chapter" and "Photo"

image

demiankatz:
Thanks, @eocarragain, it appears that my local translation map was missing a couple of entries, which is why some formats were missing. I've fixed that and am running yet another reindex; I'll report back later today.

In the meantime, one very minor thing that I just noticed: some of your edits to FormatCalculator.java have unnecessary trailing whitespace on them. You might want to adjust your editor settings to strip trailing whitespace.

demiankatz:
@eocarragain, just an update to say that the final round of indexing has finished, and the results seem pretty reasonable. I sent some notes along to our cataloger to see what he thinks of it -- one area of possible concern is that our count of maps has gone down significantly, and we're investigating -- but I suspect it's just a case of more accurately identifying Atlases (which we map to "Book"). I'll share another update once that conversation is done, and would welcome input from others in the meantime!

eocarragain:
Thanks @demiankatz. 

I also did a last read through the spec, and have pushed some more changes to cover:

- Better handling of items coded as "Computer file" in LDR/06 (these were mostly being ignored; i'm not sure why?)
- Other best-guess fallbacks from LDR/06 and LDR/07 before default to Unknown
- Additional comments to indicate why some LDR values aren't being used in particular places
- Additional mappings and translations for Dataset, Videogame, Text
- I think I've fixed trailing whitespace now

Sorry for these late changes & yet another re-index. They are related to the overall pull request (better format parsing when 007 missing), so I thought better to put them in now. 

I think these are the last changes for this PR. Looking forward to feedback from your cataloguing colleague & others on this list.

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -261,26 +287,108 @@ protected String getFormatFromBibLevel(Record record, char bibLevel, char format
...
+                // Check 008/26 Type of computer file
+                switch (get008Value(marc008, 26)) {
+                    case 'a': // Numeric data
+                        return "Dataset";
+                    case 'b': // Computer program
+                        return "Software";
+                    case 'c': // Representational
+                        return "Image";
+                    case 'd': // Document
+                        return "Software";
```
eocarragain: @demiankatz 'Text' might be better here, or a new value?

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -55,18 +55,23 @@ protected boolean definitelyNotBookBasedOn007(char formatCode) {
...
             case 'm':
-                // If this is a computer file containing numeric data, it is not a book:
-                if (getTypeOfComputerFile(marc008) == 'a') {
+                // Check the type of computer file:
+                // If it is 'Document', 'Interactive multimedia', 'Combination',
+                // 'Unknown', 'Other', it could be a book; otherwise, it is not a book:
+                char fileType = get008Value(marc008, 26);
+                if (fileType == 'd' || fileType == 'i' || fileType == 'm' || fileType == 'u' || fileType == 'z') {
+                    return false;
+                } else {
```
demiankatz: Since an "else return" is functionally equivalent to a return below the if statement, it might be more concise to remove the else entirely.

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -261,26 +287,108 @@ protected String getFormatFromBibLevel(Record record, char bibLevel, char format
...
+                        return "Video";
+                    default: break;
+                }
+                // Insufficient info in LDR and 008 to distinguish still from moving images
+                // If there is a 007 for either "Projected Graphic", "Motion Picture", or "Videorecording"
+                // that should contain more information, so return nothing here.
+                // If no such 007 exists, fall back to "ProjectedMedium"
+                if (formatCodes007.contains('g') || formatCodes007.contains('m') || formatCodes007.contains('v')) {
+                    return "";
+                } else {
```
demiankatz: Another "else return" that could potentially be condensed.

import/index_java/src/org/vufind/index/FormatCalculator.java
demiankatz: I guess the question is, is "Document" a subset of "Text" or vice versa? I wonder if it would make sense to use "Document" everywhere that "Text" is currently used. But ultimately it doesn't matter much since it can be easily remapped... but I do agree that "Software" is probably not as good as "Document" or "Text" here.

import/translation_maps/format_map.properties
```diff
@@ -48,12 +53,14 @@ SoundRecording      = Audio
...
 TapeCassette        = Software
 TapeReel            = Software
+Text                = Text
 Thesis              = Thesis
 Transparency        = Slide
 Unknown             = Unknown
 VideoCartridge      = Video
 VideoCassette       = VHS
 VideoDisc           = DVD
+VideoGame           = Video game
```
demiankatz: I would capitalize "Game" here, since multi-word formats are fully capitalized elsewhere in the file.

import/translation_maps/format_map.properties
```diff
@@ -12,6 +12,7 @@ Chart               = Photo
 ChipCartridge       = Software
 Collage             = Photo
 ConferenceProceeding= Conference Proceeding
+Dataset             = Dataset
```
demiankatz: I think "Data Set" is more common usage than "Dataset," though maybe that's an American perspective.

demiankatz (changes requested):
Thanks, @eocarragain! I'll do another test index this morning and see what happens. In the meantime, a few minor comments  (plus a reply to your query)...

import/index_java/src/org/vufind/index/FormatCalculator.java
eocarragain: Right, I think I'll go with "Document" here so as I suppose "document" here could include spreadsheet, powerpoint, etc. People can map that as they wish then, e.g. to "Electronic Document" or to "Text". 

I was thinking of "Text" as fall-back similar to "Image", in the same way as those broad categories are used by DPLA, Europeana, DataCite. Another query I had on this, was if "Text" should take over from things like "Book" in the format_map_level0.properties, again to be more consistent with how "Image" is used there. I didn't want to make that change though, as we don't use these hierarchical facets, so I don't have a good sense of how they work.

import/translation_maps/format_map.properties
eocarragain: "Data Set" is good enough for Wikipedia https://en.wikipedia.org/wiki/Data_set

import/index_java/src/org/vufind/index/FormatCalculator.java
demiankatz: Sounds reasonable to me -- so I guess we're adding two new formats, "Document" and "Text."

As for whether to use Text in the hierarchical formats, that might be useful, but I think it depends on local preferences. Probably best to leave it as-is for now, but it would be very easy to remap as needed. @EreMaijala might have stronger opinions than me, though, since he set that up originally.

demiankatz:
@eocarragain, I ran another set of tests, and my number of Book items reduced slightly, while I ended up with some new Data Sets, Video Games and Texts (for my test, I mapped both the Document and Text values to Text... but that may not have been the most useful thing to do; we can refine in the next iteration). For the most part, everything looks reasonable, except that I'm not convinced that the "Text" format is useful for me. It's only showing up in combination with other, more specific formats, so if its purpose is to simply be a fallback of last resort, I think it's getting overused. Of course, since I can't tell which of the two cases is causing it right now, maybe this feedback is not very helpful. I'll differentiate text from document and do another reindex and see what that tells me.

eocarragain:
Thanks @demiankatz. I'll address comments above later this evening. 

Agree - "Text" should really only be a last-ditch fall back. If Document gets mapped to Document, Text should only get assigned where no other value has been found (see line 603). It sounds like you may end up with some unhelpful "Documents" on next iteration though. It would be useful to see some of those records after next pass. Maybe we could:
-- just not return a value for "Document" as being too vague
-- OR only return it where there isn't a 007, i.e. where it is likely to be the only format

demiankatz:
@eocarragain, sure enough, 100% of my "Text" entries were actually the "Document" case. The majority of them are also being identified (perhaps inappropriately) as Books. Some examples:

https://library.villanova.edu/Find/Record/944923

https://library.villanova.edu/Find/Record/1955501

One is also a Journal: https://library.villanova.edu/Find/Record/459806

One is also a Conference Proceeding: https://library.villanova.edu/Find/Record/1476418

The rest are coming through as Online Serials; some examples:

https://library.villanova.edu/Find/Record/459965

https://library.villanova.edu/Find/Record/455606

I hope that sheds some light! :-)

import/index_java/src/org/vufind/index/FormatCalculator.java
eocarragain: set to ComputerDocument and mapped to Document in format_map

eocarragain:
Hi @demiankatz 
Another round of changes

- Addressed comments above re Data Set, Video Game etc
- Removed Document as not useful
- Set Books to eBooks where LDR/06 = 'm' (Comp File), except where these have already been excluded in definitelyNotBookBasedOnRecordType(), e.g. Data Set, Video Game
- Add "Electronic" to anything with LDR/06 = 'm' (Comp File). Query: I did this in isElectronic() rather than defaulting to Electronic in getFormatFromRecordType() on the basis of keeping all "Electronic" logic in one place. Let me know if you think it is cleaner to do it in getFormatFromRecordType() though (with 'Electronic' then being set in multiple places)
- Fixed bug caused by missing 'break' which was causing some material to get 'Serial' incorrectly
- Add "Nonmusical sound recording" (e.g. recorded interviews" to definitelyNotBookBasedOnRecordType()
- More strict enforcement of spec when looking up 008 based on an LDR/07 value

I'm not sure about this last one either. It is something you had raised before about the potential to misinterpret the 008. I've now enforced the following from the spec when dealing with LDR/07 (you have to look up the "FIELD DEFINITION AND SCOPE" part of each 008 type to see the rules):

| 008 Type  | LDR/06 is one of | **AND** LDR/07 is one of | 
| ------------- | ------------- | ------------- |
| 008/Books  | a, t  |  a, c, d, m |
| 008/Computer file  | m | - |
| 008/Maps  | e, f | - |
| 008/Music  | c, d, i , j | - |
| 008/Continuing Resource | a | b, i, s |
| 008/Visual Material | g, k, o, r | - |
| 008/Mixed Material | p | - |

So the code is now strictly safer, but I wonder if it might cause issues in practice. For example, the journal record you linked to above (https://library.villanova.edu/Find/Record/459806) will now be coded as "Serial" because technically we shouldn't try to distinguish other types of serial from the 008 because LDR/06 is 'm' rather than 'a'. It would be useful to see the impact on this from a re-index & I'd be happy to go back to the way it was, e.g. LDR/07='s' by itself seems like a strong enough indicator that 008 is probably coded as a serial

Also, there appears to be an issue with en.ini which is causing CI tests to fail. I can't see what's causing that, so pointers welcome

demiankatz:
@eocarragain, thanks for the latest adjustments; I'll do more reindexing and let you know what happens.

I see that you fixed the language problem before I could answer your question. For future reference, if you didn't already figure this out, you can automatically format the language files by using: `cd $VUFIND_HOME ; php public/index.php language normalize languages`

demiankatz:
@eocarragain, I should also say that I agree with your strategy of keeping electronic logic together in the isElectronic method.

One thought, though: we now have some method signature changes here; I think we should probably re-target this pull request against the dev-9.0 branch, since it is somewhat backward-breaking and may be too big of a change for inclusion in the minor 8.1 release. Any objections?

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -466,29 +608,35 @@ protected List<String> getFormatsAsList(Record record) {
...
+        String formatFromRecordType = getFormatFromRecordType(record, recordType, marc008, formatCodes007);
         if (formatFromRecordType.length() > 0) {
             result.add(formatFromRecordType);
         }
 
         // check the Leader at position 7
-        char bibLevel = Character.toLowerCase(leader.charAt(7));
         String formatFromBibLevel = getFormatFromBibLevel(
-            record, bibLevel, formatCode, marc008, couldBeBook
+            record, recordType, bibLevel, marc008, couldBeBook, formatCodes007
```
demiankatz, line 621: Is there a strong reason that you changed the position of the bibLevel parameter? If you did this for logical reasons (e.g. recordType is broader and bibLevel is more specific) then I have no objections to it, but if it just happened by chance, it might be worth swapping the order back to reduce the amount of change to the signature, to make adjustment of local customizations slightly less complicated.

eocarragain:
>  think we should probably re-target this pull request against the dev-9.0 branch, since it is somewhat backward-breaking and may be too big of a change for inclusion in the minor 8.1 release. Any objections?

None at all. That makes sense & may get more eyes on it too in the context of a major release. Done now (I think)

import/index_java/src/org/vufind/index/FormatCalculator.java
eocarragain: LDR/06 < LDR/07

So a bit esoteric. Happy to move it to the end for reasons you say.

demiankatz: I'll leave it to your discretion. I don't mind leaving it as-is for the reason you cite, especially since the signature is changing for other reasons regardless.

demiankatz:
Thanks, @eocarragain, it looks like you've retargeted the PR successfully!

demiankatz:
@eocarragain, I've finished another round of reindexing. Here is the impact: our number of Books has dropped from 2139205 to 2138290, Journals have dropped from 32493 to 32406, Serials have gone up from 16220 to 16307, Data Discs have gone up from 901 to 1602, and Audio (Non-Music) has gone up from 952 to 1132. Let me know if you want me to identify specific records involved in any of these changes!

Additionally, we normally identify electronic items based on a holdings location (in local custom code) instead of relying on the other rules. I temporarily added your new electronic rule to see what it would find, and it mostly added "Electronic" to books that have CDs in the back. I'm not sure if that's desirable or not. In any case, it won't matter to me, because in the long term, I'll just continue to use our local holdings-based approach.

import/index_java/src/org/vufind/index/FormatCalculator.java
eocarragain: let's leave it so, if that's ok. As you say the signature is already changing for other reasons.

demiankatz: No objections here -- thanks for talking it through!

eocarragain:
Thanks @demiankatz 

> Books has dropped from 2139205 to 2138290

That's quite a drop. Did you see a corresponding rise in eBook maybe? Data Set and Audio (Non-Music) may also account for some of this.

> Journals have dropped from 32493 to 32406, Serials have gone up from 16220 to 16307,

So that looks as if you had 87 records with the 008 coded for "Continuing Resource" but the LDR/06 is set to something other than 'a' (probably 'm'). Following the spec strictly we shouldn't look in 008 for these

> Data Discs have gone up from 901 to 1602

Did you mean Data Set? Is this just down to the fact that we're now identifying Data Sets based on LDR/06 = m. At least some of these probably were Books before

> Audio (Non-Music) has gone up from 952 to 1132

This most likely from add Non-music sound recording to definitelyNotBookBasedOnRecordType() so, again, these were probably going to books before. Does it seem accurate when you look at the records?

> mostly added "Electronic" to books that have CDs in the back. I'm not sure if that's desirable or not. 

I'm not sure either. I would have thought, from the spec, that putting 'm' rather than 'a' in LDR/06 should mean that the thing being described is primarily a computer file. Even so, it is probably OK to add Electronic in addition to another format for these cases to show that the item is partially electronic?

For Journals->Serials, and Electronic above, it seems to be a case of the reality of the data not matching the spec. In your case at least this hasn't caused a big change in numbers for Journals -> Serials. So it is a question of whether to follow the spec with a slight loss of detail for some non-standard records, or relax the extent that we follow the spec. Maybe your cataloguing colleague can give a steer? I'm happy to amend either way.

Is there anything outstanding that you're concerned about and would like me to dig in further at this stage? If not, I don't have other changes to make, so will wait for more feedback & happy to look again at that point.

demiankatz:
@eocarragain, thanks for the detailed responses. A couple points of clarification:

1.) Regarding the drop in Books, there is no corresponding increase in eBooks, so I think you're correct that this is reflected in the corresponding rise in Audio (Non-Music) and Data Discs.

2.) ...and speaking of Data Discs, I forgot that was a local change in our mappings. That's the label we use as the output mapping for the inputs of CDROM, ChipCartridge, DiscCartridge, FloppyDisk, Software, TapeCartridge, TapeCassette and TapeReel. (I suspect most of those categories are never found in our records at all, but it's a catch-all for software-like things).

My cataloging colleague is going to look through all my notes on remappings and odd records next week, so I'll get back to you once I've heard more from him. I'll also take some time (probably tomorrow) to check the accuracy of the "Audio (Non-Music)" records to see if there's any cause for concern there.

eocarragain:
great @demiankatz . Thanks a mil

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -261,26 +303,111 @@ protected String getFormatFromBibLevel(Record record, char bibLevel, char format
...
+            // Computer file
+            case 'm':
+                // All computer files return a format of Electronic in isElectronic()
+                // Only set more specific formats here
+                // Check 008/26 Type of computer file
+                switch (get008Value(marc008, 26)) {
+                    case 'a': // Numeric data
+                        return "DataSet";
+                    case 'b': // Computer program
+                        return "Software";
```
eocarragain, line 390: hi @demiankatz . Another query based on an internal discussion that I'd appreciate your advice on

- Until now Vufind only returned "Software" as a fall back in getFormatFrom007() where 007/0 = 'c' ("Electronric Reource") and no more specific type could be derived from 007/1 (see line 122 above). At NLI this caused issues as we have electronic things like digital images that we wouldn't call "software", so we had been mapping this to "Electronic" in translation_maps
- Now that we're parsing "computer file" in this PR we have things that you might more definitely want classed as "software", e.g. computer program, font, interactive multimedia
- From the translation_map perspective we can't map the 007 software to Electronic while leaving these computer file ones as "software"

Options to allow these things to be distinguished e.g.:

- Map the "computer file" ones (program, font, multimedia) to something else and add a new value to translation_map? ComputerProgram or something?
- Change the default in 007 (line 122 above) to something more generic. "Electronic" is the obvious one but then we're back to having "Electronic" set both in isElectronic and elsewhere. ElectronicResource? 

What do you think? We can handle locally if needed, but it does seem useful to be able to distinguish these things at translation_map level?

demiankatz: Yes, I agree that some differentiation here could be useful. Which approach is better probably depends on how much specificity we want. If we change "Software" to "ComputerProgram" in the `getFormatFromRecordType` logic, then I would argue that it might also make sense to remap the other values (font/interactive multimedia) to more specific values too.

If we don't want to get more granular there, then I would argue that changing the default up above to ElectronicResource is probably more appropriate.

Either way, we can map all the values we come up with to Software in the default translation map, so the end result is the same and we don't need to do extra i18n work.

eocarragain: Grand, if OK with then, I'll go the whole hog:

- for 007c default to ElectronicResource mapped to Software (for backwards compat, and sidesteps the isElectronic issue)
- ComputerProgram mapped to Software
- Font mapped to Software
- InteractiveMultimedia mapped to Software

Sound OK?

eocarragain: I pushed that change now @demiankatz 

In the end I left computer program as "Software" in the java code as they are really one and the same, at least more so that the others.

demiankatz: Great, works for me!

demiankatz:
@eocarragain, just a quick update to say that my collaborator in cataloging has gone through my analysis, and I think we're in pretty good shape here. I'm going to review his notes in more detail next week, and I expect that I will probably be able to approve and merge this at that time -- I just don't want to rush things toward the end of the week with limited time remaining. :-)

eocarragain:
Great @demiankatz thanks to you both!

demiankatz (approved):
Thanks, @eocarragain -- I think this is all set now! I appreciate all the time and effort put into improving this.

PR #2413, opened 2022-04-27
Make Blender's check for enabled backends more robust.

EreMaijala:
The old check caused a crash if Backends section was undefined.


demiankatz (approved):

demiankatz:
Thanks, @EreMaijala!

PR #2414, opened 2022-04-27
FOLIO: adjust renewal exception handling to cover server failures.

demiankatz:
If the Okapi server returned an unexpected response, renewals would fail in a strange way. This adjusts exception handling to fail more gracefully.


demiankatz:
@bbusenius / @todolson, I found the changes here helpful while troubleshooting a problem with FOLIO renewals. (Root cause turned out to be a missing permission, but the error handling wasn't quite right while the problem was occurring). Could one of you review and approve this when time permits? I recommend adding `?w=1` to the URL when viewing diffs to ignore whitespace; that should make the changes more readable.

jsavell (approved):

demiankatz:
Thanks, @jsavell!

PR #2417, opened 2022-05-03
[VUFIND-1550] Add support for star ratings.

EreMaijala:
This PR adds support for simple star ratings.

TODO:
- [x] Is bootstrap-rating good, or is there something better? no longer relevant
- [x] The requirement for a class-based glyphs, hot or not? no longer relevant
- [x] Is the UI good or bad?
- [x] Do we want the modal to include something like counts for different ratings? Might get complicated and prevents the login prompt.
- [X] What's with the icons not working if they are escaped in the attributes?
- [x] Do we want to limit rating by record format?
- [x] Postgresql support
- [x] Test Postgresql support
- [x] Update database changelog when merging
- [x] Componentize? star-rating.phtml + the styles would be a good candidate for a component. -- Not now.
- [x] Fix integration tests.
- [x] Resolve [VUFIND-1550](https://vufind.org/jira/browse/VUFIND-1550) when merging.
- [x] Display ratings in comments (and possibly allow adding/updating rating when posting a comment)


EreMaijala:
I'm going to try a pure css approach. I think that would allow using normal theme icons and would eliminate js and jQuery dependency.

EreMaijala:
Now modified to a js-less version based on Andrea Crawford's [Pure CSS 5-Star Rating](https://codepen.io/andreacrawford/pen/NvqJXW). I think this is much better, and this no longer shares anything with our local implementation.

EreMaijala:
Some more background information on design choices:
1.) This is decoupled from comments so that rating can be enabled separately, and comments don't get confusing as you shouldn't be able to provide more than one rating.
2.) I tried to keep the UI simple to not make it stand out too much, but still make it easy to add your own.
3.) Adding stars should be accessible by keyboard and screen readers (not really well tested though).
4.) Follows the same pattern comments and tags are implemented, but I just couldn't make the config setting the same "enabled"/"disabled" thingy.

demiankatz:
I made a number of mistakes while initially trying to test this with PostgreSQL, so I have deleted my original, potentially misleading comments.

Final outcome: all but two tests are passing, and both failures are unrelated to this PR. One of the failing tests is fixed by #2418, and the other is related to the known issue [VUFIND-1385](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1385). I'm confident that it's safe to check the "Test with Postgresql" box now.

demiankatz:
Some initial thoughts on the user interface:

* I was confused by the fact that, when there are no ratings set in the database, the rating shows as 5 stars by default (and also that when editing stars, they start out at 100% selected)
* I was initially drawn to click on the stars on the record page without opening the lightbox, and was confused when that had no effect (this is made worse by the "hover finger" mouse cursor effect)
* When I finally did find the lightbox -- which wasn't entirely obvious to me -- I clicked on the stars and expected my results to be saved immediately, so I closed the lightbox without hitting the Save button.

I should probably have known better about all of these things, but if I followed these instincts, it seems possible that others will too... I wonder if this could be streamlined at least a little bit more.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordActionsTest.php
```diff
@@ -496,6 +496,109 @@ public function testPrint(): void
...
+        // Go to a record view
+        $page = $this->gotoRecord();
+        // Check that rating is not displayed:
+        $this->unFindCss($page, 'div.rating');
+    }
+
+    /**
+     * Test star ratings on records.
+     *
+     * @retryCallback removeUsername2
```
demiankatz: I don't think you need this retryCallback since you're explicitly calling the method at the start anyway...

module/VuFind/src/VuFind/RecordDriver/AbstractBase.php
```diff
@@ -228,6 +228,46 @@ public function deleteTags($user, $tags)
...
+     * Returns an array with the following keys:
+     *
+     * rating - average rating (0-100)
+     * count  - count of ratings
+     *
+     * @param ?\VuFind\Db\Row\User $user User, or null for all users
+     *
+     * @return array
+     */
+    public function getRatingData(?\VuFind\Db\Row\User $user = null)
```
demiankatz: I don't think attaching database functionality to the record driver is a good idea -- I have regrets about this aspect of the initial design. I think it would be better to create an external class or service that manages ratings and takes the driver as an input, instead of embedding the logic here.

I suspect we're going to end up refactoring all of the record driver / database bindings when we migrate to Doctrine, so I'd prefer to follow a healthier pattern now rather than creating more technical debt for later.

That being said, if there's no more logical place to put this, I suppose I could be persuaded to live with it for now. I feel like there's some precedent for moving logic out of the record drivers, but I'm having trouble remembering what it is now. I can dig deeper if this doesn't sound familiar to you.

module/VuFind/src/VuFind/Controller/AbstractRecord.php
```diff
@@ -266,6 +266,44 @@ public function deletetagAction()
...
+
+        // Make sure ratings are enabled for the driver:
+        $recordHelper = $this->getViewRenderer()->plugin('record');
+        if (!$recordHelper($driver)->isRatingEnabled()) {
+            throw new ForbiddenException('rating_disabled');
+        }
+
+        // Save rating, if any:
+        if (null !== ($rating = $this->params()->fromPost('rating'))) {
+            $driver->addOrUpdateRating($user, intval($rating));
```
demiankatz: Should we apply range checking/validation to the value here (and/or in lower-level code) to prevent weird manipulation?

module/VuFind/src/VuFind/Db/Table/Ratings.php
```diff
@@ -0,0 +1,181 @@
...
+     */
+    public function deleteIfOwnedByUser(int $id, \VuFind\Db\Row\User $user)
+    {
+        // User must have an ID:
+        if (!isset($user->id)) {
+            return false;
+        }
+
+        // Rating row must exist:
+        $matches = $this->select(['id' => $id]);
```
demiankatz: Could you just include the user ID in the select here to save a step?

themes/bootstrap3/templates/record/addtag.phtml
```diff
@@ -20,4 +20,4 @@
   <div class="form-group">
     <input class="btn btn-primary" type="submit" value="<?=$this->transEscAttr('Save')?>"/>
   </div>
-</form
+</form>
```
demiankatz: This is clearly a bug fix -- I've just picked this one change directly to the dev branch.

demiankatz (changes requested):
This is off to a great start -- just a few initial thoughts!

themes/bootstrap3/templates/record/addtag.phtml
EreMaijala: Thanks! I was supposed to do that myself, but forgot.

module/VuFind/src/VuFind/RecordDriver/AbstractBase.php
EreMaijala: I flip-flopped around this a bit but decided to follow the pattern for comments and tags to make the Doctrine migration similar for the three, but if you feel like it's better to do more here, that's fine as well. I'm just not sure if it makes sense to do it when we know the migration is coming up.

module/VuFind/src/VuFind/Controller/AbstractRecord.php
EreMaijala: Yes, added to the lower level code.

module/VuFind/src/VuFind/Db/Table/Ratings.php
EreMaijala: Done. That was another copy-paste..

EreMaijala:
@demiankatz I fixed the issues and added some JS event handlers (quite generic, could be useful elsewhere as well) to make the use more intuitive.

module/VuFind/src/VuFind/RecordDriver/AbstractBase.php
demiankatz: I really thought we had done some work to refactor some of the database functionality away from the record drivers, but maybe that was wishful thinking; I still can't seem to find what I was thinking of. If I can't turn up a better precedent (and it's looking like I can't at this point), I think you're right that we should live with this for now and fix it in the Doctrine refactoring.

themes/bootstrap3/js/common.js
```diff
@@ -56,6 +81,27 @@ var VuFind = (function VuFind() {
...
+        if (elem.hasAttribute('data-click-callback')) {
+          return evalCallback(elem.dataset.clickCallback, event, {});
+        }
+      }
+    );
+    window.addEventListener(
+      'change',
+      function handleChange(event) {
+        let elem = event.target;
+        if (elem.hasAttribute('data-submit-on-change')) {
```
demiankatz, line 102: I wonder if some of this generic refactoring would benefit from being a separate PR... and I also wonder if this data-submit-on-change mechanism could be used for some of our other existing functionality, like the language/theme drop-downs which are also being refactored by #2378. I feel like there are a lot of balls in the air with templates/themes, and it's getting increasingly difficult to decide how to move forward. Don't let me make this worse through this suggestion, but I'm wondering if there's an opportunity to unify a couple of open threads...

EreMaijala: Yes, the idea here is that these could be used for any other functionality as well. I'm open to extracting this to a separate PR or incorporating into another, but you're right in that there's so much mid-flight at the moment that it's difficult to decide on how to proceed.

demiankatz: I'd be interested to hear what @crhallberg thinks about it... but for now, once time permits, I'll just test this out as-is and see if I have more feedback on the functionality before worrying about exactly how to refactor/commit the code.

demiankatz:
@EreMaijala, I think your latest changes make the UI significantly more intuitive. I still feel like it would be more clear if the stars were grayed out rather than filled in when there are no ratings, though. Do you feel differently? Is it possible we need a way to visually distinguish between "there is no rating yet" and either "0 stars" or "5 stars"?

EreMaijala:
@demiankatz Stars should be greyed out already when there are no ratings. I now also changed the texts a bit to make it nicer.

module/VuFind/src/VuFind/View/Helper/Root/TranslationEmpty.php
```diff
@@ -0,0 +1,59 @@
...
+namespace VuFind\View\Helper\Root;
+
+/**
+ * Helper to check if a translation is empty
+ *
+ * @category VuFind
+ * @package  View_Helpers
+ * @author   Ere Maijala <ere.maijala@helsinki.fi>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     http://vufind.org/wiki/vufind2:developer_manual Wiki
```
demiankatz: Looks like this was copy-pasted from something very old -- vufind2 should not be in the wiki link anymore.

module/VuFind/src/VuFind/View/Helper/Root/TranslationEmpty.php
```diff
@@ -0,0 +1,59 @@
...
+    use \VuFind\I18n\Translator\TranslatorAwareTrait;
+
+    /**
+     * Check if a translation is empty
+     *
+     * @param string|object $str String to translate
+     *
+     * @return bool
+     */
+    public function __invoke($str)
```
demiankatz, line 51: In the interest of not adding new code without adding tests, maybe we should give this one a unit test -- it would be almost trivial to do. :-)

demiankatz (changes requested):
A couple of pretty trivial notes based on your latest commits...

Also, one question: now, after I have rated an item, when I reopen the lightbox, it resets to no highlighted stars. Is this intentional? I would expect my rating to "stick" and was only expecting to see a blank display when I hadn't previously provided a rating.

module/VuFind/src/VuFind/View/Helper/Root/TranslationEmpty.php
EreMaijala: Fixed.

module/VuFind/src/VuFind/View/Helper/Root/TranslationEmpty.php
EreMaijala: Done!

EreMaijala:
@demiankatz Oops, I broke the display of user's current rating when refactoring. Now fixed.

ckaz:
Re: Community Call of June 7: @EreMaijala : Just went through the PR and the rating summary feature in particular looks good to me as regards accessibility.

EreMaijala:
Thanks, @ckaz!

demiankatz:
Thanks for the progress on this, @EreMaijala! Would you like me to re-test now, or should I wait until you've had time to look into the new "display ratings in comments" option?

EreMaijala:
@demiankatz I'm not expecting anything before I get the comments done. That doesn't prevent checking out the latest changes to the basic functionality, but can be done later. :)

crhallberg (approved):
The CSS-only star hover effect is very clever! Great find!

EreMaijala:
@demiankatz Rating when commenting is now implemented, so this is ready for another review. I've tried to ensure that rating is only added if the user explicitly chooses a rating so that the results don't get skewed if the user only adds a comment. It's still possible to explicitly select zero stars.

module/VuFind/src/VuFind/View/Helper/Root/Record.php
```diff
@@ -671,6 +671,17 @@ public function getLinkDetails($openUrlActive = false)
         return array_map($formatLink, $urls);
     }
 
+    /**
+     * Check if rating the record is enabled
+     *
+     * @return bool
+     */
+    public function isRatingEnabled(): bool
```
demiankatz: Do we really need this here? Since the main configuration is injected into every record driver as $this->mainConfig, would it be better to implement the config check as the default isRatingAllowed method? That might reduce the number of places where we need to inject this view helper, which feels a bit unnatural to me.

module/VuFind/src/VuFind/RecordDriver/AbstractBase.php
```diff
@@ -228,6 +235,53 @@ public function deleteTags($user, $tags)
...
+     *
+     * @param int $userId ID of the user posting the rating
+     * @param int $rating The user-provided rating
+     *
+     * @return void
+     */
+    public function addOrUpdateRating(int $userId, int $rating): void
+    {
+        // Clear rating cache:
+        $this->ratingCache = [];
```
demiankatz, line 299: Do you need to clear the whole rating cache here, or could you just update the value for the user directly?

module/VuFind/tests/unit-tests/src/VuFindTest/AjaxHandler/CommentRecordTest.php
```diff
@@ -143,17 +171,34 @@ public function testEmptyQuery()
      */
     public function testSuccessfulTransaction()
     {
-        $user = $this->getMockUser();
+        $user = $this->getmockBuilder(User::class)
```
demiankatz: Typo: getmockBuilder for getMockBuilder.

themes/bootstrap3/templates/record/addrating.phtml
```diff
@@ -0,0 +1,23 @@
...
+  <input type="hidden" name="source" value="<?=$this->escapeHtmlAttr($this->driver->getSourceIdentifier())?>" />
+  <div class="form-group">
+    <label class="control-label" for="add_rating"><?=$this->transEsc('rating_prompt')?>:</label>
+    <?=$this->render('Helpers/star-rating.phtml', ['ratingData' => $this->currentRating, 'inputAttrs' => ['data-submit-on-change']]);?>
+  </div>
+  <div class="form-group">
+    <input class="btn btn-primary" type="submit" value="<?=$this->transEscAttr('Save')?>"/>
+  </div>
+</form>
+<?php
```
demiankatz: Is there a reason for this random open tag at the end?

demiankatz (commented):
I just did another more thorough review of the code here and had a few suggestions. I'll do another round of hands-on testing once these have been discussed/addressed.

themes/bootstrap3/js/common.js
crhallberg: Would it be better to to query all elements with the data attributes and add event listeners to those elements instead of adding a top level window event listener? I'm worried about edge cases where you click on a span or a <b> inside of a link or button, etc.

themes/bootstrap3/templates/Helpers/star-rating.phtml
```diff
@@ -0,0 +1,31 @@
...
+  $inputAttrs = !empty($this->inputAttrs) ? (' ' . implode(' ', $this->inputAttrs)) : '';
+  $rating = $ratingData['count'] ? ($ratingData['rating'] ?? 0) : null;
+  // Unique name for the fields of read-only ratings so that displaying multiple ratings on a page works:
+  $fieldName = $this->readonly ? "rating_{$id}" : 'rating';
+?>
+<div class="star-rating<?=$this->readonly ? ' readonly' : ''?>">
+  <label class="rating__label hidden" for="rating_<?=$id?>_none" aria-label="<?=$this->transEsc('rating_none')?>"<?=$labelAttrs?>>
+    &nbsp;
+  </label>
+  <input class="rating__input hidden" name="<?=$fieldName?>" id="rating_<?=$id?>_none" value="" type="radio" autocomplete="off"<?=null === $rating ? 'checked' :''?><?=$inputAttrs?>>
```
crhallberg: It seems to me that this hidden element is to send a null for rating if it is still null. Since we [test if the rating is null before saving in the controller](https://github.com/vufind-org/vufind/pull/2417/files#diff-1a7fa9ee7e2ff955244ed90c58c64807c2ae2205ae588c498638bb193a594a6aR180), do we need this?

crhallberg: Perhaps we could turn this into a way to clear a previous rating.

themes/bootstrap3/templates/Helpers/star-rating.phtml
```diff
@@ -0,0 +1,31 @@
...
+  <label class="rating__label" for="rating_<?=$id?>_0" aria-label="<?=$this->transEsc('rating_0')?>"<?=$labelAttrs?>>
+    &nbsp;
```
crhallberg: Without any hover highlighting or any other visuals besides a pointer cursor, I find "click the empty space just to the left of the first star" to be a cryptic way to set 0 stars.

themes/bootstrap3/templates/Helpers/star-rating.phtml
```diff
@@ -0,0 +1,31 @@
...
+      $fieldId = "rating_{$id}_$level";
+      $rangeLow = $level;
+      $rangeHigh = 100 === $level ? 101 : ($level + 10);
+      $class = $level % 20 === 10 ? ' rating__label--half' : '';
+      $icon = $this->icon($level % 20 === 10 ? 'rating-half' : 'rating-full', 'rating__icon rating__icon--star');
+    ?>
+    <label class="rating__label<?=$class?>" for="<?=$fieldId?>" aria-label="<?=$this->transEsc("rating_$level")?>"<?=$labelAttrs?>>
+      <?=$icon?>
+    </label>
+    <input class="rating__input" name="<?=$fieldName?>" id="<?=$fieldId?>" value="<?=$level ?>" type="radio" autocomplete="off"<?=$rating >= $rangeLow && $rating < $rangeHigh ? ' checked' : ''?><?=$this->readonly ? ' disabled' : ''?><?=$inputAttrs?>>
```
crhallberg: 1. Could we abstract the label+input pair into a local function to prevent errors?
2. We can add `autocomplete=off` to the containing form if that helps simplify the HTML.

module/VuFind/src/VuFind/View/Helper/Root/Record.php
EreMaijala: You're right. Now simplified and much cleaner.

module/VuFind/src/VuFind/RecordDriver/AbstractBase.php
EreMaijala: It's not very likely that the cache would be hit during the request for update, so updating it here, in any usual case, is not worth it.

themes/bootstrap3/templates/record/addrating.phtml
EreMaijala: No, removed.

module/VuFind/tests/unit-tests/src/VuFindTest/AjaxHandler/CommentRecordTest.php
EreMaijala: Fixed.

themes/bootstrap3/templates/Helpers/star-rating.phtml
EreMaijala: WRT the change event, I don't think that a span will fire a change event. Binding to a queried list has the problem (that we have all sorts of workarounds in other places) that it needs to be done or re-done after any possible content has loaded. In this case it can be the lightbox content or comments tab.

The rating_none field is required for accessibility. Otherwise there's no indication of the current state. If you think it's not necessary, I can check if we could do without.

It's intentionally not possible to remove a previously given rating. You can revise the rating at any time, though.

EreMaijala: Actually, the rating_none field is crucial so that the stars after it don't get highlighted.

themes/bootstrap3/templates/Helpers/star-rating.phtml
EreMaijala: There's the subtle hover effect of no highlighted star at all, but I agree that it's not too clear. In the light of this, I decided to remove it altogether. I'd say that half-star as a minimum is more clear and obviously "something". What do you think?

themes/bootstrap3/templates/Helpers/star-rating.phtml
EreMaijala: Not sure I understand what this would prevent. Can you clarify?

crhallberg: 1. At the time, I thought it would simplify the code and prevent future errors modifying this code.
2. Just minimizing html.

EreMaijala: @crhallberg Sorry, I had completely lost track of this. 

1. I think the elements are suffiently different to warrant the two different cases.
2. Oops, looks like `autocomplete=off` isn't even supported for radio buttons. Now removed.

module/VuFind/src/VuFind/AjaxHandler/CommentRecord.php
```diff
@@ -142,6 +153,7 @@ public function handleRequest(Params $params)
                 self::STATUS_HTTP_BAD_REQUEST
             );
         }
+        $driver = $this->recordLoader->load($id, $source, false);
```
demiankatz, line 167: Is it worth making the handler aware of whether or not the ratings feature is turned on? Otherwise, we're going to do extra work retrieving records for no reason. Probably not a big deal since comments are rarely used, but just raising the point for consideration. Since we're already doing capability checks in the factory, it wouldn't be too hard to do.

demiankatz (commented):
One small thought, before I do some hands-on testing of this.

Also, are there any unresolved TODO checkboxes on here that you feel need further discussion before we can merge? (If everything is working, I'm happy to merge as-is and refine later, but I don't want to get ahead of things if there's more work that should be done here prior to merging).

module/VuFind/src/VuFind/AjaxHandler/CommentRecord.php
EreMaijala: It's intentionally left for the driver to tell if rating is allowed. That's because at least in Finna we need to leave some formats out.

EreMaijala: But if you feel like the Social/rating setting should work as a main switch, that could be added.

demiankatz: But if ratings are entirely disabled, do you need to check at all? I'm not suggesting that we should remove the driver check, just that maybe we don't need to load the driver at all if we know we will never have a rating.

demiankatz: (But I don't have really strong feelings about this; it's really a matter of weighing performance vs. complexity, and I'm not sure which is more important in this case).

EreMaijala: Driver's isRatingAllowed is now checked in several places and it keeps things clean. It also allows one to enable commenting e.g. only for Search2 records. A main switch would complicate the configuration and checks enough that I'd rather not do it.

demiankatz: Sounds good. I won't push this issue any further, because I don't think it's a big deal. :-)

demiankatz:
@EreMaijala, I did a little more testing and had some thoughts/observations:

1.) It might be nice to have a breakdown of ratings in the ratings pop up -- "Your Rating" at the top, and "Other People's Ratings" at the bottom, maybe grouped by number of stars with a count? If there's a reason that's difficult, it can certainly wait... but seems like it could be a nice future enhancement if nothing else.

2.) I wonder if we need a way to clear the rating -- if you accidentally click on it (for example, while writing a comment) there doesn't seem to be any way to undo. Maybe a small "clear" link next to the stars?

3.) Maybe not an issue we need to solve here, but I keep accidentally clicking on the thumbnail when I'm trying to click on the stars, and then it takes me to the image in the same browser tab. Should we open larger images in a lightbox or second tab? (Maybe that's a separate PR, but it's been annoying me today enough to mention it).

4.) I think I found a bug: if you add a comment without populating the rating control at all, it automatically shows up as 5 stars... but I don't think the 5-star rating is actually getting stored in the database. Is there a bug in the way we display comments without an accompanying rating?

EreMaijala:
@demiankatz 

1.) Not difficult other than deciding how to display it. It might e.g. be necessary to display them at the top, because buttons in the middle of a lightbox would be weird. So needs a bit of design.

2.) Added.

3.) I added a bit of top margin to leave more room between the elements. Is it better? How to display an image is another story. See e.g. https://www.finna.fi/Record/doria.10024_118689?lng=en-gb :)

4.) Should be fixed now. I think I've changed things so many times that I got confused about how they work. I hope it's all sorted now.

I also added a configurable (globally!) option for removing one's rating.

EreMaijala:
@demiankatz I just realized that I had added a note to the TODO list that displaying a breakdown in the lightbox doesn't quite work as it will interfere with the forced login flow. Maybe something like dropdown after current stars would be better.

demiankatz:
Thanks, @EreMaijala!

Regarding image display, the link that you shared doesn't seem to be working right for me -- I saw loading spinners in place of images for a very long time, and then the images eventually just went blank. Is it possible I don't have permission to see what you are trying to show me?

Regarding the star ratings, it seems to me that having the breakdown in the lightbox might be informative at the time that the user is making the rating (and also would make the lightbox feel less empty). Is there a reason we need to force login on the lightbox? Couldn't we always allow the lightbox to open, but replace "Your Rating" with a "log in to rate" link, like we do for some other features? That being said, my feelings about this aren't particularly strong, and if a drop-down for the full breakdown is significantly easier to implement, I have no objections to taking that approach instead.

module/VuFind/src/VuFind/Db/Row/Resource.php
```diff
@@ -161,6 +161,46 @@ public function addComment($comment, $user)
         return $row->id;
     }
 
+    /**
+     * Add or update user's rating for the current resource.
+     *
+     * @param int  $userId User ID
+     * @param ?int $rating Rating
```
demiankatz: Worth adding `(null to delete)` here?

module/VuFind/src/VuFind/Db/Row/Resource.php
```diff
@@ -161,6 +161,46 @@ public function addComment($comment, $user)
...
+            if (null === $rating) {
+                $existing->delete();
+            } else {
+                $existing->rating = $rating;
+                $existing->save();
+            }
+            return $existing->id;
+        }
+
+        $row = $ratings->createRow();
```
demiankatz, line 199: Should we add a safety to not create a row if `$rating` is null? Should we return null in that case?

module/VuFind/src/VuFind/View/Helper/Root/Config.php
```diff
@@ -122,4 +122,14 @@ public function alwaysDisplayIndexRecordInStaffView(): bool
         return (bool)($this->get('config')->Record
             ->alwaysDisplayIndexRecordInStaffView ?? false);
     }
+
+    /**
+     * Check if record ratings can be removed
+     *
+     * @return bool
+     */
+    public function allowsRemoveRating(): bool
```
demiankatz: Might this be a better fit for VuFind\Config\AccountCapabilities rather than the generic Config helper?

demiankatz (changes requested):
A few more minor comments based on a review of the latest code changes... I haven't done another hands-on test yet, since I expect there may be a few more updates coming soon based on ongoing discussion.

EreMaijala:
@demiankatz The image link worked for me before I sent it, but doesn't now. Will be investiged. This one really works: https://www.finna.fi/Record/siirtolaisuusinstituutti.271_m436?lng=en-gb

I'll revise the lightbox. It could be argued that previous ratings shouldn't affect the user posting a new one, but I'll just leave it here. :)

EreMaijala:
Breakdown implemented and everything done, I hope.

demiankatz (changes requested):
It looks like the Mink test may need adjustment; I'm getting this failure:

```
1) VuFindTest\Mink\RecordActionsTest::testRating
Element not found: .modal.in [name="username"] index 0

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:365
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordActionsTest.php:525
```

I suspect this is only a minor problem, so I'm going to go ahead and update the database changelog and do some more hands-on testing. I'll comment further if I run into any problems, so please consider no news to be good news. :-)

demiankatz:
@EreMaijala, I did find one small problem: the "clear" button under the Comments control doesn't seem to work correctly... at least, I can't figure out what it's supposed to do. If I change the rating when I submit a comment, it changes my rating everywhere (which is expected). If I clear the rating when I submit a comment, it appears to have no effect at all (the previous rating remains unchanged). Do we expect it to remove the rating? If not, then what is the purpose of the control? I guess it makes it possible to not submit a rating if you click on the stars accidentally, but it seems kind of misleading under other circumstances. If clearing the rating doesn't have any effect, maybe we should disable the "clear" control if a previous rating has been saved, or something like that.

I also note that the "remove_rating" setting has no impact on the clear control. This seems consistent with the thought that the purpose of this control is to prevent accidental submission of ratings (which I did suggest earlier!)... but it seems confusing in combination with the new option to remove ratings entirely.

EreMaijala:
@demiankatz The "Clear" link originally just resetted the start selection. I've now changed it so that if comment removal is enabled, clearing the stars will remove the rating. Added complexity comes from the fact that the "Clear" link needs to be removed if a rating is added and rating removal is disabled.

demiankatz (approved):
Looking good now; thanks for all the work on this, @EreMaijala!

PR #2418, opened 2022-05-04
Simplify the way "edit list" title is determined.

demiankatz:
This addresses a problem discovered while testing #2417: in PostgreSQL, new rows fetched from the database come with the ID pre-populated from the sequence, even if they are never saved. This interferes with correctly differentiating between "create row" and "edit row" titles. Fortunately, an existing flag offers an easier way to do this correctly.

Note that the underlying cause of the problem may still be a problem, since hitting "create list" repeatedly exhausts IDs unnecessarily. However, I'm not going to worry too much about that right now, since I'm hopeful that switching everything to Doctrine will likely change the situation anyway (see #2233).

TODO
- [x] Run full test suite with MySQL
- [x] Run full test suite with PostgreSQL

### DIFF
--- a/themes/bootstrap3/templates/myresearch/editlist.phtml
+++ b/themes/bootstrap3/templates/myresearch/editlist.phtml
@@ -1,6 +1,6 @@
 <?php
   // Set up page title:
-  $pageTitle = empty($this->list->id) ? 'Create a List' : "edit_list";
+  $pageTitle = $this->newList ? 'Create a List' : "edit_list";
   $this->headTitle($this->translate($pageTitle));
 
   // Set up breadcrumbs:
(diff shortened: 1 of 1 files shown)


EreMaijala (approved):

EreMaijala (changes requested):
The change is good, but should https://github.com/vufind-org/vufind/blob/ea04a81dcd215a053f1ae1af57984e7377533402/themes/bootstrap3/templates/myresearch/editlist.phtml#L22 be changed as well?

demiankatz:
Good catch, @EreMaijala -- I've made the suggested edit and re-run tests in both MySQL and PostgreSQL. I think all is well now, but I'll await your approval before merging.

EreMaijala (approved):

PR #2419, opened 2022-05-05
Redi-Resolver - Add missing array indexes to parseDOI return array and set service_type to getDOI

lahmann:
In template errors might show up when the DOI in Redi resolver results is shown as it does not containt `access` and `coverarge` index. This has been fixed plus a more specific value of `service_type` for the DOI.


demiankatz (changes requested):
@lahmann, thanks for the fix, but it looks like the corresponding unit test needs to be adjusted to match!

lahmann:
@demiankatz yes, thanks for noticing! I just pushed the adjusted tests.

demiankatz (approved):

PR #2420, opened 2022-05-05
[VUFIND-1493] Improve visibility of scheduled search feature

demiankatz:
This pull request makes the scheduled search functionality more visible by adding a link to the search result toolbar.

image

It also creates a "manage subscription" page where additional options could potentially be added in the future.

image

TODO
- [x] Select an icon for the "Subscribe to Search" / "Manage Search Subscription" links in the results toolbar
- [x] Address unattractive wrapping behavior with "Manage Search Subscription" link in results toolbar
- [x] Discuss content/behavior of manage subscription screen (Should we add a submit button instead of auto-submitting on change? Should we add information about email address? Should we add information about the search being saved?)
- [x] Add Mink test coverage
- [x] Resolve [VUFIND-1493](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1493) when merging.


demiankatz:
@crhallberg and @EreMaijala: this work was done in response to @livedvd's observation that the scheduled search functionality is currently pretty deeply hidden in the interface. This surfaces it a little bit more. The implementation so far is very much a proof of concept, with all decisions driven by getting it to work as quickly as possible. I'm interested in your thoughts on what the ideal workflow looks like, and what might be missing here, so we can tweak this into something release-worthy.

demiankatz:
I've begun work on making the search link smarter, so if you already have a subscription, it shows a different message. Unfortunately, the text gets a bit long, and we'll need to figure out how to deal with this problem:

image

EreMaijala:
@demiankatz I didn't quite understand what would lead to the "Manage Search Subscription" page. Would it also be used in the Search History page instead of the dropdown in the list?

Also, I'm not sure if I'd understand "subscribing" to a search, but that could be just me.. In any case, we need to make sure the terminology is consistent, as currently it's "Scheduled alerts" in the history page. 

Perhaps the texts could be streamlined a bit since the whole block is "Search Tools". Here are some possibilities for that come to mind (sorry, these are just a flow of ideas in no particular order):

Search Tools: * RSS Feed – * Email Link – * Save Search – * Enable Notifications

Search Tools: * RSS Feed – * Email Link – * Remove Saved – * Disable Notifications

The last link could also be unspecific to current state:

Search Tools: * RSS Feed – * Email Link – * Remove Saved – * Manage Notifications

Instead of saving searches we could also "add" and "remove" them from history:

Search Tools: * RSS Feed – * Email Link – * Add to History – * Manage Notifications

Perhaps, if the search has been saved, we could indicate that instead of a direct removal link, and allow further actions in a dropdown, like (dropdown items inside [ ]):

```
Search Tools: * RSS Feed – * Email Link – * Saved in History ˇ – * Notifications: Daily ˇ
                                            [Remove from History]   [Disable]
                                            [Show Search History]   [Daily]
                                                                    [Weekly]
                                                                    [Monthly]                                                                    
```

Or combine the "Saved in History" and Notifications into a single selection since notifications require the search to be saved anyway:

```
Search Tools: * RSS Feed – * Email Link – * Saved - Notifications: Daily ˇ
                                            [Remove from History]
                                            [Manage Notifications]
```
or:
```
Search Tools: * RSS Feed – * Email Link – * Saved - Notifications: Daily ˇ
                                            [Remove from History]
                                            [Notifications for new results: Daily]
                                            [Notifications for new results: Weekly]
                                            [Notifications for new results: Monthly]
```

For an unsaved search:

```
Search Tools: * RSS Feed – * Email Link – * Save to History ˇ
                                            [Save to History]
                                            [Save and Enable Notifications]
```

demiankatz:
@EreMaijala, to answer your question about what would lead to the "Manage Search Subscription" page, my thought is that for now, it is somewhat redundant -- just a mechanism for managing a single subscription so that we can link from search results to somewhere easier to understand than the entire search history screen. However, in the future, if we decide to add additional options associated with notifications, we could add more options here and then link to this from search history in place of the frequency drop-down. I'm not sure if we're likely to go in that direction, but I think it's useful to have the option, even if just to support local customizations.

Thanks also for all the thoughts about consistent wording and streamlining the search tools links. When time permits, I'll experiment with some of these and see which I prefer -- stay tuned! :-)

demiankatz:
Thanks again for the feedback, @EreMaijala. I've made a few adjustments based on your suggestions.

1.) I've changed "Manage Search Subscriptions" to "Manage Scheduled Alerts" for consistency of language.

2.) I've adjusted some existing Search Tools link language in en.ini to make it shorter ("Get RSS Feed" --> "RSS Feed" and "Email this Search" --> "Email Search"). This not only makes more room but also feels a little more clear to me (the unnecessary words get in the way). Obviously, we still have the problem of wrapping in other languages, but that is a pre-existing problem that already impacts just the basic options in some places (e.g. German). I think the best overall solution may be to restructure the markup of the Search Tools control entirely (to something like a `<ul>`), but maybe that doesn't need to be addressed as part of this PR.

3.) I've added plumbing so that we can display the current alert schedule setting directly in the Search Tools... so it can look like this:

image

...or like this:

image

I think that some of your more sophisticated suggestions involving drop-downs, etc., are interesting... but this seemed like the best approach I could come up with that didn't require adding new mechanisms. If @crhallberg or somebody else more front-end oriented wants to make this fancier, I think the necessary data now exists to enable such a thing.

Obviously this is still not done, but is it at least making more sense to you at this point?

demiankatz:
One more small tweak; I adjusted the m-dashes to n-dashes to make more room, and added an icon to the alert option:

image

demiankatz:
One last set of changes: I've added the same email address alert to the notification screen that we offer on the search history screen, and I've added logic to include the search query as part of the display for context. I'm sure we can make this look better (and potentially add more details drawn from the search object), but in terms of infrastructure, I think this is now fairly feature-complete.

image

EreMaijala:
@demiankatz Yes, this is much better and works for me.

module/VuFind/src/VuFind/Config/AccountCapabilities.php
```diff
@@ -149,6 +149,17 @@ public function getListTagSetting()
         return $this->config->Social->listTags ?? 'disabled';
     }
 
+    /**
+     * Get scheduled search setting ('enabled' or 'disabled').
+     *
+     * @return string
+     */
+    public function getScheduledSearchSetting()
```
EreMaijala: I know the other methods return 'enabled' or 'disabled', but could this still be more alike `isAccountAvailable` so that it returns a bool?

EreMaijala (changes requested):

module/VuFind/src/VuFind/Config/AccountCapabilities.php
demiankatz: Good point. I've simplified it.

demiankatz:
I've realized that the explanatory text on the MyResearch/ScheduleSearch page is no longer necessary because the email address info-box provides sufficient context, so I've removed that (it was always a bit out-of-context and potentially confusing). I'm feeling pretty good about this, but will await @crhallberg's review before locking it in with Mink test coverage.

crhallberg (approved):
What would be a good name for this icon for the Icon Helper? `search-alert`?

demiankatz:
Thanks, @crhallberg. Maybe `search-scheduled-alert` would be a better icon name, to make it more clear that this is not an alert message occurring during search.

EreMaijala (approved):

demiankatz:
Just for the record, @crhallberg, I merged this to dev-9.0 in commit 8f8931b6210b80c2def47c72e9aaa906cd637199 and opted for `search-schedule-alert` for the icon name (to make it more active).

PR #2422, opened 2022-05-09
[VUFIND-1554] Create generic mechanism for loading content with caching support

mtrojan-ub:
- (dev-9.0 because it's not urgent. Several tests seem to be failing in current dev-9.0 branch.)
- AwareInterface + Trait is optional
- Do we want to re-use this at some point before merging (e.g. in VuFind\Cover\CachingProxy)?

TODO
- [x] Add unit test coverage.
- [x] Resolve [VUFIND-1554](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1554) when merging.


EreMaijala:
The following are just my initial reactions, and subject to scrutiny and further discussion. :) Everything also depends on how generally useful we want to make this.

1.) I think it might be worth considering making this a special type of HTTP adapter that encapsulates the actual adapter and proxies all but the cacheable requests to the actual client. This would make it possible to use it anywhere desired without changing much in the existing code, just handing it a different client.

2.) The url alone as a cache key may not always be enough. I'd consider adding more of the request data (cookies, headers?) to the key.

3.) Why am I not seeing where addControllerCache is defined?

4.) Name it `Caching*`?

demiankatz:
Thanks, @EreMaijala, for the review! I'll wait to see how @mtrojan-ub responds before adding further comments of my own. One question for @mtrojan-ub, though: you mention failing tests in dev-9.0, but the build seems to be passing. Is it possible there is a problem in your local environment? (Feel free to respond to this question via email instead of on this PR, if that's easier -- we probably don't want to have a long conversation about test setup in the context of this PR).

mtrojan-ub:
Thanks a lot for the Feedback! 

1) I think i get the basic idea, but i don't understand entirely what you mean. Could you please specify the exact classes that should be part of it, especially the adapters that you mean?

2) We might be able to change the setClientOptions in a way that it adds the options to the cache key via a class variable and re-use this in the download function. But then we also have to consider that the setClientOptions method can be called multiple times, so we might have to implement a similar functionality like setOptions() and the $this->config array in the Laminas\Http\Client (or extend the client and add a public getter function).

3) Sorry for that! I simply forgot to add the Cache Manager changes... so i added this in the meantime and change the name a bit so that it matches the more generic approach of the downloader.

4) Why not! Then it matches the name of the existing CachingProxy... better. I will tend to that when time permits!

EreMaijala:
@mtrojan-ub,

1.) What I meant is that the HTTP Client always uses an adapter (anything that implements Laminas\Http\Client\Adapter\AdapterInterface) to do the actual request. A caching adapter could be one that would be used in place of the real adapter and would proxy the uncached requests to a real adapter. Now, I realize that this might be a bit complicated since AdapterInterface has connect, write, read and close which makes it more difficult to handle. Another option would be a descendant of \Laminas\Http\Client that would override the doRequest method and check for cached/cacheable request before passing it to the parent.

2.) I'd collect as much as possible of the details for the cache key when the request is being made so that they don't need to be tracked separately. setOptions is indeed a bit of an issue, but I think headers etc. would be more important.

All this said, I think we need to refactor everything that passes a Client around, and change it to something that consolidates HTTP requests to a single place (like VuFindHttp\HttpService) and abstracts the underlying classes used. Our current mechanism doesn't allow e.g. the proxy bypass list to work properly in many cases.

demiankatz:
Thanks, @EreMaijala, and I agree that reducing VuFind's knowledge of specific client objects is likely a good direction to move in, since the laminas-http library is essentially at end-of-life, and sooner or later we're going to need to replace it. Anything that makes that eventual project easier is a good thing! Perhaps that is an argument against building a custom client or adapter object, though, since that work might need to be redone sooner or later. If we can manage it a higher level of abstraction, it might be more future-proof.

EreMaijala:
@demiankatz You're right, and I think my proposal about an adapter wasn't up to snuff, and was particularly non-future-proof.

EreMaijala (changes requested):
I'll just flag this for requested changes. Please re-request review when done.

mtrojan-ub:
Thanks!

So i've taken care of the renaming in the meantime & added the client options to the cache key. As far as i understand how the discussion went, the proposal for higher level refactoring has been canceled again. Could you please have another look?

Of course, if you still see a necessity for major changes to the architecture we might as well discuss this in the upcoming community call.

module/VuFind/src/VuFind/Http/CachingDownloaderFactory.php
```diff
@@ -0,0 +1,72 @@
...
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options passed to factory.');
+        }
+
+        return new $requestedName(
+            $container->get(\VuFindHttp\HttpService::class)->createClient(),
```
EreMaijala: I think it would be better to just pass HttpService in the constructor and use its `get()` method. So, no need to create a client.

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * @var StorageInterface
+     */
+    protected $cache;
+
+    /**
+     * These result modes indicate whether the responses should be decoded.
+     */
+    public const RESULT_MODE_RAW = 'RAW';
+    public const RESULT_MODE_JSON = 'JSON';
```
EreMaijala: Why should the downloader know anything about the content? I'd find it better to keep it as simple as possible, so just return what you received. Also, if it was JSON-aware, it should allow decoding to an associative array as well.

module/VuFind/src/VuFind/Http/CachingDownloaderAwareTrait.php
```diff
@@ -0,0 +1,77 @@
...
+ * @category VuFind
+ * @package  Http
+ * @author   Mario Trojan <mario.trojan@uni-tuebingen.de>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Site
+ */
+trait CachingDownloaderAwareTrait
+{
+    /**
+     * Cache ID. This can be overridden by child classes if we want to use
```
EreMaijala: .. if *they* want to...

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,179 @@
...
+        // Return item if exists in cache
+        if ($this->cache->hasItem($cacheItemKey)) {
+            $body = $this->cache->getItem($cacheItemKey);
+            return $this->decodeBody($body, $resultMode, $url);
+        }
+
+        // Add new item to cache if not exists
+        $response = $this->client->setUri($url)->send();
+        if (!$response->isOk()) {
+            throw new \Exception('Could not resolve URL: ' . $url);
```
EreMaijala: This message could be misleading. Also, should probably catch exceptions as well.

EreMaijala (changes requested):

module/VuFind/src/VuFind/Http/CachingDownloader.php
mtrojan-ub: Initially this was just meant to be a comfort function. But, if the downloader knows about the content format it might also store the decoded version in the cache to save some performance on repeated downloads. This might also be useful to detect corrupt data as soon as possible and keep the memory usage of the cache low if we run into corrupt JSON, XML or any other format.

But that's just a thought, if you still think it's a better idea to keep it simple and remove this part, i would also be ok for me!

demiankatz: If you want that type of functionality, might it be a better design to allow passing in a callable for decoding or validation, rather than building specific modes into this class? Without looking deeper, I'm not sure if it would make more sense to inject some kind of support class via the constructor, or to pass something as an extra parameter to download.

Anyway, feel free to ignore this if it's unhelpful; I'm just brainstorming...

EreMaijala: I think @demiankatz's suggestion makes sense, and while JSON is common I'd rather avoid making it a special mode. Even so there could be default validation implementation for JSON. Trying to save anything doesn't seem feasible to me. The cache will serialize/unserialize the item as a string in any case. By default Laminas Cache uses the PhpSerialize serializer that uses the `serialize` and `unserialize` methods. Most benchmarks I've seen indicate that `serialize` is a bit slower than `json_encode`, but `unserialize` is markedly faster than `json_decode`, and faster decoding makes sense for a cache that's supposed to be hit several times to be of use.

EreMaijala (changes requested):

module/VuFind/src/VuFind/Http/CachingDownloader.php
mtrojan-ub: Okay, so i removed the `RESULT_MODE` and added callbacks to the download function that can be passed either for validation (before storing to cache) or decoding (after getting from cache).

I also added a default implemetation for JSON that uses PhpSerialize for validation.

Please have another look!

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,168 @@
...
+        callable $decodeCallback=null
+    ) {
+        $cacheItemKey = md5($url);
+        foreach ($params as $paramKey => $paramValue) {
+            $cacheItemKey .= '#' . $paramKey . '§' . $paramValue;
+        }
+
+        // Add new item to cache if not exists
+        if (!$this->cache->hasItem($cacheItemKey)) {
+            $response = $this->httpService->get($url, $params);
```
demiankatz: @EreMaijala had previously suggested catching exceptions here, though I'm not exactly sure why (since we're throwing other exceptions, is there a benefit to catching and rethrowing rather than just allowing existing exceptions through?). I didn't see a final resolution to that aspect of the earlier conversation, so I'm just putting in this comment to make sure we don't have any unresolved issues.

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,168 @@
...
+     * Constructor
+     *
+     * @param HttpService $httpService  HTTP service
+     * @param string      $cacheManager Base directory for cache
+     */
+    public function __construct(HttpService $httpService, CacheManager $cacheManager)
+    {
+        $this->httpService = $httpService;
+        $this->cacheManager = $cacheManager;
+        $this->setCacheId('downloader');
```
demiankatz: Would it make sense for the cache ID to be a constructor parameter? Since setCacheId is a public method, it seems strange to always initialize a cache with a hard-coded ID if there's a possibility that we will never use it. Maybe this would be cleaner with a "lazy initialization" pattern -- i.e. instead of creating the cache as a side effect of setCacheId, just store the cache ID in a property, and add a getCache() method that initializes the cache with the appropriate ID "on demand" -- then access the cache exclusively through the getter. Does that make any sense? setCacheId could simply update the ID property, and clear the cache property if the ID changes.

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,168 @@
...
+        $cacheName = $this->cacheManager->addDownloaderCache($cacheId);
+        $this->cache = $this->cacheManager->getCache($cacheName);
+    }
+
+    /**
+     * Download a resource using the cache in the background.
+     *
+     * @param string   $url              URL
+     * @param array    $params           Request parameters (e.g. additional headers)
+     * @param callable $validateCallback Callback for validation
```
demiankatz: These should be `?callable` to indicate that they're optional.

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,168 @@
...
+     * @return mixed
+     */
+    public function download(
+        $url,
+        $params=[],
+        callable $validateCallback=null,
+        callable $decodeCallback=null
+    ) {
+        $cacheItemKey = md5($url);
+        foreach ($params as $paramKey => $paramValue) {
```
demiankatz: Is there any reason not to use `http_build_query` here? Might be more concise. Also, any reason to include the parameters outside the md5 hash? I think including them in the hash would result in more consistent -- and shorter -- keys.

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,168 @@
...
+            }
+            $body = $response->getBody();
+            if ($validateCallback !== null && ($validateCallback($body) === false)) {
+                throw new \Exception('Invalid response body from URL: ' . $url);
+            }
+            $this->cache->addItem($cacheItemKey, $body);
+        }
+
+        $body = $this->cache->getItem($cacheItemKey);
+        if ($decodeCallback === null) {
```
demiankatz: You could make this a little more concise with `return ($decodeCallback === null) ? $body : $decodeCallback($body);`

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,168 @@
...
+     * @param array  $params Request parameters (e.g. additional headers)
+     *
+     * @return stdClass
+     */
+    public function downloadJson($url, $params=[])
+    {
+        $validateJson = function ($json) {
+            // Use PhpSerialize instead of json_decode for better performance
+            $serializer = \Laminas\Serializer\Serializer::factory(
+                \Laminas\Serializer\Adapter\PhpSerialize::class
```
demiankatz: Does this really work? Wouldn't you want to use the Json adapter to decode JSON? I don't understand how applying PHP unserialization to JSON data could work correctly.

module/VuFind/src/VuFind/ServiceManager/ServiceInitializer.php
```diff
@@ -92,6 +92,11 @@ public function __invoke(ContainerInterface $sm, $instance)
         if ($instance instanceof \VuFind\I18n\Translator\TranslatorAwareInterface) {
             $instance->setTranslator($sm->get(\Laminas\Mvc\I18n\Translator::class));
         }
+        if ($instance instanceof \VuFind\Http\CachingDownloaderAwareInterface) {
```
demiankatz: Is it worth adding "magic injection" support for this? I think this is a pattern Laminas is trying to move away from, so I'm not sure if we want to continue to embrace it with new functionality. (If you think it's important/useful, I won't object further... but in general I'd prefer to do more with factories and less with initializers). That's not to say that I wouldn't keep the associated Interface and Trait, as those are useful independently of the initializer.

demiankatz (changes requested):
Thanks, @mtrojan-ub -- I've left a few comments below.

One last question/thought: is there any logical place where we can use this in the existing code? It's great to add new features, but even better if we can make use of them so they don't get orphaned/broken by regressions. Perhaps this could be integrated into the cover loader, for example. I'm happy to deal with any integrations as a subsequent pass/PR if you're eager to get this work done and merged sooner... but it's something we should think about!

Also, we shouldn't forget about writing unit tests for this code, once the design is stable. I've added a TODO checkbox as a reminder.

module/VuFind/src/VuFind/Http/CachingDownloader.php
EreMaijala: There are actually two aspects here, and they both can be important for meaningful error log messages etc.:

1.) Since the download method never returns a response object, the exception thrown when a request fails should contain all necessary information for the caller to log and handle the error. One option would be a new exception class that would include the url, status code, response headers and body for a failed request, and another for when the HTTP service throws (like when unable to connect) that would include the url and the previous exception.

2.) You need to catch and re-throw to include the url in the exception message, but if the above can alleviate the need.

module/VuFind/src/VuFind/Http/CachingDownloader.php
EreMaijala: I think the callbacks should have access to the full response, or at least status code and headers in addition to body. This would mean they need to be cached as well.

module/VuFind/src/VuFind/Http/CachingDownloader.php
mtrojan-ub: I added lazy initialization, please let me know if it is ok like this now or you want me to add the initial ID as a constructor parameter as well.

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -113,13 +135,11 @@ public function download(
         callable $validateCallback=null,
         callable $decodeCallback=null
     ) {
-        $cacheItemKey = md5($url);
-        foreach ($params as $paramKey => $paramValue) {
-            $cacheItemKey .= '#' . $paramKey . '§' . $paramValue;
-        }
+        $cache = $this->getCache();
+        $cacheItemKey = md5(http_build_query(array_merge([$url], $params)));
```
demiankatz: I'm not sure if including the $url in the query makes sense; what about replacing `http_build_query(array_merge([$url], $params))` with `$url . http_build_query($params)`? I think that's a little simpler but should maintain the same degree of uniqueness.

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -90,20 +111,21 @@ public function __construct(HttpService $httpService, CacheManager $cacheManager
      *
      * @return void
      */
-    public function setCacheId($cacheId)
+    public function setCache($cacheId)
```
demiankatz: This is still setting the cache ID rather than the cache object. Should we keep the original method name of `setCacheId`? Or alternatively use a name like `resetCache` or `setUpCache` to make it clearer what is happening here. I think it's misleading to make this resemble a setter that mirrors `getCache`, since that's not what it is.

demiankatz (changes requested):
Thanks, @mtrojan-ub, this is looking better and better! I still have a couple suggestions. Also, it looks like there's still an open thread about exception handling that requires further input.

module/VuFind/src/VuFind/Http/CachingDownloader.php
mtrojan-ub: Since we use `VuFindHttp\HttpService` here: Should that kind of Exception instead be defined in `VuFindHttp\Exception` and just be re-used in higher levels like this? And should it wrap the whole Response object?

(And since HttpService seems to use cURL, we might also think about using CURLOPT_FAILONERROR so that explicit checking for isOk() is no longer necessary)

demiankatz: @mtrojan-ub, we can't assume that HttpService will use cURL -- it depends on the adapter setting in the [Http] section of config.ini.

EreMaijala: @mtrojan-ub if HttpService doesn't use the exception, I don't feel it belongs there.

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * @param string $url    URL
+     * @param array  $params Request parameters (e.g. additional headers)
+     *
+     * @return stdClass
+     */
+    public function downloadJson($url, $params=[])
+    {
+        $validateJson = function ($response) {
+            // Use PhpSerialize instead of json_decode for better performance
```
EreMaijala: Is this a stray comment?

EreMaijala (changes requested):

EreMaijala:
I understand the need for validation callback but I don't really like it. For the first request for a JSON response it means that the JSON needs to be decoded twice.

Is there a reason for using Laminas Serializer to decode the JSON, like a benchmark indicating that it's faster than pure json_decode?

module/VuFind/src/VuFind/Http/CachingDownloader.php
mtrojan-ub: OK, so i added a class `VuFind\Exception\HttpDownloadException` and tried to set additional properties depending on the error scenario. What do you think?

mtrojan-ub:
So, do you think we should use some kind of cache to avoid decoding it twice? 

/edit: There are multiple solutions... we could e.g. either store it to the existing cache before validation, or use a different cache just for the decoding, or remove the validation part entirely and remove something from cache if decoding fails... Or would you prefer something entirely different?

Regarding to json_decode, i think it would  be more Laminas-compliant to do things like this; however it seems that it will just be passed through to json_decode, see also:
https://docs.laminas.dev/laminas-json/basics/

demiankatz:
@mtrojan-ub, I don't think it's a problem to use `json_decode` directly -- since Laminas has broken up into components, I think there is a lot less pressure to "use Laminas for everything" -- I think the philosophy should actually be, most of the time, "use Laminas when you need to." Since Laminas has been deprecating and discontinuing various components, it's better not to use them unless there's strong utility. Obviously, we're very invested in Laminas ServiceManager and MVC, and I have no desire/expectation to change that, but everything else can probably be viewed as optional.

mtrojan-ub:
OK, so now json_decode is used directly. Also, i removed the validateCallback and stored the result to the cache after decoding so we do not have to decode multiple times. Does it look better now?

mtrojan-ub:
And @demiankatz: Could you please help me getting started with the unit test? I don't really understand how to create a proper `$cacheManager` mock object that won't fail during execution.

/edit: I added a test case in the meantime, but the question is which cases we actually want to cover (e.g. exceptions and so on?)

module/VuFind/src/VuFind/Exception/HttpDownloadException.php
```diff
@@ -0,0 +1,135 @@
...
+     * URL we tried to download.
+     *
+     * @var string
+     */
+    protected $url;
+
+    /**
+     * HTTP status associated with this exception.
+     *
+     * @var string
```
demiankatz: The type hints on these should have ? similar to the return types of the associated getters, since they could be null.

module/VuFind/tests/unit-tests/src/VuFindTest/Http/CachingDownloaderTest.php
```diff
@@ -0,0 +1,87 @@
...
+        $response->expects($this->once())->method('getBody')->willReturn($testBody);
+
+        $service->expects($this->once())->method('get')->with($testUrl)->willReturn($response);
+
+        // cacheManager
+        $storage = $this->getMockBuilder(\Laminas\Cache\Storage\StorageInterface::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $storage->expects($this->once())->method('hasItem')->willReturn(false);
```
demiankatz: Is it worth adding `->with()` clauses to these assertions?

demiankatz (changes requested):
Thanks, @mtrojan-ub, this is looking pretty good to me -- just a couple of very minor suggestions.

Let me know if you still have any questions about the unit testing. I do think it's probably worth creating a simple unit test for the new exception class; it should be pretty trivial, but if we're adding code we might as well cover it with tests!

I'm also interested in hearing @EreMaijala's thoughts on the latest iteration, when time permits.

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Constructor
+     *
+     * @param string $cacheManager Base directory for cache
+     */
+    public function __construct(CacheManager $cacheManager)
+    {
+        $this->cacheManager = $cacheManager;
+        $this->setUpCache('downloader');
```
EreMaijala: Is it intentional that this results in the cache being named `downloader-downloader`? I think something like 'default' (resulting in `downloader-default`) might make sense.

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Get cache and initialize it, if necessary.
+     *
+     * @return StorageInterface
+     */
+    protected function getCache()
+    {
+        if ($this->cache == null) {
+            $cacheName = $this->cacheManager->addDownloaderCache($this->cacheId);
```
EreMaijala, line 94: Since the name isn't really needed anywhere, would it make sense to name the method `getDownloaderCache` and make it return the cache directly?

EreMaijala (changes requested):
This looks pretty good to me now, just a couple of minor things about the cache handling.

module/VuFind/src/VuFind/Http/CachingDownloader.php
mtrojan-ub: Not sure if i understand correctly => Do you mean the method in VuFind\Cache\CacheManager itself? I tried to keep the name + usage similar to other existing methods, especially `addLanguageCacheForTheme`.

EreMaijala: That's what I meant, but following the existing pattern here makes sense.

EreMaijala (approved):

module/VuFind/src/VuFind/Exception/HttpDownloadException.php
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * "Format Unavailable" Exception
+ *
+ * @category VuFind
+ * @package  Exceptions
+ * @author   Mario Trojan <mario.trojan@uni-tuebingen.de>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+class HttpDownloadException extends \Exception implements HttpStatusInterface
```
demiankatz, line 41: Can we add a unit test for this new class, just to keep coverage up? Should be trivially easy. :-)

demiankatz (changes requested):
Thanks, @EreMaijala and @mtrojan-ub -- just two things before we merge this:

1.) I think we should add a simple test for the new exception class (see below).

2.) Is there anywhere in the existing code where we should apply this mechanism, so it's not just orphaned in the code base when we introduce it? Or do we plan to add something that leverages it in the near future? I'm happy to merge regardless of your answer, but if there's an easy place to integrate it, maybe we should do that here as a finishing touch before we merge.

module/VuFind/src/VuFind/Exception/HttpDownloadException.php
demiankatz: I'm also willing to help if nobody else has time, though I won't get to it right away.

mtrojan-ub: I added a simple unit test for `HttpDownloadException`.

mtrojan-ub:
> 2.) Is there anywhere in the existing code where we should apply this mechanism, so it's not just orphaned in the code base when we introduce it? Or do we plan to add something that leverages it in the near future? I'm happy to merge regardless of your answer, but if there's an easy place to integrate it, maybe we should do that here as a finishing touch before we merge.

Maybe when querying an external API that doesn't change every day, like e.g. in `VuFind\Content\Covers\Google`?

module/VuFind/src/VuFind/Exception/HttpDownloadException.php
demiankatz: @mtrojan-ub, it's good to test the exception handling logic of the caching downloader, but I had actually meant to create a VuFindTest\Exception\HttpDownloadException test which simply creates an exception and confirms that the getters return the expected values, so every new file has a corresponding test. As previously offered, if you don't have time for this, I can probably throw it together fairly quickly. :-)

demiankatz (changes requested):
Thanks again, @mtrojan-ub -- apologies about the apparent misunderstanding about unit tests.

Regarding the Google cover plugin as a potential proof of concept of this work, I agree that it looks like a promising possibility. Thinking about this a little further does raise an important question (and apologies if this is answered in the code or past conversation and I've simply forgotten): what controls the lifetime of this cache? Is there a possible use case where we have a single downloader cache but use it for multiple APIs for which we might want different cache lifetimes?

mtrojan-ub:
That could indeed be an important aspect. The consumer is already able to call the public function `setUpCache` if he wants to use any special cache ID, so i added another parameter `$cacheOptions` which can be used to override the `ttl` or maybe also other default options, if necessary. Would this be an appropriate solution?

demiankatz:
Thanks, @mtrojan-ub, I think your approach makes sense! So then it's just a question of the last test detail, and whether we want to do a proof of concept with the Google cover loader, or prefer to address that in a separate, follow-up PR (I'd lean toward doing it separately, so it doesn't hold up merging the basic functionality).

module/VuFind/src/VuFind/Exception/HttpDownloadException.php
mtrojan-ub: OK, so i added a separate HttpDownloadExceptionTest, hope it's better this time :smile:

mtrojan-ub:
The test has been added, and i would agree that doing the proof of concept in a separate PR might be better to keep a good overview (there have been quite a lot comments in this PR - but thanks a lot for your feedback!).

demiankatz:
Thanks, @mtrojan-ub, this is great! I made one small adjustment to the test (in assertions, the expected value should be the first parameter and the actual value the second, to ensure messages are accurate if something fails). I'll merge now, and we can do the follow-up with the Google cover handler when time permits. Let me know if you'd like to do that, or if I should put it on my own to-do list.

demiankatz (approved):

PR #2423, opened 2022-05-09
Fix PHP 8.1 deprecation issues.

EreMaijala:
I know we don't have official PHP 8.1 support in dev branch, but this PR fixes quite a few issues anyway (in anticipation of further tweaks in dev-9.0).

### DIFF
--- a/module/VuFind/src/VuFind/Session/AbstractBase.php
+++ b/module/VuFind/src/VuFind/Session/AbstractBase.php
@@ -106,7 +106,7 @@ abstract class AbstractBase implements HandlerInterface
      *
      * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function open($sess_path, $sess_name)
+    public function open($sess_path, $sess_name): bool
     {
         return true;
     }
@@ -117,7 +117,7 @@ abstract class AbstractBase implements HandlerInterface
      *
      * @return bool
      */
-    public function close()
+    public function close(): bool
     {
         return true;
     }
@@ -134,7 +134,7 @@ abstract class AbstractBase implements HandlerInterface
      *
      * @return bool
      */
-    public function destroy($sessId)
+    public function destroy($sessId): bool
     {
         $searchTable = $this->getTable('Search');
         $searchTable->destroySession($sessId);
@@ -153,6 +153,7 @@ abstract class AbstractBase implements HandlerInterface
      *
      * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
+    #[\ReturnTypeWillChange]
     public function gc($sessMaxLifetime)
     {
         // how often does this get called (if at all)?
@@ -177,7 +178,7 @@ abstract class AbstractBase implements HandlerInterface
      *
      * @return bool
      */
-    public function write($sessId, $data)
+    public function write($sessId, $data): bool
     {
         if ($this->writesDisabled) {
             return true;
@@ -193,5 +194,5 @@ abstract class AbstractBase implements HandlerInterface
      *
      * @return bool
      */
-    abstract protected function saveSession($sessId, $data);
+    abstract protected function saveSession($sessId, $data): bool;
 }
--- a/module/VuFind/src/VuFind/Session/Database.php
+++ b/module/VuFind/src/VuFind/Session/Database.php
@@ -48,7 +48,7 @@ class Database extends AbstractBase
      *
      * @return string
      */
-    public function read($sessId)
+    public function read($sessId): string
     {
         // Try to read the session, but destroy it if it has expired:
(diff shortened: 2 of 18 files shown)


module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/WorldCat/ConnectorTest.php
```diff
@@ -85,6 +85,8 @@ public function testGetHoldingsHttpFailure()
         $response = $this->createMock(\Laminas\Http\Response::class);
         $response->expects($this->any())->method('isSuccess')
             ->will($this->returnValue(false));
+        $response->expects($this->once())->method('getStatusCode')
```
demiankatz, line 88: I'm a bit curious what this change has to do with PHP 8.1... or why it's necessary to change a test when the corresponding class didn't change.

module/VuFind/src/VuFind/Session/AbstractBase.php
```diff
@@ -106,7 +106,7 @@ public function disableWrites()
      *
      * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
-    public function open($sess_path, $sess_name)
+    public function open($sess_path, $sess_name): bool
```
demiankatz, line 109: Correct me if I'm misremembering things, but I suspect that all of these added types are a backward-compatibility break since any class extending these classes will need to be adjusted to match. If that is the price of PHP 8.1 compatibility, it's probably worth it, but we'll definitely want a changelog note summarizing all of the changed classes (which I'm happy to write). Would you agree?

module/VuFind/src/VuFind/Session/Database.php
```diff
@@ -87,6 +87,7 @@ public function destroy($sessId)
      *
      * @return bool
      */
+    #[\ReturnTypeWillChange]
```
demiankatz, line 90: I see from the spec that as of PHP 7.1, this should be returning the count of rows deleted rather than true. Should we try to address that in a separate PR, or at least open a JIRA ticket about it?

demiankatz (commented):
Thanks, @EreMaijala -- see below for a few questions/comments.

Also, a broader comment: there is no conscious reason that the dev branch (and by extension VuFind 8.1) is not set up to support PHP 8.1, except for the fact that PHP 8.1 did not exist when the VuFind 8.x series was started. If you'd like to add 8.1 support to continuous integration as part of this PR (or make any other changes you deem necessary to better reflect 8.1 compatibility in the next VuFind release), I certainly would not object.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/WorldCat/ConnectorTest.php
EreMaijala: If `getStatusCode()` is not mocked, calling it will return null, and HttpErrorException's constructor expects an int instead. Without this change the test will cause the following deprecation warning:

`Deprecated: Exception::__construct(): Passing null to parameter #2 ($code) of type int is deprecated in /Users/eremaijala/src/VuFind2/module/VuFindSearch/src/VuFindSearch/Backend/Exception/HttpErrorException.php on line 98`

module/VuFind/src/VuFind/Session/Database.php
EreMaijala: A JIRA ticket is probably the only option, because the type in PHP 8.1 is a union `int|false` that's not supported pre-PHP 8.0.0. I'll open a ticket.

module/VuFind/src/VuFind/Session/AbstractBase.php
EreMaijala: True, and I agree.

module/VuFind/src/VuFind/Session/Database.php
EreMaijala: But you're right that the actual return value could be corrected right away. I'd do that in a separate PR.

EreMaijala: JIRA issue for the return type hint: https://openlibraryfoundation.atlassian.net/browse/VUFIND-1557

EreMaijala:
@demiankatz I'll add PHP 8.1 to CI.

module/VuFind/src/VuFind/Session/Database.php
demiankatz: Thanks!

demiankatz:
All tests are passing, and I've updated the changelog, so I'm merging this now. Thanks, @EreMaijala!

demiankatz (approved):

PR #2424, opened 2022-05-09
Some variables fixes found by phpstan

xmorave2:


module/VuFind/src/VuFind/Connection/Relais.php
```diff
@@ -87,8 +87,8 @@ protected function getDefaultData()
     /**
      * Format the parameters needed to look up an OCLC number in the API.
      *
-     * @param string $oclc   OCLC number to look up
-     * @param string $patron Patron ID (null to use default from config)
+     * @param string      $oclc   OCLC number to look up
+     * @param string|null $patron Patron ID (null to use default from config)
```
demiankatz: Would `?string` be a more concise way of representing this?

module/VuFind/src/VuFind/Connection/Wikipedia.php
```diff
@@ -418,7 +418,10 @@ protected function parseWikipedia($rawBody)
         $imageName = null;
         $imageCaption = null;
         // Check if data exists or not
-        if (isset($rawBody['query']['pages']['-1'])) {
+        if (isset($rawBody['query'])
```
demiankatz: Why is it necessary to do three levels of checking here? I thought isset was safe to call on deeply nested structures; if it is not, I'm sure there are many other places in the code where revisions would be necessary... but that seems to work against readability!

module/VuFind/src/VuFind/Db/Table/Session.php
```diff
@@ -137,9 +137,7 @@ public function writeSession($sid, $data)
     public function destroySession($sid)
     {
         $s = $this->getBySessionId($sid, false);
-        if (!empty($s)) {
-            $s->delete();
-        }
+        $s->delete();
```
demiankatz: getBySessionId is not guaranteed to return an object; I think we need to fix the return type on that function rather than change the logic here.

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
```diff
@@ -1042,11 +1042,11 @@ public function getAccountBlocks($patron)
      *
      * @return array An array with key-value pairs.
      */
-    public function getConfig($function, $params = null)
+    public function getConfig($function, $params = [])
```
demiankatz: Just curious if @EreMaijala has any concerns about the changes to this file.

module/VuFind/src/VuFind/ILS/Driver/Symphony.php
```diff
@@ -414,11 +414,11 @@ protected function getStatuses999Holdings($ids)
                     $this->translatePolicyID('LOCN', $result['current location']) :
                     $home_loc;
 
-                $available = (empty($curr_loc) || $curr_loc == $home_loc)
+                $available = ($curr_loc == $home_loc)
```
demiankatz: These changes seem to assume that $curr_loc can never be empty, but I'm not convinced that is true.

module/VuFind/src/VuFind/Search/EDS/Options.php
```diff
@@ -304,7 +304,7 @@ protected function setCommonSettings($searchSettings, $setting, $list, $target)
         if (isset($searchSettings->General->$setting)) {
             $userValues = explode(',', $searchSettings->General->$setting);
 
-            if (!empty($userValues) && isset($this->$list) && !empty($this->$list)) {
+            if (isset($this->$list) && !empty($this->$list)) {
```
demiankatz: Can we also eliminate the isset here since that is implied by the empty check?

module/VuFind/src/VuFind/Sitemap/Generator.php
```diff
@@ -374,7 +374,7 @@ protected function buildIndex($sitemaps)
                 }
             }
 
-            foreach ($sitemaps ?? [] as $sitemap) {
+            foreach ($sitemaps as $sitemap) {
```
demiankatz, line 377: Should we add a typehint to the argument above, to ensure this is really true?

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/Locale/LocaleDetectorFactoryTest.php
```diff
@@ -59,6 +59,7 @@ protected function getStrategyClasses()
     {
         $factory = new LocaleDetectorFactory();
         $strategies = $this->callMethod($factory, 'getStrategies', func_get_args());
+        $strategyClasses = [];
```
demiankatz: How about reducing this to `return array_map('get_class', $strategies);`?

demiankatz (changes requested):
Thanks, @xmorave2! I've left a few questions and comments.

module/VuFind/src/VuFind/Connection/Relais.php
xmorave2: I agree, fixed

module/VuFind/src/VuFind/Connection/Wikipedia.php
xmorave2: You are right, I overlooked bad types in doc comments, should be fixed now

module/VuFind/src/VuFind/Db/Table/Session.php
xmorave2: You are right, I changed the return type of getSessionById

module/VuFind/src/VuFind/ILS/Driver/Symphony.php
xmorave2: Again, there was missing return type for translatePolicyID I have overlooked

module/VuFind/src/VuFind/Search/EDS/Options.php
xmorave2: Sure

module/VuFind/src/VuFind/Sitemap/Generator.php
xmorave2: I agree, done

module/VuFind/tests/unit-tests/src/VuFindTest/I18n/Locale/LocaleDetectorFactoryTest.php
xmorave2: I am always for simplifying code. Done (with little help of `iterator_to_array`).

xmorave2:
Thanks @demiankatz for the review. I tried to address all your comments and made the changes as you suggest.

module/VuFind/src/VuFind/ILS/Driver/Symphony.php
```diff
@@ -417,8 +417,8 @@ protected function getStatuses999Holdings($ids)
                 $available = (empty($curr_loc) || $curr_loc == $home_loc)
                     || $result['circulate flag'] == 'Y';
                 $callnumber = $result['call number'];
-                $location   = $library . ' - ' . ($available && !empty($curr_loc)
-                    ? $curr_loc : $home_loc);
+                $location   = $library . ' - ' .
+                    ($available ? $curr_loc : $home_loc);
```
demiankatz: I'm still not sure I understand the motivation of this change. Isn't it still possible that $available could be true because $result['circulate_flag'] is 'Y' yet $curr_loc could be empty (say, because $result['current location'] was set to a blank string)? In this case, wouldn't we still want to fall back on $home_loc?

Of course, I am not familiar with Symfony or this code, so I don't know realistically what is likely or unlikely... but just looking at the code in a purely abstract way, this feels like a potential behavior change.

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
EreMaijala: I think the change is fine, but should be done for all ILS drivers.

module/VuFind/src/VuFind/Controller/Plugin/Captcha.php
```diff
@@ -127,7 +128,7 @@ public function verify(): bool
             }
         }
 
-        if (!$captchaPassed && $this->errorMode != 'none') {
+        if (isset($captcha) && !$captchaPassed && $this->errorMode != 'none') {
```
EreMaijala: This doesn't seem necessary. `active()` should always return false above unless there are captchas, and it can be argued that if this was active for any reason without captchas, the check should fail..

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
```diff
@@ -1083,13 +1083,13 @@ public function getConfig($function, $params = null)
      * @return bool True if the method can be called with the given parameters,
      * false otherwise.
      */
-    public function supportsMethod($method, $params)
+    public function supportsMethod(string $method, array $params = [])
```
EreMaijala: `$params` is not optional, so it shouldn't have a default value.

module/VuFind/src/VuFind/ILS/Driver/VoyagerRestful.php
```diff
@@ -351,11 +351,9 @@ protected function isHoldable($statusArray)
             $valid_hold_statuses_array
                 = explode(':', $this->config['Holds']['valid_hold_statuses']);
 
-            if (!empty($valid_hold_statuses_array)) {
-                foreach ($statusArray as $status) {
-                    if (!in_array($status, $valid_hold_statuses_array)) {
-                        $is_holdable = false;
-                    }
+            foreach ($statusArray as $status) {
```
EreMaijala, line 354: It would make sense to make the `isset()` check above a `!empty()` as well.

module/VuFind/src/VuFind/Search/EDS/Options.php
```diff
@@ -304,7 +304,7 @@ protected function setCommonSettings($searchSettings, $setting, $list, $target)
         if (isset($searchSettings->General->$setting)) {
```
EreMaijala: Would make sense to make this a `!empty()` as well.

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
demiankatz: Very good point, @EreMaijala. @xmorave2, do you think it would make sense to create a separate PR to adjust getConfig for all drivers, rather than making this one bigger?

EreMaijala (changes requested):

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
xmorave2: Done in #2429

demiankatz: Thanks, that looks good to me. I have merged it, so you can merge dev-9.0 into this branch whenever you're ready.

module/VuFind/src/VuFind/ILS/Driver/Symphony.php
xmorave2: Looking a bit more into it I think you are right, I changed it back.

module/VuFind/src/VuFind/Controller/Plugin/Captcha.php
xmorave2: That is true, but phpstan can't see it... So I removed `isset` for now and I am going to add `@phpstan-ignore-line`, instead of code change in future PR for phpstan I am working on (final update the level to **1**)

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
xmorave2: I removed the default value.

module/VuFind/src/VuFind/ILS/Driver/VoyagerRestful.php
xmorave2: Done.

module/VuFind/src/VuFind/Search/EDS/Options.php
xmorave2: Done.

demiankatz (approved):
This looks good to me now! I'll wait until next week when @EreMaijala is back and has an opportunity to re-check things before merging.

module/VuFind/src/VuFind/Controller/Plugin/Captcha.php
demiankatz: Maybe a more elegant solution to this would be to default $message to an empty value at the top of the function, add:

```
if (!$captchaPassed) {
    $message = $captcha->getErrorMessage();
}
```

inside the try..catch block, and check for `!empty($message)` instead of `!$captchaPassed` in the final if statement.

It appears that phpstan is correctly identifying a problem where if the CAPTCHA handler throws an exception, we will try to call a method on a non-object. This refactoring should fix that. It also gives us an opportunity to provide a generic failure message inside the catch block if we want to (which might be a good idea, though I'm not sure exactly what that message should say... it's essentially "CAPTCHA failed due to technical difficulties" or something).

demiankatz: (Perhaps you want to make that change in your level 1 PR rather than here, which is why I approved this PR in spite of this one unresolved issue).

xmorave2: Yes, I think it would be better to take care of this in next PR

EreMaijala (approved):

PR #2425, opened 2022-05-09
Add properties to PHPDoc of database row objects

xmorave2:
This is added only to some of db row objects to make phpstan level 1 happy - I am not going to add them everywhere as I know there is migration to Doctrine in progress.


demiankatz (approved):

demiankatz:
Thanks, @xmorave2, happy to merge this as a stopgap until we have more time to work on #2233.

PR #2426, opened 2022-05-09
Fix various phpstan level 1 issues

xmorave2:
Another step towards phpstan level 1

### DIFF
--- a/module/VuFind/src/VuFind/Controller/ContentController.php
+++ b/module/VuFind/src/VuFind/Controller/ContentController.php
@@ -75,21 +75,6 @@ class ContentController extends AbstractBase
     }
 
     /**
-     * Action called if matched action does not exist
-     *
-     * @return ViewModel
-     */
-    public function notFoundAction(): ViewModel
-    {
-        $response   = $this->response;
-
-        if ($response instanceof \Laminas\Http\Response) {
-            return $this->createHttpNotFoundModel($response);
-        }
-        return $this->createConsoleNotFoundModel($response);
-    }
-
-    /**
      * Get ViewModel for markdown based page
      *
      * @param string $page Page name/route (if applicable)
--- a/module/VuFind/src/VuFind/Controller/LibraryCardsController.php
+++ b/module/VuFind/src/VuFind/Controller/LibraryCardsController.php
@@ -54,29 +54,6 @@ class LibraryCardsController extends AbstractBase
             return $this->forceLogin();
         }
 
-        // Check for "delete card" request; parameter may be in GET or POST depending
-        // on calling context.
-        $deleteId = $this->params()->fromPost(
-            'delete',
-            $this->params()->fromQuery('delete')
-        );
-        if ($deleteId) {
-            // If the user already confirmed the operation, perform the delete now;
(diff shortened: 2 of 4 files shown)


module/VuFind/src/VuFind/Controller/ContentController.php
```diff
@@ -74,21 +74,6 @@ public function contentAction()
             : $this->notFoundAction();
     }
 
-    /**
-     * Action called if matched action does not exist
-     *
-     * @return ViewModel
-     */
-    public function notFoundAction(): ViewModel
```
demiankatz, line 82: Why is this method being removed? It is called by line 74 above!

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
```diff
@@ -54,29 +54,6 @@ public function homeAction()
             return $this->forceLogin();
         }
 
-        // Check for "delete card" request; parameter may be in GET or POST depending
-        // on calling context.
-        $deleteId = $this->params()->fromPost(
```
demiankatz, line 59: Similarly, it is not clear to me why this logic is being removed; I believe it is actually used.

demiankatz (changes requested):
Thanks, @xmorave2. Half of these changes make sense to me, and the other half do not... :-)

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
xmorave2: I don't think it is used - delete library card is its own controller action. Two methods `performDeleteLibraryCard` and `confirmDeleteLibraryCard` used in code below are not defined anywhere.

module/VuFind/src/VuFind/Controller/ContentController.php
xmorave2: But the method is defined also in parent `AbstractActionController` and the behaviour is the same as far as I know.

But we can of course use a safer approach and just call `parent::notFoundAction` instead of non-existant `createConsoleNotFoundModel` method...

module/VuFind/src/VuFind/Controller/LibraryCardsController.php
demiankatz: Ahh, you're right -- I had forgotten that this got moved to a separate controller.

module/VuFind/src/VuFind/Controller/ContentController.php
demiankatz: I see -- I had been looking inside VuFind code but forgot that this was defined by the framework. And I believe that createConsoleNotFoundModel is a relic of the old console code, before we switched to Symfony-Console. I'm not sure why this was ever necessary, but I think you're right to remove it. I'll do a bit of testing to be sure!

demiankatz (approved):
Thanks again, @xmorave2. You've convinced me that these changes are appropriate, and testing backs it up. I'm just going to leave this open a little longer in case @EreMaijala has anything to add, but I think we're in good shape.

EreMaijala (approved):
Looks good to me as well!

PR #2428, opened 2022-05-10
Do not suggest using AggressiveOpts; it is deprecated.

demiankatz:
Some comments in import-marc scripts suggested possibly using AggressiveOpts, but it is now deprecated; see: https://dzone.com/articles/jdk-13-what-is-aggressiveopts for details. This PR simply removes the obsolete suggestion.

### DIFF
--- a/import-marc.bat
+++ b/import-marc.bat
@@ -44,7 +44,6 @@ rem #   Tweak these in accordance to your needs
 rem # Xmx and Xms set the heap size for the Java Virtual Machine
 rem # You may also want to add the following:
 rem # -XX:+UseParallelGC
-rem # -XX:+AggressiveOpts
 rem ##################################################
 if not "!%INDEX_OPTIONS%!"=="!!" goto indexoptionsfound
 set INDEX_OPTIONS=-Xms512m -Xmx512m -DentityExpansionLimit=0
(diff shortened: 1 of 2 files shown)


demiankatz:
@xmorave2, I don't think this is a controversial change, but I need an approval to merge -- hoping you don't mind taking a quick look. :-)

xmorave2 (changes requested):
There are some more occurrences of `AggressiveOpts` in files **import/bin/import.bat**  and **import/bin/import.sh**

Otherwise, as you said, it is deprecated, we should not suggest to use it.

demiankatz:
Thanks, @xmorave2! I hadn't touched those scripts because I thought they were essentially vendor files coming from the SolrMarc distribution. Your comment inspired me to fix the problem upstream in SolrMarc, and I then discovered that the files do not exist there. I think they are just unused legacy scripts, so I have opened a separate PR (#2430) to remove them.  (That allows us to keep this harmless fix targeted against dev while making that slightly riskier change against dev-9.0). If you agree with this approach, please approve this PR so I can merge it. If not, I'm happy to discuss further!

xmorave2 (approved):

xmorave2:
@demiankatz: Removing unused code is always good. I do agree with this.

PR #2430, opened 2022-05-11
Remove obsolete import scripts.

demiankatz:
I thought that these scripts were part of the SolrMarc package, so I haven't touched them in a long time... however, it appears that they are just outdated files in VuFind that we no longer need (obsolete as of SolrMarc 3), so I am removing them. MARC import should be done using `$VUFIND_HOME/import-marc.sh`, not `$VUFIND_HOME/import/bin/import.sh`.

TODO


xmorave2 (approved):
Does make sense, thanks @demiankatz

PR #2432, opened 2022-05-16
KohaRest: Use correct datetime format and content type for hold suspe…

EreMaijala:
…nsion.

Koha 22.05 is stricter than before when validating the input. These changes make sure the requests are valid and also work with older Koha versions.

Koha requires the suspended_until to be an RFC3339 date+time, though currently it doesn't actually use the time part. It also expects the hold suspension requests to use `application/json` as the content type regardless of whether there are parameters. Older versions had the same requirements but had less strict validation.

The additional null coalescing operators avoid notices e.g. in case OPAC holds are disabled in Koha and it doesn't return any valid pick up locations.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
@@ -876,8 +876,8 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
                 if (empty($result['data'])) {
                     return [];
                 }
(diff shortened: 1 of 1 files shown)


xmorave2 (approved):
Does make sense to me.

demiankatz:
Thanks, @EreMaijala and @xmorave2!

EreMaijala:
Oops, I forgot to submit the PR adding the parameter that allows date converter to return a DateTime, so this doesn't actually work properly yet. I'm working on it.

PR #2436, opened 2022-05-16
Escape email address in search history alert.

demiankatz:


demiankatz:
@EreMaijala, I doubt there's much chance of problem characters here, but it seems better safe than sorry!

EreMaijala (approved):

PR #2438, opened 2022-05-17
Raise phpstan rule level to 1

xmorave2:
Now it checks:
* possibly undefined variables
* unknown magic methods and properties

TODO:
- [ ] Update changelog when merging (note change to SMS AbstractBase constructor)


module/VuFind/src/VuFind/Controller/Plugin/Captcha.php
```diff
@@ -51,24 +54,30 @@ class Captcha extends AbstractPlugin
 
     /**
      * String array of forms where Captcha is active
+     *
+     * @var array
```
demiankatz: Would `string[]` be more specific here, while we're making changes anyway?

module/VuFind/src/VuFind/Controller/Plugin/Captcha.php
```diff
@@ -113,27 +122,32 @@ public function verify(): bool
             return true;
         }
         $captchaPassed = false;
+        $message = '';
```
demiankatz: Maybe we should rename this variable to `$errorMessage` to be more explicit about what it is for. I think that makes the if statement at the end a little more immediately understandable.

module/VuFind/src/VuFind/I18n/Translator/LanguageInitializerTrait.php
```diff
@@ -51,7 +51,7 @@ protected function getTextDomains()
         $base = APPLICATION_PATH;
         $local = LOCAL_OVERRIDE_DIR;
         $languagePathParts = ["$base/languages"];
-        if (!empty($local)) {
+        if (strlen($local) > 0) {
```
demiankatz, line 54: What is the motivation for this change? We use `empty` to check for empty strings in quite a few places, and it seems like a reasonable approach as long as there is no realistic chance of other false-y values being included in the string. Is there a particular reason why phpstan doesn't like it here but isn't complaining about other instances?

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
```diff
@@ -1900,7 +1900,7 @@ public function handleCancelRequest($cancelDetails, $type = 'Hold')
                 'ns1:CancelRequestItemResponse/' .
                 'ns1:UserId/ns1:UserIdentifierValue'
             );
-            $itemId = $itemId ?? $requestId;
+            $itemId = empty($itemId) ? $requestId : $itemId;
```
demiankatz: Would `$itemId ?: $requestId` also have the desired effect here?

module/VuFindTheme/src/VuFindTheme/View/Helper/ConcatTrait.php
```diff
@@ -368,6 +368,7 @@ protected function outputInOrder($indent)
         $output = [];
         foreach ($this->groups as $group) {
             if (isset($group['other'])) {
+                /* @phpstan-ignore-next-line */
```
demiankatz: Do we have to ignore this because phpstan gets confused by calls from a Trait to methods undefined inside that Trait? Are we unable to use `@method` annotations to work around this? I just want to be absolutely sure there's no better option before we resort to ignore annotations.

demiankatz (changes requested):
Thanks, @xmorave2, this looks really good! Just a few minor suggestions and questions.

module/VuFind/src/VuFind/I18n/Translator/LanguageInitializerTrait.php
xmorave2: I couldn't find the real problem behind this. But I think, this is problem only for phpstan, as it looks on `$local` as it would be constant and thinks it is always not empty, but I am not sure why. Other question is, if we need the the check of emptiness at all as this is the only place in code, we are checking the LOCAL_OVERRIDE_DIR constant to not be empty...

module/VuFind/src/VuFind/ILS/Driver/XCNCIP2.php
xmorave2: Yes, I believe it's equivalent.

module/VuFindTheme/src/VuFindTheme/View/Helper/ConcatTrait.php
xmorave2: No, phpstan could analyse traits in context of classes using them. There is problem with incompatible signature of `itemToString` methods in HeadingLink and HeadingScript view helpers - one versus four arguments.

So it could probably be fixed, but with some refactoring of the trait and/or view helpers.

module/VuFind/src/VuFind/Controller/Plugin/Captcha.php
xmorave2: I will change it to `bool|string[]` as it could be also boolean value.

module/VuFind/src/VuFind/Controller/Plugin/Captcha.php
xmorave2: I'll change it.

module/VuFind/src/VuFind/I18n/Translator/LanguageInitializerTrait.php
demiankatz: I do believe this check is necessary, because we're creating a stack of paths to search, and LOCAL_OVERRIDE_DIR is optional, with empty string indicating that there is no local directory set (a rare but legal case). If phpstan insists on changing the check, there's nothing wrong with the proposed alternate approach, and it is more specific... but it seems to me that the prior code is also valid. I'm fine with leaving the change here, in any case.

module/VuFindTheme/src/VuFindTheme/View/Helper/ConcatTrait.php
demiankatz: Ahh, I see! Maybe a compromise would be to add a comment explaining why we're ignoring the next line, so we can address it in the future... e.g.:

```
/**
 * PHPStan doesn't like this because of incompatible itemToString signatures in HeadLink/HeadScript,
 * but it is safe to use because the extra parameters will be ignored appropriately.
 *
 * @phpstan-ignore-next-line
 */
```

Obviously, refactoring to solve the problem would be even better, but I can see how that might get complicated, and changing the base HeadLink/HeadScript helpers is not an option since they're coming from the framework.

xmorave2: Thanks @demiankatz. I added the comment.

demiankatz (approved):
Thanks, @xmorave2! Merging now.

PR #2439, opened 2022-05-17
Refactor NextItemToken emptiness check in XCNCIP2 ILS driver

xmorave2:
Some NCIP responders send '0' in next item token


demiankatz:
@xmorave2, is this really correct? Does the spec indicate that a next item token of 0 means there is no next item? I thought the token could be any arbitrary value -- might we run into a situation where one system returns 0 to indicate no items (perhaps erroneously/inconsistently with spec) and a different system uses 0 as a valid token?

(Sorry if this is all accounted for and I'm worried over nothing -- but it seems like it could be problematic).

xmorave2:
@demiankatz It is not correct in the meaning of NCIP standard - but there are NCIP responders with this behaviour. I refactor it a bit and am going to make this change in our local module.

demiankatz (approved):

demiankatz (approved):
Thanks, @xmorave2! I imagine we could also add a "treat_zero_as_empty_token" configuration to the .ini file in the future if that would be helpful... but this is fine to merge as-is for now.

PR #2440, opened 2022-05-18
Update Node.js and dependencies for CI. Use local eslint and jshint.

EreMaijala:
- Updates CI to use Node.js 16. This fixes slow installation due to old git protocol.
- Adds caching of NPM dependencies.
- Updates eslint and jshint and installs them locally.


demiankatz (approved):
Thanks!

PR #2441, opened 2022-05-18
Fix PHP 8.1 deprecation issues.

EreMaijala:
This is a set of dev-9.0 specific issues in addition to the ones fixed in dev.


demiankatz (approved):
I backported some fixes to the dev branch, and I'll merge the dev-9.0-specific ones here now!

PR #2442, opened 2022-05-18
Stop using error_log for connector cache debug messages.

EreMaijala:
Only uses logger if available to avoid dumping debug messages e.g. when running unit tests.

Error messages will still be logged.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2443, opened 2022-05-19
Remove Laminas\Soap from Shibboleth logout handler and add Mink test.

EreMaijala:
This removes dependency on laminas-soap, which is in security-only maintenance mode, and paves the way for several other dependency updates.

### DIFF
--- a/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ShibbolethLogoutNotificationTest.php
+++ b/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ShibbolethLogoutNotificationTest.php
@@ -1,0 +1,111 @@
+<?php
+/**
+ * Shibboleth logout notification test class.
+ *
+ * PHP version 7
+ *
+ * Copyright (C) The National Library of Finland 2022.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ShibbolethLogoutNotificationTest.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->findCss($page, 'table#recent-searches');
+
+        // Add a session id mapping to external_session table:
+        $sessionId = $session->getCookie('VUFIND_SESSION');
+        $table = $this->getTable(ExternalSession::class);
+        $table->addSessionMapping($sessionId, 'EXTERNAL_SESSION_ID');
+
+        // Call the notification endpoint:
+        $http = new \VuFindHttp\HttpService();
+        $result = $http->post(
```
demiankatz, line 100: Would it make sense to do this post twice -- once before updating the database and again after? That way we can confirm that it returns a bad status when the data is missing, and a good status when the data is present.

demiankatz (changes requested):
Looks good to me -- just one small suggestion!

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ShibbolethLogoutNotificationTest.php
EreMaijala: It won't reveal whether the request was successful or not, but returns OK always.

demiankatz: Ahh, okay, I see -- so it's validating the emptiness of the search history that confirms that it actually did something. I understand now!

demiankatz (approved):

PR #2444, opened 2022-05-19
Add cover provider for Koha local images.

demiankatz:
TODO
- [x] Determine if we need to use different parameters for sizes other than "small"
- [x] Perform real-world testing
- [x] Add unit test coverage


gmcharlt:
Regarding the first TODO, I can confirm that Koha's local cover images come in only two sizes with the following maximum dimensions applied when Koha scales the images upon upload:

- thumbnail: 140w x 200h
- full size: 600w x 800h

The image size limits aren't absolute, but the current scaling process guarantees that the image will be no wider than the maximum width. Upon skimming Koha's bug database, I don't see anybody advocating for a 'medium' size option.

demiankatz:
Thanks, @gmcharlt! Sounds like using "full size" for medium and large is probably reasonable, since the maximum isn't too extreme.

Are you in a position to confirm whether or not this code actually works with a real Koha system that uses cover images? If you are able to handle the hands-on testing, I can finish up by writing a unit test, and then we can get this feature merged and shipped. This has been stalled for a while for lack of a real-world tester.

gmcharlt:
Sure, I can commit to putting it through its paces this week with a test Koha system.

demiankatz:
Much appreciated, @gmcharlt!

module/VuFind/src/VuFind/Content/Covers/Koha.php
```diff
@@ -0,0 +1,83 @@
...
+    protected $url;
+
+    /**
+     * Constructor
+     *
+     * @param string $url Base URL for Koha covers
+     */
+    public function __construct(string $url)
+    {
+        $this->url = $url;
```
gmcharlt, line 63: I needed to add this to get it working (and state an expectation that anybody who is using Koha's local covers feature probably prefers that the results get cached.)

```
$this->supportsRecordid = true;
$this->cacheAllowed = true;
```

config/vufind/config.ini
```diff
@@ -915,6 +915,7 @@ verify_server_certificate = false
 ;       through AJAX at this time.
 ;   Note: Summon service takes a Serials Solutions client key, NOT Summon API key!
 ;   Note: For Orb, don't forget to complete [Orb] section.
+;   Note: For Koha, don't forget to add koha_cover_url below.
```
gmcharlt: And some additional documentation notes:

```
;      For Koha to be a source of covers, the LocalCoverImages and
;      OPACLocalCoverImages system preferences must be turned on and
;      local cover images uploaded to the Koha system. Koha emits only two
;      sizes of cover images, so medium and large are treated the same in VuFind.
```

gmcharlt (changes requested):
I was able to get it working with a couple changes.

gmcharlt:
@demiankatz For your own testing, please feel free to use `https://sandbox-2211.kohacatalog.com/cgi-bin/koha/opac-image.pl` in the short term. No guarantees that it will stay available permanently, but it has a record ID 1 with a cover image.

demiankatz:
@gmcharlt, thanks for the feedback and the test instance! I've incorporated your feedback, but based on your test image, I've adjusted this to use the smaller image for both "small" and "medium" and the full size only for "large". VuFind uses "medium" for search result thumbnails, and loading the full-size image for that seems excessive. If your image is representative of Koha's typical dimensions, it's more than adequate for VuFind's purposes.

gmcharlt:
@demiankatz The revisions work for me, including your tweak to the image size selection.

I can't speak to what's the most typical practice among Koha users for sizing the cover images; I suspect they're not typically as large as that picture of Luna the cat, but that picture is close to the maximum size that Koha currently permits.

You may be interested to know that I ran across a [small Koha bug](https://bugs.koha-community.org/bugzilla3/show_bug.cgi?id=33047) in the course of testing this. However, I don't think anything in VuFind will depend on the resolution of the bug, as it looks like the cover loader will treat a 500, a 404, or a 1x1 pixel image response the same.

demiankatz:
Thanks again, @gmcharlt! I've added a unit test to cover all the new code, and I've refactored slightly so that we now use a class property to determine which VuFind sizes map to Koha's thumbnail size -- this makes it easy to change the behavior through a custom subclass (and we could pretty easily add a config setting too, though I'm not inclined to add more noise in config.ini if nobody is asking for it).

EreMaijala (approved):
Looks good to me. This obviously won't work if records are harvested from multiple Koha instances, but if need arises, we could extend this to support getting the Koha cover URL from [datasources.ini](https://vufind.org/wiki/indexing:deduplication#sample_datasourcesini).

PR #2446, opened 2022-05-20
Fix PHP 8.x deprecation warnings.

EreMaijala:
* Consists mostly of fixes to null being passed to functions that take a string.
* ResultFeedTest changes ensure that translate() always returns a valid string instead of null (this caused null to be passed as format to the Dublin Core renderer).


demiankatz (approved):
Thanks, @EreMaijala!

PR #2447, opened 2022-05-26
Ajax Status Standardize and Modernize

crhallberg:
Wrote a [small AJAX queue handler](https://github.com/vufind-org/vufind/blob/80d6cd34b178730e901ff033487b05a81923c226/themes/bootstrap3/js/lib/status_ajax_queue.js) and applied it to item statuses, save list statuses, and record versions so that they all work the same and are easier to read.

**Do we want to fully remove JQuery?** I removed it in many places but left the AJAX and the record formatting mostly alone.

TODO
- [x] Finish replacing Hunt with ObserverManager.


themes/bootstrap3/js/lib/autocomplete.js
```diff
@@ -269,26 +269,26 @@
             return;
           }
           switch (event.which) {
-            case 9: // tab
```
crhallberg: It looks like eslint --fix changed the indentation of this file.

demiankatz: Maybe you should open a separate PR to just run `eslint --fix` on existing code so that we can commit these changes separately to avoid confusion. I'm not sure why the build is passing if eslint needs to make changes, so maybe we also need to adjust some CI settings....

demiankatz:
@crhallberg, please take a look at the latest alerts from jshint; it sounds like there are some features in use here that are not supported in ES6.

themes/bootstrap3/js/lib/autocomplete.js
crhallberg: This file is in the lib which isn't usually covered by eslint. Since I added a new lib file, I ran this folder but didn't notice these changes until after commit. I can revert them.

demiankatz: Is there a good reason not to check files in lib? If this is our code, it should meet our standards. If this is not our code, it should probably be moved to vendor.

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -1,232 +1,260 @@
-/*global Hunt, VuFind */
+/*global Hunt, StatusAjaxQueue, VuFind */
 
-VuFind.register('itemStatuses', function ItemStatuses() {
+VuFind.register("itemStatuses", function ItemStatuses() {
   function formatCallnumbers(callnumber, callnumber_handler) {
-    var cns = callnumber.split(',\t');
+    var cns = callnumber.split(",\t");
```
demiankatz: @crhallberg, many of these changes look like eslint style improvements too. Should we apply those as a separate commit/PR to make it easier to focus on the active changes here?

crhallberg: I don't think eslint cares about quotes. This is likely from my own formatter use. I didn't think much of it since I was drastically overhauling these files.

demiankatz: I think it's possible to set up eslint with quote preferences (at least, I'm pretty sure I remember seeing it switch quotes around in other projects), but maybe we haven't done that here. I don't have strong feelings -- this is just a huge amount of code to review, and I'm looking for anything that might reduce the mental load necessary to process it all. ;-)

In any case, since this is not yet urgent, I'll keep focusing on the 8.1 release items for now, and will circle back around to this one when there's a bit more free time to roll up my sleeves and dig in.

crhallberg: Agreed but there are no extraneous files in the PR after rolling back the eslint changes to js/lib/.

demiankatz: Yes, that's a good start! And speaking of js/lib/, did you see the thread above about lib vs. vendor?

module/VuFind/src/VuFind/AjaxHandler/GetRecordVersions.php
```diff
@@ -85,6 +85,26 @@ public function __construct(
         $this->tabManager = $tm;
     }
 
+    /**
+     * Load a single record and render the link template
+     *
+     * @param string $id     Record id
+     * @param string $source Record source
+     *
+     * @return array [response data, HTTP status code]
```
demiankatz: This returns a string, doesn't it?

module/VuFind/src/VuFind/AjaxHandler/GetRecordVersions.php
```diff
@@ -98,15 +118,18 @@ public function handleRequest(Params $params)
...
         $source = $params->fromPost('source') ?: $params->fromQuery('source');
-        $driver = $this->recordLoader->load($id, $source);
-        $tabs = $this->tabManager->getTabsForRecord($driver);
-        $full = true;
 
-        $html = ($this->recordPlugin)($driver)->renderTemplate(
-            'versions-link.phtml',
-            compact('driver', 'tabs', 'full')
-        );
+        if (gettype($id) != 'array') {
```
demiankatz: I think you can use `!is_array($id)` to make this more concise.

demiankatz (changes requested):
Just took a look at your PHP refactoring today and had a couple of small suggestions.

themes/bootstrap3/js/lib/autocomplete.js
crhallberg: I think it makes sense to do this but that should be another PR to avoid clogging this one again.

demiankatz: Agreed. I've opened [VUFIND-1566](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1566) so we don't forget to do this.

sturkel89:
@crhallberg - My testing shows that everything is working properly. I checked search results pages, saved lists, print output, and related items, and the expected information loaded correctly when using the Sample driver. 

(On the other hand, when I used the Demo driver, I saw all sorts of intermittent problems loading the location/holdings/call number info in each of those situations. So, intermittent issues that were noted before probably had more to do with irregularities in  the holdings data than any issue with the new code for AJAX.)

demiankatz:
Thanks, @sturkel89! When requesting testing assistance, I forgot that the Demo driver does not persist its randomly generated item statuses, which caused the inconsistencies that you encountered. Those strange behaviors were just a side effect of random holdings generation, and in no way related to the changes here. Apologies for any confusion!

In any case, now that the functionality has been fairly thoroughly tested, I'm going to give this another code review as soon as time permits (which will probably be sometime next week).

demiankatz:
@crhallberg, it looks like there are some new conflicts here. Can you resolve them when time permits? I wasn't sure whether or not the removal of explicit whitespace between icon and text in the HTML constructed in check_item_statuses.js was intentional, so I didn't want to resolve this myself without your input.

themes/bootstrap3/theme.config.php
```diff
@@ -245,6 +246,7 @@
             'status-pending' => 'FontAwesome:clock-o',
             'status-ready' => 'FontAwesome:bell',
             'status-unavailable' => 'FontAwesome:times',
+            'status-unknown' => 'FontAwesome:circle',
```
demiankatz, line 248: @crhallberg, why is this added here? Is this something that was accidentally omitted and needs to be restored? Is it the result of a merge error? It doesn't seem directly related to the rest of the PR, so I'm wondering what to do with it....

crhallberg: It was something that came up as an important missing option. I can put it elsewhere but I think this is a new status was introduced in this update.

demiankatz: I don't see any code that uses status-unknown in this PR... I do see code related to unknown statuses that uses the status-indicator icon, however. Is that a bug that needs correction, or am I missing something else? If that does need changing, I'd propose waiting until the rest of this work is completed and fixing that in a subsequent pass, just to avoid confusion.

crhallberg: It is something I intended to use for easier customization of [this part of the code](https://github.com/vufind-org/vufind/pull/2447/files/0e93724cc7cb6eb8f1eca9e90a6900af9b86de93#diff-6702ea8e4bf5c523d3dede9b16eabf9c02fbfd62f714e242cf993ebe3f16d26aR63) but had not added yet. There's much about this part of the code that irks me. Why are we checking for location only if the status is unknown?, for example.

demiankatz: I think the idea is that we only have a status when an item has a location. If it is available or unavailable, we can assume that the status has a location, since otherwise we wouldn't have been able to determine a binary status. When the status is unknown, it could either mean there is no location (in which case we don't know the status because the status is unknowable) or that something is wrong (we know where it is, but not what its status is). It's only useful to display the unknown status when there is an associated location. Does that make any sense? Maybe it's worth adding some code comments to clarify this.

crhallberg: I think the code could be clearer by putting the location check as an early escape on the loop.

demiankatz: I think the question is whether there are any scenarios where location is undefined but status is known. I don't _think_ that is likely, but I wouldn't swear to it without closer review.

demiankatz: In any case, unless you feel very strongly otherwise, we should probably get this code settled as-is before introducing further refactoring.

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -47,27 +51,24 @@ VuFind.register('itemStatuses', function ItemStatuses() {
       for (var x = 0; x < result.locationList.length; x++) {
         locationListHTML += '<div class="groupLocation">';
         if (result.locationList[x].availability) {
-          locationListHTML +=
-            '<span class="text-success">' +
-              VuFind.icon("status-available") + " " +
-              result.locationList[x].location +
-            '</span> ';
+          locationListHTML += '<span class="text-success">'
+                      + VuFind.icon("status-available")
```
demiankatz: Any idea why there is so much indentation here? Is it an eslint thing?

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -47,27 +51,24 @@ VuFind.register('itemStatuses', function ItemStatuses() {
...
+                      + result.locationList[x].location
+                      + '</span> ';
         } else if (typeof(result.locationList[x].status_unknown) !== 'undefined'
                   && result.locationList[x].status_unknown
         ) {
           if (result.locationList[x].location) {
-            locationListHTML +=
-              '<span class="text-warning">' +
-                VuFind.icon("status-indicator") + " " +
```
demiankatz, line 61: I think this was the only use of the status-indicator icon. Should we remove it or flag it as deprecated?

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -88,145 +89,141 @@ VuFind.register('itemStatuses', function ItemStatuses() {
...
+      delay,
+    });
+  }
 
   //store the handlers in a "hash" obj
   var checkItemHandlers = {
-    'ils': ItemStatusHandler,
-    'overdrive': OdItemStatusHandler,
+    ils: makeItemStatusQueue(),
+    overdrive: makeItemStatusQueue({ url: "/Overdrive/getStatus" }),
```
demiankatz, line 175: We don't have a way to test this. Should we try to solicit help before merging?

themes/bootstrap3/less/components/search.less
```diff
@@ -209,6 +209,11 @@ header .container.navbar { margin-bottom: 0; }
   display: none;
 }
 
+.savedLists__ul {
```
demiankatz: Is this style change directly related to other changes, or something that got tweaked along the way?

themes/bootstrap3/js/check_item_statuses.js
demiankatz: Related note: don't we need to add status-unknown to js-icons.phtml in order for it to work?

demiankatz (changes requested):
@crhallberg, I took a closer look at some of the files here and had a few more questions and observations.

demiankatz:
@crhallberg, with the merging of #2569, there are now some conflicts here that need to be resolved. Can you address them?

@LuomaJuha, once conflicts are resolved here, would you like to finish the ObserverManager work as part of this PR, or would it be easier to do that separately and merge it into this?

demiankatz:
@crhallberg, thanks for the progress here, but please note that there seem to be a bunch of unresolved comments from my previous review. Have you had a chance to look at any of those? (No rush, but I don't want to re-review if the previous review is not yet fully resolved).

themes/bootstrap3/js/check_item_statuses.js
crhallberg: Yes, please!

themes/bootstrap3/js/check_item_statuses.js
crhallberg: Our linters don't seem to mind sane indenting now.

LuomaJuha:
@demiankatz Hmm, I think I could do it into this same pr.

demiankatz:
Thanks, @LuomaJuha. If I'm not mistaken, I think @crhallberg may have already done it here at this point... but if you have time to review/test, your input would be welcome! (I also plan to take a closer look at this soon, time permitting).

LuomaJuha:
@demiankatz @crhallberg Oh, thats neat. Well, I can try it out with our theme and tell what I can find.

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -88,147 +89,143 @@ VuFind.register('itemStatuses', function ItemStatuses() {
...
 
     if (VuFind.isPrinting()) {
-      checkItemStatuses(container);
-      return;
+      checkAllItemStatuses(container);
+    } else {
+      VuFind.observerManager.createIntersectionObserver(
+        'itemStatuses',
+        checkItemStatus,
+        Array.from((container).querySelectorAll('.ajaxItem'))
```
LuomaJuha: was just wondering if the array.from is redundancy in here as it should work also with nodelists.

themes/bootstrap3/js/check_item_statuses.js
demiankatz: I'm not sure if @bpalme is in a position to test this PR, but he'd be the logical person to start with...

themes/bootstrap3/js/check_item_statuses.js
crhallberg: That's great to know! I was just following the `toArray()`s I was seeing, but I guess that's just for jQuery!

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -88,147 +89,143 @@ VuFind.register('itemStatuses', function ItemStatuses() {
...
 
     if (VuFind.isPrinting()) {
-      checkItemStatuses(container);
-      return;
+      checkAllItemStatuses(container);
+    } else {
+      VuFind.observerManager.createIntersectionObserver(
+        'itemStatuses',
+        checkItemStatus,
+        (container).querySelectorAll('.ajaxItem')
```
demiankatz: I don't think the parentheses around `(container)` serve any purpose here. Worth simplifying?

demiankatz (changes requested):
One small question...

I haven't done deeper testing yet, as I thought I'd wait and see what @LuomaJuha finds first. (But if @LuomaJuha is finished or doesn't currently have time, just let me know and I can give this more priority on my end).

themes/bootstrap3/js/check_item_statuses.js
LuomaJuha: I'm gonna test this out soon as I get my cup of coffee and check if there is something to be done.

One thing I'm not sure about is the removal of check: checkItemStatuses, checkRecord: checkItemStatus in check_item_statuses.js return part. If somebody is using them directly, it might come as a suprise that those are removed. Should they be marked as deprecated first and be removed later?

themes/bootstrap3/js/openurl.js
```diff
@@ -58,12 +58,13 @@ VuFind.register('openurl', function OpenUrl() {
...
+    if (VuFind.isPrinting()) {
       container.find('.openUrlEmbed.openUrlEmbedAutoLoad a').trigger('click');
     } else {
-      new Hunt(
-        container.find('.openUrlEmbed.openUrlEmbedAutoLoad a').toArray(),
-        { enter: embedOpenUrlLinks }
+      VuFind.observerManager.createIntersectionObserver(
+        'openUrlEmbed',
+        embedOpenUrlLinks,
+        Array.from(container.querySelectorAll('.openUrlEmbed.openUrlEmbedAutoLoad a'))
```
LuomaJuha: The container is transformed into a jquery object in line 46. Should the jquery be dropped from this init function?

themes/bootstrap3/js/doi.js
```diff
@@ -49,12 +49,13 @@ VuFind.register('doi', function Doi() {
   function init(_container) {
     var container = _container || $('body');
```
LuomaJuha: Here is also a reference to jquery object.

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -88,147 +89,143 @@ VuFind.register('itemStatuses', function ItemStatuses() {
...
+  function checkAllItemStatuses(container = document) {
+    container.querySelectorAll(".ajaxItem").forEach(checkItemStatus);
   }
-  function init(_container) {
-    var container = typeof _container === 'undefined'
-      ? document.body
-      : _container;
+
+  function init($container = document) {
+    const container = $container instanceof Node ? $container : $container[0];
```
LuomaJuha: Was just thinking should there be a global function, which would check if the object is instanceof jquery and return the 0 index element if true. It could make the transition from jquery to native js a littlebit smoother and replacing them later would be nicer?

demiankatz:
Thanks for the review, @LuomaJuha! I'll let @crhallberg take it from here as soon as time permits. :-)

themes/bootstrap3/js/openurl.js
crhallberg: It was a mixed case. I made it all jQuery now, for minimal diffs.

themes/bootstrap3/js/doi.js
crhallberg: jQuery removed here.

demiankatz:
@LuomaJuha, do you mind giving the latest updates a look so you can approve the PR if your feedback is fully addressed? I'll also take one last look once you're satisfied. :-)

themes/bootstrap3/js/openurl.js
```diff
@@ -64,7 +64,7 @@ VuFind.register('openurl', function OpenUrl() {
       VuFind.observerManager.createIntersectionObserver(
         'openUrlEmbed',
         embedOpenUrlLinks,
-        Array.from(container.querySelectorAll('.openUrlEmbed.openUrlEmbedAutoLoad a'))
+        container.findAll('.openUrlEmbed.openUrlEmbedAutoLoad a').toArray()
```
LuomaJuha: the findAll() function is a new one to me?

LuomaJuha:
@demiankatz sorry, took a while until I noticed your message. Found only one odd function reference but otherwise nothing extra odd here.

demiankatz:
Thanks for the catch, @LuomaJuha! I'll give this one more test/review run when time permits, and then hopefully we can get it merged.

demiankatz (changes requested):
@crhallberg, I just ran the test suite on this branch to see if everything is passing, but I'm running into this:

```
There was 1 failure:

1) VuFindTest\Mink\ListViewsTest::testFavoritesInTabMode
Element not found: #information_cd588d8723d65ca0ce9439e79755fa0a-content .savedLists ul index 0

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:365
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ListViewsTest.php:118

FAILURES!
Tests: 2192, Assertions: 11221, Failures: 1, Skipped: 1.
```

Any ideas?

demiankatz:
@crhallberg, I've merged the latest dev branch into this and retried the tests, but I'm still getting the same failure:

```
1) VuFindTest\Mink\ListViewsTest::testFavoritesInAccordionMode
Element not found: #information_cd588d8723d65ca0ce9439e79755fa0a-content .savedLists ul index 0

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:365
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ListViewsTest.php:155

FAILURES!
Tests: 2213, Assertions: 11341, Failures: 1.
```

I'm also not entirely clear on why you added a clickCss in commit 49a1cb626c65c3819ee901c9ebf9992efc900d2a -- I don't think you should have to click anything for the save status to show up, but maybe I'm misunderstanding the context (I honestly didn't spend too much time digging deep).

Can you take another look at this when time permits?

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ListViewsTest.php
```diff
@@ -153,7 +153,7 @@ public function testFavoritesInAccordionMode()
         // Check saved items status
         // Not visible, but still exists
         $this->clickCss($page, '#information_cd588d8723d65ca0ce9439e79755fa0a');
-        $this->snooze();
+        $this->waitForPageLoad($page);
```
crhallberg, line 158: Didn't know we had access to this! Much preferred.

themes/bootstrap3/js/check_item_statuses.js
demiankatz: I was able to use the new Overdrive test harness from #2707 to confirm that this is working correctly.

demiankatz (approved):
The full test suite is passing, and all of my manual testing seems to work as well (including Overdrive support).

I did discover a significant problem while testing this: OpenURL embedding doesn't seem to work right while the default content security policy is enforced. This will require more investigation, which I plan to do separately. However, I have confirmed that the bug is no different here than in the dev branch, so I think we can merge this without impacting that other issue.

Since this is a big change, I'm going to ask @EreMaijala to give it one last look as well before I merge.

demiankatz:
UPDATE: Please disregard the OpenURL problem I reported earlier. I see that this behavior is fully covered by Mink tests, which have been passing all along. It also works when I manually test it. I'm not sure what I was doing wrong earlier, but I can no longer seem to reproduce the issue. It's possible I just had the wrong expectations!

module/VuFind/src/VuFind/AjaxHandler/GetRecordVersions.php
```diff
@@ -85,6 +85,26 @@ public function __construct(
...
 
+    /**
+     * Load a single record and render the link template
+     *
+     * @param string $id     Record id
+     * @param string $source Record source
+     *
+     * @return string
+     */
+    protected function getVersionsForRecord($id, $source)
```
EreMaijala: I'd call this `getVersionsLinkForRecord`.

module/VuFind/src/VuFind/AjaxHandler/GetRecordVersions.php
```diff
@@ -98,15 +118,18 @@ public function handleRequest(Params $params)
...
+
+        $htmlByRecord = [];
+        for ($i=0; $i < count($id); $i++) {
+            $key = $source[$i] . '|' . $id[$i];
+
+            $htmlByRecord[$key] = $this->getVersionsForRecord($id[$i], $source[$i]);
+        }
 
-        return $this->formatResponse($html);
+        return $this->formatResponse($htmlByRecord);
```
EreMaijala: Let's return the data in a named array like `$this->formatResponse(['records' => $htmlByRecord]);`. This makes it possible to extend the output without breaking back-compatibility.

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -88,147 +89,143 @@ VuFind.register('itemStatuses', function ItemStatuses() {
...
-        if ('done' === this.items[id].state) {
-          displayItemStatus(this.items[id].result, el);
-        } else {
-          this.items[id].elements.push(el);
-        }
-        return;
+  function itemStatusAjaxSuccess(items, response) {
+    let idMap = {};
+
+    // make map if ids to element arrays
```
EreMaijala: if => of?

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -88,147 +89,143 @@ VuFind.register('itemStatuses', function ItemStatuses() {
...
-    runItemAjaxForQueue: function runItemAjaxForQueue() {
-      if (this.itemStatusRunning) {
-        this.itemStatusTimer = setTimeout(this.runItemAjaxForQueue.bind(this), this.itemStatusDelay);
+
+      idMap[item.id].push(item.el);
+    });
+
+    // display data
+    response.data.statuses.forEach(function displayItemStatusResponse(status) {
+      if (typeof idMap[status.id] == "undefined") {
```
EreMaijala: Should this be `===`?

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -88,147 +89,143 @@ VuFind.register('itemStatuses', function ItemStatuses() {
...
     }
-    var id = $item.find('.hiddenId').val();
-    var handlerName = 'ils';
-    if ($item.data("handler-name")) {
-      handlerName = $item.data("handler-name");
-    } else if ($item.find('.handler-name').length > 0) {
-      handlerName = $item.find('.handler-name').val();
+
+    // update element to reflect lookup
+    el.classList.add("js-item-pending");
```
EreMaijala: This should be done before anything else. Determination of handlerName can happen later.

themes/bootstrap3/js/check_save_statuses.js
```diff
@@ -1,160 +1,148 @@
...
+/*global htmlEncode, userIsLoggedIn, StatusAjaxQueue, VuFind, unwrapJQuery */
+
+VuFind.register("saveStatuses", function ItemStatuses() {
+  function displaySaveStatus(itemLists, el) {
+    const $item = $(el);
+
+    if (itemLists.length > 0) {
+      // If we got lists back, display them!
+      var html = '<ul>' + itemLists.map(function convertToLi(l) {
+        return '<li><a href="' + l.list_url + '">' + htmlEncode(l.list_title) + '</a></li>';
```
EreMaijala: I know this code is same as before, but since it's being touched... It's dangerous to just stick a url in the attribute without any escaping/encoding. This should be rewritten to add nodes and their contents properly.

themes/bootstrap3/js/embedded_record.js
```diff
@@ -165,8 +165,8 @@ VuFind.register('embedded', function embedded() {
             });
             // Add events to record toolbar
             VuFind.lightbox.bind(longNode);
-            if (typeof checkSaveStatuses == 'function') {
-              checkSaveStatuses(longNode);
+            if (typeof VuFind.saveStatuses.init == 'function') {
```
EreMaijala: `===`?

themes/bootstrap3/js/lib/status_ajax_queue.js
```diff
@@ -0,0 +1,63 @@
...
+    this.payload = [];
+
+    const noop = () => {};
+    const promiseNoop = () => Promise.reject();
+
+    this.runFn = run ? run : promiseNoop;
+    this.successFn = success ? success : noop;
+    this.failureFn = failure ? failure : noop;
+
+    // move once Javascript is modularized
```
EreMaijala: Mark this with `TODO`

themes/bootstrap3/js/lib/status_ajax_queue.js
```diff
@@ -0,0 +1,63 @@
+class StatusAjaxQueue {
```
EreMaijala: Since this doesn't have anything specific to statuses and is used for versions as well, could we call it just `AjaxQueue` or `AjaxRequestQueue`?

EreMaijala (changes requested):

themes/bootstrap3/js/check_item_statuses.js
crhallberg: I'd gotten into the habit of using `==` for typeof comparisons due to an [eslint rule](https://eslint.org/docs/latest/rules/eqeqeq#smart), which is either deprecated or no longer used in VuFind. It's not a semantic issue, but I'm happy to adjust the syntax.

EreMaijala: Ok, no need to change if you think it's better to keep as is!

crhallberg:
Thank you for the detailed feedback, @EreMaijala!

demiankatz (changes requested):
Two things:

1.) There seem to be some eslint failures in the new code.

2.) Should we rename status_ajax_queue.js to match the new class name it contains?

EreMaijala:
@crhallberg I think it would be better to also change the file name from status_ajax_queue.js to ajax_request_queue.js.

themes/bootstrap3/templates/search/home.phtml
```diff
@@ -15,6 +15,10 @@
 ?>
 
 <div class="searchHomeContent">
+  <?=$this->component('confirm-menu', [
```
demiankatz: I think this change accidentally leaked in from a different PR...

demiankatz (changes requested):
@crhallberg, I think there's at least one unintentional change that's been added here -- see below. Do you mind double checking that there aren't others?

themes/bootstrap3/theme.config.php
```diff
@@ -42,6 +42,7 @@
         ['file' => 'vendor/validator.min.js', 'priority' => 140],
         ['file' => 'lib/form-attr-polyfill.js', 'priority' => 210], // input[form] polyfill, cannot load conditionally, since we need all versions of IE
         ['file' => 'lib/autocomplete.js', 'priority' => 220],
+        ['file' => 'lib/ajax_request_queue.js', 'priority' => 230],
```
demiankatz, line 43: @crhallberg, since we're emptying out the lib directory in #2755, should we put this new file somewhere else? Or alternatively, should we just make sure that proper style checking is getting applied to the lib directory, since I believe it was being skipped/ignored in the past?

demiankatz (changes requested):
One minor point we should discuss in light of other recent changes...

themes/bootstrap3/theme.config.php
demiankatz: I just took a quick look at eslint settings in build.xml, and I'm not seeing any explicit exclusion of either the lib or vendor subdirectories. This makes me wonder if perhaps eslint is operating in a non-recursive mode, and only checking files in the root of the js directory. Do we need to adjust the command and/or .eslintrc file to make things smarter?

crhallberg: Since this particular file has a utility that doesn't do anything when loaded (as compared to every other file), I think it is the only candidate for this folder. Then again, if it is the only one, we don't need the folder.

I'll look into the eslint config though.

crhallberg: I think we could fix this by [changing the blobs in build.xml](https://github.com/vufind-org/vufind/blob/dev/build.xml#L179-L188), ~~or we could make build.xml more concise by using the [`files` and `ignore` properties](https://eslint.org/docs/latest/use/configure/configuration-files-new#excluding-files-with-ignores) in `.eslintrc.js`.~~

Unfortunately, this is a future feature. In the meantime, I've expanded eslint's scope here: #2780.

demiankatz: I would favor moving the configuration to .eslintrc.js, since I think that makes things a little easier to manage. Do you want to address that through a separate PR?

demiankatz (changes requested):
@crhallberg, I've gone ahead and updated the changelog on the assumption that this will be merged soon, but it can't be merged yet -- tests are failing:

```
There were 3 failures:

1) VuFindTest\Mink\RecordVersionsTest::testVersions
Element not found: div.record-versions a index 0

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:387
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordVersionsTest.php:59
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordVersionsTest.php:92
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:100

2) VuFindTest\Mink\RecordVersionsTest::testVersionsInSearch2
Element not found: div.record-versions a index 0

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:387
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordVersionsTest.php:59
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordVersionsTest.php:102
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:100

3) VuFindTest\Mink\RecordVersionsTest::testDisabledVersionsTab
Element not found: div.record-versions a index 0

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:387
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordVersionsTest.php:128
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:100

FAILURES!
Tests: 2312, Assertions: 11802, Failures: 3.
```

demiankatz:
@crhallberg, I did some manual testing to confirm what the automated tests are finding: if I do a search for `id:0001732009-0` in the standard test environment, I expect to see a version link immediately above the call number in the first search result, but I see an error message instead. (If I check out the dev branch, this works correctly, so there's no question that the problem is caused by some change in this PR).

I also noticed that the code was still trying to load Hunt even though it no longer existed. I fixed that problem, but it seems unrelated to the other error. Can you please take another look?

crhallberg:
@demiankatz I fixed the record version error with `id:0001732009-0` and the like. Would you mind running the full suite when you have the chance?

demiankatz:
@crhallberg, it looks like your latest commit broke the build with some jshint errors. Once that's fixed, I can run tests!

demiankatz (approved):
Thanks, @crhallberg! All tests are now passing and my manual tests of saving favorites, looking at availability statuses, and navigating related versions seem to indicate normal expected behavior.

@EreMaijala, do you mind taking one last look at this before I merge (since you currently have an outstanding request for changes)?

EreMaijala (changes requested):
Everything else checks out, but the "Loading..." message is missing after "Located:":

dev:
dev - loading

This PR:
pr - loading

demiankatz:
Thanks for the catch, @EreMaijala and for the fix, @crhallberg. I've merged the latest dev into this, fixed style issues, confirmed that the fix works for me, and run the full test suite. I think we're all set to merge now, as soon as @EreMaijala gives final approval.

EreMaijala (approved):

PR #2448, opened 2022-05-31
advanced search: set focus to first input after adding new search field or group

RLangeUni:
When a search field or group is added in the Advanced Search page, the new group of input fields appears above the current focus. Blind users may not notice this new content.

### DIFF
--- a/themes/bootstrap3/js/advanced_search.js
+++ b/themes/bootstrap3/js/advanced_search.js
@@ -56,8 +56,6 @@ function addSearch(group, _fieldValues, isUser = false) {
 
   if (isUser) {
     $newSearch.find('input.form-control').focus();
-  } else if (group === 0 || fieldValues.term === '') {
-    $('#group' + group).find('input.form-control').first().focus();
   }
 
   return false;
@@ -87,7 +85,7 @@ deleteSearch = function _deleteSearch(group, sindex) {
   return false;
 };
 
-function addGroup(_firstTerm, _firstField, _join) {
+function addGroup(_firstTerm, _firstField, _join, isUser = false) {
   var firstTerm = _firstTerm || '';
   var firstField = _firstField || '';
   var join = _join || '';
@@ -123,7 +121,7 @@ function addGroup(_firstTerm, _firstField, _join) {
   $('#groupPlaceHolder').before($newGroup);
   // Populate
   groupLength[nextGroup] = 0;
-  addSearch(nextGroup, {term: firstTerm, field: firstField}, false);
+  addSearch(nextGroup, {term: firstTerm, field: firstField}, isUser);
   // Show join menu
   if (nextGroup > 0) {
     $('#groupJoin').removeClass('hidden');
--- a/themes/bootstrap3/js/advanced_search.js
+++ b/themes/bootstrap3/js/advanced_search.js
@@ -54,7 +54,6 @@ function addSearch(group, _fieldValues, isUser) {
   }
   groupLength[group]++;
 
-  // #17985 finc barf
   if (isUser) {
     $newSearch.find('input.form-control').focus();
   } else {
@@ -132,7 +131,6 @@ function addGroup(_firstTerm, _firstField, _join) {
     $('.adv-group-close').removeClass('hidden');
   }
 
-  // #17985 finc barf
   $newGroup.children('input.form-control').first().focus();
 
   return nextGroup++;
--- a/themes/bootstrap3/js/advanced_search.js
+++ b/themes/bootstrap3/js/advanced_search.js
@@ -3,7 +3,7 @@ var nextGroup = 0;
 var groupLength = [];
 var deleteGroup, deleteSearch;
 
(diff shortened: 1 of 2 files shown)


ckaz:
> When a search field or group is added in the Advanced Search page, the new group of input fields appears above the current focus. Blind users may not notice this new content.

This is what our accessibility reviewers stated. The idea was to solve this by setting the focus on the newly created input.

themes/bootstrap3/js/advanced_search.js
```diff
@@ -53,6 +53,13 @@ function addSearch(group, _fieldValues) {
     $('#group' + group + ' .adv-term-remove').removeClass('hidden');
   }
   groupLength[group]++;
+
+  if (isUser) {
```
demiankatz, line 57: Do we need to be concerned about the fact that `isUser` may be `undefined` here, since the `addUser` function is called from templates as well as from internal calls? I'm assuming that `undefined` essentially acts like `false` so it should be okay, but I wasn't sure if it might throw warnings into the console in some cases, or if it would be better to be more formal (either by including an `undefined` check, or by adding the parameter to the impacted template(s)). Feel free to ignore if you feel this is unnecessary; I'm just not sure what the best practice is -- maybe @crhallberg can weigh in.

demiankatz (commented):
One minor comment; I've also run the full integration test suite and everything still passes. I'll await input from @crhallberg before merging.

themes/bootstrap3/js/advanced_search.js
crhallberg: This boolean check should work fine. I'd recommend a default in the function definition to be sure:

```JS
function addSearch(group, _fieldValues, isUser = false) {
```

crhallberg (approved):
My only concern is that I think there's functionality somewhere that builds the page by triggering click events. In the case of editing an advanced search for example.

I did a look around but didn't find any examples. This all looks good as long as someone else takes a look for this potentially case. Default of false will likely protected against this issue.

themes/bootstrap3/js/advanced_search.js
demiankatz: @crhallberg, I don't believe there's any logic that builds things through click events -- the logic for rebuilding the form with a pre-existing search can be found [here](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/search/advanced/build_page.phtml#L16-L30), and it works by calling the Javascript functions directly.

I also agree that adding a default parameter value to isUser would make the code cleaner. Since we now require ES6, it shouldn't be a compatibility concern. @RLangeUni, would you like to make that change?

crhallberg (approved):

demiankatz (approved):
This looks good to me. One question for @RLangeUni before I merge it, however: I notice that when I edit an existing advanced search, the focus is now automatically set on the last search group (previously there was no focus at all). I don't think this is necessarily a problem, but I'm wondering if it's cause for any concern (would focusing the first search group be better in this case, for example?).

RLangeUni:
Thanks for your hint. It also seems to me, that it makes more sense to keep the focus on the first input of the first group after build_page.phtml created all search groups. Although I am not totally sure.

EreMaijala:
As far as I know it's generally not recommended to automatically focus an element after page load since it can be really confusing for visually impaired and also mobile users, dumping the user in the middle of a page without proper context.

demiankatz:
@RLangeUni / @EreMaijala, is the best solution here to simply omit the else clause of the `if (isUser)` check, so we only focus in response to the user pushing a button? That seems reasonable/conservative to me.

EreMaijala:
@demiankatz At a quick glance I believe that would work.

RLangeUni:
However, then the focus would remain on the "Add Search Group"-button after adding a new group. On the other hand, it is more consistent, because after click on "Remove Search Group" there is no focus on any input either.

demiankatz:
@RLangeUni, I may be misunderstanding, but I thought when buttons were pushed, the isUser flag was set to true, which WOULD focus on the newly created group. The only time isUser is false is when we're recreating a previous search through the templates, in which case we don't want to change focus. Or am I missing a detail?

RLangeUni:
Actually, addGroup calls addSearch in l126 always with false, so that focus was set on first input of that new group. I wanted to avoid an isUser in addSearch. But perhaps it makes more sense to differentiate on BOTH buttons.

RLangeUni:
I also added the "isUser" to addGroup, which should only be true when the user clicks.

demiankatz:
Thanks, @RLangeUni, this all makes sense to me now. I'm going to request one more review from @EreMaijala just to be sure we're on the same page, and if he approves, I think we can merge!

EreMaijala:
I'll confirm with our usability expert. I'm a bit concerned about the focus after deleting a group, but it could be fine.

EreMaijala (approved):
Got green light, so all good for me!

demiankatz:
Thanks, everyone, for taking the time to make sure this is as good as possible! :-)

PR #1319, opened 2019-02-13
Paia driver add error handling and scopes

hajoseng:
removed try-catch-blocks without further action
replaced paia_missing_scope_* strings
completed output of getMyProfile method

PR #2449, opened 2022-05-31
Support for PAIA client_credential authorization

lahmann:
PAIA supports for login authorization `client_credential` besides the standard `password` authorization (cf. http://gbv.github.io/paia/paia.html#login). Until now there was no option to configure the `grant_type` but always `password` was used. This adds functionality to support `client_credentials` authorization.
As the `grant_type`needs to be configured in PAIA.ini (but is set to `password` by default) this might result in broken configurations.
Up for discussion might be setting `password` as default for `grant_type`.


lahmann:
Finally all tests past - sorry for the somewhat messy commits.

@olli-gold it would be great, if you could have a look at the code and let me know if this works for you.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
```diff
@@ -59,6 +59,13 @@ class PAIA extends DAIA
      */
     protected $paiaURL;
 
+    /**
+     * Accepted grant_type for authorization
+     *
+     * @var string
+     */
+    protected $grantType;
```
demiankatz: As you suggested in your comments, I think setting this to "password" by default for backward compatibility makes sense.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
```diff
@@ -1027,7 +1039,8 @@ protected function paiaHandleErrors($array)
 
             default:
                 throw new ILSException(
-                    $array['error_description'] ?? $array['error'],
+                    $array['error_description'] ?? $array['error']
+                    . ' : ' .
```
demiankatz: Any particular reason why you're taking away the second integer parameter here and making it part of the first string parameter?

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
```diff
@@ -1774,6 +1787,43 @@ protected function paiaPostAsArray($file, $data)
      */
     protected function paiaLogin($username, $password)
     {
+        // as PAIA supports two authentication methods (defined as grant_type:
+        // password or client_credentials), check which one is configured
+        if (!in_array($this->grantType, ['password', 'client_credentials'])) {
+            throw new ILSException(
+                'Unsupported PAIA grantType configured.'
```
demiankatz: Might be worth including the illegal grant type value in the exception for easy reference.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
```diff
@@ -1774,6 +1787,43 @@ protected function paiaPostAsArray($file, $data)
...
+            $post_data = [
+                "username"   => $username,
+                "password"   => $password
+            ];
+            break;
+        case 'client_credentials':
+            // client_credentials only works if we have client_credentials
+            // username and password
+            if (isset($this->config['PAIA']['client_username'])
+                && isset($this->config['PAIA']['client_password'])
```
demiankatz: Should you have an else that throws an exception here in case of misconfiguration?

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
```diff
@@ -1785,9 +1835,39 @@ protected function paiaLogin($username, $password)
...
+        try {
+            $result = $this->httpService->post(
+                $this->paiaURL . 'auth/login',
+                json_encode($post_data),
+                'application/json; charset=UTF-8',
+                $this->paiaTimeout,
+                $header_data
+            );
+        } catch (\Exception $e) {
+            throw new ILSException($e->getMessage());
```
demiankatz: Should you pass along the exception code as well as the message?

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
```diff
@@ -1785,9 +1835,39 @@ protected function paiaLogin($username, $password)
...
+        }
+
+        // continue with result data
+        $responseJson = $result->getBody();
+
+        try {
+            $responseArray = $this->paiaParseJsonAsArray($responseJson);
+        } catch (\Exception $e) {
+            // all error handling is done in paiaHandleErrors so pass on the
+            // excpetion
```
demiankatz: Typo: should be "exception"

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/PAIATest.php
```diff
@@ -56,6 +56,7 @@ class PAIATest extends \VuFindTest\Unit\ILSDriverTestCase
         'PAIA' =>
             [
                 'baseUrl'            => 'http://paia.gbv.de/',
+                'grantType'          => 'password',
```
demiankatz, line 59: Should we also add some test cases to cover the new logic (config validation, etc.)?

demiankatz (changes requested):
I've left a few comments and suggestions; I'll also await input from @olli-gold before merging.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
lahmann: Ok, I've gone ahead and made `password` the default.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
lahmann: Thanks for catching this. As we are still catching up with VuFind 6+ the bug @hajoseng fixed in #1915 was still present in our code and this was a temporary workaround for debugging which slipped through. I just reverted the change.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
lahmann: Good point, I added the illegal grant_type.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
lahmann: Good point, added an else and an explicit exception message.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
lahmann: I have removed this try-catch block in reference to #1319. A more in depth exception-handling should not be the scope of this pull request.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
lahmann: I have removed this try-catch block in reference to #1319. A more in depth exception-handling should not be the scope of this pull request.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/PAIATest.php
lahmann: I would like to hear @olli-gold thoughts on this as I think when going with password as default and a more explicit exception message for missing configuration values in case client_credentials are configured, this might suffice. On the other hand validating correct basic authentication (cf. https://datatracker.ietf.org/doc/html/rfc2617) would help sticking with the PAIA specification.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
demiankatz: Thanks! However, I notice that you changed grantType to grant_type in the message, even though "grantType" is used in both the code and configuration.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
demiankatz: Looks like this new message also uses grant_type instead of grantType. I think this is a sign that it may be worthwhile at some point to clean up the inconsistent naming conventions within this class. Generally, VuFind uses camelCase, but this code has quite a bit of snake_case in it. (Obviously I'm not proposing that we need to fix that right now, but something to consider for a future pass).

demiankatz: Might also be worth changing "client_credentials" to "client credentials" since I don't think that's actually a setting name, it's just a description.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/PAIATest.php
demiankatz: It should be fairly straightforward to mock the httpService and assert that appropriate headers are sent in when client credentials are configured... but it's definitely worth making sure @olli-gold agrees with the approach/implementation before spending time locking it in place with tests!

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
olli-gold: client_credentials is not just a description, it's fixed vocabulary from the PAIA specification (cf. http://gbv.github.io/paia/paia.html#login). So I'd stick with it in this case. The same applies to grant_type, as it's also a fixed parameter name in PAIA (but of course we could map it from a different name in VuFind's PAIA config, as Andre did).
I agree, that it would be nice to have it a bit more consistent, but there is obviously a collision between PAIA terminology (where mostly snake_case is used) and VuFind (which is using camelCase, as you mentioned), which makes it hard to achieve consistency...

demiankatz: Thanks for the clarification, @olli-gold -- I hadn't realized that some of this terminology was coming directly from PAIA. In that case, I suspect that most of my feedback can be safely ignored. :-)

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
demiankatz: Resolving this comment based on @olli-gold's note below.

config/vufind/PAIA.ini
```diff
@@ -18,6 +18,19 @@ baseUrl = ""
...
+; specification.
+; Server specific grant type should be set ('password' or 'client_credentials') but
+; defaults to 'password'.
+;grantType = "password"
+
+; If grant type is set to client_credentials the client credentials must be
+; set below.
+;client_username = 'client'
+;client_password = 'password'
+
```
olli-gold, line 37: I don't understand, why we need the credentials to be set here in the config. The PAIA login is always patron specific and as far as I understand it, the difference between client_credentials and password mode is the way the authentication is done (password authentication sends the username and password in the POST request, while client_credentials sends them in the HTTP header in an base64 encoded form). So, why would we need to set (global) credentials in the configuration? Shouldn't we rather use the credentials the user provided in the login form? If we cannot use those, which credentials should be provided in the config (do we need a dummy user for that?)?

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
```diff
@@ -1774,6 +1786,48 @@ protected function paiaPostAsArray($file, $data)
...
+            // username and password
+            if (isset($this->config['PAIA']['client_username'])
+                && isset($this->config['PAIA']['client_password'])
+            ) {
+                $header_data["Authorization"] = 'Basic ' .
+                    base64_encode(
+                        $this->config['PAIA']['client_username'] . ':' .
+                        $this->config['PAIA']['client_password']
+                    );
+                $post_data = [
```
olli-gold, line 1819: As I mentioned in the above comment, I do not understand, why we need global config settings at this point. Shouldn't we rather use the username and password the user provided in the login form (stored in $username and $password)?

olli-gold (commented):
I do only have a question, why it is necessary to configure the client credentials in the configuration. If this is necessary, I am missing some more comments on that in the configuration file, what credentials that are. If it is not necessary, I'd request to change it and to drop it from configuration.

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
lahmann: Thanks @olli-gold exactly that's the case. So far I opted for camelCase whenever it is VuFind related (variables, configuration) but stick with PAIA syntax in documentation and configuration. The only divergence from this rule I am aware of are the configuration values `client_password`and `client_username`. Would changing those to `clientPassword`and `clientUsername`add to more consitency? Or would `client_credentialPassword` and `client_credentialUsername` be even more explicit and consistent?

config/vufind/PAIA.ini
lahmann: The `grant_type` `client_credential` in PAIA is not patron centered, i.e. the credentials of the patron as a client, but service centered, i.e. the credentials for the service accessing the PAIA API as a client. Therefore a shared secret is used and needs to be stored in the configuration file. Of course this will allow to the VuFind client to act as any patron without proper authentication. Therefore this method is of best use when combined with authorization via Shibboleth or equivalent.
Maybe this should be pointed out more precisely in the configuration comments?

demiankatz: Is it possible that we're conflating two different things here? One issue is whether the credentials come from the end user or are hard-coded in the configuration file. The other issue is how we pass the credentials to the server -- via HTTP POST or HTTP Basic Authentication. Is it necessarily true that user-provided credentials must always be POSTed while pre-configured credentials must always be used for Basic Authentication? Would it make more sense to have two separate switches for these things? Is it possible that trying to use formal OAuth/PAIA language actually makes some things less clear in this case?

(I don't presuppose answers to any of those questions, but I think that answering some of them may prove helpful). :-)

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
demiankatz: I think clientPassword and clientUsername might be slightly preferable to client_password and client_username in the name of consistency... but based on my other comment, maybe we'll end up renaming them entirely. We should probably make sure we're all on the same page about intent before we finalize naming. (For example, if we want to be able to override user-entered credentials regardless of login method, maybe these become overrideUsername / overridePassword).

config/vufind/PAIA.ini
lahmann: PAIA login (cf. http://gbv.github.io/paia/paia.html#login) supports two ways of authentication identified by the POST parameter `grant_type`:
1. `password` - athentication is done by providing patron `id` and `password` in POST data (GET is optional)
2. `client_credential` - authentication is done by providing shared secret consisting of concatenated client-username and client-password plus base64 encoding. This authenticates the client service (i.e. VuFind) and a patron identifier needs to be passed along to actually retrieve patron data or perform actions for the patron.

> Is it necessarily true that user-provided credentials must always be POSTed while pre-configured credentials must always be used for Basic Authentication?

According to the PAIA specs, yes: "The request header “Access” is required for client credentials grant with HTTP basic authentification as defined in [RFC 2617](http://tools.ietf.org/html/rfc2617)."

But both authentication methods need POST data: `grant_type`, `scope` for both, `patron` only for `grant_type=client_credential` and `username`, `password` only for `grant_type=password`.

Does this make things clearer? :)

demiankatz: Thanks, @lahmann, that's very helpful! I would suggest adding something like this to the .ini comments:

When grantType is set to password (the default), user-entered credentials will be passed along to the PAIA server for authentication against the API. When grantType is set to client_credentials, then the "shared secret" credentials provided in clientUsername and clientPassword below will always be used for PAIA API authentication.

I think that helps to clarify what will actually happen, in case people do not understand the specifications being referenced in the existing comments (which I think you could also retain, or not, depending on your preferences).

lahmann: Thanks @demiankatz, I added the comment and refactored the credential names. I think this should make things more understandable?

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
lahmann: I refactored the variable names in the latest commit.

config/vufind/PAIA.ini
demiankatz: I was thinking of putting the comment slightly higher in the file, but I think this works too. Maybe @olli-gold can comment on the clarity since he can look at it with a fresh/objective perspective. :-)

olli-gold: Thanks for clarification. I understand better now, why we need the central username and password for client authentication in the config. But I think the comment is a bit too technical and would suggest to refactor it a little bit. 

Thus, I would cut off the first sentence in line 29f, as it should be clear, that you do not need to fill out this section for password mode, but only for client_credentials. And for the second sentence I would propose something like that:
When grantType is set to client_credentials, please insert the "shared secret" credentials provided from your PAIA provider in clientUsername and clientPassword below.

olli-gold (approved):
Basically I am happy with this approach and I have tested it successfully here. So, besides the comment about the comment in PAIA.ini above, I have nothing to add and would approve this PR.

config/vufind/PAIA.ini
demiankatz: @olli-gold and @lahmann, please take a look at commit [5159cc5](https://github.com/vufind-org/vufind/pull/2449/commits/5159cc5bdba4c47d477e29aaa7627cd481907b70) -- I've rearranged and reformatted the comments, taking inspiration from @olli-gold's suggestion. I think this is a little bit more clear/accessible. Any objections or suggestions for further improvement?

demiankatz (approved):
Since @olli-gold also approves this, I think it is ready to go -- but I'll wait for feedback on my latest revision to the .ini comments in case more slight adjustments are needed. Once I hear from both @lahmann and @olli-gold that all is well, I will merge this. Thanks again, both of you!

config/vufind/PAIA.ini
olli-gold: Just giving you a thumb up, that looks good to me. Thank you!

lahmann:
Thanks @demiankatz for your commit, this improves readability. I think this should be ready to merge.

demiankatz:
Thanks again -- merging now!

PR #2450, opened 2022-06-01
Calculate date threshold once in expiration commands.

EreMaijala:
This avoids the timestamp from creeping forward during the expiration loop when it takes a second or more to process a single batch. That could create a semi-infinite loop if there are e.g. searches to delete for every second.


module/VuFind/src/VuFind/Db/Table/AuthHash.php
```diff
@@ -108,14 +108,14 @@ public function getLatestBySessionId($sessionId)
...
+     * 'Y-m-d H:i:s'.
      *
      * @return void
      */
-    protected function expirationCallback($select, $daysOld)
+    protected function expirationCallback($select, $dateLimit)
     {
-        $expireDate = time() - $daysOld * 24 * 60 * 60;
-        $select->where->lessThan('created', date('Y-m-d H:i:s', $expireDate));
+        $select->where->lessThan('created', strtotime($dateLimit));
```
demiankatz: Why are you calling strtotime here? It seems like you probably shouldn't...

demiankatz (changes requested):
I think I spotted one small bug here... once that's fixed, I will do some hands-on testing before merging, but I think this is nearly ready to go.

module/VuFind/src/VuFind/Db/Table/AuthHash.php
EreMaijala: Oops, fixed. Also tested that it works correctly now. Previously it looked like it worked but didn't remove anything. Apparently a bad value like that will get truncated with just a warning that's not visible anywhere.

demiankatz (approved):
Thanks, @EreMaijala. This looks good to me, and I've updated the changelog to reflect the adjusted trait.

PR #2451, opened 2022-06-01
Add an index to search table to improve performance when expiring sea…

EreMaijala:
…rches.

TODO
- [x] Test MySQL upgrade
- [x] Test PostgreSQL upgrade
- [ ] Add note to database changelog when merging

### DIFF
--- a/module/VuFind/sql/migrations/pgsql/9.0/001-add-search-index.sql
+++ b/module/VuFind/sql/migrations/pgsql/9.0/001-add-search-index.sql
@@ -1,0 +1,5 @@
+--
+-- Modifications to table `search`
+--
+
+CREATE INDEX search_created_saved_idx ON search (created, saved);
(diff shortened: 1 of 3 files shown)


module/VuFind/sql/migrations/pgsql/9.0/001-add-search-index.sql
```diff
@@ -0,0 +1,5 @@
+--
+-- Modifications to table `search`
+--
+
+CREATE INDEX search_created_saved_idx ON search (created, saved);
```
demiankatz, line 5: Is there a reason that the index name in the migration is different from the index name in the pgsql.sql file?

demiankatz (changes requested):
I tested this out and it seems to work, but I noticed one small discrepancy that we should probably address before merging.

module/VuFind/sql/migrations/pgsql/9.0/001-add-search-index.sql
EreMaijala: Apart from an unsaved file in editor, no, not really. :) Thanks for catching that!

demiankatz (approved):

PR #2452, opened 2022-06-01
Blender: Make sure to ignore hidden filters in initFromRequest.

EreMaijala:
Previously only hiddenFilter used in combined search was ignored, while hiddenFilters is used elsewhere.

### DIFF
--- a/module/VuFind/src/VuFind/Search/Blender/Params.php
+++ b/module/VuFind/src/VuFind/Search/Blender/Params.php
@@ -119,7 +119,8 @@ class Params extends \VuFind\Search\Solr\Params
             'type',
             'sort',
             'filter',
-            'hiddenFilter'
+            'hiddenFilter',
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/Search/Blender/Params.php
```diff
@@ -119,7 +119,8 @@ public function initFromRequest($request)
             'type',
             'sort',
             'filter',
-            'hiddenFilter'
+            'hiddenFilter',
```
demiankatz: Do we really need to support both hiddenFilter (singular) and hiddenFilters (plural), or do we just need to change one to the other? The only reference I can find in the code to the singular form is a configuration setting used in combined search, but that gets mapped to the plural form when added to the request.

demiankatz (approved):
Thanks for the fix!

PR #2453, opened 2022-06-02
Deduplication listener: Split active record source determination from…

EreMaijala:
… fetchLocalRecord.

This makes it easier to customize deduplication. No functional changes.


demiankatz (approved):
Looks good to me!

Longer-term, as we may have discussed before, it would be nice to get test coverage for this class and perhaps reduce its internal reliance on the service locator (i.e. inject the config manager instead of the top-level one)... but that's a project for another day.

EreMaijala:
@demiankatz Agreed. Also the method could be further split to smaller parts, but I wanted to keep this one light.

PR #2454, opened 2022-06-02
Truncation for hierarchical facet, regarding VUFIND-1559

j4lib:
Regarding the truncation of the hierarchical facet, here comes the promised pull request.
What I did:
- added an extra truncation initiation for hierarchical facet via the class "truncate-hierarchy" instead of "truncate-facets". This was necessary because otherwise the truncation is done before the tree nodes are created and the nodes are set to "truncate-done" beforehand. So now the truncation for hierarchical facets is initiated when the tree nodes are done loading.
- added the usual buttons around the hierarchical facets and added the facets_before_more option. It would be worth considering to move this 'LESS and SEE MORE links' part and following into its own template as I simply copied it from the usual facets.

Still puzzling: 
- I still get this layout bug that I mentioned in slack. Which seems to be introduced by .truncate-done {overflow: hidden} in css. But that might be on my part, so it would be nice if someone could test it with their layout.
- And it of course needs to be double checked if everything else still works as expected :)

TODO
- [x] Fix SideFacetsDeferred
- [x] Run full test suite one more time before merging


demiankatz:
Thanks, @j4lib, this is great! I've done a bit of simplification and updating here. The logic you proposed moving to a separate template actually isn't needed at all now that the Javascript does all the button creation, so I removed it (and things still work for me). There were also a couple of very small things (a typo in an event name, and some language strings that could be omitted to use equivalent defaults) which I also fixed up.

I didn't see any layout issues, but maybe I didn't try hard enough -- is the current simplified version here looking any better for you?

I'm also interested to hear whether @EreMaijala has any concerns about this... but it seems pretty straightforward to me!

j4lib:
Great that it could be done even simpler! I can confirm that the simplification didn't do any harm, still working on my side. 
Then what about https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/Recommend/SideFacets/cluster-list.phtml where I copied those lines from? Can those lines maybe be removed/simplified as well?

The layout issue that I see is easily fixed by adding 
.jstree-container-ul {
  margin-bottom: 0px;
}
The extra margin-bottom comes straight out of the bootstrap vendor css (https://github.com/vufind-org/vufind/blob/11d86b56bfce949086627274883e49730755db5a/themes/bootstrap3/less/vendor/bootstrap/type.less#L163) that adds a 10px margin to the ul of jstree-container-ul in my case. 
It is absolutely possible that my less/css has just grown too different from the current version over the years and at your end this value is overwritten at the right place already. I misspelled facet so I wouldn't be surprised :sweat_smile:

j4lib:
Ah I found an error, instead of '.jstree-node' it needs to be 
VuFind.truncate.initTruncate('.truncate-hierarchy', '.list-group-item'); 

Otherwise further 'more' labels are added to subtrees in case you choose a facet value in a subtree.

EreMaijala (changes requested):
Looks pretty good to me, but I'd like the event to include the node in question so that `registerHierarchicalFacetTruncation` could initialize truncate for the specific tree node instead of going through all nodes for every event.

demiankatz:
Thanks, @j4lib! Regarding the existing code in cluster-list.phtml, the difference is that cluster-list.phtml actually renders the facet values in PHP, while the hierarchical version creates a container that gets filled in later via Javascript. The code you copied is the value-rendering part, which doesn't do anything for hierarchies. This code also supports some "open facet list in lightbox" functionality that I think would be challenging to implement for hierarchies (but might be a future project, if anyone actually cares about it).

Do you see the style problem you describe if you use the default bootstrap3/bootprint3/sandal themes?

Do you want to fix the initialization issue you described, or should I?

Any thoughts on @EreMaijala's suggestion?

I'm out of the office until Monday, but I can help further next week if necessary!

j4lib:
Ah okay! That's good to know. From my side, there is no need for a lightbox there.

grafik
So I see the problem in bootprint3 and bootstrap3, but not in sandal. In sandal it looks as expected. I will compare some more files automatically to be sure I did not introduce this issue somehow in the code.

I can fix it, I am not quite sure if I know how to fix Ere's suggestion though. It sounds like a good idea, but as I said, my JS/JQuery experience is rather limited.

EreMaijala:
@j4lib Something like this should be enough to initialize truncation only once and for the tree that's loaded:

```
function buildFacetTree(treeNode, facetData, inSidebar) {
  // Enable keyboard navigation also when a screen reader is active
  treeNode.bind('select_node.jstree', VuFind.sideFacets.showLoadingOverlay);

  var currentPath = treeNode.data('path');
  var allowExclude = treeNode.data('exclude');
  var excludeTitle = treeNode.data('exclude-title');

  var results = buildFacetNodes(facetData, currentPath, allowExclude, excludeTitle, inSidebar);
  treeNode.find('.loading-spinner').parent().remove();
  if (inSidebar) {
    treeNode.on('loaded.jstree open_node.jstree', function treeNodeOpen(/*e, data*/) {
      treeNode.find('ul.jstree-container-ul > li.jstree-node').addClass('list-group-item');
      treeNode.find('a.exclude').click(VuFind.sideFacets.showLoadingOverlay);
    });
    if (treeNode.parent().hasClass('truncate-hierarchy')) {
      treeNode.on('loaded.jstree', function initHierarchyTruncate(/*e, data*/) {
        VuFind.truncate.initTruncate(treeNode.parent(), '.list-group-item');
      });
    }
  }
  treeNode.jstree({
    'core': {
      'data': results
    }
  });
}
```

Now this leaves a couple of further issues: 
1.) The whole thing doesn't seem to work with SideFacetsDeferred.
2.) I think there should be a way to disable truncation, but currently the `facets_before_more` setting is forced to a positive value and `$truncateSettings` in hierarchical-facet.phtml is always set to non-false value.

demiankatz:
@j4lib, I've just merged the latest dev branch into this PR to keep things up to date. Do you want to try to apply @EreMaijala's suggested code, or should I?

j4lib:
Sorry for not replying earlier, Whit Monday is a public holiday in Germany.
@EreMaijala, regarding the issues:
1) I actually never tried SideFacetsDeferred. It really doesn't work. I'll try and see what is necessary. 
2) Is there an option for disabling truncation in general? I am wondering if this concerns all facets?

EreMaijala:
@j4lib I believe there has been no need for disabling truncation with normal facets since they, by nature, need to be limited to a certain number of entries. Hierarchical facets, as I see them, are a bit different.

demiankatz:
Thanks for the ongoing progress. I've added a TODO checkbox to the top of this to remind us that SideFacetsDeferred needs work. Are there other outstanding issues we still need to track there?

Regarding disabling truncation, it does appear that the `showMore` setting from facets.ini always gets forced to a positive value, so there is not currently a mechanism for disabling truncation. (For regular facets, truncation gets disabled as a side effect of setting `showMoreInLightbox` to `true`, but that special case doesn't apply to hierarchical facets, as discussed above. I suppose we could set up something like if you set `showMore` to `PHP_INT_MAX` then the code is automatically disabled... but maybe that's silly.

EreMaijala:
Could we consider removing the automatic bump to 6 in VuFind 9.0 or at least ignore it for hierarchical facets? That would allow cleaner implementation.

demiankatz:
@EreMaijala, what if we add a second parameter to the method that retrieves the configuration setting to specify the default value? Then for hierarchical facets, we can just call the method when we render the subtemplate using a default of 0 instead of 6, and in the hierarchical template, we can disable truncation when the value is < 1. Nothing else will be impacted at all. I think that would be a safe change for 8.1.

It also has the positive side effect of removing a "magic number" from the code (or at least reducing the degree to which it seems magical).

EreMaijala:
@demiankatz Sounds good to me!

demiankatz:
Thanks, @EreMaijala. Commit 4d77f566daa09b55826cb4268584a9b94daff903 adds the proposed functionality and also cleans up/expands some brief/inaccurate comments.

I think that just leaves the SideFacetsDeferred problem. Is @j4lib going to look at that, or should I? I'm not overly familiar with the code, but my guess is that it's just a matter of figuring out how to trigger the truncation event at the appropriate time.

j4lib:
I had a look at it but I could not figure it out yet. I am wondering about this line https://github.com/vufind-org/vufind/blob/7d14cfa0f41e43d3a8d5f32fa740a8276e3710b2/themes/bootstrap3/js/facets.js#L217: there is some emit treenodeloaded already in the code, but I don't think it does anything? At least, I think it needs to be somewhere around there but I am also not very familiar with the code of SideFacetsDeferred.

demiankatz:
I just took a look, and I think @EreMaijala's input might be helpful. I think the problem is probably that the deferred side facets load hierarchical facets through inline code here: https://github.com/vufind-org/vufind/blob/7d14cfa0f41e43d3a8d5f32fa740a8276e3710b2/themes/bootstrap3/templates/Recommend/SideFacetsDeferred.phtml#L83-L92 instead of by including SideFacets/hierarchical-facet.phtml (where we added the truncation markup). I'm not sure if this is intentional (because SideFacetsDeferred actually needs to be structured differently) or if it's an oversight (because SideFacets refactoring may have occurred at a different time than the creation of SideFacetsDeferred -- for example, the fact that the deferred version uses a div container instead of an li may be related to an oversight when some accessibility-related changes were made). So I'm not sure if we need to adjust SideFacetsDeferred to include hierarchical-facet.phtml, or if we need to modify the existing markup there. Any thoughts?

EreMaijala:
@demiankatz I believe that it was an oversight from refactoring. I've now pushed a fix to @j4lib's branch that at least from a quick glance fixes things, but I didn't have time for testing everything properly.

themes/bootstrap3/templates/Recommend/SideFacets/hierarchical-facet.phtml
```diff
@@ -32,7 +32,16 @@ JS;
...
-<li id="facet_<?=$this->escapeHtml($this->title)?>" class="jstree-facet"
+<?php
+  $truncateSettings = $this->facets_before_more < 1 ? false : [
+    'rows' => $this->facets_before_more,
+    'btn-class' => 'facet',
+    'wrapper-class' => false,
+    'wrapper-tagname' => 'div',
+  ];
+?>
+<div<?php if ($truncateSettings): ?> class="truncate-hierarchy" data-truncate="<?=$this->escapeHtml(json_encode($truncateSettings))?>"<?php endif; ?>>
```
demiankatz, line 43: Is it legal to wrap a `<div>` around an `<li>` or does the `<li>` need to be directly inside the containing list? If the latter, maybe we should be applying the truncation settings directly to the `<li>` instead of building a container -- and if so, we may need to adjust the truncation selector settings.

demiankatz (commented):
Thanks for the fix, @EreMaijala! I'm probably not going to have time to test until next week. @j4lib, do you have time before that? (I'll look at it next week one way or the other, but if you can offer feedback sooner, it wouldn't hurt!).

The observation that we're using `<li>` instead of `<div>` for facet groups now also led me to raise the question below....

themes/bootstrap3/templates/Recommend/SideFacets/hierarchical-facet.phtml
demiankatz: I'm happy to try to adjust this next week if it is likely to be necessary/helpful.

demiankatz: Just to follow up, I tried eliminating the div with this patch:

```
-<div<?php if ($truncateSettings): ?> class="truncate-hierarchy" data-truncate="<?=$this->escapeHtml(json_encode($truncateSettings))?>"<?php endif; ?>>
-  <li id="facet_<?=$this->escapeHtml($this->title)?>" class="jstree-facet"
-    data-source="<?=$this->escapeHtml($this->searchClassId)?>"
-    data-facet="<?=$this->escapeHtmlAttr($this->title)?>"
-    data-path="<?=$this->currentPath()?>"
-    data-exclude="<?=$this->allowExclude?>"
-    data-operator="<?=$this->recommend->getFacetOperator($this->title)?>"
-    data-exclude-title="<?=$this->transEscAttr('exclude_facet')?>"
-    data-sort="<?=$this->sortOptions ?? ''?>"
-    data-query="<?=$this->escapeHtmlAttr($urlQuery)?>"
-    data-query-suppressed="<?=$querySuppressed ? '1' : '0' ?>"
-    data-extra-fields="<?=$this->escapeHtml(implode(',', $extraUrlFields))?>">
-  </li>
-</div>
+<li id="facet_<?=$this->escapeHtml($this->title)?>"
+  class="jstree-facet<?php if ($truncateSettings): ?> truncate-hierarchy<?php endif; ?>"
+  <?php if ($truncateSettings): ?>data-truncate="<?=$this->escapeHtml(json_encode($truncateSettings))?>"<?php endif; ?>
+  data-source="<?=$this->escapeHtml($this->searchClassId)?>"
+  data-facet="<?=$this->escapeHtmlAttr($this->title)?>"
+  data-path="<?=$this->currentPath()?>"
+  data-exclude="<?=$this->allowExclude?>"
+  data-operator="<?=$this->recommend->getFacetOperator($this->title)?>"
+  data-exclude-title="<?=$this->transEscAttr('exclude_facet')?>"
+  data-sort="<?=$this->sortOptions ?? ''?>"
+  data-query="<?=$this->escapeHtmlAttr($urlQuery)?>"
+  data-query-suppressed="<?=$querySuppressed ? '1' : '0' ?>"
+  data-extra-fields="<?=$this->escapeHtml(implode(',', $extraUrlFields))?>">
+</li>
```

...but it didn't work. I don't have time to troubleshoot further this week, but maybe it's not worth the effort. It appears that we already have facet entries as `<li>` tags without containing lists, and some of them are wrapped in `<div>` tags for collapsing, so if this is wrong, it's not more wrong than what we're already doing. This may need to be revisited, but that's probably a separate issue.

demiankatz (approved):
I think at this point, this looks mergeable to me... but I'd like to hear whether @j4lib or @EreMaijala have any final concerns (particularly related to the markup issues I raised above) before I hit the button!

j4lib:
Sorry for the late reply. I had to write an important report which was due this week. The changes look good to me and no errors coming up when I switch to the SideFacetsDeferred version. ~But in the deferred version the hierarchical facet is not truncated for me. I hope to find some time tomorrow morning to take a closer look.~ Nevermind, I was just stupid. It's working just fine!

themes/bootstrap3/templates/Recommend/SideFacets/hierarchical-facet.phtml
j4lib: In mozilla developer (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li ) it is stated that permitted parents for `<li>` are `<ul>`, `<ol>`, or `<menu>` element. "Though not a conforming usage, the obsolete `<dir>` can also be a parent." So it seems like div is not allowed as parent. I assume this would need to change at several places.

demiankatz: @j4lib, I've done some digging and discovered that I was wrong about the `<li>` for hierarchical facets being related to recent accessibility changes. It has been that way since it was introduced many years ago. I wonder if this was a simple error, since individual facets are represented as `<li>`'s, but in the case of hierarchical facets, the tag is just a placeholder that gets filled in via AJAX. I think we might just want to change this `<li>` to a `<div>` to make the HTML semantically correct. @EreMaijala, what do you think?

demiankatz:
Thanks, @j4lib, glad to hear everything seems to be working. I'll wait to see what @EreMaijala has to say about the li vs. div situation, but given that that markup problem has existed for years, it might make sense to merge this as-is and address that in a separate commit.

themes/bootstrap3/templates/Recommend/SideFacets/hierarchical-facet.phtml
ckaz: > Is it legal to wrap a `<div>` around an `<li>` or does the `<li>` need to be directly inside the containing list? If the latter, maybe we should be applying the truncation settings directly to the `<li>` instead of building a container -- and if so, we may need to adjust the truncation selector settings.

Hi there, just chanced upon this:

As @j4lib mentions, you mustn't wrap `<li>` in `<div>` -- however, we were asked by our accessibility testing firm, to **consistently style all facets using `<ul>` since they are lists**. This is a bigger chunk which needs changes in several files, so definitely something for VuFind 9, imo.

demiankatz: Thanks, @ckaz. I think there's a very strong possibility that we can just change the `<li>` to a `<div>` in this case and the hierarchical facets will be okay (I can try this when I get back next week, if nobody objects or does it in the meantime) -- I'm pretty sure that the jstree library used for rendering them already uses lists internally and is reasonably accessible. Then, as you say, the other project is reformatting the other facets as lists, which is a separate project that would be appropriate for 9.0 -- I welcome a PR on that if anyone has time.

ckaz: > Then, as you say, the other project is reformatting the other facets as lists, which is a separate project that would be appropriate for 9.0 -- I welcome a PR on that if anyone has time.

@Demian: I think we have this implemented well enough to warant a PR on this soon!

EreMaijala (approved):
Let's get this merged and work on fixing the markup separately.

demiankatz:
See #2475 for the markup fix.

PR #2455, opened 2022-06-03
Rename EZB resolver driver to JOP, and improve code

RLangeUni:
* rename ezb to jop for clarity reasons
* make xpath title selector a static variable that can be overwritten for better inheritance
* add result state 10 for Unknown Print / Unknown Electronics
* check date format in downgradeOpenUrl

TODO
- [x] Update changelog when merging (note renamed driver)


module/VuFind/src/VuFind/Resolver/Driver/PluginManager.php
```diff
@@ -49,7 +49,7 @@ class PluginManager extends \VuFind\ServiceManager\AbstractPluginManager
         '360link' => Threesixtylink::class,
         'alma' => Alma::class,
         'demo' => Demo::class,
-        'ezb' => Ezb::class,
+        'jop' => Jop::class,
```
demiankatz, line 53: We should set up some aliases from ezb and the old class name to point to the new jop name, so people with outdated configurations won't break their resolving.

demiankatz (changes requested):
One quick suggestion regarding backward compatibility -- once that's fixed, I'll take a closer look at the rest of the code changes (I'm a little short on time today, but I'll have more availability next week). Thanks for this!

RLangeUni:
I have added an extra alias for the ol ezb resolver name.

module/VuFind/src/VuFind/Resolver/Driver/Jop.php
```diff
@@ -187,58 +198,64 @@ public function getResolverUrl($openURL)
...
             }
         }
+        if (!empty($downgraded)) {
+            // we need 'genre' but only the values
+            // article or journal are allowed...
+            return "genre=article&" . implode('&', $downgraded);
+        }
 
-        return implode('&', $downgraded);
+        return implode('&', $parsed);
```
demiankatz: Is there a reason we pass back the original settings if we were unable to find anything to downgrade? This seems like a behavior change from before, and I'm not sure that it will end up with good results. If there's a reason for it, maybe a comment would help explain why.

module/VuFind/src/VuFind/Resolver/Driver/Jop.php
```diff
@@ -50,8 +50,17 @@
...
+class Jop extends AbstractBase
 {
+    /**
+     * As the JOP resolver provides also generic labels 'Article', 'Journal'
+     * etc. in element AccessLevel this label can be used as title for
+     * resolver results by setting this variable to 'AccessLevel'
+     *
+     * @var string
+     */
+    protected static $xpathTitleSelector = 'Title';
```
demiankatz: Is there a strong reason to make this static, rather than a regular property?

demiankatz (commented):
Thanks, @RLangeUni! I gave this a closer review now and had a couple of small questions.

I'll also ask around and see if there are any other EZB/JOP users who might have input on this.

Since this is a backward-breaking change, I've marked this for the 9.0 milestone and will plan to merge it after 8.1 is released (or into the dev-9.0 branch, if it's ready before that happens).

I've also added a TODO checkbox to the description as a reminder to update the changelog to note the driver name change when we merge this.

demiankatz:
Quick update: I spoke with Markus Fischer, who has worked on the EZB driver in the past, and he approves the changes... so I think we can merge this once the small questions I asked above have been addressed.

module/VuFind/src/VuFind/Resolver/Driver/Jop.php
RLangeUni: Yes, you were right. I fixed that. Thanks.

module/VuFind/src/VuFind/Resolver/Driver/Jop.php
RLangeUni: In finc we are using a trait and a child class of EZB / JOP with a constructor, in which we do not call the parent constructor. So with a static property or const we would minimize changes.

demiankatz: @RLangeUni, as long as you initialize the property directly in the property declaration rather than in the contructor, the value should be correctly populated whether or not the parent constructor gets explicitly called. The property doesn't need to be static for this to work. (Unless I'm misunderstanding something).

If you want to be able to explicitly refer to Jop::$xpathTitleSelector from a trait, as opposed to having the trait simply assume that $this->xpathTitleSelector is set, maybe it would be preferable to use a constant rather than a static property.

Anyway, if you still prefer to keep this as-is, I don't mind -- but I just want to be absolutely sure it shouldn't be simplified before I merge it. :-)

module/VuFind/src/VuFind/Resolver/Driver/Jop.php
demiankatz: Thanks!

module/VuFind/src/VuFind/Resolver/Driver/Jop.php
robertlange81: Yes, You were right. I might have thought to cumbersome and can no longer understand my thoughts from back then, so I made xpathTitleSelector a regular property like the others.

demiankatz: Great, thanks! I believe this is ready to merge now.

demiankatz (approved):

PR #2457, opened 2022-06-06
Move RTL styling comment out of HTML into PHP code.

EreMaijala:
It was also output even if RTL styling was not used.


demiankatz (approved):

demiankatz:
Thanks!

PR #2458, opened 2022-06-06
Fix double-escaping of title in layout.phtml.

EreMaijala:
The unescaped string is needed for construction of the full title to avoid double-escaping the page-specific title part.


demiankatz (approved):
Good catch -- thanks!

PR #2460, opened 2022-06-06
[VUFIND-1558] Prevent duplicate searches from being saved

demiankatz:
TODO
- [x] Test that fix also works with notification frequency preferences
- [x] Add test coverage
- [x] Update changelog (note saveSearch signature change) when merging
- [x] Close [VUFIND-1558](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1558) when merging


demiankatz:
@EreMaijala, this is still work in progress, but I'd be interested to hear your thoughts on my approach so far before I invest further time in it. Thanks!

module/VuFind/src/VuFind/Db/Table/Search.php
```diff
@@ -250,34 +244,65 @@ public function saveSearch(
...
+            // Update the old search only if it wasn't saved:
+            if (!$oldSearch->saved) {
+                $oldSearch->created = date('Y-m-d H:i:s');
+                // Keep the ID of the old search:
+                $newSearchMinified = $normalized->getMinified();
+                $newSearchMinified->id = $oldSearch->getSearchObject()->id;
+                $oldSearch->search_object = serialize($newSearchMinified);
+                $oldSearch->save();
+            }
+            // Update the new search from the existing one
```
EreMaijala: Perhaps something like `// Update the search being saved with values from the existing search:` would make this more clear. Or add a mention that it's for the caller's benefit.

EreMaijala:
@demiankatz This makes sense to me. The only thing I'm a bit worried about is that the duplicate search row is deleted. If you then try to e.g. go back to edit it (provided it's an advanced search), it won't be found with the existing id. I'm not sure if there's a good solution for that without further complicating things, though. One way might be flagging the duplicate search hidden from history instead of deleting it, and then leave it to the search table cleanup to take the garbage out.

module/VuFind/src/VuFind/Db/Table/Search.php
demiankatz: See commit [15c216b](https://github.com/vufind-org/vufind/pull/2460/commits/15c216b2cfc7c781b8e81a4b8650d30f6c51176e) -- I renamed some variables and rewrote comments in an effort to make things a bit more clear, but I'm open to further suggestions.

demiankatz:
> @demiankatz This makes sense to me. The only thing I'm a bit worried about is that the duplicate search row is deleted. If you then try to e.g. go back to edit it (provided it's an advanced search), it won't be found with the existing id. I'm not sure if there's a good solution for that without further complicating things, though. One way might be flagging the duplicate search hidden from history instead of deleting it, and then leave it to the search table cleanup to take the garbage out.

Perhaps I'm overlooking something, but trying to edit an old duplicate advanced search seems like a fairly unlikely scenario unless the user is going deep into their browser history or operating using multiple browser tabs at the same time. I think in more common situations, the existing code should work fine by matching up the URL query parameters with the equivalent search in the database. The only time we pass search IDs around is in the save/delete links (which should be ephemeral) and the "edit advanced search" link (which might cause problems in some rare cases, but the worst thing that happens is you hit the back button and end up on an advanced search screen that says it can't find your search... and if you hit back one more time from there and re-click edit, it will sort itself out).

A different/related problem that I encountered during testing of this scenario is that when you click "save" while not logged in, the login screen appears at the Save URL, which then adds an entry to your browser search history for the save URL... and then subsequently going back to that page can cause weird things to happen. I added commit 6fc69a0 to make the error in question a little less fatal-looking... but this seems like an existing issue if we're worried about clean browser history and avoiding weird behavior when going back. I mean, fundamentally we really shouldn't be using GET links to take actions at all... but addressing that might be a whole separate project.

module/VuFind/src/VuFind/Db/Table/Search.php
EreMaijala: Looks good to me now.

EreMaijala:
@demiankatz I agree the scenario is unlikely and I don't think it needs to be addressed.

demiankatz:
Thanks, @EreMaijala! I've now taken care of adding equivalent logic to the scheduled search functionality -- it has some of the same pitfalls in that duplicate searches will be deleted and could in rare cases cause odd back button behavior... but I tried to minimize this possibility as much as I could. Additionally, if you have saved notifications, perform a new search, and set notification preferences prior to logging in, your new preference will overwrite your old one. I didn't see this as a problem, but maybe it's cause for some concern.

If you're happy with this, I'll do the final step of adding test coverage; I just didn't want to do the work of pinning everything down in case you suggest significant changes.

EreMaijala (approved):
Looks good to me, and I think overwriting the schedule is expected behavior. Another option would be to disable setting schedule altogether until logged in, but I think that the current behavior allows more streamlined functionality.

demiankatz:
Thanks, @EreMaijala! I've just added Mink tests for all three of the scenarios this PR is designed to correct for. I also re-targeted this PR to dev-9.0 so I could work against the latest versions of the tests.

Next week, I'll try to find time to write unit tests for the newly-added classes, and then I think this is ready to merge. Let me know if you'd like me to flag you for a re-review to look at the tests, or if you want me to just go ahead and merge as long as they pass. :-)

EreMaijala:
@demiankatz Unless the tests uncover something that needs fixing in the core code, feel free to go ahead merge. :)

PR #2462, opened 2022-06-06
Czech translation for 8.1

xmorave2:
- Adds Czech translation for VuFind 8.1
- Does not add translations for DDC - it is usually not used in Czech libraries, almost all libraries here use UDC, so there is now complete translation for DDC as far as I know


demiankatz (approved):

demiankatz:
Thanks, @xmorave2!

PR #2463, opened 2022-06-07
Allow filtering of optional text domains in language dev tool

demiankatz:
VuFind's language files now include some specialized text domains that may not be desired in every language (e.g. call number schemes). This PR adds a parameter to filter the language dev tool to focus in on the "most important" non-optional strings, in case we need to prioritize translation work.

The user interface is very crude and can likely be improved. It might also make sense to offer something more robust where the user can specify exactly which text domains to exclude. This is just a rough first pass proof of concept.


crhallberg (approved):
The toggle is very clear and the functionality looks good!

PR #2464, opened 2022-06-08
Fix holds problem in FOLIO Lotus release.

demiankatz:
This PR fixes a compatibility issue introduced by the FOLIO Lotus release and discussed on [CIRC-1546](https://issues.folio.org/browse/CIRC-1546). That ticket also discusses backward-breaking API changes that will be corrected in a FOLIO hotfix. Since that is being treated as a FOLIO bug, adding complexity to VuFind to work around it does not seem worthwhile.

This PR originally had some work to add title-level hold support, but since that feature is not fully functional yet on the FOLIO side, it has been separated into PR #2479.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -1087,7 +1087,8 @@ class Folio extends AbstractAPI implements
                 'id' => $this->getBibId(null, null, $hold->itemId),
                 'item_id' => $hold->itemId,
                 'reqnum' => $hold->id,
-                'title' => $hold->item->title
+                // Title moved from item to instance in Lotus release:
+                'title' => $hold->instance->title ?? $hold->item->title ?? '',
             ];
         }
         return $holds;


demiankatz:
Robert Wilson has kindly tested these changes against Kiwi and found that they break backward compatibility, resulting in this error (with identifiers fictionalized):

`Unrecognized field "holdingsRecordId" (class org.folio.rest.jaxrs.model.Request), not marked as ignorable (24 known properties: "awaitingPickupRequestClosedDate", "patronComments", "fulfilmentPreference", "holdShelfExpirationDate", "requestType", "requestExpirationDate", "itemId", "metadata", "status", "item", "proxyUserId", "deliveryAddressTypeId", "requesterId", "tags", "id", "requester", "requestDate", "cancelledDate", "cancellationReasonId", "position", "cancelledByUserId", "pickupServicePointId", "cancellationAdditionalInformation", "proxy"]) at [Source: (String)"{"itemId":"xxx","holdingsRecordId":"yyy","instanceId":"zzz","requestLevel":"Item","requestType":"Page","requesterId":"aaaa","requestDate":"2022-06-09T09:47:15-05:00","fulfilmentPreference":"Hold Shelf","requestExpirationDate":"2022-09-10T04:59:59.000Z","pickupServicePointId":"bbb","status":"Open - Not yet filled","item":{"title":"Time"[truncated 656 chars]; line: 1, column: 70] (through reference chain: org.folio.rest.jaxrs.model.Request["holdingsRecordId"])`

This raises some questions about how to proceed. Do we need to put some kind of version number in FOLIO.ini so that we can adjust behavior in a version-specific way? Do we just need to state that certain versions of the driver are only compatible with certain versions of FOLIO? On the FOLIO side, should this change be considered a bug, or is it a conscious change? If this is a conscious change, how can we plan for more changes like this, and avoid future breakages?

demiankatz:
I suppose another option could be to just add a switch to FOLIO.ini to control whether or not we send more detailed hold requests to FOLIO, with a note that it needs to be on for Lotus and off for Kiwi. That's not very elegant, though -- I'll wait and see what develops on the FOLIO JIRA ticket before making a decision (though I'm open to input from others in the meantime).

demiankatz:
I've just revised the top-level comment here -- please take a look there to see the current state of this PR and some recent decisions.

demiankatz:
@bbusenius or @mis306lu (or anyone else with an interest in VuFind/FOLIO integration), could somebody please give me an approval on this so I can merge it (assuming you don't see any problems/concerns)?

bbusenius (approved):
This looks good to me. I tested it against Lotus. Also, you keyed me in to a couple of problems I might have.

demiankatz:
Thanks, @bbusenius!

PR #2465, opened 2022-06-08
Add DOIs to MLA citations.

demiankatz:
This PR adds DOIs to MLA citations -- they were excluded before because many common examples exclude them, but the style does support their inclusion. Here is an example of how this looks for an article:

image

...and here's a book example (note that the DOI here is fictitious):

image

Thanks to @sturkel89 for background research on this project.


mtrojan-ub (approved):
Thanks for taking care of this!

PR #2466, opened 2022-06-09
Use doi.org instead of dx.doi.org in all citations.

demiankatz:
- dx.doi.org was deprecated in 2017; see https://www.crossref.org/display-guidelines/
- thanks to @sturkel89 for guidance on this issue


demiankatz:
@mtrojan-ub, here's another very small citation issue that came up in discussion following #2465 -- we were using an outdated DOI base URL in some situations. This PR brings us up to current standards.

mtrojan-ub (approved):

PR #2467, opened 2022-06-09
Add ingest settings for InTech Open.

demiankatz:
Import rules for InTech Open (a source of open access scientific books and other materials) was contributed to the wiki many years ago but never made its way into the core project. This PR is an effort to update that work and make it official so it can be more consistently maintained.

TODO
- [x] Implement author name formatting
- [x] Implement ISBN indexing
- [x] Implement DOI indexing
- [x] Implement format mapping (if necessary/useful)
- [x] Add test coverage for new XSLT methods
- [x] Update [wiki page](https://vufind.org/wiki/indexing:open_data_sources:intech) when merging


demiankatz:
@xmorave2 / @EreMaijala, I'm opening this PR to move some suggested import rules out of the wiki and into the code so that it is easier to keep up to date. I had to significantly rewrite the circa-2012 code to work in modern VuFind. I think this should be ready to merge, and it should be very low-risk since it won't break anything if people don't use it... but I'm hoping to get a quick review from somebody so I can merge it. Thanks!

xmorave2 (changes requested):
Looking into intech data, for example here: http://mts.intechopen.com/oai/?verb=ListRecords&metadataPrefix=oai_dc&set=978-953-7619-59-6

Sometimes the names are inverted and sometimes not. We probably need to test if there is `,` character in name and not invert name parts in that case.

demiankatz:
@xmorave2, do you think the check should go into the XSLT, or into the invertNames function? (I'd be inclined to put it in the XSLT, though I'm not sure what the best syntax would be... maybe we need an isInvertedName function that takes edge cases like ", Jr." into consideration).

It's also worth noting that some of the InTech Open records have clear bad data in them -- some records have a `<dc:creator>,</dc:creator>` value while others have `<dc:creator>Jr</dc:creator>`. Obviously we can't fix things that are broken upstream!

xmorave2:
@demiankatz `isInvertedName` function looks like good idea as it adds flexibility to call it from XSL or PHP code.

But I agree we should not to try to fix data issues coming from external source.

demiankatz:
@xmorave2, I've added inverted name detection in [657a8f5](https://github.com/vufind-org/vufind/pull/2467/commits/657a8f5654b3087da73a4df017ecd70a0bebcbf6). I'm sure this is imperfect and there are still some problematic edge cases, but it should significantly improve the overall accuracy. What do you think?

xmorave2 (approved):
@demiankatz I do agree. This is much better, and good enough - we can't do miracles on bad (and inconsistent) data. And it could be easily expanded if needed. I think we could merge it now.

PR #2468, opened 2022-06-10
Disable HTTP keep-alive in KohaRest and SierraRest.

EreMaijala:
With Koha, keep-alive has been found to cause the HTTP connection to intermittently stall e.g. when retrieving holds, which does multiple requests. With Sierra, this is precautionary.

This is unfortunate since keep-alive should be faster, but I also encountered issues when trying it with the Solr connection, so I think it will be safer to disable it for the time being.


demiankatz (approved):

demiankatz:
Thanks, @EreMaijala!

PR #2469, opened 2022-06-14
KohaRest: Add a workaround for a mapping error in enumchron field of …

EreMaijala:
…checkouts.


demiankatz (approved):
Thanks for the fix!

PR #2470, opened 2022-06-20
Add support for OAuth2 and OpenID Connect providers. 

EreMaijala:
This PR adds support for VuFind to act as an identity provider for external services using OpenID Connect (OIDC).

This can be very useful e.g. when VuFind provides access to ebooks etc. with a provider that needs to have some information about the user, such as birthdate, to control access. OIDC gives the user control over release of their information, which helps also with the GDPR requirements.

The PR does not currently include any functionality to remember and/or revoke user consents, so users will need to allow access every time. Consent management could be added in a follow-up task.

A large part of the code is plumbing for league/oauth2-server. The implementation loosely follows oauth2-server-example and is also inspired by Drupal's simple_oauth module.

[idp-oidc-tester](https://medium.com/application-security/openid-connect-tester-for-developers-63427324faea) is a very useful tool for testing the OIDC provider. Example to run it with docker: `docker run --rm --name idp-oidc-tester -p 8080:80 registry.gitlab.com/guenoledc-perso/idp-oidc-tester:latest` (make sure that the service in Docker can access token, jwks and user info endpoints; with Mac use e.g. http://docker.for.mac.localhost/vufind2/OAuth2/token).

TODO:
- [x] Add a scope for patron's blocks
- [x] Tests
- [x] Complete birthdate support in a separate PR (#2502)
- [x] [Wiki page](https://vufind.org/wiki/configuration:oauth2_oidc)
- [x] Identify and fix any places where the code could be cleaned up
- [x] Add and test Postgresql support


languages/en.ini
```diff
@@ -826,6 +826,23 @@ Number = "Number"
 number_decimal_point = "."
 number_thousands_separator = ","
 OAI Server = "OAI Server"
+oauth2_allow_access = "Allow Access"
```
demiankatz: I wonder if we should use a more generic prefix than oauth2 -- I assume that this text would be the same for any sort of external permission system, whether the technology is oauth2 or something else. Might be more future-proof to not tie the translation strings to a specific protocol version. Maybe "external_auth_..."?

demiankatz (commented):
I'll give this a deeper review later in the week, but I had five minutes just now and started with one minor comment... :-)

languages/en.ini
EreMaijala: Makes sense, I'll revise. :)

config/vufind/OAuth2Server.yaml
```diff
@@ -0,0 +1,110 @@
...
+  family_name: lastname
+  email: email
+  birthdate: birthdate
+  locale: last_language
+  phone: phone
+  address: address_json # special field including all address fields
+# Grants configuration
+Grants:
+  # Life time for an authorization code (should be kept short). Default is 1 minute.
+  authCodeLifeTime: PT1M
```
demiankatz, line 155: I'm not familiar with the `PT...` syntax. Does this need to be explained in the comments?

languages/en.ini
```diff
@@ -410,8 +410,25 @@ export_selected_favorites = "Export Selected Items"
...
 export_success = "Export Ready"
 export_unsupported_format = "Unsupported Export Format"
+external_auth_access_heading = "Request for Access"
+external_auth_access_login_message = "An external application is requesting access to your data. Please log in to review and authorize the request."
+external_auth_allow_access = "Allow Access"
+external_auth_deny_access = "Deny Access"
 external_auth_heading = "Access to licensed material"
+external_auth_ils_prompt = "Some of the requested information is retrieved from your library catalog profile."
 external_auth_login_message = "Login to access licensed material"
+external_auth_prompt_html = "is requesting the following access rights:"
```
demiankatz, line 427: This seems to be missing the client name token.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -904,7 +907,8 @@ public function getMyProfile($patron)
             'phone'           => '1900 CALL ME',
             'mobile_phone'    => '1234567890',
             'group'           => 'Library Staff',
-            'expiration_date' => 'Someday'
+            'expiration_date' => 'Someday',
+            'birthdate'       => $birthDate->format('Y-m-d')
```
demiankatz: I don't think birthdate is currently part of the driver spec. Would it make sense to add this as a separate PR?

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -554,6 +554,7 @@ public function getMyProfile($patron)
             'address1' => $address,
             'zip' => $zip,
             'city' => $city,
+            'birthdate' => $result['birthDate'] ?? '',
```
demiankatz: Is the format of this guaranteed across all Sierra installations?

demiankatz (changes requested):
Thanks for the progress here. I still haven't completed a full review yet, but I found a few more things to comment on as I continued chipping away at the list of files.

config/vufind/OAuth2Server.yaml
EreMaijala: Comment expanded.

languages/en.ini
EreMaijala: The token was a leftover. The client name is in the heading.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: Yes, and that was my intention. I expanded the TODO item.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Yes. See https://sandbox.iii.com/iii/sierra-api/swagger/index.html#!/patrons/Get_a_list_of_patrons_get_1. It's been stable across API versions.

module/VuFind/src/VuFind/Controller/OAuth2ControllerFactory.php
```diff
@@ -0,0 +1,289 @@
...
+
+    /**
+     * Return a factory function for creating the authorization server.
+     *
+     * @return callable
+     */
+    protected function getAuthorizationServerFactory(): callable
+    {
+        return function (?string $clientId): AuthorizationServer {
+            if (!($keyPath = $this->oauth2Config['Server']['privateKeyPath'] ?? '')
```
demiankatz: Adding a `getOauth2ServerSetting` method might make the code a little more concise, since you could centralize the value checking and exception throwing to a single place.

module/VuFind/src/VuFind/Controller/OAuth2ControllerFactory.php
```diff
@@ -0,0 +1,289 @@
...
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options sent to factory.');
+        }
+        $yamlReader = $container->get(\VuFind\Config\YamlReader::class);
+        $oauth2Config = $yamlReader->get('OAuth2Server.yaml');
+
+        $this->container = $container;
+        $this->oauth2Config = $oauth2Config;
+        $tablePluginManager = $this->container
```
demiankatz: Any particular reason why you use `$this->container` here but `$container` everywhere else in this method?

module/VuFind/src/VuFind/Controller/OAuth2Controller.php
```diff
@@ -0,0 +1,438 @@
...
+                Psr7ServerRequest::fromLaminas($this->getRequest())
+            );
+        } catch (OAuthServerException $exception) {
+            return $this->convertOAuthServerExceptionToResponse($exception);
+        } catch (\Exception $e) {
+            return $this->handleException('Authorization request', $e);
+        }
+
+        if ($this->formWasSubmitted('allow') || $this->formWasSubmitted('deny')
+        ) {
```
demiankatz: Does this line really need to be wrapped?

module/VuFind/src/VuFind/Controller/OAuth2Controller.php
```diff
@@ -0,0 +1,438 @@
...
+            if (!$this->csrf->isValid($this->getRequest()->getPost()->get('csrf'))) {
+                throw new \VuFind\Exception\BadRequest(
+                    'error_inconsistent_parameters'
+                );
+            }
+
+            // Store OpenID nonce in the access token table so that it can be
+            // retrieved for token or user info action:
+            $this->accessTokenTable
+                ->storeNonce($user->id, $laminasRequest->getQuery('nonce'));
```
demiankatz, line 233: Do we need to worry about the nonce being absent?

module/VuFind/src/VuFind/Controller/OAuth2ControllerFactory.php
```diff
@@ -0,0 +1,289 @@
...
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options sent to factory.');
+        }
+        $yamlReader = $container->get(\VuFind\Config\YamlReader::class);
+        $oauth2Config = $yamlReader->get('OAuth2Server.yaml');
```
demiankatz: You could save a line of code by assigning directly to `$this->oauth2Config` here and then changing the access to `$oauth2Config` below.

module/VuFind/sql/mysql.sql
```diff
@@ -346,3 +346,21 @@ CREATE TABLE `auth_hash` (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE utf8mb4_unicode_ci;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
+--
+-- Table structure for table `access_tokens`
+--
+
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8mb4 */;
+CREATE TABLE `access_token` (
```
demiankatz, line 382: We should make sure that PostgreSQL support is on the TODO list -- though I'm fine with doing that toward the end of the project so we don't have to spend time porting anything that changes along the way.

module/VuFind/src/VuFind/OAuth2/Repository/ScopeRepository.php
```diff
@@ -0,0 +1,98 @@
...
+     * Return information about a scope.
+     *
+     * @param string $identifier The scope identifier
+     *
+     * @return ScopeEntityInterface|null
+     */
+    public function getScopeEntityByIdentifier($identifier)
+    {
+        if (!($config = $this->oauth2Config['Scopes'][$identifier] ?? [])) {
+            $config['hidden'] = true;
```
demiankatz: Should this actually be `$config = ['hidden' => true]` in case we get a non-array value out of the configuration, or is that impossible?

module/VuFind/src/VuFind/Controller/OAuth2ControllerFactory.php
```diff
@@ -0,0 +1,289 @@
...
+            $encryptionKey
+                = trim($this->oauth2Config['Server']['encryptionKey'] ?? '');
+            if (!$encryptionKey) {
+                throw new \Exception(
+                    'Server/encryptionKey missing from OAuth2Server.yaml'
+                );
+            }
+
+            $server = new AuthorizationServer(
+                new ClientRepository($this->oauth2Config),
```
demiankatz: Should we have separate getters (or perhaps properties containing class names) for the various repositories and other VuFind-specific classes so it's easier to locally override them?

module/VuFind/src/VuFind/OAuth2/Entity/ScopeEntity.php
```diff
@@ -0,0 +1,151 @@
...
+    public function __construct(array $config)
+    {
+        foreach (['identifier'] as $required) {
+            if (!isset($config[$required])) {
+                throw new \Exception("OAuth2 scope config missing '$required'");
+            }
+        }
+        $this->setIdentifier($config['identifier']);
+        $this->setDescription($config['description'] ?? '');
+        $this->setILSNeeded(boolval($config['ils'] ?? false));
```
demiankatz: Any strong reason to use `boolval` instead of `(bool)` typecasting? Nothing wrong with it, but I feel like it's not how we usually force values to booleans in the code base.

demiankatz (commented):
I don't claim to know a lot of the Oauth2/OpenID-specific details, but I gave this a general read-through and had a few (mostly minor stylistic) suggestions.

module/VuFind/src/VuFind/Controller/OAuth2Controller.php
EreMaijala: storeNonce allows it to be null as well, which would be fine. I'll update the comment to clarify that.

module/VuFind/src/VuFind/Controller/OAuth2Controller.php
EreMaijala: No, fixed.

module/VuFind/src/VuFind/Controller/OAuth2ControllerFactory.php
EreMaijala: No, fixed.

module/VuFind/src/VuFind/Controller/OAuth2ControllerFactory.php
EreMaijala: Fixed.

module/VuFind/sql/mysql.sql
EreMaijala: I added a TODO entry and will add it.

module/VuFind/src/VuFind/OAuth2/Repository/ScopeRepository.php
EreMaijala: Yes, that's better. Fixed.

module/VuFind/src/VuFind/Controller/OAuth2ControllerFactory.php
EreMaijala: I added some getters and then decided that it's better to actually utilize the service manager, so now it all goes through it. I also split some methods to smaller parts for easier overriding.

module/VuFind/src/VuFind/OAuth2/Entity/ScopeEntity.php
EreMaijala: No, changed.

module/VuFind/src/VuFind/Controller/OAuth2ControllerFactory.php
demiankatz: Thanks, @EreMaijala, this is definitely cleaner now. I did momentarily wonder if we should have an Oauth2-specific plugin manager to avoid putting so many narrow-purpose definitions into the top-level service manager, but given that these classes don't have a common base class or interface, I think what you've done makes the most sense. If we wanted to separate these concerns from the rest of the code, maybe it would actually make the most sense to factor them to a separate VuFind\Oauth2 Laminas module... but I don't think that's something we need to contemplate right now, unless it appeals to you.

EreMaijala: I thought about a plugin manager, but it didn't feel appropriate. A separate module is worth a thought, but I feel it would be so intertwined with the base module that it gets complicated easily, particularly if the functionality is expanded in the future so that it'd be possible to remember and manage access grants.

demiankatz: Agreed -- worth consideration, but not an urgent priority right now. We can see how this evolves and refactor later if justified.

config/vufind/OAuth2Server.yaml
```diff
@@ -0,0 +1,159 @@
+# OAuth2 Server Configuration
```
demiankatz, line 1: Maybe we should provide a little more context here, like a link to a wiki page, so somebody curious about the purpose of this file can quickly figure it out.

config/vufind/OAuth2Server.yaml
```diff
@@ -0,0 +1,159 @@
...
+
+# Authorization server configuration
+Server:
+  # Create keys e.g. as above and point the following setting to it. The paths can be
+  # absolute or relative to the config directory.
+  privateKeyPath: "oauth2_private.key"
+  publicKeyPath: "oauth2_public.key"
+  # Encryption key used to encrypt payloads. Make sure it is random and long enough
+  # (at least 32 characters).
+  encryptionKey: ""
```
demiankatz, line 22: Since we're defaulting to blank here, will the software fail if the user does not fill it in? I agree that it's a good idea not to provide a default key, since that could compromise security... just want to be sure there's a check to ensure we don't go keyless by accident.

demiankatz: (Same for the salt below).

demiankatz (commented):
I still have more to review, but a few comments from progress I made today...

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
```diff
@@ -58,6 +59,13 @@ abstract class MinkTestCase extends \PHPUnit\Framework\TestCase
      */
     protected $modifiedConfigs = [];
 
+    /**
+     * Modified yaml configurations
+     *
+     * @var array
+     */
+    protected $modifiedYamlConfigs = [];
```
demiankatz: This YAML functionality is a nice addition; maybe we should commit it separately from the rest. I can merge it early to dev if you like, just to reduce files in this PR.

module/VuFind/src/VuFind/Controller/OAuth2Controller.php
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    /**
+     * Get a JSON response from an array of data
+     *
+     * @param array $data Data to encode
+     *
+     * @return Response
+     */
+    protected function getJsonResponse(array $data): Response
```
demiankatz: Might it make sense to put some of these utility methods like getJsonResponse and addCorsHeaders into a trait for reuse? I'd eventually like to clean up the existing controller class hierarchy by moving more logic out of the base class and into traits... maybe this is an opportunity to begin to move in that direction.

module/VuFind/src/VuFind/OAuth2/Repository/ScopeRepository.php
```diff
@@ -0,0 +1,99 @@
...
+     * Given a client, grant type and optional user identifier validate the set of
+     * scopes requested are valid and optionally append additional scopes or remove
+     * requested scopes.
+     *
+     * @param ScopeEntityInterface[] $scopes         Scopes
+     * @param string                 $grantType      Grant type
+     * @param ClientEntityInterface  $clientEntity   Client
+     * @param null|string            $userIdentifier User ID
+     *
+     * @return ScopeEntityInterface[]
```
demiankatz, line 89: Do we need an unused parameters annotation here?

config/vufind/OAuth2Server.yaml
EreMaijala: `getOAuth2ServerSetting` will ensure that there is at least something in the key. It doesn't ensure the key is great, but I think it's okay to leave that to the administrator. Also the grants that encrypt something will throw if the key is empty. And yes, leaving the default empty helps avoid [trouble](﻿https://it.slashdot.org/story/22/08/22/227245/hyundai-uses-example-keys-for-encryption-system).

EreMaijala: And salt is also retrieved via `getOAuth2ServerSetting` that ensures it's set.

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
EreMaijala: Sure, please do!

demiankatz:
Okay, @EreMaijala, I think that completes my review... but on the understanding that I'm not familiar with the OAuth2 libraries and did not study that code in great depth -- I focused mainly on the changes to existing code and skimmed the rest. If we merge, we should check the coverage report and make sure everything new is fully tested (if you haven't already looked at coverage reports locally, of course). We also need to remember to adapt all this when we eventually merge it from dev into #2233.

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
demiankatz: Done: commit c906c31584dcdb8b33eb3bfc804159ee8e89b30d.

module/VuFind/src/VuFind/OAuth2/Repository/ScopeRepository.php
EreMaijala: Yes, added.

module/VuFind/src/VuFind/Controller/OAuth2Controller.php
EreMaijala: Makes sense. Trait and accompanying tests added.

EreMaijala:
@demiankatz Thanks! All issues should be resolved now. I'll try to verify the test coverage once more.

EreMaijala:
@demiankatz Code coverage for unit tests looks good to me.

demiankatz (approved):
Thanks, @EreMaijala!

PR #2473, opened 2022-06-21
Improve MARC DOI extraction.

demiankatz:
This PR offers more sophisticated, regex-driven DOI extraction -- this method makes it possible to extract DOIs from a wider variety of URLs without having to add many lines of prefix-based checking. It also works better when URLs have proxy prefixes included.

TODO
- [x] Add URL decoding capabilities


demiankatz:
@mtrojan-ub / @EreMaijala, it was brought to my attention that VuFind's default MARC DOI extraction left quite a bit to be desired - in addition to failing to account for variations of the DOI base URLs, it also did not properly URL-decode DOIs containing special characters. I believe that this PR should fix both of those issues. What do you think?

EreMaijala (approved):
Looks good to me, but I'll defer final judgement to @mtrojan-ub.

import/index_java/src/org/vufind/index/DoiTools.java
```diff
@@ -44,7 +64,36 @@ public Set<String> getDoiFromUrl(final Record record, String fieldSpec, String b
...
+     * Extract DOIs from URLs matching the specified regular expression
+     * @param record MARC record
+     * @param fieldSpec taglist for URL fields
+     * @param regEx Regular expression for matching URLs
+     * @param groupIndex Index of the grouped expression in regEx containing the DOI to extract (passed as a String instead of an Integer due to SolrMarc limitations)
+     * @return Set of DOIs
+     */
+    public Set<String> getDoisFromUrlWithRegEx(final Record record, String fieldSpec, String regEx, String groupIndex) {
+        // Build the regular expression:
+        Pattern pattern = Pattern.compile(regEx);
```
mtrojan-ub: This method might be called millions of times, and it will always compile the same pattern over and over again. Can we use some kind of cache for compiled patterns?

Or maybe we could use something like the CachedPattern class in this article:
https://stackoverflow.com/questions/13420321/does-pattern-compile-cache

(We should also keep in mind that the cache should be thread-safe)

mtrojan-ub (changes requested):
see comment related to CachedPattern

import/index_java/src/org/vufind/index/DoiTools.java
demiankatz: Thanks, @mtrojan-ub -- what do you think of [e24ee32](https://github.com/vufind-org/vufind/pull/2473/commits/e24ee32bc84a7e2660fb9e55846121e259557e2f)?

mtrojan-ub: That's definitely better! However, for an optimal Java-like solution you might wanna use computeIfAbsent() instead of a hardcoded if ... get ... else ... put block, because it's the most elementary thread-safe operation. This is already used in several other parts of the code, here's an example:

https://github.com/vufind-org/vufind/blob/f738bbb07733504c2995f40bd900bc4b73b62e78/import/index_java/src/org/vufind/index/WorkKeys.java#L71

https://github.com/vufind-org/vufind/blob/e6ece437af9b322f7f712b947e4371a435b91a83/import/index_java/src/org/vufind/index/ConfigManager.java#L179

So it should basically be one line of code:
`Pattern pattern = patternCache.computeIfAbsent(compiledPattern -> Pattern.compile(regEx));`

demiankatz: Good point, @mtrojan-ub, that's a lot nicer -- see [1d867dc](https://github.com/vufind-org/vufind/pull/2473/commits/1d867dce856f4e194835ceb8b93b8f1428059a37).

demiankatz: (Also accidentally introduced unwanted whitespace into the comment in that commit -- fixed in a subsequent commit; sorry about that!)

mtrojan-ub: Looks fine now, thanks!

mtrojan-ub (approved):

import/index_java/src/org/vufind/index/DoiTools.java
demiankatz: Thanks again for the very helpful input, @mtrojan-ub!

PR #2474, opened 2022-06-22
Fix search tabs versions.

EreMaijala:
Using bare `$showCounts` will cause a notice when it's not passed to the renderer.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2475, opened 2022-06-22
Fix hierarchical facet markup.

demiankatz:
Hierarchical facets previously used an inappropriate `<li>` tag as a placeholder. This PR changes it to a `<div>`, which is legal in context. No functionality or styling appears to be impacted by the change.

This is a follow-up to discussion on #2454.


EreMaijala (approved):
Looks good to me!

PR #2477, opened 2022-06-23
Remove no-JS support from cart add/remove.

EreMaijala:
It was breaking the HTML markup while adding quite a bit of markup to each result.

Mink tests are passing.

TODO
- [x] Create "legacy" branch so feature can be reinstated if needed

### DIFF
--- a/themes/bootstrap3/templates/record/cart-buttons.phtml
+++ b/themes/bootstrap3/templates/record/cart-buttons.phtml
@@ -8,15 +8,5 @@
     <a href="#" class="cart-remove hidden<?php if($cart->contains($cartId)): ?> correct<?php endif ?>">
       <span class="cart-link-label"><?=$this->icon('cart-remove') ?><?=$this->transEsc('Remove from Book Bag') ?></span>
     </a>
-    <noscript>
-      <form method="post" name="addForm" action="<?=$this->url('cart-processor')?>">
-        <input type="hidden" name="ids[]" value="<?=$this->escapeHtmlAttr($cartId)?>" />
-        <?php if ($cart->contains($cartId)): ?>
-          <input class="btn btn-default" type="submit" name="delete" value="<?=$this->transEscAttr('Remove from Book Bag')?>"/>
-        <?php else: ?>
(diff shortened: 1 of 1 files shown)


ckaz:
Nice!! We have something like this in our PR pipeline: We kept the form element but removed the <noscript> braces like so:

`      <form method="post" name="addForm" action="<?=$this->url('cart-processor')?>" class="cartProcessorNoJs">
        <input type="hidden" name="ids[]" value="<?=$this->escapeHtmlAttr($cartId)?>" />
        <?php if ($cart->contains($cartId)): ?>
          <input class="btn btn-default" type="submit" name="delete" value="<?=$this->transEsc('Remove from Book Bag')?>"/>
        <?php else: ?>
          <input class="btn btn-default" type="submit" name="add" value="<?=$this->transEsc('Add to Book Bag')?>"/>
        <?php endif; ?>
      </form>`

@RLangeUni then added the `class="cartProcessorNoJs"` and the following script for better W3C validation:

`<script>$(document).ready(function() { $(".cartProcessorNoJs").css('display', 'none'); });</script>`

Sorry, I can't make a proper commit of this right now -- I'm on the run.

demiankatz:
Thanks, @EreMaijala. I think it's probably safe to remove this functionality -- the need to run websites with no Javascript support in the current era of the web doesn't seem particularly realistic, so I think it's probably safe to remove some of these things in places where they potentially cause problems. Would you agree, @ckaz and @crhallberg?

My proposal is that I merge this PR, then create a legacy/no-js-bookbag branch which re-adds it, so it can be easily reinstated if anyone needs it. In theory, that legacy branch could incorporate @ckaz's proposed improvement (though that approach seems potentially prone to a flash of unstyled content).

crhallberg (approved):
Makes sense to keep it in a branch as a legacy feature.

demiankatz:
Thanks, @EreMaijala and @crhallberg. I have merged this and created a legacy branch here: https://github.com/vufind-org/vufind/tree/legacy/nojs-cart

PR #2480, opened 2022-06-29
Confirm functionality to renewals.

ghost:
TODO:

- [x] Run full test suite once design is finalized


demiankatz:
Thanks for the contribution, @aidaluuppala! I'm asking @crhallberg to look at this from a technical/stylistic perspective. However, I'm also curious about the motivation behind the change: why do we need a confirmation on renewals? Since this is essentially a non-destructive action, is there a strong reason to add an extra step to the process? This seems doubly true for the "renew selected" feature, which is already a multi-step process, since the user has to select one or more items before pressing the button.

demiankatz:
@aidaluuppala, some Czech translations of your new strings were contributed via #2482. Since this PR is still in progress, I thought it would be best to move them here, but I do not have write permission to your branch, so I have opened https://github.com/aidaluuppala/NDL-VuFind2/pull/1 -- can you please merge that when time permits?

ghost:
> Thanks for the contribution, @aidaluuppala! I'm asking @crhallberg to look at this from a technical/stylistic perspective. However, I'm also curious about the motivation behind the change: why do we need a confirmation on renewals? Since this is essentially a non-destructive action, is there a strong reason to add an extra step to the process? This seems doubly true for the "renew selected" feature, which is already a multi-step process, since the user has to select one or more items before pressing the button.

Thank you for your comments @demiankatz, they are very good points. The motivation behind the change came from user feedback at Finna. Customers are used to having the confirm functionality in the holds and wished for it for the renewals as well to prevent accidental renewals. I asked about the selected renewals confirmation from our UI designer when I started this development, and we came to conclusion that sense the selected holds have the confirmation, maybe it would be consistent for the users if the renewals worked with same logic. But however, your point about the fact that renewals are a non-destructive action and that user clicks the items to renew in selection, maybe the renew selected doesn't need confirmation. The main concern for our users is the renew all confirmation.

demiankatz:
Thanks, @aidaluuppala, that's a good point regarding consistency with holds functionality. Perhaps this is an argument for removing the confirmation from "cancel selected holds," though I guess that's a separate conversation.

In any case, I appreciate the additional context! Now I'll wait and see what @crhallberg has to say. :-)

crhallberg (approved):
Looks good to me on the tech and style end.

demiankatz:
Thanks for the review, @crhallberg. I think we might as well go forward with this as-is. If we want to remove confirmation for the "selected" options, we could do that in a separate PR and apply the change everywhere that it's relevant.

However, before we can merge this, we need to adjust the test suite to reflect the new functionality. Right now, I'm getting one test failure:

```
1) VuFindTest\Mink\IlsActionsTest::testRenewal
Element not found: .alert.alert-danger index 0

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:290
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/IlsActionsTest.php:556
```

@aidaluuppala, are you interested in adjusting the test? If so, I'll leave it to you, but if you don't have time, or want to see an example before you try it yourself, let me know and I can work on it next week.

ghost:
Thank you for the review @crhallberg @demiankatz and sorry for the late reply, I was on a summer holiday. :) I'm interested in adjusting the tests and I can try and do that.

demiankatz (approved):
Welcome back, @aidaluuppala, and thanks for fixing the tests. I think this is all ready to go now!

EreMaijala:
For the record: renewals are only almost non-destructive since there's often a maximum number of times an item can be renewed. Losing your last chance of renewal accidentally would be irritating. Additionally, it's quite easy to humble a click with a mobile device.

PR #2481, opened 2022-06-29
Show request counts in MyResearch menu

ghost:
TODO:

- [x] Run full test suite once design is finalized


demiankatz:
Thanks for this contribution as well, @aidaluuppala! Once again, I'm requesting input from @crhallberg when he has a chance.

Here's how this looks on my test server with the Demo driver activated:

image

I have a few questions:

1.) If we're displaying counts, do we really need the bell icon, or could the numbers potentially supersede the icon?

2.) Why is the hold count in a round border, and the checked out count in a square border?

3.) Is there any reason not to apply consistent behavior to the storage retrieval and ILL requests? It seems to me that all three of these should behave the same way, once we decide on exactly what behavior is desired.

4.) Do we potentially want a yellow-tinted count of requests that are in the system but not yet ready for pickup, to provide a fuller context? I think perhaps the original idea here was that you only see a notification if something is ready... but if we start displaying numbers, maybe it's confusing if we only display a count of a subset.

Thanks again!

ghost:
> Thanks for this contribution as well, @aidaluuppala! Once again, I'm requesting input from @crhallberg when he has a chance.
> 
> Here's how this looks on my test server with the Demo driver activated:
> 
> image
> 
> I have a few questions:
> 
> 1.) If we're displaying counts, do we really need the bell icon, or could the numbers potentially supersede the icon?
> 
> 2.) Why is the hold count in a round border, and the checked out count in a square border?
> 
> 3.) Is there any reason not to apply consistent behavior to the storage retrieval and ILL requests? It seems to me that all three of these should behave the same way, once we decide on exactly what behavior is desired.
> 
> 4.) Do we potentially want a yellow-tinted count of requests that are in the system but not yet ready for pickup, to provide a fuller context? I think perhaps the original idea here was that you only see a notification if something is ready... but if we start displaying numbers, maybe it's confusing if we only display a count of a subset.
> 
> Thanks again!

Thank you @demiankatz for your questions and excellent points.

1.) Very good point. Our designers at Finna wished for the count to be next to the bell icon, but since the checked outs don't have the icon either and I agree that the count supersedes the icon and leaves the icon redundant. I can remove it in this pull request.

2.) The count is in a round border, because of the fines are in round border. But it's a very good point that it should be more similar with checked outs. I'll change it to square border.

3.) Our designers asked for only the holds to have the count, but it's excellent point that for consistency it should be in the storage retrieval and ILL requests as well. I think for our users the holds are more used and that's why the count was wished only for holds. But considering consistency and other organizations possible needs, I think counts should be added to storage retrieval and ILL requests too. I can add them in this pull request.

4.) Excellent point, our designers asked only for the available holds count, because it's the information our users are most interested in, but I agree that it's confusing if we don't show the other numbers as well. I can add them in this pull request. And if they are not needed in our theme, they can be hidden from there only at our end.

demiankatz:
Thanks, @aidaluuppala, that all makes sense and sounds great. I'll await your next revision and will then give this another look.

ghost:
> Thanks, @aidaluuppala, that all makes sense and sounds great. I'll await your next revision and will then give this another look.

Thank you @demiankatz, I implemented the changes we discussed.

demiankatz:
Thanks, @aidaluuppala, I think this is looking very good. Just a couple more comments:

1.) When making changes to Less files, it's a good idea to run `grunt lessToSass` and `grunt less` and commit the results to the PR. Otherwise, the continuous integration will fail. I assume you have an environment where you can do this, but if you need any help, let me know!

2.) Do you think we should have a red box with the count of all the remaining requests that are not either available or in transit?  If you think it's worth doing, it would probably need a new translation string (maybe something like "Still pending" or "Other status"). I'm mainly thinking of the scenario where somebody places a new request and maybe expects to see it reflected in a count somewhere. But maybe we're only interested in surfacing the numbers that are immediately actionable. Just something to think about!

ghost:
Thanks @demiankatz
1) I ran the grunt commands and will commit the results with other changes.
2) Good point, I think it's worth doing and will add them as "Other status" in this pull request.

module/VuFind/src/VuFind/AjaxHandler/AbstractUserRequestAction.php
```diff
@@ -67,14 +67,16 @@ public function handleRequest(Params $params)
...
         foreach ($requests as $request) {
-            if ($request['available'] ?? false) {
+            if (!empty($request['available'])) {
                 $status['available'] += 1;
-            }
-            if ($request['in_transit'] ?? false) {
+            } elseif (!empty($request['in_transit'])) {
                 $status['in_transit'] += 1;
+            } else {
+                $status['other'] += 1;
```
demiankatz: Tiny nitpick, but we could use `++` instead of all these `+= 1` as a slight optimization. :-)

themes/bootstrap3/js/account_ajax.js
```diff
@@ -223,86 +223,90 @@ $(document).ready(function registerAccountAjax() {
...
-      } else if (status.in_transit > 0) {
-        $element.html(
-          statusIcon("my-account-warning", "text-warning", "account_requests_in_transit")
-        );
+      }
+      if (status.in_transit > 0) {
+        html += '<span class="badge warn" data-toggle="tooltip" title="' + VuFind.translate('account_requests_in_transit') + '">' + status.in_transit + '</span>';
+      }
+      if (status.other > 0) {
+        html += '<span class="badge overdue" data-toggle="tooltip" title="' + VuFind.translate('account_requests_other') + '">' + status.other + '</span>';
```
demiankatz: I wonder if we should create a style with a more generic name than `overdue`, since `overdue` isn't really accurate in this situation. Maybe something like `alert`?

demiankatz (commented):
Thanks, this is looking great! A couple of small suggestions; also hoping to hear more from @crhallberg, but that probably won't happen until next week (since it's the Independence Day holiday this weekend and we'll be out of the office until Tuesday).

themes/bootstrap3/js/account_ajax.js
ckaz: @aidaluuppala Good idea, this! We use something more of a flat design and would much prefer the item numbers over the bell icon. The numbers convey much more information than the bell and the entire menu would look more consistent.
my-account-bell
.

crhallberg (approved):

demiankatz:
@aidaluuppala, it looks like we have @crhallberg's and @ckaz's approval, so I'm happy to merge this soon -- but before I do, did you have any thoughts about my final round of suggestions?

ghost:
Thank you for the review @crhallberg @ckaz @demiankatz and sorry for the late reply, I was on a summer holiday. :) Good points, I agree and will implement the final round of suggestions.

aidaluuppala:
Thank you for the review @crhallberg @demiankatz and sorry for the late reply, I was on a summer holiday. :) I'm interested in adjusting the tests and I will try to do that.

ghost:
@demiankatz I implemented the suggestions, but I used class name danger, because alert was already taken.

demiankatz:
Thanks, @aidaluuppala! A couple of observations:

1.) The red "overdue" box has been replaced with a grey "danger" box. Was this intentional, or did the color change by accident?

Before:
image

After: 
image

2.) We have rounded badges for the numbers on requests, but square boxes for the numbers on checked out items. Is there a reason for this inconsistency? The rounded badges seem to have a problem where they get stretched out for items that have a two-line link in them. See:

image

3.) Since we're changing the .overdue class to .danger, I wonder if we should consider retaining the .overdue styles as an alternative for legacy compatibility. (My inclination is to say no, and just put a note in the changelog... but I wonder if @crhallberg or others feel differently).

ghost:
Thanks @demiankatz, good that you noticed, 1.) and 2.) were both accidental. They seem to have happened because of the name change of the class. I'll changed it again.

demiankatz:
Thanks, @aidaluuppala, this looks good now, and the test suite is passing.

However, there is one very small problem: it looks like you didn't commit the updates to compiled.css after running `grunt less`, so I had to do this on my end to get everything up to date. Can you please commit those changes and push them up? (I would do it myself, but I don't have write access to your branch, and creating a separate PR for this seemed overcomplicated).

Once the CSS is updated in this PR, I'm happy to merge it as long as @crhallberg has no concerns.

ghost:
Thanks @demiankatz, I committed and pushed the compiled css.

demiankatz:
Thanks, @aidaluuppala! I've re-requested review from @crhallberg just in case he has any thoughts on naming/backward compatibility, and once I get the all-clear from him, I'll merge this.

demiankatz (approved):

themes/bootstrap3/js/account_ajax.js
```diff
@@ -211,7 +211,7 @@ $(document).ready(function registerAccountAjax() {
         level = ICON_LEVELS.WARNING;
       }
       if (status.overdue > 0) {
-        html += '<span class="badge overdue" data-toggle="tooltip" title="' + VuFind.translate('account_checkouts_overdue') + '">' + status.overdue + '</span>';
+        html += '<span class="badge account-alert" data-toggle="tooltip" title="' + VuFind.translate('account_checkouts_overdue') + '">' + status.overdue + '</span>';
```
crhallberg, line 214: If we're going to give the badges clearer names, we should also change `.ok` and `.warn` to something suitable (`.account-good` and `.account-warning`?)

crhallberg (dismissed):

themes/bootstrap3/js/account_ajax.js
demiankatz: For simplicity's sake, would you object to merging this and then renaming the other two classes in a separate PR?

As far as names go, I think `account-warning` makes sense... would `account-info` or `account-ok` be a better alternative to `account-good`? (I don't have strong feelings about this, just brainstorming options).

crhallberg: No objection as long as it is not forgotten.

ghost: I renamed the badges to `account-warning` and `account-info` in this PR, so that they won't be forgotten.

demiankatz:
Thanks for the updates, @aidaluuppala. Unfortunately, though, the CSS doesn't seem to have compiled correctly -- it looks like maybe some kind of encoding problem was introduced. I have recompiled the CSS and merged the latest dev branch to fix the problem -- see https://github.com/aidaluuppala/NDL-VuFind2/pull/2. If you can merge that into here, I think we can finally wrap this up!

I'm curious what process / software versions you're using to compile your CSS; hopefully we can identify what went wrong here so that it won't be an issue again in the future.

ghost:
Sorry I used https://lesscss.org/, so that probably caused the problem. I will use grunt in the future.

demiankatz:
Thanks, @aidaluuppala! I believe that our grunt task has a few extra processing steps, so running the compiler directly probably skipped those, which is why things came out a little weird. We're planning to improve our tooling in the future, but for now it's definitely best to use the grunt version to compile CSS. In any case, I'm going to merge this now. Thanks for your patience through the process of getting everything polished! :-)

PR #2482, opened 2022-06-29
Update cs.ini

black23:
Added new renewal texts.


demiankatz (approved):
Thanks for this! Since two of the strings here are not yet part of the dev branch, I have moved them into PR #2480 where the relevant functionality is being developed. That way, if anything changes, we can keep all the work in sync. I'm merging the remaining strings to dev now.

PR #2483, opened 2022-07-01
Add !default flags to accessibility.scss

ckaz:
This adds missing !default flags to variables in components/accessibility.scss. The !default flags are necessary to allow overwriting variables in inheriting themes.

All releases since 6.x are affected.

### DIFF
--- a/themes/bootstrap3/less/components/accessibility.less
+++ b/themes/bootstrap3/less/components/accessibility.less
@@ -8,12 +8,35 @@ a {
 }
 
 /* AAA Color Contrasts */
+/* #LESS> */
 @badge-bg: #595959;
+/* <#LESS */
+/* #SCSS>
+$badge-bg: #595959 !default;
+<#SCSS */
 
+/* #LESS> */
 @breadcrumb-color: #535353;
-@breadcrumb-active-color: #444;
+/* <#LESS */
+/* #SCSS>
+@breadcrumb-color: #535353 !default;
+<#SCSS */
 
+/* #LESS> */
+@breadcrumb-active-color: #444;
+/* <#LESS */
+/* #SCSS>
+@breadcrumb-active-color: #444 !default;
+<#SCSS */
+
+/* #LESS> */
 @state-danger-text: #8a211e;
+/* <#LESS */
+/* #SCSS>
+@state-danger-text: #8a211e !default;
+<#SCSS */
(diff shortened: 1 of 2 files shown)


demiankatz:
@ckaz, the .scss files are automatically generated using a script that converts them from LESS... so if something is wrong with them, we probably need to adjust the conversion process rather than editing directly. (This is the reason the build is failing here: it checks to confirm that the .scss files are in sync with the .less files by using the auto-conversion, and since you made manual changes, they don't match up). @crhallberg, can you look into this?

ckaz:
Apologies for not getting this in the first place. Yes, I guess it might be something to do with the LESS to SCSS conversion -- I just can't find where the required `!default` string is added to variables.
Happy holiday!

crhallberg:
We should be able to fix this for SASS conversion by using the comment flags we created.

You can see an example at the top of [bootstrap.less](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/less/bootstrap.less).

```less
/* #LESS> */
$badge-bg: #595959;
/* <#LESS */
/* #SCSS>
$badge-bg: #595959 !default;
<#SCSS */
```

demiankatz:
Thanks, @crhallberg! @ckaz, please let us know if you need help making these adjustments.

ckaz:
> We should be able to fix this for SASS conversion by using the comment flags we created.
> 
> You can see an example at the top of [bootstrap.less](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/less/bootstrap.less).
> 
> ```css-less
> /* #LESS> */
> $badge-bg: #595959;
> /* <#LESS */
> /* #SCSS>
> $badge-bg: #595959 !default;
> <#SCSS */
> ```

Neat! Didn't know it was so easy. Will make a proper PR!

demiankatz:
Thanks, @ckaz -- remember that you can just push changes to this existing branch to update this PR; that might be simpler than opening a whole new one. :-)

ckaz:
Apologies for the huge delay. Just updated this by removing the flags from accessibility.scss and adding the comments to the less file.
Cheers!

demiankatz (approved):
Thanks, @ckaz. I just ran `grunt lessToSass` to get everything back in sync, and I think this is ready to merge now!

PR #2484, opened 2022-07-01
VuFind 1564   add due date to folio driver

cwolfebsco:
Work relates to:
https://openlibraryfoundation.atlassian.net/browse/VUFIND-1564

Current State:
Folio ILS Driver does not return a dueDate, as a result when you view a record that is checked out you are not given any information as to when the item is expected back

New State:
Folio ILS Driver returns a dueDate (j F Y format, e.g. 2 July 2022) and Folio will display the dueDate in the detailed record

Screenshot 2022-07-01 at 10 18 11
Note: screenshot shows d F Y formatting

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -544,6 +544,11 @@ class Folio extends AbstractAPI implements
      */
     public function getHolding($bibId, array $patron = null, array $options = [])
     {
+        $showDueDate = $this->config['Availability']['showDueDate'] ?? true;
+        $showTime = $this->config['Availability']['showTime'] ?? true;
+        $maxNumDueDateItems = $this->config['Availability']['maxNumberItems'] ?? 5;
+        $dueDateItemCount = 0;
+
         $instance = $this->getInstanceByBibId($bibId);
         $query = [
             'query' => '(instanceId=="' . $instance->id
@@ -608,7 +613,10 @@ class Folio extends AbstractAPI implements
                 );
 
                 $dueDateValue = '';
-                if ($item->status->name == 'Checked out') {
+                if ($item->status->name == 'Checked out'
+                    && $showDueDate
+                    && $dueDateItemCount < $maxNumDueDateItems
+                ) {
                     // extract itemId & construct query
                     $query = [
                         'query' => 'itemId==' . $item->id
@@ -618,16 +626,29 @@ class Folio extends AbstractAPI implements
                         'loans',
                         '/circulation/loans',
                         $query
-                    ) as $loan) { 
-                        $dueDate = new DateTime(
-                            $loan->dueDate,
-                            new DateTimeZone('UTC')
-                        );
-                        $loc = (new DateTime)->getTimezone();
-                        $dueDate->setTimezone($loc);
-                        $dueDateValue = $this->dateConverter
-                            ->convertToDisplayDateAndTime('U', $dueDate->format('U'));
+                    ) as $loan) {
+                        if (!isset($loan->returnDate)) {
+                            $dueDate = new DateTime(
+                                $loan->dueDate,
+                                new DateTimeZone('UTC')
+                            );
+                            $localTimezone = (new DateTime)->getTimezone();
+                            $dueDate->setTimezone($localTimezone);
+                            $dueDateValue = $this->dateConverter
+                                ->convertToDisplayDate(
+                                    'U',
+                                    $dueDate->format('U')
+                                );
+                            if ($showTime) {
+                                $dueDateValue = $this->dateConverter
+                                    ->convertToDisplayDateAndTime(
+                                        'U',
+                                        $dueDate->format('U')
+                                    );
+                            }
+                        }
                     }
+                    $dueDateItemCount += 1;
                 }
 
                 $items[] = $callNumberData + [
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -27,6 +27,8 @@
  */
 namespace VuFind\ILS\Driver;
 
+use DateTime;
+use DateTimeZone;
 use Exception;
 use VuFind\Exception\ILS as ILSException;
 use VuFind\I18n\Translator\TranslatorAwareInterface;
@@ -605,7 +607,10 @@ class Folio extends AbstractAPI implements
                     $item->itemLevelCallNumber ?? ''
                 );
 
-                $dueDate = new DateTime($item->status->date, new DateTimeZone('UTC'));
+                $dueDate = new DateTime(
+                    $item->status->date,
+                    new DateTimeZone('UTC')
+                );
                 $loc = (new DateTime)->getTimezone();
                 $dueDate->setTimezone($loc);
 
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -607,12 +607,16 @@ class Folio extends AbstractAPI implements
                     $item->itemLevelCallNumber ?? ''
                 );
 
-                $dueDate = new DateTime(
-                    $item->status->date,
-                    new DateTimeZone('UTC')
-                );
-                $loc = (new DateTime)->getTimezone();
-                $dueDate->setTimezone($loc);
+                $dueDateValue = '';
+                if ($item->status->name == 'Checked out') {
+                    $dueDate = new DateTime(
+                        $item->status->date,
+                        new DateTimeZone('UTC')
+                    );
+                    $loc = (new DateTime)->getTimezone();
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -604,13 +606,22 @@ public function getHolding($bibId, array $patron = null, array $options = [])
                     $item->itemLevelCallNumberPrefix ?? '',
                     $item->itemLevelCallNumber ?? ''
                 );
+
+                $dueDate = new DateTime(
```
demiankatz: Should we do this conditionally, only when $item->status->date is set? (I assume many items will not have a due date, if they are not checked out).

demiankatz (commented):
Thanks, @cwolfebsco! I have just one question (see separate comment). I'm out of the office until Tuesday because of the Independence Day holiday, but I'll test this out in more detail once I'm back.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
cwolfebsco: I had found a date on every item that I had checked and thus made an assumption it would always be there

`"status": {
                "name": "Available",
                "date": "2021-06-25T15:47:44.956+00:00"
            },`

Thanks to your question, I found that while this is true, there are other status messages that also returned a date and that will cause a problem:

`"status": {
                "name": "In process",
                "date": "2022-06-21T10:32:07.860+00:00"
            }`

I'll investigate this further, as I think it is less a function of whether the date is set, but whether the status is one for which we would want to show a date. Thanks for pointing this out, I had looked only at available and unavailable items during testing and it looked good - but I found a "in process" item that then showed a due date, that was obviously wrong

demiankatz: Thanks for the progress on this, @cwolfebsco. Just one suggestion: you might consider changing the assignment to $dueDateValue to read:

```
                    $dueDateValue = $this->dateConverter
                        ->convertToDisplayDateAndTime('U', $dueDate->format('U'));
```

This way, instead of a hard-coded display date format in the driver, this will respect the displayDateFormat/displayTimeFormat settings in config.ini. What do you think?

cwolfebsco: Makes perfect sense :) I'll submit a change...

cwolfebsco: BTW I just found a fallacy in this implementation during more testing :( - this needs more than just a small change, I am afraid

demiankatz: Sorry to hear that; let me know if I can be of any help with testing or general brainstorming.

cwolfebsco: Thanks for the offer, I think I need some brain storming here in a moment.

Basically I made a massive mistake as to where to get the dueDate from. In the code you will notice that I was trying to get it from $item->status->date and I had a screenshot attached, but it clearly shows 02 July 2021. I am obviously living in the past.

As far as I can tell the correct way is to call /circulation/loans with the itemId and get the dueDate from there.

I do however want to ask about the Universal date format:

For this time String:
2022-12-15T23:59:00.000+00:00. (<== UTC)

This is the output in my environment
15-12-2022 23:59 (<== currently London is BST = UTC+1, but December is correctly UTC)

Unless you have short term loans in your library, is the time truly desired? If so, would it make sense to let users configure the desired format in folio.ini? 
Screenshot 2022-07-05 at 19 04 27

demiankatz: @cwolfebsco, it's too bad that an additional API call is needed to fetch the due date; that could slow things down considerably, especially in cases where there are a lot of copies. We may want to incorporate some safeties (for example, a configurable limit on how many due dates to look up, perhaps -- which could be set to 0 to disable the feature entirely).

You're also right that showing date + time may not be desirable in many situations, though it can be important for short term loans. I guess that opens two possibilities:

Option 1: if we can detect a short term loan, we could show times only when appropriate. For example, might we be able to assume that 23:59 is never a useful time to display, but any other time should be displayed? Or is there some other attribute that could give us a clue?

Option 2: as you suggest, we could put in a setting in Folio.ini to control whether we use convertToDisplayDateAndTime or just plain convertToDisplayDate.

Thanks!

cwolfebsco: I had quick look at https://github.com/folio-org/edge-rtac to see if this could solve the problem, but the responses doesn't include the itemId, which Folio expects (see https://github.com/folio-org/edge-rtac/blob/master/ramls/holdings.xsd for definition of)

I might not find time to noodle this problem further this week - my apologies in advance

demiankatz: Is it possible that the id in the "holding" section of the RTAC response is actually an item ID rather than a holding ID, and that the "holding" section is simply misleadingly named? I don't understand why we would want to expose the holding ID for every item but not the item ID... but maybe there's a reason, or I'm just misreading something.

In any case, up to this point, we have avoided using the edge APIs in the FOLIO driver because they are optional, and we wanted to have the broadest possible compatibility. However, if it turns out that the edge API could offer a more efficient behavior, perhaps we could consider adding a Folio.ini setting to switch between edge and core APIs -- this would allow people the option of using edge if they have it, but would also provide a fallback if they didn't. But maybe that's more complexity than we need to bother with!

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -542,6 +544,11 @@ protected function chooseCallNumber($hCallNumP, $hCallNum, $iCallNumP, $iCallNum
      */
     public function getHolding($bibId, array $patron = null, array $options = [])
     {
+        $showDueDate = $this->config['Availability']['showDueDate'] ?? true;
+        $showTime = $this->config['Availability']['showTime'] ?? true;
```
demiankatz: Maybe the default in the code should match the default in Folio.ini for consistency.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -604,13 +611,54 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
                     $item->itemLevelCallNumber ?? ''
                 );
+
+                $dueDateValue = '';
+                if ($item->status->name == 'Checked out'
+                    && $showDueDate
+                    && $dueDateItemCount < $maxNumDueDateItems
+                ) {
+                    // extract itemId & construct query
+                    $query = [
```
demiankatz: It might improve readability to extract the logic here to a `protected getDueDate($item, $showTime)` method. It will not only make this function shorter and more readable, but it will also make it easier for people to customize the due date retrieval in local custom code if they wish to.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -604,13 +611,54 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
+                            if ($showTime) {
+                                $dueDateValue = $this->dateConverter
+                                    ->convertToDisplayDateAndTime(
+                                        'U',
+                                        $dueDate->format('U')
+                                    );
+                            }
+                        }
+                    }
+                    $dueDateItemCount += 1;
```
demiankatz: You could make this even shorter with `$dueDateItemCount++;`.

demiankatz (changes requested):
Thanks for the progress on this; it's looking great to me. Just a few suggestions for minor improvements. Let me know if you'd like me to help apply any of the suggestions. I'll do some hands-on testing once the code has been fully updated.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
cwolfebsco: Absolutely, sorry that this mismatched - I'll set showTime to false by default

module/VuFind/src/VuFind/ILS/Driver/Folio.php
cwolfebsco: You are totally right, I do a lot of programming in Swift these days, where the ++ increment was never implemented, as it was deemed ambiguous. So I unlearned a good thing :( I'll update that

module/VuFind/src/VuFind/ILS/Driver/Folio.php
cwolfebsco: Very fair point and very true. I'll commit code soon, would appreciate your review and your testing in your live system, I did test those changes, of course, but you know how it goes - more eyes = more better (sorry for the grammar)

demiankatz:
Thanks, @cwolfebsco, the changes look great! I ran out of time before I could do hands-on testing, but I plan to work on that first thing tomorrow morning. I'll let you know how it goes!

demiankatz (approved):
Thanks again, @cwolfebsco, this looks great and is working for me.

PR #2485, opened 2022-07-01
Handle empty index related dates in admin module

xmorave2:
Last modified date could be empty in newly created index without ever any record indexed


languages/en.ini
```diff
@@ -617,6 +617,7 @@ In This Collection = "In This Collection"
 in_collection_label = "In collection:"
 include_synonyms = "Expand results using synonyms"
 Index Terms = "Index Terms"
+index_unknown_datetime = "Unknown"
```
demiankatz: There's an existing "Unknown" translation. You could either use that, or else use the console tool to copy language strings to populate all the language files with it.

themes/bootstrap3/templates/admin/home.phtml
```diff
@@ -33,12 +33,26 @@
         <tr>
           <th><?=$this->transEsc('Start Time')?>: </th>
           <?php $startTime = $core->xpath('//lst[@name="' . $coreName . '"]/date[@name="startTime"]') ?>
-          <td><?=$this->escapeHtml($this->dateTime()->convertToDisplayDateAndTime(DATE_RFC3339_EXTENDED, (string)array_pop($startTime)))?></td>
+          <td>
+            <?php $startTime = (string)array_pop($startTime); ?>
+            <?php if (!empty($startTime)): ?>
```
demiankatz: Might be worth considering converting this to use a ternary for conciseness:

```
            <?=!empty($lastModified)
              ? $this->escapeHtml($this->dateTime()->convertToDisplayDateAndTime(DATE_RFC3339_EXTENDED, $lastModified))
              : $this->transEsc('index_unknown_datetime');
            ?>
```

demiankatz (commented):
Thanks! A couple of minor comments...

languages/en.ini
xmorave2: I would like to keep it different as it is used in specific context, so it may have different translations in some languages. So I copied the current translation to other language files. I manually removed the string from pt and pt-br translation, because it was obviously bad for this context (index_unknown_datetime = online does not look right)

themes/bootstrap3/templates/admin/home.phtml
xmorave2: Looks more readable to me, so I've changed it.

languages/en.ini
demiankatz: Sounds good. I've sent a note to our Portuguese translation volunteers to see if we can get this corrected -- thanks for bringing this to my attention. I suspect that this is an error that has existed since the early days of the project.

demiankatz:
@xmorave2, I think this is ready to merge, but before I do, one last question about the wording: it seems to me that if the last modified date is missing, it doesn't really mean "unknown" so much as "not yet modified." Would it be better to either create a different translation text (like "Never"), or avoid the translation issue entirely by using something like `-` to indicate no value?

demiankatz:
Another option would be to omit the "Last Modified" row entirely, which completely avoids the problem of language/presentation.

xmorave2:
I think using `-` character is good idea. It avoids using translation and keep the date if available. I've updated the patch this way.

demiankatz (approved):

demiankatz:
Great, thank you! Merging now!

PR #2486, opened 2022-07-01
Folio Driver | getMyTransactions | add correctly formatted dueDate and dueStatus

cwolfebsco:
Current State:
Folio Driver would not indicate whether an loaned item (from /circulation/loans endpoint) was due or overdue

New State:
Folio Driver will:
* consider an item due if it's dueDate timestamp is within 24 hours of now
* consider an item overdue, if now is greater than dueTime
* set false (neither due nor overdue) in all other cases

This will display bootstrap alert divs next to relevant loans, additionally it will update the pills in the my account menu appropriately

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -27,6 +27,8 @@
  */
 namespace VuFind\ILS\Driver;
 
+use DateTime;
+use DateTimeZone;
 use Exception;
 use VuFind\Exception\ILS as ILSException;
 use VuFind\I18n\Translator\TranslatorAwareInterface;
@@ -897,7 +899,6 @@ class Folio extends AbstractAPI implements
             '/circulation/loans',
             $query
         ) as $trans) {
-            
             $date = new DateTime($trans->dueDate, new DateTimeZone('UTC'));
             $loc = (new DateTime)->getTimezone();
             $date->setTimezone($loc);
@@ -906,10 +907,9 @@ class Folio extends AbstractAPI implements
             $tmpDueDate = strtotime($trans->dueDate);
 
             $now = time();
-            if($now > $tmpDueDate){
+            if ($now > $tmpDueDate) {
                 $dueStatus = 'overdue';
-            }
-            elseif($now > $tmpDueDate - (1 * 24 * 60 * 60)){
+            } elseif ($now > $tmpDueDate - (1 * 24 * 60 * 60)) {
                 $dueStatus = 'due';
             }
             $transactions[] = [
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -900,21 +900,29 @@ class Folio extends AbstractAPI implements
             $query
         ) as $trans) {
             $date = new DateTime($trans->dueDate, new DateTimeZone('UTC'));
-            $loc = (new DateTime)->getTimezone();
-            $date->setTimezone($loc);
+            $localTimezone = (new DateTime)->getTimezone();
+            $date->setTimezone($localTimezone);
 
             $dueStatus = false;
-            $tmpDueDate = $date->getTimestamp();
+            $dueDateTimestamp = $date->getTimestamp();
 
             $now = time();
-            if ($now > $tmpDueDate) {
+            if ($now > $dueDateTimestamp) {
                 $dueStatus = 'overdue';
-            } elseif ($now > $tmpDueDate - (1 * 24 * 60 * 60)) {
+            } elseif ($now > $dueDateTimestamp - (1 * 24 * 60 * 60)) {
                 $dueStatus = 'due';
             }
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -897,12 +899,23 @@ public function getMyTransactions($patron)
             '/circulation/loans',
             $query
         ) as $trans) {
-            $date = date_create($trans->dueDate);
+            $date = new DateTime($trans->dueDate, new DateTimeZone('UTC'));
+            $loc = (new DateTime)->getTimezone();
+            $date->setTimezone($loc);
+
+            $dueStatus = false;
+            $tmpDueDate = strtotime($trans->dueDate);
```
demiankatz: Should you derive $tmpDueDate from $date above, instead of going back to the raw $trans->dueDate? Just wondering if this impacts the time zone handling.

demiankatz (commented):
Thanks for this, @cwolfebsco -- just one question!

module/VuFind/src/VuFind/ILS/Driver/Folio.php
cwolfebsco: Another good question :) I was under the impression that the strtotime of a ISO8601 string (e.g. 2022-06-21T10:32:07.860+00:00) would always result in the correct Unix timestamp as the timezone was indicated at the end. On the other hand there would be no harm done by using `$tmpDueDate = $date->getTimestamp()`

I've had this snippet run in Trinity's VuFind for a while (but I must also admit that it originated from a much older version of the Driver). one to ponder for sure

demiankatz: You are probably correct that it will work either way, but I suspect switching it to use `$date->getTimestamp()` would be more efficient since it would eliminate the need to parse the date twice, which I'm sure is a more expensive operation than fetching an already-calculated timestamp.

cwolfebsco: I like the efficiency argument and I am rather embarrassed that I didn't think about that :) Thanks for pointing it out!

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -897,12 +899,23 @@ public function getMyTransactions($patron)
             '/circulation/loans',
             $query
         ) as $trans) {
-            $date = date_create($trans->dueDate);
+            $date = new DateTime($trans->dueDate, new DateTimeZone('UTC'));
+            $loc = (new DateTime)->getTimezone();
+            $date->setTimezone($loc);
+
+            $dueStatus = false;
+            $tmpDueDate = $date->getTimestamp();
```
demiankatz: Maybe `$dueDateTimestamp` would be a more clear name for this variable.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -897,12 +899,23 @@ public function getMyTransactions($patron)
...
+                $dueStatus = 'overdue';
+            } elseif ($now > $tmpDueDate - (1 * 24 * 60 * 60)) {
+                $dueStatus = 'due';
+            }
             $transactions[] = [
-                'duedate' => date_format($date, "j M Y"),
-                'dueTime' => date_format($date, "g:i:s a"),
-                // TODO: Due Status
-                // 'dueStatus' => $trans['itemId'],
+                'duedate' => $date->format('d F Y'),
```
demiankatz: Maybe we could here use `$this->dateConverter->convertToDisplayDate('U', $tmpDueDate)` and `$this->dateConverter->convertToDisplayTime('U', $tmpDueDate)` to take advantage of configurable date/time formats, similar to what we discussed on #2484. (And of course, please replace $tmpDueDate with the appropriate value if you choose to rename it per my suggestion above).

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -897,12 +899,23 @@ public function getMyTransactions($patron)
             '/circulation/loans',
             $query
         ) as $trans) {
-            $date = date_create($trans->dueDate);
+            $date = new DateTime($trans->dueDate, new DateTimeZone('UTC'));
+            $loc = (new DateTime)->getTimezone();
```
demiankatz: Maybe `$localTimezone` instead of just `$loc` for greater clarity?

demiankatz (changes requested):
Thanks for the progress on this. A few more very minor suggestions for polish, if you don't mind!

demiankatz:
Thanks, @cwolfebsco, this looks good! I'm going to do a little testing later today to make sure this works as expected in my test environment, and if all goes well, I will merge it.

demiankatz (approved):
I've confirmed that this corrects inaccurate dates in my test environment. Thanks again, @cwolfebsco!

demiankatz:
Since this is a bug fix, I've also backported it to the release-8.1 branch.

PR #2487, opened 2022-07-07
Tag admin improvements

demiankatz:
This PR simplifies some of the tag administration code and fixes minor bugs. It is inspired by work/discussion on #2399.

### DIFF
--- a/module/VuFindAdmin/src/VuFindAdmin/Controller/TagsController.php
+++ b/module/VuFindAdmin/src/VuFindAdmin/Controller/TagsController.php
@@ -191,7 +191,7 @@ class TagsController extends AbstractAdmin
     protected function getConfirmDeleteMessages($count)
     {
         // Default all messages to "All"; we'll make them more specific as needed:
-        $userMsg = $tagMsg = $resourceMsg = 'All';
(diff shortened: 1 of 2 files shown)


module/VuFindAdmin/src/VuFindAdmin/Controller/TagsController.php
```diff
@@ -197,7 +193,7 @@ protected function getConfirmDeleteMessages($count)
         // Default all messages to "All"; we'll make them more specific as needed:
         $userMsg = $tagMsg = $resourceMsg = 'All';
```
xmorave2: Could you translate the 'All' string here? I know that not all string in admin module are translated, but when you do refactoring here it should not harm ta add this minor change.

xmorave2 (changes requested):
Looks good to me, just one minor suggestion.

module/VuFindAdmin/src/VuFindAdmin/Controller/TagsController.php
demiankatz: Thanks for pointing this out -- should be fixed now.

xmorave2 (approved):
Thanks @demiankatz.

PR #2488, opened 2022-07-07
Add UDC (Universal Decimal Classification) support

demiankatz:
This PR is intended to add UDC support as an alternative to Dewey and LC. Currently there is a great deal of work still to do, so this is just a TODO list and some translations; work will be filled in as time permits.

TODO
- [x] Decide whether we need separate ones/tens/hundreds translations, or if we can use a unified language file for all of UDC
- [ ] Determine how to account for versioning (since different versions of the UDC standard have some significantly different assignments of numbers, and mixing them together will cause confusion)
- [ ] Add indexing support
- [ ] Add search support
- [ ] Add browse support
- [ ] Add alphabrowse support
- [ ] Add UDC namespaces to "optional" list in language devTool
- [ ] Obtain further translations


demiankatz:
On today's Community Call, @EreMaijala pointed out that different versions of UDC can have significantly different classifications. I've added a TODO checkbox above as a reminder that we should figure out a way to differentiate (or at least report) versions before we can fully implement this.

PR #2489, opened 2022-07-08
Adjust InstallController to verify PHP version based on composer.json

xmorave2:
We do not updated the minimal version of PHP in InstallController check.

This updates the version and refactors to have it defined at one place only in the controller. There is still second place where minimal PHP version is defined - in composer.json. So maybe it is a question if want it also in InstallController even if composer make this check during installation of dependencies.


demiankatz:
Thanks, @xmorave2! The main reason to keep the version check in the InstallController is that if a user installs VuFind using a .deb package or other pre-bundled archive, they do not run Composer, and so will bypass the version checks. This provides an additional layer where we can warn them. It might be worth considering revising this further to parse the version number out of composer.json, however. If we did that in the controller and added a unit test to confirm that the parsing works, I think we would be much less likely to forget this detail in the future. What do you think?

demiankatz:
I also committed a much [smaller/simpler fix](https://github.com/vufind-org/vufind/commit/2346d8a610507bb877dbebb8d561a7183693acf3) to fix the inaccurate message immediately for the next release, while we work out a better long-term solution here.

xmorave2:
@demiankatz I tried to implement reading PHP version from composer.json and added some basic tests. 

If the composer.json file is not found or does not contents php version (read from 'require' > 'php' with 'config' > 'platform' > 'php' as fallback) it throws exception.

We probably want to make exception class more specific (I am using  `\Exception` now). What do you think?

I Add also some basic tests, but they could be probably expanded with some more cases...

module/VuFind/src/VuFind/Controller/InstallController.php
```diff
@@ -922,4 +923,57 @@ public function homeAction()
...
+        if (empty($composer)) {
+            throw new \Exception('Cannot find composer.json');
+        }
+        $rawVersion = $composer['require']['php'] ?? '';
+        $version = preg_replace('/[^0-9. ]/', '', $rawVersion);
+        $version = preg_split('/[. ]/', $version);
+        if (count($version) >= 3) {
+            return sprintf('%d.%d.%d', $version[0], $version[1], $version[2]);
+        }
+        $version = $composer['config']['platform']['php'] ?? '';
```
demiankatz: Why do you do validation/normalization on `$composer['require']['php']` but not on `$composer['config']['platform']['php']`? Would it be better to just check both locations in the assignment to `$rawVersion` prior to normalization? I realize that one field may have special characters that the other does not, but I don't think it hurts to normalize both just to be safe.

module/VuFind/src/VuFind/Controller/InstallController.php
```diff
@@ -922,4 +923,57 @@ public function homeAction()
...
+    {
+        $composer = $this->getComposerJson();
+        if (empty($composer)) {
+            throw new \Exception('Cannot find composer.json');
+        }
+        $rawVersion = $composer['require']['php'] ?? '';
+        $version = preg_replace('/[^0-9. ]/', '', $rawVersion);
+        $version = preg_split('/[. ]/', $version);
+        if (count($version) >= 3) {
+            return sprintf('%d.%d.%d', $version[0], $version[1], $version[2]);
```
demiankatz: This might be a good place to use a spread operator; I think `return sprintf('%d.%d.%d', ...$version);` may achieve the same effect more concisely.

module/VuFind/src/VuFind/Controller/InstallController.php
```diff
@@ -922,4 +923,57 @@ public function homeAction()
...
+
+    /**
+     * Get minimal PHP version ID required for VuFind to run.
+     *
+     * @return int
+     */
+    protected function getMinimalPhpVersionId(): int
+    {
+        $version = explode('.', $this->getMinimalPhpVersion());
+        return $version[0] * 10000 + $version[1] * 100 + $version[2];
```
demiankatz, line 958: During normalization, should we pad the array with zeroes to ensure that at least three elements are set? e.g. if the PHP version is just set to "7" we want the array to be `[7, 0, 0]`. (Some cases demonstrating this scenario would be a good addition to your test suite, as you alluded to in your comments).

module/VuFind/tests/unit-tests/src/VuFindTest/Controller/InstallControllerTest.php
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Return method InstallController::getMinimalPhpVersion
+     *
+     * @return \ReflectionMethod
+     */
+    protected function getMinimalPhpVersionMethod(): \ReflectionMethod
+    {
+        $method = new \ReflectionMethod(
+            '\VuFind\Controller\InstallController',
```
demiankatz: I'd use `InstallController::class` here to be more direct.

module/VuFind/tests/unit-tests/src/VuFindTest/Controller/InstallControllerTest.php
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * Mock controller
+     *
+     * @param array $json JSON data
+     *
+     * @return InstallController
+     */
+    protected function mockController(array $json): InstallController
```
demiankatz: I'd rename this to getMockControllerWithComposerJson or something else more specific, since this method is only really relevant to tests that are working with composer.json, and if we expand the test suite, we may want to create different mock controllers for different purposes.

module/VuFind/src/VuFind/Controller/InstallController.php
```diff
@@ -227,8 +227,8 @@ protected function phpVersionIsNewEnough()
             return false;
         }
 
-        // We need at least PHP v7.3.0:
-        return PHP_VERSION_ID >= 70300;
+        // We need at least PHP version as defined in MINIMAL_PHP_VERSION:
```
demiankatz: This comment seems to suggest that there is a constant called MINIMAL_PHP_VERSION; maybe it should be revised.

demiankatz (commented):
Thanks, @xmorave2, this is coming along nicely. Just a few comments/suggestions.

Regarding exceptions, as you say, we might be able to use more specific ones, but I don't think it's essential. We throw basic exceptions in many places in the code, and in the unexpected situation that composer.json is missing at install time, I don't know that we need to do something more nuanced.

module/VuFind/src/VuFind/Controller/InstallController.php
xmorave2: Yes, that was the reason. But as you say, it does not hurt and simplifies the code a bit so I've changed it as you suggested.

module/VuFind/src/VuFind/Controller/InstallController.php
xmorave2: Oh, you are right, I completely missed this opportunity to use spread operator.

module/VuFind/src/VuFind/Controller/InstallController.php
xmorave2: It should never happen in this method. As if there aren't at least 3 parts of version, method `getMinimalPhpVersion` throws an exception. But it could be possible, that in the future we may want to define only one or two parts. So I expanded the ability of `getMinimalPhpVersion` to parse also 1 or 2 parts versions and added more test cases.

module/VuFind/tests/unit-tests/src/VuFindTest/Controller/InstallControllerTest.php
xmorave2: I'm almost sure I had it here, but it looks like I somehow messed it up... Nevermind, it is fixed now.

module/VuFind/tests/unit-tests/src/VuFindTest/Controller/InstallControllerTest.php
xmorave2: I agree it would be better name. Fixed.

module/VuFind/src/VuFind/Controller/InstallController.php
xmorave2: Fixed, now it references composer.json file.

xmorave2:
Thanks for review @demiankatz I tried to address all of your comments.

demiankatz (approved):
Thanks; this looks good to me now!

PR #2490, opened 2022-07-14
FOLIO driver: optimize date handling code

demiankatz:
After merging @cwolfebsco's latest improvements to FOLIO date handling, I noticed a couple of opportunities for code simplification. This PR simply optimizes the code without changing any functionality.

@cwolfebsco, would you mind reviewing/testing when you have a chance, just in case I've overlooked anything?

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -647,6 +647,21 @@ class Folio extends AbstractAPI implements
     }
 
     /**
+     * Convert a FOLIO date string to a DateTime object.
+     *
+     * @param string $str FOLIO date string
+     *
+     * @return DateTime
+     */
+    protected function getDateTimeFromString(string $str): DateTime
+    {
+        $dateTime = new DateTime($str, new DateTimeZone('UTC'));
+        $localTimezone = (new DateTime)->getTimezone();
+        $dateTime->setTimezone($localTimezone);
+        return $dateTime;
+    }
+
+    /**
      * Support method for getHolding(): obtaining the Due Date from OKAPI
      * by calling /circulation/loans with the item->id, adjusting the
      * timezone and formatting in universal time with or without due time
@@ -667,12 +682,7 @@ class Folio extends AbstractAPI implements
             // many loans are returned for an item, the one we want
             // is the one without a returnDate
             if (!isset($loan->returnDate) && isset($loan->dueDate)) {
-                $dueDate = new DateTime(
-                    $loan->dueDate,
-                    new DateTimeZone('UTC')
-                );
-                $localTimezone = (new DateTime)->getTimezone();
-                $dueDate->setTimezone($localTimezone);
+                $dueDate = $this->getDateTimeFromString($loan->dueDate);
                 $method = $showTime
                     ? 'convertToDisplayDateAndTime' : 'convertToDisplayDate';
                 return $this->dateConverter->$method('U', $dueDate->format('U'));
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -646,6 +646,21 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
     }
 
+    /**
+     * Convert a FOLIO date string to a DateTime object.
+     *
+     * @param string $str FOLIO date string
+     *
+     * @return DateTime
+     */
+    protected function getDateTimeFromString(string $str): DateTime
```
demiankatz, line 656: Created a shared support method since we did the same date conversion in two places.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -658,39 +673,22 @@ public function getHolding($bibId, array $patron = null, array $options = [])
      */
     protected function getDueDate($itemId, $showTime)
     {
-        $dueDateValue = '';
```
demiankatz, line 661: Removed the $dueDateValue string, since we can just return directly when we find a relevant value (which also potentially reduces the amount of looping we do below).

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -658,39 +673,22 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
-        $query = [
-            'query' => 'itemId==' . $itemId
-        ];
+        $query = 'itemId==' . $itemId;
         foreach ($this->getPagedResults(
             'loans',
             '/circulation/loans',
-            $query
+            compact('query')
```
demiankatz, line 680: Used compact to build an array here so we don't have to build it "by hand" above.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -658,39 +673,22 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
-                        $dueDate->format('U')
-                    );
-                if ($showTime) {
-                    $dueDateValue = $this->dateConverter
-                        ->convertToDisplayDateAndTime(
-                            'U',
-                            $dueDate->format('U')
-                        );
-                }
+            if (!isset($loan->returnDate) && isset($loan->dueDate)) {
```
demiankatz, line 684: Added a safety check so we don't try to access `$loan->dueDate` if it is undefined.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -658,39 +673,22 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
+                $dueDate = $this->getDateTimeFromString($loan->dueDate);
+                $method = $showTime
```
demiankatz, line 686: Used a variable method name instead of writing out two different calls with exactly the same parameters. This also prevents us from converting the date twice when `$showTime` is true (which was happening previously, but didn't need to).

demiankatz (commented):
Here's a summary of my changes and the reasoning behind them....

demiankatz:
@crhallberg, I'm requesting a review from you as well because @cwolfebsco is not currently a committer and so I can't request a review from him directly. :-)

cwolfebsco:
@demiankatz Thank you so much for making this code so much better and for your patience throughout the process. I tested with data from Trinity College and it worked as expected.

demiankatz:
Thanks, @cwolfebsco, I appreciate your help with this (and your receptiveness to feedback). :-)

PR #2491, opened 2022-07-18
Aleph: set Holds config defaults without erasing other settings.

demiankatz:
@welblaud discovered a problem where after #2306, some Holds-related configuration settings (e.g. the itemLimit for pagination, or the helpText settings) are lost by the Aleph driver. This PR refactors the code to set defaults without erasing additional settings. I was unable to test (since I don't have an Aleph system) but hope that @welblaud and/or @xmorave2 can try it out and let me know if I broke anything.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Aleph.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Aleph.php
@@ -1877,14 +1877,12 @@ class Aleph extends AbstractBase implements \Laminas\Log\LoggerAwareInterface,
     {
         if ($func == "Holds") {
             $holdsConfig = $this->config['Holds'] ?? [];
-            return [
-                "HMACKeys" => $holdsConfig['HMACKeys'] ?? "id:item_id",
-                "extraHoldFields"
-                    => $holdsConfig['extraHoldFields']
-                        ?? "comments:requiredByDate:pickUpLocation",
-                "defaultRequiredDate"
-                    => $holdsConfig['defaultRequiredDate'] ?? "0:1:0"
+            $defaults = [
+                'HMACKeys' => 'id:item_id',
+                'extraHoldFields' => 'comments:requiredByDate:pickUpLocation',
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -1877,14 +1877,12 @@ public function getConfig($func, $params = null)
...
-                    => $holdsConfig['extraHoldFields']
-                        ?? "comments:requiredByDate:pickUpLocation",
-                "defaultRequiredDate"
-                    => $holdsConfig['defaultRequiredDate'] ?? "0:1:0"
+            $defaults = [
+                'HMACKeys' => 'id:item_id',
+                'extraHoldFields' => 'comments:requiredByDate:pickUpLocation',
+                'defaultRequiredDate' => '0:1:0',
             ];
+            return $holdsConfig + $defaults;
```
janhejzl, line 1885: @demiankatz, does this actually mean `array_merge($holdsConfig, $defaults)`? 🤔

xmorave2: @welblaud not exactly, see this: https://www.texelate.co.uk/blog/combining-arrays-using-plus-versus-array-merge-in-php

You could read this code like: Get holds config from ini file, if there is any mandatory key missing, add it from defaults (but not overwrite it - that's how the plus operator works here)

xmorave2:
@demiankatz @welblaud I tried it on live library and it worked.

xmorave2 (approved):
The patch fixes issue and the code is much more readable. Thanks @demiankatz

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
janhejzl: Ok, got it, thanks for the explanation :)

demiankatz:
Thanks, everyone!

PR #2493, opened 2022-07-20
Adds explanatory text to filters (facets) for more accessibility

ckaz:
This is related to the other loss of focus issue (https://github.com/vufind-org/vufind/pull/2492): This is what our accessibility testers said: When selecting or deselecting facets/filters, the page reloads without warning which can be confusing to non-sighted users.

Our suggestion to this is fleshing out the sr-only and aria text hints in the sidebar and in the active filters list. 
So the desired behaviour is this:
Browsing filters/facets  ('Online', for example), the screen reader will read:
"Page will refresh when filter 'Online' is selected"

For selected facets/filters ('Online', for example) the screen reader will read:
"Applied Filter - Page will refresh when filter 'Online' is deselected"

For the active filters in the header, it should read something like:
"Remove filter - page will reload when filter is removed" and/or "Reset all filters - page will reload when all filters are reset"

Certainly this leaves room for further improvement/fine tuning.

According to our accessibility testers, ideally, the focus after reloading should be on the last selected filter/facet again but I have no idea as to how this can be accomplished. @lahmann mentioned the session store in the browser as a possibility, but as a front end person, this is beyond me.

Looking forward to what you think! Cheers!


demiankatz:
Thanks, @ckaz. Definitely interested in @crhallberg's input here. A couple of observations in the meantime:

1.) It looks like this branch has some conflicts with dev; let me know if you need help resolving them.

2.) I'm not entirely sure why page reloads are unexpected in the context of filters -- they are rendered as links, and isn't the normal behavior of a link a page reload? I could understand if we're specifically talking about checkbox filters (though even those are rendered as links rather than "real" checkboxes), but I'm not sure why this causes confusion in general. Is it possible there's some higher-level markup issue that could be improved to make this more clear, rather than having to put explanatory titles on everything to compensate for the behavior? As usual, I'm not entirely opposed to this, but I want to be sure we find the simplest possible solution to the problem. :-)

ckaz:
> 1.) It looks like this branch has some conflicts with dev; let me know if you need help resolving them.

Yes, sorry, I based in on 8.1 since our docker environment doesn't have a PhP 7.4.1 version yet. Do I need to push it to 8.1 then or can I pull dev into the PR?

> 2.) I'm not entirely sure why page reloads are unexpected in the context of filters -- they are rendered as links, and isn't the normal behavior of a link a page reload? I could understand if we're specifically talking about checkbox filters (though even those are rendered as links rather than "real" checkboxes), but I'm not sure why this causes confusion in general. Is it possible there's some higher-level markup issue that could be improved to make this more clear, rather than having to put explanatory titles on everything to compensate for the behavior? As usual, I'm not entirely opposed to this, but I want to be sure we find the simplest possible solution to the problem. :-)

My mistake -- I expressed this clumsily: The page reload as such isn't so much unexpected -- nevertheless, our testers found it safer to announce it to non-sighted users. If you activate a facet/filter by accident, your focus lands on top of the page and it takes to keyboard navigate back to where you were.
The loss of focus when you activate a facet/filter, however, is inacceptable. Non-sighted users can expect to have the focus returned to where they last operated on the page.

So our workaround suggested two things: 
1. It adds explanatory text to the facets so non-sighted users can have their screen readers read out to them what exactly the facet elements do (activate/exclude/add/remove ... + page reload). With this information, it is still annoying that the focus lands on top of the page and not on the facet activated, but a skip link to the facet list may shorten the effort required to get to where I was. 
[As a side note: Non-sighted users may have become familiar with ajax loading mechanisms by now so a full page reload isn't necessarily what they may expect when they hit a filter/facet.]
2. It was going to add a skip link that takes you to the facet list. However, the new active facets list in the header may render a skip link to facets less useful since it might be just as meaningful to wanting to skip to the active facets section. I guess, we would need to render the active facets as a live section that announces any changes when they occur (similar to the bookbag, but that's another PR) -- I'm thinking of `aria-live="polite"` and `aria-atomic="true" `or something in that vein.

A consensus was that when you use the catalog with eyes closed and a screen reader, it has parts that are almost in-navigable or impossible to imagine -- in particular if one hasn't _seen_ the entire thing before. 

Facets/filters, for example, announce themselves as chunks of terms. The facet list in the sidebar may be read as something like  "Refine Results -- heading level 2 -- Institution -- filled triangle pointing up -- opened -- Villanova University 36,896 -- link -- Library -- filled triangle pointing up -- opened -- Falvey Libary 36,896 -- link..." and so forth. Every time I hear this I go: What??

If we are aiming at an accessible catalog, I believe we should provide more guidance/explanatory text/hints. This could be useful for other people with other special needs to (legasthenics?). My suggestion would be something like (this example with exclude filters on)

"Refine Results -- heading level 2 -- Institution -- Please select from or exclude one or more of the following filters to narrow down search results:  -- heading level 3 -- dropdown menu opened -- List with ... entries -- Villanova University -- 36,896 results -- Select filter 'Villanova University' (Page will refresh when filter is selected). -- Exclude filter 'Villanova University': Page will refresh when filter is selected for exclusion from the search results."

Hope I've made myself a little clearer now. My best!

demiankatz:
> Yes, sorry, I based in on 8.1 since our docker environment doesn't have a PhP 7.4.1 version yet. Do I need to push it to 8.1 then or can I pull dev into the PR?

You should be able to merge dev into this PR without any trouble (though if that prevents you from doing further work on the code, I understand if you prefer to wait until after you get an updated Docker environment set up... it's just something we'll have to address before we can merge this).

> My mistake -- I expressed this clumsily: The page reload as such isn't so much unexpected -- nevertheless, our testers found it safer to announce it to non-sighted users. If you activate a facet/filter by accident, your focus lands on top of the page and it takes to keyboard navigate back to where you were. The loss of focus when you activate a facet/filter, however, is inacceptable. Non-sighted users can expect to have the focus returned to where they last operated on the page.

Is it possible that this could be addressed to some extent by setting up anchors and using them in the links? Would it be better if the various filter links anchored to either the "applied facets" area or the facet group from which they were chosen? That would not be difficult to implement, and I can see how it might be more convenient for everyone by drawing attention to the effect of the filter. Of course, it might also "feel weird." I would have to try it to have a stronger opinion one way or the other.

>     1. It adds explanatory text to the facets so non-sighted users can have their screen readers read out to them what exactly the facet elements do (activate/exclude/add/remove ... + page reload). With this information, it is still annoying that the focus lands on top of the page and not on the facet activated, but a skip link to the facet list may shorten the effort required to get to where I was.

If we jump to the applied filters list after applying a filter, would it be useful to have jump links for each facet field (like "skip to more values like this one")?

>        [As a side note: Non-sighted users may have become familiar with ajax loading mechanisms by now so a full page reload isn't necessarily what they may expect when they hit a filter/facet.]

That's a good point. In theory, I imagine there are some ways we could make the VuFind search results behave in a more AJAX-y way, though I'm not particularly eager to do this -- while it's counter-intuitive, I often find that single-page apps perform much worse than "old-fashioned" ones. I think the cost of dynamically redrawing large parts of the page through the DOM is significantly greater than a simple page reload, and it can make things laggy. (I could be mistaken, or just prejudiced, but I frequently feel like I'm wading through mud when trying to use really AJAX-heavy web apps).

>     2. It was going to add a skip link that takes you to the facet list. However, the new active facets list in the header may render a skip link to facets less useful since it might be just as meaningful to wanting to skip to the active facets section. I guess, we would need to render the active facets as a live section that announces any changes when they occur (similar to the bookbag, but that's another PR) -- I'm thinking of `aria-live="polite"` and `aria-atomic="true" `or something in that vein.

If there's markup we could use to improve the way this is communicated, I'm open to it -- but I'm assuming that "live section" markup wouldn't be meaningful unless we did an AJAX reimplementation of the search results page. Please correct me if I'm wrong!

> A consensus was that when you use the catalog with eyes closed and a screen reader, it has parts that are almost in-navigable or impossible to imagine -- in particular if one hasn't _seen_ the entire thing before.
> 
> Facets/filters, for example, announce themselves as chunks of terms. The facet list in the sidebar may be read as something like "Refine Results -- heading level 2 -- Institution -- filled triangle pointing up -- opened -- Villanova University 36,896 -- link -- Library -- filled triangle pointing up -- opened -- Falvey Libary 36,896 -- link..." and so forth. Every time I hear this I go: What??
> 
> If we are aiming at an accessible catalog, I believe we should provide more guidance/explanatory text/hints. This could be useful for other people with other special needs to (legasthenics?). My suggestion would be something like (this example with exclude filters on)
> 
> "Refine Results -- heading level 2 -- Institution -- Please select from or exclude one or more of the following filters to narrow down search results: -- heading level 3 -- dropdown menu opened -- List with ... entries -- Villanova University -- 36,896 results -- Select filter 'Villanova University' (Page will refresh when filter is selected). -- Exclude filter 'Villanova University': Page will refresh when filter is selected for exclusion from the search results."

Yes, I agree that it seems there is room for improvement here. I wonder if we really need two pull requests to address all of this: one to take care of the focus-related issues, and another to simply improve the readability of the facets. Maybe the two things are so intertwined that we should do it all here, but it also seems possible that there are some "low hanging fruit" we could get merged right away while we're discussing these more complex aspects of the problem (like the fact that reading out "filled triangle pointing up" is not going to help anyone, or that introducing alternate sr-only text to contextualize sections of the control and to explain the meaning of the numbers, will improve the experience). :-)

I do think that if we can figure out a technical way to make the focus behave more predictably, that is probably preferable to putting explanatory notes about page reloads all over the place. On pages with hundreds or thousands of facet values, just repeating that text so many times is going to impact the rendered page size and potentially slow things down... and while I realize that screen reader users are accustomed to rapidly skipping through values without listening to the whole thing every time, I still think it might be annoying to have the same message stated so many times. I'm still open to being persuaded otherwise, but this feels like the solution we should choose if we can't find a more elegant way, and I haven't given up hope on finding a more elegant way yet. :-)

What do you think?

ckaz:
> Is it possible that this could be addressed to some extent by setting up anchors and using them in the links? Would it be better if the various filter links anchored to either the "applied facets" area or the facet group from which they were chosen? That would not be difficult to implement, and I can see how it might be more convenient for everyone by drawing attention to the effect of the filter. Of course, it might also "feel weird." I would have to try it to have a stronger opinion one way or the other.

Sounds like a workable idea to me, and I guess many things feel weird until one gets used to them. It would at least be worth a try.


> In theory, I imagine there are some ways we could make the VuFind search results behave in a more AJAX-y way, though I'm not particularly eager to do this -- while it's counter-intuitive, I often find that single-page apps perform much worse than "old-fashioned" ones. I think the cost of dynamically redrawing large parts of the page through the DOM is significantly greater than a simple page reload, and it can make things laggy. (I could be mistaken, or just prejudiced, but I frequently feel like I'm wading through mud when trying to use really AJAX-heavy web apps).

I personally have a great dislike for AJAX web sites since you always get that weird delay untill more results plop in and you cannot search exhaustibly inside the page until the last item has loaded ...

> > 2. [...] I guess, we would need to render the active facets as a live section that announces any changes when they occur (similar to the bookbag, but that's another PR) -- I'm thinking of `aria-live="polite"` and `aria-atomic="true" `or something in that vein.
> > ```
> 
> If there's markup we could use to improve the way this is communicated, I'm open to it -- but I'm assuming that "live section" markup wouldn't be meaningful unless we did an AJAX reimplementation of the search results page. Please correct me if I'm wrong!

```
It works with standard pages. For our bookbag we use:

<ul class="nav navbar-nav navbar-right flip">
  <li id="cartSummary" aria-live="polite" aria-atomic="true">
    <a id="cartItems" class="btn" data-lightbox="" title="View Book Bag" href="/vufind2/local/Cart/Home">
      <i class="fa fa-clipboard" aria-hidden="true"></i>
      <span role="status">
        <span class="sr-only">Book Bag:</span> <strong>11</strong> <span class="cart-label">items</span>
        <span class="sr-only full hidden">(Your Book Bag is full)</span>
      </span>
    </a>
  </li>
```
So, when I add an item through the result list, the new number of items in the back is read out and if the bag is full the 'Your bag is full' section is added. I think we have this live in our local catalog: https://katalog.ub.uni-leipzig.de/Search/Results?lookfor=Demian -- if you add items with the clipboard icon, the screen reader should announce the changes.
          

> I wonder if we really need two pull requests to address all of this: one to take care of the focus-related issues, and another to simply improve the readability of the facets. Maybe the two things are so intertwined that we should do it all here, but it also seems possible that there are some "low hanging fruit" we could get merged right away while we're discussing these more complex aspects of the problem (like the fact that reading out "filled triangle pointing up" is not going to help anyone, or that introducing alternate sr-only text to contextualize sections of the control and to explain the meaning of the numbers, will improve the experience). :-)

I fully agree. Getting rid of the read-out icon should be fairly straightforward, for example.

> I do think that if we can figure out a technical way to make the focus behave more predictably, that is probably preferable to putting explanatory notes about page reloads all over the place. On pages with hundreds or thousands of facet values, just repeating that text so many times is going to impact the rendered page size and potentially slow things down... and while I realize that screen reader users are accustomed to rapidly skipping through values without listening to the whole thing every time, I still think it might be annoying to have the same message stated so many times. I'm still open to being persuaded otherwise, but this feels like the solution we should choose if we can't find a more elegant way, and I haven't given up hope on finding a more elegant way yet. :-)

Another good point! It might of course be easier to issue the hint just once in the parent element and have the focus stay on/return to the element last used.

demiankatz:
> Sounds like a workable idea to me, and I guess many things feel weird until one gets used to them. It would at least be worth a try.

Is this something you would be interested in trying, or should @crhallberg and I do some experiments on our end? (I realize that either way, we probably won't get to this right away because of vacations).

> I personally have a great dislike for AJAX web sites since you always get that weird delay untill more results plop in and you cannot search exhaustibly inside the page until the last item has loaded ...

I'm glad it's not just me!

> > If there's markup we could use to improve the way this is communicated, I'm open to it -- but I'm assuming that "live section" markup wouldn't be meaningful unless we did an AJAX reimplementation of the search results page. Please correct me if I'm wrong!
> 
> ```
> It works with standard pages. For our bookbag we use:
> 
> <ul class="nav navbar-nav navbar-right flip">
>   <li id="cartSummary" aria-live="polite" aria-atomic="true">
>     <a id="cartItems" class="btn" data-lightbox="" title="View Book Bag" href="/vufind2/local/Cart/Home">
>       <i class="fa fa-clipboard" aria-hidden="true"></i>
>       <span role="status">
>         <span class="sr-only">Book Bag:</span> <strong>11</strong> <span class="cart-label">items</span>
>         <span class="sr-only full hidden">(Your Book Bag is full)</span>
>       </span>
>     </a>
>   </li>
> ```
> 
> So, when I add an item through the result list, the new number of items in the back is read out and if the bag is full the 'Your bag is full' section is added. I think we have this live in our local catalog: https://katalog.ub.uni-leipzig.de/Search/Results?lookfor=Demian -- if you add items with the clipboard icon, the screen reader should announce the changes.

The difference, though, is that the book bag does use AJAX to dynamically change the existing web page without a full reload (the scenario that ARIA "live updates" are designed to capture) whereas applying a filter always causes a full-page reload (so I don't think there's any scenario where just the filter box would be changing without the rest of the page coming along with it).

> I fully agree. Getting rid of the read-out icon should be fairly straightforward, for example.

As with the above, is this something you're interested in working on when time permits? It might make sense to do this in a separate PR since it would be a small and simple change that could get merged while we continue to work through these more complex points.

> Another good point! It might of course be easier to issue the hint just once in the parent element and have the focus stay on/return to the element last used.

Agreed!

crhallberg:
Is this PR still relevant after the changes to #2492?

ckaz:
> Is this PR still relevant after the changes to #2492?

Hi Chris, yes it is. Sorry for the delays. To sum this up quickly: 

- During the accessibility tests, our catalogs where criticized for unexpected losses of focus when applying filters, i.e. the focus must return to the triggering element after the page reload.
- Since we couldn't come up with a satisfactory solution (using the session storage might be an option, says @lahmann), we decided to warn non-sighted users using sr-only texts on the facets. This works well, so I suggested the same here.
- (The bookbag screen-reader announcement is a separate issue and was mainly used for reference. I would suggest a separate PR for this.)
- The main concern, if I'm correct, is that to much text might clutter things.
- @demiankatz also suggested that using anchors could improve this issue -- the rationale is that when clicking on anchors, page reloads can be expected, hence we could avoid explanatory text. I have tried it and it still feels like a workaround, so I wouldn't favor it.
- My current take is: 
  - We should at least inform non-sighted users that the facets will trigger page reloads.
  - If, in the future we could come up with a solution that places focus back on the re-load-triggering facet, this would much improve usabiliity. (This is out of my range of capabilities.)

@demiankatz Yes, I could revise/re-submit this for VuFind 9.

@crhallberg Do you have any ideas on that?

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
```diff
@@ -33,30 +33,48 @@
...
+          (<?=$this->transEsc('applied_filter')?> - <?=$this->transEsc('page_reload_on_deselect_hint', ['%%filter_name%%' => $this->facet['displayText']])?>)<?php else: ?>(<?=$this->transEsc('page_reload_on_select_hint', ['%%filter_name%%' => $this->facet['displayText']])?>)
+        <?php endif;?>
+      </span>
     </span>
   <?php endif; ?>
 
   <?php if (!$this->facet['isApplied']): ?>
     <span class="badge">
-      <?=$this->localizedNumber($this->facet['count']) ?>
+      <?=$this->localizedNumber($this->facet['count']) ?><span class="sr-only"><?=$this->transEsc('results') ?></span>
```
demiankatz: This seems like a problem for two reasons: 1.) I don't think there will be a space between the number and the word "results" and 2.) it would be better to use a string containing a token so that the order will be correct in languages where "results" might come before the number (which I believe is possible).

languages/de.ini
```diff
@@ -906,6 +906,10 @@ page_last = "Zur letzten Seite springen"
 page_next = "Zur nächsten Seite"
 page_num = "Seite %%page%%"
 page_prev = "Zur vorherigen Seite"
+page_reload_on_select_hint = "Page will refresh when filter '%%filter_name%%' is selected"
```
demiankatz: I notice that the German translations are in English for some reason. I don't think it's worth changing these until we come up with our final design, but just wanted to highlight the issue so we don't forget!

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
```diff
@@ -33,30 +33,48 @@
 <?php if ($hasSubLinks): ?>
   <div class="<?=implode(' ', $classList) ?>">
 <?php else: ?>
-  <a href="<?=$toggleUrl ?>" class="<?=implode(' ', $classList) ?>" data-title="<?=$this->escapeHtmlAttr($this->facet['displayText']) ?>" data-count="<?=$this->facet['count'] ?>"<?php if($this->facet['isApplied']): ?> title="<?=$this->transEscAttr('applied_filter') ?>"<?php endif;?> data-lightbox-ignore>
+  <a href="<?=$toggleUrl ?>"
+     class="<?=implode(' ', $classList) ?>"
+     data-title="<?=$this->escapeHtmlAttr($this->facet['displayText']) ?>"
+     data-count="<?=$this->facet['count'] ?>"
+     title="<?php if($this->facet['isApplied']): ?><?=$this->transEsc('applied_filter') ?> - <?=$this->transEsc('page_reload_on_deselect_hint', ['%%filter_name%%' => $this->facet['displayText']])?><?php else: ?><?=$this->transEsc('page_reload_on_select_hint', ['%%filter_name%%' => $this->facet['displayText']])?>
```
demiankatz: This attribute might be more readable if we just assign the title in a PHP block above the link -- as it is here, the wrapping is a bit confusing.

demiankatz (commented):
Thanks, @ckaz -- I've taken another look at this to refresh my memory and left a few new comments (though I don't think we should take action until we've decided on our course).

My opinion remains about the same: I would much rather try to figure out a way to address the underlying issue so the behavior is intuitive, rather than repeatedly tell the user that the link they are about to click will do something unintuitive. I think this will make navigating through the facet lists slower and more frustrating (though maybe I misunderstand exactly how it will act in a screen reader). Better to fix the road than put a warning sign in every pothole. ;-)

One thought regarding anchoring: I don't think it would be difficult to anchor all the "add facet" links to the resulting "remove facet" area, or to anchor the "remove facet" links back to the associated facet group. Sorry if I've suggested that already (memory is limited these days), but if that would solve the problem, I can propose technical details.

Beyond that, I'm definitely interested in @crhallberg's thoughts.

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
crhallberg: 1. The span element should cause a screen reader to read this as two words however.
2. A tokenized string would be better for translation. Perhaps as a title or aria-label on the `<span class="badge">` parent.

languages/de.ini
crhallberg: Since all of the facet selectors are links, I think it is expected that navigation will occur after activation (reloading in this case).

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
```diff
@@ -33,30 +33,48 @@
...
+    <a class="text" href="<?=$toggleUrl ?>" data-lightbox-ignore data-title="<?=$this->escapeHtmlAttr($this->facet['displayText']) ?>" data-count="<?=$this->facet['count'] ?>"
+        title="<?php if($this->facet['isApplied']): ?><?=$this->transEsc('applied_filter') ?> - <?=$this->transEsc('page_reload_on_deselect_hint', ['%%filter_name%%' => $this->facet['displayText']])?> <?=$this->escapeHtmlAttr($this->facet['displayText']) ?><?php else: ?><?=$this->transEsc('page_reload_on_select_hint', ['%%filter_name%%' => $this->facet['displayText']])?><?php endif;?>">
       <?=$displayText ?>
+      <span class="sr-only">
+        <?php if($this->facet['isApplied']): ?>(<?=$this->transEsc('applied_filter')?> - <?=$this->transEsc('page_reload_on_deselect_hint', ['%%filter_name%%' => $this->facet['displayText']])?> <?=$this->escapeHtmlAttr($this->facet['displayText']) ?>)<?php else: ?>(<?=$this->transEsc('page_reload_on_select_hint', ['%%filter_name%%' => $this->facet['displayText']])?>)<?php endif;?>
+      </span>
```
crhallberg: This is quite verbose to put on every single facet. It might be good to put this text by the sidebar header or even near the top of each section so it's conveyed once. As it is now, this will make many of the links on the page unnecessarily wordy.

crhallberg (changes requested):

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
ckaz: You are absolutely right. Putting it at the top of the facets so it can be read once when users arrive there might be the best option.

demiankatz: Thanks, @ckaz, do you have time to try for a revision, and to address some of the other feedback here? I also note there are some merge conflicts that need to be resolved -- let me know if you need any help with those.

It appears that we haven't set a milestone for this. Do you feel this is something we should add with a target of release 9? If so, we should try to expedite work on it so that we can finalize the new language strings, so that this doesn't block our ability to start work on translation for the new release.

ckaz: Yes, I will give this a shot on monday. How do I bring my branch to VuFind 9? I could check it out afresh based on the VF9-branch?

demiankatz: You should be able to merge in the dev branch to get things up to date, but there will definitely be conflicts to resolve. If it's easier to just reapply changes to dev and then force-push to this branch, that's another viable approach.

ckaz:
I finally managed to get to this: The commit should cater to all the comments above so:

- there is now a _short_ **sr-only explanatory text on top of the facets** saying that page will reload on facet select/exclude
- the **bages** with the number of hits now have an **aria-label saying "Results:"**
- the **Filters section** now has a _short_ sr-only **explanatory text** saying that page will reload on facet de-select
- the Filters section has the **remove-all button** text clarified to **"Remove all filters"**
- the Filters section with the **selected filters** now has an sr-only **explanatory text** saying that this is where you can remove individual filters
- the individual filters now have a much increased touch/click size (previously only the x-icon, now the entire text + icon) -- to explain: The x-icon felt much too flimsy to operate, considering people with shaking hands, poor vision etc. [AAA accessibilty even assumes a 42x42px minimum size for touch/click elements!]

**To Dos:**
- The **focus/hover contrast** on the Filters section elements still seems too poor to be easily recognisable.
- There is not focus/hover contast on the selected filters in the sidebar, but I guess, this would be advisable, along with improving hover/focus contrast on all sidebar elements (this is stuff for a separate ticket, I think)

Oh, I just noticed (again) there was an "improved interface for active filter removal" (#2522), so there will be overlap. The only filter section-related changes worth considering after the merge of #2522 from this ticket here would therefore be the explanatory texts and the clarified remove-all button text. I can move them to a separate ticket once #2522 has been merged.

Sorry for my oversight,
Cheers!

demiankatz:
Thanks, @ckaz. I merged #2522 today, so I think the next step is to resolve the resulting conflicts -- once that's done, I'll be happy to take a closer look at this, but I figure I shouldn't start reviewing until it's in a more stable state. :-)

ckaz:
I'll revise based on the current main branch after the merge of #2522 -- this should clarify things.

demiankatz:
> I'll revise based on the current main branch after the merge of #2522 -- this should clarify things.

Thanks, @ckaz -- #2522 is merged now, so you can go ahead at your convenience. (And sorry if I'm repeating myself and you already know this -- just making sure :-) ).

ckaz:
Ok, its done: merged dev into this and removed conflicting code. Hope it makes more sense now.

data/cache/.gitignore
```diff
@@ -1,2 +0,0 @@
-*
```
demiankatz: This file appears to have been deleted by accident.

languages/en.ini
```diff
@@ -1119,7 +1122,7 @@ Requests = "Requests"
 Reserves = "Reserves"
 Reserves Search = "Reserves Search"
 Reserves Search Results = "Reserves Search Results"
-reset_filters_button = "Reset Filters"
+reset_filters_button = "Reset All Filters"
```
demiankatz: The "reset filters" button can be a bit strange, because it resets filters to their default state, based on configuration. In most cases, this means removing all filters, but in situations where visible filters are configured to appear by default, the reset button will restore the settings to the initial state... so it can potentially add filters as well as remove them. I'm not sure if changing the wording of the button helps, and it might make things worse in some cases. Would it be better to leave this to local customization?

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
```diff
@@ -31,7 +31,11 @@
 <?php if ($hasSubLinks): ?>
   <div class="<?=implode(' ', $classList) ?>">
 <?php else: ?>
-  <a href="<?=$toggleUrl ?>" class="<?=implode(' ', $classList) ?>" data-title="<?=$this->escapeHtmlAttr($this->facet['displayText']) ?>" data-count="<?=$this->facet['count'] ?>"<?php if($this->facet['isApplied']): ?> title="<?=$this->transEscAttr('applied_filter') ?>"<?php endif;?> data-lightbox-ignore>
+  <a href="<?=$toggleUrl ?>"
+     class="<?=implode(' ', $classList) ?>"
+     data-title="<?=$this->escapeHtmlAttr($this->facet['displayText']) ?>"
+     data-count="<?=$this->facet['count'] ?>"
+     <?php if($this->facet['isApplied']): ?>title="<?=$this->transEscAttr('applied_filter') ?>:"<?php endif;?> data-lightbox-ignore>
```
demiankatz: I definitely appreciate the improved readability here, but I wonder if the style might need a little adjusting for consistency. This is using a four-space indent instead of a two-space indent (which might be justified, since it makes the attribute names align... but it's a little different from everything else). Maybe it would help to put the closing `>` character de-indented on a line by its own to make it more clear at a glance that this is all one big tag. @crhallberg might have thoughts on this too.

Another alternative might be to refactor the code to use the `htmlAttributes` view helper.

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
```diff
@@ -45,7 +49,7 @@
   <?php endif; ?>
 
   <?php if (!$this->facet['isApplied'] && null !== $this->facet['count']): ?>
-    <span class="badge">
+    <span class="badge" aria-label="<?=$this->transEsc('results') ?>:">
```
demiankatz: Might it be more clear to add a new label for "Result count:"?

languages/en.ini
```diff
@@ -953,6 +953,8 @@ page_last = "Go to Last Page"
 page_next = "Go to Next Page"
 page_num = "Page %%page%%"
 page_prev = "Go to Previous Page"
+page_reload_on_deselect_hint = "Page will reload when a filter removed."
```
demiankatz: missing "is" -- should be "when a filter is removed."

languages/en.ini
```diff
@@ -1092,6 +1094,7 @@ relais_success_message = "Request id #%%id%% was created. You will receive a con
 Related Author = "Related Author"
 Related Items = "Related Items"
 Related Subjects = "Related Subjects"
+remove_selected_filters = "Remove selected filters"
```
demiankatz: Is this wording potentially confusing because it implies that you can select multiple filters and then remove them in a single operation? Would it be better for the heading to be something more like "Currently applied filters" and then have each individual link say "remove [filter]"?

demiankatz (changes requested):
Thanks, @ckaz, I think this is a great improvement. Just a few comments (some very nitpicky). :-)

languages/en.ini
ckaz: Oh, I wasn't aware of that. Sure, leave it as it was then!

languages/en.ini
ckaz: Good find!

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
ckaz: Oh, I just did it for readability and forgot to mush it all back together.

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
ckaz: I don't quite understand what you mean -- this is exactly what I had?

languages/en.ini
ckaz: We already have some sr-only text inside the removable individual filters (just the translation was missing), so I added that and added the plural for the Applied filters string.

ckaz:
Just included your suggestions and commited again. Cheers!

data/cache/.gitignore
demiankatz: I see that you resolved this comment, but the file is still marked as deleted. Could you try again to restore it?

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
demiankatz: Maybe I'm misunderstanding something, but as currently written, it's just going to label the block with the word "results". Won't that read out like "results 150"? That's not going to be grammatically correct in most languages, so I thought it might be less confusing if we made the label more like a label, so it reads out as "Result count: 150". But maybe I'm mistaken about how the aria-label attribute is applied in this context.

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
demiankatz: Thanks. I think it would be a good idea for us to rewrite a number of templates using `htmlAttributes` and other helpers for increased long-term readability, but I guess that's better handled as a separate PR so it doesn't mix multiple unrelated tasks into a single commit.

Now that this is a single line it's easier to compare what you changed, which leads me to two comments:

1.) You've moved the space from in front of the "title" attribute to be in front of the "<?php if" statement associated with inserting a title. I think it was better where it was; otherwise, there will be an extra unnecessary space in the tag when title is not displayed.

2.) You've added a colon onto the end of the title. I think it would be better to include the punctuation in the language files rather than hard-coding it into the template, so it's easier to apply i18n. Maybe it's worth creating a new applied_filter_label or applied_filter_title translation string that just copies the existing string but adds the colon. It should be possible to automate this with the command line language tools. If you agree with the approach but don't know how to use the tools, I'm happy to help with this step.

languages/en.ini
demiankatz: Similar to the above, maybe it's worth renaming the new string to applied_filters_label and including the colon in the string itself.

(And I do realize that VuFind has some existing situations where the colons are applied in the templates instead of in the language files, so there is a precedent for doing it the way it's done here... I just think internationalizing the punctuation is likely to be more flexible in the long run, and if we decide to change the other usages in the future, it will be easier if we don't add new instances).

demiankatz:
Thanks for the progress, @ckaz -- I've just replied to a few of the open threads, and resolved the ones that are done. :-)

languages/en.ini
ckaz: this is gone completely from the code as it is no longer necessary

themes/bootstrap3/templates/search/filters.phtml
```diff
@@ -128,18 +128,21 @@
...
   <?php // Normal view ?>
   <div class="active-filters hidden-xs">
+    <span class="sr-only"><?=$this->transEscAttr('page_reload_on_deselect_hint') ?></span>
     <?php if ($resetLink && $options->getRetainFilterSetting()): ?>
       <a class="reset-filters-btn" href="<?=$resetLink?>"><?=$this->transEsc('reset_filters_button')?></a>
     <?php elseif ($advancedSearch): ?>
       <p class="adv_search_filters"><?=$this->transEsc('adv_search_filters')?>:</p>
     <?php endif; ?>
     <div class="filters">
+      <span class="sr-only"><?=$this->transEscAttr('applied_filters') ?>:</span>
```
demiankatz: The comment you pointed out as outdated was actually talking about this template line and the associated translation string. Sorry for confusing matters by talking about it in entirely the wrong context! :-)

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
ckaz: good point, I didn't consider other languages ... I'd stick with "results" because we already have       
`result_count = "%%count%% Treffer"` but changed the order

themes/bootstrap3/templates/search/filters.phtml
ckaz: Ah, ok. I get it now. Sorry. Committed the change

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
```diff
@@ -47,6 +47,7 @@
   <?php if (!$this->facet['isApplied'] && null !== $this->facet['count']): ?>
     <span class="badge">
       <?=$this->localizedNumber($this->facet['count']) ?>
+      <span class="sr-only"><?=$this->transEsc('results') ?></span>
```
demiankatz: In order to more fully take advantage of i18n, what if you did it like this?

```
<?php $currentCount = $this->localizedNumber($this->facet['count']); ?>
<span class="sr-only"><?=$this->transEsc('result_count', ['%%count%%' => $currentCount]) ?></span>
<span aria-hidden><?=$currentCount ?></span>
```

This way, the screen reader user gets a fully-articulated count appropriate to their local language and the stray number is hidden, and the visual user sees no change.

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
ckaz: It should be resolved to be back the way it was + the colon in the language files. I'd rather not mess with the command line as I'm too inexperienced with this. But please, feel free to improve on this!

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
ckaz: very clever. feels like i'm wasting a lot of your time! Resolved it.

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
ckaz: Done in the comment below

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
demiankatz: The reason I had suggested creating a new string for the version with the colon is that it's easier to remember to get translation strings for all the languages when we introduce new strings than it is to keep track of changes when we edit existing strings. Of course, if `applied_filter` is only used in one place in all the code, then we might as well just edit it in place instead of introducing something new. Have you done that analysis yet, or would you like me to dig into it when time permits?

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
demiankatz: Not a waste, it's a good collaboration! :-)

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
ckaz: It is only used in single-facets.phtml. But of course people might have used it elsewhere in their installations ...

demiankatz: I've added the colon to all the language files. We can easily undo this if we change our minds -- is it appropriate for a title attribute to end in a colon?

data/cache/.gitignore
demiankatz: I went ahead and fixed this while I was working in the PR anyway.

demiankatz:
Thanks again, @ckaz. I ran the full test suite here and found that a few tests broke due to the changes to the text, so I have fixed them. Everything passes now. I think this looks pretty good, but I'd like to hear @crhallberg's feedback one more time before merging.

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
ckaz: Yes, as far as I know, it is perfectly fine.

ckaz:
Great: Pls. let me know if there's anything else I can do here.

languages/ar.ini
```diff
@@ -71,7 +71,7 @@ and = "و"
 anonymous_tags = "وسوم مجهولة"
 APA Citation = "APA استشهاد"
 APA Edition Citation = "توثيق جمعية علم النفس الأمريكية APA (الطبعة السابعة)"
-applied_filter = "مرشح مطبّق"
+applied_filter = "مرشح مطبّق:"
```
crhallberg, line 74: Are we sure this colon is in the right place? And also, do we want punctuation in a translation that doesn't include tokens?

crhallberg (approved):

languages/ar.ini
demiankatz: I'm not 100% certain this is correct, but it does seem to be consistent with other existing translation strings in the file that contain a colon.

As far as including a colon in a string that does not include tokens, I think it's appropriate for a value that is used as a label, because there are other structural aspects of the HTML that logically separate the parts (e.g. label and value are in separate markup areas, different table cells, etc.).

demiankatz: Update: I've sent an email to our Arabic translators to get verification of this. If it's okay, I think we're ready to merge.

demiankatz: Further update: I have received confirmation that this is correct. It's quite confusing, because in GitHub, the .ini file displays with the colon on the right side, which would be incorrect... but when the language files get interpreted by VuFind and displayed in right-to-left mode, the colon is correctly on the left where it belongs. See:

image

demiankatz (approved):
I think we're all set now; thanks again, @ckaz and @crhallberg!

PR #2494, opened 2022-07-21
Implement TranslatorAwareInterface in AbstractBase controller

LuomaJuha:
Use translatorawareinterface so translate function doesn't need to be implemented again and get access to all needed functions.

Also removed unused Use clause. Was there any reason that it was there?


demiankatz:
Thanks, @LuomaJuha. I'm going to run the full test suite to be absolutely sure this has no side effects, but it looks to me like a very clear optimization. Thanks for catching it!

Regarding the unused use statement, the interface is mentioned inside a comment, and php-cs-fixer won't remove use statements for classes that are referenced in comments. However, in this instance, I do believe it is clear enough that we don't need an unnecessary use statement, so I support your decision to remove it.

demiankatz (approved):
All tests are passing; merging now!

PR #2495, opened 2022-07-26
Fix Swagger UI API documentation

xmorave2:
In swagger-ui 4.9.0 was dist/index.html file refactored and the url is no more defined here, but in file swagger-initializer.js

I also removed 'defaultModelRendering' setting change as it is not part of dist/index.html from version 3.0.0


demiankatz (approved):

demiankatz:
Thanks for catching and fixing this!

PR #2496, opened 2022-07-27
FOLIO use item effective call number

ihardy:
Prefer to use effective call number. FOLIO generates this one. I think it should always exist, but left a fallback to item level JIC.


ihardy:
sorry, failed the syntax, let me fix

demiankatz:
Thanks, @ihardy, I think the problem here is just related to line length. I'm happy to fix it for you if that would save any time!

demiankatz (approved):
This works for me and received positive feedback in the FOLIO `#vufind-integrations` channel, so I'm merging it now. I took care of fixing the minor style issue.

PR #2497, opened 2022-07-29
Begin using RDA fields in format calculator.

demiankatz:
I've begun running into records where the 33x fields are the only accurate source of format information. This PR creates some mechanisms to allow 33x fields to override other methods of format identification when they are present. For the moment, it is only being used for videos and online resources, but it should be expandable from here as needed.

This PR also fixes a small order-of-operations bug where iterators were being created BEFORE checking for non-null values.


demiankatz:
I'd love to get a review of this from @eocarragain if he has time to take a look!

mtrojan-ub:
@demiankatz: You added me as a Reviewer on this PR, but didn't you want to add @eocarragain as reviewer instead?

demiankatz:
@mtrojan-ub, I can't add @eocarragain as a reviewer because he is not currently a project committer (though I'd be willing to nominate him if he's interested), so I added you since you often make good Java-related suggestions. :-)

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -548,6 +553,55 @@ protected boolean hasSerialHost(Record record) {
         return false;
     }
 
+    protected boolean isOnlineAccordingTo338(Record record) {
+        // Does the RDA carrier indicate that this is online?
+        List carrierFields = record.getVariableFields("338");
+        if (carrierFields != null) {
+            Iterator carrierFieldsIter = carrierFields.iterator();
+            while (carrierFieldsIter.hasNext()) {
```
mtrojan-ub: You should be able to save multiple lines of code by using a `for` loop instead:
```
for (VariableField variableField : record.getVariableFields("338")) {
    DataField carrierField = (DataField) variableField;
```

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -548,6 +553,55 @@ protected boolean hasSerialHost(Record record) {
...
+     * @param  Record record
+     * @return Set format(s) of record
+     */
+    protected List<String> getFormatsFrom33xFields(Record record) {
+        boolean isOnline = isOnlineAccordingTo338(record);
+        List<String> formats = new ArrayList<String>();
+        List typeFields = record.getVariableFields("336");
+        if (typeFields != null) {
+            Iterator typeFieldsIter = typeFields.iterator();
+            while (typeFieldsIter.hasNext()) {
```
mtrojan-ub: here as well

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -548,6 +553,55 @@ protected boolean hasSerialHost(Record record) {
...
     }
 
+    protected boolean isOnlineAccordingTo338(Record record) {
+        // Does the RDA carrier indicate that this is online?
+        List carrierFields = record.getVariableFields("338");
+        if (carrierFields != null) {
+            Iterator carrierFieldsIter = carrierFields.iterator();
+            while (carrierFieldsIter.hasNext()) {
+                DataField carrierField = (DataField) carrierFieldsIter.next();
+                String desc = carrierField.getSubfield('a') == null ? "" : carrierField.getSubfield('a').getData();
```
mtrojan-ub: When getting values with a fallback like this, you're always calling getSubfield() twice. In PHP we have the null coalesce operator `??` which unfortunately doesn't seem to be available in Java. To avoid this, the only other efficient option would be to implement some kind of function `getSubfieldOrDefault(char code, String default)` that stores the subfield in a variable and calls getData() if it is != null. 

But maybe i'm exaggerating here, since it would only improve performance on tags with lots of subfields...

mtrojan-ub (commented):
Just a few minor optimizations

demiankatz:
Thanks for the helpful feedback, @mtrojan-ub! I just returned from vacation and probably won't get back to this for a little while, but I'll apply your suggestions as soon as time permits.

demiankatz:
Thanks again for the review, @mtrojan-ub. I've finally found time to apply your suggestions. Do you mind taking a second look? I'm also still interested in hearing from @eocarragain if he has any time to look at this.

eocarragain:
hi @demiankatz was on holiday myself - apologies. 

I haven't run this but logic looks fine to me. I'm fine with skipping RecordType check if we already found useful data in 33X fields. The RecordType check was really a fall back for very sparse records where no 007, so if 33X has been completed that is likely a better source.

demiankatz:
Thanks, @eocarragain, I hope you had a good holiday! Do you think it's worth building this out any further, or would you be content to merge this as-is and expand it later as needed? Right now it's very much meeting one very specific need at Villanova rather than doing anything systematic, but maybe that's good enough for the moment. :-)

import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -580,12 +634,20 @@ protected List<String> getFormatsAsList(Record record) {
             result.add("ConferenceProceeding");
         }
 
+        // check the 33x fields; these may give us clear information in newer records
+        // and if we find something here, it currently indicates non-book content.
```
eocarragain: Minor thing but this comment could probably be clearer re why couldBeBook is being set to false, so that we reconsider if/when implementing getFormatsFrom33xFields() more fully e.g. 

```
        // check the 33x fields; these may give us clear information in newer records;
        // in current partial implementation of getFormatsFrom33xFields(), if we find 
        // something here, it indicates non-book content.
```

demiankatz: Good idea, I've adjusted as you suggested.

eocarragain:
It was great, thanks! 

**TL;DR: I think it is fine to do this & expand out when/if it starts to become an issue for people**

I looked at the 33x when doing the RecordType work. The clearer delineation of content/media/carrier is useful in theory but in practice I couldn't see that it was going to help a whole lot over what we were getting from 007/008 (without thinking about splitting Vufind's "format" into discrete content/carrier facets or maybe some sort of hierarchical facet?).

I don't have a good sense of exactly how the RDA fields are being used or how widely it has been adopted by libraries. If/when its the case that lots of libraries are completing 33X _instead of_ 007 rather than _in addition to_ it, it probably does make sense to fill out that getFormatsFrom33xFields() method to handle all 336/337/338 values & codes.  In which case, you'd probably do the same thing as you've done with the RecordType check and not check the 007 anymore if we found good data in 33x

demiankatz:
Thanks, @eocarragain, I think we're on the same page. Based on my experience with these video records, I think we may get more and more questions of "why does record X have format Y" and the answer is going to be "because the metadata creator used 33x and totally ignored the traditional coding." I imagine we could just organically grow the 33x processing to meet cases as they crop up. As you say, trying to do it systematically ahead of time doesn't seem hugely productive, because even though there's a huge number of codes, they don't seem obviously more useful than what we've done in the past, and some are open to interpretation. I think it will be easier to respond to real-world usage patterns. Of course, if somebody wants to take on a project to be systematic, I'm not going to stop them -- I'm just not volunteering myself right now. ;-)

mtrojan-ub:
> Thanks again for the review, @mtrojan-ub. I've finally found time to apply your suggestions. Do you mind taking a second look? I'm also still interested in hearing from @eocarragain if he has any time to look at this.

The mentioned code looks better now! But there still seem to be multiple other occasions where the Iterator+While Pattern could be replaced by a foreach pattern as well.

demiankatz:
@mtrojan-ub, I decided not to optimize existing code that was unrelated to this PR in order to retain focus, but I'll try to find time to open a separate PR to upgrade legacy code.

demiankatz:
@mtrojan-ub, if you are satisfied with what's been done here, please approve the PR so that I can merge it (right now it's not mergeable without an approval). Then I'll get started on the "clean up inefficient older code PR" as soon as I can!

mtrojan-ub (approved):

demiankatz:
See #2520 for the follow-up PR discussed above.

PR #2498, opened 2022-08-01
Adding configurable ignoreFilterReasons in Alma Resolver

he247:
This change was made based on the discussion in the mailing-list "[VuFind-Tech] [EXTERNAL]  Alma Resolver - Date Filter".

### DIFF
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -1281,8 +1281,11 @@ rfr_id          = vufind.svn.sourceforge.net
 ; "demo" generates fake values for use in testing the embed setting below).
 ;resolver        = sfx
 
+; If the Alma OpenURL resolver returns reasons to filter specific resource links,
+; these reasons may be ignored by adding the "filter reason" to this list.
 ; List of filter reasons that are ignored (displayed regardless of filtering).
-; You may also use a comma-separated list of filter reasons
+; You may also use a comma-separated list of filter reasons.
+; (default is "" if commented out: non filter reason ignored)
 ; ignoredFilterReasons            = "Date Filter"
 
 ; If you want OpenURL links to open in a new window, set this setting to the
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -1286,7 +1286,7 @@ rfr_id          = vufind.svn.sourceforge.net
 ; List of filter reasons that are ignored (displayed regardless of filtering).
 ; You may also use a comma-separated list of filter reasons.
 ; (default is "" if commented out: non filter reason ignored)
-; ignoredFilterReasons            = "Date Filter"
+ignoredFilterReasons            = "Date Filter"
 
 ; If you want OpenURL links to open in a new window, set this setting to the
 ; desired Javascript window.open parameters.  If you do not want a new window
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -1281,12 +1281,15 @@ rfr_id          = vufind.svn.sourceforge.net
 ; "demo" generates fake values for use in testing the embed setting below).
 ;resolver        = sfx
 
-; If the Alma OpenURL resolver returns reasons to filter specific resource links,
-; these reasons may be ignored by adding the "filter reason" to this list.
-; List of filter reasons that are ignored (displayed regardless of filtering).
-; You may also use a comma-separated list of filter reasons.
-; (default is "" if commented out: non filter reason ignored)
-ignoredFilterReasons            = "Date Filter"
+; Some link resolver drivers can filter resource links based on specific criteria.
+; This setting indicates which filter reasons should NOT be used to exclude
+; resource links. This is currently ONLY supported by the Alma resolver driver.
+; (default is "Date Filter" if commented out, for backwards compatibility)
+; You may add new filters to the array by adding new lines in the format shown below
+; Currently known valid values include: Date Filter
+; ignoredFilterReasons[] = "Date Filter"
+; To deactivate this setting you can set
+; ignoredFilterReasons = false or one line ignoredFilterReasons[] = ""
 
 ; If you want OpenURL links to open in a new window, set this setting to the
 ; desired Javascript window.open parameters.  If you do not want a new window
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -1289,7 +1289,7 @@ rfr_id          = vufind.svn.sourceforge.net
 ; Currently known valid values include: Date Filter
 ; ignoredFilterReasons[] = "Date Filter"
 ; To deactivate this setting you can set
-; ignoredFilterReasons = false or one line ignoredFilterReasons[] = ""
+; ignoredFilterReasons = false or one line ignoredFilterReasons = ""
 
 ; If you want OpenURL links to open in a new window, set this setting to the
 ; desired Javascript window.open parameters.  If you do not want a new window
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -1281,6 +1281,10 @@ rfr_id          = vufind.svn.sourceforge.net
 ; "demo" generates fake values for use in testing the embed setting below).
 ;resolver        = sfx
 
+; List of filter reasons that are ignored (displayed regardless of filtering).
+; You may also use a comma-separated list of filter reasons
+; ignoredFilterReasons            = "Date Filter"
+
 ; If you want OpenURL links to open in a new window, set this setting to the
 ; desired Javascript window.open parameters.  If you do not want a new window
 ; to open, set this to false or comment it out.
--- a/module/VuFind/src/VuFind/Resolver/Driver/Alma.php
+++ b/module/VuFind/src/VuFind/Resolver/Driver/Alma.php
@@ -69,7 +69,7 @@ class Alma extends AbstractBase
         if (isset($options['ignoredFilterReasons'])) {
             $this->ignoredFilterReasons
                 = empty($options['ignoredFilterReasons'])
-                    ? [] : (array)$options['ignoredFilterReasons'];
+                    ? [] : array_filter((array)$options['ignoredFilterReasons']);
         }
     }
 
--- a/module/VuFind/src/VuFind/Resolver/Driver/Alma.php
+++ b/module/VuFind/src/VuFind/Resolver/Driver/Alma.php
@@ -55,19 +55,21 @@ class Alma extends AbstractBase
     /**
      * Constructor
      *
-     * @param string               $baseUrl    Base URL for link resolver
-     * @param \Laminas\Http\Client $httpClient HTTP client
-     * @param \Laminas\Config\Config $config VuFind Configuration (optional)
+     * @param string                 $baseUrl    Base URL for link resolver
+     * @param \Laminas\Http\Client   $httpClient HTTP client
+     * @param \Laminas\Config\Config $config     VuFind Configuration (optional)
      */
     public function __construct(
         $baseUrl,
         \Laminas\Http\Client $httpClient,
-        \Laminas\Config\Config $config = null)
-    {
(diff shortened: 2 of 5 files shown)


he247:
@demiankatz could you help me with the UnitTest? I'm little lost as I set the parameter type to "\Laminas\Config\Config".
Should I reduce the parameter to the config value (string) of "ignoredFilterReasons"?

demiankatz:
Thanks, @he247! I'll give this a closer look tomorrow, but the test failure is because the test is constructing the Alma object with only two parameters, and your changes here add a third mandatory parameter. You'll either need to change the test to build the object differently, or else make the constructor parameter optional (e.g. by defaulting to null).

A bigger question, though: is this something that belongs in config.ini, or should it really be part of Alma.ini? That might have some impact on how things are designed....

[and please forgive my earlier revision of this comment, which contained some inaccurate information, since I once again forgot that we were talking about resolver drivers and not ILS drivers].

he247:
Hi @demiankatz thank you for the hint about defaulting to null. One question would be if I use the entire config-object as parameter or just the string stored in "ignoredFilterReasons".

I'd prefer keeping it in the general config.ini for OpenUrl, so it wouldn't be spread into several config-files. And I can imagine that it might also be useful for other Link Resolvers to not have static filter settings (even though it might not be implemented yet).

[no worries, I always appreciate your input :) ]

EreMaijala:
@he247 I agree the right place is config.ini since the resolver driver is not related to the ILS driver.

demiankatz:
@he247 and @EreMaijala, my concern about including the setting in config.ini is that if it only works for Alma and/or contains Alma-specific values, it will be confusing for users of other services. If we keep it there, I think we should either implement it universally or put "alma" in the setting name to indicate the relationship. A clear comment might also help, indicating context and legal values.

EreMaijala:
@demiankatz A setting with "alma" in it, or an Alma resolver specific section would work. My comment was just meant to sat that this is unrelated to any of the settings in Alma.ini and it would be super-confusing to intertwine ILS driver and resolver driver settings in a single configuration file.

he247:
@demiankatz and @EreMaijala , if you agree I'd just add a part to the comment in the config.ini stating that this is Alma-Resolver specific

demiankatz:
That sounds reasonable, @he247. If in future we decide to extend the setting to work with other drivers, then we can just update the comment.

config/vufind/config.ini
```diff
@@ -1281,6 +1281,13 @@ rfr_id          = vufind.svn.sourceforge.net
 ; "demo" generates fake values for use in testing the embed setting below).
 ;resolver        = sfx
 
+; If the Alma OpenURL resolver returns reasons to filter specific resource links,
+; these reasons may be ignored by adding the "filter reason" to this list.
+; List of filter reasons that are ignored (displayed regardless of filtering).
+; You may also use a comma-separated list of filter reasons.
+; (default is "" if commented out: non filter reason ignored)
+ignoredFilterReasons            = "Date Filter"
```
demiankatz: Would it make sense to use an array-type setting here, rather than a delimited list? e.g.

```
ignoredFilterReasons[] = "reason 1"
ignoredFilterReasons[] = "reason 2"
```

rather than

```
ignoredFilterReasons = "reason 1,reason2"
```

?

This would require less parsing to deal with in the driver, and might avoid potentially problematic edge cases. You could support having the setting set to false (`ignoredFilterReasons = false`) to disable filtering.

Also, is there a strong reason to change the default to "no filters" rather than allowing the driver to have default settings? Making this change could introduce unexpected changes in behavior if people fail to update their configurations when upgrading. I'm not sure how likely that is to be a problem but wanted to raise the question.

module/VuFind/src/VuFind/Resolver/Driver/Alma.php
```diff
@@ -50,18 +50,26 @@ class Alma extends AbstractBase
...
-    public function __construct($baseUrl, \Laminas\Http\Client $httpClient)
-    {
+    public function __construct(
+        $baseUrl,
+        \Laminas\Http\Client $httpClient,
+        \Laminas\Config\Config $config = null
+    ) {
         parent::__construct($baseUrl);
         $this->httpClient = $httpClient;
+        $this->ignoredFilterReasons = explode(
```
demiankatz: What if you just accepted `array $options` instead of a Config object? You could convert the OpenURL Config section to an array in the factory, and then this class has one less dependency. (@EreMaijala and I have talked about reducing dependency on the Config class, since it's not as useful as it used to be, and if it should ever go out of support in the future, it will be that much less work to factor it out later).

My thought is that you could default the property to whatever value you like (either the old default, if we decide to keep it, or an empty array if we don't). Then the code here could be something like:

```
if (isset($options['ignoredFilterReasons'])) {
    $this->ignoredFilterReasons = empty($options['ignoredFilterReasons']) ? [] : (array)$options['ignoredFilterReasons'];
}
```

(Of course, that assumes you take my suggestions above about using array-type configuration settings and supporting false to disable filtering... if you prefer to stick with the delimited string, the code would look a little different).

module/VuFind/src/VuFind/Resolver/Driver/AlmaFactory.php
```diff
@@ -0,0 +1,69 @@
...
+     * creating a service.
+     * @throws ContainerException&\Throwable if any other error occurs
+     */
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        $options = [$container->get(\VuFind\Config\PluginManager::class)
+            ->get('config')];
```
demiankatz: I think you could just add `->OpenURL->toArray()` to the end of this if you wanted to pass an array instead of a Config object as noted above.

config/vufind/config.ini
demiankatz: Also, a suggested rewording of the comment that might be more "future-proof":

```
; Some link resolver drivers can filter resource links based on specific criteria.
; This setting indicates which filter reasons should NOT be used to exclude
; resource links.
; [note about format, depending on discussion above]
; [note about defaults, depending on discussion above]
; This is currently ONLY supported by the Alma resolver driver.
; Legal values include: Date Filter (any others of note?)
```

demiankatz (commented):
Thanks again, @he247, I think this is coming together now. I've done a more thorough review with some possible suggestions. As always, you can feel free to ignore any that are unhelpful... but hopefully at least some are relevant. (I was pretty far off-base yesterday ;-) ).

module/VuFind/src/VuFind/Resolver/Driver/AlmaFactory.php
he247: Dear @demiankatz, I really appreciate your suggestions. These changes will also prevent the UnitTest to fail. So I will apply your suggestions and then push again :)

config/vufind/config.ini
he247: There is another Filter Reason I found "Available for - the ip groups are not part of the service AF groups" but unfortunately I wasn't able to find any documentation about those reasons.

config/vufind/config.ini
```diff
@@ -1281,6 +1281,16 @@ rfr_id          = vufind.svn.sourceforge.net
...
 
+; Some link resolver drivers can filter resource links based on specific criteria.
+; This setting indicates which filter reasons should NOT be used to exclude
+; resource links. This is currently ONLY supported by the Alma resolver driver.
+; (default is "Date Filter" if commented out, for backwards compatibility)
+; You may add new filters to the array by adding new lines in the format shown below
+; Currently known valid values include: Date Filter
+; ignoredFilterReasons[] = "Date Filter"
+; To deactivate this setting you can set
+; ignoredFilterReasons = false or one line ignoredFilterReasons[] = ""
```
demiankatz: I think `ignoredFilterReasons = ""` would be better than `ignoredFilterReasons[] = ""`. Without the brackets, it will resolve to empty and be ignored. With the brackets, it will be a non-empty array containing an empty string, and the options probably won't be set as expected. (I don't think this would hurt anything, but I also don't think it's what is desired/expected).

module/VuFind/src/VuFind/Resolver/Driver/AlmaFactory.php
```diff
@@ -0,0 +1,69 @@
+<?php
+/**
+ * Factory for Alma resolver driver.
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2020.
```
demiankatz, line 7: Might want to use 2022 since this is a new file.

module/VuFind/src/VuFind/Resolver/Driver/AlmaFactory.php
```diff
@@ -0,0 +1,69 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Resolver_Drivers
+ * @author   Demian Katz <demian.katz@villanova.edu>
```
demiankatz, line 25: Feel free to give yourself credit if you wish. :-)

demiankatz (changes requested):
Thanks, @he247, this is looking good to me. I've just left a few very small comments about comments.

Also, it would be nice if we could add a case to the test (https://github.com/vufind-org/vufind/blob/dev/module/VuFind/tests/unit-tests/src/VuFindTest/Resolver/Driver/AlmaTest.php) to cover this new functionality. Since the existing test fixture includes a date filter, I think it would be pretty simple to copy `testParseLinks` to `testParseLinksWithoutFiltering`, add a new parameter to `createConnector` to accept options, and then set a custom filter option in the new test and adjust the expectations accordingly. I can help if you need me to!

config/vufind/config.ini
demiankatz: I suppose another alternative would be to add an `array_filter` to the code processing the configuration, which will filter out empty strings and then make both versions work as expected...

he247:
Thanks, @demiankatz, I will apply the changes and add a test. :)

module/VuFind/tests/unit-tests/src/VuFindTest/Resolver/Driver/AlmaTest.php
```diff
@@ -170,6 +173,96 @@ public function testParseLinks()
...
+                'coverage' => '',
+                'access' => '',
+                'href' => 'https://www.google.com/search?Testingrft.oclcnum=437189463&q=Fundamental+Data+Compression&rft.archive=9942811800561',
+                'notes' => '',
+                'authentication' => '',
+                'service_type' => 'getWebService',
+            ],
+        ];
+
+        $this->assertEquals($result, $testResult);
```
demiankatz: In PHPUnit assertions, the first parameter is the "expected" value and the second is the "actual" value -- so I think these should be reversed.

demiankatz (changes requested):
Thanks, @he247 -- there's one very small fix needed to the test, and then I believe this can be merged. If you're busy, let me know and I can fix this on my end if you prefer!

module/VuFind/tests/unit-tests/src/VuFindTest/Resolver/Driver/AlmaTest.php
he247: Thank you for reviewing, @demiankatz I corrected the parameter order in both assertions.

demiankatz (approved):
Thanks, @he247 -- I see now that the test bug was copied and pasted from the existing test, so thanks for fixing that as well. I think we're all set now. I'll merge this right away!

PR #2502, opened 2022-08-04
Add birthdate support to several ILS drivers.

EreMaijala:
The field is now returned in patron profile with Alma, Demo, KohaRest and SierraRest drivers. It isn't displayed since that's rarely useful for the user, but is provided so that it could be used to determine if the user should be able to access age-restricted material.

TODO:
- [x] Update ILS driver wiki page.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Demo.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Demo.php
@@ -893,7 +893,7 @@ class Demo extends AbstractBase implements \VuFind\I18n\HasSorterInterface
     public function getMyProfile($patron)
     {
         $this->checkIntermittentFailure();
-        $age = 13 + rand(0, 100);
+        $age = rand(13, 113);
         $birthDate = new \DateTime();
         $birthDate->sub(new \DateInterval("P{$age}Y"));
         $patron = [
--- a/module/VuFind/src/VuFind/ILS/Driver/Demo.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Demo.php
@@ -893,6 +893,9 @@ class Demo extends AbstractBase implements \VuFind\I18n\HasSorterInterface
     public function getMyProfile($patron)
     {
         $this->checkIntermittentFailure();
+        $age = 13 + rand(0, 100);
+        $birthDate = new \DateTime();
+        $birthDate->sub(new \DateInterval("P{$age}Y"));
         $patron = [
             'firstname'       => 'Lib-' . $patron['cat_username'],
             'lastname'        => 'Rarian',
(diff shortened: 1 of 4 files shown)


demiankatz:
Thanks, @EreMaijala. Do we need to worry about normalizing the date to a particular format across the various drivers? I'm not sure how stable/reliable the format is for the various platforms, or whether there might be any locale-specific considerations to worry about.

Apart from settling that detail, this all makes sense to me!

EreMaijala:
@demiankatz All of these return the date in Y-m-d format, and that's what I added to the wiki as well. Though now that you mentioned it, I checked again, and Alma adds a time zone indicator, so I'll need to clean that up.

demiankatz:
Thanks for double-checking, @EreMaijala! I had a feeling it was too good to be true for everything to be uniform by default. :-)

EreMaijala:
@demiankatz I'm not sure I understand why Alma returns dates with the UTC designator, and I'm also not sure it's quite right in this case, so I decided to just drop it like the parseDate method does as well.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -893,6 +893,9 @@ public function patronLogin($username, $password)
     public function getMyProfile($patron)
     {
         $this->checkIntermittentFailure();
+        $age = 13 + rand(0, 100);
```
demiankatz: Silly observation, but why do math here? Could just be `rand(13, 113)` (or maybe the more realistic `rand(13, 100)`).

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -621,7 +621,8 @@ public function getMyProfile($patron)
             'zip' => $result['postal_code'],
             'city' => $result['city'],
             'country' => $result['country'],
-            'expiration_date' => $this->convertDate($result['expiry_date'] ?? null)
+            'expiration_date' => $this->convertDate($result['expiry_date'] ?? null),
```
demiankatz, line 624: Isn't convertDate going to return whatever display date format is configured in config.ini, rather than the desired Y-m-d?

demiankatz (changes requested):
Thanks for the progress -- I gave this a more thorough review and had a couple of small comments.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: Oops. Yes. And it's for display purposes anyway. I just stuffed up when updating the wiki page.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: That was to make it clear that 13 is the minimum age a Demo user can have, but admittedly your proposal is not that much harder to read, so I'll just change it.

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: Wiki fixed.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: This changed as well.

demiankatz (approved):
Thanks, looking good now (along with the wiki page)!

PR #2503, opened 2022-08-08
Fix PHP 8.1 deprecation notice.

EreMaijala:
Second parameter to BackendException should be an int. Use original exception's code for the converted exception as well (like in EDS).


demiankatz (approved):
Thanks, @EreMaijala!

PR #2504, opened 2022-08-09
Catch ILS exceptions in HoldsController.

EreMaijala:
Moves the existing OnDispatch handler that catches ILS exceptions from MyResearchController to a trait and uses it in MyResearchController and HoldsController.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2509, opened 2022-08-12
Handle collapsed facets properly when loading hierarchical facets.

EreMaijala:
Moves the check for collapsed facet into JS to take into account the client-side session storage that overrides collapsedFacets from configuration.

Before the change a collapsed facet could have been loaded in the background. While this doesn't cause any actual problems, it's just extra work for the server in case the user never sees the results.


EreMaijala:
Mink tests are passing.

demiankatz (approved):
Thanks, @EreMaijala!

PR #2510, opened 2022-08-12
Redirect to profile page after email verification.

EreMaijala:
Since it was probably initiated from the profile page, it's a logical target as well. It also ensures the alert messages are shown. Otherwise they could be hidden until a page that displays them is shown. This would happen e.g. when defaultLoggedInModule = Search in config.ini.


demiankatz (approved):
Makes sense to me! Thanks, @EreMaijala.

PR #2511, opened 2022-08-15
Avoid double-load of hierarchical facets in deferred facets.

EreMaijala:
Removes the data-facet attribute from the collapse container to avoid triggering the getSideFacets loader.


demiankatz (approved):

demiankatz:
Thanks, @EreMaijala!

One issue this raises, which does not need to be addressed here, is a question of style: should we use `<?php endif ?>` or <?php endif; ?>`? I realize that from a technical perspective it doesn't matter, but from a style perspective I think it might be better to pick one style or the other and normalize everything.

EreMaijala:
@demiankatz I picked the format around the one I added. Here I think it would make sense to use the htmlAttributes helper to get rid of the really long lines and make the code much more clear, but that should be done separately without functional changes.

demiankatz:
Agreed, @EreMaijala, there are really two separate code quality things that would be worth doing when time permits:

1.) Simplify any long lines in templates containing lots of attributes (particularly conditional ones) to improve readability using a helper.

2.) Investigate whether php-cs-fixer has a style rule to control whether or not single-line `<?php ... ?>` tags end with a semi-colon, since I think it would be nice to standardize that one way or the other.

PR #2512, opened 2022-08-15
Fix and unify facet load indicators.

EreMaijala:
- Adds whitespace between spinner and text in JS-rendered loading indicator for uniformity with phtml rendered versions.
- Adds a class to hierarchical facet load indicator to remove extra borders from the list-group-item.
- Restores load indicator for normal (non-hierarchical) deferred facets and tweaks styles to ensure it looks good (avoids flex to keep the spinner steady).
- Improves the look of facet load interruption message.


demiankatz (approved):
Makes sense to me!

PR #2513, opened 2022-08-19
Remove password check in PAIA patronLogin() if client_credentials is used as grant_type

lahmann:
The PAIA grant_type `client_credentials` is only used when patron authentication is done e.g. by Shibboleth which usually provides only the patron/library card id for the authenticated user. Therefore `patronLogin()` only uses the client credentials with the patron identifier so checking for the patron password will always fail, which is being fixed by this pull request.
In addition to that the prior implementation was always overwriting the `post_data` array resulting in non working logins which is fixed with this pull request as well.


demiankatz:
@lahmann, is there any other data element that could be used as a double-check in the absence of a password? I often like to perform some kind of additional verification in scenarios like this, because otherwise it would be very easy to spoof a user's account under certain circumstances (for example, if there's a temporary system outage that causes VuFind to display the "catalog login" form, somebody could enter any user name to access that account).

demiankatz:
Thanks for the update, @lahmann! Did you have any thoughts on my comments from earlier? Also, does @olli-gold have time for the requested review?

lahmann:
To reflect the result from #2570 I added a comment to PAIA.ini pointing to the `allowUserLogin` setting. In combination with that setting spoofing of usernames should be avoidable.

@demiankatz do you think this is good enough to be merged?

demiankatz (approved):
Thanks, @lahmann, this addresses my concerns. However, I think it would also be beneficial to have a review from @olli-gold before merging, just in case any further PAIA-specific concerns can be identified (since I can't do any hands-on testing of this myself).

olli-gold (approved):
Sorry for the delayed review!
I have done a quick test of the new driver now and it's working fine on my side (at least as far as I saw that). So I can finally approve this one; no objections about merging it!

demiankatz:
Thanks, @olli-gold! Merging now!

PR #2515, opened 2022-08-24
Use a callback to create a HTTP client in Solr connector.

EreMaijala:
This makes it possible to create the client when the URL is available so that any proxy can be bypassed for local addresses.

This is provided as a starting point for solving the issue where proxy configuration prevents connection to a local Solr instance. If deemed appropriate, similar mechanism could be used to fix other cases where just passing HttpService in constructor instead of a Client is not sufficient.

TODO

### DIFF
--- a/module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
+++ b/module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
@@ -48,6 +48,7 @@ use VuFindSearch\Backend\Solr\HandlerMap;
 class WriterTest extends \PHPUnit\Framework\TestCase
 {
     use \VuFindTest\Feature\SearchServiceTrait;
+    use \VuFindTest\Feature\ReflectionTrait;
 
     /**
      * Test commit
@@ -61,7 +62,17 @@ class WriterTest extends \PHPUnit\Framework\TestCase
         );
         $connector = $bm->get('Solr')->getConnector();
         $connector->expects($this->once())->method('write')
-            ->with($this->isInstanceOf(CommitDocument::class));
+            ->with($this->isInstanceOf(CommitDocument::class))
+            ->will(
+                $this->returnCallback(
+                    function () use ($connector) {
+                        // Call client factory for expectations to be met:
+                        $factory = $this->getProperty($connector, 'clientFactory');
+                        $factory('');
+                        return true;
+                    }
+                )
+            );
         $writer = new Writer($this->getSearchService($bm), $this->getMockChangeTracker());
         $writer->commit('Solr');
     }
@@ -121,7 +132,17 @@ class WriterTest extends \PHPUnit\Framework\TestCase
         );
         $connector = $bm->get('Solr')->getConnector();
         $connector->expects($this->once())->method('write')
-            ->with($this->isInstanceOf(OptimizeDocument::class));
+            ->with($this->isInstanceOf(OptimizeDocument::class))
+            ->will(
+                $this->returnCallback(
+                    function () use ($connector) {
+                        // Call client factory for expectations to be met:
+                        $factory = $this->getProperty($connector, 'clientFactory');
+                        $factory('');
+                        return true;
+                    }
+                )
+            );
         $writer = new Writer($this->getSearchService($bm), $this->getMockChangeTracker());
         $writer->optimize('Solr');
     }
@@ -179,11 +200,21 @@ class WriterTest extends \PHPUnit\Framework\TestCase
             ->onlyMethods(['getConnector', 'getIdentifier'])
             ->getMock();
         $handlerMap = new HandlerMap();
-        $client = $client ?? $this->getMockBuilder(\Laminas\Http\Client::class)
-            ->onlyMethods([])
-            ->getMock();
         $mockConnector = $this->getMockBuilder(\VuFindSearch\Backend\Solr\Connector::class)
-            ->setConstructorArgs(['http://localhost:8983/solr/biblio', $handlerMap, $client])
+            ->setConstructorArgs(
+                [
+                    'http://localhost:8983/solr/biblio',
+                    $handlerMap,
+                    function () use ($client) {
+                        // If client is provided, return it since it may have test
+                        // expectations:
+                        return $client
+                            ?? $this->getMockBuilder(\Laminas\Http\Client::class)
+                                ->onlyMethods([])
+                                ->getMock();
+                    }
+                ]
+            )
             ->onlyMethods(['write'])
             ->getMock();
         $mockBackend->expects($this->any())->method('getConnector')->will($this->returnValue($mockConnector));
--- a/module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
+++ b/module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
@@ -28,11 +28,11 @@
 namespace VuFindTest\Solr;
 
 use VuFind\Db\Table\ChangeTracker;
-use VuFind\Search\BackendManager;
 use VuFind\Solr\Writer;
+use VuFindSearch\Backend\Solr\Command\WriteDocumentCommand;
 use VuFindSearch\Backend\Solr\Document\CommitDocument;
+use VuFindSearch\Backend\Solr\Document\DeleteDocument;
 use VuFindSearch\Backend\Solr\Document\OptimizeDocument;
-use VuFindSearch\Backend\Solr\HandlerMap;
 
 /**
  * Solr Utils Test Class
@@ -45,9 +45,6 @@ use VuFindSearch\Backend\Solr\HandlerMap;
  */
 class WriterTest extends \PHPUnit\Framework\TestCase
 {
-    use \VuFindTest\Feature\SearchServiceTrait;
-    use \VuFindTest\Feature\ReflectionTrait;
-
     /**
      * Test commit
      *
@@ -55,12 +52,8 @@ class WriterTest extends \PHPUnit\Framework\TestCase
      */
     public function testCommit()
     {
-        $bm = $this->getBackendManagerWithMockSolr();
-        $connector = $bm->get('Solr')->getConnector();
-        $connector->expects($this->once())->method('write')
-            ->with($this->isInstanceOf(CommitDocument::class));
-        $writer = new Writer($this->getSearchService($bm), $this->getMockChangeTracker());
-        $writer->commit('Solr');
+        $command = new WriteDocumentCommand('Solr', new CommitDocument(), 60 * 60);
+        $this->getWriter($command)->commit('Solr');
     }
 
     /**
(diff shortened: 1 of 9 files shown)


module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
```diff
@@ -102,19 +102,19 @@ class Connector implements \Laminas\Log\LoggerAwareInterface
...
         HandlerMap $map,
-        HttpClient $client,
+        callable $cf,
         $uniqueKey = 'id'
     ) {
         $this->url = $url;
         $this->map = $map;
         $this->uniqueKey = $uniqueKey;
-        $this->client = $client;
+        $this->clientFactory = $cf;
```
demiankatz: If we wanted to maintain backward compatibility, we could type sniff $cf, and if it's a Client object, define the factory as `function () use ($cf) { return clone $cf; }`. Not sure that there's a strong reason to do this (probably simpler not to), but just sharing a passing thought that might be useful if we want to prioritize backward-compatibility over simplicity.

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
```diff
@@ -61,7 +62,17 @@ public function testCommit()
         );
         $connector = $bm->get('Solr')->getConnector();
         $connector->expects($this->once())->method('write')
-            ->with($this->isInstanceOf(CommitDocument::class));
+            ->with($this->isInstanceOf(CommitDocument::class))
+            ->will(
+                $this->returnCallback(
```
demiankatz: This seems like kind of a weird hack -- forcing a method to have a side effect it wouldn't normally have in order to replicate behavior that is being mocked out elsewhere (if I'm understanding correctly). Is there a more elegant way to do this?

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php
```diff
@@ -582,6 +583,45 @@ public function testWriteDocument()
...
+     *
+     * @return void
+     */
+    public function testDelayedClientCreation()
+    {
+        $doc = new CommitDocument();
+        $client = $this->getMockBuilder(\Laminas\Http\Client::class)
+            ->onlyMethods(['setOptions'])
+            ->getMock();
+        $client->expects($this->exactly(0))->method('setOptions')
```
demiankatz: Why use `exactly(0)` rather than `never()`? If it's not going to be called, why are we specifying `with` parameters? Or am I misunderstanding the order of precedence here?

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php
```diff
@@ -582,6 +583,45 @@ public function testWriteDocument()
...
+        $connector->expects($this->once())->method('write')
+            ->with(
+                $this->equalTo($doc),
+                $this->equalTo('update'),
+                $this->isNull()
+            )
+            ->will(
+                $this->returnCallback(
+                    function () use ($connector) {
+                        // Call client factory for expectations to be met:
```
demiankatz, line 595: Another instance of this hack.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php
```diff
@@ -644,10 +684,19 @@ protected function loadResponse($fixture)
...
         $map = new HandlerMap(['select' => ['fallback' => true]]);
-        $client = $client ?? new \Laminas\Http\Client();
         return $this->getMockBuilder(\VuFindSearch\Backend\Solr\Connector::class)
             ->onlyMethods($mock)
-            ->setConstructorArgs(['http://example.org/', $map, $client])
+            ->setConstructorArgs(
+                [
+                    'http://localhost/',
+                    $map,
+                    function () use (&$client) {
```
demiankatz: Is there a reason you're passing a reference here, but not in WriterTest? Just curious about the apparent inconsistency.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/ConnectorTest.php
```diff
@@ -161,7 +170,16 @@ public function testWriteJSON()
             ->with($this->equalTo($jsonData));
         $conn = $this->getMockBuilder(Connector::class)
```
demiankatz: Is it worth having a `getMockConnector` function for this repeated code now that it's significantly longer?

demiankatz (changes requested):
Thanks, @EreMaijala, this makes sense to me in principle! I have a few questions/comments, mostly to do with the test suite. It seems to me that some of this might be able to be made more clear. If you want me to spend time coming up with a more tangible proposal for how, let me know. :-)

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
EreMaijala: That makes sense, added. There's no way to avoid changing the constructor signature, so it's a breaking change regardless in that way.

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
EreMaijala: I removed the expectations that were not relevant to the functionality being tested in WriterTest.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php
EreMaijala: That was a mistake. Tests have now been rearranged and cleaned up.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php
EreMaijala: Just a mistake.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/ConnectorTest.php
EreMaijala: Added, though named `getConnectorMock` for parity with the one in BackendTest.

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
demiankatz: I think WriterTest is actually still testing more than it needs to, and we've lost the part of the test that confirms that expected timeout values are passed along. Since the actual behavior of the WriteDocumentCommand is covered by its own test, I don't think we need to create a whole environment to test that behavior here. Why mock a Solr Connector at all? As a unit test, we should just confirm that the Writer passes an appropriately populated WriteDocumentCommand to the search service... which I think we could do by just mocking the search service and putting a custom callback with expectations on the invoke method. What do you think? That would allow us to further simplify both testCommit and testOptimize. (It might also be worth rewriting other test methods, but it seems like commit/optimize is the bare minimum to keep things passing here, so I wouldn't ask for more than that).

EreMaijala: You're right of course. I've now pretty much rewritten all the tests. What do you think?

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
```diff
@@ -145,63 +120,64 @@ public function testDeleteAll()
...
+     * @param object $expectedCommand Expected command class
+     * @param mixed  $result          Result to return for the invoked command
+     *
+     * @return MockObject&SearchService
      */
-    protected function getMockChangeTracker()
+    protected function getMockSearchService($expectedCommand, $result)
     {
-        return $this->getMockBuilder(\VuFind\Db\Table\ChangeTracker::class)
+        $resultCommand = $this->getMockBuilder(WriteDocumentCommand::class)
```
demiankatz: Would it make sense for this to be `get_class($expectedCommand)` rather than a hard-coded value, or am I misunderstanding something?

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
```diff
@@ -45,22 +45,15 @@
...
      */
     public function testCommit()
     {
-        $bm = $this->getBackendManagerWithMockSolr();
-        $connector = $bm->get('Solr')->getConnector();
-        $connector->expects($this->once())->method('write')
-            ->with($this->isInstanceOf(CommitDocument::class));
-        $writer = new Writer($this->getSearchService($bm), $this->getMockChangeTracker());
-        $writer->commit('Solr');
+        $command = new WriteDocumentCommand('Solr', new CommitDocument(), 60 * 60);
```
demiankatz: Maybe it would make sense to rename all these $command variables to $expectedCommand, just to make it clearer what role they play in the test.

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
```diff
@@ -145,63 +120,64 @@ public function testDeleteAll()
...
-                ->getMock();
-        $mockBackend->expects($this->any())->method('getConnector')
-            ->will($this->returnValue($mockConnector));
-        $mockBackend->expects($this->any())->method('getIdentifier')
-            ->will($this->returnValue('Solr'));
-        $sm->setService('Solr', $mockBackend);
-        return $pm;
+        return $this->getMockBuilder(\VuFind\Db\Table\ChangeTracker::class)
+            ->disableOriginalConstructor()
+            ->onlyMethods(['markDeleted'])
```
demiankatz: I don't think this onlyMethods serves a purpose anymore, but I could be wrong.

demiankatz (commented):
Thanks for the test revisions, I think this is a whole lot better now... just a few minor suggestions.

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
EreMaijala: Makes sense, done. They all were instances of `WriteDocumentCommand`, but `get_class` makes the intention more obvious.

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
EreMaijala: Done.

module/VuFind/tests/unit-tests/src/VuFindTest/Solr/WriterTest.php
EreMaijala: Right, removed.

demiankatz (approved):

demiankatz:
@EreMaijala, this looks good to me now. I just have one remaining question: since it's no longer required to pass an HTTP factory function, should we revert test classes that were only changed to meet that requirement? It might make this overall PR more readable. No pressure, though!

I've approved the PR, and I'll leave the final decision to you whether you'd like to roll back a couple of tests or just merge as-is. Either way, feel free to go ahead as long as CI is passing.

PR #2517, opened 2022-08-29
UnitTest for Reviews.php

skellamp:


module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/ReviewsTest.php
```diff
@@ -0,0 +1,52 @@
+<?php
+/**
+ * HoldingsILS Test Class
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2019.
```
demiankatz: You might want to put the current year in this message.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/ReviewsTest.php
```diff
@@ -0,0 +1,52 @@
...
+ * @category VuFind
+ * @package  Tests
+ * @author   Sudharma Kellampalli <skellamp@villanova.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
+ */
+
+class ReviewsTest extends \PHPUnit\Framework\TestCase
+{
+    Public function testGetDescription()
```
demiankatz: In our code style, keywords are lowercase, so I would change `Public` to `public`.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/ReviewsTest.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $obj = new Reviews();
+        $expected = 'Reviews';
+
+        $this->assertSame($expected, $obj->getDescription());
+
```
demiankatz: We remove blank lines at the end of functions.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/ReviewsTest.php
demiankatz: We also put PHPDoc comments on every function, so I'd suggest adding something like:

```
/**
 * Test getting description.
 *
 * @return void
 */
```

demiankatz: Finally, while types are optional in PHP, I find it helpful to include typehints in tests to make things as explicit as possible (I expect that in future, we'll be more strongly typing all of VuFind's code, since this is a direction the PHP language seems to be heading in). So you could add `: void` to the end of the function definition to indicate the void return type.

demiankatz (changes requested):
Thanks, @skellamp, this is a great start! I've put in a lot of picky suggestions just to give you a sense of project style. I think the automated CI checks might also suggest at least some of these changes -- if you run `phing.sh php-cs-fixer` and/or `phing.sh phpcbf` these tasks may automatically apply some of the changes for you. (Each of those Phing tasks applies a different subset of style-related normalization routines -- we use two different complementary tools for this).

demiankatz (approved):
Looks good, and all CI checks are passing. Thanks again; merging now!

PR #2518, opened 2022-08-30
Fix default sender name and address in form email handler.

EreMaijala:


EreMaijala:
Apparently this isn't covered by tests, but I couldn't spend time adding one here.

demiankatz:
Thanks, @EreMaijala. I've gone ahead and established a test to cover this situation. It would definitely be a good idea to expand the test suite here in the future.

demiankatz (approved):

PR #2519, opened 2022-09-05
Add support for cookie consent dialog and conditional services.

EreMaijala:
Adds cookie consent dialog, conditional cookie/cookieless tracking with Matomo and conditional checks in JS or template code.

Notes:
- Uses [orestbida/cookieconsent via npm](https://www.npmjs.com/package/vanilla-cookieconsent) for the actual functionality, but includes converted styles to better integrate with VuFind's themes. The one thing I couldn't quite manage to replicate here is the fancy out-of-bounds close button that lightbox uses.
- Piwik helper is using legacy invokation style where it's not possible to support consent without more extensive changes, so I decided against implementing it.

TODO:
- [x] Uses a more recent commit (orestbida/cookieconsent@777f4849785f7c3c7140143c841195ee626dacec) of vanilla-cookieconsent than actually available from npm. Also, it's still a pre-release, but is so much nicer that it's probably worth it. Needs a new release regardless, or the automatic update needs to be disabled for now.
- [x] Missing any cookies?
- [ ] Anyone need it with Google Analytics (apparently not GDPR compliant, so not needed for that)?
- [ ] Any other services that need to be loaded conditionally?
- [x] Check English translations.
- [x] See if can get Swedish translations.
- [x] Mink tests.
- [x] Unit tests for the helper.
- [x] [Documentation](https://vufind.org/wiki/configuration:cookie_consent) including linking to the privacy policy page.
- [x] Decide whether it makes sense to include iframemanager, fix the faq page.
- [x] Fix relative link.
- [x] Possibility to display information about given consent e.g. on privary policy page.


composer.json
```diff
@@ -121,7 +121,10 @@
     "scripts": {
         "phing-install-dependencies": ["phing patch-dependencies", "phing installsolr installswaggerui"],
         "post-install-cmd": "@phing-install-dependencies",
-        "post-update-cmd": "@phing-install-dependencies",
+        "post-update-cmd": [
+            "@phing-install-dependencies",
+            "[ $COMPOSER_DEV_MODE -eq 0 ] || (npm install && phing copynodemodules)"
```
demiankatz: What is the purpose of the dev mode flag here? Don't we need the node modules even if we're not in dev mode?

languages/CookieConsent/en.ini
```diff
@@ -0,0 +1,42 @@
+Accept All Cookies = "Accept All Cookies"
+Accept Only Essential Cookies = "Accept Only Essential Cookies"
+allow_video_always = "Show Video and Always Allow"
+allow_video_once = "Show Video"
+analytics_cookies_description_html = "Analytics collected with these cookies is used for development of the site."
+analytics_cookies_title_html = "Analytics Cookies"
+category_description_html = "Cookies used on the site are categorized based on their purpose. See below for description of each category as well as the option to accept or reject it. <a href="Content/privacy" target="_blank">Privacy information for the site</a>."
```
demiankatz: I think "See below for descriptions" would be better than the singular form.

I notice there's a relative URL in the `<a>` tag. Will that reliably work correctly, or do we need to use a token?

languages/CookieConsent/en.ini
```diff
@@ -0,0 +1,42 @@
...
+cookie_description_cart_html = "Used to store the contents of the book bag."
+cookie_description_consent_html = "Used to store information about whether visitors have given consent or declined the use of cookie categories used on the site."
+cookie_description_language_html = "Used to store the current user interface language."
+cookie_description_matomo_id_html = "Used for website statistics. Placed by Matomo to store a unique user ID."
+cookie_description_matomo_ref_html = "Used for website statistics. Placed by Matomo to store referrer ID's."
+cookie_description_matomo_ses_html = "Used for website statistics. Placed by Matomo to store a unique session ID."
+cookie_description_search_open_html = "Used to store open records in search results."
+cookie_description_session_html = "Used to maintain user's session state."
+cookie_description_sidefacet_html = "Used to maintain facet expand/collapse state."
+cookie_description_ui_html = "Used to store the current user interface."
```
demiankatz: Maybe "selected theme" would make more sense than "user interface" here?

languages/CookieConsent/en.ini
```diff
@@ -0,0 +1,42 @@
...
+cookie_description_youtube_bandwidth_html = "Used to estimate bandwidth."
+cookie_description_youtube_gps_html = "Used to store user's location."
+cookie_description_youtube_marketing_html = "Used to track user's actions."
+cookie_description_youtube_preferences_html = "Used to store user's preferences."
```
demiankatz: Does it make sense to include YouTube stuff in the core project, since we don't have any core YouTube support? (I'm not opposed to it, but I'm not sure I understand why this is here... is it possible there's something we should be adding via a separate PR?).

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -98,6 +96,9 @@ JS;
           $this->headScript()->appendFile("covers.js");
       }
     ?>
+    <?=$this->cookieConsent()?>
+    <?=$this->headLink()?>
+    <?=$this->headStyle()?>
```
demiankatz: Previously, all the script-related logic was together. This rearrangement separates the more special-purpose script logic from the general script output. Maybe that's unavoidable, but I wonder if there's a way to make this slightly more clear.

themes/bootstrap3/templates/content/faq.phtml
```diff
@@ -12,3 +12,12 @@
   Q: Why does the service use cookies?<br/>
   A: Cookies are used to manage sessions, which include selected language, last search, book bag contents, any login information, etc.
 </p>
+<p>
+  Q: What kinds of services have been built on top of VuFind?<br>
```
demiankatz: Is this a temporary test or something that's really meant to be included? If we're keeping it, we need the (R) symbol on VuFind (though it looks like a test to me). :-)

themes/bootstrap3/templates/footer.phtml
```diff
@@ -28,6 +28,9 @@
           <li><a href="<?=$this->url('help-home')?>?topic=search&amp;_=<?=time() ?>" data-lightbox class="help-link"><?=$this->transEsc('Search Tips')?></a></li>
           <li><a href="<?=$this->url('content-page', ['page' => 'asklibrary']) ?>"><?=$this->transEsc('Ask a Librarian')?></a></li>
           <li><a href="<?=$this->url('content-page', ['page' => 'faq']) ?>"><?=$this->transEsc('FAQs')?></a></li>
+          <?php if ($this->plugin('cookieConsent')->isEnabled()): ?>
```
demiankatz: What is the `$this->plugin` syntax here? I'm guessing this is a more verbose way of accessing view helpers, which might make sense in some contexts, but I'm not sure if there's a reason to set a new precedent. (This seems to be used in more than one place in the PR... just making sure I understand what's going on).

themes/bootstrap3/css/vendor/cookieconsent.css
```diff
@@ -0,0 +1 @@
...m__header{padding:1em 1.15em!important}#cc-main .pm__badge{display:none}#cc-main .pm__section-table,#cc-main .pm__table-body,#cc-main .pm__table-td,#cc-main .pm__table-th,#cc-main .pm__table-tr{display:block}#cc-main .pm__table-head{display:none}#cc-main .pm__table-tr{border-top:1px solid #f0f4f7;border-top:1px solid var(--cc-separator-border-color)}#cc-main .pm__table-td{display:flex;justify-content:space-between}#cc-main .pm__table-td:before{color:#2c2f31;color:var(--cc-primary-color);content:attr(data-column);flex:1;font-weight:600;min-width:100px;overflow:hidden;padding-right:2em;text-overflow:ellipsis}#cc-main .pm__table-td>div{flex:3}#cc-main .pm__btn-group,#cc-main .pm__footer{flex-direction:column!important}#cc-main .pm__btn-group{display:flex!important}#cc-main .pm__btn-group:nth-child(2),#cc-main .pm__btn:nth-child(2){margin:6px 0 0!important}#cc-main .pm--flip .pm__btn-group,#cc-main .pm--flip .pm__footer{flex-direction:column-reverse!important}#cc-main .pm--flip .pm__btn-group:nth-child(2),#cc-main .pm--flip .pm__btn:nth-child(2){margin:0 0 6px!important}.show--preferences #cc-main.c--anim .pm{-webkit-transform:translateY(0)!important;transform:translateY(0)!important}}
```
demiankatz: Should this file be copied in by Phing as well?

demiankatz (commented):
@EreMaijala, I haven't looked at everything in depth yet, but here are a few quick comments that might be helpful...

composer.json
EreMaijala: The idea I had was that updating the dependencies would be a dev operation since the node modules are not needed for normal operation but are only used for updating the files in theme's js/vendor. This also means that files tracked by git could change by the update operation. But this was just an idea, and doing it regardless of mode would probably work fine since at least as long as versions are pinned.

themes/bootstrap3/templates/content/faq.phtml
EreMaijala: That's just a test for iframemanager, I'll need to think about it. It's also possible to leave out the iframemanager support, but I think that it can be a relatively popular choice to include some videos or social media content e.g. on the front page.

themes/bootstrap3/css/vendor/cookieconsent.css
EreMaijala: No, it's covered by the new less files that give more control over the colors and make it blend with VuFind better.

composer.json
demiankatz: Ahh, good point, since the files are in Git, it's not essential. Makes sense; I hadn't fully thought it through. :-)

(My initial concern is that I build the official release packages with a non-dev installation, but that won't be harmed by this approach).

themes/bootstrap3/templates/content/faq.phtml
demiankatz: Makes sense. Maybe we just need a comment somewhere to clarify that VuFind doesn't use YouTube for anything by default, but this setting supports local use where applicable.

themes/bootstrap3/css/vendor/cookieconsent.css
demiankatz: If it's being locally built, should it be somewhere other than vendor? If it relies on third party files, where are they coming from? (Sorry, I haven't studied the less files closely yet, so maybe I'm missing something obvious).

themes/bootstrap3/templates/footer.phtml
EreMaijala: That's to get the plugin without calling __invoke. I suppose the other option would be to change __invoke, but I'm not sure which one is better.

demiankatz: Ahh, I see! If we need to avoid calling __invoke, is that a sign that the invoke method is doing too much work? (The helper code is another piece I haven't reviewed yet, so again, my apologies if the answer is obvious... I'll find out in a couple of days when I have more time!)

EreMaijala: Indeed. I flip-flopped a bit about it but left it as is for now. The primary function of outputting the code needed for the cookie consent component is now done via __invoke the same way it's done with most simple helpers, but this can obviously be changed.

themes/bootstrap3/templates/content/faq.phtml
EreMaijala: Maybe it's enough to document in the wiki how it works so that no embedded video or such is included in the core package. At least the configuration should be available since it took some effort to list all the cookies YouTube uses (and I'm still not sure I didn't miss any!).

languages/CookieConsent/en.ini
EreMaijala: See below for some explanation. Not important to us, though, since we'll just add it locally as necessary, but could be useful as an example.

languages/CookieConsent/en.ini
EreMaijala: Yes, fixed.

languages/CookieConsent/en.ini
EreMaijala: Fixed.

themes/bootstrap3/templates/layout/layout.phtml
EreMaijala: That was done because the helper added css files which would have been missing, but now that the styles come from less files it's not needed anymore. Now reverted.

themes/bootstrap3/css/vendor/cookieconsent.css
EreMaijala: I adopted the less files into the core so they're included with `grunt less`. I wouldn't consider them third-party anymore since they're pretty VuFind-specific.

languages/CookieConsent/en.ini
demiankatz: Since the YouTube strings don't actually mention YouTube in any way, is there a way to make the strings more generic/reusable so that other similar services can leverage them without the need to customize language files?

themes/bootstrap3/templates/content/faq.phtml
demiankatz: I guess a related question (which may not be answerable without the experience gained by time) is, "how fragile is this?" I would imagine that YouTube could change its cookie strategy at any time without notice and break things. Perhaps that's an argument that we should put service-specific settings on a wiki page, and put a link to that wiki page in the config file. Then, if people want to add something, they can find the latest working settings easily, even if they're running a slightly out-of-date VuFind release. And then we could set up a directory of other services like Vimeo, etc., as needed. (Though I still could be persuaded to keep these in if you have strong feelings about it -- still in brainstorming mode at the moment).

languages/CookieConsent/en.ini
EreMaijala: Good idea, done!

themes/bootstrap3/templates/content/privacy.phtml
```diff
@@ -0,0 +1,7 @@
+<?php $this->headTitle($this->translate('Privacy Policy')); ?>
```
demiankatz, line 1: I noticed that this file slipped in as part of a seemingly-unrelated commit. Was that intentional?

EreMaijala: Should have been a part of another commit, but having a template for privacy policy and link to it from the default texts was intentional. If you feel that's too much here, I'll just leave them out, but having it would help with Mink tests too.

themes/bootstrap3/templates/content/faq.phtml
EreMaijala: That is an issue, but apparently unavoidable. I've discussed this very issue with our specialist and he said that we must display all the information in the cookie consent preferences. I suppose that we can only do our best. From what I've gathered, there are quite a few cookie consent dialogs around with less cookies for YouTube, which shows how difficult it is to get right.

EreMaijala: Also, some of the cookies, or at least their expiration times, are regional or dependent on whether the user is logged in to Google services, so the wiki docs must include a point to verify their correctness for any particular case.

themes/bootstrap3/templates/content/privacy.phtml
demiankatz: No objection, just wanted to be sure it was intentional and not an orphaned file. As long as there's a link to it somewhere, it's fine with me. We should probably make a note in an appropriate place to document the need for users to populate this if they want to show a privacy policy (if that hasn't already been done). It might also be valuable to have a setting that lets the privacy policy link to an external URL if desired (but maybe that's easily achievable through template customization, which would be fine too).

EreMaijala: We should probably have something like a "Checklist for Production Use" wiki page. We have a lot of different topics, but I'm not sure if there's a good place where to put all the things you should verify before switch to production. A typical place for a link to privacy policy would be the footer, and I suppose that's one the templates that would be modified in most cases anyway.

demiankatz: Yes, a production checklist is a good idea -- and something that we could link to as a last step from all the installation notes pages. Is it worth adding a TODO to this PR to create a "production notes" section on the privacy policy wiki page (which I assume we will eventually create), and then once that is built, we can create a higher-level page to link to that along with other relevant production notes?

themes/bootstrap3/templates/content/faq.phtml
demiankatz: Understood! Do you have a preference at this point about whether to lean toward the wiki or the built-in configuration? If you like the idea of the wiki, please feel free to start building out pages whenever it's convenient (and let me know if you need help establishing anything or deciding where to put things).

EreMaijala: Built-in is a must for us. Otherwise we'd have to somehow embed the wiki page into the consent preferences dialog, which would be ugly and dependent on the wiki being available. Making the user go to a third-party site, no matter how friendly, is a no-no.

EreMaijala: Regardless, a wiki page is needed for instructions on how to set up additional services and verify current settings.

demiankatz: I think I may not have clearly expressed my idea -- I wasn't talking about redirecting end users to our wiki; I was proposing that we have YAML sections pre-built in the wiki that can be copied and pasted into the local YAML, so we can publish configurations for a variety of third-party services users might potentially want to embed in VuFind, without having to include them in a single huge file in the project... and that would allow us to keep them up to date without having to issue new VuFind releases. Then the YAML file in the project could just reference the wiki and suggest pulling in whatever section(s) are needed.

demiankatz: ...but maybe there are compelling reasons to argue FOR including everything in the project, in which case I suppose we could also include a comment pointing to the current version of the file in GitHub and suggest that people watch it for changes if they're using older releases.

EreMaijala: Sorry, I didn't get it! Having yaml pieces on the wiki page would make sense. That leaves the question whether it makes sense to include something in the distributed yaml file as an example or leave it out.

demiankatz: Maybe if we include an example, it could be a fictitious one (pointing at a myuniversity.edu site, or something) so we don't need to worry about it getting out of date and causing confusion... and as long as we have a comment pointing at the wiki for real settings, hopefully that will allow people to get what they need fairly easily.

EreMaijala: One potential pitfall with having yaml in the wiki is that we need to make sure not to introduce changes that are incompatible with older versions. I suppose that shouldn't be a huge issue since we try to maintain back-compatibility for configuration in any case.

demiankatz: Yes, no solution is entirely foolproof, but as you say, this is just a further reason to be careful about backward-compatibility, which we already try to do. And worst case scenario where something changes dramatically, we can always put a note on the top of the wiki saying "These notes require version X at a minimum" and link to a prior revision for legacy users.

config/vufind/CookieConsent.yaml
```diff
@@ -0,0 +1,132 @@
...
+# supported: matomo
+#
+# The following placeholders can be used in settings and translated texts:
+# {{consent_cookie_name}}         - Name of the consent cookie (default cc_cookie)
+# {{consent_cookie_expiration}}   - Consent cookie expiration time
+# {{current_host_name}}           - Server name of the VuFind server
+# {{vufind_cookie_domain}}        - Domain for VuFind's cookies
+# {{vufind_session_cookie}}       - VuFind session cookie name
+CookieConsent:
+  #CookieExpiration: 365
```
demiankatz, line 61: What is the purpose of this commented-out setting?

config/vufind/CookieConsent.yaml
```diff
@@ -0,0 +1,132 @@
...
+          ThirdParty: true
+          Description: CookieConsent::cookie_description_third_party_marketing_html
+          Expiration: 8
+          ExpirationUnit: months
+      AutoClearCookies:
+        - Name: '/^(yt-player-.*|GPS|VISITOR_INFO1_LIVE|PREF|__Secure-YEC|YSC|CONSENT|SOCS)$/'
+  IframeServices:
+    youtube:
+      EmbedUrl: 'https://www.youtube.com/embed/{data-id}'
+      #ThumbnailUrl: 'https://i3.ytimg.com/vi/{data-id}/hqdefault.jpg'
```
demiankatz: Another commented-out service that I'm curious about. (Just figure we should document or remove, as appropriate).

themes/bootstrap3/less/components/cookie-consent-modal.less
```diff
@@ -0,0 +1,646 @@
...
+            .cm__btn-group {
+                flex-direction: column-reverse!important;
+            }
+
+            .cm__btn-group:nth-child(2),
+            .cm__btn:nth-child(2) {
+                margin: 0 0 @cc-cm-btn-margin 0!important;
+            }
+        }
+	}
```
demiankatz, line 608: Is there an indentation problem here?

themes/bootstrap3/js/vendor/iframemanager.js
```diff
@@ -0,0 +1,17 @@
+/*
+ iframemanager v1.0
```
demiankatz: Should we be loading this via NPM as well?

module/VuFind/src/VuFind/View/Helper/Root/CookieConsent.php
```diff
@@ -0,0 +1,441 @@
...
+        array $config,
+        array $consentConfig,
+        CookieManager $cookieManager
+    ) {
+        $this->config = $config;
+        $this->consentConfig = $consentConfig;
+        $this->cookieManager = $cookieManager;
+        $this->consentCookieName = $this->consentConfig['CookieName'] ?? 'cc_cookie';
+        $this->consentCookieExpiration
+            = $this->consentConfig['CookieExpiration'] ?? 182;
```
demiankatz: Why 182? Just seems like an oddly specific magic number...

themes/bootstrap3/templates/footer.phtml
demiankatz: Makes sense. Either way, it makes some code more verbose and other code less verbose. I think having a render() method on the helper instead of rendering with __invoke might be slightly more clear, though -- generally if a helper has multiple callable methods, we just use the default __invoke (which just returns the object itself).

themes/bootstrap3/templates/content/faq.phtml
EreMaijala: I've now created the initial [wiki page](https://vufind.org/wiki/configuration:cookie_consent) with YouTube sample config.

On a second thought I'm not sure how useful it would be to include an example in the default yaml, so maybe it's easier to just leave it out since it also reduces the number of translation strings.

config/vufind/CookieConsent.yaml
EreMaijala: It can be used to change the expiration for the consent cookie.

config/vufind/CookieConsent.yaml
EreMaijala: It's not in the default config anymore, but needs to be documented in the wiki (like all the other settings).

themes/bootstrap3/less/components/cookie-consent-modal.less
EreMaijala: Ah, a couple of stray tabs, fixed.

themes/bootstrap3/js/vendor/iframemanager.js
EreMaijala: Yes, but unfortunately it's not available there.

module/VuFind/src/VuFind/View/Helper/Root/CookieConsent.php
EreMaijala: Half a year rounded down. That's the default for the cookieconsent module, but I'm not sure where it comes from. I'll let you know if I find a clarification for it.

themes/bootstrap3/templates/content/faq.phtml
demiankatz: Thanks, it looks good. I agree that reducing the amount of extra translation work we need to do is helpful!

config/vufind/CookieConsent.yaml
demiankatz: Makes sense, so I guess we just need to document it. Then the question becomes, do we document it in a big comment at the top of the YAML or in the wiki? I lean toward "big comment" since that ensures that the documentation and configuration file are always in sync. We can use the wiki for example configs and link to the YAML file itself for more detailed documentation (unless you feel otherwise).

themes/bootstrap3/less/components/cookie-consent-modal.less
demiankatz: Thanks!

themes/bootstrap3/js/vendor/iframemanager.js
demiankatz: Too bad! Maybe we could add a download task for it in Phing, then... though it's annoying to have to manage versions of things in multiple places. (I guess there's also the question of whether/how well this will be maintained, since as of right now it has only one release).

module/VuFind/src/VuFind/View/Helper/Root/CookieConsent.php
demiankatz: Ahh, of course. Maybe a comment `// half a year` at the end of the line would be enough to satisfy curiosity. :-)

config/vufind/CookieConsent.yaml
```diff
@@ -0,0 +1,141 @@
...
+#
+# Settings:
+#
+#   CookieExpiration  Days before the consent cookie expires. Default is 182 days
+#                     (half a year). One year is another typical value.
+#   Categories        Cookie categories (groups) that bring together a set of
+#                     cookies. The user can choose their consent for each group
+#                     apart from essential cookies. See below for documentation.
+#   IframeServices    Configuration for iframe services that depend on cookie
+#                     consent. See below for documenation.
```
demiankatz: Typo: documenation.

config/vufind/CookieConsent.yaml
```diff
@@ -0,0 +1,141 @@
...
+        - Name: 'cc_*'
+          Domain: '{{vufind_cookie_domain}}'
+          Description: CookieConsent::cookie_description_consent_html
+          Expiration: '{{consent_cookie_expiration}}'
+          ExpirationUnit: days
+        - Name: language
+          Domain: '{{vufind_cookie_domain}}'
+          Description: CookieConsent::cookie_description_language_html
+          Expiration: session
+        - Name: sidefacet-* (session storage)
```
demiankatz, line 83: I'm a little confused by the `(session storage)` here. I was assuming that the names were being used to actually match up cookies for functional reasons. Does that mean that `(session storage)` has a specific meaning that should be documented? Or is this field purely descriptive (in which case, how are the actual cookies matched up?)

config/vufind/CookieConsent.yaml
```diff
@@ -0,0 +1,141 @@
...
+          Expiration: session
+        - Name: '{{vufind_session_cookie}}'
+          Domain: '{{vufind_cookie_domain}}'
+          Description: CookieConsent::cookie_description_session_html
+          Expiration: session
+        - Name: ui
+          Domain: '{{vufind_cookie_domain}}'
+          Description: CookieConsent::cookie_description_ui_html
+          Expiration: session
+        - Name: vufind_cart, vufind_cart_src
```
demiankatz, line 95: Is comma-separated list of names something else that needs to be documented?

demiankatz (changes requested):
As we discussed on Slack, I gave this another look with a focus on CookieConsent.yaml. I left a few comments regarding the text/formatting of the file. I also did a review of VuFind code in case we missed any cookies. Here's what I found -- I don't think most are important, but I wanted to be comprehensive:

  * VuFindTestSuiteRunning is checked in some places, though obviously it only matters in test mode; do we need to account for it?
  * The upgrade process sets some `vfup*` cookies, though this is only going to be used by administrators; again, do we need to account for it?
  * VuFind\Session\SecureDelegator uses a [session name]_KEY cookie for session encryption. I think that might need to be addressed for scenarios where secure sessions are enabled (unless it already is and I missed it).

I hope this helps! Please let me know when you're ready for me to look again and/or do hands-on testing.

config/vufind/CookieConsent.yaml
EreMaijala: Name in the Cookies array is just descriptive. I'll add a note. It's only in `AutoClearCookies` where it has functionality.

config/vufind/CookieConsent.yaml
EreMaijala: Yes, in the essence that it's free text.

EreMaijala:
@demiankatz Thanks! I've added the vfup* and *_KEY cookies but left them commented out by default. I think `VuFindTestSuiteRunning` can be left out as it should never be set for end users.

EreMaijala:
vanilla-cookieconsent is now up to date and updatable via npm. I also redid the less files as more direct conversions of the original scss files.

demiankatz:
Thanks for the progress on this, @EreMaijala! I think I've resolved most, if not all, of the conversation threads that are now finished. I think there might still be more wiki work to do in future (beyond the scope of this PR). I didn't have time for hands-on testing today, but I'll try to get to that soon!

themes/bootstrap3/js/vendor/iframemanager.js
EreMaijala: Making it reliable and somehow integrate with npm's or composer's 'outdated' check might be bit of a burden. The library hasn't had any frequent updates either, so while this is not optimal, I feel like the additional work is not worth the trouble.

EreMaijala:
@demiankatz If you have any specific wishes for the wiki, please let me know and I'll make sure to document everything needed. E.g. the privacy policy part is currently just [one topic under User Privacy](https://vufind.org/wiki/configuration:user_privacy#privacy_policy), but I didn't feel like it would warrant its own page. The switch to production checklist would be nice, but out of scope here, I think.

themes/bootstrap3/templates/footer.phtml
EreMaijala: I failed to mention that I added the render() method instead of __invoke doing it. Now that the plugin has several more methods available, it makes even more sense this way.

themes/bootstrap3/js/vendor/iframemanager.js
demiankatz: From my perspective, the real question is whether this code has a future. If it does not, are we just setting ourselves up for some kind of emergency rewrite/replacement down the road? How heavily do we rely on this, and is there anything we can/should do now to protect ourselves against future problems?

EreMaijala: That's a difficult question. There are no commits this year, though since it's doing what it says that's not a good indication. There are several feature wishes and one PR with no reaction, though.

Since this is somewhat problematic, I'd be fine stripping iframemanager out of the default setup and leave handling of videos etc. to a local method. In its simplest form it would be a check for active category in the template and display of message about cookies instead of the video if the category is not enabled. The events etc. should allow one to use iframemanager locally if deemed desirable. What do you think?

demiankatz:
@EreMaijala, I agree that, documentation-wise, the outstanding item is the production checklist, which is beyond the scope of this PR. I'll put a note about it on the next PMC meeting agenda, so we can brainstorm there as a group exactly what it should include, where it should live, and what should link to it. That way we won't lose track of the task, and we can end discussion of it here. :-)

themes/bootstrap3/js/vendor/iframemanager.js
demiankatz: If it's easy to separate the iframemanager piece, I'd even be open to moving that to a separate PR -- then we could merge the base behavior right away, and leave the iframemanager PR open a little longer to see how things develop and consider alternate approaches if necessary.

EreMaijala: It's now gone from here, and I replaced the sample template on the wiki page with one that doesn't need iframemanager.

EreMaijala: One approach that would eliminate the need for iframemanager or such but would make it easier to embed videos is to add a view helper for that. It would then render the template code like in the [YouTube example](https://vufind.org/wiki/configuration:cookie_consent#youtube).

demiankatz: Yes, definitely worth considering! I guess to a large extent it may come down to community needs and whatever makes your local code easier to maintain. I'm open to a variety of possibilities; I just wanted to be careful about setting a bad precedent for accepting questionable dependencies. :-)

EreMaijala: And I appreciate that! :)

demiankatz:
Thanks, @EreMaijala, this looks good and seems to work as expected. One question: is there a way to bring the consent dialog back up again and change settings? This seems like it might be a useful feature -- could we perhaps have an "edit" link on the privacy page to trigger this?

EreMaijala:
@demiankatz There's a link in footer, but I forgot to add one to the privacy page where it quite obviously should be as well. Now added in a very simple form.

demiankatz (approved):

config/vufind/CookieConsent.yaml
```diff
@@ -0,0 +1,134 @@
...
+        - Name: '{{consent_cookie_name}}'
+          Domain: '{{vufind_cookie_domain}}'
+          Description: CookieConsent::cookie_description_consent_html
+          Expiration: '{{consent_cookie_expiration}}'
+          ExpirationUnit: days
+        - Name: 'cc_*'
+          Domain: '{{vufind_cookie_domain}}'
+          Description: CookieConsent::cookie_description_consent_html
+          Expiration: '{{consent_cookie_expiration}}'
+          ExpirationUnit: days
```
stweil, line 78: @EreMaijala, can these lines be removed? Are there such cookies which are not already handled by the lines 69...73?

demiankatz: @stweil, @EreMaijala is on vacation for several weeks, so he will probably not answer this question for some time. I am not entirely sure of the answer, though it seems likely that this changed when the cookie consent system was refactored in #5163 -- as far as I can tell, we're only using the primary consent cookie, so any other cc_* cookies may have been related to the old consent library, and not needed for backward compatibility.

I can think of two possible approaches:

1.) Open a PR to remove the potentially redundant lines, request @EreMaijala's review, and wait until he returns to get it approved and merged (this is much less likely to get lost/forgotten than a comment on a merged PR).

2.) Open a PR to consolidate the cookie consent lines, so it's a single entry named `{{consent_cookie_name}}, cc_*` instead of two separate redundant entries. We seem to use comma-separated cookie name lists elsewhere, so I think this is legal. This makes the display more compact but reserves the option to use additional cookie consent cookies in the future if a need should arise.

If the goal is to fully eliminate the redundant entry and there's no rush to do it, option 1 may be better. If you think we should take some action before release 11.1, option 2 may be better, because I would feel more confident in merging that without @EreMaijala's input, and he won't be back before the release is completed.

EreMaijala: @stweil Ah, I missed this one. There should no longer be a need for other cookies with `cc_` prefix, so the forementioned lines can be removed. I'd appreciate a pull request to do that.

stweil: Thanks! See PR #5494.

PR #2520, opened 2022-09-06
Optimize use of iterators in custom SolrMarc methods.

demiankatz:
This is a follow-up to #2497, optimizing existing SolrMarc iterator code to use for loops instead.

There is also one minor functional change: the illustration detection logic has been revised to recognize RDA-style, non-abbreviated "illustrations" in the description (in addition to legacy styles).

As of this writing, I have confirmed that everything compiles and that records can still be indexed, but I haven't done granular testing of all of the affected code. I will try to do that when time permits (see TODO list below).

TODO
- [x] Actually test all changed methods.


demiankatz:
@mtrojan-ub, I've requested a review from you, though as noted above, I haven't done thorough hands-on testing of anything yet. Please let me know if you have a preference as to whether you'd like to review now, pre-testing, to potentially catch any errors I have made, or if you would rather wait until after I have done fuller testing before doing your review. I'm happy with either approach!

mtrojan-ub (approved):
@demiankatz, thanks for the follow-up! I think these changes are pretty straightforward so i will gladly approve the changes even before you do the testing.

demiankatz:
Thanks, @mtrojan-ub! I expect it should be fine, but I like to test everything by hand just to be 100% sure there's not some tiny subtle bug I'm overlooking. I'll merge this once I've had a chance to do that (which may not be for a little while...).

demiankatz:
I finally found time to test everything, and it all checks out. Merging now!

PR #2521, opened 2022-09-07
fix loading of FacetList content block

xrosecky:


demiankatz:
Thanks, @xrosecky, is this to fix a problem with hierarchical facets? If so, do you have any idea when the problem was introduced? Does this need to be backported to the release-8.1 branch, or does this only affect dev? I imagine it might be valuable to add a Mink test to cover the situation and prevent future regressions.

I'm going to be busy with the VuFind Summit this week, but please let me know if you'd like help with anything after that! I'll do some hands-on testing at that time as well, but thought it might be worthwhile to ask these questions now while I had a moment to do so. :-)

xrosecky:
It seems to be broken from this commit: https://github.com/vufind-org/vufind/commit/035d3329fec3c799a622983ef7a19d8aa5255617

demiankatz:
Thanks again, @xrosecky -- looks like the bug was introduced recently, so there's no need to backport anything. I have just updated this PR with a test case to prevent regressions, and I will merge as soon as CI passes.

demiankatz (approved):

PR #2522, opened 2022-09-09
Improved interface for active filter removal

crhallberg:
- Remove icon
- Show removal with line-through
- Improve contrast
- Move sr text to more clearly identify links

image


demiankatz:
Thanks, @crhallberg. This looks reasonable to me across all three themes, but I don't have really strong opinions about the change. Curious to hear what @EreMaijala thinks, since a lot of the original design comes from him.

EreMaijala (changes requested):
While this looks (and works) fine when you know what you're looking at, I think this affects the discoverability of the function particularly on mobile devices where you can't rely on the hover effect.

crhallberg:
To address @EreMaijala's concern, I could keep the x icon for mobile but this slashing and red styling is definitely much clearer to me.

EreMaijala:
@crhallberg Is it possible to have the effect also on keyboard focus? A focus rectangle would be great as well. It didn't have that before either, but the x changed color on focus so you had at least some sort of input.

I'm still a bit hesitant on the underlining changing to overstrike. Without underlining the addition of overstrike would be great, but the transition feels weird to me, kind of like I'd be losing the originally implied action.

EreMaijala (changes requested):
Having thought about this, I think the link shouldn't be underlined. The whole element is more of a button, so underlining doesn't convey more information. The focus indication would be important to have.

demiankatz:
Thanks, @EreMaijala; I'm inclined to agree that this is more clear without the underlining, and that keyboard focus highlighting would help a lot.

@crhallberg, I noticed there were some conflicts here, so I resolved them in order to test; you might want to double-check my work.

I'm also bringing @ckaz into this conversation, because I think he might have helpful opinions as well.

ckaz:
Great idea but I fully agree with @EreMaijala : For accessibility we need the focus indication and it would make more sense to me too, if there was no underline (which I wouldn't expect of a button type-looking interaction element anyway). So my preference would be 
* Strikethrough on hover/focus / x-icon on mobile
* Focus rectangle on focus
* No underline

demiankatz (approved):
Thanks, @crhallberg, this looks much better to me. Curious what @EreMaijala and @ckaz think at this point; I'll wait for their feedback before merging.

EreMaijala (approved):
Looks good to me!

I noticed the strikethrough overflowing on the right, but I tested it without, and it looked worse to me, so I've no problem with that.

ckaz (approved):
Looks good to me. Pls. go ahead!

demiankatz:
Thanks, everyone -- just merged!

PR #2523, opened 2022-09-12
Upgrade sqlite-jdbc from 3.27.2 to 3.39.3.0 for improved platform sup…

EreMaijala:
…port.

Adds support for more recent Java version at least up to JDK 17, and macOS with Apple Silicon. The latter has been tested, and I've also run all VuFind tests with JDK 18.


demiankatz (approved):
Thanks!

PR #2534, opened 2022-09-14
Add sodium to dependency check in installer.

EreMaijala:
I haven't been able to really test this since the php packages I've tried come with sodium by default.


demiankatz (approved):
I tested this on a Linux distribution that lacks sodium by default, and it worked as expected. Thanks!

Now the question is why some distributions lack sodium, since the PHP manual says it's been bundled into the core since PHP 7.2 ... but this is clearly not true in all cases. (I encountered this problem with both PHP 7.4 and PHP 8.0 installed from the remi repositories into Enterprise Linux).

PR #2535, opened 2022-09-14
Make it possible to add HTTP options in Solr Connector.

EreMaijala:
This makes it easier to specify any additional options for the HTTP client in Solr Connector without having to override the whole createConnector method and restores flexibility that was lost in the previous refactoring.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2536, opened 2022-09-15
Avoid calling urlencode() with null.

EreMaijala:
At least EDS has sort set to null.

Note that a change could be in order in the EDS Params code, but it explicitly says it needs to be null, so changing that to an empty string or such without more extensive research was too scary.


demiankatz (approved):
It makes sense to address this at a low level, since null values could potentially come from a lot of places. Thanks!

PR #2538, opened 2022-09-15
Disable default filters when using an alphabetic search result

damien-git:
See JIRA [VUFIND-1577](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1577) ("Default filters should not be applied when clicking on an alphabetic search result.")

### DIFF
--- a/config/vufind/config.ini
+++ b/config/vufind/config.ini
@@ -1900,6 +1900,14 @@ page_size = 20
 rows_before = 0
 ; highlight the match row (or spot where match would have been)? default false
 highlighting = false
+; AlphaBrowse results are not subject to dynamic filtering. If you have default
+; filters defined in searches.ini, you most likely will want to disable them when
+; users navigate from browse results to search results, to ensure that the result
+; count is consistent between the browse listing and the search listing. However,
+; in the rare situation that you have configured default filters that BROADEN
+; rather than NARROW search results, you will likely want to change this setting
+; to false to avoid inconsistencies.
+bypass_default_filters = true
 ; SEE ALSO: the General/includeAlphaBrowse setting in searchbox.ini, for including
 ; alphabrowse options in the main search drop-down options.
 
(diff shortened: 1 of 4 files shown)


demiankatz:
Thanks, @damien-git, I'm glad this worked for you, and I appreciate you taking the time to get this PR started!

It occurs to me that we should probably make this behavior configurable (maybe via a `bypass_default_filters` setting in the `[AlphaBrowse]` section of config.ini). In the vast majority of cases, you are correct that this is likely to be helpful... but since it is theoretically possible to create default filters that narrow, rather than expand, the search results, there might be cases where clearing the default filters actually makes things worse. It might be helpful to have a setting to fix this in those scenarios.

What do you think? If you like the idea, I'm happy to help with the implementation (if you don't mind me pushing to this branch).

damien-git:
@demiankatz Do you mean some filters would expand the search results ? In my case the filter was narrowing the results, and I didn't realize it could be the opposite. In that case it would make sense to add a config setting. I certainly don't mind you changing this branch.

demiankatz:
Thanks, @damien-git. As you say, it is possible to set up filters that expand search results rather than limit them (by using the "inverted facets" feature introduced in release 8.1 -- see PR #2340 for details). I think this is fairly rarely used, but it's still an option we should account for.

I found that I did not have permission to push to your branch, so I instead created a pull request against it (https://github.com/MSU-Libraries/vufind/pull/1) with some proposed changes. Please merge that PR if you are happy with it, and then we can finish up the work here.

I also notice that you created this PR using the dev branch of your fork. In future, I would recommend creating a new branch with a more specific name for each PR you create -- it will make things less confusing for you if you ever have more than one thing going on at any given time. :-)

damien-git:
@demiankatz OK. Your PR looks good, I merged it.

demiankatz (approved):
Thanks again, @damien-git! I will merge this now, so it will be included in release 9.0.

PR #2539, opened 2022-09-16
Fix MarcReaderTrait::getFirstFieldValue when field does not exist.

EreMaijala:
- Avoids returning null when string is the only documented return type.
- Adds tests to ensure several different kinds of methods return the correct result when the requested metadata is not available.
- Fixes PHP 8.1 deprecation notices.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2543, opened 2022-09-20
ThemeInfo: Fix getMergedConfig

crhallberg:
- [x] Fixed flattened method to match actual theme behavior
- [x] Remove non-flattened parameter
- [x] Upgrade tests
- [x] Add changelog note about backward compatibility breaks when merging

### DIFF
--- a/module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
+++ b/module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
@@ -241,6 +241,11 @@ class ThemeInfo
             return $current;
         }
 
+        // Early escape for string values
+        if (!is_array($merged) || !is_array($current)) {
+            return $merged;
+        }
+
         if ($this->isStringKeyedArray($merged)) {
             foreach ($current as $key => $val) {
                 if (!array_key_exists($key, $merged)) {
@@ -258,12 +263,8 @@ class ThemeInfo
             return $merged;
         }
 
-        if (is_array($merged)) {
-            // capture unique or missing array items
-            return array_merge($current, $merged);
-        }
-
-        return $merged;
+        // capture unique or missing array items
+        return array_merge((array)$current, (array)$merged);
     }
 
     /**
--- a/module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
+++ b/module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
@@ -232,44 +232,49 @@ class ThemeInfo
      * Using the string-keyed array format of theme config info,
      * recursively walk the array, capturing unique or missing values.
      *
-     * @param array|string $merged  Merged theme info, overrides parent theme
-     * @param array|string $current Theme info to be merged
+     * @param array|string $children  Merged theme info, overrides parent theme
+     * @param array|string $parent Theme info to be merged
      *
      * @return array|string merged theme info, favoring child themes (merged)
      */
-    protected function mergeWithoutOverride($merged, $current)
+    protected function mergeWithoutOverride($children, $parent)
     {
-        if (empty($merged)) {
-            return $current;
+        if (empty($children)) {
+            return $parent;
         }
 
         // Early escape for string, number, etc. values
-        if (!is_array($merged) && !is_array($current)) {
-            return $merged;
+        if (!is_array($children) && !is_array($parent)) {
+            return $children;
         }
 
-        if ($this->isStringKeyedArray($merged)) {
-            foreach ($current as $key => $val) {
-                if (!array_key_exists($key, $merged)) {
+        if ($this->isStringKeyedArray($children)) {
+            if (!$this->isStringKeyedArray($parent)) {
+                // don't override if incompatible
+                return $children;
+            }
+
+            foreach ($parent as $key => $val) {
+                if (!array_key_exists($key, $children)) {
                     // capture missing string keys
-                    $merged[$key] = $val;
+                    $children[$key] = $val;
                 } elseif ($this->isStringKeyedArray($val)) {
                     // recurse
-                    $merged[$key] = $this->mergeWithoutOverride($merged[$key], $val);
+                    $children[$key] = $this->mergeWithoutOverride($children[$key], $val);
                 } elseif (is_array($val)) {
                     // capture unique or missing array items
-                    $merged[$key] = array_merge($val, (array)$merged[$key]);
-                } elseif (is_array($merged[$key])) {
+                    $children[$key] = array_merge($val, (array)$children[$key]);
+                } elseif (is_array($children[$key])) {
                     // string -> array
-                    $merged[$key] = array_merge((array)$val, $merged[$key]);
+                    $children[$key] = array_merge((array)$val, $children[$key]);
                 }
             }
 
-            return $merged;
+            return $children;
         }
 
         // capture unique or missing array items
-        return array_merge((array)$current, (array)$merged);
+        return array_merge((array)$parent, (array)$children);
     }
 
     /**
@@ -303,6 +308,7 @@ class ThemeInfo
                 $allThemeInfo[$currentTheme]['mixins'] ?? [],
             );
 
+            // from child to parent
             foreach ($currentThemeSet as $theme) {
                 if (isset($allThemeInfo[$theme])
                     && (empty($key) || isset($allThemeInfo[$theme][$key]))
--- a/module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
+++ b/module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
@@ -211,34 +211,80 @@ class ThemeInfo
     }
 
     /**
+     * Determine if a variable is a string-keyed array
+     *
+     * @param mixed $op Variable to test
+     *
+     * @return boolean
+     */
+    protected function is_string_keyed_array($op) {
+        if (!is_array($op)) {
+            return false;
+        }
+
+        foreach ($op as $key => $val) {
+            if (is_string($key)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Add parent theme information to a collection of merged theme info.
+     * Using the string-keyed array format of theme config info,
+     * recursively walk the array, capturing unique or missing values.
+     *
+     * @param array $merged Merged theme info, overrides parent theme
+     * @param array $current Theme info to be merged
+     *
+     * @return array merged theme info, favoring child themes (merged)
+     */
+    protected function mergeWithoutOverride($merged, $current) {
+        if (empty($merged)) {
+            return $current;
+        }
+
+        if ($this->is_string_keyed_array($merged)) {
+            foreach ($current as $key => $val) {
+                if (!array_key_exists($key, $merged)) {
+                    // capture missing string keys
+                    $merged[$key] = $val;
+                } elseif ($this->is_string_keyed_array($val)) {
+                    // recurse
+                    $merged[$key] = $this->mergeWithoutOverride($merged[$key], $val);
+                } elseif (is_array($val)) {
+                    // capture unique or missing array items
+                    $merged[$key] = array_merge($merged[$key], $val);
+                }
+            }
+
+            return $merged;
+        }
+
+        if (is_array($merged)) {
+            // capture unique or missing array items
+            return array_merge($merged, $current);
+        }
+
+        return $merged;
+    }
+
+    /**
      * Get a configuration element, merged to reflect theme inheritance.
      *
-     * @param string $key     Configuration key to retrieve (or empty string to
+     * @param string $key Configuration key to retrieve (or empty string to
      * retrieve full configuration)
-     * @param bool   $flatten Use array_replace to flatten values
      *
-     * @return array
+     * @return array|string
      */
-    public function getMergedConfig(string $key = '', bool $flatten = false): array
+    public function getMergedConfig(string $key = ''): array|string
     {
         $currentTheme = $this->getTheme();
         $allThemeInfo = $this->getThemeInfo();
 
-        /**
-         * Assume a parent value 'a' and a child value 'b'
-         *
-         * Using array_merge (default) will merge them into ['b', 'a']
-         * Using array_replace ($flatten = true) will merge them into 'b'
-         *
-         * We're using an anonymous function here to swap the arguments in the
-         * flatten case. This is to make sure child values override parent values
-         * with replace but parent values are appended to the end of merged values
-         */
-        $deepFunc = !$flatten
-            ? 'array_merge_recursive'
-            : 'array_replace_recursive';
-
-        $cacheKey = ($flatten ? '1_' : '0_') . $currentTheme . '_' . $key;
+        $cacheKey = $currentTheme . '_' . $key;
 
         if ($this->cache !== null) {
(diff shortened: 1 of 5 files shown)


demiankatz:
@crhallberg, it looks like this PR is using some PHP 8 syntax, but we still need to support 7.4 for at least a little longer (union types aren't available in 7.x). I'll take a deeper look later, but that's the reason for the CI failure here.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -210,35 +210,81 @@ public function getThemeInfo()
...
+     * @param mixed $op Variable to test
+     *
+     * @return boolean
+     */
+    protected function is_string_keyed_array($op) {
+        if (!is_array($op)) {
+            return false;
+        }
+
+        foreach ($op as $key => $val) {
```
demiankatz: This might be more efficient (or at least concise):

```
reset($op);
if (is_string(key($op)) {
    return true;
}
```

...though of course there are various edge cases that might cause unexpected behavior regardless of how you implement this.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -210,35 +210,81 @@ public function getThemeInfo()
...
     }
 
+    /**
+     * Determine if a variable is a string-keyed array
+     *
+     * @param mixed $op Variable to test
+     *
+     * @return boolean
+     */
+    protected function is_string_keyed_array($op) {
```
demiankatz: I don't think we want to use snake case for a method name.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -210,35 +210,81 @@ public function getThemeInfo()
...
-     * @param string $key     Configuration key to retrieve (or empty string to
+     * @param string $key Configuration key to retrieve (or empty string to
      * retrieve full configuration)
-     * @param bool   $flatten Use array_replace to flatten values
      *
-     * @return array
+     * @return array|string
      */
-    public function getMergedConfig(string $key = '', bool $flatten = false): array
+    public function getMergedConfig(string $key = ''): array|string
```
demiankatz: I think this is what's breaking the build; we just need to leave off the return type until the PHP requirement is raised to 8.

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
```diff
@@ -223,10 +223,10 @@ public function testGetMergedConfigParentOnly()
     {
         // Parent
         $ti = $this->getThemeInfo();
-        $parentJS = $ti->getMergedConfig('js');
+        $parentJS = $ti->getMergedConfig('js', false);
         $this->assertEquals(['hello.js'], $parentJS);
         // recursive
-        $parentHelpers = $ti->getMergedConfig('helpers');
+        $parentHelpers = $ti->getMergedConfig('helpers', false);
```
demiankatz: Why are you adding false here and in line 226? Hasn't the second parameter been eliminated?

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
```diff
@@ -244,29 +244,9 @@ public function testGetMergedConfigChild()
         $ti = $this->getThemeInfo();
         $ti->setTheme('child');
         $childJS = $ti->getMergedConfig('js');
-        $this->assertEquals(['hello.js', 'extra.js'], $childJS);
+        $this->assertEquals(['extra.js', 'hello.js'], $childJS);
```
demiankatz: Is it a problem that the order is reversing here? Don't we want child JS to load AFTER core JS (unless specifically configured otherwise)?

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
demiankatz: Also related to this line of code, but a separate issue: have you examined the ThemeConfig view helper? It also sets $flatten to true, and it does not appear to be changed in this PR. At very least, the obsolete parameter should be removed... but I imagine we also need to analyze how/where it is used and make sure these changes don't break any of that.

demiankatz: (Fortunately, at a glance, it looks like use of the helper is confined to the example theme... but we should definitely make sure that still works as expected!)

demiankatz (changes requested):
@crhallberg, I didn't go all the way down in analyzing the logic here, but I left a few quick comments that might lead to some revisions. I'll wait for you to respond to these before delving in deeper.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
crhallberg: Probably don't have to worry about edge cases for the theme configs. ~~What does reset do in this case?~~

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
crhallberg: I was trying to match `is_array` and the like, but I will change this.

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
crhallberg: Not sure how these slipped through, I grepped for usage. Fixing.

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
crhallberg: It could easily be either way ([just need to edit this line](https://github.com/vufind-org/vufind/pull/2543/files#diff-487ea02c96725131d0a8acbd87fd1c59bb49d0793f966115fa95541d0450ca6dR258)). Child first made more sense to me but loading order makes sense too. Which do you prefer?

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
demiankatz: (I'm not sure that the reset is really necessary, but I put it in on the principle of "better safe than sorry").

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
demiankatz: I think loading order will be less disruptive, since that's how the current behavior works.

demiankatz:
@crhallberg, looks like some minor style issues are still causing a build failure, but `vendor/bin/phing phpcbf` should take care of them automatically.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -210,35 +210,80 @@ public function getThemeInfo()
...
+     *
+     * @return boolean
+     */
+    protected function isStringKeyedArray($op) {
+        if (!is_array($op)) {
+            return false;
+        }
+
+        reset($op);
+        if (is_string(key($op))) {
```
demiankatz: Further optimization: just `return is_string(key($op));` instead of an if and two different returns.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -210,35 +210,80 @@ public function getThemeInfo()
...
+
+    /**
+     * Add parent theme information to a collection of merged theme info.
+     * Using the string-keyed array format of theme config info,
+     * recursively walk the array, capturing unique or missing values.
+     *
+     * @param array $merged Merged theme info, overrides parent theme
+     * @param array $current Theme info to be merged
+     *
+     * @return array merged theme info, favoring child themes (merged)
```
demiankatz: Are these types really correct? Isn't it possible to be array or string in some situations? Are there some edge cases that might be problematic, like if a parent theme contains a string and a child theme contains an array (which might lead to `foreach` being called on a string)?

Sorry if I'm misreading the context in which this is used, but it seems like we might want to tighten this up a little more (and maybe create a weird test case to exercise the potentially problematic edges).

demiankatz (changes requested):
@crhallberg, I think this is looking pretty good, but I have a couple more suggestions to (hopefully) make it more efficient and bullet-proof.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -210,35 +210,76 @@ public function getThemeInfo()
...
+     *
+     * @return array|string merged theme info, favoring child themes (merged)
+     */
+    protected function mergeWithoutOverride($merged, $current) {
+        if (empty($merged)) {
+            return $current;
+        }
+
+        if ($this->isStringKeyedArray($merged)) {
+            foreach ($current as $key => $val) {
```
demiankatz: I think we need to define a behavior for what happens when $merged is a string-keyed array and $current is a string... because otherwise we'll make an illegal foreach call and get a nasty error of some sort. Probably worth a test case!

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -210,35 +210,76 @@ public function getThemeInfo()
...
+                    $merged[$key] = array_merge($val, $merged[$key]);
+                }
+            }
+
+            return $merged;
+        }
+
+        if (is_array($merged)) {
+            // capture unique or missing array items
+            return array_merge($current, $merged);
```
demiankatz: Similarly, should we cast $current to an array here, just in case?

demiankatz (changes requested):
Just a couple more thoughts on types...

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
demiankatz: Thanks for the progress on this, @crhallberg, but is it really the case that when either $merged or $current is a non-array, we always want to return $merged? That does seem consistent with the method name of "mergeWithoutOverride," but don't we *want* to override string values in sub-themes? Or is the idea that we're applying these in reverse order from child to parent? If that's the case, maybe it is okay -- but I think this further reinforces my suggestion that we should be sure to have a test case that demonstrates this scenario, to be 100% clear about what we expect to happen.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
```diff
@@ -252,22 +296,19 @@ public function getMergedConfig(string $key = '', bool $flatten = false): array
 
         while (!empty($currentTheme)) {
             $currentThemeSet = array_merge(
-                (array)$currentTheme,
                 $allThemeInfo[$currentTheme]['mixins'] ?? [],
+                (array)$currentTheme,
```
demiankatz: Related to my comment above, why is the order of mixin vs. theme changing here? Do we expect mixins to override the theme they are mixed into, or do we expect the current theme to override its own mixins? I'm not sure what's preferable (I can see arguments for both approaches), but we should have a test case for this if we don't already, just to be sure the behavior is well-defined and doesn't get accidentally changed in future.

crhallberg: I would expect mixins to override the theme they are included in (specific trumps general). I would also expect mixins to override earlier mixins, similar to CSS. I see now that my code doesn't reflect this expectation.

demiankatz: Thanks, @crhallberg, I agree! I guess if you want to override a setting from a mixin, you would have to create a local mixin to load subsequent to the mixin you are overriding.

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
```diff
@@ -339,15 +300,67 @@ public function testGetMergedConfigNoKey()
...
+     * Stress-test our merging algorithm
+     *
+     * @return void
+     */
+    public function testMergeWithoutOverrideEdgeCases()
+    {
+        $ti = $this->getThemeInfo();
+
+        // string
+        $merged = $ti->mergeWithoutOverride('original', 'override');
```
demiankatz: If you want to directly test mergeWithoutOverride without having to actually make the method public, you can `use \VuFindTest\Feature\ReflectionTrait;` at the top of the test class, and then use (for example) `$this->callMethod($ti, 'mergeWithoutOverride', ['original', 'override'])`.

demiankatz: Refactoring this part of the test to use a dataProvider might also make things a bit cleaner, since then you could have a series of named test cases in the data provider, and you'd only have to rely on the reflection call in one spot in the code.

module/VuFindTheme/src/VuFindTheme/ThemeInfo.php
demiankatz: To return to my original comment, it looks to me like it may still be the case that things will go wrong if `$merged` is a string-keyed array and `$current` is not an array. Do we lack a test for that, or am I just missing/misunderstanding something?

crhallberg: I believe these cases are covered here. https://github.com/vufind-org/vufind/pull/2543/files#diff-9dda0badc9ada5e653e0704a4e20ac71ebcb9fdc46c23c092442afa0d1909c43R364-R380

demiankatz: @crhallberg, aren't those cases covering merging string-keyed and non-string-keyed arrays together? I'm talking about the case of merging a string-keyed array with a plain string (not an array at all). In other words, this test case:

```
            [
                [
                    ['mixed' => ['array']],
                    'not an array',
                ],
                // I'm not sure what we actually expect to happen in this case... just guessing!
                ['mixed' => ['array'], 0 => 'not an array'],
            ],
```

crhallberg: @demiankatz I don't expect this case to ever show up since we're merging themeConfigs, but I've added a handle where it just skips the attempted override.

module/VuFindTheme/tests/unit-tests/src/VuFindTest/ThemeInfoTest.php
```diff
@@ -378,6 +378,18 @@ public static function mergeEdgeCasesProvider(): array
...
+
+            // arrays and strings
+            [
+                [
+                    ['mixed' => ['array']],
+                    'not an array',
+                ],
+                [
+                    'mixed' => ['array'],
+                    // 0 => 'not an array'
```
demiankatz: Probably should remove this comment...

demiankatz (changes requested):
Thanks, @crhallberg, this makes a lot more sense now. Just one trivial comment, because I think you copied and pasted my suggestion and didn't clean it up after implementation. :-)

demiankatz (approved):
This looks good to me. I just corrected some minor style issues, and now I believe we're ready to merge. Thanks, @crhallberg!

PR #2544, opened 2022-09-21
Add status-available and status-unavailable to js-icons.

EreMaijala:
The icons are used in check_item_statuses.js.


demiankatz (approved):
Thanks!

PR #2545, opened 2022-09-21
FOLIO: Added check for patron login failure

damien-git:
See Jira [VUFIND-1578](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1578).


demiankatz (approved):
I have successfully reproduced the problem and confirmed the fix. Thanks, @damien-git!

PR #2546, opened 2022-09-22
Add more options to preferred library selection.

EreMaijala:
Makes it possible to reset back to default or make requests always ask for pick up location.


module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -1428,15 +1440,15 @@ public function getHoldDefaultRequiredDate($patron, $holdInfo)
      * placeHold, minus the patron data.  May be used to limit the pickup options
      * or may be ignored.
      *
-     * @return string A location ID
+     * @return false|string      The default pickup location for the patron or false
```
demiankatz, line 1443: We should probably keep wording in the return hint to indicate that it's a location ID we're looking for when a string is returned. Also, once we have standard return value wording, should we update the driver spec and copy it into all the other drivers that implement this method?

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AccountActionsTest.php
```diff
@@ -214,13 +218,127 @@ public function testChangeEmail()
...
+        // If emptying the selection is available, test it:
+        if (in_array('', $expectedPickUpChoices)) {
+            $this->findCss($page, '#home_library')->setValue('');
+            $this->clickCss($page, '#profile_form .btn');
+            $this->waitForPageLoad($page);
+            $this->assertEquals(
+                '',
+                $this->findCss($page, '#home_library')->getValue()
+            );
+            echo "CHECKED!\n";
```
demiankatz: I don't think you want this echo statement.

demiankatz (changes requested):
Thanks, @EreMaijala -- just a couple of comments. Also, a general question: you mention some limitations of the current database design; do you think these are worth attempting to overcome, or is this change adequate?

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: Ah, yes. I just copied it from another driver. Does "Default pickup location ID for the patron, or false for no default value" make sense for the wording?

demiankatz: Works for me! And I thought this was a new feature, but apparently you're just leveraging an existing one. Sorry for the confusion. :-)

EreMaijala: No problem! It's existing functionality supported by several drivers, and it looks like our specs have this case covered, so no need to fix there. I was a bit worried for a minute! :)

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AccountActionsTest.php
EreMaijala: Oops. Removed.

EreMaijala:
@demiankatz The issue with the database is that home_library cannot be null. Having null as "use system's default" would be convenient and would allow user to reset their choice (`home_library === ''`) to indicate that they always want to choose. Then it would be easy to display the home library selection accordingly and it would match the behavior when placing a request. I'd be happy to change the database, but I'm not sure how disruptive this would be. I'd suppose home_library is not typically used in many places other than the profile place and request dialog.

demiankatz:
@EreMaijala, I don't see much risk in removing the "NOT NULL" from the home_library definition; if we keep the default value the same, then it really shouldn't impact VuFind at all, except to add this new possibility. As you say, VuFind only uses this value in a limited number of places.

The question, then, is whether you want to get this PR wrapped up as-is and address that separately, or if you want to handle it here all in one step. I'm fine with either approach!

EreMaijala:
@demiankatz I think it would actually be nicer to do it all here since the current conditional approach is not quite optimal. Stay tuned!

EreMaijala:
@demiankatz The change is now done, but the database upgrade requires #2562. I kept the default value for home_library as empty string for compatibility, but it's not intuitive and causes some of the code to be a bit more complicated than if it was null by default meaning that the user hasn't set any preference.

EreMaijala:
Mink tests are passing for me and I've tested the pqsql scripts as well.

demiankatz:
Thanks, @EreMaijala! I'll do some hands-on testing of this once #2562 is completed. In the meantime, one question: is it worth putting a comment above the processing code in MyResearchController and/or in the changeHomeLibrary method to explain the historical reason for the meanings of `''` and `null`, just in case we need a reminder in the future?

EreMaijala:
@demiankatz Good idea, done. I also changed the order of if's in hold.phtml and applied the cxhanges to storageretrievalrequest.phtml that I had missed. illrequest.phtml is different, it never uses a default.

themes/bootstrap3/templates/record/hold.phtml
```diff
@@ -66,12 +66,18 @@
 
   <?php if (in_array("pickUpLocation", $this->extraHoldFields)): ?>
     <?php
+      // Empty string as home library means that default pick up location should be
+      // used. null means that there should be no default
```
demiankatz, line 70: While we're refactoring this anyway, it seems like it might be an opportunity to simplify the line below to:

```
if (($this->gatheredDetails['pickUpLocation'] ?? '') !== '') {
```

...and an equivalent change can also be made to the storage retrieval template.

demiankatz (commented):
One small suggestion. I don't think I'm going to have time to test this before tomorrow, so apologies if doing this hands-on would make the answer obvious... but I'm a little confused about how the profile page should work.

The idea here is that the user now has two different generic settings: null, meaning that the home library dialog box will be pre-populated to "no selection", and empty string, which means that the home library dialog box will be pre-populated to whatever the current default value happens to be. But as far as I can tell, there's no way to actually select both of those values from the profile page. It looks like you can either pick "null" or a specific value, but there's no way to choose the "pick the default" option. Is that intentional and/or a problem? Just wondering if this is going to cause confusion....

themes/bootstrap3/templates/record/hold.phtml
EreMaijala: Good idea, now done!

EreMaijala:
@demiankatz You're right, there's no setting that allows you to select "system default". It could be added, but I can't see it as a useful option. It's good as the default when the user hasn't chosen otherwise, but if the user is making a choice, "default" is really vague and probably not very useful. Perhaps middle ground solution would be to mark the default in the select list?

EreMaijala:
@demiankatz No, wait, I can see the point if getDefaultPickUpLocation returns dynamic information (in which case we can't display that in the select list). Perhaps that needs to be an option then.

demiankatz:
@EreMaijala, this is a tricky one -- making sure we offer the user all appropriate options without confusing them in the process. I suppose one way to word the menu could be something like "Always ask me" vs. "Default to the best available option" vs. "Always use X". But since not all of these options are meaningful in all possible contexts (e.g. because different drivers support different behaviors), that makes it trickier. Do we need config settings? Do we need drivers to report capabilities better? Is all of that too complicated and we should just rely on template customization where needed? I'm not sure what's best.

demiankatz:
(I should also note that, because my library only ever has one pickup location, I'm not heavily invested in this on a personal level ;-) ).

EreMaijala:
@demiankatz There's an additional terminology issue in that we call the selection "Home Library", though it really is just the default pick up location. But for "Home Library" something like "Always ask me" doesn't really work. If it was "Default Pick Up Library", then that would be perfect.

Now, adding the "No selection" option here at least doesn't make things worse than before, and it's always possible to add the "best available" option later. I also realized that we don't have a very good option value for "Default to the best available option" since anything but an empty string could in theory be a valid id.

demiankatz:
Good points. Maybe we should consider relabeling "Home Library" -- I don't know how attached anyone is to that particular wording. As for how to store "no selection" vs. "best available", I thought that was the distinction between null and empty string in the database... but is your point that we can't make that distinction in an HTML form? If so, that's true, though I imagine we could come up with some kind of token value that would be safe to assume would never be a real-world library identifier (e.g. "__vufind__use__default__").

demiankatz:
@EreMaijala, I just started testing this and see that the label in the Profile page is already "Preferred Library" rather than "Home Library," so I think that does open up flexible possibilities already.

demiankatz:
@EreMaijala, the Mink tests are passing locally for me as well, and the behavior looks reasonable during hands-on testing for the most part. However, I noticed one anomaly that may need to be addressed: if I have no default set, and I try to submit a hold without selecting a location, I get blocked and forced to fill in the field. However, if I do the same thing for a storage retrieval request, the request just goes through even though I didn't pick a location. Is that intentional? It seems like we should probably block submission of the form until a selection is made. (I tested this using the Demo driver with all default settings, in case that makes a difference).

EreMaijala:
> @EreMaijala, I just started testing this and see that the label in the Profile page is already "Preferred Library" rather than "Home Library," so I think that does open up flexible possibilities already.

Oops! How did I miss that? Well, that changes everything! I think I'll tweak the choices right away.

And yes, I meant the problem of encoding different defaults in the html select options. But I'll use `vufind__use__default` or something like that.

EreMaijala:
@demiankatz I believe the idea was that with storage retrieval requests the check is done in the ILS driver, so it's redundant at a higher level. KohaRest and VoyagerRestful (with wrong message) at least do that, but the Demo driver was missing it. Now added.

demiankatz (approved):
Thanks, @EreMaijala -- tests are passing and it looks good to me!

PR #2547, opened 2022-09-22
MultiBackend: Skip null values when adding or removing prefixes.

EreMaijala:
Fixes a PHP 8.1 deprecation notice and prevents mangling of null.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2549, opened 2022-09-23
Fix cookie consent mink test.

EreMaijala:
By default the dialog is hidden from bots, and this caused the dialog to be hidden from headless Chrome.

I used non-headless Chrome when testing to see what happens and failed to test headless properly. The cookie consent library is checking [navigator.webdriver](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/webdriver) in addition to user agent, and it's set when Chrome is controlled by automation, but apparently only in headless mode.


EreMaijala:
I'll merge this bypassing protections since this is a small change that only affects the Mink test.

PR #2550, opened 2022-09-23
Make sure to not set cookie domain for consent unless configured.

EreMaijala:
Setting it to any value overrides the default in the cookie consent component.

This is only an issue when host name isn't localhost, which prevented it from coming up in tests.


demiankatz (approved):
Makes sense. Thanks, @EreMaijala!

PR #2551, opened 2022-09-26
Fix a PHP 8.1 deprecation issue; getDisplayString expects a string.

EreMaijala:


demiankatz (approved):
Thanks for the fix, @EreMaijala!

PR #2553, opened 2022-09-26
Add config setting for display of items without barcodes.

demiankatz:
This was implemented in response to discussion #2548.

TODO:
- [x] Discuss name/placement of setting
- [x] Discuss whether/how to impact status-full.phtml
- [x] Add test coverage for new helper
- [x] Update changelog to note changed behavior when merging


demiankatz:
Based on discussion on #2548, I'm beginning to think that the best approach here is:

1.) Change the default behavior to allow empty barcodes (which will require a changelog note, but will also eliminate the inconsistency of behavior between holdings view and full status display).

2.) Add a note to the config file indicating that advanced filtering can be achieved by overriding/extending the holdings view helpers.

I'm definitely interested in other people's opinions on this!

demiankatz:
As suggested above, I've gone ahead and added expanded comments while flipping the default behavior. I still need to write a test for the new helper, which I will do once the basic approach is agreed upon.

EreMaijala (approved):
Looks good to me!

demiankatz:
Thanks @EreMaijala (and @damien-git). I've added a test and updated the changelog, so I think this is clear to merge now!

PR #2554, opened 2022-09-26
Make AbstractApi::makeRequest more generic/safe/flexible.

demiankatz:
This PR is a response to [VUFIND-1579](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1579). As of this writing, nothing has been tested yet -- this is meant to be a proof of concept for initial discussion.

TODO
- [x] Test thoroughly
- [x] Resolve [VUFIND-1579](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1579) when merging
- [x] Document BC breaks in changelog when merging: "The signature and behavior of \VuFind\ILS\Driver\AbstractAPI::makeRequest() has been changed for more consistent and reliable error handling; this impacts the FOLIO and GeniePlus ILS drivers; see pull request #2554 for details."


demiankatz:
@crhallberg, I think you may have been an original contributor to the AbstractAPI base class; I'm also interested in your thoughts on this (and the associated JIRA ticket). I think these changes reduce the number of assumptions built into the code and generally increase safety with the "throw a generic exception on any error, unless instructed otherwise" approach... but maybe you know some history that I'm not accounting for in the changes.

module/VuFind/src/VuFind/ILS/Driver/AbstractAPI.php
```diff
@@ -133,17 +153,21 @@ public function makeRequest(
...
-        case 403:
-            throw new Forbidden($response->getBody());
-        case 404:
-            throw new RecordMissing($response->getBody());
-        case 500:
-            throw new ILSException("500: Internal Server Error");
+        try {
+            $response = $client->send();
+        } catch (\Exception $e) {
+            $this->logError("Unexpected " . get_class($e) . ": " . $e->getMessage());
```
EreMaijala: I'd log `(string)$e` instead of `$e->getMessage()`. I've found it to be very useful to get a stack leading to the exception.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1031,7 +1025,9 @@ public function renewMyItems($renewDetails)
                 $response = $this->makeRequest(
                     'POST',
                     '/circulation/renew-by-id',
-                    json_encode($requestbody)
+                    json_encode($requestbody),
+                    [],
+                    '/.*/'
```
EreMaijala: Could we also allow something like `true` that would equal `/.*/` but be a bit cleaner?

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1283,18 +1281,21 @@ public function cancelHolds($cancelDetails)
...
-                json_encode($request_json)
-            );
-            if ($cancel_response->getStatusCode() == 204) {
+            try {
+                $cancel_response = $this->makeRequest(
+                    'PUT',
+                    '/circulation/requests/' . $requestId,
+                    json_encode($request_json)
+                );
+                if ($cancel_response->getStatusCode() !== 204) {
```
EreMaijala: I don't really like this. I think we shouldn't rely on exceptions for normal operation, as it seems here. I think it would be better to just allow any status code to be returned and either mark the result success or failure based on that.

EreMaijala (changes requested):

crhallberg (approved):
Approved once Ere's feedback is addressed.

demiankatz:
Thanks, @EreMaijala. I believe that I've addressed all of your review comments. Since the cancel holds method is significantly refactored, I also added test coverage and fixed a small problem (lack of default config value) while I was at it.

I still want to take a little more time to test everything else before merging, but please let me know if you think we're on the right track here!

EreMaijala (approved):
Looks good to me now. I didn't dig too deep in the new tests thoughs.

EreMaijala:
@demiankatz I also didn't test anything hands-on as I don't have access to Folio, but the code now seems way safer and cleaner than before.

demiankatz:
Thanks, @EreMaijala! I'll re-request review if anything substantially changes during the course of my testing; otherwise, I'll merge this once I'm satisfied that I haven't broken any of the impacted code. May take me a few days/weeks to find time to do all the necessary testing, though!

demiankatz:
I've ensured we have unit test coverage for all impacted code, plus I've done hands-on testing of both affected ILS drivers to ensure that they still work correctly. Merging now!

PR #2555, opened 2022-09-26
CSP Headers Update

crhallberg:
Updates to the default Content Security Policy (CSP) headers to allow for:

1. SVG images to use loaded and used.
2. Updates to Safari to support the more secure and recommended [strict-dynamic](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#strict-dynamic).

### DIFF
--- a/config/vufind/contentsecuritypolicy.ini
+++ b/config/vufind/contentsecuritypolicy.ini
@@ -25,7 +25,11 @@ use_nonce = true
 ; See also the VuFind wiki for additional recommendations and tools:
 ; https://vufind.org/wiki/administration:security:content_security_policy
 [Directives]
-default-src[] = "'none'"
+; default of 'self' with 'none' on child, object, prefetch allows SVG requests.
+default-src[] = "'self'"
+child-src[] = "'none'"
+object-src[] = "'none'"
+prefetch-src[] = "'none'"
 ; 'strict-dynamic' would allow any trusted script to load other scripts. However, it
 ; is currently not supported by Safari.
 ;script-src[] = "'strict-dynamic'"
--- a/config/vufind/contentsecuritypolicy.ini
+++ b/config/vufind/contentsecuritypolicy.ini
@@ -31,8 +31,10 @@ child-src[] = "'none'"
 object-src[] = "'none'"
 prefetch-src[] = "'none'"
 ; 'strict-dynamic' allows any trusted script to load other scripts with a hash.
-; Recent versions of Safari do not support this feature, so it is disabled by default.
-; https://caniuse.com/mdn-http_headers_content-security-policy_strict-dynamic
+;   Safari 15.3 and earlier does not support this feature. Since these browser
+;   versions constitute a significant portion of users, especially mobile users,
+;   'string-dynamic' is disabled by default.
+;   https://caniuse.com/mdn-http_headers_content-security-policy_strict-dynamic
 ;script-src[] = "'strict-dynamic'"
 ; backwards compatible to CSP 2
 script-src[] = "http:"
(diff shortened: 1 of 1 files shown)


demiankatz:
This looks good to me, but I'm interested in @EreMaijala's input, especially regarding the Safari issue.

EreMaijala (changes requested):
The problem is that while Safari 15.4 finally supports 'strict-dynamic', older versions are still widely used particularly on iOS devices as Safari version there is tied to the OS version.

While it's possible to add 'strict-dynamic' and depend on fallbacks for versions that don't support it, I'm not in favor of it for the following reasons:

1.) If 'strict-dynamic' is needed, it means that any fallbacks lose something. It can be functionality or security, but it's something.

2.) When a large subset of browsers uses different rules than others, making sure everything works across the devices becomes complicated, and you can't ensure correct functionality just by testing on desktop or with a device that has the latest iOS version.

demiankatz:
Makes sense, @EreMaijala. Do you think we should just roll back the strict-dynamic change but keep the other changes for SVG compatibility? Maybe it would be helpful to get @xmorave2's opinion too!

EreMaijala:
@demiankatz Yes, that would make sense.

crhallberg:
Just wanted to [double-check the statistics](https://caniuse.com/mdn-http_headers_content-security-policy_strict-dynamic), old versions of Safari on iOS account for 3% of web traffic. That might be high enough to pull it, since that would effect about 10% of mobile users.

EreMaijala:
@crhallberg That's just one stat, and even if it was accurate, 3% is still high enough to count. Maybe I'm a bit conservative here, but since the older iOS versions are still supported by Apple, I wouldn't want to disregard them. Also, percentages can be misleading. For us, in one single VuFind instance, we only had 2.2% in September by older iOS devices, but that still accounted for ~6700 unique visitors.

demiankatz:
I think it's probably okay to be conservative about compatibility right now and keep the Safari-back-compatible defaults... but this is the sort of item that might be worth mentioning on the "production setup checklist" page we're going to discuss at the next PMC meeting.

config/vufind/contentsecuritypolicy.ini
```diff
@@ -25,10 +25,16 @@ use_nonce = true
...
-default-src[] = "'none'"
-; 'strict-dynamic' would allow any trusted script to load other scripts. However, it
-; is currently not supported by Safari.
+; default of 'self' with 'none' on child, object, prefetch allows SVG requests.
+default-src[] = "'self'"
+child-src[] = "'none'"
+object-src[] = "'none'"
+prefetch-src[] = "'none'"
+; 'strict-dynamic' allows any trusted script to load other scripts with a hash.
+; Recent versions of Safari do not support this feature, so it is disabled by default.
```
EreMaijala: "Recent"?

EreMaijala (changes requested):

config/vufind/contentsecuritypolicy.ini
demiankatz: Do we have a specific version number we can use for cut-off, like "Versions before X (released in YEAR) do not..."

crhallberg: Agreed. I definitely tried too hard to be concise.

demiankatz (approved):

demiankatz:
Thanks, @crhallberg, this makes sense to me now. @EreMaijala, do you mind giving it one more look? Feel free to merge if you're happy with it.

EreMaijala (approved):
Fixed a typo, now all good.

xmorave2 (approved):
I do agree with the idea of maintaining wide compatibility by default.

EreMaijala:
Merged, thanks everyone!

PR #2556, opened 2022-09-27
Add support for merging sections of yaml config files.

EreMaijala:
Two commits: first one adds support for passing config file path locator callback to YamlReader for easier unit tests and future work, second one adds the merge support and tests.

All tests are including Mink tests are passing for me.

TODO:
- [x] Document the options in the wiki. Where? We only have a change log entry for the existing `@parent_yaml` directive. [Local Settings Directory](https://vufind.org/wiki/configuration:local_settings_directory) is at least related...

### DIFF
--- a/module/VuFind/src/VuFind/Config/Locator.php
+++ b/module/VuFind/src/VuFind/Config/Locator.php
@@ -39,6 +39,28 @@ namespace VuFind\Config;
 class Locator
 {
     /**
+     * Mode for getConfigPath: try to find a local file but fall back to base file
+     * if not available.
+     *
+     * @const int
+     */
+    public const MODE_AUTO = 0;
+
+    /**
+     * Mode for getConfigPath: try to find a local file.
+     *
+     * @const int
+     */
+    public const MODE_LOCAL = 1;
+
+    /**
+     * Mode for getConfigPath: get the base configuration file path.
+     *
+     * @const int
+     */
+    public const MODE_BASE = 2;
+
+    /**
      * Get the file path to the local configuration file (null if none found).
      *
      * @param string $filename config file name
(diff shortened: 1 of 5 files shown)


module/VuFind/src/VuFind/Config/Locator.php
```diff
@@ -83,21 +105,31 @@ public static function getBaseConfigPath($filename, $path = 'config/vufind')
     /**
      * Get the file path to a config file.
      *
-     * @param string $filename config file name
-     * @param string $path     path relative to VuFind base (optional; defaults
+     * @param string  $filename Config file name
+     * @param ?string $path     Path relative to VuFind base (optional; defaults
```
demiankatz: Maybe change the comment to say "use null for default" since that behavior has changed slightly?

module/VuFind/src/VuFind/Config/Locator.php
```diff
@@ -83,21 +105,31 @@ public static function getBaseConfigPath($filename, $path = 'config/vufind')
...
-        $local = static::getLocalConfigPath($filename, $path);
-        if (!empty($local)) {
-            return $local;
+    public static function getConfigPath(
+        $filename,
+        $path = null,
+        int $mode = self::MODE_AUTO
+    ) {
+        if (null === $path) {
+            $path = 'config/vufind';
```
demiankatz: Would it make sense to move the default path to a constant or property?

module/VuFind/src/VuFind/Config/YamlReader.php
```diff
@@ -161,10 +180,33 @@ protected function parseYaml($file, $defaultParent = null)
...
 
-        // Now load in missing sections from parent, if applicable:
+        // Now load in merged or missing sections from parent, if applicable:
         if (null !== $defaultParent) {
-            foreach ($this->parseYaml($defaultParent) as $section => $contents) {
+            $parentSections = $this->parseYaml($defaultParent);
+            // Process merged sections:
+            foreach ($mergedSections as $path) {
+                $resultElemRef = &$this->getArrayElemRefByPath($results, $path);
+                if (is_array($resultElemRef)) {
```
demiankatz: Do we really want to skip logic if the merged path does not exist in the child? We're only going to auto-merge missing top-level sections, so what if we want to merge in a more deeply-nested section from the parent that is not defined in the child? I might be misreading something, but it looks to me like that will not work. Does getArrayElemRefByPath perhaps need a flag to allow it to create missing nodes, and then if the path exists in the parent but not the child, we can build it into the child?

demiankatz (changes requested):
Thanks, this looks good, but I had a few questions/comments.

Also, regarding wiki documentation, I agree that it makes sense to link something from the "local settings directory" page. Maybe a new "configuration inheritance" page under the configuration section would make the most sense -- we could document the [Parent_Config] .ini capabilities along with the new YAML behavior. It looks like [Parent_Config] is mentioned in a few places but doesn't really have any higher-level documentation.

module/VuFind/src/VuFind/Config/Locator.php
EreMaijala: I added a const. Making it a property felt too scary!

module/VuFind/src/VuFind/Config/YamlReader.php
EreMaijala: You're right, it wasn't working properly for anything but top level. Should be fixed now.

EreMaijala:
@demiankatz I added [a wiki page](https://vufind.org/wiki/configuration:inheritance). It's a bit messy and probably missing something, but at least I tried! It's now also linked to from the local configuration page.

demiankatz (approved):
Looks great! Thanks, @EreMaijala.

PR #2558, opened 2022-09-27
Add a sensible default cookie consent revision update message.

EreMaijala:
This message is displayed when cookie consent revision has been updated.


demiankatz (approved):
Definitely better; thanks, @EreMaijala!

PR #2560, opened 2022-09-27
Add missing elements to MakeTag

crhallberg:
The spec I was using only listed elements that can be direct children of the body element, leaving many elements that are necessarily nested out. I found a new source and fixed the list, and added a deprecated element warning, since I had that info available.


demiankatz (approved):
Thanks, @crhallberg! I fixed a couple of minor typos, and I added "last checked" dates to help us decide when to review the spec in future if things start to get out of date.

PR #2561, opened 2022-09-27
Remove duplicate links

damien-git:
In MarcAdvancedTrait, links are extracted using a number of MARC subfields. There can be some redundancy in the data, especially with FOLIO where both instance records and holding records can have links. This change ensures the displayed links are unique (url, desc) pairs.

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/Record.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/Record.php
@@ -668,7 +668,7 @@ class Record extends \Laminas\View\Helper\AbstractHelper
             return $link;
         };
 
-        return array_unique(array_map($formatLink, $urls), SORT_REGULAR);
+        return array_values(array_unique(array_map($formatLink, $urls), SORT_REGULAR));
     }
 
     /**
--- a/module/VuFind/src/VuFind/View/Helper/Root/Record.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/Record.php
@@ -688,7 +688,7 @@ class Record extends \Laminas\View\Helper\AbstractHelper
      * Remove duplicates from the array. All keys and values are being used
      * recursively to compare, so if there are 2 links with the same url
      * but different desc, they will both be preserved.
-     * 
+     *
      * @param array $links array of associative arrays,
      * each containing 'desc' and 'url' keys
      *
--- a/module/VuFind/src/VuFind/View/Helper/Root/Record.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/Record.php
@@ -668,8 +668,7 @@ class Record extends \Laminas\View\Helper\AbstractHelper
             return $link;
         };
 
-        $arr = array_map($formatLink, $urls);
-        return array_values(array_unique($arr, SORT_REGULAR));
+        return $this->deduplicateLinks(array_map($formatLink, $urls));
     }
 
     /**
@@ -683,5 +682,20 @@ class Record extends \Laminas\View\Helper\AbstractHelper
     {
         return isset($this->config->OpenURL->replace_other_urls)
             && $this->config->OpenURL->replace_other_urls;
+    }
+
+    /**
+     * Remove duplicates from the array. All keys and values are being used
+     * recursively to compare, so if there are 2 links with the same url
+     * but different desc, they will both be preserved.
+     * 
+     * @param array $links array of associative arrays,
+     * each containing 'desc' and 'url' keys
+     *
+     * @return array
+     */
+    protected function deduplicateLinks($links)
+    {
+        return array_values(array_unique($links, SORT_REGULAR));
     }
 }
--- a/module/VuFind/src/VuFind/View/Helper/Root/Record.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/Record.php
@@ -668,7 +668,7 @@ class Record extends \Laminas\View\Helper\AbstractHelper
             return $link;
         };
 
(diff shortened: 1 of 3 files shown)


demiankatz:
Thanks, @damien-git -- this relates to an open JIRA ticket ([VUFIND-1277](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1277)) which contains some history you might find interesting.

I think this PR is a start, but I wonder if it's worth trying to address this more comprehensively while we're working on the problem anyway.

It would also be a good idea to add some test cases to exercise this change, if you have any sample records you'd like to share.

What do you think?

damien-git:
@demiankatz Interesting. I took a look at the ticket and the old PR. One notable difference is that my code is at a different place, and only applies to MARC records. The advantage is that I know for sure that the array will only have (url, desc) pairs and nothing else. The previous attempt was also meant to resolve issues with MARC records, so only dealing with this case seems reasonable to me, at least until we are aware of a different use case. Regarding the question of what to do when there are several urls with different descs, I think it's safer to display several independant links in that case. It might also be better in terms of layout, because otherwise it could create very long descriptions. I have not seen that case so far anyway.
I will look at adding tests.

demiankatz:
Thanks, @damien-git. As you note, the earlier PR did this in a different place: the Record helper's getLinkDetails method. There are MANY different implementations of getURLs in VuFind for a variety of different data sources, but the data always passes through getLinkDetails before getting rendered. That provides an opportunity to solve the problem in a single consistent way, rather than having varying implementations for each data type. I wonder if simply applying your deduplication at the end of that method, after all the data has been formatted and normalized, might have the same effect but apply it more broadly. As you say, it gets more complicated if we try to identify situations where different descriptions of the same URL get prioritized or merged, but even setting that scenario entirely aside, a straightforward deduplication seems like an obvious improvement.

damien-git:
@demiankatz It might work to put that in `getLinkDetails`. My main concern is that because there are many implementations of `getURLs`, some might return data that would break `array_unique($retVal, SORT_REGULAR)`.  Do you think it's safe ? Also you mentioned that in the other PR:
> There's another problem here: the URLs are currently being deduplicated before they are processed. While the most common type of link just has url/desc keys, some also use VuFind routes to generate URLs dynamically. We can't deduplicate the URLs before they have been generated, and in cases where they have been generated, there are going to be additional values in the array beyond 'url' and 'desc'. Thus, you could theoretically end up with duplicate url/desc combinations that would not be removed by the serialize/unserialize trick due to the presence of other keys.

So deduplication with array_unique might not work in all cases. I like that solution, though, because it's just a 1 line change and it's less likely to break (well, assuming we know what's in the array).

demiankatz:
@damien-git, the purpose of getLinkDetails is to take the potentially varied outputs from the various record drivers and reduce them down to a standard array format... so as long as that doesn't introduce any elements that make duplicate values distinct from each other (and I don't think it should), I think it's likely safe to apply the array_unique at the end. We can always look at a variety of record drivers and build diverse unit tests to reassure ourselves, if necessary. (I'm happy to help with that as needed). :-)

damien-git:
@demiankatz I moved it to getLinkDetails and added a test.

module/VuFind/src/VuFind/View/Helper/Root/Record.php
```diff
@@ -668,7 +668,8 @@ public function getLinkDetails($openUrlActive = false)
             return $link;
         };
 
-        return array_map($formatLink, $urls);
+        $arr = array_map($formatLink, $urls);
+        return array_values(array_unique($arr, SORT_REGULAR));
```
demiankatz: Maybe it's worth wrapping `array_values(array_unique($arr, SORT_REGULAR));` into a `protected function deduplicateUrls($urls)` definition (similar to the strategy of the earlier PR) so that if people want to apply a different algorithm, there's an easy extension point for them to utilize.

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/RecordTest.php
```diff
@@ -502,6 +502,35 @@ public function testGetLinkDetailsFailure()
...
+    public function testGetLinkDetailsWithDuplicateURLs()
+    {
+        $driver = new \VuFindTest\RecordDriver\TestHarness();
+        $driver->setRawData(
+            [
+                'URLs' => [
+                    ['desc' => 'link 1', 'url' => 'http://foo/baz1'],
+                    ['desc' => 'link 2', 'url' => 'http://foo/baz2'],
+                    ['desc' => 'link 1', 'url' => 'http://foo/baz1'],
+                    ['desc' => 'link 3', 'url' => 'http://foo/baz1']
```
demiankatz: Is it intentional that the 'url' value here is a duplicate? I think that's a useful thing to test, but I wonder if that's the case, it would be worth having a 'desc' like 'link 1 (alternate description)' to make the intention of the test case more clear. It might also be interesting to add a `['url' => 'http://foo/baz1']` entry in the array to exercise the code that injects a description when it is missing, just to further demonstrate that this works correctly in combination with pre-processing.

demiankatz (changes requested):
Thanks, @damien-git, this looks good to me; I just have a couple of suggestions that might make it even better.

damien-git:
@demiankatz I added your suggested changes.

demiankatz (approved):
Thanks, @damien-git, this looks good to me! I'm just going to ask @EreMaijala for one more opinion before I merge it, in case there's anything we're both overlooking...

EreMaijala (approved):

PR #2562, opened 2022-09-28
DbUpgrade fixes

EreMaijala:
When preparing a database change for PR #2546, I stumbled on several issues that the independent commits here address:

- Nullability was not checked
- MySQL stores json as longtext with a check constraint, and we couldn't handle the discepancy.

### DIFF
--- a/module/VuFind/src/VuFind/Controller/Plugin/DbUpgrade.php
+++ b/module/VuFind/src/VuFind/Controller/Plugin/DbUpgrade.php
@@ -775,6 +775,22 @@ class DbUpgrade extends AbstractPlugin
     }
 
     /**
+     * Given a current row default, return true if the current nullability matches
+     * the one found in the SQL provided as the $sql parameter. Return false if there
+     * is a mismatch that will require table structure updates.
+     *
+     * @param bool   $currentNullable Current nullability
+     * @param string $sql             SQL to compare against
+     *
+     * @return bool
+     */
+    protected function nullableMatches(bool $currentNullable, string $sql): bool
+    {
+        $expectedNullable = stripos($sql, 'NOT NULL') ? false : true;
+        return $expectedNullable === $currentNullable;
+    }
+
+    /**
      * Given a table column object, return true if the object's type matches the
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/Controller/Plugin/DbUpgrade.php
```diff
@@ -185,17 +185,33 @@ protected function getCharsetAndCollationProblemsForTableColumns(
...
-        foreach ($results as $current) {
-            if (!empty($current->Collation)
-                && strtolower($current->Collation) !== strtolower($collation)
-            ) {
-                $retVal[$current->Field] = (array)$current;
+        foreach ($results as $i => $current) {
+            if (!empty($current->Collation)) {
+                $normalizedCollation = strtolower($current->Collation);
+                if ($normalizedCollation !== $collation
+                    && $expectedTypes[$i] !== 'json'
```
demiankatz: I'm not sure I understand why we completely ignore the collation check for json. Shouldn't it get the correct collation for the table, regardless of other details?

If we really do never want to touch collation on json fields, I wonder if we should move the expectedTypes check earlier, since I don't think there's any reason to normalize and compare collations if we're going to skip them anyway.

demiankatz (commented):
Thanks, @EreMaijala. I've gone ahead and cherry-picked the first two commits from here into dev, but I had one question about the JSON piece, so I've left that commit unmerged until we can discuss further.

module/VuFind/src/VuFind/Controller/Plugin/DbUpgrade.php
EreMaijala: No, collation is always utf8mb4_bin for json fields unless something else is specified for the column, regardless of the table collation. I don't think there's a reason to change that, but this can be revisited if necessary.

I moved the check earlier and changed to an associative array to ensure that any discrepancy in column order or extra columns don't break the check.

demiankatz (approved):
Thanks, @EreMaijala!

PR #2563, opened 2022-09-30
Rename cookie consent events and add event for initial consent.

EreMaijala:
I had missed the recommendation in common.js of prefixing events with `vf-`. Also added an event for the initial consent given by user.

I already updated the [wiki](https://vufind.org/wiki/configuration:cookie_consent#javascript_events) in anticipation of this change.


demiankatz (approved):
Thanks; makes sense to me!

PR #2564, opened 2022-09-30
Make JSON DB upgrade compatible with MySQL + MariaDB.

demiankatz:


EreMaijala (approved):
Thanks! Verified to work with MariaDB.

PR #2565, opened 2022-09-30
Add tests for RecordTab classes.

skellamp:
Tests for RecordTab Classes.


demiankatz:
Thanks, @skellamp, this is a great start! If you look at the files here on GitHub, you'll see that the continuous integration is highlighting some issues in need of correction. Fixing those might be a good start toward getting the build to a green checkmark (you can also click on the details of the failed build to see what other issues, if any, are coming up). Once the build is passing, let me know and I'll give it a closer review... and then hopefully we can merge it!

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/CollectionHierarchyTreeTest.php
```diff
@@ -0,0 +1,72 @@
...
+ * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
+ */
+class CollectionHierarchyTreeTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * Test getting Description.
+     *
+     * @return void
+     */
+    public function testGetActiveRecord(): void
```
demiankatz, line 48: To get full coverage here, it might also be worth adding a second test like:

```
    /**
     * Test that getActiveRecord returns false when no ID parameter is provided.
     *
     * @return void
     */
    public function testGetActiveRecordWithEmptyId(): void
    {
        $conf = $this->getMockBuilder(\Laminas\Config\Config::class)
            ->getMock();
        $load = $this->getMockBuilder(\VuFind\Record\Loader::class)
            ->getMock();
        $request = $this->getMockBuilder(\Laminas\Http\Request::class)
            ->getMock();
        $request->expects($this->once())->method('getQuery')
            ->with($this->equalTo('recordID'), $this->equalTo(false))
            ->will($this->returnValue(false));
        $load->expects($this->never())->method('load');
        $obj = new CollectionHierarchyTree($conf, $load);
        $obj->setRequest($request);
        $this->assertFalse($obj->getActiveRecord());
    }
```

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/CollectionHierarchyTreeTest.php
```diff
@@ -0,0 +1,72 @@
...
+ * @category VuFind
+ * @package  Tests
+ * @author   Sudharma Kellampalli <skellamp@villanova.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
+ */
+class CollectionHierarchyTreeTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * Test getting Description.
```
demiankatz: I think this test description was pasted from elsewhere; it should be updated to be more accurate (like "Test that getActiveRecord loads the correct record ID.")

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/ComponentPartsTest.php
```diff
@@ -0,0 +1,133 @@
...
+        $obj = new ComponentParts($searchObj);
+        $this->assertSame(100, $obj->getMaxResults());
+    }
+
+    /**
+     * Test if the tab is active.
+     *
+     * @return void
+     */
+    public function testIsActive(): void
```
demiankatz: This might be a good test to add a `@dataProvider` to testIsActive() -- the data provider could provide the count of records reported by the record driver and the expected return value of isActive, and then you could test two cases: 0 / false and 10 / true. You'd end up with something like:

```
    /**
     * Data provider for testIsActive.
     *
     * @return array
     */
    public function isActiveProvider(): array
    {
        return ['no children' => [0, false], 'children' => [10, true]];
    }

    /**
     * Test if the tab is active.
     *
     * @param int  $childCount   Child count for record driver to report
     * @param bool $expectedResult Expected return value from isActive
     *
     * @return void
     *
     * @dataProvider isActiveProvider
     */
    public function testIsActive(int $childCount, bool $expectedResult): void
    {
        $searchObj = $this->getService();
        $obj = new ComponentParts($searchObj);
        $recordDriver = $this->getMockBuilder(\VuFind\RecordDriver\DefaultRecord::class)
            ->getMock();
        $recordDriver->expects($this->any())->method('tryMethod')
            ->with($this->equalTo('getChildRecordCount'))
            ->will($this->returnValue($childCount));
        $obj->setRecordDriver($recordDriver);
        $this->assertSame($expectedResult, $obj->isActive());
    }
```

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/ComponentPartsTest.php
```diff
@@ -0,0 +1,133 @@
...
+     */
+    public function testGetResults(): void
+    {
+        $searchObj = $this->getMockBuilder(\VuFindSearch\Service::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $rci = $this->getMockBuilder(
+            \VuFindSearch\Response\RecordCollectionInterface::class
+        )->disableOriginalConstructor()->getMock();
+        $searchObj->expects($this->any())->method('search')
```
demiankatz: I think you can probably make this more specific -- it should only be called once, shouldn't it? And I think you can replace the anythings with equalTo's -- e.g. second parameter should be `$this->equalTo(new \VuFindSearch\Query\Query('hierarchy_parent_id:"foo"'))`, etc., etc.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/FormatsTest.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Test if the tab is active.
+     *
+     * @return void
+     */
+    public function testisActive(): void
+    {
+        $obj=$this->getFormats();
```
demiankatz: For style reasons, I prefer to put spaces around equal signs.

Also, this method might be a good candidate for a `@dataProvider` -- you can test that isActive returns true when the Formats object is initialized with true, and false for false.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/HoldingsWorldCatTest.php
```diff
@@ -0,0 +1,119 @@
...
+    public function testIsActive(): void
+    {
+        $searchObj = $this->getService();
+        $obj = new HoldingsWorldCat($searchObj);
+        $recordDriver = $this->getMockBuilder(\VuFind\RecordDriver\DefaultRecord::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $recordDriver->expects($this->once())->method('tryMethod')
+            ->with($this->equalTo('getCleanOCLCNum'))
+            ->will($this->returnValue("foo"));
```
demiankatz: This would be another good candidate for a data provider -- so we can test the case where the record driver returns an OCLC number, and the case where it returns null. I realize that this won't work right as long as we retain the static caching in `HoldingsWorldCat::getOCLCNum`, but as we discussed yesterday, I would be in favor of removing that logic entirely to simplify the code. I see no real benefit to caching that value, and its potential for causing confusing behavior was proven yesterday. :-)

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/HoldingsWorldCatTest.php
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * @return void
+     */
+    public function testGetHoldings(): void
+    {
+        $searchObj = $this->getMockBuilder(\VuFindSearch\Service::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $obj = new HoldingsWorldCat($searchObj);
+        $recordDriver = $this->getMockBuilder(\VuFind\RecordDriver\SolrDefault::class)
```
demiankatz: I think we changed this from DefaultRecord to SolrDefault while troubleshooting yesterday, but in the end I don't think that was necessary, and DefaultRecord probably makes more sense. If you want to use a specific class, WorldCat would be the best choice, since that's the class most likely to get passed here in normal operation.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/MapTest.php
```diff
@@ -0,0 +1,191 @@
...
+        $obj = new Map();
+        $this->assertFalse($obj->supportsAjax());
+    }
+
+    /**
+     * get a Map object
+     *
+     * @return Map
+     */
+    public function getMap()
```
demiankatz: Would it make sense to use getMap everywhere for consistency (instead of instantiating Map directly in the two tests above) or else to rename this method getConfiguredMap to better explain its purpose?

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/MapTest.php
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * Test getting Graticule.
+     *
+     * @return void
+     */
+    public function testGetMapGraticule(): void
+    {
+        $obj = $this->getMap();
+        $expected = false;
```
demiankatz: Graticule defaults to false. Maybe it would be interesting to have two tests for this method -- one that confirms the default value is false, and another that confirms that you can override it to true via configuration.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/MapTest.php
```diff
@@ -0,0 +1,191 @@
...
+     *
+     * @return void
+     */
+    public function testIsActive(): void
+    {
+        $obj = $this->getMap();
+        $recordDriver = $this->getMockBuilder(\VuFind\RecordDriver\SolrDefault::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $recordDriver->expects($this->any())->method('tryMethod')
```
demiankatz: I think you can change any to exactly(2) here -- might as well be as specific as possible when it's safe to do so... or you could refactor this to use a dataProvider if you prefer. (Entirely your choice!)

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/MapTest.php
```diff
@@ -0,0 +1,191 @@
...
+     * @return void
+     */
+    public function testGetDisplayCoords(): void
+    {
+        $obj = $this->getMap();
+        $coordinates=["00 00 56 56", "45 56 87 89"];
+        $recordDriver = $this->getMockBuilder(\VuFind\RecordDriver\SolrDefault::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $recordDriver->expects($this->any())->method('tryMethod')
```
demiankatz: I think you can change any to once here.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/MapTest.php
```diff
@@ -0,0 +1,191 @@
...
+     * @return void
+     */
+    public function testGetGeoLocationCoords(): void
+    {
+        $obj = $this->getMap();
+        $coordinates= ["ENVELOPE(25.8,43.9,5.0,4.6)"];
+        $recordDriver = $this->getMockBuilder(\VuFind\RecordDriver\SolrDefault::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $recordDriver->expects($this->any())->method('tryMethod')
```
demiankatz: Another safe any --> once change.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/PreviewTest.php
```diff
@@ -0,0 +1,105 @@
...
+ * @author   Sudharma Kellampalli <skellamp@villanova.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:testing:unit_tests Wiki
+ */
+namespace VuFindTest\RecordTab;
+
+use VuFind\RecordTab\Preview;
+
+/**
+ * Reviews Test Class
```
demiankatz: Typo in comment (doesn't match the one above).

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/PreviewTest.php
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @return void
+     */
+    public function testisActive(): void
+    {
+        $expected_false = false;
+        $expected_true = true;
+        $obj1=$this->getPreview(false);
+        $obj2=$this->getPreview(true);
+        $this->assertSame($expected_false, $obj1->isActive());
```
demiankatz: You could use assertFalse/assertTrue here to eliminate the need for your two $expected_* variables.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/PreviewTest.php
```diff
@@ -0,0 +1,105 @@
...
+        $this->assertSame($expected_true, $obj2->isActive());
+    }
+
+    /**
+     * Test if the tab is intially visible.
```
demiankatz: Typo: "intially"

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/PreviewTest.php
demiankatz: There are some more assertSames below that could also be simplified in the same way.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/TOCTest.php
```diff
@@ -0,0 +1,74 @@
...
+     * @return void
+     */
+    public function testIsActive(): void
+    {
+        $recordDriver = $this->getMockBuilder(\VuFind\RecordDriver\SolrDefault::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $recordDriver->expects($this->any())->method('tryMethod')
+            ->with($this->equalTo('getTOC'))
+            ->will($this->returnValue(true));
```
demiankatz: It would be more realistic to return a string rather than a bool here -- any non empty string should make the tab active.

Probably another test where a dataProvider could help, so we could test both empty and non-empty strings and the expected results.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/TOCTest.php
```diff
@@ -0,0 +1,74 @@
...
+        // considering the complexity to create mock objects
+        // for parent class, only one test case was tested
+        $obj=new TOC();
```
demiankatz: Style: missing spaces around assignment operator.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/UserCommentsTest.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * Test if the tab is active
+     *
+     * @return void
+     */
+    public function testIsActive(): void
+    {
+        $obj = new UserComments();
```
demiankatz: Would be good to use a dataProvider here to confirm that the $enabled constructor parameter controls the behavior of the isActive method.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/UserCommentsTest.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Test if Captcha is active
+     *
+     * @return void
+     */
+    public function testIsCaptchaActive(): void
+    {
+        $obj = new UserComments();
+        $this->assertFalse($obj->isCaptchaActive());
```
demiankatz: Similarly, a dataProvider to confirm the relationship between the second constructor parameter and isCaptchaActive would be valuable.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/VersionsTest.php
```diff
@@ -0,0 +1,130 @@
...
+        $obj= new Versions($config, $som);
+        $obj->setRecordDriver($recordDriver);
+        $translator = $this->getMockBuilder(\Laminas\I18n\Translator\TranslatorInterface::class)
+            ->getMock();
+        $translator->expects($this->any())->method('translate')
+            ->with($this->equalTo('other_versions_title'), $this->equalTo('default'))
+            ->will($this->returnValue("Count:%%count%%"));
+        $obj->setTranslator($translator);
+        $obj->getDescription();
+        $this->assertEquals("Count:5", $obj->getDescription());
```
demiankatz: It would probably make sense for this to be "Count:$count" rather than "Count:5" to tie things together more explicitly.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/VersionsTest.php
```diff
@@ -0,0 +1,130 @@
...
+     */
+    public function testGetDescription(): void
+    {
+        $count=5;
+        $som = $this->getMockPluginManager();
+        $config = $this->getMockConfig();
+        $recordDriver = $this->getMockBuilder(\VuFind\RecordDriver\SolrDefault::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $recordDriver->expects($this->any())->method('tryMethod')
```
demiankatz, line 57: This any could probably safely be once.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/VersionsTest.php
```diff
@@ -0,0 +1,130 @@
...
+            ->with($this->equalTo('getOtherVersionCount'))
+            ->will($this->returnValue($count));
+        $obj= new Versions($config, $som);
+        $obj->setRecordDriver($recordDriver);
+        $translator = $this->getMockBuilder(\Laminas\I18n\Translator\TranslatorInterface::class)
+            ->getMock();
+        $translator->expects($this->any())->method('translate')
```
demiankatz, line 64: Can we change this any to either once or exactly(n) as appropriate?

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/VersionsTest.php
```diff
@@ -0,0 +1,130 @@
...
+            ->getMock();
+        $recordDriver->expects($this->any())->method('getSourceIdentifier')
+            ->will($this->returnValue('foo'));
+        $recordDriver->expects($this->any())->method('tryMethod')
+            ->with($this->equalTo('getOtherVersionCount'))
+            ->willReturnOnConsecutiveCalls(1, 0);
+        $obj= new Versions($config, $som);
+        $obj->setRecordDriver($recordDriver);
+        //Test when the tab is active
+        $this->assertTrue($obj->isActive());
```
demiankatz: If you rewrite this to use a dataProvider, I think you can tighten things up and simplify code a bit -- instead of using many $this->any assertions and the willReturnOnConsecutiveCalls logic, you could just use $this->once() for everything and return a mock value injected as a parameter by the data provider.

demiankatz (changes requested):
Thanks for all the work on this, @skellamp -- this looks very good! I left a few comments with suggested improvements, many of which are along similar lines. I think in a few places you can use PHPUnit's `@dataProvider` feature to test more thoroughly and/or make your code simpler and more readable.

Please let me know what you would like to work on after putting the finishing touches on this pull request! There is, of course, plenty more code in need of testing, but if you're interested in doing something more active in the code, I'm sure I can find something for you to do. If you are able to attend next week's VuFind developers call (9am on October 11) we will be discussing all outstanding work required for the 9.0 release, so you might find something to volunteer for there.

demiankatz (approved):
Thank you for the help improving our test coverage -- everything looks good to me!

PR #2566, opened 2022-10-03
Add Finnish translations for external authentication.

EreMaijala:
I had somehow missed adding these while adding Swedish ones.


demiankatz (approved):
Thanks!

PR #2567, opened 2022-10-03
OAuth2Controller: Add an overridable wrapper around config Locator.

EreMaijala:


demiankatz (approved):
Makes sense to me! In theory, it might make sense to make the constructor parameter optional and default to the locator in the controller if nothing is provided, but given that this is both newly-added functionality and also because the factory utilizes the same behavior, I don't think it's worth chaning. Merging as is. Thanks!

PR #2568, opened 2022-10-04
Refactor configuration file path resolution

EreMaijala:
Finished, still not yet tested. Test suite including Mink tests is passing.

TODO
- [x] Update changelog when merging (deprecation of \VuFind\Config\Locator; changes to internals of \VuFind\Config\PluginFactory and \VuFind\Config\YamlReader)
- [x] Refactor path stack to work in a sensible way with DEFAULT_CONFIG_PATH and the `$path` parameter.


module/VuFind/src/VuFind/Config/Locator.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * VF Configuration Locator
+ * VuFind Configuration Locator - A static compatibility wrapper around PathResolver
```
demiankatz, line 3: It's probably worth flagging all the methods in this class deprecated.

module/VuFind/src/VuFind/Config/PathResolver.php
```diff
@@ -0,0 +1,157 @@
...
+     * @return ?string
+     */
+    public function getConfigPath(
+        string $filename,
+        ?string $path = null,
+        int $mode = self::MODE_AUTO
+    ) {
+        if (self::MODE_BASE !== $mode) {
+            // Check if config exists in local dir:
+            $local = static::getLocalConfigPath(
```
demiankatz: Should we call internal methods non-statically now that this is not a static class?

module/VuFind/src/VuFind/Config/PathResolver.php
```diff
@@ -0,0 +1,157 @@
...
+     */
+    public function getLocalConfigPath(
+        string $filename,
+        ?string $path = null,
+        bool $force = false
+    ): ?string {
+        if (null === $path) {
+            $path = self::DEFAULT_CONFIG_PATH;
+        }
+        if (defined('LOCAL_OVERRIDE_DIR') && strlen(trim(LOCAL_OVERRIDE_DIR)) > 0) {
```
demiankatz: How would you feel about eliminating internal reliance on constants like LOCAL_OVERRIDE_DIR and APPLICATION_PATH in favor of local properties, so we can move the constant processing into a factory and pass relevant values via constructor calls?

module/VuFind/src/VuFind/Config/PluginFactory.php
```diff
@@ -65,17 +65,23 @@ public function __construct()
...
+     * file path (optional; defaults to Locator::getConfigPath)
      *
      * @return Config
      */
-    protected function loadConfigFile($filename, $path = 'config/vufind')
-    {
+    protected function loadConfigFile(
+        $filename,
+        $path = null,
+        callable $pathCallback = null
```
demiankatz: Would it make sense to make this non-optional, since it's an internal protected method, and we know we'll have access to the PathResolver? Then we'd have one less ongoing reference to the deprecated Locator...

module/VuFind/src/VuFind/Controller/Feature/ConfigPathTrait.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Get path to local configuration file
+     *
+     * @param string $filename Configuration file name
+     *
+     * @return string
+     */
+    protected function getLocalConfigFilePath(string $filename): string
```
demiankatz: is there a reason not to proxy the other parameters of the method?

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
```diff
@@ -126,15 +125,16 @@ protected function changeYamlConfigs($configs, $replace = [])
      */
     protected function changeConfigFile($configName, $settings, $replace = false)
     {
+        $resolver = new \VuFind\Config\PathResolver();
```
demiankatz: I wonder if it would be better to have a single instance of this accessed via a getter rather than building it on the fly every time -- especially if we eventually need a factory to build this with proper defaults.

module/VuFind/tests/bootstrap.php
```diff
@@ -31,8 +31,8 @@
 }
 
 // Make sure local config dir exists:
-if ($localDir = \VuFind\Config\Locator::getLocalConfigPath('', null, true)) {
-    if (!file_exists($localDir)) {
-        mkdir($localDir, 0777, true);
-    }
+$resolver = new \VuFind\Config\PathResolver();
```
demiankatz: Using this in the bootstrapper seems to pose some chicken and egg problems. I wonder if this might actually be an instance where simply using the environment variable directly would actually make more sense. (Maybe not, just brainstorming...)

module/VuFindAdmin/src/VuFindAdmin/Controller/ConfigController.php
```diff
@@ -47,7 +47,8 @@ public function homeAction()
     {
         $view = $this->createViewModel();
         $view->setTemplate('admin/config/home');
-        $view->baseConfigPath = \VuFind\Config\Locator::getBaseConfigPath('');
+        $resolver = $this->serviceLocator->get(\VuFind\Config\PathResolver::class);
```
demiankatz, line 50: Any reason not to use the ConfigPathTrait here?

module/VuFindTheme/src/VuFindTheme/ThemeGenerator.php
```diff
@@ -46,6 +47,25 @@ class ThemeGenerator extends AbstractThemeUtility implements GeneratorInterface
...
+     */
+    protected $pathResolver;
+
+    /**
+     * Constructor
+     *
+     * @param ThemeInfo    $info         Theme info object
+     * @param PathResolver $pathResolver Config file path resolver
+     */
+    public function __construct(ThemeInfo $info, PathResolver $pathResolver = null)
```
demiankatz, line 63: We need to adjust a factory to inject this, don't we?

demiankatz (changes requested):
Thanks, @EreMaijala, this is an excellent start. Thanks, in particular, for cleaning up those nasty tests that were writing config files into directories!

I commented as I read, so some of my early comments may not be informed by things I learned later. I think the bottom line here is that we need to decide whether we embrace keeping Locator as a quick and easy way to get paths, or if we want to try to thoroughly deprecate and eliminate it, to enforce better habits. I currently find myself leaning more toward deprecating it, since I don't think it's _too_ painful to get by without it, and it will break us of bad habits... but I could be persuaded otherwise.

Let me know if there's more I can do to help!

module/VuFind/src/VuFind/Config/Locator.php
EreMaijala: Done.

module/VuFind/src/VuFind/Config/PathResolver.php
EreMaijala: Yes, mistake.

module/VuFind/src/VuFind/Config/PathResolver.php
EreMaijala: Good idea, now done.

module/VuFind/src/VuFind/Config/PluginFactory.php
EreMaijala: I ended up changing it to just take the full filename, which seemed much more intuitive.

module/VuFind/src/VuFind/Controller/Feature/ConfigPathTrait.php
EreMaijala: Yes, these were meant to be simple wrappers to repeated calls with same parameters. I added a clarifying comment.

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
EreMaijala: Added.

module/VuFind/tests/bootstrap.php
EreMaijala: Agreed. I changed it to check LOCAL_OVERRIDE_DIR and verify that it exists.

module/VuFindAdmin/src/VuFindAdmin/Controller/ConfigController.php
EreMaijala: The trait is in another module so I'm not sure if that's ok. Also I just replicated the functionality in `enableautoconfigAction`, but the code might be wrong; if local config.ini doesn't exist, it would write to base config.ini.

module/VuFindTheme/src/VuFindTheme/ThemeGenerator.php
EreMaijala: Oops, added.

module/VuFind/src/VuFind/Config/Locator.php
```diff
@@ -38,75 +38,42 @@
...
+            ->getLocalConfigPath($filename, $path, $force);
     }
 
     /**
      * Get the file path to the base configuration file.
      *
      * @param string $filename config file name
-     * @param string $path     path relative to VuFind base (optional; defaults
-     * to config/vufind
+     * @param string $path     path relative to VuFind base (optional; use null for
```
demiankatz: This should be a ?string now (as should the one in getLocalConfigPath, though that's been wrong for a while).

module/VuFind/src/VuFind/Config/PathResolver.php
```diff
@@ -0,0 +1,147 @@
...
+    public function getLocalConfigPath(
+        string $filename,
+        ?string $path = null,
+        bool $force = false
+    ): ?string {
+        if (null === $path) {
+            $path = self::DEFAULT_CONFIG_PATH;
+        }
+        $fallbackResult = null;
+        foreach ($this->localConfigDirStack as $localDir) {
```
demiankatz: If localConfigDirStack is actually a stack, shouldn't we search it in reverse order to get the newest match off the top?

module/VuFind/src/VuFind/Controller/Feature/ConfigPathTrait.php
demiankatz: This still feels potentially confusing to me, since we're proxying a method using the same name but changing the default behavior in just one context. Could we tweak the method name for clarity? Maybe `getForcedLocalConfigFilePath`? Or even `getLocalConfigFullPath` which is not more specific but is at least distinct.

module/VuFind/src/VuFindTest/Feature/LiveDatabaseTrait.php
```diff
@@ -73,6 +73,12 @@ public function getLiveTableManager()
                 $config['vufind']['config_reader']
             );
             $container->set(\VuFind\Config\PluginManager::class, $configManager);
+            $localDirs = strlen(trim(LOCAL_OVERRIDE_DIR)) > 0
```
demiankatz: Would it be more concise/less redundant to do it like this?

```
$prFactory = new \VuFind\Config\PathResolverFactory();
$container->set(
    \VuFind\Config\PathResolver::class,
    $prFactory($container, \VuFind\Config\PathResolver::class)
);
```

module/VuFind/src/VuFindTest/Feature/LiveSolrTrait.php
```diff
@@ -67,6 +67,12 @@ protected function createLiveSolrContainer()
             $config['vufind']['config_reader']
         );
         $container->set(\VuFind\Config\PluginManager::class, $configManager);
+        $localDirs = strlen(trim(LOCAL_OVERRIDE_DIR)) > 0
```
demiankatz: If you change LiveDatabaseTrait, this should be made to match.

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
```diff
@@ -731,14 +737,21 @@ protected function logWarning(string $consoleMsg, string $logMsg = ''): void
...
         if (!$this->continuousIntegrationRunning()) {
             $this->markTestSkipped('Continuous integration not running.');
             return;
         }
 
         // Reset the modified configs list.
         $this->modifiedConfigs = [];
+
+        // Create a pathResolver:
+        $localDirs = strlen(trim(LOCAL_OVERRIDE_DIR)) > 0
```
demiankatz: ... and this, too, though I guess in this context you'd have to pass it a dummy container to satisfy the interface.

module/VuFind/tests/bootstrap.php
demiankatz: Great, seems safer and more robust!

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/OAuth2Test.php
```diff
@@ -73,11 +72,8 @@ public static function setUpBeforeClass(): void
      */
     public function setUp(): void
     {
-        // Give up if we're not running in CI:
-        if (!$this->continuousIntegrationRunning()) {
-            $this->markTestSkipped('Continuous integration not running.');
-            return;
-        }
+        parent::setUp();
```
demiankatz: Thanks for catching and fixing all this redundancy. I'm going to see if I can cherry-pick d09789ca4401bc85ac178349094c004f16986513 to dev so we can benefit from the cleanup immediately and reduce the number of changed files in this PR.

module/VuFind/tests/unit-tests/src/VuFindTest/CSV/ImporterTest.php
```diff
@@ -67,6 +68,10 @@ public function setUp(): void
     {
         $this->csvFixtureDir = $this->getFixtureDir() . 'csv/';
         $this->container = new MockContainer($this);
+        $this->container->set(
+            PathResolver::class,
+            new PathResolver(APPLICATION_PATH, [])
```
demiankatz: I suppose we could use the factory here for consistency with other tests, if we go down that road (though don't take all these related comments as insistence that we need to -- just want to document everything in case we decide that way).

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/RecordDataFormatterTest.php
```diff
@@ -151,6 +151,10 @@ protected function getFormatter()
             \VuFind\Config\PluginManager::class,
             new \VuFind\Config\PluginManager($container)
         );
+        $container->set(
+            \VuFind\Config\PathResolver::class,
+            new \VuFind\Config\PathResolver(APPLICATION_PATH, [])
```
demiankatz: Another one...

module/VuFind/tests/unit-tests/src/VuFindTest/XSLT/Import/VuFindTest.php
```diff
@@ -78,6 +78,10 @@ public function testGetChangeTracker()
     public function testGetConfig()
     {
         $container = $this->getMockContainer();
+        $container->set(
+            \VuFind\Config\PathResolver::class,
+            new \VuFind\Config\PathResolver(APPLICATION_PATH, [])
```
demiankatz: ...and another one.

module/VuFindConsole/tests/unit-tests/src/VuFindTest/Command/Util/SwitchDbHashCommandTest.php
```diff
@@ -285,4 +284,18 @@ public function testSuccessWithUser()
...
     }
+
+    /**
+     * Get path to local config.ini
+     *
+     * @return string
+     */
+    protected function getLocalConfigIniPath(): string
+    {
+        $localDirs = defined('LOCAL_OVERRIDE_DIR')
```
demiankatz: Maybe this would be better done in a setup method so we don't keep rebuilding the object over and over... and we could theoretically use the factory.

module/VuFindAdmin/src/VuFindAdmin/Controller/ConfigController.php
demiankatz: Good point. I view the VuFindAdmin module as having a dependency on the VuFind module, but I suppose it's better to keep things as disentangled as possible.

Good point also regarding enableautoconfigAction; probably would be better to call getLocalConfigPath and fail if it's null, to ensure we don't mess up the core config. Should we make a TODO about that and fix it after the rest of the PR, to keep things simpler here?

demiankatz (changes requested):
Thanks for the ongoing work, @EreMaijala -- this is really coming together nicely. I've done another review and left some more comments (many of which are redundant, and which you're welcome to ignore if they're unhelpful).

module/VuFind/src/VuFind/Config/Locator.php
EreMaijala: Yes, fixed.

module/VuFind/src/VuFind/Config/PathResolver.php
EreMaijala: Fixed and tests added.

module/VuFind/src/VuFindTest/Feature/LiveDatabaseTrait.php
EreMaijala: Method renamed and feature trait for test added to injecting PathResolver to a container.

EreMaijala:
@demiankatz I added a test feature trait for PathResolver injection to simplify the tests. Also other cleanup. Still need to think about the path stack and its problematic relation to DEFAULT_CONFIG_PATH.

module/VuFind/src/VuFindTest/Feature/PathResolverTrait.php
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Add PathResolverFactory to a mock container
+     *
+     * @param \VuFindTest\Container\MockContainer $container Mock Container
+     *
+     * @return void
+     */
+    protected function addPathResolverFactory(
```
demiankatz: This isn't really adding the factory -- it's just initializing the service using the factory. Perhaps this would be more accurately named `addPathResolverUsingFactory` or `addPathResolverToContainer`.

module/VuFind/src/VuFind/Config/PathResolver.php
demiankatz: I'm not seeing a PathResolverTest -- is there a new file that didn't get committed/pushed up?

demiankatz (changes requested):
Thanks for the latest round! A few more minor things...

module/VuFind/tests/unit-tests/src/VuFindTest/Config/PathResolverTest.php
```diff
@@ -0,0 +1,157 @@
...
+     * @return void
+     */
+    public function setUp(): void
+    {
+        $fixtureDir = $this->getStackedFixtureDir();
+        $this->stackedResolver = new PathResolver(
+            APPLICATION_PATH,
+            [
+                $fixtureDir . 'middle',
+                $fixtureDir . 'bottom',
```
demiankatz: I'm a little confused by this test, because the "bottom" of the stack based on the file name is actually what I would consider the "top" of the stack. Am I missing something?

EreMaijala: Maybe it's better not to call it a stack at all. :) I'll revise.

module/VuFind/src/VuFind/Config/PathResolver.php
EreMaijala: Forgot it, now added.

module/VuFind/src/VuFindTest/Feature/PathResolverTrait.php
EreMaijala: Fixed.

module/VuFind/tests/unit-tests/src/VuFindTest/Config/PathResolverTest.php
EreMaijala: Ended up renaming the fixtures.

module/VuFind/src/VuFindTest/Feature/PathResolverTrait.php
```diff
@@ -44,13 +44,27 @@ trait PathResolverTrait
...
+            ? [
+                [
+                    'directory' => LOCAL_OVERRIDE_DIR,
+                    'defaultConfigSubdir' => PathResolver::DEFAULT_CONFIG_SUBDIR
+                ]
+            ] : [];
+        return new \VuFind\Config\PathResolver(
+            [
+                'directory' => $baseDirectory ?? APPLICATION_PATH,
+                'defaultConfigSubdir' => 'config/vufind'
```
demiankatz: Any reason not to use `PathResolver::DEFAULT_CONFIG_SUBDIR` here as well?

module/VuFind/src/VuFind/Config/Locator.php
```diff
@@ -115,28 +82,37 @@ public static function getBaseConfigPath($filename, $path = 'config/vufind')
...
+            ? [
+                [
+                    'directory' => LOCAL_OVERRIDE_DIR,
+                    'defaultConfigSubdir' => PathResolver::DEFAULT_CONFIG_SUBDIR
+                ]
+            ] : [];
+        return new \VuFind\Config\PathResolver(
+            [
+                'directory' => APPLICATION_PATH,
+                'defaultConfigSubdir' => 'config/vufind'
```
demiankatz: Same question; should we use the constant here for consistency?

module/VuFind/src/VuFind/Config/PathResolver.php
```diff
@@ -0,0 +1,176 @@
...
+     * @var string
+     */
+    public const DEFAULT_CONFIG_SUBDIR = 'config/vufind';
+
+    /**
+     * Base directory
+     *
+     * Must contain the following keys:
+     *
+     * directory           - The local configuration directory
```
demiankatz: Should we use "base" instead of "local" in this comment?

module/VuFind/src/VuFind/Config/PathResolverFactory.php
```diff
@@ -0,0 +1,92 @@
...
+                ? [
+                    [
+                        'directory' => LOCAL_OVERRIDE_DIR,
+                        'defaultConfigSubdir' => $this->defaultConfigDirectory
+                    ]
+                ] : [];
+        return new $requestedName(
+            [
+                'directory' => APPLICATION_PATH,
+                'defaultConfigSubdir' => 'config/vufind'
```
demiankatz: Another hard-coded 'config/vufind'. I can see arguments for keeping this distinct instead of using $this->defaultConfigDirectory... but at the same time, I can't really think of a scenario where you would want to change how the local directory works independently of the core directory...

demiankatz (changes requested):
This looks good. I just have a couple of questions (or, more accurately, the same question three times) and one minor nitpick. Nearly there, I think!

EreMaijala:
@demiankatz All fixed, I hope!

demiankatz (approved):
Looks good, and all tests are passing!

PR #2569, opened 2022-10-05
Add new VuFind ObserverManager to replace hunt.

LuomaJuha:
Here is a little js script for creating intersectionobservers. Easier to add observable elements even after the observer has been created.

### DIFF
--- a/themes/bootstrap3/js/check_item_statuses.js
+++ b/themes/bootstrap3/js/check_item_statuses.js
@@ -223,12 +223,11 @@ VuFind.register('itemStatuses', function ItemStatuses() {
       checkItemStatuses(container);
       return;
     }
-    VuFind.observer.create(
-      'IntersectionObserver',
+    VuFind.observerManager.createIntersectionObserver(
       'itemStatuses',
       checkItemStatus
     );
-    VuFind.observer.observe(
+    VuFind.observerManager.observe(
       'itemStatuses',
       $(container).find('.ajaxItem').toArray()
     );
--- a/themes/bootstrap3/js/check_item_statuses.js
+++ b/themes/bootstrap3/js/check_item_statuses.js
@@ -1,4 +1,4 @@
-/*global Hunt, VuFind */
+/*global VuFind */
 
 VuFind.register('itemStatuses', function ItemStatuses() {
   function formatCallnumbers(callnumber, callnumber_handler) {
@@ -215,17 +215,23 @@ VuFind.register('itemStatuses', function ItemStatuses() {
     }
   }
   function init(_container) {
-    if (typeof Hunt === 'undefined' || VuFind.isPrinting()) {
-      checkItemStatuses(_container);
-    } else {
-      var container = typeof _container === 'undefined'
-        ? document.body
-        : _container;
-      new Hunt(
-        $(container).find('.ajaxItem').toArray(),
-        { enter: checkItemStatus }
-      );
+    var container = typeof _container === 'undefined'
+      ? document.body
+      : _container;
+
+    if (VuFind.isPrinting()) {
+      checkItemStatuses(container);
+      return;
     }
+    VuFind.observer.create(
+      'IntersectionObserver',
+      'itemStatuses',
+      checkItemStatus
+    );
+    VuFind.observer.observe(
+      'itemStatuses',
+      $(container).find('.ajaxItem').toArray()
+    );
   }
 
   return { init: init, check: checkItemStatuses, checkRecord: checkItemStatus };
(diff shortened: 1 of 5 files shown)


demiankatz:
Thanks, @LuomaJuha! I'm going to see if @crhallberg has any feedback on this. In the meantime, a couple of questions/comments:

1.) Hunt is used in quite a few places. Do you intend to replace all of them before this PR is merged? If so, maybe we should add a TODO checkbox to the top description as a reminder. (I understand that you might want to get this fully working and tested in one place before refactoring everything).

2.) @crhallberg has done some complementary work in #2447 (it doesn't replace Hunt, but it refactors the code that queues up work). We should make sure that these two pull requests will interact safely. If conflicts are likely, we should decide which one to prioritize first and try to get it fully reviewed and merged before finishing the other, to avoid creating extra work for ourselves.

LuomaJuha:
@Demiankatz Hey, at first, I thought about doing them in smaller pull requests. And it would seem like a good idea now that @crhallberg is doing the ajax rework pr.

Still thinking though that is the observer a good name for it, or should I switch it to ObserverController or something like that as it can control multiple different observers.

demiankatz:
@LuomaJuha, how about ObserverManager? Since "controller" has a fairly specific meaning in VuFind, "manager" might be more clear in this situation.

LuomaJuha:
@demiankatz Well that sounds like a better solution.

LuomaJuha:
@demiankatz @crhallberg I removed the check_save_statuses changes and left the check_item_statuses so there is not too much happening on the same time. Changed the name also.

demiankatz:
Thanks again, @LuomaJuha! I'll wait for input from @crhallberg before proceeding, but if Chris is too busy to review in the near future, I can try to do some testing of my own.

In the meantime, for reference, in case anyone is concerned, here's compatibility information for Intersection Observer: https://caniuse.com/intersectionobserver

crhallberg (approved):
I really like this solution! We mustn't forget to remove `js/vendor/hunt.min.js`.

themes/bootstrap3/js/check_item_statuses.js
```diff
@@ -215,17 +215,22 @@ VuFind.register('itemStatuses', function ItemStatuses() {
...
+      ? document.body
+      : _container;
+
+    if (VuFind.isPrinting()) {
+      checkItemStatuses(container);
+      return;
     }
+    VuFind.observerManager.createIntersectionObserver(
+      'itemStatuses',
+      checkItemStatus
```
demiankatz: Might it make sense to add an optional third "initial observation target" parameter to createIntersectionObserver so that everything can be set up in a single step? It might improve the readability of the code (while still retaining the separate method for adding additional targets later, of course).

demiankatz (commented):
Thanks, @LuomaJuha and @crhallberg: one very minor suggestion before I merge this (though if you disagree with the suggestion, I am happy to merge as-is).

I've run the full integration test suite, and everything is passing. This also works as expected when I do hands-on testing.

In response to @crhallberg's comment about removing the Hunt library, we can do that in future, but first we'll need to apply the ObserverManager everywhere that Hunt is used. I think the plan is to merge this PR first, then incorporate the remaining work into #2447. (Perhaps @crhallberg and @LuomaJuha can discuss how best to break that up).

themes/bootstrap3/js/check_item_statuses.js
LuomaJuha: Sounds great. Will do.

demiankatz (approved):
Looks great to me -- merging now! Thanks again, @LuomaJuha. Let's discuss next steps over on #2447.

demiankatz:
@LuomaJuha, I am extremely puzzled by this, but merging this PR caused a build failure -- the CombinedSearch integration test fails, because AJAX statuses do not display correctly when combined search results are loaded via AJAX. I have reviewed your refactoring several times and don't understand why these changes are causing things to break. However, for whatever reason, calling `observe` separately rather than taking advantage of the new `elements` parameter fixes the problem. I don't see how these are not functionally equivalent, but I must be overlooking something! For now, I have rolled back that part of the change in commit 16fdc65e0544d106e52a10a875ae8038a264b06f to get the build to pass again. Do you have any idea what's going on here?

LuomaJuha:
Well, thats odd. I have to look at it a littlebit more then!

LuomaJuha:
@demiankatz oh, the reason is that combined-search.js is calling the init function. Edit: I put a return check for createIntObs function so i think i could encapsulate the creation part so it would pass.

PR #2573, opened 2022-10-06
Validate MultiILS login target.

EreMaijala:
This doesn't show up unless you have a buggy template or you modify the login target option before sending it, but should be guarded against in any case. 

Note: not really a vulnerabiliity since you would still need to have valid credentials for the given target.

### DIFF
--- a/module/VuFind/src/VuFind/Auth/MultiILS.php
+++ b/module/VuFind/src/VuFind/Auth/MultiILS.php
@@ -65,6 +65,13 @@ class MultiILS extends ILS
         // We should have target either separately or already embedded into username
         if ($target) {
             $username = "$target.$username";
+        } else {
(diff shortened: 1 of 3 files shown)


demiankatz:
Thanks, @EreMaijala. It looks like we don't currently have a test class for this module. Is it feasible/worthwhile to establish a MultiILSTest to cover this new code? (Not expecting a full test class, but perhaps we could at least cover what has changed).

EreMaijala:
@demiankatz Makes sense. I'll work on some initial tests.

EreMaijala:
@demiankatz Needed quite hefty plumbing to get things working, but basic tests now implemented.

demiankatz (approved):
Thanks, @EreMaijala, you've gone above and beyond on the testing. :-)

PR #2574, opened 2022-10-11
Fix decoding of MARC ISO2709 record with multibyte subfield codes.

EreMaijala:


demiankatz (approved):
Makes sense. Thanks for the fix!

PR #2575, opened 2022-10-12
Check that there are pick up locations to choose from. 

EreMaijala:
When placing a hold it doesn't make sense to let the user to the hold form if we know that a pick up location is needed but there are none available. Otherwise the request will fail when the user tries to place with a confusing error message "An invalid pick up location was entered. Please try again" when the user never even saw the field.

### DIFF
--- a/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php
+++ b/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php
@@ -40,7 +40,7 @@ use Behat\Mink\Element\Element;
  * @author   Demian Katz <demian.katz@villanova.edu>
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org Main Page
- * @retry    2
+ * @retry    4
  */
 final class HoldsTest extends \VuFindTest\Integration\MinkTestCase
 {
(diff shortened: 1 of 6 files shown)


module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php
```diff
@@ -40,7 +40,7 @@
  * @author   Demian Katz <demian.katz@villanova.edu>
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org Main Page
- * @retry    4
+ * @retry    2
```
demiankatz: Did you intend to change the retry value here? I'm not opposed to reducing retries now that code is more stable/predictable, but I'm not sure why we'd make just one test class vary from the overall pattern.

demiankatz (changes requested):
This makes sense to me, but one small question about what looks like it may be an inadvertent change.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php
EreMaijala: No, just forgot to change it back (or mangle it so that it doesn't pass QA). Now reverted.

demiankatz (approved):
Thanks for the fix. Tests are passing for me; merging now.

PR #2576, opened 2022-10-12
Fix decoding of MARC ISO2709 record with multibyte indicator values.

EreMaijala:
I missed another way a record can be broken with the previous fix; an indicator can also contain a multibyte value.


module/VuFind/tests/fixtures/marc/multibyte_indicators.mrc
```diff
@@ -0,0 +1 @@
...
  anidebnc2rdacarrier
  aSisältää bibliografisia viitteitä (sivut 135-138) ja hakemiston.
  a"All academics need to write, and many struggle to finish their dissertations, articles, books, or grant proposals. Writing is hard work and can be difficult to wedge into a frenetic academic schedule. This revised and updated edition of Paul Silvia's popular guide provides practical, lighthearted advice to help academics overcome common barriers and become productive writers. Silvia's expert tips have been updated to apply to a wide variety of disciplines, and this edition has a new chapter devoted to grant and fellowship writing"--cJulkaisija.
  aEnglish languagexRhetoric.
  aAcademic writing.
12aWriting.
22aResearch.
 8iVerkkoaineisto:aSilvia, Paul  J., 1976- author.tHow to write a lotbSecond edition.dWashington, DC : American Psychological Association, [2019]z9781433829789
8 bminervacKTAiLaihTutkimuskirjallisuus82265015310006246
  bminervacKTAiLaioTutkimuskirjallisuus SIL
```
demiankatz: You have added a fixture but not a test; is something missing?

demiankatz (changes requested):
One specific question (see below) and also a general one: can we safely assume UTF-8 at this point? Are MARC records with other encodings like MARC-8 getting automatically converted at some other stage? If not, do we need to either add more encoding-awareness to the code or else include some kind of warning in the documentation about converting things prior to ingest?

module/VuFind/tests/fixtures/marc/multibyte_indicators.mrc
EreMaijala: Oops. Now added. I also realized that encoding wasn't working properly either, so that's now fixed as well.

EreMaijala:
@demiankatz As far as I can see Marc4j and thus SolrMarc convert MARC8 to UTF-8. We don't have any provision for anything else but UTF-8.

EreMaijala:
@demiankatz Fortunately we have `marc.to_utf_8 = true` in import properties files.

demiankatz (approved):

demiankatz:
Thanks, @EreMaijala, and glad to hear that the encoding issue seems settled -- just wanted to double-check that we aren't overly likely to run into problems with the 9.x transition, but it sounds promising!

PR #2577, opened 2022-10-12
Remove support for Apache 2.2

skellamp:
TODO
- [x] Close [VUFIND-1497](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1497) when merging


config/vufind/httpd-vufind.conf
```diff
@@ -1,38 +1,23 @@
 # Configuration for theme-specific resources:
 AliasMatch ^/vufind/themes/([0-9a-zA-Z-_]*)/(assets|css|images|js)/(.*)$ /usr/local/vufind/themes/$1/$2/$3
 <Directory ~ "^/usr/local/vufind/themes/([0-9a-zA-Z-_]*)/(assets|css|images|js)/">
-  <IfModule !mod_authz_core.c>
-    Order allow,deny
-    Allow from all
-  </IfModule>
   <IfModule mod_authz_core.c>
-    Require all granted
```
demiankatz, line 9: We need to keep `require all granted` but get rid of the `<IfModule>` and `</IfModule>` blocks above and below it. That is the Apache 2.4 syntax, so we want to keep it while eliminating the conditions for detecting Apache 2.4.

demiankatz (changes requested):
Thanks! One detail needs to be adjusted, however.

demiankatz (approved):
Thanks, @skellamp, this is working as expected!

PR #2580, opened 2022-10-13
Upgrade to Solr 9

demiankatz:
This is a PR for working on the upgrade to Solr 9. As of this writing it is work in progress.

TODO:
- [x] Figure out why Phing Solr download action is not working.
- [x] Get Solr to start up successfully
- [x] Test everything
  - [x] Basic startup
  - [x] SolrMarc indexing (under Linux)
  - [x] SolrMarc indexing (under Windows)
  - [x] Browse indexing/handling
  - [x] Full VuFind test suite
  - [x] In Windows
- [x] Investigate possible issues
  - [x] Slower indexing times?
  - [x] Errors on shutdown?
- [x] Update/recompile browse handler (see [vufind-browse-handler#40](https://github.com/vufind-org/vufind-browse-handler/pull/40))
- [x] Close relevant JIRA ticket(s) when merging


demiankatz:
With help from @marktriggs (see https://github.com/vufind-org/vufind-browse-handler/pull/40 -- thanks, Mark!) I was able to get everything working. The full test suite now passes using Solr 9, and I've also tested that all of the updated scripts (and alphabrowse functionality) work correctly under Windows as well as Linux.

I think this is now theoretically mergeable, though we still need to decide whether we want to target this to VuFind 9 or VuFind 10. There are a couple of outstanding issues that may require further investigation:

1.) Indexing with SolrMarc seems slower with Solr 9 than with Solr 8. This seems even more severe under Windows than under Linux (at least in my limited experience). The slowdown seems to occur at the "compiling Java" step. I suspect this may be related to the fact that we had to change the SolrJ directory in the import-marc scripts to reflect some restructuring of the Solr release. Previously, SolrJ had its own directory; now it is found in the general lib directory. I suspect that searching/linking a much larger number of .jar files may be impacting performance. I'm not sure if there is a simple solution to this problem.

2.) I have noticed that when I stop Solr (either with `./solr.sh stop` or via the `phing shutdown` action) it sometimes takes a very long time and then dumps out multiple error messages. I haven't seen the problem today, but it was occurring for me earlier in the week; perhaps it was related to the alphabrowse / Java security manager issues that were recently resolved, but I'd be interested to hear if it happens again for anyone.

@EreMaijala, do you have any other thoughts/concerns to add to this list? Maybe we should brainstorm possible solutions to the import speed problem first (perhaps starting by seeing whether we can point directly to the SolrJ jar instead of the whole lib directory), and then solicit further testing of this PR by more users.

EreMaijala:
@demiankatz The shutdown delay and errors typically indicate that Solr wasn't able to shut down cleanly, so it's always a bit worrying. I wonder if it could be related to alphabrowse or something. If it was only happening when there were issues with browse handler, perhaps it's no longer an issue?

The slowdown is probably really quite meaningless in production, but it makes startup for Mink tests slower, which is annoying. I'd still say that it's something we can live with.

EreMaijala (approved):
All changes look good to me.

demiankatz:
Thanks, @EreMaijala, it's very possible that the shutdown errors were related to the Java permission problems. I'll definitely test this a few more times as time permits and report if I see that issue again. Regarding the performance, I'll also try to do a bit of experimentation to see if my SolrJ path theory is correct. Maybe we can solve that problem in a simple way if it's just a matter of pointing to the right .jar file.

The bigger question at this point is whether we want to merge this for VuFind 9, or if we prefer to wait for VuFind 10. I'm a little bit inclined toward waiting, just on the principle of never relying on an x.0.0 release... but if Solr 9.0.1 comes out in the near future, maybe that argument will weaken. ;-)

demiankatz:
I've been doing more testing and haven't seen the shutdown errors again, so hopefully that was just down to the browse handler problems.

I've been doing some more testing related to timing, and it does seem that SolrMarc consistently takes about 10 seconds longer to run with Solr 9 than it did with Solr 8. The first run is always the slowest (because of compilation of custom Java), but subsequent runs are also slower. I'm pretty sure this has to do with time spent searching for SolrJ. I'll test that theory and look for possible workarounds when I have a little more time....

demiankatz:
Update: I have confirmed my theory that the slowdown is related to searching for SolrJ in the lib. If I copy solr*.jar and http*.jar from the Solr lib directory into a separate directory, then point SolrMarc's SolrJ setting at the copy, indexing time improves by 15-20 seconds.

So the next question is, is there some way to specify multiple specific jars with the SolrJ setting so we can point at just the ones we actually need, or if not, is it worth copying files or setting up symlinks in order to improve performance? It's a pretty substantial gain, so it may be worth the effort... but if so, we need to figure out how/where.

demiankatz:
I put together a quick proof of concept in https://github.com/vufind-org/vufind/pull/2580/commits/2b1cdc30d3971001a59299c7ef7d26db6ae7638f which greatly improves indexing performance. However, it has some obvious potential problems/limitations:

1.) I don't think there's an easy way to build equivalent logic for the Windows version (at least, I'm not interested in wasting time and effort on trying). We may just have to live with Windows being slower -- since it's not a good production environment anyway, I don't think this is likely to be a major issue.

2.) When upgrading Solr, these symlinks will go bad. We could add logic to the installsolr task to wipe them out. Or maybe we should put them under solr/vendor instead of solr/vufind so that they're part of the Solr installation... I just thought putting them in vufind expressed the fact that we're making them ourselves instead of pulling them out of the package. Another option might be to wipe out and recreate the directory every time we reindex, which is less efficient but probably not significantly time-consuming.

3.) It would be better to address this in the SolrMarc code, and add a smarter way to load SolrJ... but I think a workaround in VuFind is more likely to see fruition quickly.

Any thoughts, @EreMaijala?

EreMaijala:
@demiankatz How about relinking the files on every run, i.e. loop through the jars, remove any existing symlinks and add new ones? I suppose that wouldn't cause a perceptible slowdown and would ensure correct jars are available.

EreMaijala (changes requested):
See previous question.

demiankatz:
@EreMaijala, on further reflection, I think a more elegant solution is to put the SolrJ links under the vendor directory. This way, we only generate them once, and when Solr gets upgraded, the old links are automatically removed as part of the upgrade process.

I had originally resisted this idea because I didn't want to add anything to the vendor directory... but then I had the idea of putting a version number into the SolrJ symlink path to differentiate versions, and that reminded me that we put a hidden file in the vendor directory to track the installed version. Rather than reading that file and using it to create a longer path outside of vendor, it seemed easier to follow the precedent and just insert an additional hidden item into the vendor path.

What do you think?

EreMaijala:
@demiankatz Works for me!

EreMaijala (approved):
@demiankatz Everything now looks good to me.

demiankatz:
Thanks, @EreMaijala. We should also update the lucene match version in the Solr XML configuration, plus we should make sure the browse handler builds against the latest binaries and recompile everything. I'm happy to take care of those next week when I'm back in the office if you like!

EreMaijala:
@demiankatz I'd be happy to leave that to you. I just wanted to ensure everything still works with Solr 9.1.

demiankatz:
@EreMaijala, I've made the necessary updates here but ran into a snag with the classpath needing to be adjusted due to https://github.com/vufind-org/vufind-browse-handler/pull/39. I think it's sorted out, but I ran out of time before I could fully test everything. I'll do that as soon as time permits!

demiankatz:
@EreMaijala, I think I've gotten everything sorted out, and this now works when manually tested in Windows and passes the full test suite in Linux. I think it's ready to merge, but it's probably best to save this for discussion at the next Community Call, in case anyone has any concerns.

demiankatz:
@EreMaijala, we discussed this on today's Community Call, and nobody objected to merging it (though nobody was vocally in favor either). I thought I'd check with you one last time to see if you've already tested Solr 9 in production, or if you have any objections/concerns about merging. If you're comfortable moving forward, I will merge this once I hear from you! If you'd rather wait, we can reschedule this for 10.0.

EreMaijala:
@demiankatz I've on objections to merging now. We don't have any experience from production yet, but I still think that upgrading now makes sense.

PR #2581, opened 2022-10-13
Add username policy alongside the existing password policy.

EreMaijala:
TODO: 

- [x] Consider enabling a default policy.
- [x] Update changelog when merging (default username policy, deprecation of AbstractBase::getCannedPasswordPolicyHint, removal of Database::validateUsernameAndPassword)


module/VuFind/src/VuFind/Auth/AbstractBase.php
```diff
@@ -316,49 +316,73 @@ public function supportsConnectingLibraryCard()
...
+     * Return a canned username policy hint when available
      *
-     * @param string $pattern Current policy pattern
+     * @param string  $type    Policy type (password or username)
+     * @param ?string $pattern Current policy pattern
      *
-     * @return string
+     * @return ?string
      */
-    protected function getCannedPasswordPolicyHint($pattern)
```
demiankatz, line 325: Do you think there's any value in keeping this as a deprecated method for backward compatibility, or should we just go ahead and remove it? I don't mind removing it, but if we do, it probably deserves a changelog note, though I imagine local customizations depending on this are pretty rare/unlikely.

module/VuFind/src/VuFind/Auth/Database.php
```diff
@@ -197,19 +200,35 @@ public function updatePassword($request)
...
+     * Make sure username isn't blank and matches the policy.
      *
-     * @param array $params request parameters
+     * @param array $params      Request parameters
+     * @param bool  $checkPolicy Whether to check the policy as well (default is
+     * true)
      *
      * @return void
      */
-    protected function validateUsernameAndPassword($params)
```
demiankatz, line 207: Same question here... this might be slightly more likely to be getting called in local custom classes.

module/VuFind/src/VuFind/Auth/AbstractBase.php
```diff
@@ -378,56 +416,69 @@ public function getUserTable()
...
+    ): void {
+        $policy = $this->getPolicyConfig($type);
         if (isset($policy['minLength'])
-            && strlen($password) < $policy['minLength']
+            && mb_strlen($string, 'UTF-8') < $policy['minLength']
         ) {
             throw new AuthException(
                 $this->translate(
-                    'password_minimum_length',
+                    "{$type}_minimum_length",
```
demiankatz, line 482: I wonder if it's worth including a comment like `// e.g. password_minimum_length or username_minimum_length` so that if somebody searches the codebase for those translation strings, they get a hit and don't assume that they are orphaned.

demiankatz (commented):
I ran out of time before I could do a full review, but here are some initial comments...

config/vufind/config.ini
```diff
@@ -529,6 +529,19 @@ ils_encryption_key = false
...
 ;legal_domains[] = "mail.myuniversity.edu"
 
+; Specify default minimum and maximum new username length (Auth method may override
+; this).
+;minimum_username_length = 3
+;maximum_username_length = 32
+; Specify default limit of accepted characters in the username. Allowed values
+; are "numeric", "alphanumeric" or a regular expression
+;username_pattern = "(?=.*\d)(?=.*[a-z])(?=.*[A-Z])"
+;username_pattern = "([\p{L}\p{N}]+)"
```
demiankatz: Why do we have two examples here? If you intend to have multiple example patterns, we should probably also have multiple corresponding hints.

demiankatz (changes requested):
Thanks again, @EreMaijala -- one more comment.

Also, I tried hands-on testing of this, and when my username failed to match the regex pattern, I got an untranslated "username_error_invalid" message instead of the username_hint I configured. Any idea where that's coming from or why the translation is missing?

(I should note that I tested in a clean environment -- so no old language cache -- and simply uncommented the first username_pattern and username_hint lines in config.ini to test with -- so I didn't do anything particularly weird).

module/VuFind/src/VuFind/Auth/AbstractBase.php
EreMaijala: Yes, I added it back as a wrapper for `getCannedPolicyHint`.

module/VuFind/src/VuFind/Auth/Database.php
EreMaijala: This is unfortunately not as straightforward, as the split includes the important piece of ignoring username rules when changing password. Without it any previously added user whose username doesn't match current policy would be unable to change password.

module/VuFind/src/VuFind/Auth/AbstractBase.php
EreMaijala: That's a very valid point. I added comments about the translation strings as well as config settings.

config/vufind/config.ini
EreMaijala: I intended to leave only the second one in there. Now cleaned and explanation added.

EreMaijala:
@demiankatz The translation for `username_error_invalid` was missing. You wouldn't normally see it, but I had also forgotten to enable the length limits and pattern rule on the form after testing without it. So, now you'd get the message only if you managed to avoid the validation on the form.

module/VuFind/src/VuFind/Auth/Database.php
demiankatz: I'll add a changelog note.

demiankatz (changes requested):
@EreMaijala, I've gone ahead and updated the changelog on the assumption that this will be merged soon. However, before merging it, I think we need to revise the default username policy, as it currently blocks some characters that I think are quite commonly included in usernames -- e.g. underscores, periods, etc. Additionally, it makes it impossible to use an email address as a username, which is a common preference. Can we continue to block exotic Unicode characters while allowing basic ASCII non-control characters? Maybe that would be a little safer...

demiankatz:
In any case, all tests are passing, so I think this is ready to merge once we agree on what the default username pattern should be!

EreMaijala:
@demiankatz I added some characters to the pattern. Also tests with same pattern. This only leaves a minor issue in that it's not easy to describe clearly what's allowed. Displaying all the characters is messy, so I jsut called it punctuation even though I understand it's not just punctuation.

demiankatz (approved):
Looks good to me, and all tests are passing. Thanks!

PR #2582, opened 2022-10-13
Fix translation of Back link in account creation.

EreMaijala:


EreMaijala:
I have a feeling we've been here before, but I can't see anything supporting it in git blame.

demiankatz (approved):
Good catch! I'm going to backport this to release-8.1 as well since it's a bug fix.

PR #2583, opened 2022-10-13
Ignore consent cookie contents if revision doesn't match.

EreMaijala:
If revision doesn't match, the consent should be considered invalid.

The server side checks were missing a check for matching revision, which caused discrepancy between what the view helper says and what the user actually sees.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2586, opened 2022-10-14
LESS Compiler Deprecation

skellamp:
TODO:
- [x] Run full integration suite
- [x] Update changelog and LESS documentation in wiki when merging


demiankatz:
Thanks, @skellamp!

@crhallberg, I suggested removing the active LESS compilation along with the deprecated command line LESS compiler, since both of these depend on the same underlying PHP LESS compiler, which we know does not work consistently with the Javascript-based compiler. I think getting rid of this functionality will increase our flexibility for tooling in the future theme, and I doubt that anyone is using it in production. Do you have any concerns before I merge this?

demiankatz (changes requested):
@skellamp, I see that tests are failing here -- it looks like there are some tests that need to be removed, because the code that they cover has also been removed. If you look at the test output in the build, it should help guide you, but please let me know if you have any questions!

module/VuFindTheme/src/VuFindTheme/ResourceContainer.php
```diff
@@ -80,25 +80,6 @@ class ResourceContainer
...
     protected $generator = '';
 
-    /**
-     * Add a Less CSS file.
-     *
-     * @param array|string $less Less CSS file (or array of Less CSS files) to add
-     *
-     * @return void
-     */
-    public function addLessCss($less)
```
demiankatz, line 90: There is still one remaining reference to this method that needs to be removed from \VuFindTheme\Initializer.

demiankatz (changes requested):
Thanks again for this -- I think it's nearly done, but I found one remaining loose end to clean up.

demiankatz (approved):
I've run the full integration test suite, and everything is passing. I believe that this is ready to merge (and I'm happy to take care of the necessary documentation updates, once the decision to merge is made). I'm just going to wait for feedback from @crhallberg before taking those last steps.

crhallberg (approved):
I've always found the PHP LESS compiler convenient on systems that don't have Node installed, but I won't let my crutch continue to stand in the way of progress. This has been a security and deprecation concern for a while.

demiankatz:
Thanks, @skellamp and @crhallberg. All documentation is now up to date, so I am merging this!

PR #2587, opened 2022-10-17
Solr 8 deprecation cleanup

demiankatz:
This PR removes settings from solrconfig.xml that have been deprecated as of Solr 8.

HashDocSet has apparently been meaningless since Solr 1.4 more than a decade ago (see [SolrPerformanceFactors documentation](https://cwiki.apache.org/confluence/display/SOLR/SolrPerformanceFactors#SolrPerformanceFactors-MaxSizeConsiderations)), so I don't anticipate that removing it will have any impact.

The `<jmx />` setting has been replaced by `<metrics><reporter /></metrics>`, but I think it's probably best to omit this by default and let people set up reporting locally as needed.

I'm targeting this PR on the release-8.1 branch because this seems like an opportunity to eliminate warning messages in people's logs without hurting anything... but if anyone thinks that the jmx change is potentially "breaking" I can just re-target this to dev and we can wait a little longer to get it out in an official release.

TODO:
- [x] Run full test suite
- [x] Close [VUFIND-1571](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1571) when merging

### DIFF
--- a/solr/vufind/authority/conf/solrconfig.xml
+++ b/solr/vufind/authority/conf/solrconfig.xml
@@ -89,20 +89,6 @@
 
   </indexConfig>
 
-  <!--	Enables JMX if and only if an existing MBeanServer is found, use
-  		this if you want to configure JMX through JVM parameters. Remove
-  		this to disable exposing Solr configuration and statistics to JMX.
-
-		If you want to connect to a particular server, specify the agentId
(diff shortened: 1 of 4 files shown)


demiankatz:
Unsurprisingly, all tests are passing.

EreMaijala (approved):

PR #2589, opened 2022-10-17
Remove RecordLink View Helper

skellamp:


demiankatz (changes requested):
This looks good, but there's one more file to edit: themes/root/theme.config.php has some configurations related to the RecordLink helper that need to be removed.

demiankatz (approved):

PR #2591, opened 2022-10-19
Always pass a string to Bcrypt verify method.

EreMaijala:
Fixes a PHP 8.1 deprecation warning and wrong error message (authentication_error_technical instead of authentication_error_invalid) if user's password hash is null.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2592, opened 2022-10-19
Remove invalid attribute from cookie settings links.

EreMaijala:
I must have accidentally left the type attribute there when changing a button to a link.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2593, opened 2022-10-19
Upgrade vanilla-cookieconsent to 3.0.0-rc.7.

EreMaijala:
Most significant and relevant fix is that saving the preferences no longer breaks if categories are changed without increasing consent revision.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2594, opened 2022-10-20
End session if the user of the session cannot be found.

EreMaijala:
If we cannot find the user, we need to ensure that any session containers with anything related to the session state are cleared, and I believe the easiest and most appropriate action is to end the session. Otherwise logging out is not possible and there could be e.g. ChoiceAuth state information in the session making it impossible to log in again.


demiankatz (approved):
Thanks, @EreMaijala, this makes sense to me.

Since isLoggedIn gets called in a lot of places, I spent a little time investigating whether there was any possibility that logout could somehow get called after output had begun, which would cause a cookie-write error... but I think that's impossible for multiple reasons (I think the framework is naturally going to execute the code before rendering anything, plus the call to isLoggedIn in Bootstrapper::initUserLanguage is always going to get called early), and I can't think of any other reason to object to this.

PR #2595, opened 2022-10-20
[VUFIND-1552] Authority module: improve record routing

demiankatz:
This PR makes authority record URLs prettier. As it happens, there was already infrastructure in place to support clean authority record URLs, but it was not being fully utilized. This PR uses it for search results, eliminates the obsolete static route, and fixes a situation where an ID of "Record" would be impossible to retrieve. Because we share a controller between authority search and authority record display, it is currently impossible to fetch an authority record with an ID of "Search." It might be worth considering creating a separate AuthorityRecord controller to alleviate this situation (and make it less problematic to add new Authority actions in the future). However, given that this is using existing routes, I don't think I'm introducing any new problems here, so it's probably safe to address those concerns in the future on an as-needed basis.

Thoughts, anyone?

TODO
- [x] Resolve [VUFIND-1552](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1552) when merging


EreMaijala (changes requested):
This seems to break the old style URLs like `http://localhost/vufind2/Authority/Record?id=xyz`. I think it's important to keep support for it.

mtrojan-ub:
I don't think it would be critical to keep support for the old routes, but if @EreMaijala has a good reason for that of course i wouldn't mind either.

EreMaijala:
It'd be a problem if they're bookmarked, linked from web pages, harvested to search engines etc. and they'd suddenly break. Deprecating and removing somewhere in the future might be allright provided there's a redirect in place for the time between.

demiankatz:
@EreMaijala, it was not my intention to break the old URLs, and they still worked while I was testing this. I must have somehow broken it during one of my last changes. I'm honestly still not sure I understand why the routing is working the way it is. In any case, I was able to easily restore the legacy functionality with 5dd6773ee87c6c2e4e3fb42ce8ec20d9bc719377, so I think all is well now!

EreMaijala (changes requested):
I know the route `/Authority/<id>` is an existing one, but it's not a good one and I'm hesitating about promoting it. I think it should be `/AuthorityRecord/<id>`, similar to all other record routes, because otherwise there's a clash between `/Authority/<id>` and `/Authority/Search`. I suspect this may be the reason you're seeing the routing weirdness, and while it's possible to work around, I don't like the fact that we can't have an authority record with unique id `Search`.

EreMaijala:
@demiankatz Legacy route is working now, so that part is good.

demiankatz:
Thanks, @EreMaijala, I've done this in a more foolproof, standardized way, and I believe that both forms of legacy routing should continue to work as well. At some point we may want to consider dropping the authority/record.phtml template and leveraging more standard functionality, but for now I opted to keep the functionality the same so we can focus on routing changes alone in this PR.

module/VuFind/src/VuFind/Controller/AuthorityController.php
```diff
@@ -58,31 +58,20 @@ public function __construct(ServiceLocatorInterface $sm)
...
     {
-        // If we came in with a record ID, forward to the record action:
+        // If we came in with a record ID, forward to the record action; this
+        // provides backward compatibility with multiple legacy routes.
         if ($id = $this->params()->fromRoute('id', false)) {
-            $this->getRequest()->getQuery()->set('id', $id);
-            return $this->forwardTo('Authority', 'Record');
+            if ($id === 'Record') {
+                $queryId = $this->params()->fromQuery('id');
+                $id = $queryId === null ? $id : $queryId;
```
EreMaijala: These two lines can be simplified as `$id = $this->params()->fromQuery('id', $id);`, right?

EreMaijala (changes requested):

module/VuFind/src/VuFind/Controller/AuthorityController.php
demiankatz: No, I don't think so -- we have to process the route and query parameters differently because of the multiple paths that can lead to this point. Since there's been quite a bit of refactoring, maybe all of the scenarios are no longer likely/possible, but there at least used to be a scenario where going to the legacy record URL would cause the home route to get triggered with a record ID of "Record", so then it was necessary to fall back to the query parameter. In other scenarios, the route parameter IS the ID we want. That's what this code is intended to address (as expressed by the comment above). If the comment needs clarification, or if I'm misunderstanding your recommendation, I am of course still happy to make further adjustments!

EreMaijala: Sorry, I meant replacing just lines 65 and 66 as:

```
            if ($id === 'Record') {
                $id = $this->params()->fromQuery('id', $id);
            }

```

demiankatz: Ahh, sorry for misunderstanding; I believe you are correct and will fix it now...

EreMaijala (approved):

demiankatz:
Thanks, @EreMaijala!

PR #2596, opened 2022-10-20
[VUFIND-1556] Handle relator URIs during indexing

demiankatz:
TODO
- [x] Close [VUFIND-1556](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1556) when merging.


import/index_java/src/org/vufind/index/RelatorContainer.java
```diff
@@ -50,6 +58,24 @@ public Set<String> getKnownRelators()
         return knownRelators;
     }
 
+    public Set<String> getRelatorPrefixesToStrip()
+    {
+        // Populate set if empty:
+        if (relatorPrefixesToStrip.isEmpty()) {
```
mtrojan-ub: If somebody doesn't use these values and resets the config to an empty list, getConfigSection() will be called over and over again, which might produce lots of warnings. Woulnd't it be better to use null as default for the class variable and test against that?

demiankatz: Thanks, @mtrojan-ub, that's a very good point. This code was modeled on PunctuationContainer.java, which contains the same issue. I've opened #2597 to fix that, and closer examination of the code here made me realize that this could be done more simply, so I've further simplified the logic. I also removed a `System.out.println` debug message that I had not intended to commit.

import/index_java/src/org/vufind/index/CreatorTools.java
```diff
@@ -592,6 +593,12 @@ protected List<String> normalizeRelatorStringList(List<String> stringList)
      */
     protected String normalizeRelatorString(String string)
     {
+        for (String prefix : relatorPrefixesToStrip) {
```
EreMaijala, line 597: Would it make sense to do this last? Otherwise e.g. leading whitespace will cause the stripping to seemingly not work.

demiankatz: Good point. I've moved the trim above the prefix stripping. I think it may make more sense to keep the lowercasing and punctuation stripping later (since otherwise we would have to apply the same operations to the prefixes themselves, and that might result in unexpected behaviors).

EreMaijala: Makes sense!

EreMaijala (approved):

demiankatz:
Thanks, @EreMaijala! I'm going to leave this open just a little longer in case @mtrojan-ub or @lahmann have any further comments. (I believe that I addressed @mtrojan-ub's earlier feedback, but I'm not sure if he's reviewed again since then... and still curious if @lahmann has any suggestions/concerns about this in principle).

demiankatz:
It's been another week without feedback, so I'm going to go ahead and merge this. Of course, others are free to open follow-up PRs as needed if changes/improvements are found after the fact.

PR #2600, opened 2022-10-24
[VUFIND-1474] Remove deprecated wrapper methods for search backend calls

skellamp:
TODO
- [x] Fix deprecation warning: `PHP Deprecated:  Laminas\ServiceManager\AbstractPluginManager::__construct now expects a Psr\Container\ContainerInterface instance representing the parent container; please update your code in /usr/local/vufind/vendor/laminas/laminas-servicemanager/src/AbstractPluginManager.php on line 97`
- [x] Resolve [VUFIND-1474](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1474) when merging


skellamp (commented):
Some tests still need to be added. There is a var_dump in one of the test classes. Will fix those.

demiankatz:
Thanks, @skellamp, this looks like it's off to a good start! Please request a review from me once you're ready for me to take a closer look.

module/VuFind/src/VuFind/Record/Loader.php
```diff
@@ -207,9 +210,9 @@ public function loadBatchForSource(
         $genuineRecords = [];
         if ($list->hasUnchecked()) {
             try {
+                $command = new RetrieveBatchCommand($source, $list->getUnchecked(), $params);
```
demiankatz: The build failure is caused by this line being longer than the 86-character limit. Try changing it to:

```
                $command = new RetrieveBatchCommand(
                    $source,
                    $list->getUnchecked(),
                    $params
                );
```

demiankatz (changes requested):

demiankatz:
Thanks for the progress, @skellamp! I see that some tests are still failing -- please let me know if you need help with anything as you keep chipping away at this!

demiankatz (changes requested):
Thanks for the progress on this, @skellamp, it's looking great. As soon as time permits, I'll take a closer look and make sure nothing is missing. At a quick glance, though, I have one question: why did you entirely delete VuFindTest\SearchServiceTest? It looks to me like at least some of it is still going to be relevant (e.g. tests that cover the invoke method, like `testFailedResolveWithCommand`).

demiankatz (changes requested):
@skellamp, I ran the full integration test suite and encountered 2 errors, 33 failures and 18 skipped tests. Some of these are likely chain reactions. In any case, I think there are still some changes that haven't been fully completed yet.

One problem that seems to be causing at least some of these failures is that calls to `$this->searchService->workExpressions()` have not been refactored (see `VuFind\RecordDriver\Feature\VersionAwareTrait`).

I also see some calls to `->search` and `->retrieve` that haven't been updated yet. 

Additionally, it looks to me like the SearchServiceTest was providing coverage for some of the Command objects, and now that we have refactored code, we'll need to create new command tests to maintain test coverage. Fortunately, many commands are already covered, and most of the remaining commands should be fairly easy to cover by using SearchCommandTest as a model. Some are a little bit complex, though, like the RetrieveBatchCommand which checks to see if a backend supports a batch retrieval interface, and falls back to retreiving records one by one if batch support does not exist. In cases like this, the old SearchCommandTest methods can probably be used as models to build new command tests. If any of this is too confusing, let me know and I'm happy to help!

module/VuFind/tests/unit-tests/src/VuFindTest/Recommend/RandomRecommendTest.php
```diff
@@ -186,10 +203,26 @@ public function testCanInitialiseInDisregardMode()
         $params->setBasicSearch($query->getString(), $query->getHandler());
         $request = $this->createMock(\Laminas\Stdlib\Parameters::class);
 
-        $service->expects($this->once())->method('random')
-            ->with($this->equalTo("Solr"))
+        $commandObj = $this->getMockBuilder(\VuFindSearch\Command\AbstractBase::class)
+        ->disableOriginalConstructor()
```
demiankatz: Minor style issue -- this line and the one below should be indented one more level. Not sure why this didn't get caught by php-cs-fixer, but I think the tests are less rigorously style-checked than the actual code.

module/VuFind/tests/unit-tests/src/VuFindTest/Record/FallbackLoader/SolrTest.php
```diff
@@ -59,13 +60,22 @@ public function testEnabledLoader(): void
...
+        $commandObj = $this->getMockBuilder(\VuFindSearch\Command\AbstractBase::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $commandObj->expects($this->once())->method('getResult')
+            ->will($this->returnValue($collection));
+        $checkCommand = function ($command) {
+            return get_class($command) === \VuFindSearch\Command\SearchCommand::class
+                && $command->getTargetIdentifier() === "Solr"
+                && $command->getArguments()[0]->getString() ===
+                'previous_id_str_mv:"oldId"';
```
demiankatz, line 71: Can you compare the first argument directly to `$expectedQuery`, or if not, at least compare this to `$expectedQuery->getString()`? Otherwise, the `$expectedQuery` parameter does not appear to be getting used for anything.

module/VuFind/tests/unit-tests/src/VuFindTest/Record/FallbackLoader/SummonTest.php
```diff
@@ -64,13 +64,24 @@ public function testLoader(): void
...
+        $commandObj = $this->getMockBuilder(\VuFindSearch\Command\AbstractBase::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $commandObj->expects($this->once())->method('getResult')
+            ->will($this->returnValue($collection));
+        $checkCommand = function ($command) {
+            return get_class($command) === \VuFindSearch\Command\RetrieveCommand::class
+                && $command->getTargetIdentifier() === "Summon"
+                && $command->getArguments()[0] === 'bar'
+                && $command->getArguments()[1]->getArrayCopy() === ["summonIdType" => [2]];
```
demiankatz: Can you utilize `$expectedParams` here in some way so the variable is not unused?

module/VuFind/tests/unit-tests/src/VuFindTest/Record/LoaderTest.php
```diff
@@ -136,14 +183,24 @@ public function testSingleRecordWithBackendParameters()
...
+            ->getMock();
+        $commandObj->expects($this->once())->method('getResult')
             ->will($this->returnValue($collection));
+        $service = $this->getMockBuilder(\VuFindSearch\Service::class)
+            ->disableOriginalConstructor()->getMock();
+        $checkCommand = function ($command) {
+            return get_class($command) === \VuFindSearch\Command\RetrieveCommand::class
+                && $command->getTargetIdentifier() === "Solr"
+                && $command->getArguments()[0] === "test"
+                && $command->getArguments()[1]->getArrayCopy() === ['fq'=> ['id:test']];
```
demiankatz: Do we literally expect $params here? Can we do a direct comparison instead of pulling the array copy?

demiankatz: I also wonder if this test might be simplified (or at least made more readable) by adding a protected support method like:

```
protected function getCommandChecker($class, $args = [], $target = 'Solr')
{
    return function ($command) use ($class, $args, $target) {
        return get_class($command) === $class
            && $command->getArguments() === $args
            && $command->getTargetIdentifer() === $target;
    };
}
```

Then, instead of building custom functions repeatedly, you can just do things like `->with($this->callback($this->getCommandChecker(...)))`.

This is an entirely optional suggestion, and maybe the argument comparison won't work as well as I hope it will... but something you can consider if you like!

module/VuFind/tests/unit-tests/src/VuFindTest/Record/LoaderTest.php
```diff
@@ -170,19 +227,49 @@ public function testBatchLoad()
...
+            ->willReturnOnConsecutiveCalls($collection1, $collection2, $collection3);
+
+        $service = $this->getMockBuilder(\VuFindSearch\Service::class)
+            ->disableOriginalConstructor()->getMock();
+
+        $checkCommand1 = function ($command) {
+            return get_class($command) === \VuFindSearch\Command\RetrieveBatchCommand::class
+                && $command->getTargetIdentifier() === "Solr"
+                && $command->getArguments()[0] === ["test1", "test2"]
+                && $command->getArguments()[1]->getArrayCopy() === ['fq'=> ['id:test1']];
```
demiankatz: This is another place where you might be able to compare literally against the params object instead of using getArrayCopy, since I believe we're actually checking for a specific instance and not just for a generic set of parameters.

demiankatz: (See also the two other check commands below).

module/VuFind/tests/unit-tests/src/VuFindTest/Record/LoaderTest.php
```diff
@@ -216,12 +303,35 @@ public function testBatchLoadWithFallback()
...
+            ->willReturnOnConsecutiveCalls($collection1, $collection2);
+
+        $service = $this->getMockBuilder(\VuFindSearch\Service::class)
+            ->disableOriginalConstructor()->getMock();
+
+        $checkCommand1  = function ($command) {
+            return get_class($command) === \VuFindSearch\Command\RetrieveBatchCommand::class
+                && ($command->getTargetIdentifier() === "Solr")
+                && $command->getArguments()[0] === ["test1", "test2"]
+                && ($command->getArguments()[1]->getArrayCopy() === ['fq'=> ['id:test1']]);
```
demiankatz: Same comment here: might be worth checking all instances of getArrayCopy in this test and seeing if some/all can be eliminated.

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
```diff
@@ -171,16 +171,27 @@ protected function getSearchServiceWithMockSearchMethod(
...
+        $commandObj->expects($this->once())->method('getResult')
+            ->will($this->returnValue($collection));
+
+        $checkCommand = function ($command) {
+            return get_class($command) === \VuFindSearch\Command\SearchCommand::class
+                && $command->getTargetIdentifier() === 'Solr'
+                && get_class($command->getArguments()[0]) === \VuFindSearch\Query\Query::class
+                && $command->getArguments()[1] === 0
+                && $command->getArguments()[2] === 20
+                && get_class($command->getArguments()[3]) == \VuFindSearch\ParamBag::class;
```
demiankatz: Would it be more meaningful to compare the array copy of argument 3 against `$expectedParams` rather than only checking the class?

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
```diff
@@ -435,11 +446,25 @@ protected function getResultsFromResponse(
...
+            ->disableOriginalConstructor()
+            ->getMock();
+        $searchService = $this->getMockBuilder(\VuFindSearch\Service::class)
+            ->disableOriginalConstructor()
+            ->getMock();
         // No need to validate the parameters, just return the requested results:
-        $searchService->expects($this->once())
-            ->method('search')
+        $commandObj = $this->getMockBuilder(\VuFindSearch\Command\AbstractBase::class)
+        ->disableOriginalConstructor()
```
demiankatz: More indentation needed here, as above.

module/VuFind/tests/unit-tests/src/VuFindTest/Sitemap/Plugin/Index/CursorMarkIdFetcherTest.php
```diff
@@ -181,21 +180,29 @@ public function testFetching(): void
...
+                [$this->isInstanceOf(GetIdsCommand::class)],
             )->willReturnOnConsecutiveCalls(
                 $this->getMockKeyCommand(),
+                $commandObj,
                 $this->getMockKeyCommand(),
+                $commandObj
             );
-        $service->expects($this->exactly(2))->method('legacyInvoke')
-            ->withConsecutive(
-                [$this->callback($this->getIdsExpectation('*'))],
```
demiankatz, line 196: Have these expectations gotten lost in the refactoring? Do we need to add consecutive expectations to `$commandObj` above to retain this logic?

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Command/RetrieveBatchCommandTest.php
```diff
@@ -0,0 +1,133 @@
...
+        $command = new RetrieveBatchCommand($backendId, $ids, $params);
+        $backend = $this->getMockBuilder(\VuFindSearch\Backend\BackendInterface::class)
+            ->disableOriginalConstructor()->getMock();
+        $rci = $this->getMockBuilder(\VuFindSearch\Response\RecordCollectionInterface::class)
+            ->disableOriginalConstructor()->getMock();
+        $record = $this->getMockBuilder(\VuFindSearch\Response\RecordInterface::class)
+        ->disableOriginalConstructor()->getMock();
+        $backend->expects($this->exactly(2))->method('retrieve')
+            ->withConsecutive(
+                [ $this->equalTo('id1'),$this->equalTo($params)],
```
demiankatz: Minor style issues: extra space after opening bracket, missing space after comma in this line and next.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Command/RetrieveBatchCommandTest.php
```diff
@@ -0,0 +1,133 @@
...
+            )->will($this->returnValue('result'));
+        $this->assertEquals('result', $command->execute($backend)->getResult());
+    }
+
+    /**
+     * Test RetrieveBatch without RetrieveBatchInterface
+     *
+     * @return void
+     */
+    public function testExecuteNotInstance(): void
```
demiankatz: Maybe rename this `testExecuteWithoutRetrieveBatchInterface`?

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Command/SimilarCommandTest.php
```diff
@@ -0,0 +1,110 @@
...
+            ->with(
+                $this->equalTo("id"),
+                $this->equalTo($params)
+            )->will($this->returnValue('result'));
+
+        $command = new SimilarCommand($backendId, "id", $params);
+
+        $this->assertEquals('result', $command->execute($backend)->getResult());
+
+        $this->assertEquals('similar', $command->getContext());
```
demiankatz: Might it make sense to break this test into a few smaller tests, since you seem to be testing pairs of getters/setters after completing the main part of the execute test? It might make things more clear should a test eventually fail.

demiankatz (changes requested):
This is really great work, @skellamp, thank you very much! I just have a few minor questions and suggestions about some of the tests. I also plan to do one last search for any missing method calls when time permits, though I'm pretty confident that you have been very thorough already. :-)

Once all of this is addressed, I might go ahead and merge the new Command tests separately from the rest of the PR, because I think those are of value immediately. The remaining logic here may take a little longer to merge because I want to allow a little more time for community feedback (and I'll request some reviews after we've finished the current round of test cleanup).

One last thing: it might be worth adding tests for the ChannelProvider classes that are modified here, since they currently lack coverage... but I don't think that's essential as part of this PR; it could just be something we keep in mind as a follow-up action. I've done direct hands-on testing of all of those classes and they seem to work as expected.

demiankatz (approved):
Thanks, @skellamp, this is looking good to me! However, since this is a major change, I'm going to ask for reviews from @EreMaijala and @aleksip in case they have any further input.

In particular, I wonder if this is a time to begin a conversation (though it shouldn't hold up the work at hand) about whether the inability to extend/override command objects in the current code is going to be a problem for anyone. Do we need to have some kind of CommandManager that constructs commands, so it's easier to extend things? Or is this okay as-is, and potentially customizable with event listeners?

I'm not looking to over-engineer anything before there's a need/use case, but just curious if anyone sees any problems on the horizon.

EreMaijala (dismissed):
Technically this looks good to me, but I think it would be better not to lose the original copyright year. Use year ranges, or for old files something like `Copyright (C) Villanova University 2010, 2022.` I'd also recommend adding `@author` comments, but I wouldn't mandate that.

aleksip:
@demiankatz I will have a look at this tomorrow, when my brain hopefully functions better...

aleksip:
Looks all good to me!

aleksip (approved):

PR #2601, opened 2022-10-25
Adjust createIntersectionObserver method

LuomaJuha:
Prevents from overriding the intersection observer but allows elements to be observed if it is called.

### DIFF
--- a/themes/bootstrap3/js/check_item_statuses.js
+++ b/themes/bootstrap3/js/check_item_statuses.js
@@ -225,10 +225,7 @@ VuFind.register('itemStatuses', function ItemStatuses() {
     }
     VuFind.observerManager.createIntersectionObserver(
       'itemStatuses',
-      checkItemStatus
-    );
-    VuFind.observerManager.observe(
-      'itemStatuses',
+      checkItemStatus,
       $(container).find('.ajaxItem').toArray()
     );
   }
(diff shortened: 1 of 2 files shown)


demiankatz (commented):
Thanks, @LuomaJuha, that makes sense. One question, though: is it possible there's a scenario where two different onIntersect functions get passed using the same identifier, and then the second one simply gets ignored? Should we try to detect that and at least issue a warning, to avoid hard-to-diagnose problems? I'm not sure how likely that is, but it seems like a potential pitfall here.

LuomaJuha:
@demiankatz Hmm, so a warning when the function is called second time with same identifier would likely be in place?

demiankatz:
@LuomaJuha, I'm not exactly sure what is best -- if the function gets repeatedly called under normal circumstances, we don't want to throw warnings during routine operation, because that will confuse people. Ideally, we would only warn if a different function gets passed in the second time... but there probably isn't an easy way to detect that. So I'm not entirely sure what to do; it just feels like this is a potential problem....

LuomaJuha:
@demiankatz actually there might be, I just have to see where the function is saved inside the observer and compare toString() returns of those both

demiankatz:
@LuomaJuha, that should work -- I couldn't remember if it was possible to get a reasonable string representation of a function for comparison purposes, but if that is reliable, then it sounds good to me.

demiankatz (approved):
Thanks, @LuomaJuha, this makes sense to me, and all tests are passing. I made a small edit to the wording of the warning message in an effort to be a little more specific, and now I'm going to merge this!

PR #2603, opened 2022-10-31
Fix password reset when using ChoiceAuth.

EreMaijala:
`setUpAuthenticationFromRequest()` needs to be called before `$this->getAuthManager()->getPasswordPolicy()`. Otherwise ChoiceAuth state is invalid and it returns false instead of correct information.


demiankatz (approved):
Thanks, @EreMaijala, makes sense to me.

PR #2605, opened 2022-11-03
Fix admin panel tags menu template PHP 8.1 deprecation notice

aleksip:
Just a quick fix for something I noticed testing the Doctrine PR on PHP 8.1.


demiankatz (approved):
Thanks for the catch, @aleksip!

PR #2606, opened 2022-11-03
Add support for validating HTML after each Mink test.

EreMaijala:
Requires that the NU Validator is running as a service.

NU can be downloaded from https://github.com/validator/validator/releases/. Get the vnu.jar package, e.g. vnu.jar_20.6.30.zip. Unzip and start, example:
`java -cp ./vnu.jar -Dnu.validator.servlet.bind-address=127.0.0.1 nu.validator.servlet.Main 8888`.

Then set html_validator to the address the validator runs at when running tests. You can also use html_validator_fail_tests=0 to just report errors and html_validator_log_file to log messges to a file instead of console.

Example:
`./phing.sh phpunitfast -Dhtml_validator=http://localhost:8888/ -Dhtml_validator_fail_tests=0 -Dhtml_validator_log_file=html_validation.log`


EreMaijala:
Note: validation can also be performed as necessary by calling `MinkTestCase::validateHtml()`

demiankatz:
Thanks, @EreMaijala, this looks like a great start! A few thoughts:

1.) I wonder if it would make sense to implement a custom annotation (similar to our retry mechanism) to trigger the HTML validation. Then we could turn it on where it makes sense and avoid doing redundant checking and generating excessive error reports. (Or we could do it explicitly through code, of course... but wondering if an annotation is a useful compromise between always doing it, and doing it manually through code).

2.) I suspect it might be useful to add some build.xml features to support this (at very least, a property to set the new environment variable... but maybe something to start/stop the validation server, if that would make sense).

3.) It might be worth investigating the feasibility of outputting the validation results in the same checkstyle XML format that other tools use... then we could easily wire this into Jenkins.

I haven't done a deep enough dive to investigate the complexity or feasibility of any of these things, so please don't consider this a request to do anything specific -- just brainstorming ideas for starters!

EreMaijala:
@demiankatz An annotation won't cut it if you want to validate the HTML at several points during a test. Since many tests load multiple screens checking both could be useful. It would be fine for validation at the end or excluding a test from validation or suppressing a message if need be. 

The important bit here is to find the balance where validation checks as much as possible without being excessive. My concern about opt-in validation is that we'd forget (I certainly would) to annotate new tests properly.

I'm not so concerned about redundant messages. I'd expect normal runs to result in no messages at all, and if something breaks, nagging about it is just fine. :) We could of course keep track of recent messages and avoid repeating them if it turns out to be an issue.

EreMaijala:
Support in build.xml added, and names of environment variables changes to align with others.

EreMaijala:
@demiankatz The checkstyle output is a nice idea, but ultimately I'd like to just fail tests when issues are detected. I didn't add any support for starting the validator since it's not shipped with VuFind. If we add support for downloading it from GitHub, we could also add support for starting and stopping it, but that's another decision to make. For now I'd consider it an external requirement similar to Selenium or Chrome.

demiankatz (commented):
Thanks for the progress on this, @EreMaijala! A few more thoughts:

1.) You're right that if we "opt-in" with annotations, we're likely to forget about the detail. But what if we add an "opt-out" annotation, so we can skip validation on specific tests? (I suppose the value of this depends on how performant the validation process is -- but if it takes significant time to run each check, skipping some might be of significant value).

2.) Treating this as an external requirement for now seems fine to me.

3.) Regarding the checkstyle output idea, I agree that simply failing tests will make sense once we have a generally-passing baseline... but I thought having a report to work through might be helpful in the meantime if we have a lot to chip away at. Of course, I haven't actually tried this yet, so maybe my opinion will change once I actually see what it looks like in practice.

4.) I notice that we always display info messages, regardless of settings. Is that going to create noise in the middle of the test output, or am I misunderstanding where those messages are going? Again, I haven't actually seen this in practice yet, so maybe the result is self-explanatory, but I'm wondering if we might need a verbosity level setting or equivalent, in case some of those info messages are loud but unhelpful.

demiankatz:
@EreMaijala, just looking through older PRs and I see this hasn't moved in a while. Is it worth picking up this conversation any time soon?

EreMaijala:
@demiankatz Ah, yes, perhaps it is!

1.) Annotation added.

3.) I'll need to fix a couple of generic errors need to be fixed first. Then it's easier to see how much work there is. 

4.) It can be quite verbose indeed. I added an option for logging into a file instead of console. The file could then be checked or its existence used as an indicator that something's reported. What do you think?

build.xml
```diff
@@ -48,9 +48,14 @@
...
   <property name="skip_phpdoc" value="false" />
   <property name="phpdoc_version" value="3.3.1" />
+  <property name="html_validator" value="" /><!-- set to base address of NU Validator (e.g. http://localhost:8888/) for HTML validation during Mink tests -->
+  <property name="html_validator_fail_tests" value="1" /><!-- set to 0 to just log HTML validation errors without causing tests to fail -->
+  <property name="html_validator_log_file" value="" /><!-- set to a file name to log HTML validation messages to it instead of console -->
+  <property name="phpunit_html_validation_params" value="VUFIND_HTML_VALIDATOR=${html_validator} VUFIND_HTML_VALIDATOR_FAIL_TESTS=${html_validator_fail_tests} VUFIND_HTML_VALIDATOR_LOG_FILE=${html_validator_log_file}" />
 
   <property name="version" value="8.1.2" />
 
+
```
demiankatz: Is there a reason for this new blank line?

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
```diff
@@ -752,6 +753,127 @@ protected function logWarning(string $consoleMsg, string $logMsg = ''): void
...
+                } else {
+                    $this->logWarning($message);
+                }
+            }
+            if ($errors) {
+                $message = 'HTML validation errors for '
+                    . $this->session->getCurrentUrl() . ': ' . PHP_EOL . PHP_EOL
+                    . implode(PHP_EOL . PHP_EOL, $errors);
+
+                if (getenv('VUFIND_HTML_VALIDATOR_FAIL_TESTS') !== '0') {
```
demiankatz, line 832: Should the exception throwing be independent of the logging/displaying of messages? I'm thinking we might want to log the error before we fail the test, if we have both logging and test failures turned on at once.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AccountMenuTest.php
```diff
@@ -38,6 +38,7 @@
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org Main Page
  * @retry    4
+ * @skip_html_validation true
```
demiankatz: Is there a compelling reason to skip validation here, or is this just a proof of concept? (Just wondering if this was intentionally committed).

demiankatz (changes requested):
Thanks, @EreMaijala, this seems to be moving in the right direction! Just a few small comments. Beyond that, maybe it's worth putting some TODO checkboxes at the top of the PR so we can keep track of what work remains, since it sounds like you have a few items in process on your end.

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
EreMaijala: Yes, revised. There's now a new html_validator_quiet option for suppressing output during tests as well.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AccountMenuTest.php
EreMaijala: Oops, that was just the result of committing test change..

build.xml
```diff
@@ -48,6 +48,11 @@
   <property name="solr_version" value="9.1.1" />
   <property name="skip_phpdoc" value="false" />
   <property name="phpdoc_version" value="3.3.1" />
+  <property name="html_validator" value="" /><!-- set to base address of NU Validator (e.g. http://localhost:8888/) for HTML validation during Mink tests -->
+  <property name="html_validator_fail_tests" value="1" /><!-- set to 0 to just log HTML validation errors without causing tests to fail -->
+  <property name="html_validator_log_file" value="" /><!-- set to a file name to log HTML validation messages to it instead of console -->
```
demiankatz: This comment still says "instead of console," but I don't think that's true now that you've created the quiet setting.

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
```diff
@@ -752,6 +753,131 @@ protected function logWarning(string $consoleMsg, string $logMsg = ''): void
...
+                if ('info' === $message['type']) {
+                    $info[] = $this->htmlValidationMsgToStr($message);
+                } else {
+                    $errors[] = $this->htmlValidationMsgToStr($message);
+                }
+            }
+            $logFile = getenv('VUFIND_HTML_VALIDATOR_LOG_FILE');
+            $quiet = getenv('VUFIND_HTML_VALIDATOR_QUIET');
+            if ($info) {
+                $message = 'HTML validation messages for '
```
demiankatz: The similarity between the info and error code might lend itself to refactoring to a support method (e.g. `outputHtmlValidationMessages($messages, $level, $logFile, $quiet)`.

demiankatz (changes requested):
Thanks, @EreMaijala -- a couple more thoughts!

Once these are addressed, is this ready to merge from your perspective, or is there still more work to be done?

demiankatz (approved):
Looking good! As we discussed on Slack, we can work on setting this up in Jenkins after some more baseline cleanup work has been done.

PR #2607, opened 2022-11-03
PHPUnit: convert deprecations to exceptions.

demiankatz:
Newer versions of PHPUnit no longer automatically fail tests that emit deprecation warnings. This PR turns on converting deprecations to exceptions, as this makes it much easier to identify and troubleshoot these types of problems. I discovered this while working on a deprecation issue in #2600.


EreMaijala (approved):
A good catch and a welcome change!

PR #2609, opened 2022-11-04
Use aes instead of blowfish in switch_db_hash command test.

EreMaijala:
A minimal change to fix test failures occurring on some of the GitHub Actions runners.


demiankatz (changes requested):
Thanks, @EreMaijala, a good stopgap measure. However, while we're going to the trouble of changing every reference to blowfish in this test anyway, is it worth creating an `$encryptionAlgorithm` property in the test class so we can more easily change this again in future if we need to test something different?

EreMaijala:
@demiankatz Good idea, added!

demiankatz (approved):

demiankatz:
Thanks, @EreMaijala!

PR #2610, opened 2022-11-04
Display deprecations in CI, but don't fail VuFind module

EreMaijala:
We have some known deprecation warnings in third-party libraries, so we need to allow them for the time being.


demiankatz (approved):
Thanks, @EreMaijala -- it's never simple, is it? :-)

EreMaijala:
@demiankatz No, not even when you think it is. :)

PR #2612, opened 2022-11-07
[VUFIND-1588] Add support for WorldCat Search v2 API

demiankatz:
This PR adds a new search backend to connect to the [WorldCat Search v2 API](https://developer.api.oclc.org/wcv2).

Given that the shutdown of the v1 API has been announced for December 31, 2024, we should aim to get this work completed as soon as possible.

TODO
- [x] Build out skeleton of code
  - [x] Search backend
  - [x] Controllers
    - [x] Associated templates
  - [x] Routes
  - [x] Search options/params/results objects
  - [x] Record driver
  - [x] Configuration (ini) file
- [x] Get things working "for real"
  - [x] Flesh out record driver methods
  - [x] Implement sorting
  - [x] Fix link templates (for author, subject, etc.)
- [x] Build equivalent to WorldCatSimilar "related record" module if necessary/appropriate/possible.
- [x] Add facet support.
- [x] Improve itemtype format mapping based on [lists in documentation](https://developer.api.oclc.org/wcv2#/Bibliographic%20Resources/search-bibs).
- [x] Implement holdings tab.
- [x] Investigate additional features:
  - [x] Adding configuration or UI toggles for different types of record grouping
  - [x] Blocking excessively large (>30 terms) queries to avoid API failures (including smarter behavior in similarity recommendations)
  - [x] Supporting rate limiting via 429 messages (which may be implemented in future)
- [x] Add tests
- [x] Resolve [VUFIND-1588](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1588) when finished

Possible future enhancements:
  - Using the "show holding indicators" flag to identify records that are locally held (since the filtering may not be 100% effective). (As of this writing, the feature is not fully implemented in the API; it's in OCLC's backlog, so we can't take further action until something changes on their end).
  - Geographic support (not currently working in the API, but would be useful to support when it is fixed).


demiankatz:
On today's Community Call, @lahmann offered to help with this. His team is currently awaiting API keys. I can help with building the basic "plumbing" to support the new backend -- he can feel free to suggest a timeline if he needs that work completed by a particular date. Obviously, I can't make promises about my availability, but I can at least try to prioritize accordingly. :-)

We're also setting this with a 10.0 milestone, since multiple institutions have a need for/interest in this project, but it may not be completed in time for 9.1 due to its size and scope.

demiankatz:
@lahmann, as promised, I have built the framework for the WorldCat v2 code -- now, if you go to /Worldcat2 under your VuFind instance with this branch, you can perform a search, and you will always see the same fake simulated record.

The next step is to add some real code to \VuFindSearch\Backend\WorldCat2\Connector to actually perform API calls to retrieve real records and search results. I'm sure making that fully functional will also require changes in other places (e.g. to properly format backend parameters in \VuFind\Search\WorldCat2\Params, to perform appropriate logic in \VuFind\Search\WorldCat2\QueryBuilder, and to properly represent the data through the record driver by changing \VuFindSearch\Backend\WorldCat2\Response\RecordCollectionFactory and/or \VuFind\RecordDriver\WorldCat2), but hopefully having a fully-functional baseline to build upon will make the rest of the work a bit easier!

I hope this is helpful; please let me know if you need anything more from my end in order to move forward with this.

jpkanter:
I looked a bit into the code, but frankly, I had other things on my plate and hadn't the time available to me that this PR deserves. I was especially looking into the `bib-holdings` part. The vanilla variant functions as expected, its a bit unfortunated that the geo location doesn't work from OCLCs side, I experimented a bit with the parameters but the V1 functionality wasn't really restore able. I wanted to say that it would be nice to have some way to dynamically set geo coordinates (for example, if a given user wants records closest to them dynamically), but as the function doesn't work anyway properly, therefore I will refrain from such a comment.

demiankatz:
Thanks, @jpkanter, I agree that adding some geographic mechanisms would be a good idea in the long term, but as you say, we need the features to actually work on the OCLC side first. :-)

jpkanter:
Hello,

I tested a bit more around, it was mostly functional testing I am still stuck on actually going through the code line by line to see if I can spot anything amiss.

A minor thing I noticed when I was going through the config.ini, under [SearchTabs] there should be an uncommented "Worldcat2" option but there isn't. That is, if I am not mistaken.

I wanted to leave this here so you don't have to assume that the check was abandoned.

demiankatz:
> I tested a bit more around, it was mostly functional testing I am still stuck on actually going through the code line by line to see if I can spot anything amiss.
> 
> A minor thing I noticed when I was going through the config.ini, under [SearchTabs] there should be an uncommented "Worldcat2" option but there isn't. That is, if I am not mistaken.
> 
> I wanted to leave this here so you don't have to assume that the check was abandoned.

Thanks, @jpkanter! You're right, it's a good idea to update the WorldCat example in config.ini SearchTabs before we forget about it; I've made that change as suggested.

I understand the difficulty of reviewing the code here -- there are a lot of files changed. A lot of it is mostly boilerplate code, though; I think if you focus on the code in the VuFind\RecordDriver, VuFind\Related, VuFind\Search\WorldCat2 and VuFindSearch\Backend namespaces, that's probably the most interesting code to look at. I realize that's still quite a lot, but hopefully that makes the task a bit simpler.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Response/RecordCollectionFactory.php
```diff
@@ -0,0 +1,112 @@
...
+            throw new InvalidArgumentException('Record factory must be callable.');
+        }
+        $this->recordFactory = $recordFactory;
+        $this->collectionClass = $collectionClass ?? RecordCollection::class;
+    }
+
+    /**
+     * Return record collection.
+     *
+     * @param array $response Collection of XML documents
```
maccabeelevine: Haven't done the review yet but coincidentally someone just asked about whether this new API could return XML, and this line hints yes, but I think it's a lie because https://github.com/vufind-org/vufind/pull/2612/files#diff-d8b51aac0547dd71cd90709b806fd885c5ea954bec687dd59da4e90eeda53453R50

demiankatz: Good catch -- copy and paste strikes again; I have revised the comment.

config/vufind/WorldCat2.ini
```diff
@@ -0,0 +1,167 @@
...
+[Record]
+; Because the WorldCat database changes frequently, this feature sometimes
+; behaves unpredictably; it is recommended that you leave it disabled:
+next_prev_navigation = false
+
+; Set to true to include URLs in records; many WorldCat URLs do not work, so they are disabled
+; by default.
+show_urls = false
+
+related[] = "WorldCat2Similar"
```
maccabeelevine, line 140: A comment on this would be helpful, although I see it doesn't have one in WorldCat.ini.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * Get an OAuth2 token.
+     *
+     * @return string
+     * @throws IdentityProviderException
+     */
+    protected function getToken(): string
+    {
+        return $this->authProvider->getAccessToken('client_credentials', ['scope' => 'wcapi'])->getToken();
```
maccabeelevine, line 93: I am not clear if `wcapi` is the complete scope string.  The list on [oclc.org](https://developer.api.oclc.org/wcv2#/Member%20Shared%20Print%20Holdings/find-bib-retained-holdings) (click Authorize) shows several sub-scopes but not that general one.  That said, I use `wcapi` in a local app.  I wonder if it's deprecated.

module/VuFind/src/VuFind/RecordDriver/WorldCat2.php
```diff
@@ -0,0 +1,551 @@
...
+     *
+     * @return array
+     */
+    public function getFormats()
+    {
+        $formats = [];
+        foreach (['generalFormat', 'specificFormat'] as $key) {
+            if (isset($this->fields['format'][$key])) {
+                $formats[] = $this->translate(
+                    'WorldCatFormats::' . strtolower($this->fields['format'][$key]),
```
maccabeelevine, line 105: It is worth revisiting WorldCatFormats/en.ini at some point.  Most results I see have format `PrintBook` (no space) which could be translated to have a space.  Guessing any other two-word format could use something similar.

module/VuFind/src/VuFind/RecordTab/HoldingsWorldCat2.php
```diff
@@ -0,0 +1,111 @@
...
+    {
+        return 'Holdings';
+    }
+
+    /**
+     * Get holdings information from WorldCat (null if none available).
+     *
+     * @return ?array
+     */
+    public function getHoldings()
```
maccabeelevine, line 72: Not clear what this means in the context of WorldCat.  Current institution's holdings?  All known holdings (within the limit of the API, which I think is closest 50 institutions)?  I'm getting an empty tab.  Is there something required in [Holdings] in the .ini?

config/vufind/WorldCat2.ini
```diff
@@ -0,0 +1,167 @@
...
+; on-screen text will be run through the translator, so be sure to update language
+; files if necessary.  The order of these settings will be maintained in the
+; drop-down list in the UI.
+;
+; For a complete list of legal values, see the Bibliographic record indexes documentation:
+; https://help.oclc.org/Librarian_Toolbox/Searching_WorldCat_Indexes/Bibliographic_records/Bibliographic_record_indexes
+[Basic_Searches]
+kw              = "All Fields"
+ti:se           = Title
+au              = Author
```
maccabeelevine, line 91: I'm unable to search by Author successfully, getting a timeout.  Could just be me, and I see that `au` is a valid index.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
```diff
@@ -0,0 +1,211 @@
...
+     * @return Response
+     * @throws IdentityProviderException
+     * @throws InvalidArgumentException
+     * @throws RuntimeException
+     * @throws ExceptionRuntimeException
+     */
+    protected function makeApiCall(string $path): Response
+    {
+        if (!isset($this->session->token)) {
+            $this->session->token = $this->getToken();
```
maccabeelevine, line 110: If the auth credentials are wrong (or never set) in the .ini, this throws a not-very-helpful exception from the oauth2 library deep within `getToken`.  

image

Worth catching somehow and faking a 401 so it can be handled a few lines down.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
```diff
@@ -0,0 +1,211 @@
...
+        }
+        $headers = ['Authorization: Bearer ' . $this->session->token];
+        $this->client->setHeaders($headers);
+        $this->client->setUri($this->baseUrl . $path);
+        $this->debug($path);
+        $response = $this->client->send();
+        // If authorization failed, the token may be expired; re-request and try again:
```
maccabeelevine: 'Authorization' would be 403 not 401 (authentication), but actually the line below should perhaps catch both?  People may have a variety of OCLC API keys, some authorized for different APIs so would get a 403.

maccabeelevine (changes requested):
I'm unable to get the holdings tab working.  Otherwise it works well, and the code seems clean as far as I can see, just some minor things below.

config/vufind/WorldCat2.ini
demiankatz: Comment added.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
demiankatz: Sounds like a question for @librarywebchic. I'm not sure where this code originally came from, but I'm sure she can clarify whether or not it will remain valid going forward.

module/VuFind/src/VuFind/RecordDriver/WorldCat2.php
demiankatz: Agreed. In the long term, I'd like to do a more general overhaul of format handling -- move the core format names out of the default text domain and into a format-specific one, and use aliasing instead of in-code conversion to map equivalent formats from various backends. But alas, it's a lot of work, and not a top priority, so it hasn't gotten done yet. For now, my goal here was to have reasonable support for the top-level WorldCat formats, but I didn't attempt to address all of the sub-categories. If there's sufficient interest, I am of course happy to, but I didn't want to create tons of work for the translators if nobody cared. :-)

module/VuFind/src/VuFind/RecordTab/HoldingsWorldCat2.php
demiankatz: This means libraries holding the item... I could rename this method to getHoldingInstitutionData() if you think that would be more clear. Here's an example of what it looks like for me:

image

Last I checked (which was a few weeks ago), all of the geographic functionality in the API was broken, so I wasn't able to use it. So right now it's just going to be an arbitrary list of holding institutions, without much useful information.

If you're getting an empty tab for everything, I wonder if perhaps some extra permission is needed on your API key to activate the functionality. I could ask @librarywebchic if that would help!

config/vufind/WorldCat2.ini
demiankatz: Author search is working fine for me. Are you still having issues? If so, for what search(es)? Here's an example of a successful one:

image

demiankatz: (Note that it seems a little wonky that my top hit seems to be ranking based on a title match even though I specified an author search; not sure what that's about... I do get entirely different results if I use different search types, so it's not like the criterion is being ignored, though!)

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
librarywebchic: wcapi is sort of an overarching scope that gets you all the others. if you want to be more precise you'd only ask for the ones you are using which I would guess would be wcapi:view_bib and wcapi:view_institution_holdings

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
demiankatz: I'm not sure I understand what you mean. If there is no token in the session, we try to get a fresh token. The logic a few lines down attempts to get a fresh token if the one in the session has expired. If failing to get the token the first time led to us simulating a 401, we'd just end up repeating the same call that failed in the first place. I don't think that would help! Am I misunderstanding something?

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
demiankatz: I think a 403 would indicate a valid token was being sent to an inappropriate service, and in that context, I don't think requesting a refresh of the token would make any difference. You're right that my comment is misleading, though; I have just fixed it.

librarywebchic: Token expired will return a 401. Token for the wrong set of scopes a 403.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
demiankatz: Thanks, @librarywebchic! Sounds like it's probably okay to leave this as-is, but we could certainly make this configurable in future if that would be valuable.

module/VuFind/src/VuFind/RecordTab/HoldingsWorldCat2.php
librarywebchic: @maccabeelevine if you send me the key you're using I can look in the log files to see what's going on with the holdings request.

module/VuFind/src/VuFind/RecordDriver/WorldCat2.php
maccabeelevine: Yeah I think what is done in this PR is certainly sufficient, and as you say doesn't create work for the translators.  A broader review sometime in the future would make sense.

module/VuFind/src/VuFind/RecordTab/HoldingsWorldCat2.php
maccabeelevine: In debug, I don't see this HoldingsWorldCat2 class ever being called.  The path I see is that holdingsils.phtml calls the record driver's `getRealTimeHoldings()`.  WorldCat2 record driver doesn't define that, so it falls back to the DefaultRecord implementation which returns an empty result.  ILSAwareTrait would do the right thing (maybe?) but the WorldCat2 record driver doesn't use it.

I did email Karen with my API key in case that makes the difference somehow, but seems not.

config/vufind/WorldCat2.ini
maccabeelevine: Yes, today author search is working fine!  (Although yeah the results are maybe not great.)  Who knows.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
maccabeelevine: Ok yeah "simulating" a 401 response is not a good idea.  My point (badly stated) was that when I have an actual authentication problem (wrong password), there is an uncaught exception thrown, not a response with a 401.  So the code handling the 401 never gets triggered.  May be different if the reason for the 401 is actually an expired token instead of a bad password.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
maccabeelevine: Ok I agree, there's no point in retrying after a 403.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
demiankatz: My point is that this code is about dealing with a 401 coming from WorldCat, not with sending a 401 to the end user. But I'm not sure that it's appropriate to send a 401 to the VuFind user when this happens -- it's more of a 500 situation, since the server is misconfigured...

module/VuFind/src/VuFind/RecordTab/HoldingsWorldCat2.php
demiankatz: If you're looking at a WorldCat2 record, you should never see holdingsils.phtml at all. Is it possible you somehow have an outdated RecordTabs.ini in your test instance that's preventing the appropriate WorldCat2-specific tabs from being loaded?

maccabeelevine: sigh yes.  I have one in `local` that was overriding your upstream change.  Tab looks good!  Having not seen this functionality in the original WorldCat driver, I assume it's the same -- it would be useful to linkify those holdings to WorldCat so the patron can do something about them, but assuming this is consistent with the old WorldCat API implementation, no reason to iterate on that right now.

Also FWIW I tried adding `heldInState = "US-HI"` to the .ini and it works great, I can see all the places I could borrow the book if I were lucky enough to be near a Hawaiian beach instead of in my office.

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
maccabeelevine: Ok I think that's a fair point.  I hate how opaque the error message is for the developer, but ultimately it's a 500 which as you say is correct.

maccabeelevine (approved):
All of my feedback has been discussed/addressed.

module/VuFind/src/VuFind/RecordTab/HoldingsWorldCat2.php
demiankatz: The original WorldCat holdings tab had more functionality than the current version, because the original API had more functionality. There's some uncertainty about whether a separate API (possibly for-fee) is needed to get some of the data going forward. @librarywebchic is investigating, and we can enhance the tab in a separate future PR once progress is made.

Glad to hear that the limit option is working for you!

module/VuFindSearch/src/VuFindSearch/Backend/WorldCat2/Connector.php
demiankatz: Yeah, it's not ideal. If this turns out to be a recurring support problem/source of frustration, we might at least be able to wrap the low-level exception in a higher-level exception... but that could also make things even more confusing if the failures differ from what we expect them to be. I'll leave it alone for now but we can revisit if needed!

demiankatz:
It's been a long road, but I'm finally merging this. Thanks for helping us over the finish line, @maccabeelevine!

PR #2616, opened 2022-11-09
Improve BrowZine DOI lookup configurability.

EreMaijala:
- Makes services configurable and re-orderable and adds settings for retraction notices.
- Adds an option to use local icons instead of remote assets.
- Adds an option to load remote icons via cover loader proxy.
- Adds an option to load DOI lookup links in a new window/tab.

An example of retraction notice can be found at least from EDS by searching for "10.23736/S1973-9087.21.06696-X" or "Multidisciplinary program based on early management of psychological factors reduces disability of patients with subacute low back pain: one-year results of a randomized controlled study" (see [Retraction Watch](https://retractionwatch.com/2022/11/09/pain-researchers-lose-three-papers-after-cochrane-group-questioned-data/#more-125984) for more information).

The constructors etc. are a bit complicated as I tried to maintain back-compatibility.

TODO:
- [x] Note BC breaks in changelog when merging.
- [x] Mark [VUFIND-1538](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1538) resolved when this is merged.


EreMaijala:
For Primo support, #2615 is required.

config/vufind/BrowZine.ini
```diff
@@ -23,10 +23,26 @@ default_view         = list
 ; setting. The default is "none," which performs no filtering.
 ;filterType = exclude
 ; This repeatable section can be used to filter based on link type; legal
-; options are "browzineWebLink" and "fullTextFile"
+; options are the keys defined in DOIServices section.
+; Note that this setting is available for convenience and compatibility with previous
+; versions. You can get the same results by changing the Services section.
```
demiankatz: Do you mean DOIServices?

config/vufind/config.ini
```diff
@@ -1308,6 +1310,14 @@ show_in_results = true      ; include in search results
 show_in_record = false      ; include in core record metadata
 show_in_holdings = false    ; include in holdings tab of record view
 
+; Whether to load any third-party icons for the DOI services via VuFind's cover
+; loader proxy to avoid any privacy implications. Ensure that the necessary domains
+; are allowed in Content/coverproxyCache[] setting. Default is false.
+proxy_icons = true
```
demiankatz, line 1316: This says default is false, but it's set to true. Is that intentional? I'm not opposed to having default configuration that differs from default behavior when configuration is disabled, but that can be a bit confusing!

module/VuFind/src/VuFind/AjaxHandler/DoiLookup.php
```diff
@@ -118,4 +155,41 @@ public function handleRequest(Params $params)
...
+     *
+     * @return array
+     */
+    protected function proxifyIconLinks(array $dois): array
+    {
+        if (!$this->proxyIcons) {
+            return $dois;
+        }
+
+        $serverHelper = $this->viewRenderer->plugin('serverurl');
```
demiankatz, line 160: I realize that it _should_ be impossible to reach this code without using the updated factory, but maybe it's worth throwing an exception of the viewRenderer is missing, just to be on the safe side.

module/VuFind/src/VuFind/AjaxHandler/DoiLookupFactory.php
```diff
@@ -68,12 +68,11 @@ public function __invoke(
             throw new \Exception('Unexpected options passed to factory.');
         }
         $config = $container->get(\VuFind\Config\PluginManager::class)
-            ->get('config');
-        $pluginManager = $container->get(\VuFind\DoiLinker\PluginManager::class);
+            ->get('config')->toArray();
```
demiankatz, line 71: Is there a reason to send the whole configuration, or would it be cleaner to send just the DOI section? (I'm fine with sending it all if you think it might improve future flexibility, but it would be less verbose to send the relevant section otherwise).

demiankatz (changes requested):
Thanks, @EreMaijala, this looks great -- just a few comments now, and I'll do hands-on testing after you've had a chance to apply the finishing touches.

Also note that you might want to consider updating the test for the BrowZine DOI handler to cover the new functionality; hopefully it's not much work since there's already existing coverage.

I'm not exactly sure why LGTM is complaining about URLs since you haven't added anything new, but maybe it's just warning because things have moved around.

config/vufind/BrowZine.ini
EreMaijala: Fixed.

config/vufind/config.ini
EreMaijala: I changed default to false so that current behavior is kept.

module/VuFind/src/VuFind/AjaxHandler/DoiLookup.php
EreMaijala: Fixed and then I realized that local icons were not properly rendered when proxying was disabled. Had to revise so that the method does everything when view renderer is available and nothing otherwise. That's not quite optimal, so I'm wondering if a BC break would be better in this case.

module/VuFind/src/VuFind/AjaxHandler/DoiLookupFactory.php
EreMaijala: It would be cleaner, but I did it this way to avoid having to change the signature if something else from the config is needed in future, like happened with BrowZine.php here, or in a derived class.

EreMaijala:
Not sure why LGTM flagged this now. The link is nothing new, but I've now added `rel="noreferrer"` to it.

module/VuFind/src/VuFind/AjaxHandler/DoiLookup.php
demiankatz: Yeah, maybe a BC break is justified in this case. I'm rather doubtful that many (any?) people have customized this, so it's unlikely to have much impact... and then we can have less confusing code.

EreMaijala: I've changed the code to always expect new constructor style. I also added test coverage for the new functionality.

EreMaijala:
`rel="noreferrer"` didn't help with LGTM. Not sure what to do about it.

demiankatz:
> `rel="noreferrer"` didn't help with LGTM. Not sure what to do about it.

I think they want `rel="noopener noreferrer"` to be fully satisfied. I think that's probably safe for us to add; I don't think we need `Window.opener` for anything.

EreMaijala:
Added, hope it helps. A bit odd since noreferrer implies noopener according to [the spec](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer).

demiankatz:
@EreMaijala, for whatever reason, adding noopener made no difference to the LGTM alert... but maybe we shouldn't worry about that for now, since perhaps #2617 will turn up different (better?) results here.

This small matter aside, is this now ready for another test/review, or are you still working on something?

EreMaijala:
@demiankatz That's odd. Ignoring LGTM, this is ready for review and testing.

config/vufind/BrowZine.ini
```diff
@@ -23,10 +23,26 @@ default_view         = list
...
+local_icons = false
+
+; This section defines the services to display from BrowZine and their display order.
+; Each key is a service name (see https://thirdiron.atlassian.net/wiki/x/WIDqAw for
+; more information) and value contains translation key, local icon and Third Iron's
+; icon asset (optional) separated by a pipe character.
+[DOIServices]
+browzineWebLink = "View Complete Issue|browzine-issue|https://assets.thirdiron.com/images/integrations/browzine-open-book-icon.svg"
+fullTextFile = "PDF Full Text|browzine-pdf|https://assets.thirdiron.com/images/integrations/browzine-pdf-download-icon.svg"
+retractionNoticeUrl = "View Retraction Notice|browzine-retraction"
```
demiankatz, line 44: Do you think we should fall back on a local icon if now BrowZine icon URL is provided? Otherwise, by default, retraction notice has no icon at all.

demiankatz (changes requested):
@EreMaijala, I did a quick test of this and it looks good to me, though I have one small suggestion.

On a side note, I kept running into a problem where the EDS search results auto-redirected me to the DOI when I searched for raw `10.23736/S1973-9087.21.06696-X`. I haven't had time to dig into this in much depth, but I think there may be a bad default setting for the DOI recommendation module that's causing undesirable results in this situation. Did you encounter the same thing? If not, I'll spend a little more time on my end figuring out what's going on.

config/vufind/BrowZine.ini
EreMaijala: Makes sense, now added.

EreMaijala:
@demiankatz I didn't encounter automatic redirection of any kind. I would have been alarmed by such behavior!

demiankatz:
Thanks, @EreMaijala, I'll test this again tomorrow when I'm back in the office.

I also figured out the redirection thing: I was using a copy of my production EDS.ini for testing, and I forgot that it was configured to auto-redirect pure DOI searches. That's a decision we'll need to revisit in light of retractions!

EreMaijala:
@demiankatz Yes, at least in the light of accessibility, and perhaps GDPR, automatically redirecting to an external site is not preferable.

demiankatz (approved):
Looking good to me, and all tests are passing. Thanks again, @EreMaijala!

PR #2617, opened 2022-11-09
Set up GitHub code scanning with CodeQL.

demiankatz:
This replaces LGTM, which is shutting down in December.

TODO:
- [x] Clean up alerts if practical
- [ ] After merging, examine other possible code scanning workflows to add

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'java', 'javascript' ]
+        language: [ 'javascript' ]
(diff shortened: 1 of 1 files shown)


EreMaijala:
@demiankatz I checked the cookieconsent alert, and it's a false positive. The result is added to a node's textContent property, so even if the sanitization didn't work, it wouldn't cause any issues. The tag removal is there to just make a label from HTML input for accessibility, and the input comes from configuration and translations, so I'd consider it trusted.

EreMaijala:
@demiankatz The other issues look like false positives to me, so I'd mark them ignored as well.

EreMaijala (approved):
The new workflow looks good to me.

demiankatz:
Thanks, @EreMaijala, it's possible to dismiss false positives through the GitHub UI, so I've done that on the critical alert you analyzed, and now the build is showing as passing here. Merging now!

PR #2618, opened 2022-11-10
PAIA: delete cached user data after successfully placing hold

charbugs:


demiankatz:
Thanks, @charbugs! This looks reasonable to me, but I'm going to wait for a review from another PAIA user like @lahmann or @olli-gold since I can't test this directly.

Also, I wonder if this PR points at an existing problem that needs to be addressed: we appear to be using the same cache for both user and item data, without prefixing cache keys. Isn't there at least a small chance of a cache collision between these different identifiers? Should we do something about that? (If this is a problem, it should obviously be addressed in a separate PR, and I don't think we need to hold off on merging this one... but perhaps that is future work that somebody should consider).

lahmann:
@demiankatz I have no objections, your question regarding prefixing cache keys is valid. @charbugs we could address this in another pull request?

demiankatz (approved):
Thanks, @lahmann. I'll merge now and will look forward to a follow-up PR as time permits.

PR #2619, opened 2022-11-10
GitHub actions: Enable automated dependency review.

demiankatz:


demiankatz:
@EreMaijala, as a follow-on to #2617, I'm adding this action to check for vulnerable dependencies. According to the documentation, it should support PHP/Composer and Javascript/NPM. It's not finding any vulnerabilities right now, so it's a little hard to tell if it's actually working, but since there's nothing really to configure except to turn it on, I assume it is.

I also notice that there are pre-built workflows for eslint and some other tools; I wonder if it would make sense at some point for us to separate the code review tasks from the test tasks, as it would simplify our matrix a bit. No rush on that, though, just something to think about!

EreMaijala (approved):
`npm update` is currently reporting some issues, but as far as I've understood the dependency review only checks for changes that would introduce new ones, and obviously adding the review doesn't introduce any new issues.

demiankatz:
Thanks, @EreMaijala, that makes sense... I just thought there was some possibility that the first time it ran, it might consider all dependencies to be new... Probably for the best that it isn't working like that!

Maybe it's worth doing some general dependency updating in the near future to give it some exercise. My last round of dependency updates stalled out when I ran into annoying new conflicts between php-cs-fixer and PHP CodeSniffer, but that's no reason to avoid updating other dependencies.

PR #2620, opened 2022-11-10
Prevent duplicate bib_id values in Solr for course reserves

damien-git:
Currently, when a course has several items in the same instance in FOLIO, the number of items returned for a search does not match the number of returned instances when the user clicks on a course. This is because of duplicate values for bib_id in Solr (one for each item instead of one for each instance). This fix prevents the duplicates, which fixes the issue with the number of items.


demiankatz (approved):
Thanks for the fix!

PR #2622, opened 2022-11-10
Folio: optimized course reserve lookup

damien-git:
Currently the Folio driver is slow to find reserve information in findReserves(). This is mostly because for each reserve item, it looks for the item, the holdings and the instance in Folio. This is not necessary because we can use the returned copiedItem.instanceId or copiedItem.instanceHrid, depending on getBibIdType().
After this optimization, indexing the course reserves takes a third of the time it took before.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -1463,16 +1463,17 @@ class Folio extends AbstractAPI implements
     public function findReserves($course, $inst, $dept)
     {
         $retVal = [];
+        $idType = $this->getBibIdType();
 
         // Results can be paginated, so let's loop until we've gotten everything:
         foreach ($this->getPagedResults(
             'reserves',
             '/coursereserves/reserves'
         ) as $item) {
-            try {
-                $bibId = $this->getBibId(null, null, $item->itemId);
-            } catch (\Exception $e) {
-                $bibId = null;
+            if ($idType == 'hrid') {
+                $bibId = $item->copiedItem->instanceHrid ?? null;
+            } else {
+                $bibId = $item->copiedItem->instanceId ?? null;
             }
             if ($bibId !== null) {
                 $courseData = $this->getCourseDetails(


demiankatz:
Thanks, @damien-git, this is great! The driver course reserves code was originally developed prior to the 1.0.0 release of mod-courses, and I'm not sure if the copiedItem elements existed at that time (though it's also entirely possible I just overlooked them)... either way, it looks like they've been there since the first official release, so this seems like a safe and obvious improvement.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1463,16 +1463,17 @@ protected function getInstructorIds($courseListingId)
...
         // Results can be paginated, so let's loop until we've gotten everything:
         foreach ($this->getPagedResults(
             'reserves',
             '/coursereserves/reserves'
         ) as $item) {
-            try {
-                $bibId = $this->getBibId(null, null, $item->itemId);
-            } catch (\Exception $e) {
-                $bibId = null;
+            if ($idType == 'hrid') {
```
demiankatz, line 1473: If you wanted to make this even more concise, you could consider something like:

```
$idProperty = $this->getBibIdType() === 'hrid' ? 'instanceHrid' : 'instanceId';
$bibId = $item->copiedItem->$idProperty ?? null;
```

demiankatz (changes requested):
Before merging this, it occurred to me that it could be simplified... but if you dislike my suggestion, feel free to say so, and I'm happy to merge as-is. :-)

module/VuFind/src/VuFind/ILS/Driver/Folio.php
damien-git: Works for me, please edit as you like.

demiankatz: Thanks! It looks like I don't have write access to this branch, so I can't make changes directly. If you'd like me to, then please give me access. If not, I can just modify as part of the merge-to-dev process. Either way, I'm currently waiting to get some course data set up in my local test environment, and I'll give this a closer look once that's ready.

demiankatz:
@damien-git, it seems like the remaining huge problem here is that retrieving the instructor data requires iterating through all the courses. My test environment has 6,000+ courses, and so it takes hours. I wonder if there are any plans to offer a better option here (or if there's already something we could do to improve performance). As it is, I still haven't been able to test this, because my last attempt got interrupted by a network outage, and I don't have enough time left today to start over again.

demiankatz:
Thanks again, @damien-git. I was finally able to test this properly and it looks good, so I'm merging my simplified version as commit https://github.com/vufind-org/vufind/commit/82daaf82ef8568f7fd9e26c18839dab7888070c0.

damien-git:
@demiankatz Sorry about the permission, I thought I already granted it earlier. You are right that there are still loops, and maybe it could be further optimized.

demiankatz:
Thanks, @damien-git! Regarding the instructor loop, I eventually ended up adding caching logic to the driver so that when my tests got interrupted, I could resume from where I left off instead of having to start over. I didn't keep this code after I was finished testing, but maybe some degree of configurable caching might be useful as a future feature. I'm happy to share what I did if there's any interest, but I also feel that introducing a cache in this context could cause problems if the expiration time were not tuned correctly.

PR #2623, opened 2022-11-11
Increase flexibility in batch import marc script extension matching

meganschanz:
Currently the batch import marc scripts match files ending in .xml and .mrc. This update includes the additional common MARC extension, .marc, as well as adding in case-insensitive matching on the extension. So for example, now .XML and .MARC will now both be included and processed by the import scripts.

In the windows .bat scripts, the additional extension was added, but was not tested since I am not on a Windows environment. Those scripts should already include case-insensitive matching.

For context, we (The Michigan State University Libraries) are receiving MARC files from EBSCO with the .marc extension and authority files we receive from Backstage have a .MRC extension. So we thought it would be simpler to make the import script more flexible rather than rename the files every time we receive them from both places.

### DIFF
--- a/harvest/batch-import-marc-auth.bat
+++ b/harvest/batch-import-marc-auth.bat
@@ -102,7 +102,7 @@ md %BASEPATH%\processed
 :processedfound
 
 rem Process all the files in the target directory:
-for %%a in (%BASEPATH%\*.xml %BASEPATH%\*.mrc) do (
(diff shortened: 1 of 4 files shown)


demiankatz (approved):
Thanks, @meganschanz, this is a great enhancement!

demiankatz:
(I've tested under Windows and Linux to confirm that it works as expected in both environments).

PR #2625, opened 2022-11-15
Set TranslatableString::displayString default value as an empty string

LuomaJuha:
Using null in php 8.1 gives dep notice when passed into explode.

### DIFF
--- a/module/VuFind/src/VuFind/I18n/TranslatableString.php
+++ b/module/VuFind/src/VuFind/I18n/TranslatableString.php
@@ -50,10 +50,12 @@ class TranslatableString implements TranslatableStringInterface
      *
      * @var string
      */
-    protected $displayString;
+    protected $displayString = '';
 
     /**
      * Whether translation is allowed
+     *
+     * @var bool
      */
     protected $translatable;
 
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/I18n/TranslatableString.php
```diff
@@ -50,10 +50,12 @@ class TranslatableString implements TranslatableStringInterface
      *
      * @var string
      */
-    protected $displayString;
+    protected $displayString = '';
```
demiankatz, line 53: I didn't think it was necessary to initialize a property that is initialized by the constructor. Did adding this specifically change something?

LuomaJuha: Oh, actually I must've been wrong with this one. Was doing a second task at the same time and mixed them up.

demiankatz: If you don't mind reverting this unnecessary change, we can at least merge the rest to add the missing annotation!

LuomaJuha: Will convert this to draft and find the real culprit.

LuomaJuha: Seems like someone is actually putting a null variable into the translate function and it starts to fill the php_error.log with dep notices in our end. Will need to find the culprit there. Sorry for the confusion!

PR #2626, opened 2022-11-15
CookieConsent: Do not pass null description into translate function

LuomaJuha:
@demiankatz this was the real culprit it seems.


EreMaijala (approved):
Description should always be specified in the configuration, but this can still happen.

PR #2628, opened 2022-11-16
Clarify solrmarc.path in SolrMarc's import.properties

natecollins:
Addresses issue when a user copies `import/import.properties` to `local/import/import.properties` but does not modify `solrmarc.path` variable (there is no indication to the user that they ought to do this, and would otherwise be easy to miss). This results in solrmarc failing to read in overrides from `local/import/marc_local.properties` _unless_ the original `import/marc_local.properties` file is first removed.

Commenting out the default solrmarc.path from `import/import.properties` prevents this unexpected behavior. In testing this change, solrmarc still successfully loads the default `import/marc.properties` and `import/marc_local.properties` with that line now commented. Users are still free to uncomment the line and set the value if needed.

### DIFF
--- a/import/import.properties
+++ b/import/import.properties
@@ -8,8 +8,11 @@ solr.core.name = biblio
 solr.indexer.properties = marc.properties, marc_local.properties
 solr.hosturl = http://localhost:8983/solr/biblio/update
 
-# where to look for properties files, translation maps, and custom scripts
-# note that . refers to the directory where the jarfile for SolrMarc is located.
+# Where to look for properties files, translation maps, and custom scripts.
+# Note that . refers to the directory where the jarfile for SolrMarc is located.
+# You can use a pipe (|) character to create a multi-directory search path.
+# You can use install.php to automatically create and configure your local
+# override of this file for common use cases.
 solrmarc.path = /usr/local/vufind/import
 
 # Path to your marc file
--- a/import/import_auth.properties
+++ b/import/import_auth.properties
@@ -8,8 +8,11 @@ solr.core.name = authority
 solr.indexer.properties = marc_auth.properties
 solr.hosturl = http://localhost:8983/solr/authority/update
 
-# where to look for properties files, translation maps, and custom scripts
-# note that . refers to the directory where the jarfile for SolrMarc is located.
+# Where to look for properties files, translation maps, and custom scripts.
+# Note that . refers to the directory where the jarfile for SolrMarc is located.
+# You can use a pipe (|) character to create a multi-directory search path.
+# You can use install.php to automatically create and configure your local
+# override of this file for common use cases.
 solrmarc.path = /usr/local/vufind/import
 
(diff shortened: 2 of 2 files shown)


demiankatz:
@natecollins, thanks for the contribution. However, as-is, this will cause some problems...

Users are not expected to manually copy import.properties into place; the local/import/import.properties is intended to be automatically created by VuFind's install script, which copies the default version and does some string replacing to put correct paths in place. See:

https://github.com/vufind-org/vufind/blob/23eed813745e5ef2018705e91b52a9dcdedd8d42/module/VuFindConsole/src/VuFindConsole/Command/Install/InstallCommand.php#L632-L636

If you comment out the default setting, then the install script will not work correctly, because the setting will be commented out in the resulting file.

This suggests that we need to approach this a little differently, with one of at least two possible options:

1.) Instead of changing the default setting, add a comment highlighting the potential problem.

2.) Change the install script so it is smart enough to uncomment the setting, if necessary. (Possibly as simple as changing `\s*` to `[\s#]*`).

What do you think?

demiankatz:
Another consideration: there is a separate configuration file for authority records (https://github.com/vufind-org/vufind/blob/23eed813745e5ef2018705e91b52a9dcdedd8d42/import/import_auth.properties). Whatever we decide to do here, we should also do for that file, to keep things consistent.

natecollins:
Ah, thanks @demiankatz for explaining the missing pieces. Our build process had been using a template from the original file, but this does clarify the intended workflow.

For increased clarity, noting the potential issue in a comment would have reduced a bit of confusion on our side. Also, we weren't aware that you could specify multiple paths using a pipe; a comment about that could be another improvement!

As for adding the comment in the default file and then removing it during the install copy, I defer to your preference. It would have helped us in our build process, but at the same time the issue was largely our own fault as we apparently weren't following the recommended setup procedure. Happy to help in doing any updates - just let me know what you decide.

demiankatz:
@natecollins, it seems that simply leaving the solrmarc.path setting uncommented and expanding upon the comment is the least risky option, so if that feels helpful enough to you, that would be the easiest path forward. Then it's just a question of deciding on the appropriate wording. Maybe something like:

```
# Where to look for properties files, translation maps, and custom scripts.
# Note that . refers to the directory where the jarfile for SolrMarc is located.
# You can use a pipe (|) character to create a multi-directory search path.
# You can use install.php to automatically configure this file for common
# use cases.
```

If you like that, you can revise this PR to include it in place of the previous changes. Feel free to make your own tweaks! If there's anything else you'd like to discuss, just let me know.

natecollins:
All set @demiankatz; the pull branch has been updated. Thank you!

demiankatz (approved):
Thanks, @natecollins, I appreciate the help improving the documentation -- I hope this helps the next person to run into it. :-)

If there's anything else I can assist you with, just let me know!

PR #2400, opened 2022-04-20
Improve styles for links containing icons (and related issues)

crhallberg:
TODO:
- [x] Author links may present a problem (discuss)
- [x] Titles of related works
- [x] Feedback link missing underlining
- [x] Delete links in search history need adjustment
- [x] View links above search results need adjustment
- [x] Checkbox facets need adjustment
- [x] Search tool links at bottom of results need adjustment
- [x] OpenURL links need adjustment
- [x] DOI links (doi.js) need adjustment 
- [x] Add to Book Bag on record page needs underlining removed
- [x] First/Prev/Next/Last buttons on record pages need review and to be made consistent (are they links or buttons?)
- [x] Place Title Level Request link needs adjustment
- [x] Show Other Versions (in results and on record view) needs adjustment
- [x] Inconsistent styles on request links (place hold underlined, storage retrieval / ILL not underlined)
- [x] All action links in Library Cards have also the icon underlined (Add Library Card, Edit, Delete)
- [x] Fix failing integration tests
- [x] Fix search history delete button
- [x] Fix advanced search link underlining
- [x] Fix delete button on "edit list item" page
- [x] Fix result-list.phtml full text link formatting
- [x] Fix wrapping of long status display in MyResearch menu
- [x] Extended holdings template needs icon alignment help
- [x] Close [VUFIND-1533](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1533) when merging

PR #2629, opened 2022-11-17
Include DOI link icon in the anchor.

EreMaijala:
I believe the icon should be part of the link. We're currently quite inconsistent [1] with that, but the results view type links can be used as an example. It causes the unsightly underlining issue, but still better for usability and the icon getting the link color.

[1] e.g. `Show QR Code` and `Save to List` don't include the icon, while `Add to Book Bag` does.

### DIFF
--- a/themes/bootstrap3/js/doi.js
+++ b/themes/bootstrap3/js/doi.js
@@ -38,9 +38,9 @@ VuFind.register('doi', function Doi() {
                 var icon = $('<img />');
                 icon.attr('src', response.data[currentDoi][i].icon);
                 icon.attr('class', 'doi-icon');
-                $(doiEl).append(icon);
+                newLink.prepend(icon);
               } else if (typeof response.data[currentDoi][i].localIcon !== 'undefined') {
-                $(doiEl).append(response.data[currentDoi][i].localIcon);
+                newLink.prepend(response.data[currentDoi][i].localIcon);
               }
               $(doiEl).append(newLink);
               $(doiEl).append("<br />");


demiankatz:
@EreMaijala, #2400 is work in progress to normalize all the link/icon behavior. @crhallberg just finished another round of bug-fixing there, so I'm planning to test it again this morning. Maybe we should hold off on finishing this until that one is merged.

EreMaijala:
@demiankatz #2400 is not affecting this one, but I'm fine with incorporating the fixes there if it seems to be nearing completion.

demiankatz:
@EreMaijala, if you don't expect #2400 to cause a need for further changes here, I'm happy to review/merge this now... I just thought if the styling there would be useful for improving this code, then we might as well wait instead of changing the code twice.

EreMaijala:
@demiankatz Further changes are needed here to get the underlining nice and uniform like in #2400. But since #2400 is pretty extensive, I can't merge it to our tree without fair amount of work, so my wish would be to merge this (then it could be cherry-picked or merged before #2400 lands), then merge dev to #2400 and add the tweaks to doi.js there. But if it's too much trouble, I can just override doi.js locally and undo that later.

demiankatz:
@EreMaijala, if that makes your life easier, it's fine with me -- I just want to be sure we don't forget about this, but that's easily done by adding a TODO checkbox on #2400 as soon as this is merged. I'll try to take a closer look here as soon as I'm done with my current #2400 review.

EreMaijala:
@demiankatz I've already added a TODO checkbox in #2400 since it needs attention in any case.

demiankatz (approved):
Looks fine to me!

PR #2630, opened 2022-11-17
Fix setting of language cookie.

EreMaijala:
- Overrides the cookie strategy to avoid it setting a cookie with wrong parameters.
- Only sets the language cookie if it doesn't already exist or match the detected language.


demiankatz (approved):
Thanks, @EreMaijala -- I've backported this to the release-8.1 branch and merged it from there to dev, so we can include it in our next bug fix release.

PR #2632, opened 2022-11-21
Support for filters in SolrPrefix autocomplete

xrosecky:


module/VuFind/src/VuFind/Autocomplete/SolrPrefix.php
```diff
@@ -121,6 +128,9 @@ public function getSuggestions($query)
             $options = $params->getOptions();
             $options->disableHighlighting();
             $options->spellcheckEnabled(false);
+            foreach ($this->filters as $current) {
+                $this->searchObject->getParams()->addFilter($current);
```
demiankatz: Would it make more sense to add these directly to the $params object a few lines earlier, since we have direct access in this method?

demiankatz (changes requested):
Thanks for this -- one small suggestion for improvement below.

Also, a question: what mechanism causes the filters to get added in the first place? Do we need to add a config setting, etc.?

demiankatz (approved):
Thanks, @xrosecky -- I see now that this is part of standard "duck typing" in the VuFind\Autocomplete\Suggester.

PR #2633, opened 2022-11-23
Add ability to get cover using multiple ISBNs

xmorave2:


module/VuFind/src/VuFind/Cover/Loader.php
```diff
@@ -101,6 +101,13 @@ class Loader extends \VuFind\ImageLoader
      */
     protected $isbn = null;
 
+    /**
+     * User ISBNs parameter
+     *
+     * @var array[ISBN]
```
demiankatz: Would the syntax for this be `ISBN[]`?

module/VuFind/src/VuFind/Cover/Loader.php
```diff
@@ -302,6 +310,12 @@ protected function storeSanitizedSettings($settings)
     {
         $settings = array_merge($this->getDefaultSettings(), $settings);
         $this->isbn = new ISBN($settings['isbn'] ?? '');
+        $this->isbns = array_map(
```
demiankatz, line 304: Should we array_filter this to ensure validity?

module/VuFind/src/VuFind/Cover/Loader.php
```diff
@@ -438,6 +453,9 @@ protected function getIdentifiers()
         if ($this->isbn && $this->isbn->isValid()) {
             $ids['isbn'] = $this->isbn;
         }
+        if (!empty($this->isbns)) {
```
demiankatz, line 444: Do we actually need a separate parameter for plural isbns vs singular isbn? Would it be better to just use the existing isbn parameter but add extra logic to handle the possibility of it being either a single value or an array? That would reduce redundancy in URLs and code.

demiankatz: (I realize that we probably need separate values to get passed into the actual plugins for backward compatibility... but the rest of the "plumbing" in between could probably be adjusted to work with arrays and then pass the first value as the singular instead of keeping those things separate the whole way through).

demiankatz (changes requested):
Thanks, @xmorave2 -- a few questions and suggestions.

module/VuFind/src/VuFind/Cover/Loader.php
xmorave2: This simplification dose make sense to. I reduce code a bit in that way.

module/VuFind/src/VuFind/Cover/Loader.php
xmorave2: isValid is called in getCleanISBNs in record driver, so there should not be any invalid isbn here

module/VuFind/src/VuFind/Cover/Loader.php
xmorave2: fixed

module/VuFind/src/VuFind/Content/ObalkyKnihService.php
```diff
@@ -176,7 +176,15 @@ protected function getFromService(array $ids): ?\stdClass
...
-        $isbn = isset($ids['isbn']) ? $ids['isbn']->get13() : null;
+        $isbn = null;
+        if (!empty($ids['isbns'])) {
+            $isbn = array_map(
+                function ($isbn) {
+                    return $isbn->get13();
+                },
+                $ids['isbns']
+            );
+        }
```
demiankatz, line 189: Is it worth adding an `elseif (!empty($ids['isbn'])) {` case? I realize that that should never happen post-refactoring, but it might be useful to maintain that part of the API in case this plugin is ever being used in a different context in custom code.

module/VuFind/src/VuFind/Cover/Loader.php
demiankatz: Right!

module/VuFind/src/VuFind/Cover/Loader.php
```diff
@@ -301,7 +301,12 @@ protected function getImageSettingsFromLegacyArgs($args)
     protected function storeSanitizedSettings($settings)
     {
         $settings = array_merge($this->getDefaultSettings(), $settings);
-        $this->isbn = new ISBN($settings['isbn'] ?? '');
+        $this->isbns = array_map(
+            function ($isbn) {
+                return new ISBN($isbn);
+            },
+            $settings['isbns'] ?? []
```
demiankatz: Similarly, for backward compatibility, maybe we should support the possibility of the legacy 'isbn' value still being used, like:

`$settings['isbns'] ?? (empty($settings['isbn']) ? [] : (array)$settings['isbn'])`

...just in case a custom record driver somewhere is still using the old setting.

module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
```diff
@@ -282,6 +282,28 @@ public function getCleanISBN()
         return $isbn13;
     }
 
+    /**
+     * Return all valid ISBNs found in the record.
+     *
+     * @return array
+     */
+    public function getCleanISBNs(): array
```
demiankatz: Would it make sense to include a flag or option array to potentially allow the caller to filter for ISBN-10 only, or to normalize to ISBN-13? If you don't think this would be useful, I don't insist on it... just thinking about potentially useful features while we're designing something new.

demiankatz (commented):
Thanks for the progress on this, @xmorave2 -- in reviewing again, I had a couple of thoughts about improving backward compatibility and potentially adding useful future functionality.

module/VuFind/src/VuFind/Content/ObalkyKnihService.php
xmorave2: I've done it.

module/VuFind/src/VuFind/Cover/Loader.php
xmorave2: Done. Thanks for suggestions.

module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
xmorave2: I tried to add the flags array. Is that what you have in mind?

module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
```diff
@@ -282,6 +282,40 @@ public function getCleanISBN()
...
+     * Return all valid ISBNs found in the record.
+     *
+     * @param array $flags Flags to prefer ISBN-10 or ISBN-13. If 'only10' is true,
+     * only ISBN-10s will be returned. If 'normalize13' is true, the ISBN is always
+     * converted to ISBN-13. If both are false (default), then each ISBN is returned
+     * as found in record.
+     *
+     * @return array
+     */
+    public function getCleanISBNs(
```
demiankatz, line 295: Thanks, @xmorave2! I had a few ideas about additional potentially useful flags and some refactoring to ensure that flags are never unset. Here's my proposed (but currently untested) revision:

```
    public function getCleanISBNs(array $userFlags = []): array
    {
        $defaultFlags = [
             'only10' =>  false,
             'prefer10' => true,
             'filterInvalid' => true,
             'normalize13' => false
        ];
        $flags = $userFlags + $defaultFlags;
        $isbns = $this->getISBNs();
        $all = $tens = $thirteens = $invalid = [];
        foreach ($isbns as $isbn) {
            // Strip off any unwanted notes:
            if ($pos = strpos($isbn, ' ')) {
                $isbn = substr($isbn, 0, $pos);
            }
            $isbnObj = new ISBN($isbn);
            if ($isbnObj->isValid()) {
                if ($isbn10 = $isbnObj->get10()) {
                    $normalized = $flags['normalize13'] ? $isbnObj->get13() : $isbn10;
                    $tens[] = $normalized;
                    $all[] = $normalized;
                } elseif ($isbn13 = $isbnObj->get13()) {
                    $thirteens[] = $isbn13;
                    $all[] = $isbn13;
                } elseif (!$flags['filterInvalid']) {
                    $invalid[] = $isbn;
                    $all[] = $isbn;
                }
            }
        }
        return $flags['prefer10'] ? array_merge($tens, $thirteens, $invalid) : $all;
    }
```

What do you think? This adds the "prefer 10" flag set to true by default so that we can retain backward compatibility with the getCleanISBN logic which favors ISBN10-compatible values in case some APIs are not 13-compatible (probably not a major concern at this point in time, but a bit of compatibility we might as well maintain if we easily can).

Because of all the possible code paths through this, we should add some unit tests to cover this logic. I'm happy to write something next week when I'm back from vacation if that would be helpful.

xmorave2: I tried to refactor it this way. 
What do you think about it now?

module/VuFind/tests/fixtures/misc/testbug2.json
```diff
@@ -59,7 +59,7 @@
           "Vico, Giambattista, 1668-1744.",
           "La congiura dei Principi Napoletani 1701 : (prima e seconda stesura) /"],
         "isbn":[
-          "8820737493"],
+          "8820737493", "9788072815562"],
```
demiankatz: Might be worth also adding a 979 ISBN (one that can't be converted to ISBN-10, e.g. 9791032917633) to the test fixture so that some of the flags can be more thoroughly tested. I'd recommend making the 979 ISBN first, so we can ensure that prefer10 is actually doing something.

module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
```diff
@@ -281,6 +281,54 @@ public function getCleanISBN()
...
 
+    /**
+     * Return all ISBNs found in the record.
+     *
+     * @param array $userFlags Flags to control behavior:
+     *  - 'only10' returns only ISBN-10s
+     *  - 'prefer10' returns ISBN-10s if available, otherwise ISBN-13s (default)
+     *  - 'normalize13' returns ISBN-13s, normalizing ISBN-10s to ISBN-13s
+     *  - 'filterInvalid' filters out invalid ISBNs
+     * Only one of 'only10', 'prefer10', or 'normalize13' should be set to true.
```
demiankatz: Should we add some logic to enforce this, like:

```
$exclusiveFlags = ($flags['only10'] ? 1 : 0) + ($flags['prefer10'] ? 1 : 0) + ($flags['normalize13'] ? 1 : 0);
if ($exclusiveFlags > 1) {
    throw new \Exception('Invalid combination of flags provided.');
}
````

module/VuFind/tests/fixtures/misc/testbug2.json
demiankatz: Adding an invalid ISBN (like "invalid-isbn!") would also be a good idea, so we can ensure that filterInvalid is working.

demiankatz (changes requested):
Thanks, @xmorave2, this looks good to me. I just have a couple of small suggestions, mainly to ensure that the test coverage is as thorough as possible.

module/VuFind/tests/fixtures/misc/testbug2.json
xmorave2: I have added 979 ISBN and invalid isbn and updated the tests to test all variant more thoughtfully

module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
xmorave2: I refactored the parameters a bit - now there is no flags array but two params - mode (only10, prefer10, normalize13) and one boolean $filterInvalid

demiankatz (approved):
Thanks, @xmorave2, these changes make sense to me, and the test looks very thorough now!

PR #2634, opened 2022-11-23
Fix processing spelling suggestions in Summon

xmorave2:


demiankatz:
Thanks, @xmorave2 -- I'm no longer in a position to test this, though I assume you're submitting this because it works for you. Did the API change, or do you think this bug was introduced by refactoring?

xmorave2:
@demiankatz: Yes, we have currently trial access to Summon and this is working for me.

Looking into git history I don't think this is because of refactoring. I bet the API has changed. I checked the raw API response and really has this structure.

demiankatz (approved):

demiankatz:
Thanks, @xmorave2, I've backported this to the release-8.1 branch so we can release the fix sooner.

PR #2635, opened 2022-11-24
Improve normalization of parentheses for Solr queries.

EreMaijala:
Leaves quoted content unchanged.

This allows one to search for weird titles such as [(((](https://www.finna.fi/Record/samk.991457813605968?lng=en-gb) (ISBN 1-953035-71-X) by targeting a string field (or any other field that does not disregard parentheses) in the index, e.g. `title_sort:"((("`.

### DIFF
--- a/module/VuFindSearch/src/VuFindSearch/Backend/Solr/LuceneSyntaxHelper.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/Solr/LuceneSyntaxHelper.php
@@ -282,7 +282,6 @@ class LuceneSyntaxHelper
     public function extractSearchTerms($query)
     {
         $result = [];
-        $inQuotes = false;
         $collected = '';
         $discardParens = 0;
         // Discard local parameters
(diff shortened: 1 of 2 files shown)


module/VuFindSearch/src/VuFindSearch/Backend/Solr/LuceneSyntaxHelper.php
```diff
@@ -666,4 +665,63 @@ protected function capitalizeRangesCallback($match)
...
+     */
+    protected function removeNonQuoted(array $needles, string $haystack): string
+    {
+        $result = '';
+        $inQuotes = false;
+        $lastCh = '';
+        $lastLastCh = '';
+        foreach (str_split($haystack) as $ch) {
+            // Check for non-escaped quote:
+            if ('"' === $ch && ('\\' !== $lastCh || '\\' === $lastLastCh)) {
```
demiankatz: Is there a reason that the logic is a little bit different between countNonQuoted and removeNonQuoted? If not, I'm wondering if it might actually make sense to create a protected helper method so the non-quoted logic is only maintained in one place, and the rest of the work could be done with a callback... e.g., something like:

```
    protected function countNonQuoted(string $needle, string $haystack): int
    {
        $result = 0;
        $callback = function ($ch) use ($result) {
            $result++;
        };
        $this->applyCallbackToNonQuotedChars([$needle], $haystack, $callback);
        return $result;
    }

    protected function removeNonQuoted(array $needles, string $haystack): string
    {
        $result = '';
        $callback = function ($ch) use ($result) {
            $result .= $ch;
        };
        $this->applyCallbackToNonQuotedChars($needles, $haystack, $callback);
        return $result;
    }

    protected function applyCallbackToNonQuotedChars(array $needles, string $haystack, callable $callback): void
    {
        $inQuotes = false;
        $lastCh = '';
        $lastLastCh = '';
        foreach (str_split($haystack) as $ch) {
            // Check for non-escaped quote:
            if ('"' === $ch && ('\\' !== $lastCh || '\\' === $lastLastCh)) {
                $inQuotes = !$inQuotes;
            }
            if (!$inQuotes && !in_array($ch, $needles)) {
                $callback($ch);
            }
            $lastLastCh = $lastCh;
            $lastCh = $ch;
        }
    }
```

Too much abstraction? Maybe... but just trying to avoid replication of somewhat complex logic.

demiankatz (commented):
Thanks, @EreMaijala -- one question/suggestion for your consideration!

module/VuFindSearch/src/VuFindSearch/Backend/Solr/LuceneSyntaxHelper.php
EreMaijala: Oops, my mistake. I'll try to find a way to avoid duplication. Also the escape handling is still not quite correct.

demiankatz (approved):
Thanks, @EreMaijala, this is tremendously improved -- a great application of functional programming. I've done a little more work to expand/modernize the associated test suite (basically finishing some of the work that you started, while we were already in the area anyway). I'll merge now!

PR #2636, opened 2022-11-25
Clean up header login/logout links and fix issues.

EreMaijala:
Only relevant links are now included in header html. This uncovered issues in our tests, and fixing the tests uncovered issues with lightbox referrer handling for multi-step flows. The original referrer was lost when there were many steps in a lightbox, but the tests still worked because they used the hidden '.logoutOptions .logout' link to log out, but e.g. the following flow ended up with a broken lightbox:

1. Enable record rating, Database login and account creation
2. start a new session (not logged in)
3. Navigate to a record
4. Start adding a rating by clicking the stars
5. Click the "You must be logged in first" button
6. Click the "Create New Account" button
7. Enter user information and click Submit
8. If catalog login is enabled, observe prompt for Library Catalog Profile and enter credentials
9. Observe a broken lightbox

### DIFF
--- a/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php
+++ b/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php
@@ -154,6 +154,12 @@ final class HoldsTest extends \VuFindTest\Integration\MinkTestCase
         $this->fillInAccountForm($page);
         $this->clickCss($page, 'input.btn.btn-primary');
 
+        // Start establishing library catalog profile
+        $this->waitForPageLoad($page);
+        $element = $this->findCss($page, '.alert.alert-info a');
+        $this->assertEquals('Library Catalog Profile', $element->getText());
+        $element->click();
+
         // Test invalid patron login
         $this->submitCatalogLoginForm($page, 'bad', 'incorrect');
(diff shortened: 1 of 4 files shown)


demiankatz:
I'm interested in hearing @crhallberg's review of this one since it touches the lightbox.

I thought the purpose of the hidden login/logout links was so that the lightbox could update the header when the user logs in during a lightbox flow. Does this break that behavior? Is it possible that behavior was already broken? (I notice that it _is_ working on vufind.org/demo, though the header update doesn't happen until the lightbox actually closes).

EreMaijala:
@demiankatz I have a faint recollection that it could have been the case that lightbox does the hiding/unhiding, but I can't find anything in the codebase to support that. Perhaps it was refactored out at some point.

demiankatz:
@EreMaijala, maybe that logic got simplified into a full page refresh at some point, in which case this shouldn't be a problem. @crhallberg might remember better. In any case, I'll look again when I return to the office tomorrow.

crhallberg:
The dynamic hiding/unhiding of login/logout was refactored in favor of refreshing the whole page when the user logged in or out.

crhallberg (approved):

demiankatz:
Thanks, @EreMaijala and @crhallberg!

PR #2637, opened 2022-11-29
Add Google Tag Manager support

maccabeelevine:
Google Tag Manager is a useful way of deploying multiple tags without individually deploying each in the app's JS.  For VuFind it's particularly useful to deploy both out-of-the-box Google Analytics tracking as well as additional tracking for user events, as [done here for Summon](https://journal.code4lib.org/articles/14853).

This code is patterned after the existing GoogleAnalytics view helper.  It uses the nonce-aware version of the GTM script.


maccabeelevine:
I'm not clear how to indicate the necessary configuration in config.ini.  It's a new section with one property:

```
[GoogleTagManager]
propertyId       = "GTM-XXXXXXX"
```

Also I know that a unit test is needed, but I was hoping for a general review first since this is my first VuFind PR.

module/VuFind/src/VuFind/View/Helper/Root/GoogleTagManager.php
```diff
@@ -0,0 +1,93 @@
+<?php
+/**
+ * GoogleTagManager view helper
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2010.
```
demiankatz: Might want to use 2022 here.

module/VuFind/src/VuFind/View/Helper/Root/GoogleTagManager.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Returns GA code (if active) or empty string if not.
+     *
+     * @param string $customUrl override URL to send to Google Analytics
+     *
+     * @return string
+     */
+    public function __invoke($isHeadCode = false)
```
demiankatz: Would it be more clear, rather than creating an __invoke method with a switch, to just have `getHeaderCode()` and `getFooterCode()` methods? If you leave the default __invoke from the parent class, its behavior is simply to return the helper, so then your template code would be `<?=$this->googleTagManager()->getHeaderCode()?>`, etc.

I don't insist on the change if you prefer this approach, but I feel that it might be more readable than a non-obvious boolean value.

demiankatz (changes requested):
Thanks, @maccabeelevine, this is off to a great start!

Regarding config.ini, I'd suggest looking at the existing [GoogleAnalytics] section and adding a similar commented out [GoogleTagManager] section below it.

It might also be a good idea to add some documentation to the [relevant wiki page](https://vufind.org/wiki/configuration:usage_stats) when we get closer to merging this. Would you be interested in doing that?

module/VuFind/src/VuFind/View/Helper/Root/GoogleTagManager.php
maccabeelevine: Yes thanks, will correct.  How does IP work for VuFind -- do all contributors assign copyright to Villanova?

module/VuFind/src/VuFind/View/Helper/Root/GoogleTagManager.php
maccabeelevine: Your approach is much better.  I wasn't clear on the invoke() behavior.

maccabeelevine:
Ok the changes above are addressed.  I'd be happy to edit the wiki page as well.  It's a little odd because GTM is not a simple alternative to GA or Matomo, since it's a higher-level tool that can use both, so in theory it's not directly about "[Collecting Usage Stats](https://vufind.org/wiki/configuration:usage_stats)".  However in practice that's what it's most used for and certainly what I intend to use it for.  So that's the page I should edit.  Just maybe a slight page reorg as part of it.

I still have to write up a unit test.

maccabeelevine:
And I'll address the unit test failure on code style.

module/VuFind/src/VuFind/View/Helper/Root/GoogleTagManager.php
demiankatz: The copyright does not have to be assigned to Villanova; as long as it's GPL'ed, we're not overly concerned whose name is on it. :-)

demiankatz:
Thanks, @maccabeelevine, this is looking pretty good to me! A few remaining comments:

1.) Should we do any work to document how to make this work in combination with a content security policy? I suspect at least parts of it will get blocked if you enable VuFind's default. Maybe we should add some notes to the wiki when we get to that part, and then link to the wiki page from config.ini, or something like that.

2.) I agree that GTM isn't exactly the same as GA or Matomo, but since it can be used for the same purpose, it's appropriate to document it on the usage stats page. We can always link to the notes from elsewhere if other use cases arise.

3.) Regarding the style issues, some are very straightforward, but others are related to line length in the code snippets. You might be able to fix them by simply de-indenting the contents of the heredoc block, but if that doesn't work (or you don't want to do that), there are also annotations you can use to skip checks for blocks of code, and I think it would be appropriate to use those here if a more elegant option does not present itself.

4.) I think we're close enough to the final design that you can safely start to build a unit test. I imagine the output of the helper might change a little bit if we have to adjust markup for CSP, etc., but the shape of the logic and associated test shouldn't.

demiankatz:
@maccabeelevine, the current CI failures appear to be caused by trailing whitespace on some of the lines. You can automatically clean this up by running `vendor/bin/phing php-cs-fixer` in your instance... but you might also want to consider setting your editor to auto-trim trailing whitespace for you.

maccabeelevine:
Hi @demiankatz, I think this is finally ready.

> 1.) Should we do any work to document how to make this work in combination with a content security policy? I suspect at least parts of it will get blocked if you enable VuFind's default. Maybe we should add some notes to the wiki when we get to that part, and then link to the wiki page from config.ini, or something like that.

So, I did use the CSP-aware version of the GTM head code.  It loads the code snippet either way and does send through GTM events.  (My GTM is setup itself to pass data along to GA, which then fails without modifying connect-src to allow it.  But that's out of scope for GTM.)

> 2.) I agree that GTM isn't exactly the same as GA or Matomo, but since it can be used for the same purpose, it's appropriate to document it on the usage stats page. We can always link to the notes from elsewhere if other use cases arise.

Ok, I can add something to the wiki.  Just let me know the appropriate timing, i.e. how to indicate the feature won't be available until release X.

> 3.) Regarding the style issues, some are very straightforward, but others are related to line length in the code snippets. You might be able to fix them by simply de-indenting the contents of the heredoc block, but if that doesn't work (or you don't want to do that), there are also annotations you can use to skip checks for blocks of code, and I think it would be appropriate to use those here if a more elegant option does not present itself.

These are resolved.

> 4.) I think we're close enough to the final design that you can safely start to build a unit test. I imagine the output of the helper might change a little bit if we have to adjust markup for CSP, etc., but the shape of the logic and associated test shouldn't.

Done.

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/GoogleTagManagerTest.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Test the helper (disabled mode)
+     *
+     * @return void
+     */
+    public function testDisabled(): void
+    {
+        $this->assertEquals('', $this->renderGTMHeadCode(false));
```
demiankatz, line 76: Should we also assert an empty return from renderGTMBodyCode here?

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/GoogleTagManagerTest.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Render the GTM Head element code
+     *
+     * @param string     $gtmContainerId     GTM Container ID (false for disabled)
```
demiankatz: There's some extra whitespace in this param comment -- should only be one space between each section (same issue on the next function below). Not very important, but might as well fix it while you're in here anyway. :-)

demiankatz (changes requested):
@maccabeelevine, this looks really good! I just had a couple of fairly trivial suggestions about the unit test, and then I believe this is ready to merge.

Next steps/questions:

1.) You can add a Google Tag Manager section to [this wiki page](https://vufind.org/wiki/configuration:usage_stats) whenever you like. You can include an "available starting with 9.0" message modeled on the 8.0 message found in the existing Matomo section. If you feel this is not the most appropriate page, I'm also happy to discuss other alternatives.

2.) Do you have any thoughts on what we should do with #2402? Is it worth providing direct GA4 support in place of existing (and soon to be useless) legacy GA support? Or should we deprecate the whole thing and point people toward using the tag manager? I don't have really strong opinions, but I'd love to get this resolved one way or another in time for the 9.0 release.

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/GoogleTagManagerTest.php
maccabeelevine: Absolutely.  Fixed.

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/GoogleTagManagerTest.php
maccabeelevine: You know I did that on purpose as a copy-paste from GoogleAnalyticsTest, but I see it only made sense to line up with the second param ($options) which is not needed for GTM.  Oops.  Also fixed.

maccabeelevine:
> 1.) You can add a Google Tag Manager section to [this wiki page](https://vufind.org/wiki/configuration:usage_stats) whenever you like. You can include an "available starting with 9.0" message modeled on the 8.0 message found in the existing Matomo section. If you feel this is not the most appropriate page, I'm also happy to discuss other alternatives.


Ok, will do that.  I'm not sure it's the ideal page but better to get the info down somewhere and then we can always reorganize right?

> 2.) Do you have any thoughts on what we should do with https://github.com/vufind-org/vufind/pull/2402? Is it worth providing direct GA4 support in place of existing (and soon to be useless) legacy GA support? Or should we deprecate the whole thing and point people toward using the tag manager? I don't have really strong opinions, but I'd love to get this resolved one way or another in time for the 9.0 release.

Yeah this is a good question and I can see it either way.  GA4 is still a standalone product, same as it was in V3 (Universal Analytics), so there's no good reason to drop support, except for the work involved in building the GA4 version.  I would argue that anyone using GA should also be using GTM as it just makes GA easier, and you can add additional tracking (whether GA-related tracking or whatever else) much more easily.  However in theory someone could have a larger privacy concern with GTM than they do with GA.  So in principle VuFind should still support GA4 but I'd suggest finding at least one person who cares (vs GTM) before anyone takes the time.  I should also mention that I intend to share my custom user VuFind event tracking as a GTM library for others to use if they like.  So all the more reason to dump standalone GA tags.

To be clear I would definitely either deprecate/remove or replace with GA4 code.  No point in leaving what's there for GA3 and prior.

demiankatz (approved):
Thanks, @maccabeelevine -- merging now!

demiankatz:
I agree that getting the documentation somewhere first, and moving it later as needed, makes sense. Let me know if I can do anything more to help with the wiki side of things.

Your assessment of GA4 also makes sense to me. I'll talk with the local team at VU and see if there are any strong opinions here.

PR #2638, opened 2022-11-30
Set RSS feed type to avoid PHP deprecation notices.

EreMaijala:
Feed type doesn't have a default, but gets used with strtolower.


demiankatz (approved):
Thanks, @EreMaijala.

PR #2639, opened 2022-11-30
Fix PHP 8 deprecation notice from getSessionInitiator call.

EreMaijala:
The target parameter should be a string.


demiankatz (approved):
Thanks, @EreMaijala.

PR #2640, opened 2022-11-30
Add critical Upgrade step for blowfish deprecation.

crhallberg:
Adds an explanatory step to the UpgradeController and an error check for blowfish at the most likely point of failure.

The [Blowfish encryption algorithm](https://en.wikipedia.org/wiki/Blowfish_(cipher)) is vulnerable to certain attacks, particularly in contexts like HTTPS. [Versions of OpenSSL starting with 3.0 will no longer support blowfish by default](https://wiki.openssl.org/index.php/OpenSSL_3.0#Providers), which can break your VuFind installation. You can [read more about this issue on the VuFind JIRA ticket](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1563).


module/VuFind/sql/migrations/pgsql/9.0/006-add-user-cypher-method.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "user"
+  ADD COLUMN `cypher_method` varchar(50) DEFAULT -1; -- -1 == undefined
```
demiankatz: Can we delete this file?

themes/bootstrap3/templates/upgrade/criticalfixblowfish.phtml
```diff
@@ -0,0 +1,93 @@
...
+activate = 1</pre>
+		</div>
+	</div>
+<?php endif; ?>
+
+<h3 id="conversion">2. Convert existing data to a new encryption method</h3>
+
+<p>You can use VuFind's CLI tool <kbd>switch_db_hash</kbd> to convert all passwords from blowfish to another encryption method. Below, you can find a pre-filled example command that will convert your encryption to AES-256 (the new default VuFind encryption method) with a random 32 character string. You can use <kbd>openssl_get_cipher_methods()</kbd> to see all encryption methods available on your system.</p>
+
+<pre id="example-cmd">php public/index.php util switch_db_hash blowfish:<?=$this->oldkey ?> aes-256:<?=$this->exampleKey ?></pre>
```
demiankatz: This is not the right syntax -- parameter 1 = new encryption method, parameter 2 = new key. The old details are automatically loaded from the configuration.

Also, I had to enclose the new key in single quotes, because otherwise if the randomly generated key includes exclamation points, it causes parsing errors for the Unix shell.

The "aes-256" algorithm didn't work for me -- I got "The algorithm aes-256 is not supported by Laminas\Crypt\Symmetric\Openssl" under Ubuntu 20 (I haven't tested on Ubuntu 22 yet). Using plain "aes" without a number did work for me. Is this simply an error that needs to be corrected, or do we potentially have a problem where older versions of Linux won't support newer encryption algorithms?

It's probably also a good idea to put `$VUFIND_HOME/` in front of `public/index.php` to make the command work in more contexts.

config/vufind/config.ini
```diff
@@ -536,7 +538,7 @@ ils_encryption_key = false
 ; Common choices: blowfish (default), aes
 ; If you want to convert from one algorithm to another, run this from $VUFIND_HOME:
 ;   php public/index.php util switch_db_hash oldhash:oldkey (or none) newhash:newkey
-;ils_encryption_algo = "blowfish"
+;ils_encryption_algo = "aes-256"
```
demiankatz: As discussed elsewhere, aes-256 does not seem to work in Ubuntu 20, so we may need to correct this (if aes-256 is simply an invalid value) or add notes (if different situations will require different approaches).

themes/bootstrap3/templates/upgrade/criticalfixblowfish.phtml
```diff
@@ -0,0 +1,93 @@
...
+		<div class="panel-body">
+
+			<p>Your system's version of OpenSSL may not configured to provide the Blowfish algorithm. You may need to <a href="https://wiki.openssl.org/index.php/OpenSSL_3.0#Providers">follow these steps to enable the legacy provider in OpenSSL</a>.
+
+			<ol>
+				<li>
+					<p><b>Identify the location of your OpenSSL configuration</b></p>
+
+					<pre id="find-openssl">php -i | grep "Openssl default config"</pre>
+					<?=$this->copyToClipboardButton('find-openss') ?>
```
demiankatz: This button didn't work for me -- I think you're missing an 'l' on the end of the ID.

themes/bootstrap3/templates/upgrade/criticalfixblowfish.phtml
```diff
@@ -0,0 +1,93 @@
...
+					<p><b>Add a new section</b> somewhere below:</p>
+
+					<pre>
+[legacy_sect]
+activate = 1</pre>
+				</li>
+
+				<li><b>Restart Apache</b></li>
+
+				<li>Refresh this page or <a href="#conversion">move onto the conversion step below</a>.</li>
```
demiankatz: Nitpick: I'd change "onto" to "on to".

module/VuFind/src/VuFind/Controller/UpgradeController.php
```diff
@@ -849,6 +851,12 @@ public function homeAction()
             return $this->forwardTo('Upgrade', 'EstablishVersions');
         }
 
+        // Check for critical upgrades
+        $forward = $this->criticalCheckForBlowfishEncryption();
```
demiankatz: Rather than directly calling this from homeAction, maybe it would be worth wrapping it in a `performCriticalChecks` method, so if we have additional critical checks in the future, there's an obvious place to add them -- it keeps the homeAction a bit more generic. Of course, since there's only one check at this point, we could also choose to do that refactoring later. I don't have strong feelings about what to do now, I just think that's the eventual place we should land if we end up with multiple critical checks.

module/VuFind/src/VuFind/Controller/UpgradeController.php
```diff
@@ -938,4 +946,53 @@ protected function fixshortlinks()
...
+            $newcipher = new BlockCipher(new Openssl(['algorithm' => 'blowfish']));
+            $newcipher->setKey('akeyforatest');
+            $newcipher->encrypt('youfoundtheeasteregg!');
+        } catch (Exception $e) {
+            $blowfishIsWorking = false;
+        }
+
+        // Get key for example command
+        $config = $this->getConfig();
+        $oldkey = $config->Authentication->ils_encryption_key;
```
demiankatz: As noted in the template, I don't think we actually need to expose the old key, which is a good thing, because sending the real current key to the web interface under any circumstances is pretty dangerous (in case somebody forgets to disable auto-config).

module/VuFind/src/VuFind/Db/Row/User.php
```diff
@@ -232,21 +232,38 @@ protected function encryptOrDecrypt($text, $encrypt = true)
...
         }
 
+        $configAuth = $this->config->Authentication;
+
         // Load encryption key from configuration if not already present:
-        if (null === $this->encryptionKey) {
-            if (!isset($this->config->Authentication->ils_encryption_key)
-                || empty($this->config->Authentication->ils_encryption_key)
+        if ($this->encryptionKey === null) {
+            if (!isset($configAuth->ils_encryption_key)
```
demiankatz: `empty` includes a built in `isset` check and does not trigger notices for missing values, so this code can be simplified to only make the empty check.

module/VuFind/src/VuFind/Db/Row/User.php
```diff
@@ -232,21 +232,38 @@ protected function encryptOrDecrypt($text, $encrypt = true)
...
             }
-            $this->encryptionKey = $this->config->Authentication->ils_encryption_key;
+
+            $this->encryptionKey = $configAuth->ils_encryption_key;
         }
 
         // Perform encryption:
-        $algo = $this->config->Authentication->ils_encryption_algo ?? 'blowfish';
-        $cipher = new BlockCipher(new Openssl(['algorithm' => $algo]));
+        $algo = $configAuth->ils_encryption_algo ?? 'aes-256';
```
demiankatz: I wonder if we should define the default encryption algorithm as a constant somewhere (maybe even within this class) so it exists in only one place. Maybe we should also define a key length constant, to help with key generation in a generic way. I think we need to do some work on InstallController::fixSecurityConfiguration, since I don't think it's going to set things up correctly with the new default.

demiankatz: Also, upon further reflection, I think we have to leave the default at 'blowfish' here -- otherwise, by upgrading the code, you'll break legacy environments that use blowfish encryption. I think this is the reason we had talked about storing the algorithm in the database... but I think an easier solution is to simply make specifying the algorithm in the config file mandatory going forward. For now, we'll keep the blowfish default. Maybe starting in VuFind 10, we can throw an exception if the algorithm is unspecified in the configuration.

demiankatz (changes requested):
Thanks, @crhallberg, this is a great start, though I ran into a few issues. In addition to the notes below, some general comments:

1.) For some reason, this PR originally showed a bunch of changes that were already part of dev. I merged the latest dev into your branch and pushed it up to sort things out.

2.) It's very misleading that we have two separate security functions: password hashing and ILS credential encryption -- and yet our command line utility to change the encryption algorithm has the word "hash" in it. I got very confused as a result of this. I wonder if we should consider renaming that command for clarity (though of course we'd have to retain an alias for backward compatibility purposes).

3.) I mentioned this in passing below, but just to draw attention to it: I think we need to do some work on the InstallController::fixSecurityConfiguration method to bring it up to date with the new default.

4.) I've retitled [VUFIND-1563](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1563) and edited the description to make it more immediately useful to people who follow the reference link there. We may need to edit it further to reflect the aes vs. aes-256 problems discussed below.

config/vufind/config.ini
```diff
@@ -528,15 +528,17 @@ encrypt_ils_password = false
 
 ; This is the key used to encrypt and decrypt catalog passwords.  This must be
 ; filled in with a random string value when encrypt_ils_passwords is set to true.
+; aes-256 (default) requires a key exactly 32 characters long.
```
demiankatz: Do we need to change this instance of aes-256 as well?

module/VuFind/src/VuFind/Controller/UpgradeController.php
```diff
@@ -821,6 +823,20 @@ public function getsourceversionAction()
...
 
+    /**
+     * Organize and run critical, blocking checks
+     *
+     * @return string|null
+     */
+    protected function performCriticalChecks()
+    {
+        // Prevent the use of the deprecated blowfish cipher algo
+        $forward = $this->criticalCheckForBlowfishEncryption();
```
demiankatz: I think performCriticalChecks should just `return $this->criticalCheckForBlowfishEncryption();` and then the calling code can do the null check and forward. This way, if we get more critical checks, we can just chain them here with `??`. This change would address the fact that right now, I don't think the return value from this method is getting handled correctly.

themes/bootstrap3/templates/upgrade/criticalfixblowfish.phtml
```diff
@@ -0,0 +1,93 @@
...
+activate = 1</pre>
+		</div>
+	</div>
+<?php endif; ?>
+
+<h3 id="conversion">2. Convert existing data to a new encryption method</h3>
+
+<p>You can use VuFind's CLI tool <kbd>switch_db_hash</kbd> to convert all passwords from blowfish to another encryption method. Below, you can find a pre-filled example command that will convert your encryption to AES (the new default VuFind encryption method) with a random 32 character string. You can use <kbd>openssl_get_cipher_methods()</kbd> to see all encryption methods available on your system.</p>
+
+<pre id="example-cmd">php public/index.php util switch_db_hash aes:<?=$this->exampleKey ?></pre>
```
demiankatz: I think you need to remove the colon after aes in this line.

demiankatz (changes requested):
A bit of feedback on your latest changes, but it looks like there are also some previous comments that haven't been fully addressed yet. No rush, I'm just trying to keep on top of things as they come. :-)

config/vufind/config.ini
```diff
@@ -528,15 +528,17 @@ encrypt_ils_password = false
...
 
 ; This is the algorithm used to encrypt and decrypt catalog passwords.
 ; A symmetrical encryption algorithm must be used.
 ; You can use openssl_get_cipher_methods() to see available options on your system.
-; Common choices: blowfish (default), aes
+; Common choices: aes (default), cast, sm4
 ; If you want to convert from one algorithm to another, run this from $VUFIND_HOME:
 ;   php public/index.php util switch_db_hash oldhash:oldkey (or none) newhash:newkey
-;ils_encryption_algo = "blowfish"
+;ils_encryption_algo = "aes"
```
demiankatz: As noted earlier, I think we need to keep blowfish as the default fallback for the VuFind\Db\Row\User class, since otherwise we'll misinterpret legacy configuration files and try to apply the wrong algorithm. Instead, maybe what we need to do is uncomment the aes line by default in the configuration file, to enforce a reasonable default for new installations. Maybe we can change (default) in the comment to (recommended).

demiankatz (changes requested):
Thanks for the ongoing progress, @crhallberg. I've resolved all the previous comments that have been addressed, but I think there are a couple of minor points that are still outstanding. Can you take another look when time permits?

config/vufind/config.ini
crhallberg: That's a good point.

crhallberg: Although it is disabled by default... wouldn't we want to encourage good practices for people who haven't messed with the default (disabled) config?

demiankatz: My thought is that at some point, we should throw an exception if the algorithm is undefined. So maybe in VuFind 9.0, we issue a deprecation warning, like "ils_encryption_algo undefined; defaulting to blowfish" and then in VuFind 10.0, it's a fatal error to turn on encryption without specifying an algorithm.

crhallberg: Do we want the deprecation warning in this PR?

demiankatz: Probably wouldn't hurt!

demiankatz: @crhallberg, I think this detail still needs to be done. Can you double-check?

crhallberg: Forgot to push a change but I'm completely confident this is the best way to display a warning.

demiankatz:
@crhallberg, please take a look at my recent commits here and let me know what you think! I did some hands-on testing of this, and ended up making several changes/additions:

1.) The added code in the admin module did not work -- the admin module doesn't support flash messages, and the way the flash message was being constructed was using the wrong URL syntax. Rather than fixing it, I removed it, because I don't think many people ever look at the admin module. Instead, I added a deprecation warning to the VuFind\Db\Row\User class if blowfish encryption is enabled. This way, if blowfish is broken due to the use of OpenSSL 3, the user will get the exception you created. If blowfish is still working, they'll instead get a deprecation notice as a nudge in the right direction.

2.) I realized that the upgrade controller is only half of the puzzle, because there's also the install controller, which was turning on blowfish by default. A refactored some code so that install and upgrade controllers can share logic related to database security, I revised the install controller to use aes, and I added a new critical check to the upgrade controller to complain if database settings are not fully secure. (It is still possible to opt out of secure settings, but hopefully this extra nagging will be helpful).

3.) I removed a note about compatibility with VuFind 1.x, because I _hope_ nobody is contemplating connecting VuFind 9.0 to a running instance of VuFind 1.x. I hope there ARE no more running instances of VuFind 1.x by now.

4.) I changed the default config.ini to have aes encryption as an uncommented setting, to reduce the odds of anyone encountering the bad default of blowfish... but I made sure the code still treats blowfish as the default everywhere for backward compatibility.

5.) I added double quotes around the key in your command line message, because I discovered that & characters can cause trouble otherwise.

I think this might get us to a mergable state, but I wanted to get your eyes on my work first. If you're happy with it, I might request an additional review from somebody else just to make sure we're not overlooking any further details.

module/VuFindAdmin/src/VuFindAdmin/Controller/AdminController.php
```diff
@@ -64,6 +64,20 @@ public function homeAction()
         }
         $view = $this->createViewModel();
         $view->xml = $xml ? simplexml_load_string($xml) : false;
+
+        if (!isset($config->Authentication->ils_encryption_algo)
+            || $config->Authentication->ils_encryption_algo == "blowfish"
```
demiankatz: More concise might be:

```
if (($config->Authentication->ils_encryption_algo ?? 'blowfish') == "blowfish") {
```

module/VuFindAdmin/src/VuFindAdmin/Controller/AdminController.php
```diff
@@ -64,6 +64,20 @@ public function homeAction()
...
+        ) {
+            $this->flashMessenger()->addMessage(
+                'Database encryption method is set to blowfish or is set unset ' .
```
demiankatz: I'd remove "set" from "set unset"

module/VuFindAdmin/src/VuFindAdmin/Controller/AdminController.php
```diff
@@ -64,6 +64,20 @@ public function homeAction()
...
+                    '(defaulting to blowfish). This lay cause security ' .
+                    'vulnerabilities. <a href="' .
+                    $this->url()->fromRoute('Upgrade', 'CriticalFixBlowfish') .
+                    '">Go to this upgrade pagefor more information.</a> ',
```
demiankatz: Missing space: "pagefor"

demiankatz (changes requested):
Thanks, @crhallberg. I'm still not convinced that we need this (due to low usage of the admin module, and the addition of deprecation warnings elsewhere), but if you feel strongly that it's valuable, I have a few suggestions for improvement below.

crhallberg:
@demiankatz my apologies, I missed the comment where you removed the message. I don't think it's necessary either, nor that this was the best place to put it. I'm reverting it now.

config/vufind/config.ini
```diff
@@ -535,10 +535,10 @@ ils_encryption_key = false
 ; This is the algorithm used to encrypt and decrypt catalog passwords.
 ; A symmetrical encryption algorithm must be used.
 ; You can use openssl_get_cipher_methods() to see available options on your system.
-; Common choices: blowfish (default), aes, cast, sm4
+; Common choices: blowfish (deprecated -- do not use), aes (recommended), cast, sm4
 ; If you want to convert from one algorithm to another, run this from $VUFIND_HOME:
 ;   php public/index.php util switch_db_hash oldhash:oldkey (or none) newhash:newkey
-;ils_encryption_algo = "blowfish"
+ils_encryption_algo = "aes"
```
crhallberg, line 541: If we're turning this on be default, will the system prompt for or generate a random key for the cipher?

demiankatz: We're not turning encryption on by default, we're just setting a default encryption algorithm to use if the setting is enabled.

demiankatz: (The encrypt_ils_password setting above still defaults to false, and that's the key thing).

demiankatz:
Thanks again, @crhallberg -- I believe this is now ready to merge. I'd like another look from @EreMaijala if he doesn't mind, though, just to be sure we haven't overlooked anything important!

EreMaijala (approved):
Looks good to me!

demiankatz (approved):
Thanks for the double-check, @EreMaijala!

PR #2641, opened 2022-12-01
 Fix a PHP notice when a driver does not return callnumber_prefix.

EreMaijala:
This usually remains hidden as it happens during an AJAX call and just produces an additional field in the JSON response.

Another instance was fixed in commit df231793.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2643, opened 2022-12-02
Avoid calling EDS API until necessary.

EreMaijala:
This is admittedly not the prettiest thing ever, but I've tried to do something to avoid calling EDS API just to get a search tab or Blender instance. Even if it's cached, it was possible for a connection problem to prevent creation of the Options class and displaying home page.

Also removes the setUp step that, as far as I can see, didn't actually do anything useful. It established an EDS session, but that would happen in any case when needed.

The SearchTabs helper change is not beatiful, but instantiating UrlQueryHelper would necessarily include the overhead of full initialization of the Params/Options classes.

TODO
- [ ] Update changelog when merging (note removed include_facets config setting, and constructor changes)


module/VuFind/src/VuFind/Controller/EdsController.php
```diff
@@ -126,8 +115,7 @@ protected function getAdvancedFacets()
         $options = $params->getOptions();
         $availableLimiters = $options->getAdvancedLimiters();
         if (!$availableLimiters) {
-            //execute a call to search just to pull in the limiters
-            $this->setUp();
+            throw new \Exception('Gah');
```
demiankatz: I don't think this is meant to be the final code. ;-)

module/VuFind/src/VuFind/Search/EDS/Options.php
```diff
@@ -164,9 +197,20 @@ public function getModeOptions()
...
+        $this->setOptionsFromApi();
         return $this->defaultMode;
     }
 
+    /**
+     * Get an array of sort options.
+     *
+     * @return array
+     */
+    public function getSortOptions()
```
demiankatz, line 226: This is defined identically in the parent class; why are you adding it here?

module/VuFind/src/VuFind/Search/EDS/Options.php
```diff
@@ -154,6 +186,7 @@ public function __construct(
      */
     public function getModeOptions()
     {
+        $this->setOptionsFromApi();
```
demiankatz: I'm not sure if it's hugely better, but what if instead of calling setOptionsFromApi, you added something like:

```
protected getApiProperty($propertyName, $default = null)
{
    $this->setOptionsFromApi();
    return $this->$propertyName ?? $default;
}
```

That at least reduces redundancy and makes things seem a little more atomic, and if we find a more efficient way to do this, it gives us a central place to apply optimizations... but it probably also has its own set of drawbacks.

demiankatz: Also, what about parent methods like `getDefaultLimit` that depend on values populated by `setOptionsFromApi`? Don't we need to redefine those here as well?

module/VuFind/src/VuFind/Search/EDS/Params.php
```diff
@@ -356,4 +355,47 @@ public function getDisplayQuery()
...
+    /**
+     * Get information on the current state of the boolean checkbox facets.
+     *
+     * @param array $include        List of checkbox filters to return (null for all)
+     * @param bool  $includeDynamic Should we include dynamically-generated
+     * checkboxes that are not part of the include list above?
+     *
+     * @return array
+     */
+    public function getCheckboxFacets(
```
demiankatz: Is it worth introducing a getRawCheckboxFacets method to the Base class which just returns $this->checkboxFacets, so that we can then do this override in just one place?

module/VuFind/src/VuFind/View/Helper/Root/SearchTabs.php
```diff
@@ -431,7 +431,15 @@ protected function buildUrlHiddenFilters(Results $results, $filters)
         foreach ($filters as $filter) {
             $params->addHiddenFilter($filter);
         }
-        $urlParams = $results->getUrlQuery()->getParams(false);
-        return $urlParams !== '?' ? $urlParams : '';
+        $urlParams = [];
+        $hiddenFilters = $params->getHiddenFilters();
+        foreach ($hiddenFilters as $field => $values) {
+            foreach ($values as $current) {
+                $urlParams['hiddenFilters'][] = $field . ':"' . $current . '"';
```
demiankatz: This feels really fragile, and possibly incorrect -- might there be escaping problems here? Does this properly account for complex hidden filters? I think it may be worth investigating a different approach to this. Can we somehow defer action in the existing URL query helper? (I admit that I didn't have time to do much digging... but this really feels like there has to be a better way).

demiankatz (changes requested):
Thanks! A few initial suggestions/comments...

module/VuFind/src/VuFind/Controller/EdsController.php
EreMaijala: Gah.

module/VuFind/src/VuFind/Search/EDS/Options.php
EreMaijala: It was supposed to include `$this->setOptionsFromApi();`.

module/VuFind/src/VuFind/Search/EDS/Options.php
EreMaijala: Yes, any that access API options need to be overridden, and I had missed that one.

module/VuFind/src/VuFind/Search/EDS/Params.php
EreMaijala: Done.

module/VuFind/src/VuFind/View/Helper/Root/SearchTabs.php
EreMaijala: It's passed on to http_build_query, which should take care of things. But I can see that UrlQueryHelper has its own method `buildQueryString` which creates `hiddenFilters[]=...` instead of `hiddenFilters[0]...` and avoids something that messes up ampersands (not sure what). I've now changed `buildQueryString` to public static and use it instead. What do you think?

module/VuFind/src/VuFind/RecordDriver/EDS.php
```diff
@@ -485,14 +485,22 @@ public function getItemsTitleSource()
 
     /**
      * Performs a regex and replaces any url's with links containing themselves
-     * as the text
+     * as the text. Also replaces link elements with anchors.
```
demiankatz: Is this related to the other changes, or should this be a separate commit?

module/VuFind/src/VuFind/Search/EDS/Options.php
demiankatz: Thanks. Did you give any thought to my earlier suggestion about a wrapper function for retrieving API-manipulated values, or did I distract you with my follow-up question? (No pressure if you dislike the idea, just want to be sure it didn't get entirely missed).

module/VuFind/src/VuFind/View/Helper/Root/SearchTabs.php
demiankatz: That change seems like a good idea (though we should document it as a minor BC break in the changelog). However, I think it's addressing different issues than I was thinking about. I was worried, for example, about the fact that we're not escaping quotes in `$current`, which might cause problems. I also wasn't sure how this would interact with the RawHiddenFilters setting, but it looks to me like that is addressed purely on the search backend side rather than through the URL, so maybe that part is a non-issue (though it's all a bit confusing!). I guess the bottom line is that I don't like doing this reformatting operation (which seems like something we might potentially do in other places too, and which might have some tricky bits) inside a specific view helper. Is it possible we need a new Params method, like `getHiddenFiltersAsQueryParams` or something? If we move this to a more centralized/reusable/testable place, that might make me feel better about it. Maybe it's also worth searching the code to see if there are other places where we're doing similar query manipulation, in case there's something we should reuse or model this on.

demiankatz (changes requested):
Thanks, @EreMaijala! A few more comments...

module/VuFind/src/VuFind/Search/EDS/Options.php
EreMaijala: @demiankatz Weird, I thought I commented about it, but probably lost the plot somehow. While the approach you suggested would be more elegant, it would require us to know the default to return. Since e.g. `$this->defaultLimit` can be null until API options are processed, we don't know the default yet. Or I misunderstood something. :)

module/VuFind/src/VuFind/View/Helper/Root/SearchTabs.php
EreMaijala: @demiankatz Ah, I see. I don't think the old code did anything different. And I don't think we want to handle escaping for the backend in the query params. I believe it's more coherent to do that when the backend params are gathered and sent to the backend regardless of whether they came from query params or config or whatever. 

`getHiddenFiltersAsQueryParams` on the other hand sounds like it could work. I'll see how it turns out.

module/VuFind/src/VuFind/Search/EDS/Options.php
demiankatz: I don't think we need to know a default -- I just included the second parameter in my example getApiProperty method in case we needed the option to provide one, and I thought it was a good idea to have a fallback in case an unset property was requested. However, in current practice, I wouldn't expect this to get used. It seems like most of the methods have been refactored to: "call setOptionsFromApi, then return a property." So I thought if we made a function that just takes the property name and applies that pattern of behavior, it makes things a little more consistent/readable without changing functionality at all.

module/VuFind/src/VuFind/View/Helper/Root/SearchTabs.php
demiankatz: @demiankatz, I'm not sure that it's specifically escaping for the backend -- if we have a hidden `field` filter with the value `"value"`, what is the expected query parameter? `field:""value""`? Or `field:"\"value\""`? Maybe we've never encountered this situation and answered that question, but it feels like something that might cause problems one way or another.

EreMaijala: Right. I'm not sure what this should do. I just parroted the code in UrlQueryHelperFactory without giving it any further thought, but it's obviously something that needs to be checked.

demiankatz: I guess my bottom line is that I don't want to create more duplicate instances of the same potentially questionable logic. But if we can refactor things so it only happens in one place, then we can revisit the deeper implications separately in the future. :-)

module/VuFind/src/VuFind/RecordDriver/EDS.php
EreMaijala: Should have been separate. Now extracted to #2648.

module/VuFind/src/VuFind/View/Helper/Root/SearchTabs.php
EreMaijala: I've now added methods to the base Params class to return the filters. We still have a few other places that build filters, and I've tried to ensure that all of them do the same thing. Perhaps some further refactoring could be done, but I think I'll need to extract the changes into a separate pull request first.

Looks like quotes in a filter don't pose a problem provided that the filter value is enclosed in quotes in the url. `Params::parseFilter` removes the outermost quotes, so ensuring that they are always present will make sure that the actual filter value is untouched. I don't really like this, but there's no way to change parseFilter without introducing serious BC issues.

EreMaijala: Hidden filter handling now in #2649.

demiankatz:
I know you're busy, so there's no rush... but just a reminder that we should revisit this when time permits, now that #2648 and #2649 are merged.

EreMaijala:
@demiankatz I've now merged dev here, but will need to check that everything's correct now.

module/VuFind/src/VuFind/Search/EDS/Options.php
EreMaijala: @demiankatz Right, I failed to read and understand that your suggestion was to pass the property name to the method instead of it's value at the time. I'll revise.

EreMaijala: Now added.

module/VuFind/src/VuFind/Search/EDS/Options.php
```diff
@@ -612,25 +694,16 @@ public function getAvailableExpanders()
...
+        $settings = $this->apiInfo['ViewResultSettings'] ?? [];
+        // default result Limit
+        $this->defaultLimit = $this->defaultLimit
+            ?? $settings['ResultsPerPage']
+            ?? 20;
+
+        // default view
+        $this->defaultView = $this->defaultView
+            ?? $settings['ResultListView']
+            ?? 'list';
```
demiankatz: Is there a reason this last fallback value has changed from 'list|brief' to just 'list'? I confess that I don't fully understand the purpose of the pipe-delimited default value if we only return the first section of it... but I didn't do a deep dive to see if the full version is used anywhere. Since we're still using pipes in some cases, I suspect we might want to keep them everywhere for consistency.

demiankatz (commented):
Thanks, @EreMaijala -- just one remaining question. When that detail is settled, I'll do some hands-on testing of this and hopefully get it merged. It's looking good to me overall!

module/VuFind/src/VuFind/Search/EDS/Options.php
EreMaijala: Oops, fixed. That was accidentally left in while checking why the view icons don't display in EDS results. They still don't but the problem is unrelated.

module/VuFind/src/VuFind/Search/EDS/Options.php
```diff
@@ -148,23 +193,68 @@ public function __construct(
...
+
+    /**
+     * Given a basic handler name, return the corresponding label (or false
+     * if none found):
+     *
+     * @param string $handler Handler name to look up.
+     *
+     * @return string
+     */
+    public function getLabelForBasicHandler($handler)
```
demiankatz: Should we make this change in the base class to make extensibility cleaner, rather than overriding it here?

demiankatz (changes requested):
One small suggestion to make the code potentially more future-proof. Other than that, everything looks good in my hands-on testing now. Probably just as well to wait on resolving this until the new year, though, so we can both look at this with fresh eyes. :-)

module/VuFind/src/VuFind/Search/EDS/Options.php
EreMaijala: Yes, moved.

module/VuFind/src/VuFind/Search/EDS/Options.php
```diff
@@ -115,28 +146,42 @@ class Options extends \VuFind\Search\Base\Options
...
         parent::__construct($configLoader);
         // 2015-06-30 RF - Changed to unlimited
         //$this->resultLimit = 100;
         $this->viewOptions = [
             'list|title' => 'Title View', 'list|brief' => 'Brief View',
             'list|detailed' => 'Detailed View'
         ];
-        $this->apiInfo = $apiInfo;
-        $this->setOptionsFromApi();
-        $this->setOptionsFromConfig($searchSettings);
```
demiankatz, line 139: Is it possibly worth still running setOptionsFromConfig in the constructor? Or possibly splitting setOptionsFromConfig into two methods, one that overrides API-retrieved values and one that doesn't, so we can set up the non-API settings in the constructor and the rest as part of the API process? Otherwise, it seems that there might be some scenarios where we could construct an object and request a value and find that it hasn't been configured yet. (I think this is unlikely, but maybe not impossible).

What do you think?

demiankatz (commented):
@EreMaijala, I took another look at this and had one last thought (plus I went ahead and corrected a small typo rather than commenting about it). :-)

module/VuFind/src/VuFind/Search/EDS/Options.php
EreMaijala: Yes, makes sense. I decided to change setOptionsFromConfig to only process options that are not dependent on API config and moved the rest to setOptionsFromApi. I also fixed invalid default for defaultView and removed extraneous parameters passing member variables around. I hope it makes more sense now!

module/VuFind/src/VuFind/Search/EDS/Options.php
```diff
@@ -578,16 +690,16 @@ function ($limiter) {
                 );
                 return $limiter;
             },
-            $this->limiterOptions
+            $this->getApiProperty('limiterOptions')
         );
         // No setting = use all available values
-        if (empty($this->advancedLimiters)) {
+        if (!($advancedLimiters = $this->getApiProperty('advancedLimiters'))) {
```
demiankatz: Do we need to use getApiProperty here? This appears to only be set by configuration. (Though if you think it's safer to just use getApiProperty everywhere to be safe, I wouldn't object too strongly...)

module/VuFind/src/VuFind/Search/EDS/Options.php
```diff
@@ -320,104 +465,59 @@ protected function setCommonSettings($searchSettings, $setting, $list, $target)
...
-            $this->highlight = $searchSettings->General->highlighting;
+        if (isset($this->searchSettings->General->highlighting)) {
+            $this->highlight = $this->searchSettings->General->highlighting;
         }
 
         // Set up facet preferences
-        if (isset($searchSettings->General->include_facets)) {
-            $this->includeFacets = $searchSettings->General->include_facets;
+        if (isset($this->searchSettings->General->include_facets)) {
+            $this->includeFacets = $this->searchSettings->General->include_facets;
```
demiankatz: Unless I'm overlooking something, this seems like an orphaned property/config that never gets used for anything. Would you agree?

demiankatz (changes requested):
Thanks, @EreMaijala, this is really coming along nicely. Sorry it took me so long to find time for another review. In any case, I think we're nearly done, but I did notice a couple more minor things.

module/VuFind/src/VuFind/Search/EDS/Options.php
EreMaijala: Better without, I cleaned it up.

module/VuFind/src/VuFind/Search/EDS/Options.php
EreMaijala: You're right, cleaned up. The backend's SearchRequestModel is still supports it just in case someone needs it.

demiankatz (approved):
Looks good -- merging now. Thanks for your patience on this one, @EreMaijala. :-)

PR #2644, opened 2022-12-04
Tests For Channel Provider Classes 

skellamp:


module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
```diff
@@ -0,0 +1,337 @@
...
+        $driver->setRawData($data);
+        return $driver;
+    }
+
+    /**
+     * Get AlphaBrowse object
+     *
+     * @param array $options options for the provider
+     *
+     * @return AlphaBrowse
```
demiankatz: The return type doesn't seem right here.

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
```diff
@@ -0,0 +1,337 @@
...
+            ->getMock();
+        $url = $this->getMockBuilder(\Laminas\Mvc\Controller\Plugin\Url::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $router = $this->getMockBuilder(\VuFind\Record\Router::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $alpha = new AlphaBrowse($search, $url, $router, $options);
+
+        return ['search' => $search,
```
demiankatz: A useful shorthand for this kind of thing is `compact('search', 'url', 'router', 'alpha');`

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
```diff
@@ -0,0 +1,337 @@
...
+     * Test deriving channel information from a record driver object.
+     *
+     * @return void
+     */
+    public function testGetFromRecord(): void
+    {
+        $mocks = $this->getAlphaBrowse();
+        $alpha = $mocks['alpha'];
+        $recordDriver = $this->getDriver();
+        $this->assertSame([], $alpha->getFromRecord($recordDriver));
```
demiankatz: I suspect we need at least two tests for getFromRecord, since we also want to test the code path where the channel token does not match the record driver's unique ID. It might be better in the "normal" case to mock some data so that the method actually returns something, but if the search tests already cover all that logic, maybe it's not strictly necessary. It might at least be a good idea to add another assertion or two on the mocks to confirm that something is really happening, since asserting the return of an empty array doesn't really test most of what's going on.

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
```diff
@@ -0,0 +1,337 @@
...
+
+        $this->assertSame(
+            [$parameters[1]],
+            $alpha->getFromSearch($results, 'channel_token')
+        );
+    }
+
+    /**
+     * Support method to mock objects.
+     *
```
demiankatz, line 140: This should have `@param` comments, and I wonder if it would benefit from a clearer name, like `configureTestTargetAndExpectations` or something like that.

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
```diff
@@ -0,0 +1,337 @@
...
+    public function testGetFromSearch(): void
+    {
+        $results = $this->getMockBuilder(\VuFind\Search\Base\Results::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $data = ['solrField' => 'foo'];
+        $recordDriver = $this->getDriver($data);
+        $results->expects($this->once())->method('getResults')
+            ->will($this->returnValue([$recordDriver]));
+        $parameters = $this->supportMethod();
```
demiankatz: Maybe `[$parameters, $expected] = $this->supportMethod();` would be a slightly more readable syntax here.

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
```diff
@@ -0,0 +1,337 @@
...
+
+    /**
+     * Support method to mock objects.
+     *
+     * @return array
+     */
+    public function supportMethod($options = ['maxRecordsToExamine' => 1], $flag = false)
+    {
+        $obj = $this->getAlphaBrowse($options);
+        $search = $obj['search'];
```
demiankatz: You could use `extract($obj)` instead of these four assignments if you like. If you're using all the variables in the returned array, it's a little more concise.

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
```diff
@@ -0,0 +1,337 @@
...
+        $retrieveBatchArgs = [['foo_id'], new ParamBag()];
+        $retrieveArgs = ['channel_token', new ParamBag()];
+        $class = \VuFindSearch\Command\RetrieveCommand::class;
+        $retrieveBatchClass = \VuFindSearch\Command\RetrieveBatchCommand::class;
+        $collection = $this->getMockBuilder(
+            \VuFindSearch\Response\RecordCollectionInterface::class
+        )
+            ->disableOriginalConstructor()
+            ->getMock();
+        if ($flag) {
```
demiankatz: Maybe name the $flag parameter to explain what the flag is for.

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
```diff
@@ -0,0 +1,337 @@
...
+            $collection->expects($this->once())->method('first')
+                ->willReturn($this->returnValue($driver));
+
+            $commandObj->expects($this->exactly(3))->method('getResult')
+                ->willReturnOnConsecutiveCalls(
+                    $this->returnValue($collection),
```
demiankatz: Do you need to use `$this->returnValue()` inside `willReturnOnConsecutiveCalls`? Our other tests just include the return values directly.

demiankatz (changes requested):
Thanks for this, @skellamp! I've left a few suggestions on the AlphaBrowseTest -- mostly about clarifying naming/documentation, and also pointing out some PHP language features that might be helpful if you are not already familiar with them (though if you simply prefer not to use them, feel free to ignore those suggestions). I think that some of my feedback probably applies across the other tests as well, so I'm going to let you adjust the first one (and make equivalent changes to the other) before taking the time to do a deep review of everything. Please let me know when you're ready for me to take another look.

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
skellamp: Made changes except in RandomTest.php as it has a different construct and do not use all the packaged variables in all calling methods.

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
```diff
@@ -0,0 +1,347 @@
...
+{
+    /**
+     * Test deriving channel information from a record driver object.
+     *
+     * @return void
+     */
+    public function testGetFromRecord(): void
+    {
+        $recordDriver = $this->getDriver(['solrField' => 'foo']);
+        [$alpha, $expetedResult] = $this->configureTestTargetAndExpectations();
```
demiankatz: Typo: should be `$expectedResult`. (Looks like this is copied and pasted in a few places -- hopefully a search-and-replace can easily fix it :-) ).

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/AlphaBrowseTest.php
```diff
@@ -0,0 +1,347 @@
...
+        $this->assertSame($expetedResult, $alpha->getFromRecord($recordDriver));
+    }
+
+    /**
+     * Test deriving channel information from a record driver object when we
+     * have a token that do not match record driver.
```
demiankatz: Typo: should be "does not match"

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/RandomTest.php
```diff
@@ -0,0 +1,178 @@
...
+class RandomTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * Test deriving channel information from a record driver object.
+     *
+     * @return void
+     */
+    public function testGetFromRecord(): void
+    {
+        $parameters = $this->supportMethod();
```
demiankatz: I'd once again recommend giving supportMethod a more descriptive name -- maybe `setUpTestInputsAndExpectations` or something like that?

demiankatz: ...and you could also do the `[$random, $expectedResult] = ...` approach that you applied to other tests to improve readability.

module/VuFind/tests/unit-tests/src/VuFindTest/ChannelProvider/RandomTest.php
```diff
@@ -0,0 +1,178 @@
...
+        $this->assertSame($parameters[1], $random->getFromSearch($results));
+    }
+
+    /**
+     * Support method to mock objects.
+     *
+     * @return array
+     */
+    public function supportMethod()
+    {
```
demiankatz, line 83: This method is quite long, and it might be a good idea to break it into smaller parts with the help of support methods. I don't insist on this -- it's only a test class, and I don't want to slow you down from moving on to other work -- but in principle, smaller methods are easier to follow and make it easier to see how the variables relate to one another. As a first step, simply factoring out all the setup of various mock objects might help -- e.g., at the end, you could probably change:

```
        $coverRouter = $this->getMockBuilder(\VuFind\Cover\Router::class)
            ->getMock();
        $coverRouter->expects($this->once())->method('getUrl')
            ->with($this->equalTo($recordDriver), $this->equalTo('medium'))
            ->willReturn('foo_Thumbnail');
        $recordRouter = $this->getMockBuilder(\VuFind\Record\Router::class)
            ->getMock();
        $recordRouter->expects($this->once())->method('getTabRouteDetails')
            ->with($this->equalTo($recordDriver))
            ->willReturn('foo_Route');
        $random->setCoverRouter($coverRouter);
        $random->setRecordRouter($recordRouter);```

to

```
    protected function getConfiguredCoverRouterMock($recordDriver)
    {
        $coverRouter = $this->getMockBuilder(\VuFind\Cover\Router::class)
            ->getMock();
        $coverRouter->expects($this->once())->method('getUrl')
            ->with($this->equalTo($recordDriver), $this->equalTo('medium'))
            ->willReturn('foo_Thumbnail');
        return $coverRouter;
    }

    protected function getConfiguredRecordRouterMock($recordDriver)
    {
        $recordRouter = $this->getMockBuilder(\VuFind\Record\Router::class)
            ->getMock();
        $recordRouter->expects($this->once())->method('getTabRouteDetails')
            ->with($this->equalTo($recordDriver))
            ->willReturn('foo_Route');
        return $recordRouter;
    }

...

        $random->setCoverRouter($this->getConfiguredCoverRouterMock($recordDriver));
        $random->setRecordRouter($this->getConfiguredRecordRouterMock($recordDriver));
```

(And of course there are other segments of the method that could be given similar treatment; that's just one example).

demiankatz (changes requested):
Thanks, @skellamp -- looking even better; just a few more minor comments.

demiankatz (approved):
Looking good, @skellamp -- thanks again! Merging now.

PR #2645, opened 2022-12-06
set input type for search box to 'search'

charbugs:
I found a problem with the search box in older versions of Chrome for Android: When you enter a search term and then hit the enter key of the virtual keyboard the form doesn't get submitted. Instead the focus jumps to the search field selection.

I found out that the problem does not exist if the input type is set to "search" instead of "text". With that fix I also noticed that the icon of the enter key changed to a search symbol. So I suppose using "type=search" will have some andvantages when it comes to usability and accessibility.

Do you see any conflict with search box related features in VuFind like autocomplete? As far as i can see from the [MDN article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search) there are no differences between "text" and "search" regarding the element API. But according to this article browsers may handel then different at the UX level, e.g. adding a delete cross to the search box.


demiankatz:
Thanks, @charbugs! I'm interested to hear what @crhallberg and @ckaz think about this.

ckaz (approved):
Looks good to me. Any usability improvement is welcome!

crhallberg (approved):
Thank you! Was only set to `text` due to compatibility once upon a time. 👍

demiankatz:
Looks like everyone is happy with this, and the full test suite is passing, and it looks fine to me on visual inspection... so I think there's no reason to delay merging it. Thanks again, @charbugs.

PR #2647, opened 2022-12-08
Fix trying to pass null into strtolower in Browse::getSolrField

LuomaJuha:
This warning shows up with php 8.1


demiankatz (approved):
Thanks, @LuomaJuha!

PR #2648, opened 2022-12-09
Fix handling of link tags in EDS records.

EreMaijala:
EDS seems to return link elements in records. They were not processed correctly and resulted in invalid HTML.


demiankatz (changes requested):
Thanks, @EreMaijala -- this seems like a change worth covering with a unit test. There's already a test class for this, so adding it should be easy. Let me know if you'd like me to take care of that.

EreMaijala:
@demiankatz Sure thing, test now added.

demiankatz (approved):

demiankatz:
Thanks, @EreMaijala!

demiankatz:
I also backported this to release-8.1, since it may be useful to make available sooner to avoid bad data problems.

PR #2649, opened 2022-12-09
Unify handling of hidden filters in URLs.

EreMaijala:
Spin-off from #2643.

TODO
- [x] Update changelog when merging (BC break: UrlQueryHelper::buildQueryString is now static).

### DIFF
--- a/module/VuFind/src/VuFind/Search/Base/Params.php
+++ b/module/VuFind/src/VuFind/Search/Base/Params.php
@@ -1105,13 +1105,7 @@ class Params
      */
     public function getFiltersAsQueryParams(): array
     {
-        $result = [];
-        foreach ($this->getRawFilters() as $field => $values) {
-            foreach ($values as $current) {
-                $result[] = "$field:\"$current\"";
-            }
-        }
-        return $result;
+        return $this->formatFilterArrayAsQueryParams($this->getRawFilters());
     }
 
     /**
@@ -1265,6 +1259,26 @@ class Params
     protected function getRawCheckboxFacets(): array
     {
         return $this->checkboxFacets;
+    }
+
+    /**
+     * Format a raw filter array as a query parameter array.
+     *
+     * Returns an array of strings that parseFilter can parse.
+     *
+     * @param array $filterArray Filter array
+     *
+     * @return array
+     */
+    protected function formatFilterArrayAsQueryParams(array $filterArray): array
+    {
+        $result = [];
+        foreach ($filterArray as $field => $values) {
+            foreach ($values as $current) {
+                $result[] = "$field:\"$current\"";
+            }
+        }
+        return $result;
     }
 
     /**
@@ -1613,13 +1627,7 @@ class Params
      */
     public function getHiddenFiltersAsQueryParams(): array
     {
-        $result = [];
-        foreach ($this->getHiddenFilters() as $field => $values) {
-            foreach ($values as $current) {
-                $result[] = "$field:\"$current\"";
-            }
-        }
-        return $result;
+        return $this->formatFilterArrayAsQueryParams($this->getHiddenFilters());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


module/VuFind/src/VuFind/Search/Base/Params.php
```diff
@@ -1090,6 +1096,24 @@ public function getFilterList($excludeCheckboxFilters = false)
...
     }
 
+    /**
+     * Get the filter list as a query parameter array.
+     *
+     * Returns an array of strings that parseFilter can parse.
+     *
+     * @return array
+     */
+    public function getFiltersAsQueryParams(): array
```
demiankatz, line 1106: Could this and getHiddenFiltersAsQueryParams be further refactored with a support method `protected formatFilterArrayAsQueryParams($filters): array` so that the actual logic is only defined once?

module/VuFind/src/VuFind/Search/UrlQueryHelper.php
```diff
@@ -362,7 +362,7 @@ public function resetDefaultFilters()
      */
     public function getParams($escape = true)
     {
-        return '?' . $this->buildQueryString($this->urlParams, $escape);
+        return '?' . self::buildQueryString($this->urlParams, $escape);
```
demiankatz: Should we call with `static::` to allow for overriding in a subclass? (Not a likely need, but just the principle of the thing...)

module/VuFind/src/VuFind/Search/UrlQueryHelper.php
```diff
@@ -584,20 +584,47 @@ public function asHiddenFields($filter = [])
                     if (!$this->filtered($paramName, $paramValue2, $filter)) {
                         $retVal .= '<input type="hidden" name="' .
                             htmlspecialchars($paramName) . '[]" value="' .
-                            htmlspecialchars($paramValue2) . '" />';
+                            htmlspecialchars('"' . $paramValue2 . '"') . '" />';
```
demiankatz: What is the purpose of adding double quotes here? I understand wanting to standardize quoting of values in hidden filters, but do we really want to put quotes around every arbitrary query parameter?

module/VuFind/src/VuFind/View/Helper/Root/SearchTabs.php
```diff
@@ -246,30 +247,37 @@ public function getCurrentHiddenFilterParams(
     ) {
         if (!isset($this->cachedHiddenFilterParams[$searchClassId])) {
             $view = $this->getView();
-            $searchTabs = $view->plugin('searchTabs');
-            $hiddenFilters = $searchTabs->getHiddenFilters(
+            $hiddenFilters = $this->getHiddenFilters(
```
demiankatz, line 250: Well, that's just bizarre. Why did we pull this out of the view manager instead of just using $this? Glad you noticed and fixed that...

demiankatz (changes requested):
Thanks, @EreMaijala -- definitely a step in the right direction, but I have a few comments/suggestions.

Also, it seems like the SearchTabs view helper is a class in need of a unit test. Is it worth adding one before we merge this? This might be something that @skellamp can help us with if you don't have time.

module/VuFind/src/VuFind/View/Helper/Root/SearchTabs.php
EreMaijala: No idea, but it was probably my fault anyway. :)

module/VuFind/src/VuFind/Search/Base/Params.php
EreMaijala: Yes, done.

module/VuFind/src/VuFind/Search/UrlQueryHelper.php
EreMaijala: Yes, done.

module/VuFind/src/VuFind/Search/UrlQueryHelper.php
EreMaijala: That was a mistake. I got hidden fields and hidden filters mixed.

EreMaijala:
@demiankatz All issues should be addressed, so ready for re-review.

demiankatz (approved):
Thanks, @EreMaijala -- looks good to me and full test suite is passing!

PR #2650, opened 2022-12-09
Blender: Fix daterange handling and related default config.

EreMaijala:
Date range filter handling was quite broken. Thanks to Tobias Groth for reporting the issue.


demiankatz (approved):
Thanks for the fix!

PR #2652, opened 2022-12-12
Blender: count ignored facet mappings in facet values

GrothTobias:
The blender only displays the facet results from mapped facets currently. This is confusing for users since the count behind a facet indicates e.g. 600 and the search result is different since the 1000 from the ignored mapping were not counted. 

Here is a solution for this problem. Please let me know if you have improvement ideas.

### DIFF
--- a/config/vufind/BlenderMappings.yaml
+++ b/config/vufind/BlenderMappings.yaml
@@ -27,6 +27,7 @@
 # Ignore        true to ignore all values for a filter, or an array of values to
 #               ignore. Otherwise any filter value that does not have a field and
 #               value mapping causes the backend to be disabled for the search.
+# CountIgnored  Count ignored mappings to facet results. Default: true.
 # DefaultValue  Filter value used for the backend unless a filter is active.
 #               For example the pcAvailability filter by default limits results to
 #               those that have full text available, so a DefaultValue can be used to
--- a/config/vufind/BlenderMappings.yaml
+++ b/config/vufind/BlenderMappings.yaml
@@ -27,7 +27,6 @@
 # Ignore        true to ignore all values for a filter, or an array of values to
 #               ignore. Otherwise any filter value that does not have a field and
 #               value mapping causes the backend to be disabled for the search.
-# CountIgnored  Count ignored mappings to facet results. Default: true.
 # DefaultValue  Filter value used for the backend unless a filter is active.
 #               For example the pcAvailability filter by default limits results to
 #               those that have full text available, so a DefaultValue can be used to
--- a/module/VuFindSearch/src/VuFindSearch/Backend/Blender/Response/Json/RecordCollection.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/Blender/Response/Json/RecordCollection.php
@@ -380,18 +380,9 @@ class RecordCollection
                 ($ignore = $mappings['Ignore'] ?? '') &&
                 ($mappings['CountIgnored'] ?? true))
             {
-                if (is_array($ignore)) {
-                    foreach ($ignore as $ignoredValue)
-                    {
-                        $result[$ignoredValue] = ($result[$ignoredValue] ?? 0) + $collections[$backendId]->getTotal();
-                    }
-                }
-                else
+                foreach (is_array($ignore) ? $ignore : array_keys($result) as $ignoredValue)
                 {
-                    foreach ($result as $ignoredValue)
-                    {
-                        $result[$ignoredValue] = ($result[$ignoredValue] ?? 0) + $collections[$backendId]->getTotal();
-                    }
+                    $result[$ignoredValue] = ($result[$ignoredValue] ?? 0) + $collections[$backendId]->getTotal();
                 }
             }
         }
--- a/module/VuFindSearch/src/VuFindSearch/Backend/Blender/Response/Json/RecordCollection.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/Blender/Response/Json/RecordCollection.php
@@ -374,6 +374,28 @@ class RecordCollection
                 }
             }
         }
+
+        foreach ($settings['Mappings'] as $backendId => $mappings) {
+            if (($collections[$backendId] ?? false) &&
+                ($ignore = $mappings['Ignore'] ?? '') &&
+                ($mappings['CountIgnored'] ?? true))
+            {
+                if (is_array($ignore)) {
+                    foreach ($ignore as $ignoredValue)
+                    {
+                        $result[$ignoredValue] = ($result[$ignoredValue] ?? 0) + $collections[$backendId]->getTotal();
+                    }
+                }
+                else
+                {
+                    foreach ($result as $ignoredValue)
+                    {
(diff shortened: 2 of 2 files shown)


module/VuFindSearch/src/VuFindSearch/Backend/Blender/Response/Json/RecordCollection.php
```diff
@@ -374,6 +374,28 @@ protected function mapFacetValues(array $collections, array $settings): array
                 }
             }
         }
+
+        foreach ($settings['Mappings'] as $backendId => $mappings) {
+            if (($collections[$backendId] ?? false) &&
+                ($ignore = $mappings['Ignore'] ?? '') &&
+                ($mappings['CountIgnored'] ?? true))
+            {
+                if (is_array($ignore)) {
```
demiankatz: You could simplify this code significantly by reducing the big if..else down to:

```
foreach (is_array($ignore) ? $ignore : $result as $ignoredValue) {
   ...
}
```

demiankatz (changes requested):
Thanks for the contribution, @GrothTobias -- one suggestion about a potential code simplification. I'm also interested to hear @EreMaijala's thoughts on this, since he's the lead designer on Blender.

module/VuFindSearch/src/VuFindSearch/Backend/Blender/Response/Json/RecordCollection.php
GrothTobias: Yes, your code looks a lot better. I just don't know how to simplify the outer if clause or rather, i don't like the way it looks now.

demiankatz: I think in order to meet project standards, the outer if should be reformatted like:

```
        foreach ($settings['Mappings'] as $backendId => $mappings) {
            if ($collections[$backendId] ?? false
                && $ignore = $mappings['Ignore'] ?? ''
                && $mappings['CountIgnored'] ?? true
            ) {
                foreach (is_array($ignore) ? $ignore : array_keys($result) as $ignoredValue) {
                    $result[$ignoredValue] = ($result[$ignoredValue] ?? 0) + $collections[$backendId]->getTotal();
                }
            }
        }
```

I'm not sure if that makes you like it any better... but if we need that many clauses, I think that's the best we can do short of assigning separate variables for the separate conditions.

config/vufind/BlenderMappings.yaml
```diff
@@ -27,6 +27,7 @@
 # Ignore        true to ignore all values for a filter, or an array of values to
 #               ignore. Otherwise any filter value that does not have a field and
 #               value mapping causes the backend to be disabled for the search.
+# CountIgnored  Count ignored mappings to facet results. Default: true.
```
EreMaijala: I'd explain this in a bit more detail to make the purpose clear. Something like "If a filter is ignored, include the filter-less result count for the source in a facet. Should normally be left as true for expected results." Not really happy with that either, but perhaps @demiankatz can offer a better English explanation.

Alternatively we could leave this undocumented since I can't really come up with a scenaria where the old behavior makes sense.

module/VuFindSearch/src/VuFindSearch/Backend/Blender/Response/Json/RecordCollection.php
```diff
@@ -374,6 +374,19 @@ protected function mapFacetValues(array $collections, array $settings): array
                 }
             }
         }
+
+        foreach ($settings['Mappings'] as $backendId => $mappings) {
+            if (($collections[$backendId] ?? false) &&
+                ($ignore = $mappings['Ignore'] ?? '') &&
```
EreMaijala: Maybe `$mappings['Ignore'] ?? false`?

EreMaijala (changes requested):

module/VuFindSearch/src/VuFindSearch/Backend/Blender/Response/Json/RecordCollection.php
demiankatz: If we're not using `$ignore` elsewhere, we definitely don't want/need to create an unused variable here.

config/vufind/BlenderMappings.yaml
demiankatz: If we set this to false, would the number of results shown on the listed facet value be different from the number of results retrieved after applying the facet? If that's the case, I don't see any reason to make the behavior optional (except maybe debugging), so it would probably be okay to drop the setting entirely. If that's not what woud happen, then maybe I'm failing to understand what this is even for... :-)

EreMaijala: Indeed, and I'm not sure why I did it this way, or if I just failed to think this through. But if @GrothTobias had an idea, please let me know. :)

GrothTobias: I can't think of a situation where one would not count the ignored facets, too. However, @EreMaijala mentioned in the email that it would be best to make it configurable, so i made it configurable. :D

EreMaijala: @GrothTobias Sorry, I'm obviously flip-flopping here. Let's _not_ make it configurable. I thought that there might have been a case for configuration, but since I can't really come up with anything meaningful, I have to admit that I just didn't think about it during the initial implementation.

module/VuFindSearch/src/VuFindSearch/Backend/Blender/Response/Json/RecordCollection.php
GrothTobias: Yes that is better. I failed to realize it after changing the code.

config/vufind/BlenderMappings.yaml
GrothTobias: No need to worry. I just thought, that there might be some rare cases i can't think of. :)

EreMaijala: It's still possible, but we can always revisit this if it ever comes up. :)

demiankatz (changes requested):
Thanks, @GrothTobias! I think there are just some minor changes needed now to meet code standards -- you have a few long lines that should be split up. See this alert from the GitHub Actions run:

```
FILE: .../VuFindSearch/Backend/Blender/Response/Json/RecordCollection.php
----------------------------------------------------------------------
FOUND 0 ERRORS AND 3 WARNINGS AFFECTING 3 LINES
----------------------------------------------------------------------
 379 | WARNING | Line exceeds 85 characters; contains 98 characters
 380 | WARNING | Line exceeds 85 characters; contains 94 characters
 381 | WARNING | Line exceeds 85 characters; contains 114 characters
----------------------------------------------------------------------

Time: 1 mins, 17.27 secs; Memory: 36MB
```

Let me know if you need any help with the reformatting!

demiankatz (approved):
Thanks, looking good to me! I'll let @EreMaijala give this one last look and merge it, since he still has an outstanding review as well.

EreMaijala (approved):
Looks good to me!

PR #2654, opened 2022-12-13
KohaRest: Add a workaround for Koha 22.11 BC break.

EreMaijala:
Bug 30275 changed the renewals field of checkouts to renewals_count, so we need to account for both.


demiankatz (approved):

demiankatz:
Thanks, @EreMaijala! I've backported this to the release-8.1 branch so we can get the fix released more quickly.

PR #2655, opened 2022-12-13
Add local/config/vufind ini, json and yaml files to gitignore.

LuomaJuha:


demiankatz:
@LuomaJuha, this is probably a good idea, but before I merge it, it might be worth a bit of brainstorming, because some people (or, maybe just me -- but I'm a person) rely on the ability to commit local configurations in order to track local changes. I just push those changes to a local, secure Git repository instead of a web-connected one.

We might want to define/recommend a best practice to allow committing local files and document it somewhere (maybe the Git page in the wiki) before merging this. My first thought is that it may be possible to include a more specific .gitignore file inside the local/config/vufind directory that reverses these instructions, but I haven't had time to try it. What do you think?

LuomaJuha:
@demiankatz works with me. Actually I just tested and added local/config/vufind/.gitignore file and the path to /.gitignore and it works to just add *.ini, *.json, *.yaml into it.

The question is it better to prevent people from pushing development stuff or something else to repository from the start and whitelist wanted files, or the other way around.

demiankatz:
@LuomaJuha, I think your initial inclination to exclude this by default makes sense. I've confirmed that if I put a .gitignore file inside my local/config directory with:

```
!*.ini
!*.json
!*.yaml
```

...it overrides your defaults in a sensible way. I think this is a reasonable best practice!

demiankatz:
I've added a note to the bottom of the [Getting Started](https://vufind.org/wiki/development:recommended_tools:git#getting_started) section of the Git wiki page to highlight this change.

demiankatz (approved):

PR #2658, opened 2022-12-13
Adjust RecordDataFormatter to accept driver with invoke function

LuomaJuha:
Was lookin at the helper and it is used usually with only certain driver. Decided that if it should work like the record helper.


demiankatz:
Thanks, @LuomaJuha, I think this is certainly an interesting idea and probably a step in the right direction, but I think we need to be careful about compatibility breaks here. A couple of initial thoughts:

1.) What if we adjusted the public methods so that if the first parameter is a record driver, they automatically set the driver and shift the other parameters left? This would allow old-style calls to continue working while also allowing the new-and-improved syntax. Perhaps there's an argument for waiting until we raise the minimum PHP version to 8 so that we can use union types in the typehints for these situations.

2.) The refactoring of the core metadata to a separate template isn't directly related to adding functionality to the helper. Maybe we should address part that as a separate PR, especially since that change can likely be merged more quickly and with less discussion.

LuomaJuha:
@demiankatz bc was something I was wondering, but I will add functions for them also. Also template will be moved into a new pr. :)

LuomaJuha:
@demiankatz Here is the new version with bc. There was only 1 function getData which needed to be taken into account.

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
```diff
@@ -131,20 +151,32 @@ protected function render($driver, $field, $data, $options)
...
      *
      * @return array
      */
-    public function getData(RecordDriver $driver, array $spec)
+    public function getData(...$args)
     {
+        if (empty($args[0])) {
+            return [];
+        }
+        if (is_a($args[0], 'VuFind\\RecordDriver\\AbstractBase')) {
```
demiankatz: I'd suggest `if ($args[0] instanceof RecordDriver) {` instead -- using the class directly instead of a string is better for static analysis.

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
```diff
@@ -131,20 +151,32 @@ protected function render($driver, $field, $data, $options)
...
+            $this->driver = $args[0];
+            array_shift($args);
+        }
+        if (!is_array($args[0])) {
+            throw new \Exception('Argument 0 must be an array');
+        }
```
demiankatz, line 178: Maybe we should also check that $this->driver is non-null at this point and throw an exception if it is, to avoid unexpected behavior caused by mixed legacy/non-legacy invocations.

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
```diff
@@ -131,20 +151,32 @@ protected function render($driver, $field, $data, $options)
 
     /**
      * Create formatted key/value data based on a record driver and field spec.
+     * The first argument can be a descendant of RecordDriver/AbstractBase.
```
demiankatz: I think you can just say RecordDriver in this context, for consistency with the use statement above.

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
```diff
@@ -131,20 +151,32 @@ protected function render($driver, $field, $data, $options)
...
+     * If omitted, then invoke this class with the desired driver.
+     * The second or first argument is an array containing formatting specifications.
      *
-     * @param RecordDriver $driver Record driver object.
-     * @param array        $spec   Formatting specification
+     * @param array ...$args Record driver object and or formatting specifications.
```
demiankatz: I'd make "and or" "and/or" here.

demiankatz (changes requested):
Thanks, @LuomaJuha, this is looking very good. Just a few minor suggestions; once those are addressed, I'll give it a more thorough test, and hopefully merge it!

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
LuomaJuha: Oh ye, I added a null check for this->driver a little bit higher in this part as a separate check.

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
```diff
@@ -50,6 +52,27 @@ class RecordDataFormatter extends AbstractHelper
      */
     protected $defaults = [];
 
+    /**
+     * Record driver object.
+     *
+     * @var RecordDriver
+     */
+    protected $driver;
```
demiankatz: We should probably initialize this to null, in case somebody finds a way to call getData without first calling __invoke (e.g. by pulling the object of the plugin manager).

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/RecordDataFormatterTest.php
```diff
@@ -152,7 +152,7 @@ protected function getFormatter()
             new \VuFind\Config\PluginManager($container)
         );
         $this->addPathResolverToContainer($container);
-        $formatter = $factory($container, RecordDataFormatter::class);
+        $formatter = $factory($container, RecordDataFormatter::class)($this->getDriver());
```
demiankatz: In order to exercise both forms of usage, maybe we can use a `@dataProvider` that passes in the name of a method that receives a driver and spec and returns the data. Then we can run the test twice, once for legacy invocation and once for new invocation.

If you're not sure what I'm talking about, I can probably find some time next week to put together a clearer example (I'm off for the rest of this week).

demiankatz (changes requested):
Thanks, @LuomaJuha, nearly there, though I had a couple more minor thoughts!

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/RecordDataFormatterTest.php
LuomaJuha: Yeah, been doing something with the tests so I can try out and see it comes close to your thoughts!

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
```diff
@@ -131,20 +153,35 @@ protected function render($driver, $field, $data, $options)
...
      *
      * @return array
      */
-    public function getData(RecordDriver $driver, array $spec)
+    public function getData(...$args)
     {
+        if (empty($args[0])) {
+            return [];
+        }
+        if ($args[0] instanceof RecordDriver) {
```
demiankatz, line 166: Should we do the instanceof check BEFORE the empty check? Otherwise, if you invoke the method with a driver and an empty array, we'll end up doing extra work.

demiankatz (changes requested):
Thanks, @LuomaJuha, this looks really good -- I have just one very small remaining suggestion. Once that's addressed, I think this can be merged, unless you plan to do any more work in this PR. If you want to adjust templates to use the new invocation, I wouldn't mind doing that here... but it could also be handled in a follow-up PR if you prefer, along with the other changes that we already discussed handling separately.

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
LuomaJuha: I could do that! Yeah, I can alter the templates to use the new invocation.

demiankatz (approved):
Thanks, @LuomaJuha, this is looking great, and the full integration test suite is passing, so I'm going to merge now. I'll look forward to the follow-up with further template refactoring whenever time permits. :-)

PR #2659, opened 2022-12-15
Use CachingDownloader for Covers

mtrojan-ub:
Just a first small follow-up PR for #2422 

I wonder if it would also be possible to easily re-use this in similar classes that use `HttpServiceAwareTrait`, e.g. `Bokinfo`, `Orb`, or even `VuFind\Cover\CachingProxy`?

TODO
- [x] Add Orb test
- [x] Improve coverage of Google test
- [x] Fix "disabled = false" bug


demiankatz:
Thanks, @mtrojan-ub, this looks like a good start -- and nice to see that it simplifies some code. My only question is whether we want a mechanism to disable caching and/or set the cache lifetime on a case-by-case basis.

module/VuFind/src/VuFind/Content/Covers/Google.php
```diff
@@ -37,9 +39,9 @@
  * @link     https://vufind.org/wiki/development Wiki
  */
 class Google extends \VuFind\Content\AbstractCover
-    implements \VuFindHttp\HttpServiceAwareInterface
+    implements \VuFind\Http\CachingDownloaderAwareInterface
```
demiankatz: I don't think we set up "magic injection" for this interface, so I'm pretty sure the factories for both of these handlers will need to be updated to inject the caching downloader.

demiankatz (changes requested):
Thanks again, @mtrojan-ub. In addition to my earlier comment about cache lifetimes / cache disabling, I also wonder if now would be the right time to add unit tests to these two classes, since they currently lack it. I'm happy to help with that (probably in the new year) if you don't have time.

module/VuFind/src/VuFind/Content/Covers/Google.php
mtrojan-ub: Done!

mtrojan-ub:
Should we also add any specific test cases here, e.g. to make sure that the JSON properties are accessed correctly?

module/VuFind/src/VuFind/Content/Covers/GoogleFactory.php
```diff
@@ -0,0 +1,76 @@
...
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options passed to factory.');
+        }
+
+        $google = new $requestedName();
+        $cachingDownloader = $container->get(\VuFind\Http\CachingDownloader::class);
+        $google->setCachingdownloader($cachingDownloader);
```
demiankatz: I think this should be `setCachingDownloader` for consistency with the capitalization of the method in the trait. (PHP works either way, but I like to aim for consistency).

module/VuFind/src/VuFind/Content/Covers/OrbFactory.php
```diff
@@ -76,6 +76,9 @@ public function __invoke(
         if (!isset($config->Orb->key)) {
             throw new \Exception("Orb 'key' not set in VuFind config");
         }
-        return new $requestedName($url, $config->Orb->user, $config->Orb->key);
+        $orb = new $requestedName($url, $config->Orb->user, $config->Orb->key);
+        $cachingDownloader = $container->get(\VuFind\Http\CachingDownloader::class);
+        $orb->setCachingdownloader($cachingDownloader);
```
demiankatz: setCachingDownloader again.

demiankatz (changes requested):
Thanks, @mtrojan-ub! A couple of very minor requests related to the latest changes.

Regarding tests, I think we can start by just building basic tests that return JSON fixtures via mocks of the caching downloader. We could get a fixture for Google easily enough by just running the code and saving the results. I'm not sure what it would take to get Orb data, but maybe @fredericd would be willing to help us there.

If we can figure out a mechanism for disabling the cache or having different defaults for different providers, then our tests can confirm that functionality with appropriate expectations.

Is that any help? Let me know if you'd like me to contribute any test development -- might be something I can find time for in the new year. :-)

demiankatz:
Thanks for the fixes, @mtrojan-ub! Let me know if you want me to brainstorm further on cache configuration and testing; I think those are the last outstanding details for now. :-)

mtrojan-ub:
Thanks for the feedback, @demiankatz! I won't be able to continue working on this for 2 weeks or so, so i'd be very happy if you have the time to continue on the test development, as you already offered!

demiankatz:
@mtrojan-ub, I've also had a very busy two weeks, but I'm working on a Google test now. I hope to have it pushed up within the hour. :-)

module/VuFind/src/VuFind/Content/Covers/Google.php
```diff
@@ -86,23 +73,44 @@ public function getUrl($key, $size, $ids)
...
+                    'Invalid response body (json)',
+                    $url,
+                    $response->getStatusCode(),
+                    $response->getHeaders(),
+                    $response->getBody()
+                );
+            }
+        };
+
+        $json = $this->cachingDownloader->download($url, [], $decodeCallback);
```
demiankatz, line 119: We shouldn't assume that `cachingDownloader` is set here, since it is set optionally. We should either throw an exception or have a non-cached fallback.

demiankatz:
@mtrojan-ub, as promised, I've pushed up a base GoogleTest for you. This revealed a bug in the Google refactoring (a missing return statement), which I have fixed. There's still a lot to test -- e.g. the various edge/exception cases -- but this at least provides a foundation for further work.

mtrojan-ub:
@demiankatz, thanks for the GoogleTest!

I finally found the time to solve the open merge conflicts. How should we continue at this point? There has been no reaction from @fredericd so far => can we write the Orb test by ourself or should we remove the Orb-related changes from this PR and stick to the Google-related change as a first step?

module/VuFind/src/VuFind/Content/Covers/Google.php
mtrojan-ub: I don't think this will be a problem, since it is a mandatory property which will be set by the factory.

demiankatz: I still think a simple safety check would be a good idea, since we can't guarantee the code will be built by our factory. (For example, if somebody has built a custom factory in an older version and then upgrades, we don't want things to mysteriously break -- not that I think this is particularly likely in this case, but that's the principle behind my suggestion :-) ).

demiankatz:
Regarding Orb, since we haven't gotten any feedback, I think going ahead with our own test should be okay. Thanks for the progress, @mtrojan-ub!

module/VuFind/src/VuFind/Content/Covers/Google.php
mtrojan-ub: OK, so i added a safety check, can you please have another look?

demiankatz:
Thanks, @mtrojan-ub, I think the safety check you added looks good. From my perspective, these are the open issues here:

1.) We should add an equivalent safety check to the Orb code.

2.) We should add an Orb test.

3.) We should add some additional cases to the Google test to cover error conditions and get closer to 100% coverage.

4.) We should discuss how to make caching behavior configurable on a per-service basis. I think that's the biggest missing piece here. Should each service that uses the caching downloader register itself in some way -- e.g. by calling setUpCache with its own class name? Should there be a config file or section that allows service-specific settings (e.g. lifetimes, or the ability to totally disable caching? On a semi-related note, I notice that the existing CachingDownloader code includes $cacheOptions, but this variable never seems to be used for anything. Is this not fully developed yet, or am I missing something?

Let me know if you'd like me to help with any of this -- I can try to put some time into the tests, for example, after the 9.0 release.

mtrojan-ub:
Thanks for the feedback, @demiankatz!

I've already taken care of 1, however it would be very helpful if you could take care of the test development for 2 and 3.

Regarding 4, i wonder what's the best way to realize that. You're right, the `cacheOptions` did not seem to have any effect so far (the pass-through from the `CachingDownloader` to `CacheManager` has been missing so i just added that). I wonder what's the best way to implement a generic mechanism for services to specify caching settings. I definitely think it could be helpful to get these settings from a config file. How about we add class variables to the `CachingDownloaderAwareTrait` to specify which config file + section belongs to the service? Then we could just assume fixed names for the laminas cache options with a prefix (e.g. `cache_ttl` and so on, and pass them through if they exist in the config). What do you think?

https://docs.laminas.dev/laminas-cache/v3/storage/adapter/

demiankatz:
Thanks for the progress, @mtrojan-ub! I've cherry-picked 66e596e68b0cc16292679e8bd4f94730a9eb5ce6 into the release-9.0 branch, since I think this counts as a bug-fix to the existing code, and it makes sense to launch that capability as soon as possible.

I'm adding this PR to the 9.1 milestone so we can focus some effort on finishing it up for the next minor release. That will ensure that I don't forget about doing my part to help with test coverage.

I also think your configuration idea makes sense. By default, caches are set up using the defaults in the [Cache] section of config.ini. The properties in CachingDownloaderAwareTrait could then optionally specify a configuration that will be used to override those defaults (or perhaps could be left null to skip overrides). In order to disable caching, we might need to figure out a way to set an option that causes the "BlackHole" adapter to be used instead of a "real" adapter. I think that will require at least minor changes to the cache manager, but is probably not too difficult to achieve.

demiankatz:
@mtrojan-ub, I'm reviewing PRs on the 9.1 milestone for review on next week's Community Call (and merging dev in where possible, to get green checkmarks everywhere again now that we've added PHP 8.2 to the GitHub Actions grid). Let me know when it's a good time to resume our discussion about this!

mtrojan-ub:
@demiankatz: I added the config settings functionality, as well as an option to disable caching. Can you please have a look? Maybe there's a better way to acquire the ConfigPluginManager needed to get the settings... also i'm not sure whether the `disable` flag actually works.

demiankatz:
Thanks, @mtrojan-ub! If you need help testing the disabled cache functionality, I'm happy to look into that when time permits.

Regarding the approach to the ConfigPluginManager, I think we need to move in one of two directions:

1.) If we keep the current approach, where the ConfigPluginManager is injected, the trait needs to add some checks so that it throws an exception if the dependency is missing, and so that the toArray() call doesn't fail if the configuration section is missing.

2.) Alternatively, it might be better to fetch the configuration from the plugin manager in the factories, and pass that to the trait instead of the whole plugin manager. That makes the factories do more work, but I think it generally results in cleaner architecture.

I somewhat prefer approach 2, but if you have strong reasons for preferring 1, I'm certainly open to continuing that approach.

What do you think?

mtrojan-ub:
Good idea, @demiankatz! I also prefer approach 2 and made the corresponding changes. Are there any specific config sections yet that should be registered in `OrbFactory` & `GoogleFactory`?

config/vufind/config.ini
```diff
@@ -2038,6 +2038,8 @@ HMACkey = mySuperSecretValue
 ; VUFIND_CACHE_DIR (see httpd-vufind.conf). The default location is inside the
 ; local settings directory.
 [Cache]
+; Uncomment the following line to disable all caching
```
demiankatz, line 2046: Right now, I think this is only disabling file-based caching, since there's only a check for it in createFileCache and not the other cache creation methods. We should either make the comment more specific or else apply the logic more places. (I don't have a strong preference for one approach over the other -- I think both are potentially valid).

demiankatz (changes requested):
Thanks, @mtrojan-ub, this looks like a step in the right direction!

I traced the code through to see how the cache options are getting processed, and it looks to me like any options passed in via the Google and Orb factories will be applied as overrides to the defaults found in the `[Cache]` section. So I don't think we need to pass data from `[Cache]` in our factories -- only overrides that we care to apply. Thus, I think the thing to do is simply to add new sections to config.ini for each service that we want to make configurable. We should probably come up with a convention, like `[Cache_GoogleCover]`, so that we can name these consistently. What do you think?

config/vufind/config.ini
mtrojan-ub: Which other cache creation methods do you mean exactly?

As far as i could see, the VuFind\Cache\Manager only supports two creation methods: createFileCache (which is already part of the change), and createNoCache (which obviously doesn't need any options since it will create a BlackHole).

Or do you mean other parts of the code that do not use the Cache Manager at all? But that also means => if they don't use it, some special settings might have no effect at all, no matter if it's file based or not (if they even use the default [Cache] section).

demiankatz: Sorry, @mtrojan-ub, there used to be another cache creation method using APC, but that was deprecated and removed, and I forgot that we only had file caching left -- I should have looked more closely at the code before complaining. :-)

demiankatz:
@mtrojan-ub, I think my misguided comment about the File cache may have served as a distraction from the perhaps more important point I made above, suggesting coming up with a convention for cache override sections. What do you think about that?

mtrojan-ub:
@demiankatz, i didn't have much time to continue on this issue over the last couple of weeks, but the suggested convention is definitely a good thing!

demiankatz:
Thanks, @mtrojan-ub! I'm also quite busy at the moment, but let me know if you'd like me to do any proof-of-concept work on adding the new configuration options. I may get to it eventually. Or we can just see who finds time first, and let each other know when we plan to start on this to be sure we don't both do it simultaneously. :-)

module/VuFind/src/VuFind/Http/CachingDownloader.php
```diff
@@ -105,16 +115,20 @@ protected function getDownloaderCache()
...
-    public function setUpCache(string $cacheId, array $cacheOptions = [])
+    public function setUpCache(string $cacheId, string $cacheOptionsId = null)
     {
         $this->cache = null;
         $this->cacheId = $cacheId;
-        $this->cacheOptions = $cacheOptions;
+
+        if (!empty($cacheOptionsId)) {
+            $fullCacheOptionsId = 'Cache_' . $cacheOptionsId;
+            $this->cacheOptions = $this->configManager->get('config')->$fullCacheOptionsId->toArray() ?? [];
```
demiankatz: I would suggest calling this cacheOptionsSection rather than cacheOptionsId since we generally refer to these things as config sections.

Also, `$this->configManager->get('config')->$fullCacheOptionsId->toArray() ?? []` won't work, because if the section is undefined, you'll get a fatal error about calling a method on null before the null coalescing comes into play. This unfortunately requires the more verbose:

```
$section = $this->configManager->get('config')->$fullCacheOptionsId;
$this->cacheOptions = !empty($section) ? $section->toArray() : [];
```

demiankatz (changes requested):
Thanks for the progress, @mtrojan-ub -- I only had time for a quick review today, but I found one significant issue and also made one minor suggestion. If you don't have time to apply these changes, let me know and I'm happy to work on this further next week when I have more time in my schedule. :-)

module/VuFind/src/VuFind/Http/CachingDownloader.php
mtrojan-ub: this should be fixed now!

demiankatz (changes requested):
Thanks for the progress on this, @mtrojan-ub. Unfortunately, there's still some more work to do...

I made a bit of progress myself today:

1.) I revised the comments in config.ini to add notes about the new cache configuration sections (and to put things in a more logical order).

2.) I found and fixed a bug where disabling caching would also prevent content from being retrieved (because the code was populating the cache, and then fetching data from the cache... but that won't work when you use a "black hole" cache). I refactored the code to work around this issue.

However, I ran into one more problem that I don't currently have time to investigate further.

I tested a few different scenarios using this Google Cover link in my test environment: `http://localhost/vufind/Cover/Show?author=Person%2C+Fake+DOI%2C+1668-1744&callnumber=DG848.15+.V53+1992&size=medium&title=Fake+DOI+test+3&recordid=fakedoi3&source=Solr&isbns%5B0%5D=8820737493&oclc=30585539`

1.) Default settings -- I confirmed that the cover displays and data is populated into the cache.

2.) `disabled = true` in `[Cache_GoogleCover]` section -- I confirmed that the cover displays and data is NOT populated into the cache. (This is the scenario that led me to the bug fix mentioned above).

3.) `disabled = false` in `[Cache_GoogleCover]` section -- I would expect this to behave exactly the same as scenario 1 above... but instead, no cover is displayed, and no data is loaded into the cache!

4.) `randomgarbage = true` in `[Cache_GoogleCover]` section -- this behaves the same as scenario 3.

So I'm not sure exactly what is going on, but it seems like the presence of certain options in the cache override section breaks everything. Any idea what might be causing that?

I have added TODO checkboxes to cover the need to fix this bug and also to complete the other testing tasks we discussed earlier. Hopefully this will help us keep track of the work in progress.

demiankatz:
@mtrojan-ub, as I was writing up the problem I discovered, I realized what the problem was there: the cache factory was throwing an exception due to receiving illegal options. There is no 'disabled' option supported by the cache, so if the cache is explicitly NOT disabled with `disabled = false` we need to unset that option to prevent a problem. I've added the necessary fix and also re-fixed a test that I broke with my refactoring. Hopefully the code is in a better place now, but I'll need to do some more test-related work in the future when I have a little more time.

demiankatz:
@mtrojan-ub, I was able to put a little more time into this today, and I believe all of the outstanding test work is now completed. Do you mind reviewing my changes from yesterday and today and letting me know if you have any concerns? If you're happy with where this has ended up, I believe it can be merged. I'll do one last manual test once I hear back from you, then approve and merge if all goes well and you have no further suggestions.

mtrojan-ub:
Thank you very much, @demiankatz! Looks good to me!

demiankatz (approved):
Thanks, @mtrojan-ub! We can probably do follow-up work to add similar functionality to more handlers, but for now I'm going to merge this.

PR #2660, opened 2022-12-15
Configurable Sort Tie-Breaker 

skellamp:
TODO
- [x] Close [VUFIND-1529](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1529) when merging


module/VuFind/tests/unit-tests/src/VuFindTest/Search/Blender/ParamsTest.php
```diff
@@ -940,7 +940,7 @@ public function testSortMappings(): void
         $this->assertInstanceOf(ParamBag::class, $primoParams);
         $this->assertInstanceOf(ParamBag::class, $edsParams);
 
-        $this->assertEquals(['publishDateSort desc'], $solrParams->get('sort'));
+        //$this->assertEquals(['publishDateSort desc'], $solrParams->get('sort'));
```
demiankatz: Is this commented out for a specific reason?

config/vufind/searches.ini
```diff
@@ -8,6 +8,11 @@ default_handler      = AllFields    ; Search handler to use if none is specified
 ; below.
 default_sort         = relevance
 
+; This setting controls the default sort order of search results if multiple
+; records match the same value when records are sort just using one field.
```
demiankatz: Typo: "records are sort" should be "records are sorted"

module/VuFind/src/VuFind/Search/Solr/Params.php
```diff
@@ -489,6 +489,17 @@ protected function normalizeSort($sort)
                 );
             }
         }
+        $tieBreaker = $this->getOptions()->getSortTieBreaker();
```
demiankatz: I had originally thought about allowing the tie-breaker to be a fully-developed sort value -- e.g. comma-separated, potentially including asc/desc, potentially using the legacy aliases. However, on seeing this elegant solution to the problem, I'm somewhat convinced that we don't need to do all that extra work. We could always start with just a field name for now, and we could expand the functionality in the future if a use case was identified to justify the work (since a future fuller interpretation of sort fields would still be backward compatible with configuration consisting of just the name of a single Solr field).

What do you think? Is it worth adding any additional functionality here, or should this stand as-is? We can ask for more opinions once you're ready to take this out of draft mode.

module/VuFind/src/VuFind/Search/Solr/Params.php
```diff
@@ -489,6 +489,17 @@ protected function normalizeSort($sort)
...
+
+        if (!(in_array($tieBreaker . " asc", $normalized)
```
demiankatz: We should probably make sure $tieBreaker is not null here as well.

demiankatz (changes requested):
Thanks, @skellamp, I think this is developing nicely. I've left a few comments and thoughts. I think the only other thing I can think of beyond the notes below is that we'll want to be sure we have test coverage for the newly added functionality.

module/VuFind/src/VuFind/Search/Solr/Params.php
```diff
@@ -489,6 +489,18 @@ protected function normalizeSort($sort)
                 );
             }
         }
+        $tieBreaker = $this->getOptions()->getSortTieBreaker();
+
+        if (!empty($tieBreaker)
+            && !((in_array($tieBreaker . " asc", $normalized)
```
demiankatz: I think there's an extra set of parentheses here -- we seem to have `((..))` around the two in_array checks when `(..)` would suffice.

module/VuFind/src/VuFind/Search/Solr/Options.php
```diff
@@ -290,6 +300,16 @@ public function getEmptySearchRelevanceOverride()
         return $this->emptySearchRelevanceOverride;
     }
 
+    /**
+     * Get the field to be used as a sort tie-breaker.
+     *
+     * @return string Sort field or null if not set
```
demiankatz: If we want to allow null return, this should be `@return ?string`.

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Blender/ParamsTest.php
```diff
@@ -1069,9 +1069,13 @@ protected function getConfigManager()
      */
     protected function getParamsClassesArray(): array
     {
-        $solrConfigMgr = $this->createMock(\VuFind\Config\PluginManager::class);
+        $solrConfigMgr = new \VuFindTest\Container\MockConfigPluginManager($this);
+        $solrConfigMgr->set('facets', new \Laminas\Config\Config([]));
+        $solrConfigMgr->set(
+            'searches',
+            new \Laminas\Config\Config(['General' => ['tie_breaker_sort' => 'id']])
```
demiankatz: What happens if we remove these customizations to the configuration of the configuration plugin manager? Does the test actually fail, or did you just add these in order to test the new functionality? If the former, I'm curious how/why; if the latter, this might not be the best place for this test, since this is a test of Blender, and it would probably make more sense to put a test in the Solr-specific test class.

Another somewhat related question is whether perhaps we want the tie-breaker field to default to "id" if not otherwise configured. In that case, I would expect that we would need to change this test, but not customize the search configuration in the test.

demiankatz (changes requested):
Thanks, @skellamp -- this keeps getting better. I fixed a very small typo (duplicate period in comment) and left a few more questions/comments.

demiankatz:
Thanks for the progress, @skellamp! Let me know if you need any more information to proceed with the final test adjustments. There's also an open conversation thread above that might be worth continuing, though I don't think it's necessary to make this more complicated at this time. :-)

demiankatz (approved):
Thanks again, @skellamp. I've made a few minor adjustments: I tweaked some whitespace to eliminate unnecessary diffs, I renamed your test in the ParamsTest (I think it accidentally had a copied-and-pasted name from elsewhere), and I separately merged your MockConfigPluginManager class to the dev branch, since I think it might be useful in the future, but it is no longer used here and is thus no longer related to the rest of this PR. I also added a couple of tests to the OptionsTest to cover the other half of the new code. I'm asking for a review from @EreMaijala in case he has any feedback before we merge this.

config/vufind/searches.ini
```diff
@@ -8,6 +8,11 @@ default_handler      = AllFields    ; Search handler to use if none is specified
 ; below.
 default_sort         = relevance
 
+; This setting controls the default sort order of search results if multiple
+; records match the same value when records are sorted just using one field.
+; selected field is used as a tie breaker to sort the matching results.
```
EreMaijala: Starts with a lowercase s.

EreMaijala (changes requested):
I think there's a case to be made about wanting e.g. 'id desc' as tie breaker to prefer newer records that a higher id generally means. So yes, I'd like to see at least support for indicating the direction, but I don't think it's much more work to append the tie breaker to the other sort options and then just check that $normalized doesn't contain the value before adding it. Solr doesn't seem to have any issue sorting e.g. by `id asc,title,id desc`.

module/VuFind/src/VuFind/Search/Solr/Params.php
```diff
@@ -482,11 +487,14 @@ protected function normalizeSort($sort)
                     $order ?: $table[$field]['order']
                 );
             } else {
-                $normalized[] = sprintf(
+                $sortValue = sprintf(
                     '%s %s',
                     $field,
                     $order ?: 'asc'
                 );
+                if (!in_array($sortValue, $normalized)) {
```
demiankatz: This won't prevent you from having a weird sort order like "id desc,id asc". Maybe the thing to do is to initialize a `$fields` array at the same time that we initialize `$normalized`, and then change this logic to:

```
if (!in_array($field, $fields)) {
    $normalized[] = $sortValue;
    $fields[] = $field;
}
```

This will ensure that no field appears in the sort list more than once -- it helps with tie-breakers but also with problems in the core sort configuration.

I believe this also means that we can now support complex, multi-part tie-breakers, like "title desc,id asc" if we want to. It might be worth adding a couple more cases to the unit test to cover these additional scenarios.

demiankatz (changes requested):
Thanks, @skellamp, I'm very happy with the direction this has taken -- I think it's a significant improvement in flexibility and simplicity. Just a couple final suggestions below!

demiankatz (approved):
Looks great! Thanks again, @skellamp!

EreMaijala (approved):

PR #2662, opened 2022-12-16
FOLIO: Enhance getMyHolds to provide more properties

cwolfebsco:
The changes in this pull request will enhance the Folio Driver to populate the following properties of the "holds" object of the getMyHolds method:

- available
- in_transit
- last_pickup_date
- position

Populating these properties should enable VuFind to provide an improved user experience for Holds / Recalls:

- Show bell icon next to the My Holds menu item in the My Account area, if an item is available for pickup
- Show clock icon next to the My Holds menu item in the My Account area, if an item is in transit
- Show status message below hold entry if they are available or in transit
- Show last pickup date below hold entry if item is available for pickup
- Show position for hold entry if item is not yet in_transit or available

The status of a hold is described as an enum, see [https://github.com/folio-org/mod-circulation-storage/blob/v14.1.0/ramls/request.json](https://github.com/folio-org/mod-circulation-storage/blob/v14.1.0/ramls/request.json). 

Folio.ini has been extended to define, which of the status messages constitute an available or in_transit.

Question / Request for Comments and Suggestion:

- [x] Would you please look at line 1175 and the formatting of the lastPickup Date - I believe there is room for improvement

Screen Shot 2022-12-16 at 15 18 18

Kudos for this work should go to Ben Taylor of Trinity College Cambridge, who identified the fact that this functionality was missing and documented the requirements.

### DIFF
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -142,3 +142,10 @@ maxNumberItems = 5
 ; dueDates will be shown in Universal Time Format, e.g. 06 Jul 2022 14:59, set showTime
 ; to false to show 06 Jul 2022 only
 showTime = false
+
+[Holds]
+
+available[] = "Open - Awaiting pickup"
+in_transit[] = "Open - Not yet filled"
+in_transit[] = "Open - In transit"
+in_transit[] = "Open - Awaiting delivery"
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -144,8 +144,13 @@ maxNumberItems = 5
 showTime = false
 
 [Holds]
+; This section associates status text found in FOLIO with specific VuFind statuses
+; all possible Folio status messages for holds are found at (2022-12-19)
+; https://github.com/folio-org/mod-circulation-storage/blob/v14.1.0/ramls/request.json.
 
+; FOLIO statuses indicating available-for-pickup items
 available[] = "Open - Awaiting pickup"
+; FOLIO statuses indicating in-transit items
 in_transit[] = "Open - Not yet filled"
 in_transit[] = "Open - In transit"
 in_transit[] = "Open - Awaiting delivery"
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -142,10 +142,3 @@ maxNumberItems = 5
 ; dueDates will be shown in Universal Time Format, e.g. 06 Jul 2022 14:59, set showTime
 ; to false to show 06 Jul 2022 only
 showTime = false
-
-[Holds]
-
-available[] = "Open - Awaiting pickup"
-in_transit[] = "Open - Not yet filled"
-in_transit[] = "Open - In transit"
-in_transit[] = "Open - Awaiting delivery"
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -128,8 +128,8 @@ cancellation_reason = 75187e8d-e25a-47a7-89ad-23ba612338de
 ;helpText[en-gb] = "Help text for British English localization."
 
 ; This folowing settings associate status text found in FOLIO with specific VuFind 
-; statuses all possible Folio status messages for holds are found at (2022-12-19)
-; https://github.com/folio-org/mod-circulation-storage/blob/v14.1.0/ramls/request.json.
+; statuses. All possible Folio status messages for holds are found at:
+; https://github.com/folio-org/mod-circulation-storage/blob/master/ramls/request.json
 
 ; FOLIO statuses indicating available-for-pickup items
 available[] = "Open - Awaiting pickup"
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -127,6 +127,17 @@ cancellation_reason = 75187e8d-e25a-47a7-89ad-23ba612338de
 ;helpText[*] = "Default help text used if not overridden."
 ;helpText[en-gb] = "Help text for British English localization."
 
+; This folowing settings associates status text found in FOLIO with specific VuFind 
+; statuses all possible Folio status messages for holds are found at (2022-12-19)
+; https://github.com/folio-org/mod-circulation-storage/blob/v14.1.0/ramls/request.json.
+
+; FOLIO statuses indicating available-for-pickup items
+available[] = "Open - Awaiting pickup"
+; FOLIO statuses indicating in-transit items
+in_transit[] = "Open - Not yet filled"
+in_transit[] = "Open - In transit"
+in_transit[] = "Open - Awaiting delivery"
+
 [CourseReserves]
 ; If set to true, the course number will be prefixed on the course name; if false,
 ; only the name will be displayed:
@@ -142,15 +153,3 @@ maxNumberItems = 5
 ; dueDates will be shown in Universal Time Format, e.g. 06 Jul 2022 14:59, set showTime
 ; to false to show 06 Jul 2022 only
 showTime = false
-
-[Holds]
-; This section associates status text found in FOLIO with specific VuFind statuses
-; all possible Folio status messages for holds are found at (2022-12-19)
-; https://github.com/folio-org/mod-circulation-storage/blob/v14.1.0/ramls/request.json.
-
-; FOLIO statuses indicating available-for-pickup items
-available[] = "Open - Awaiting pickup"
-; FOLIO statuses indicating in-transit items
-in_transit[] = "Open - Not yet filled"
-in_transit[] = "Open - In transit"
-in_transit[] = "Open - Awaiting delivery"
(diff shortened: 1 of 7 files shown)


config/vufind/Folio.ini
```diff
@@ -142,3 +142,10 @@ maxNumberItems = 5
 ; dueDates will be shown in Universal Time Format, e.g. 06 Jul 2022 14:59, set showTime
 ; to false to show 06 Jul 2022 only
 showTime = false
+
+[Holds]
```
demiankatz: We should probably add comments here to explain what these settings mean... e.g., before `[Holds]`:

`; This section associates status text found in FOLIO with specific VuFind statuses`

...and then maybe a comment above each of the settings as well ("FOLIO statuses indicating available-for-pickup items" / "FOLIO statuses indicating in-transit items").

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1171,6 +1171,10 @@ public function getMyHolds($patron)
             // Set expire date if it was included in the response
             $expireDate = isset($hold->requestExpirationDate)
                 ? date_create($hold->requestExpirationDate) : null;
+            //set lastPickup Date if provided, format to j M Y
+            $lastPickup = isset($hold->holdShelfExpirationDate)
+                ? date_format(date_create($hold->holdShelfExpirationDate), "j M Y")
```
demiankatz: I think you probably want to use `$this->dateConverter->convertToDisplayDate("Y-m-d H:i", $hold->holdShelfExpirationDate)` here. Note that the format string (my example is "Y-m-d H:i", which may not be correct) is intended to describe the EXISTING format of the date, so that VuFind can parse it and reformat it to the display format specified by config.ini.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1181,6 +1185,16 @@ public function getMyHolds($patron)
                 'reqnum' => $hold->id,
                 // Title moved from item to instance in Lotus release:
                 'title' => $hold->instance->title ?? $hold->item->title ?? '',
+                'available' => in_array(
+                    $hold->status,
+                    $this->config['Holds']['available']
```
demiankatz, line 1221: Maybe we should provide a fallback to default values here in case the settings are missing -- e.g. `$this->config['Holds']['available'] ?? $defaultAvailableStatuses`, with $defaultAvailableStatuses defined earlier in the code somewhere.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1181,6 +1185,16 @@ public function getMyHolds($patron)
...
+                ),
+                'in_transit' => in_array(
+                    $hold->status,
+                    $this->config['Holds']['in_transit']
```
demiankatz, line 1226: A default fallback here might also be a good idea.

demiankatz (changes requested):
Thanks, @cwolfebsco! This is a great start; just a few suggestions (including the one you requested regarding dates).

Also, in addition to the below, it would probably be a good idea to add test coverage for this method while we are doing the work of changing it anyway. I can probably help with that in the new year if you're not sure where to begin. Perhaps a starting point would be if you could contribute a sample response JSON document as a test fixture in [this directory](https://github.com/vufind-org/vufind/tree/dev/module/VuFind/tests/fixtures/folio/responses) so we have something to build the test around.

cwolfebsco:
Thank you so much for all the useful comments and kind guidance. I hope I was able to accurately reflect the requested changes in the updated Pull Request. I am particularly uncertain with the Unit Tests, but happy to work on them more based on your guidance

config/vufind/Folio.ini
demiankatz: @cwolfebsco, thanks for adding the comments, but I just realized something -- there is already a `[Holds]` section in Folio.ini, and having two is going to cause problems. You should move your new settings into the existing section. Maybe you can adjust your "This section" comment to instead say "The following settings" since it no longer describes the entire section.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1167,20 +1167,51 @@ public function getMyHolds($patron)
...
+            );
             // Set expire date if it was included in the response
             $expireDate = isset($hold->requestExpirationDate)
-                ? date_create($hold->requestExpirationDate) : null;
+                ? $this->dateConverter->convertToDisplayDate(
+                    "Y-m-d H:i",
+                    $hold->requestExpirationDate
+                )
+                : null;
+            //set lastPickup Date if provided, format to j M Y
```
demiankatz: Very minor point, but I'd put a space after the `//` here for consistency with other comments.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1167,20 +1167,51 @@ public function getMyHolds($patron)
...
+                    $hold->holdingsRecordId,
+                    $hold->itemId
+                ),
                 'item_id' => $hold->itemId,
                 'reqnum' => $hold->id,
                 // Title moved from item to instance in Lotus release:
                 'title' => $hold->instance->title ?? $hold->item->title ?? '',
+                'available' => in_array(
+                    $hold->status,
+                    $this->config['Holds']['available'] ?? $availabilityArray
```
demiankatz: This is great, but maybe a further improvement would be to move this to

```
/**
 * Default availability messages, in case they are not defined in Folio.ini
 *
 * @var string[]
 */
protected $defaultAvailabilityStatuses = ['Open - Awaiting pickup'];
```

at the top of the class, and then use `$this->defaultAvailabilityStatuses` in place of `$availabilityArray` here. (And of course do something similar for the other defaults as well). This not only makes the function code less cluttered by moving the variable definitions outside, but it also makes it possible to customize the values in a local subclass if there were ever a reason to do that.

It's a minor point, and fine as it is if you prefer to leave it... but just a suggestion!

demiankatz (changes requested):
Thanks, @cwolfebsco, this is really great -- and the tests are exactly what I was hoping for! I found one significant (but easily fixed) problem with the .ini file and had a few other very minor comments. Once you've had a chance to make what fixes you wish, I'll try to do some hands-on testing (though that will have to wait until the new year, as I don't want to mess with anything while we're in the middle of migration).

Happy holidays in the meantime!

config/vufind/Folio.ini
cwolfebsco: Oh that is bad and embarrassing, I'll fix this ASAP

module/VuFind/src/VuFind/ILS/Driver/Folio.php
cwolfebsco: Sorry for oversights like this...

module/VuFind/src/VuFind/ILS/Driver/Folio.php
cwolfebsco: When it comes to creating code in a big project like this I am the most junior of junior developers, any guidance is gratefully appreciated. Despite the fact that these variables are unlikely to see much re-use (at least I cannot see that right now), I like the  thought of removing clutter from this function - so this will be in my next push.

I will just quickly run those changes on a test server with "live" data before hitting that push button.

cwolfebsco:
Hello @demiankatz - appreciate all the constructive feedback. I sometimes fear that I cause you more work with a PR than if I had just reported said thing as an issue, but I hope that at one point I can repay the favor with having actually added value to this very awesome project.

demiankatz:
Thanks, @cwolfebsco! Please don't worry about causing me work -- your contributions have been very helpful, and I really don't mind pointing out minor areas of potential improvement. I look at PHP code every day, so it costs me very little to notice these types of things, and I don't mind sharing suggestions with people who spend less time in this mindspace. ;-) Thanks for taking constructive criticism so well!

I think the code here is solid, so now just one last question about the configuration: do you really think we should treat "Open - not yet filled" as an in transit status? My understanding is that "in transit" is meant to mean "the item has been located and is currently on its way to you." If the request hasn't been filled yet, we might create the wrong user expectations by flagging "in transit" too early in the process. Of course, I don't exactly know how these statuses are applied in real-world FOLIO scenarios, and I haven't carefully reviewed how this flag is used in the VuFind code, so maybe this is more appropriate than I think... but at a glance, it did make me wonder if we'd be better off commenting that one out by default.

Please let me know what you think about this. In the meantime, I'll raise awareness of this PR in the FOLIO community in case anybody else wants to test it... and as I said earlier, I'll try it out myself in the new year once I have a stable system to test it with. (If there's any urgency about merging this, let me know so I can prioritize appropriately!).

cwolfebsco:
There is no urgency and the more I think about it, the more I feel we should remove the Open - not filled from the in_transit array.

I hope that you have a wonderful festive season and a great start to 2023. We can pick this up in the new year...

demiankatz:
Thanks, @cwolfebsco -- same to you, and I'll see you next year! ;-)

demiankatz (approved):
Thanks again, @cwolfebsco -- now that we're live on FOLIO here, I was able to test this more thoroughly. It looks good! The only tiny change I made was to add a `?? null` fallback on the position value just in case it's ever missing for some reason (which may not be likely, but I think it's probably better to potentially omit an optional field than trigger a PHP notice).

PR #2663, opened 2022-12-20
Fix defaults and error message in cache manager

EreMaijala:
Two separate commits to fix the default value of defaults in cache manager and the error message for an obsolete setting.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2664, opened 2022-12-21
Fix old mistakes in the Swedish translation.

EreMaijala:
Looks like I dropped the ball [some nine years ago](https://github.com/vufind-org/vufind/commit/df3a4e2e51bb0d04391b926dab8694be73a42235#diff-4ea6c0f92f0458d4809f6f80832069e49668c772391a5e306c37d86840c4312aR453) when updating the Swedish translations. Here's a fixup for some remaining mistakes.


demiankatz (approved):
Thanks, @EreMaijala! I expect to put out release 8.1.2 early in the new year, so I've backported these fixes to the release-8.1 branch so we can take advantage of them that much sooner.

PR #2665, opened 2022-12-22
Separate core fields into their own templates

LuomaJuha:


LuomaJuha:
Hmm, the php-cs-fixer seems very odd. Going to adjust it a little.

demiankatz:
@LuomaJuha, when php-cs-fixer gets fussy about templates, sometimes it helps to put the `<?=` on a separate line from the actual code. I'd recommend running `phing php-cs-fixer`, and if you don't like the result, try putting newlines after the `<?=` and repeating the process.

I haven't looked closely at the rest of the code yet, but I thought this might help you get to the point where the build passes. Please let me know when you're ready for more feedback, or if I can be of further help with this nuisance. :-)

LuomaJuha:
@demiankatz yea, seems like it only cared about the first parameter and broke after the first array bracket. Adjusted the end parts and now it passes.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/collection-info.phtml
```diff
@@ -39,18 +39,16 @@
...
     <?php /* Display Main Details */ ?>
-    <?php
-      $formatter = $this->recordDataFormatter($driver);
-      $fields = $formatter->getData($formatter->getDefaults('collection-info'));
+    <?=
+      $this->record($this->driver)->renderTemplate(
+          'core-fields.phtml',
+          [
+            'driver' => $this->driver,
+            'defaults' => 'core',
```
demiankatz: Should this be `collection-info` rather than `core`?

themes/bootstrap3/templates/RecordDriver/DefaultRecord/core-fields.phtml
```diff
@@ -0,0 +1,12 @@
+<?php
+  $formatter = $this->recordDataFormatter($this->driver);
+  $coreFields = $formatter->getData($formatter->getDefaults($this->defaults));
```
demiankatz: Would it make sense to use `$this->defaults ?? 'core'` here so we default to reasonable defaults?

demiankatz (changes requested):
Thanks, @LuomaJuha, this looks pretty good -- just a couple of small comments...

themes/bootstrap3/templates/RecordDriver/DefaultRecord/collection-info.phtml
LuomaJuha: Absolutely.

demiankatz (approved):
Thanks again (and happy new year), @LuomaJuha, this looks great to me. Merging now!

PR #2666, opened 2022-12-22
Clarify/simplify highlighting code/comments.

demiankatz:
I've been working with @welblaud to troubleshoot a problem he is seeing where, after upgrading to Solr 9, he is getting `java.lang.IllegalArgumentException: field 'author' was indexed without offsets, cannot highlight` errors. We haven't yet determined whether this is tied to local customizations or something else, but while troubleshooting, I discovered that the problem goes away if we disable highlighting in Solr, but turning off highlighting in searches.ini doesn't automatically prevent VuFind from requesting highlighting data. This is because highlighting data is also required for snippets, but I had forgotten about that.

While investigating, I noticed some code that could be simplified, and I thought adding a comment about the snippet detail might save somebody some time in the future -- hence this pull request. If the Solr error is identified as something related to the core project, I'll address that in a separate PR... but for now, investigation is ongoing.

### DIFF
--- a/config/vufind/searches.ini
+++ b/config/vufind/searches.ini
@@ -71,7 +71,10 @@ default_noresults_recommend[] = SpellingSuggestions
 default_noresults_recommend[] = RemoveFilters
 
 ; Set this to true in order to highlight keywords from the search query when they
-; appear in fields displayed in search results.
+; appear in fields displayed in search results. Note: while this setting will
+; always prevent highlighting from displaying in your output, if you want to
+; prevent VuFind from requesting highlighting data from Solr, you must also turn
+; off the snippets setting below, since snippets depend on highlighting data.
 highlighting = true
 
 ; Set this to restrict the list of fields that will be highlighted (the hl.fl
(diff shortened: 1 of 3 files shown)


EreMaijala:
@demiankatz We probably need `storeOffsetsWithPositions="true"` for highlight fields. See [Solr reference guide](https://solr.apache.org/guide/solr/latest/query-guide/highlighting.html#schema-options-and-performance-considerations) for more information. We've had these enabled for a long time, so I had completely forgotten about them. It's also possible to revert to the original highlighter, but unified highlighter with the underlying schema changes is so much faster that I wouldn't do it.

See e.g. https://github.com/NatLibFi/finna-solr/blob/dev/vufind/biblio/conf/schema.xml#L178 for an example.

Edit: termVectors are probably less important.

EreMaijala (approved):

demiankatz:
Thanks, @EreMaijala -- I'll work with @welblaud to try those schema adjustments, and if they help his scenario, we can evaluate whether they are also needed upstream in the core project.

PR #2667, opened 2023-01-06
Add results metadata as data attributes

maccabeelevine:
These two data attributes are useful metadata for analytics tracking

1. "recordTotal" as a data attribute on div.search-stats.  It's obviously useful to know how many results there were for a given search; especially if the number is zero.  This number is otherwise only available by parsing the "showing results" string, which varies by localization.  It could also be added directly to the span that holds the results total, but that's within the localization file and would have to be done in each translation.
2. "recordNumber" as a data attribute on the individual search result list item.  It is useful to quantify how often a user clicks on a result past the first 5-10.  This number is already present in a div within the list item so not strictly necessary, but it rightly belongs as metadata on the result itself.  For example if the "title of the result" link is clicked, we should ideally be able to get which recordNumber was clicked by traversing only to an ancestor of the link, not siblings.


maccabeelevine:
Tests TBD.

maccabeelevine:
Or actually it doesn't look like there are unit tests for the themes, unless I'm missing something.

demiankatz:
Thanks, @maccabeelevine -- we do have Mink tests to cover some UI things, so in theory we could write integration tests to verify that these numbers are present and correct... but I think it's okay to get by without them.

My only suggestion is that it might be good to wrap `$this->escapeHtmlAttr()` around both of the values. I realize that under normal operations, these should always be integers, but it never hurts to add appropriate escaping just in case something weird happens. :-)

Also interested in whether @crhallberg has any concerns/suggestions.

maccabeelevine:
@demiankatz Good idea on escaping the HTML, yes it's numeric and not user-supplied data but can't hurt.  That's complete.

demiankatz:
Thanks, @maccabeelevine -- at this point I think we're mostly just waiting for @crhallberg's review when time permits (and it might take a little while -- we're having a very busy week here at Villanova :-) ).

One other thought, though probably not a hugely important one: do you think we need to come up with an equivalent mechanism for list-grid.phtml, in case a user has "grid mode" turned on? This is probably a rarely-used feature and it's certainly not a show stopper, but if you think it's worth figuring out, I'm happy to review code and discuss further.

maccabeelevine:
@demiankatz Hmm I didn't even realize "grid mode" existed.  I believe data-record-total will work either way so this is a question about data-record-number.  That said, I think the question of how far "down" the user scrolled/paged to click a record is trickier in a grid layout.  On a practical level looking at the code, there's no result wrapper to put the attribute on (other than the result itself) so there's no direct equivalent, we'd have to modify whatever getSearchResult() returns -- I didn't get that far -- and would have to pass into that function the index of the result, which seems to create a leaky abstraction.  Actually the best way to deal would probably to create a new wrapping div around the result, but I don't know what side effects that would have.  In short no I don't think it's worth figuring out without someone who cares, but feel free to push back on that.

demiankatz:
Thanks, @maccabeelevine, I agree that there's no point doing extra work without an interested user, but thought it might be worth at least raising awareness that this feature exists. :-) As soon as @crhallberg has a moment to look at this, I think we can get it merged.

crhallberg (approved):
I'm trying to think of other useful places to apply this transparency but I am coming up empty (for the context of useful analytics).

demiankatz:
Thanks, @crhallberg -- merging now.

maccabeelevine:
> I'm trying to think of other useful places to apply this transparency but I am coming up empty (for the context of useful analytics).

@crhallberg @demiankatz  I'm thinking of this as phase one of our custom analytics.  I'm not sure if there will be a phase two that needs more data attributes but if so I will certainly do another PR.  

I can also share the GTM tags and trigger that use them if there is interest.  They can be exported/imported as JSON.

PR #2668, opened 2023-01-09
Improve behavior/documentation of getSubfieldOrDefault

demiankatz:
I just discovered a bug which could result in null pointer exceptions in recent import refactoring. This should address the problem.

TODO

### DIFF
--- a/import/index_java/src/org/vufind/index/FormatCalculator.java
+++ b/import/index_java/src/org/vufind/index/FormatCalculator.java
@@ -548,7 +548,7 @@ public class FormatCalculator
     protected String getSubfieldOrDefault(DataField field, char subfieldCode, String defaultValue) {
         Subfield subfield = field.getSubfield(subfieldCode);
         String data = subfield != null ? subfield.getData() : null;
-        return (data == null || data == "") ? defaultValue : data;
+        return (data == null || data.isEmpty()) ? defaultValue : data;
     }
 
     protected boolean isOnlineAccordingTo338(Record record) {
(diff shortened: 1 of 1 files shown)


import/index_java/src/org/vufind/index/FormatCalculator.java
```diff
@@ -547,7 +547,7 @@ protected boolean hasSerialHost(Record record) {
 
     protected String getSubfieldOrDefault(DataField field, char subfieldCode, String defaultValue) {
         Subfield subfield = field.getSubfield(subfieldCode);
-        String data = subfield.getData();
+        String data = subfield != null ? subfield.getData() : null;
```
mtrojan-ub, line 558: Good catch! By the way, what happens if the subfield exists but contains only empty string? Do we also want to return the defaultValue in this case? (I'm not sure if this is a proper case anyway)

EreMaijala: Empty subfields happen even if they shouldn't, and I'd return the default value in that case.

demiankatz: I've added a simple empty subfield check. Is this what you had in mind? Note that I haven't had time to actually test it yet! I'll add a TODO to run the test suite before merging.

EreMaijala: Yes. I think `data.isEmpty()` would be the recommended way though, but I'd be happy to be corrected on this.

mtrojan-ub: I also think data.isEmpty() is the correct way to check for an empty string, else it might compare the pointer instead of the actual string value

demiankatz: Corrected -- thanks!

mtrojan-ub (approved):

EreMaijala (approved):

demiankatz:
All tests are passing, so I've merged this. Thanks again!

PR #2669, opened 2023-01-12
Corrects non-validating cart dropdown menus

ckaz:
base: release-8.1
 head-fork: finc/vufind
 compare: pullrequest_cart-dropdown-menus

This corrects two non-validating dropdown menus in cart.phtml by adding aria roles required by role="menu" and adding href content. The bug is present in all versions since at least VuFind 6. It is a validation bug, the menu is functional without the fix. Tested with Axe DevTools on Chrome with the respective menus open.

Another issue with this is that when tabbing through the menu, the dropdown-togglers don't get proper focus highlighting because of 
```
.dropdown-toggle:focus {
  outline: 0;
}
```

in Bootstraps own _dropdown.scss -- I'd prefer a normal focus outline for this but could not decide on the correct place to put it.

TODO
- [ ] Figure out how to deal with focus highlighting issue
- [x] Resolve conflicts
- [x] Run full test suite


demiankatz:
Thanks, @ckaz, this makes sense to me, but I'll hold off on giving it a closer look until @crhallberg has a chance to weigh in. Additionally, it looks like there is a conflict of some sort, probably a result of moving this from release-8.1 to dev, where the icon helper changes have introduced some inconsistencies. Let us know if you need help resolving the conflicts! We should also run the full integration test suite on this after that is sorted out. I'll add some TODO checkboxes to keep track of where things stand.

crhallberg:
Yeah, removing outline is very outdated due to inaccessibility. We could probably override it with something like:

```css
.dropdown-toggle:focus {
  outline: unset;
}
```

themes/bootstrap3/templates/cart/cart.phtml
```diff
@@ -18,37 +18,45 @@
...
           <?=$this->transEsc('Save')?>
         </button>
       <?php endif; ?>
-      <button type="submit" class="toolbar-btn btn-type-email" name="email" value="1" title="<?=$this->transEscAttr('bookbag_email')?>" value="1">
+      <button type="submit" class="toolbar-btn btn-type-email" name="email" title="<?=$this->transEscAttr('bookbag_email')?>" value="1">
         <?=$this->transEsc('Email')?>
       </button>
       <?php $exportOptions = $this->export()->getActiveFormats('bulk'); if (count($exportOptions) > 0): ?>
-        <button type="submit" class="toolbar-btn btn-type-export" name="export" value="1" title="<?=$this->transEscAttr('bookbag_export')?>" value="1">
+        <button type="submit" class="toolbar-btn btn-type-export" name="export" title="<?=$this->transEscAttr('bookbag_export')?>" value="1">
```
crhallberg: I appreciate the catch on the double value, but why do we need value on these buttons? Is it just for validation?

themes/bootstrap3/templates/cart/cart.phtml
```diff
@@ -18,37 +18,45 @@
...
           <?=$this->transEsc('Delete')?>
         </button>
         <ul class="dropdown-menu" role="menu" aria-labelledby="cart-delete-label">
-          <li><a id="cart-confirm-delete" title="<?=$this->transEscAttr('confirm_delete')?>"><?=$this->transEsc('confirm_dialog_yes')?></a></li>
-          <li><a><?=$this->transEsc('confirm_dialog_no')?></a></li>
+          <li role="none">
+            <a href="javascript:" id="cart-confirm-delete" title="<?=$this->transEscAttr('confirm_delete')?>" role="menuitem"><?=$this->transEsc('confirm_dialog_yes')?></a>
+          </li>
+          <li role="none">
+            <a href="javascript:" role="menuitem"><?=$this->transEsc('confirm_dialog_no')?></a>
```
crhallberg: Is `href="javascript:"` a modern version of `href="#"`?

crhallberg (approved):
I have a few questions (for my own education) but for the sake of HTML validation, I see no problems with these changes.

themes/bootstrap3/templates/cart/cart.phtml
demiankatz: I believe that the value is needed by the cart processor -- it knows which action to execute based on which button is pressed and thus which value is received. See: https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/Controller/CartController.php#L82-L96

themes/bootstrap3/templates/cart/cart.phtml
ckaz: Not really. We got used to using it where an anchor action is triggered using JS -- no matter which one uses, anchors without any href value used to cause issues with screen readers and tabbing access/being non-focussable.

ckaz:
> Additionally, it looks like there is a conflict of some sort, probably a result of moving this from release-8.1 to dev, where the icon helper changes have introduced some inconsistencies. Let us know if you need help resolving the conflicts! We should also run the full integration test suite on this after that is sorted out.

Hi Demian, would the right way forward be: Checking this out against "dev" rather than "release_8.1" and resolving conflicts and pushing this again?

demiankatz:
@ckaz, you should be able to simply merge dev into this branch, resolve the conflicts, and push up the result (which is perhaps exactly what you suggest -- but there's no need to create anything new, you can just merge in the dev branch... so, for example, if your Git remote for the core project is called "upstream" you would do: `git fetch upstream ; git merge upstream/dev` then resolve conflicts and push). Does that help?

themes/bootstrap3/templates/cart/cart.phtml
crhallberg: If I recall correctly, the cart knows which button was pressed thanks to this bit of Javascript: https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/cart/cart.phtml#L67-L70.

Maybe this is something we look to refactor in the future. There's probably a better way.

demiankatz: The cart currently works with Javascript disabled. Whether we need to maintain that is a separate question, but the original design was meant to work without requiring Javascript.

ckaz:
> @ckaz, you should be able to simply merge dev into this branch, resolve the conflicts, and push up the result (which is perhaps exactly what you suggest -- but there's no need to create anything new, you can just merge in the dev branch... so, for example, if your Git remote for the core project is called "upstream" you would do: `git fetch upstream ; git merge upstream/dev` then resolve conflicts and push). Does that help?

That sounds easy. Extremely sorry for the delay -- I'll try to get it done this week. Cheers!

ckaz:
Just merged dev into the PR -- once again, sorry for the delay. We are chock-a-block ...

crhallberg (approved):
Thank you! This seems to have been overlooked in the big icon helper fix.

demiankatz (approved):
@ckaz, this did not work correctly for me as written -- the "javascript:" links did not stop the browser from loading invalid URLs and causing errors. However, a bit of research suggests that `javascript:void(0);` is more appropriate syntax to create a null link, and after making that revision, everything seems to work for me, and the full test suite is passing.

Any objections to this change before I merge this? I'm also interested in hearing from @crhallberg again if he has any opinions.

EreMaijala:
@demiankatz I suspect that won't work with content security policy enabled.

demiankatz (changes requested):
Good point, @EreMaijala. I think we need to figure out a better way to create void links. I also wonder if we should refactor this entire control (the drop down yes/no menu) into a component using @crhallberg's new component view helper, since this is a recurring pattern used in several parts of the application, and it might be cleaner to solve the problem once and apply that solution universally through a shared template.

demiankatz:
In the interest of moving things forward, I have merged the fix to the duplicate value attributes as c20ed3c56e6962560d54fef20b75014a84a73320, leaving behind the drop-down adjustments which still need further work.

ckaz:
> Good point, @EreMaijala. I think we need to figure out a better way to create void links. I also wonder if we should refactor this entire control (the drop down yes/no menu) into a component using @crhallberg's new component view helper, since this is a recurring pattern used in several parts of the application, and it might be cleaner to solve the problem once and apply that solution universally through a shared template.

We could simply turn the anchors into buttons and in this way get rid of any empty links? Having buttons there would make more sense anyway, I guess.

EreMaijala:
@ckaz The typical way has been to use `href="#"` and ensure that further processing is prevented (like `$e.preventDefault();`) in the event handler.

demiankatz:
I'm inclined to agree with @ckaz that buttons would be more semantically appropriate than links here, and I don't think the Javascript would have to change at all if we restructured from links to buttons. Cosmetically, we'd want to be sure that the change didn't alter the appearance of the drop-down menu too dramatically, but I'm pretty sure we already have a class that styles buttons to look like links. @crhallberg probably remembers what it is called if it's not easy to find.

I think it's then a separate question whether we would like a separate component template called something like drop-down-confirmation-menu to abstract this functionality. I think there might be a benefit, but we could address that in a separate PR. What does everyone think?

crhallberg:
1. Yes, in the case where a link doesn't have a valid href (or needs to cancel or void it), a button is more appropriate semantically (and accessibly).
2. A confirmation dropdown would be good in this case, especially if it is as simple as submitting the form containing it.

demiankatz:
@crhallberg, do you have time to open a PR to introduce the confirmation template discussed here, so we can move this process forward?

demiankatz:
I'm closing this as superseded by #2723 (though that work is still in progress -- just trying to clean up the number of PR's pinned to the 9.0 milestone, and I don't expect this work to be merged in its current form).

PR #2670, opened 2023-01-13
Fix for query string character prefix

welblaud:
This fixes a minor bug in getting params of the hidden filters.

Previously, the query string was cached including the prefix char (default `&amp;`), hence the intentional prefixing did not work (specifically in the [searchbox](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/search/searchbox.phtml#L29)).


demiankatz:
Thanks, @welblaud -- some of this logic was refactored recently, so maybe this bug was introduced at that time. I'm a bit short on bandwidth today; @EreMaijala, is there any chance you could take a look at this? Otherwise, I'll try to give it more attention next week.

EreMaijala (approved):
@welblaud Thanks for catching this, it was indeed caused by my recent refactoring. The changes look good to me.

It also seems that we're missing a unit test here. I'll take care of that in a separate pull request.

PR #2671, opened 2023-01-13
Hide "searches" in history that are made for the print view

jpkanter:
Reproduction: 
1.) Add Elements to the book bag from search
2.) open book bag, select elements and click print
3.) print view will open
4.) open search history
5.) search "5 elements" will be present

As far as i can see printing creates a fake search with a fixed number of IDs and goes through the normal processes via the `RecordsController->HomeAction()` which in turn calls `resultsAction()` which is mostly inherited from `AbstractSearch`

By intercepting the saving to the search history with an additional condition the bloating of the search history can be avoided.

TODO
- [x] Add mink test
- [x] Run full test suite

### DIFF
--- a/module/VuFind/src/VuFind/Controller/AbstractSearch.php
+++ b/module/VuFind/src/VuFind/Controller/AbstractSearch.php
@@ -380,16 +380,13 @@ class AbstractSearch extends AbstractBase
             if ($jump = $this->processJumpTo($results)) {
                 return $jump;
             }
-            
-            // do not remember in case of print view
-            if( !$this->params()->fromQuery('print') ) { 
-                // Remember the current URL as the last search.
-                $this->rememberSearch($results);
 
-                // Add to search history:
-                if ($this->saveToHistory) {
-                    $this->saveSearchToHistory($results);
-                }
+            // Remember the current URL as the last search.
+            $this->rememberSearch($results);
+
+            // Add to search history:
+            if ($this->saveToHistory) {
+                $this->saveSearchToHistory($results);
             }
 
             // Set up results scroller:
--- a/module/VuFind/src/VuFind/Controller/AbstractSearch.php
+++ b/module/VuFind/src/VuFind/Controller/AbstractSearch.php
@@ -380,13 +380,16 @@ class AbstractSearch extends AbstractBase
             if ($jump = $this->processJumpTo($results)) {
                 return $jump;
             }
+            
+            // do not remember in case of print view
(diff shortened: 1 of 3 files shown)


demiankatz:
Thanks, @jpkanter! I'm a bit short on time this week but will try to take a closer look at this next week. We should probably add a Mink integration test to capture the behavior and prevent regressions. I also suspect that there might be a more generic way of preventing this behavior instead of depending specifically on detecting the print parameter (though maybe not -- I'll need to give it a little more thought). In any case, thanks for reporting this and getting started on a fix!

demiankatz:
Sorry, @jpkanter, I still haven't quite gotten to this (we migrated to FOLIO last week, so it's been unusually busy over here!). It's still on my radar, though, and I'll offer more feedback as soon as I'm able.

demiankatz:
@jpkanter, I took another look at this, and I think I see a slightly more targeted way to approach the problem. What do you think of revising \VuFind\Controller\RecordsController::homeAction() like this:

```
    /**
     * Home action -- call standard results action
     *
     * @return mixed
     */
    public function homeAction()
    {
        // If there is exactly one record, send the user directly there:
        $ids = $this->params()->fromQuery('id', []);
        $print = $this->params()->fromQuery('print');
        if (count($ids) == 1) {
            $details = $this->getRecordRouter()->getTabRouteDetails($ids[0]);
            $target = $this->url()->fromRoute($details['route'], $details['params']);
            // forward print param, if necessary:
            $params = empty($print) ? '' : '?print=' . urlencode($print);
            return $this->redirect()->toUrl($target . $params);
        }
        // We don't want to remember print jobs in the search history:
        if (!empty($print)) {
            $this->saveToHistory = false;
        }
        // Not exactly one record -- show search results:
        return $this->resultsAction();
    }
```

This method was already doing analysis related to the print parameter, and this allows us to disable search history based on the parameter in the specific place where we expect the parameter to be received, instead of changing more general, lower-level code. I believe that the net effect is the same.

What do you think?

demiankatz:
(If you'd like me to push this change to a branch somewhere for you, or create a new PR, I'm happy to do so... but if you'd rather revise this one, that's fine with me too!).

jpkanter:
You method seems better, i was already feeling uncomfortable changing such a low level function for a somewhat specific change.

demiankatz (approved):
Thanks again, @jpkanter. I've added a Mink test here to prevent future regressions. I'll merge this as soon as the CI build passes.

PR #2672, opened 2023-01-13
Add tests for getCurrentHiddenFilterParams.

EreMaijala:
As promised in #2670, here are tests to cover SearchTabs helper's getCurrentHiddenFilterParams method. The helper could use more test coverage, but at least this is a start.


demiankatz (approved):
Thanks, @EreMaijala! As you say, we could stand to add more tests as time permits, but progress is better than nothing, so I'm merging this now. :-)

PR #2673, opened 2023-01-16
Fix PHP 8.1 compatibility of hierarchy tree formatter.

EreMaijala:
Trying to get titles in hierarchy for the top-level record failed with TypeError since it doesn't have hierarchy_parent_id. Also avoids getting titles in hierarchy unless the node has a parent ID.


EreMaijala:

demiankatz (approved):
Thanks, @EreMaijala -- looks good to me!

PR #2674, opened 2023-01-17
Add ability to pass params into forms

xmorave2:
This enables to create links with urls like `/Feedback/Form/SomeForm?param=value` to prefill inputs for user, or fill hidden inputs with some dynamic value.


demiankatz:
Thanks, @xmorave2! I'm curious whether @EreMaijala has any thoughts on this, since he's more invested in the form code than I am. I also wonder whether we need to add test coverage for this new functionality, and whether we should add a configuration setting to limit which parameters may be pre-filled via URL (in case there are any possible abuses of the functionality that we might want to avoid). Any thoughts?

EreMaijala (changes requested):
This is a good idea, but @demiankatz has valid points. We need to protect at least the read-only reserved elements (referrer, useragent, submit). Also need to ensure that this doesn't allow overriding a select field with an invalid value (I didn't test it yet), and here test coverage would help.

xmorave2:
Thanks for review.

I added configuration for enabling prefill only for specific fields and check if some of protected fields are not prefilled.

I didn't implement prefilling for selects and checkboxes - it turned out a bit complicated and we don't even need for our use case.

config/vufind/FeedbackForms.yaml
```diff
@@ -65,6 +65,14 @@
 #
 #     Help texts may contain HTML-formatting.
 #
+#   prefillFields (array) Fields that could be prefilled by passing parametr to form URL. Could be
```
demiankatz: Typo: parametr --> parameter

module/VuFind/src/VuFind/Form/Form.php
```diff
@@ -723,6 +725,11 @@ protected function parseConfig($formId, $config, $params)
                 }
             }
 
+            $prefill = $this->sanitizePrefill($prefill);
```
demiankatz: Should we do this at the top of the function, so we only sanitize once instead of doing it on every loop iteration?

module/VuFind/src/VuFind/Form/Form.php
```diff
@@ -1135,4 +1157,29 @@ public function getFormId(): string
...
+    /**
+     * Validates prefill data and returns only the prefill values for enabled fields
+     *
+     * @param array $prefill Prefill data
+     *
+     * @return array
+     */
+    protected function sanitizePrefill(array $prefill): array
+    {
+        foreach ($this->getProtectedFieldNames() as $fieldName) {
```
demiankatz: You could probably eliminate this extra foreach loop by incorporating protected fields into the array_filter callback below -- e.g. change that to `return in_array($key, $prefillFields) && !in_array($key, $this->getProtectedFieldNames())`

module/VuFind/tests/unit-tests/src/VuFindTest/Form/FormTest.php
```diff
@@ -843,4 +845,116 @@ public function testEmailSubjects(
             $form->getEmailSubject($form->getData())
         );
     }
+
+    public function testPrefill(): void
```
demiankatz, line 854: Should put a function header on this test.

module/VuFind/tests/unit-tests/src/VuFindTest/Form/FormTest.php
```diff
@@ -843,4 +845,116 @@ public function testEmailSubjects(
...
+                    'type' => 'submit',
+                    'name' => 'submit',
+                    'label' => 'Send',
+                ],
+            ],
+            $form->getFormElementConfig()
+        );
+    }
+
+    public function testPrefillProtectedFields(): void
```
demiankatz, line 920: Also needs a function header comment.

demiankatz (changes requested):
Thanks for the progress on this! I have just a few minor suggestions.

xmorave2:
Thanks  @demiankatz  I fixed them all.

demiankatz (approved):
Thanks, @xmorave2! I'll wait to see if @EreMaijala has any further thoughts before merging this.

EreMaijala (approved):
Looks good to me now!

PR #2676, opened 2023-01-20
Add semantics to fines table for screen readers

stephanieleary:
- Adds `<thead>`, `<tbody>`, and `<tfoot>` to better distinguish headers from body cells
- Adds `scope="col"` to column headers
- Changes cells to headers in the footer row (the fine totals) and removes inline bold styling


themes/bootstrap3/templates/myresearch/fines.phtml
```diff
@@ -67,22 +67,28 @@
...
           <?php endforeach; ?>
         </tr>
       <?php endfor; ?>
-      <tr style="font-weight:bold">
-        <td colspan="<?=count($columns) - 1?>"><?=$this->transEsc('Total Balance Due')?></td>
-        <td><?=$this->safeMoneyFormat($totalDue / 100.00) ?></td>
+      </tbody>
+      <tfoot>
+      <tr>
+        <th colspan="<?=count($columns) - 1?>"><?=$this->transEsc('Total Balance Due')?></th>
```
demiankatz: Should this have a scope of row, or is that implied?

themes/bootstrap3/templates/myresearch/fines.phtml
```diff
@@ -67,22 +67,28 @@
...
+        <th><?=$this->safeMoneyFormat($totalDue / 100.00) ?></th>
```
demiankatz: I think this might make more semantic sense as a td with bold styling, rather than a th, since this is a value rather than a heading.

demiankatz (changes requested):
Thanks, @stephanieleary, this is definitely an improvement, but I have a couple of suggestions for further improvement. Of course, feel free to disagree if either of my suggestions are off-base! :-)

themes/bootstrap3/templates/myresearch/fines.phtml
stephanieleary: Since it summarizes the contents of all the cells above it, I think it should be `<th>`. (This is almost exactly the demo on the [MDN `<tfoot>` reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tfoot).)

themes/bootstrap3/templates/myresearch/fines.phtml
stephanieleary: It should, yes... and actually, the fines' names should probably be scoped headers as well. I will amend that!

themes/bootstrap3/templates/myresearch/fines.phtml
demiankatz: @stephanieleary, sorry if I'm missing something, but in the MDN example you link to, the table cell in the tfoot containing the actual total value is a td, not a th, which seems to align with what I was originally suggesting. Is it possible we're just misunderstanding each other and not talking about the same cell? :-)

demiankatz: (I definitely agree that the other footer cell, containing the label "Total Balance Due," unquestionably should be a th).

stephanieleary: Yes, I was definitely looking at the wrong cell! My apologies.

demiankatz: No worries, glad we were able to get on the same page. :-) I'm going to run the test suite now just to be sure none of these changes require adjustments to integration tests, and if that goes well, I'll merge in a few minutes. Thanks again for these improvements!

demiankatz (changes requested):
@stephanieleary, all the tests are passing, but I notice there are some cosmetic differences that might be worth discussing:

1.) All the titles are bold now because of default `th` styling. I don't think this is necessarily a problem (it doesn't bother me), but just noting that it is a change in case we want to change it in some way.

2.) For some reason, the upper border styling is messed up by these changes. Look at before:

image

and after:

image

(Note how we now have a double-thick border under the header, instead of borders above and below it).

I'm seeing this issue in all three of VuFind's default themes (bootstrap3, bootprint3 and sandal), so I think it must somehow be related to some styling in the base bootstrap3 theme (which the others inherit from).

We should probably figure this out before merging. If you're not sure where to look, @crhallberg may be able to help with this, though he's out of office until the end of the week, so it may take a little bit.

Sorry for the delay, thanks again for the progress, and please let me know if I can do more to help!

crhallberg:
> 2.) For some reason, the upper border styling is messed up by these changes.

This is the [default way that Bootstrap styles `thead th`](https://getbootstrap.com/docs/3.3/css/#tables-example), with a bold bottom border and no border on top.

We could add the top border back in with something like

```less
.table > thead > tr > th {
    border-top: 1px solid @table-border-color;
}
```

crhallberg (approved):

demiankatz:
Thanks, @crhallberg -- I guess the question then becomes, do we prefer the new styling and merge as-is, or do we apply the changes you suggested? I don't have strong preferences. What do you think, @stephanieleary?

stephanieleary:
It's probably best to add the style @crhallberg suggested so as not to introduce any unwanted visual changes. I'll be working on all the other tables in the account screens next week, so I think a change to the main stylesheet is warranted.

demiankatz:
@stephanieleary, I did a bit more investigation and confirmed that, as @crhallberg reported, this styling is intentional for tables with a `<thead>`. Since we already have some existing instances of such tables in the code, there is no way to avoid visual changes. Either we change the existing table headers to look different, or we change the fines to look like the other table headers. I think changing the appearance of the fine table is actually the least disruptive approach and feels appropriate now that I realize it's a design choice and not a bug. Of course, if anyone dislikes it, they're free to tweak their local CSS as needed!

demiankatz (approved):
I just made one small tweak to indent the HTML in a more uniform way. I'll merge as soon as CI passes. Thanks again!

PR #2677, opened 2023-01-23
SierraRest: Add dueStatus field to transactions.

EreMaijala:
SierraRest driver did not return the dueStatus field for transactions. This caused account ajax notifications to miss notifications for items due or overdue.


demiankatz (approved):
Thanks, @EreMaijala, looks good to me. Probably wouldn't hurt to add unit test coverage for this new method since I think it would be pretty easy to do, but I'm not going to hold up the merging process for that since the SierraRestTest is pretty bare-bones, and increasing its coverage probably makes more sense as a separate project.

PR #2678, opened 2023-01-23
Add "my account" sidebar to "edit list item"

srajaei:
"Edit item" of Favorites misses the your account sidebar

TODO
- [x] Run full test suite

PR #2679, opened 2023-01-23
Fix consistency of list menu highlighting/breadcrumbs.

demiankatz:
While testing and reviewing #2678, I noticed that editing a list did not highlight the correct menu item, and that the template for editing a list item included a $listFilter variable that was never set. This PR adds the missing variable, uses it for consistent menu highlighting, and also fixes the list highlighting problem.


module/VuFind/src/VuFind/Controller/MyResearchController.php
```diff
@@ -943,9 +943,7 @@ public function editAction()
         }
 
         return $this->createViewModel(
-            [
-                'driver' => $driver, 'lists' => $lists, 'savedData' => $savedData
-            ]
+            compact('driver', 'lists', 'savedData', 'listFilter')
```
EreMaijala: Is there a reason for calling the view parameter listFilter instead of listID? Filter sounds like it could be something more than a single list ID.

themes/bootstrap3/templates/myresearch/edit.phtml
```diff
@@ -6,11 +6,11 @@
   $this->layout()->breadcrumbs = '<li><a href="' . $this->url('myresearch-home') . '">' . $this->transEsc('Your Account') . '</a></li> <li class="active">' . $this->transEsc('Edit') . '</li>';
```
EreMaijala: Could the breadcrumb link be fixed as well now that the list id is known?

EreMaijala (changes requested):

module/VuFind/src/VuFind/Controller/MyResearchController.php
demiankatz: The template already expected a parameter called listFilter, which I think is a legacy of VuFind 1.x; I kept the name for backward compatibility... but since this parameter has been orphaned/broken for many years, it's probably better to change it to something more understandable. I will revise!

themes/bootstrap3/templates/myresearch/edit.phtml
demiankatz: Good idea -- done!

EreMaijala (approved):

EreMaijala:
@demiankatz Looks good to me. A very slight concern regarding translations is that this makes it more difficult to find where a translation string is used.

demiankatz:
Thanks, @EreMaijala -- you make a good point about searchability of translation strings, so I've reverted that optimization for now. If we want to switch that to use a ternary in future, we might just want to replace those long strings with tokens... but I won't change that right now.

PR #2680, opened 2023-01-25
KohaRest: Add an option to allow canceling of holds in transit.

EreMaijala:
Some libraries want to allow canceling of holds in transit to pick up location. Koha seems to support this, so this adds a  configuration option to support it in VuFind.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
@@ -4,7 +4,7 @@
  *
  * PHP version 7
  *
- * Copyright (C) The National Library of Finland 2016-2020.
+ * Copyright (C) The National Library of Finland 2016-2023.
  * Copyright (C) Moravian Library 2019.
  *
  * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -103,6 +103,13 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
      */
     protected $defaultPickUpLocation;
 
+    /**
+     * Whether to allow canceling holds in transit. Default is false.
+     *
+     * @var bool
+     */
+    protected $allowCancelInTransit;
```
demiankatz: Probably wouldn't hurt to initialize this to false to ensure it's not unset if init() is somehow bypassed (unlikely, but maybe possible in testing, etc.).

demiankatz (changes requested):
Thanks, @EreMaijala, makes sense to me. Just one trivial suggestion before we merge. :-)

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: Done. I added another one as well while at it.

demiankatz (approved):
Thanks, @EreMaijala!

PR #2682, opened 2023-01-25
FOLIO: Add patron comment support for requests

demiankatz:
I noticed that we didn't support patron comments on requests yet; this PR adds the feature. I'd be interested in feedback from other FOLIO users. I have confirmed that this works as expected for Villanova's instance.

### DIFF
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -97,7 +97,7 @@ HMACKeys = holding_id:item_id:status:level
 defaultRequiredDate = 0:1:0
 
 ; extraHoldFields - A colon-separated list used to display extra visible fields in the
-; place holds form. Supported values are "requiredByDate",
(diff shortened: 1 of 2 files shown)


bbusenius (approved):
This looks fine to me. I have something similar in our local module that I added right before we launched. Mine is a little different because I added it before the title level holds stuff was in the code, however, the substance of it is the same.

maccabeelevine (approved):
Yes, without testing, but it's good functionality.  Lehigh has this as a customization as well, along with some other custom fields. 
image

demiankatz:
Thanks, both of you -- merging now!

On a different subject, have either of you done anything with proxy borrowers yet? That might be my next project....

maccabeelevine:
@demiankatz I don't believe we have at Lehigh.

PR #2683, opened 2023-01-25
FOLIO: more flexible hold exclusion configuration

meganschanz:
This adds a new config in the FOLIO driver to optionally exclude hold location matches by substrings instead of full string matches. This config is set to false by default for backwards compatibility.

This was useful for us because we have a dozens of locations we need to exclude that either start with or contain a string (such as "Reserve" or "Special Collection") but the full location have more specific qualifiers in them like the floor and the list would be quite long and difficult to maintain anytime a new location was added or changed.

### DIFF
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -113,6 +113,7 @@ extraHoldFields = requiredByDate:pickUpLocation
 ; See https://issues.folio.org/browse/UXPROD-2422.
 ; Checks against the location's Discovery Display Name in FOLIO.
 ; If no Discovery Display Name is configured, checks against name.
+; Legal excludeHoldLocationsCompareMode settings: exact, regex; see examples below.
 ;excludeHoldLocationsCompareMode = exact
 ; Exact mode examples
 ;excludeHoldLocations[] = "24 Hour Reserve"
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -107,15 +107,18 @@ extraHoldFields = requiredByDate:pickUpLocation
 
 ; Hide the place hold/recall/page link when an item is in the configured
 ; list of locations. This can either match the full location name exactly or
-; do a substring match depending on how you set the excludeHoldLocationsBySubString
+; do a regex match depending on how you set the excludeHoldLocationsCompareMode
 ; setting. Ideally, this would not be needed and FOLIO would include an
 ; indicator of whether or not an item is holdable with item details.
 ; See https://issues.folio.org/browse/UXPROD-2422.
 ; Checks against the location's Discovery Display Name in FOLIO.
 ; If no Discovery Display Name is configured, checks against name.
-;excludeHoldLocationsBySubString = false
+;excludeHoldLocationsCompareMode = exact
+; Exact mode examples
 ;excludeHoldLocations[] = "24 Hour Reserve"
 ;excludeHoldLocations[] = "Reference Collection"
+; Regex mode example
+;excludeHoldLocations[] = "/.*RESERVE.*/i"
 
 ; When a request is cancelled through VuFind, use this cancellation reason ID. Most users
 ; will not have to change this ID unless they have replaced the cancellation reason
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -106,11 +106,14 @@ extraHoldFields = requiredByDate:pickUpLocation
 ;default_request = Hold
 
 ; Hide the place hold/recall/page link when an item is in the configured
-; list of locations. Ideally, this would not be needed and FOLIO would include an
+; list of locations. This can either match the full location name exactly or
+; do a substring match depending on how you set the excludeHoldLocationsBySubString
+; setting. Ideally, this would not be needed and FOLIO would include an
 ; indicator of whether or not an item is holdable with item details.
 ; See https://issues.folio.org/browse/UXPROD-2422.
 ; Checks against the location's Discovery Display Name in FOLIO.
 ; If no Discovery Display Name is configured, checks against name.
+;excludeHoldLocationsBySubString = false
 ;excludeHoldLocations[] = "24 Hour Reserve"
 ;excludeHoldLocations[] = "Reference Collection"
 
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -483,10 +483,9 @@ class Folio extends AbstractAPI implements
                 }
             }
             return true;
-        } else {
-            // Otherwise exclude checking by exact match
-            return !in_array($locationName, $excludeLocs);
         }
+        // Otherwise exclude checking by exact match
+        return !in_array($locationName, $excludeLocs);
     }
 
     /**
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -463,14 +463,13 @@ class Folio extends AbstractAPI implements
      */
     protected function isHoldable($locationName)
     {
-        $setSubSt = (bool)$this->config['Holds']['excludeHoldLocationsBySubString']
-                    ?? false;
+        $mode = $this->config['Holds']['excludeHoldLocationsCompareMode'] ?? 'exact';
         $excludeLocs = (array)$this->config['Holds']['excludeHoldLocations'] ?? [];
 
-        // Exclude Checking by substring match
-        if ($setSubSt) {
-            foreach ($excludeLocs as $exclude) {
-                if (strpos($locationName, $exclude) !== false) {
+        // Exclude checking by regex match
+        if ($mode == "regex") {
+            foreach ($excludeLocs as $pattern) {
+                if (preg_match($pattern, $locationName) === 1) {
                     return false;
                 }
             }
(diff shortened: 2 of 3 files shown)


meganschanz:
Rookie mistake! I'm working on cleaning up the format issues and I'll make another commit tomorrow.

demiankatz (commented):
Thanks, @meganschanz -- I'm happy to merge this as is, but a suggestion for your consideration first: what if, instead of a binary excludeHoldLocationsBySubString  setting, we had a excludeHoldLocationsCompareMode setting, with "exact" or "substring" options? This would open the way to adding additional options (like a "regex" mode) if we need them, without having to further redesign the configuration. Or we could implement regex mode now, and make "substring" a special case of regex (we could just add `/.* ... .*/` around each value. Case sensitivity is another question: do we want to offer case-sensitive and case-insensitive versions? Using regular expressions makes it easier to configure this without having to add lots of different modes, since you can control it with regex switches.

meganschanz:
Good idea. I much prefer that. I took your suggested and made a commit that has a setting with `excludeHoldLocationsCompareMode` with the default value of `exact` but allowing for `regex`.

config/vufind/Folio.ini
```diff
@@ -106,13 +106,19 @@ extraHoldFields = requiredByDate:pickUpLocation
...
 ; Hide the place hold/recall/page link when an item is in the configured
-; list of locations. Ideally, this would not be needed and FOLIO would include an
+; list of locations. This can either match the full location name exactly or
+; do a regex match depending on how you set the excludeHoldLocationsCompareMode
+; setting. Ideally, this would not be needed and FOLIO would include an
 ; indicator of whether or not an item is holdable with item details.
 ; See https://issues.folio.org/browse/UXPROD-2422.
 ; Checks against the location's Discovery Display Name in FOLIO.
 ; If no Discovery Display Name is configured, checks against name.
+;excludeHoldLocationsCompareMode = exact
```
demiankatz, line 117: Might be worth explicitly listing legal values in a comment here -- the comment above mentions this setting but isn't totally explicit about what settings are supported (e.g. `; Legal excludeHoldLocationsCompareMode settings: exact, regex; see examples below.`).

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -463,10 +463,24 @@ public function getConfig($function, $params = [])
...
     {
-        return !in_array(
-            $locationName,
-            (array)($this->config['Holds']['excludeHoldLocations'] ?? [])
-        );
+        $mode = $this->config['Holds']['excludeHoldLocationsCompareMode'] ?? 'exact';
+        $excludeLocs = (array)$this->config['Holds']['excludeHoldLocations'] ?? [];
+
+        // Exclude checking by regex match
+        if ($mode == "regex") {
```
demiankatz: Might be worth doing `trim(strtolower($mode))` just to be a little more tolerant of non-standard input (particularly since the comments in the file capitalize "Exact" and "Regex" in the headings above the example, which might lead someone to enter a capitalized version).

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -463,10 +463,24 @@ public function getConfig($function, $params = [])
...
+            foreach ($excludeLocs as $pattern) {
+                if (preg_match($pattern, $locationName) === 1) {
+                    return false;
+                }
+            }
+            return true;
+        } else {
+            // Otherwise exclude checking by exact match
+            return !in_array(
```
demiankatz: You can probably collapse this function call into a single line -- I think it was spread out before because $excludeLocs used to be a long expression, but it should be readable in a more compact presentation now.

demiankatz (changes requested):
Looking great, @meganschanz! A few minor suggestions for finishing touches. It might also be nice to add some test coverage for the new behavior. If you'd like to try that and are not yet familiar with VuFind's test suite, I can point you to some resources and make some suggestions. If you don't have time, I can probably add a quick test myself before merging this.

meganschanz:
Those suggestions look good. I just started in on writing the unit tests and ran into a problem, the `isHoldable` function is protected. Is there a preferred way for tested those. I could have it call the public `getHolding` function and then just check the `is_holdable` attribute of item, but that seems like more overhead than necessary. 

```
Error: Call to protected method VuFind\ILS\Driver\Folio::isHoldable() from scope VuFindTest\ILS\Driver\FolioTest
```

demiankatz:
Thanks, @meganschanz! If you need to test a protected method, and there's not a straightforward way to test it via a public method, you can mix in the [ReflectionTrait](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFindTest/Feature/ReflectionTrait.php) and use $this->callMethod().

meganschanz:
I learned so much about the testing today... that `vendor/bin/phing qa-php` command that the CI runs is handy!

Let me know if you have suggestions for any rework that could be done to the tests. When working on the tests, I also added in a small change to the `isHoldable` function to better handle the scenario when a bad regex pattern is set in the config.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -463,10 +463,30 @@ public function getConfig($function, $params = [])
...
+                        $pattern
+                    );
+                    continue;
+                }
+                if ($match === 1) {
+                    return false;
+                }
+            }
+            return true;
+        } else {
```
demiankatz: Very minor point of style: since you're guaranteed to return some value in the if condition above, you don't really need an else here -- you can just return the default check at the bottom of the function, which shaves off a little bit of extra code.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/FolioTest.php
```diff
@@ -500,4 +501,69 @@ public function testSingleItemGetMyHolds(): void
...
+    public function testIsHoldable(): void
+    {
+        $driverConfig = $this->defaultDriverConfig;
+
+        // Positive test for exact compare mode
+        $driverConfig['Holds']['excludeHoldLocations'] = ['reserve'];
+        $driverConfig['Holds']['excludeHoldLocationsCompareMode'] = 'exact';
+        $this->createConnector("empty", $driverConfig);
+
+        $this->assertEquals($this->callMethod($this->driver, "isHoldable", ["reserve"]), false);
```
demiankatz: In PHPUnit assertions, the expected value comes before the actual value -- this is important because if an assertion fails, you want to be sure the failure message puts the right values in the right places. So this should actually be:

```
$this->assertEquals(false, $this->callMethod($this->driver, "isHoldable", ["reserve"]));
```

Except that there are actually special assertions for true/false, so you can further simplify down to:

```
$this->assertFalse($this->callMethod($this->driver, "isHoldable", ["reserve"]))
```

...so many of your assertions in this method can become just `$this->assertTrue()` or `$this->assertFalse()`.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/FolioTest.php
```diff
@@ -500,4 +501,69 @@ public function testSingleItemGetMyHolds(): void
...
+        $this->assertEquals($this->callMethod($this->driver, "isHoldable", ["Reserve"]), true);
+        $this->assertEquals($this->callMethod($this->driver, "isHoldable", ["library"]), true);
+
+        // Test default mode is exact
+        unset($driverConfig['Holds']['excludeHoldLocationsCompareMode']);
```
demiankatz: I would also suggest breaking down this test into many smaller tests -- rather than one long testIsHoldable(), it might be better to have testIsHoldableExactMode(), testIsHoldableDefaultBehavior(), testIsHoldableExactCaseSensitivity(), etc., etc.

While there's a little more overhead in terms of having lots of function headers in the file, it will be helpful in future if any tests fail, since we'll be able to more easily pinpoint what is wrong. In particular, if you have one long test, it's going to stop as soon as something goes wrong. If you have many tests, each test will run independently, which can provide more information -- e.g. if all the isHoldable tests start failing, that tells us one thing; if just one fails, that tells us another. The distinction can be useful!

demiankatz (changes requested):
Thanks for the ongoing progress, @meganschanz -- this is looking great, and could nearly be merged as it is... but since you asked about test style, I provided some suggestions that I think will make this even better. I also had one small stylistic comment which you can use or ignore as you see fit. The only thing that I think we absolutely must do is fix the order of parameters in the assertions; the rest is a matter of taste.

One other tool in the PHPUnit test kit that I find very useful is the `@dataProvider` annotation, which lets you separate your test logic from your test data in a useful way. You _could_ use that for some of the tests here, though I think it's probably overkill. If you're interested in learning more about it, though, feel free to experiment.

Let me know if there's anything more I can do to help!

meganschanz:
I spent some time playing around with the `@dataProvider` syntax this morning, which was really neat! I did think it probably would be overkill to do in this case since I couldn't use the same data provider for each of the separate test functions (due to the expected output changing based on the config setting of that particular test function). So for simplicity and readability I kept it as-is for now. But for future tests I will definitely keep that in mind!

I made the other corrections and suggestions to the tests and removed the unnecessary else statement in the `isHoldable` function. Let me know if you see anything else!

demiankatz (approved):
This looks great, @meganschanz -- thanks for taking the time to put it all together!

PR #2685, opened 2023-01-26
FOLIO: support proxy user requests

demiankatz:
TODO
- [x] Finish adding support to FOLIO driver
- [x] Add support to Demo driver
- [x] Add optional element on hold forms to select proxy
- [x] Make sure proxy information displays where appropriate
- [x] Add test coverage where appropriate
- [x] Add new method and settings to ILS driver spec in wiki when finalized


demiankatz:
@EreMaijala, I realize you're not a FOLIO user, but I'm interested in your feedback on the general proxy request mechanism, which should now be fully functional with the Demo driver.

Note that because proxy requests are placed on behalf of another user, they do not show up in your own account. Thus, I created a separate translation string for a success message after placing a proxy request, in case clarification is needed about this.

demiankatz:
The FOLIO implementation here is confirmed to work correctly. One detail that may be worth adding is a parameter in the list of existing requests to indicate when a different user created a request on your behalf. I haven't yet investigated whether that's possible in FOLIO, though I strongly suspect that it is.

EreMaijala (changes requested):
While all the code here looks good to me, I think we should prepare for a case where you actually can see the proxied requests in the holds list and modify them etc. I realize this means more configuration, but I was sort of expecting this to be the case so I think it's not too far-fetched that Folio or another system could support managing proxied requests after they've been placed.

demiankatz:
Thanks, @EreMaijala, that makes sense, and I'll experiment with it further when time permits. Do you envision all the requests being displayed together in a single list, or a separate "proxy requests" section of the account area? I suspect the separate list might be easier to implement, though I haven't tried anything yet.

EreMaijala:
@demiankatz I haven't thought about it enough to say really, but in my mind I was thinking about a single holds list that would just include an extra field "Proxied for: Lib Rarian". But I can see that it might depend on the ILS too.

EreMaijala:
@demiankatz We can of course revisit this later if an ILS with support for that comes up. If you think that's better than trying to predict the future, I'm okay with that. And please ensure that Folio doesn't support this yet. :)

demiankatz:
@EreMaijala, I think FOLIO might support it already -- it depends on how much functionality is granted based on having the "proxyUserId" attribute set in the request object. I'll do some experimentation later today if time permits! If I can get it to work there, I'll mimic the behavior in the Demo driver so you can try it out.

demiankatz:
@EreMaijala, I believe I have added the missing functionality to both FOLIO and the Demo driver. What do you think? If you're happy with this implementation, I'll put some work into testing. Note that this branch won't actually work until #2717 is merged.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1248,6 +1258,20 @@ public function getMyHolds($patron)
                 'last_pickup_date' => $lastPickup,
                 'position' => $hold->position ?? null,
             ];
+            // Set proxy user information
+            if (($hold->proxyUserId ?? null) !== $patron['id']
```
EreMaijala: I'm not sure I understand this comparison, but it's probably my lack of understanding on how Folio presents the proxy user.

EreMaijala (approved):
Now this looks good to me. The commented part could perhaps have more verbose comments if this needs explanation to anyone more familiar with FOLIO.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
demiankatz: I've added comments in an effort to make this logic more clear. Please let me know what you think!

EreMaijala: Better. :) One more question: if `proxyUserId` is not set, shouldn't it be equal to it being the current user? Now current user will be compared against null.

demiankatz: Yes, good point! I've adjusted both of the fallbacks in this method. I think these fields are mandatory and it probably won't come up (which is why I missed the logical flaw), but better to be safe! :-)

demiankatz:
Okay, I think this is about ready to go -- I've added some FOLIO-specific unit tests, plus a Mink integration test that places a proxy request using the Demo driver. I'm going to seek a little more feedback from the FOLIO community before finalizing the documentation and merging this.

demiankatz:
@bbusenius, this is hopefully my last FOLIO change for the moment: the addition of proxy user support. I'd be interested in your thoughts from the FOLIO perspective. I hope to merge this next week if nobody objects before then.

bbusenius:
@demiankatz, I've been looking at this and have a couple of questions. It seems like I can't get this working on a local copy pointed at Folio (Nolana, I believe). I get an `ILSException` with `Unexpected error code` when I add `proxiedUsers` to the `extraHoldFields` config. I'm guessing that I would need to complete setup for proxy borrowers in Folio to make that error go away. Is that correct?

When I switch to the Demo driver, I can see how this is supposed to work. When I place a hold I am able to select from the list of users in the `ProxiedUsers` section of the demo config. When using the demo driver, I am shown all of the proxy users in the config. I assume this is just for testing and that when using this against Folio, I would only see proxy users that have been linked with my account. I just wanted to confirm that was the case.

demiankatz:
@bbusenius, thanks for taking a look at this! I'd be very interested to learn what unexpected error code you're actually receiving in FOLIO, in case there's some scenario the current code isn't handling but needs to. I would guess the problem is coming from the call to getProxiedUsers(), which is leading to [this exception](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/ILS/Driver/AbstractAPI.php#L172). It looks like if you turn debug logging on, you'll be able to get more details about exactly what happened.

And yes, your assumptions about the Demo driver and the normal behavior are correct -- you should only see your own proxy users normally, but for demo/testing purposes, we just define a global list.

demiankatz:
@bbusenius, on further reflection, I think there's a very good chance that your problem is related to a missing permission. You might need to add `proxiesfor.collection.get` permission to your VuFind service account in order to retrieve proxy user data. There's a comment about this in the code, but we probably need it somewhere more prominent. (Indeed, a comprehensive list of FOLIO permissions for VuFind would be valuable!).

bbusenius (changes requested):
You were right @demiankatz. After we gave the vufind user the API permissions, everything began to work. It's pretty standard for Folio APIs to produce that kind of an error in VuFind when the permissions haven't been given (which for us is always the default). It might be nice in this case to throw an exception that's more informative in the event that someone drops `proxiedUsers` in the `extraHoldFields` and the API permissions haven't been granted. If that's too cumbersome for some reason, a comment in the `extraHoldFields` description that mentions this might be sufficient.

Another thing I thought of when I was looking at this is that it might be nice to see my proxies and/or who I'm a proxy for in the Profile section of my account. I'm not sure if that's beyond the scope of this pull request, so I won't hold you to it, but I thought it was worth mentioning.

Beyond all that, I have a couple of people looking at this to see if it might be something we could use. We're currently handling proxy borrowing differently and I imagine our way is less ideal. I'm not sure if our stakeholders will want to change how they operate since they've been doing it the same way for a long time but I thought it would be worth their time to take a look.

demiankatz:
Thanks, @bbusenius, I'll be interested to hear what your stakeholders think of this.

Because I'm currently in the home stretch of preparing VuFind release 9.0, I'd prefer to get this merged without adding a lot of additional functionality, though I think everything you suggest makes sense -- I'd just rather do those things as follow-ups rather than extending the duration of this existing PR.

I'd suggest opening two follow-up JIRA tickets (which I'm happy to do on your behalf if you agree with my proposal):

Ticket 1: add logic to FOLIO to improve the visibility of permission-related exceptions. Exception handling for ILS drivers is a bit complicated because there's a lot of logic to catch exceptions and fail over to NoILS, etc., etc. This makes it pretty easy for exceptions to get swallowed. We also try not to expose too much detail by default in order to avoid accidentally leaking credentials or other sensitive information. So I think it will be a non-trivial project to improve the visibility of these messages... but it is of course possible, and I think worth doing. I think if we can get it right, it will potentially save a lot of people a lot of time.

I also agree that a comment might be a helpful stopgap measure, though I wonder if a more comprehensive project to document FOLIO permissions would be better than documenting just one permission in this PR. There are lots of different permissions required, and none of them are currently documented in the .ini file. Might be worth brainstorming the best way to guide people to the right solution. Maybe that dovetails with the exception handling project, or maybe that's a whole separate one.

Ticket 2: add "view proxies" page to account menu -- I think it would be useful to be able to see my proxies, but I think we need to have a bit of design discussion before implementing that. Is this something that goes on the existing profile page? Does it get its own page? What do we call everything? I don't see all of those decisions getting made in time for release 9.0, so maybe this could be a 9.1 feature.

bbusenius (approved):
That all sounds good to me. I just wanted to raise the issues. I'll go ahead and approve.

demiankatz:
Thanks again, @bbusenius! I've opened [VUFIND-1598](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1598) and [VUFIND-1599](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1599) to track the two issues we discussed above. Please feel free to comment on those with any ideas/preferences!

PR #2686, opened 2023-01-27
Add data-attribute title to facet-groups

maccabeelevine:
I'm building GTM tracking for clicking facets, to record the specific facet as well as the facet group.  This way one can filter analytics for (say) all Format facets before seeing the individual use of each facet within Format.

Within sidebar facets, the individual facet elements have a data-title attribute with the title of the facet.  That makes getting the facet link text simpler than parsing the link text.

This PR adds the same data-title to the facet-group element.  Until recently there was no clean way to extract a facet group title; we'd have identify the first button element within the facet group (brittle), get its text and then strip it of the leading and trailing whitespace.  I see a [recent change](https://github.com/vufind-org/vufind/pull/1962/files) added a span.facet-title within one use of facet-group, but doesn't appear to be for analytics purposes (so I'm not clear if it will best stable for that) and wasn't implemented in other facet-group declarations.  This PR has the potential to be intentional for analytics and thus safer for that use.

### DIFF
--- a/themes/bootstrap3/templates/search/advanced/layout.phtml
+++ b/themes/bootstrap3/templates/search/advanced/layout.phtml
@@ -196,7 +196,8 @@
         <?php endforeach; ?>
       <?php endif; ?>
       <h2><?=$this->transEsc("Search Tips")?></h2>
-      <div class="facet-group" data-title="<?=$this->transEsc("Search Tips") ?>">
+      <?php // Data-title attribute is for analytics use.  Do not remove. ?>
+      <div class="facet-group" data-title="<?=$this->escapeHtmlAttr("Search Tips") ?>">
         <a class="facet help-link" data-lightbox href="<?=$this->url('help-home')?>?topic=advsearch&amp;_=<?=time() ?>"><?=$this->transEsc("Help with Advanced Search")?></a>
         <a class="facet help-link" data-lightbox href="<?=$this->url('help-home')?>?topic=search&amp;_=<?=time() ?>"><?=$this->transEsc("Help with Search Operators")?></a>
       </div>
--- a/themes/bootstrap3/templates/search/advanced/layout.phtml
+++ b/themes/bootstrap3/templates/search/advanced/layout.phtml
@@ -196,7 +196,7 @@
         <?php endforeach; ?>
       <?php endif; ?>
       <h2><?=$this->transEsc("Search Tips")?></h2>
-      <div class="facet-group">
+      <div class="facet-group" data-title="<?=$this->transEsc("Search Tips") ?>">
         <a class="facet help-link" data-lightbox href="<?=$this->url('help-home')?>?topic=advsearch&amp;_=<?=time() ?>"><?=$this->transEsc("Help with Advanced Search")?></a>
         <a class="facet help-link" data-lightbox href="<?=$this->url('help-home')?>?topic=search&amp;_=<?=time() ?>"><?=$this->transEsc("Help with Search Operators")?></a>
       </div>
--- a/themes/bootstrap3/templates/Recommend/SideFacets.phtml
+++ b/themes/bootstrap3/templates/Recommend/SideFacets.phtml
@@ -54,7 +54,8 @@
 <?php $hierarchicalFacetSortOptions = $this->recommend->getHierarchicalFacetSortOptions() ?>
 <?php if (!empty($sideFacetSet) && $results->getResultTotal() > 0): ?>
   <?php foreach ($sideFacetSet as $title => $cluster): ?>
-    <div class="facet-group" id="side-panel-<?=$this->escapeHtmlAttr($title) ?>" data-title="<?=$this->transEsc($cluster['label']) ?>">
+    <?php // Data-title attribute is for analytics use.  Do not remove. ?>
+    <div class="facet-group" id="side-panel-<?=$this->escapeHtmlAttr($title) ?>" data-title="<?=$this->escapeHtmlAttr($cluster['label']) ?>">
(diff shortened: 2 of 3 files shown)


demiankatz:
Thanks, @maccabeelevine! As usual, I'm interested in @crhallberg's thoughts on this. Also, a question: for analytics purposes, is it better to use translated or untranslated strings? That is, if people take the same actions in different user interface languages, do you want those to track as the same thing or different things? Right now, you're using translated values, but maybe it would actually make more sense to use constant values. (I realize that this is probably not a major issue for you at Lehigh, but it could make a big difference for some of our international users).

maccabeelevine:
Thanks @demiankatz .  As you say the issue of translation or not is moot for Lehigh, but my two cents is yes, it should use translated values.  In an analytics report, the data should be easily readable for the local audience.  The counter argument would be if you were comparing analytics results from region to region, but IMHO that's less likely to occur and less important anyway.

You can also make an argument for using an internal ID for the data attributes, such as the existing $title used in the [facet group ID](https://github.com/crhallberg/vufind/blob/51b92488b57c3e53598b3cbb49e163c23144d341/themes/bootstrap3/templates/Recommend/SideFacets.phtml#L56).  That keeps the tracking consistent even if the visible label changes.  I still err on the side of using a value that's more human readable.

maccabeelevine:
One nice change in GA4 is that you can have any number of textual event parameters, not just three (utm_category, utm_action, utm_label).  So in theory you could add as many of these data attributes as you wanted, pick them up in GTM as individual event parameters for GA, and report on any or all.  But I don't see a need.

demiankatz:
@maccabeelevine, the scenario I'm concerned about is a VuFind instance which has a language select drop-down and is used by a diverse audience. If we want to know how many people clicked on a particular facet, we can't find that out if the facet name is different depending on who the user is. I think this could be a very common scenario for some of our European partners, for example. Are we more interested in how many people faceted by subject, or how many people faceted by subject in Swedish? (I'm still not sure of the answer to that question, because like you, it doesn't impact my local use case... but I think it might be significant to others). I'd be interested to hear what @EreMaijala thinks, for example.

maccabeelevine:
Thanks @demiankatz for the clarification, I wasn't thinking of changing languages per-user.  I can go either way.  Or both if needed.

demiankatz:
I18n certainly adds some unique complexities to things! I'll wait to get some more input from other stakeholders before taking a deeper look at this. :-)

crhallberg (approved):
Demian already asked about my only concern: how to escape the titles. Looks good to me!

EreMaijala:
@demiankatz @maccabeelevine I think I'd prefer untranslated titles to keep the results consistent when multiple languages are enabled, but there could also be an argument for translation that would allow you to see how much facets are used in different languages.
The concern I have with adding the data attribute without using it here is that later on it will seem like a stray attribute not used anywhere in the base code. Perhaps a PHP comment above or alongside would clarify it (I'd avoid HTML comments here).

demiankatz:
Thanks, @EreMaijala!

@maccabeelevine, I think this makes me lean further toward not translating the data elements in order to ensure consistent data collection. If we wanted to measure facet use by language, a better approach would be to capture the selected user language as a separate data element, and then correlate everything as needed, rather than relying on a side effect of translation (especially since in some languages, the values may be identical, while in others they may not).

Also, regarding making sure nobody removes these data elements on the assumption that they are unused, I think comments are a good idea. You might also consider creating a Mink integration test to catch any future regressions. (I don't insist on that since it's a bit of work, but just an idea we could consider).

maccabeelevine:
Ok I'll comment on the data attributes, and I'll un-translate them, but then guessing @demiankatz they should be run through escapeHtmlAttr()?

demiankatz:
Yes, @maccabeelevine, that sounds right to me! :-)

themes/bootstrap3/templates/search/advanced/layout.phtml
```diff
@@ -196,7 +196,8 @@
         <?php endforeach; ?>
       <?php endif; ?>
       <h2><?=$this->transEsc("Search Tips")?></h2>
-      <div class="facet-group">
+      <?php // Data-title attribute is for analytics use.  Do not remove. ?>
+      <div class="facet-group" data-title="<?=$this->escapeHtmlAttr("Search Tips") ?>">
```
demiankatz, line 200: @maccabeelevine, this is not a real facet group -- it's just a static list of help links that happens to recycle the styling of a facet group. I don't think we want to treat as if it were a facet for analytics purposes. @crhallberg, should we re-style this so it's less misleading?

demiankatz (changes requested):
Thanks, @maccabeelevine, this is looking good to me, except for one thing... see below:

themes/bootstrap3/templates/search/advanced/layout.phtml
maccabeelevine: So yes dealing with the Search Tips sidebar here is a hack, all I can say is that it's a pre-existing hack (using facet-group) that I took advantage of.  I am equally interested in capturing clicks on those advanced search help links as I am the real facets. (Lehigh just customized that help text, and we are interested in what use it gets.)  I did the two together in one PR since they were already implemented the same way, and it made the analytics seamless.  If we have to separate it that's fine, but there should be some new sidebar heading / item standard with equivalent data-elements for each, and I'll track it separately in GTM.

demiankatz: Thanks for the context, @maccabeelevine. Since it's intentional, I'll let it slide and merge this, though I definitely don't like how hacky the whole thing feels. As you say, when we fix the pre-existing issue, we can also address this one. And if we don't get to that right away, at least you'll have a workable approach in the meantime.

demiankatz (approved):

PR #2689, opened 2023-01-31
Fix boolean fields represented as integers in database schema

skellamp:
Changed the data type of public column in user_list to tinyint.

TODO
- [x] Apply similar fixes to the saved column in the searches table
- [x] Run full test suite with MySQL
- [x] Test upgrade tool with MySQL existing data
- [x] Apply equivalent changes to PostgreSQL schema
- [x] Add migration script for PostgreSQL
- [x] Run full test suite with PostgreSQL
- [x] Test migration script in PostgreSQL with existing data
- [x] Incorporate changes from this PR into #2233 after merging.


module/VuFind/sql/mysql.sql
```diff
@@ -265,7 +265,7 @@ CREATE TABLE `user_list` (
   `title` varchar(200) NOT NULL,
   `description` text,
   `created` datetime NOT NULL DEFAULT '2000-01-01 00:00:00',
-  `public` int(11) NOT NULL DEFAULT '0',
+  `public` tinyint NOT NULL DEFAULT '0',
```
demiankatz: I think this should be `tinyint(1)` for consistency with other tinyint values in the schema.

demiankatz (changes requested):
Thanks, @skellamp -- I think we also need to change `int` to `boolean` for the equivalent field in pgsql.sql, and we'll also need to add a migration script for PostgreSQL users. I'll put TODO checkboxes in place about all of this -- we should probably test that the change is safe in MySQL before we spend time working on PostgreSQL as well.

demiankatz (changes requested):
@skellamp, I tried to do a test of upgrading existing data in PostgreSQL using the migration script, but it doesn't seem to be working. Here's what I did:

1.) I started up VuFind with the `-Ddbtype=pgsql` switch while on the dev branch
2.) I created an account in the instance, created public and private lists, created some search history, and saved a search. This gave me data where I could check that flags were being properly applied after the migration.
3.) I switched to your branch
4.) I ran the migration with `psql -U vufindtest -d vufind_test -a -f module/VuFind/sql/migrations/pgsql/9.0/006-modify-boolean-fields.sql`

Here's what I got (after entering the default test environment password of "vufindpass"):

```
--
-- Modifications to table `search`
--
ALTER TABLE "search"
  ALTER COLUMN saved TYPE boolean NOT NULL DEFAULT '0';
psql:module/VuFind/sql/migrations/pgsql/9.0/006-modify-boolean-fields.sql:6: ERROR:  syntax error at or near "NOT"
LINE 2:   ALTER COLUMN saved TYPE boolean NOT NULL DEFAULT '0';
                                          ^
--
-- Modifications to table `user_list`
--
ALTER TABLE "user_list"
  ALTER COLUMN public TYPE boolean NOT NULL DEFAULT '0';
psql:module/VuFind/sql/migrations/pgsql/9.0/006-modify-boolean-fields.sql:13: ERROR:  syntax error at or near "NOT"
LINE 2:   ALTER COLUMN public TYPE boolean NOT NULL DEFAULT '0';
```

I think PostgreSQL syntax requires you to have a separate clause for each thing you change in a row. See, for example, https://github.com/vufind-org/vufind/blob/dev/module/VuFind/sql/migrations/pgsql/3.0/004-modify-seach-columns.sql for something similar.

However, when I tried to adjust the syntax to this style, I still ran into errors about type-casting of data and defaults, so I think a little more work/research is needed. Can you investigate further? Let me know if you need any more help from my end!

demiankatz (approved):
Thanks, @skellamp, this looks good and everything is working as expected for me in MySQL and PostgreSQL now.

PR #2690, opened 2023-02-01
Hierarchy Controller: return JSON when getTreeJson produces an error

jpkanter:
While internally testing *something* i noticed that i cannot reach an internal service that serves hierarchy data, at first i didnt mostly because the actual response from the AJAX Controller was a simple "error" which could in turn not be parsed as Json.

As the function is called `gettreejsonAction()` i would assume it always returns some kind of json.

I therefore propose a slight change to the error response, although no further handling is included here


demiankatz:
Thanks, @jpkanter, this seems reasonable to me. Do you plan to look into the calling code and make use of this, or would you like me to spend some time on that when schedule permits? (I'm certainly willing, though I won't get to it right away).

module/VuFind/src/VuFind/Controller/HierarchyController.php
```diff
@@ -177,7 +177,8 @@ public function gettreejsonAction()
         }
 
         // If we got this far, something went wrong:
-        return $this->outputJSON('error', 503); // Service Unavailable
+        return $this->outputJSON('{"error": {"code": 503, "message": "Service Unavailable"}}', 503);
```
demiankatz: I see that this is causing a CI failure due to line length. You could either wrap the line like:

```
return $this->outputJSON(
    '{"error": {"code": 503, "message": "Service Unavailable"}}',
    503
);
```

Or you might consider structuring the response data in PHP and then encoding it:

```

$response = ['error' => ['code': 503, 'message' => 'Service Unavailable']];
return $this->outputJSON(json_encode($response), 503);
```

demiankatz (changes requested):
Looks like some minor adjustments are needed to get the build to pass.

jpkanter:
I fear my own schedule is somewhat stuffed these days, especially when i am not all that familiar with the underlying architecture. I wrote this because i stumbled upon it by chance and thought a small fix / awareness nudge might be helpful.

demiankatz:
@jpkanter, no worries, I fully understand! If you don't mind fixing the style issue using one of my suggestions, I'll take it from there when time permits. (And if anyone else has bandwidth to work on this, please feel free to speak up!).

module/VuFind/src/VuFind/Controller/HierarchyController.php
jpkanter: i did the second variant, it seems a bit more clear and saves us the hassle of manually typing json in a php document

demiankatz (approved):
Thanks again, @jpkanter, I've finally found time to finish this up. I think this is in good shape now, but I'll wait for a second opinion on my changes before merging.

EreMaijala (approved):

PR #2691, opened 2023-02-02
Add support for username with Redis sessions.

EreMaijala:
Redis 6 supports username in addition to password. Also use null instead of false for undefined password since that's what Credis_Client documents as a valid option.


demiankatz (approved):
Makes sense to me -- thanks!

PR #2692, opened 2023-02-02
SierraRest: Add support for checking canFreeze up front.

EreMaijala:
Since requesting the canFreeze field used to be slow, there was no option for requesting it up front. This slowness is supposedly fixed in Sierra 5.6, so I added an option to enable the check. This makes it possible to prevent trying to change a hold that doesn't have anything changeable. I intentionally kept the check in updateHolds in place at least for now to ensure VuFind doesn't request a change that should not be made.


demiankatz (approved):
Looks good to me -- thanks!

PR #2693, opened 2023-02-02
Add comments for analytics data attributes

maccabeelevine:
Given @EreMaijala's [comment](https://github.com/vufind-org/vufind/pull/2686#issuecomment-1413294627) on #2686, these data attributes should also have comments.


demiankatz (approved):
Makes sense, thanks!

PR #2694, opened 2023-02-02
Add Demo DoiLinker plug-in.

demiankatz:
While testing #2400 I realized there was not an easy way to test DOI linking using the standard test data set, so this PR adds some MARC records with fake DOIs and a Demo DOI linker that generates data for every record with a DOI -- now it's easy to test the UI for DOI links without relying on any external services. In theory, we could also build a Mink test for DOI linking, though I don't think I have time right now.


EreMaijala (approved):
Looks good to me!

PR #2695, opened 2023-02-03
Add storeOffsetsWithPositions to Solr text fields using termVectors.

EreMaijala:
Continued from #2666. This should eliminate the highlighting problem.

We have termVectors enabled for author fields. Highlighter in Solr 9 apparently requires offsets with term vectors (this isn't explicitly said but is implied by the [docs](https://solr.apache.org/guide/solr/latest/query-guide/highlighting.html#schema-options-and-performance-considerations) mentioning postings ... with term vectors).

TODO
- [x] Update schema changelog; include note in main changelog about importance of reindexing for author search


demiankatz:
@EreMaijala, according to the [Solr docs](https://solr.apache.org/guide/solr/latest/query-guide/highlighting.html), "This adds a moderate amount of extra data to the index." In your experience, has that been really significant? (Just wondering if we need a warning in the changelog).

Might there be a reason to add a configuration somewhere to override hl.method in case something is broken for somebody? Did this default change at some point in the recent past? (Not advocating for building something we'll never use... just brainstorming).

I'm still not sure I understand why @welblaud was having problems with this when the default code seems to work fine as-is. It seems like this will be a beneficial change, but I want to be sure I understand the costs/benefits and that we have mechanisms in place to work around any problems that might arise -- e.g. what happens if somebody can't rebuild their index right away when updating; will everything break, or might there be a workaround?

EreMaijala:
@demiankatz I can't remember anymore what the size increase was for our index, but the speed difference was significant, so it wasn't really a choice to us back then. For the CI test index the difference seems to be 2.9MB => 3.4MB.

Anyway, I'm now having this issue with default code, so something needs to be done..

demiankatz:
@EreMaijala, do you have a path to reproduce the problem in the standard test environment? Maybe we should add a test if there's a way to do it, in case of future regressions (and to help demonstrate why this is necessary).

If the growth rate is constant relative to the test environment, a 17% index size increase probably won't be a disaster. :-)

EreMaijala:
@demiankatz Yes:
1.) Check out dev
2.) composer install
3.) phing startup phpunitfaster

Result is an early failure:
```
There was 1 failure:

1) VuFindTest\Integration\Connection\SolrTest::testDeweyValues
Failed asserting that 4 matches expected 1.

/Users/eremaijala/src/vufind-test/module/VuFind/tests/integration-tests/src/VuFindTest/Connection/SolrTest.php:108

```

demiankatz:
@EreMaijala, I'm confused -- you're seeing failures on the dev branch, but I am not. Continuous integration is passing, and I run tests regularly on a daily basis. I've never seen the failure you describe. Am I misunderstanding something?

In any case, I'm running tests on this branch right now, and I'm seeing a failure in my environment. I'll have to wait for the process to complete before I see exactly what the failure is, but I'll share it here once I have all the information.

demiankatz:
@EreMaijala, here's what I'm seeing:

```
There was 1 failure:

1) VuFindTest\Integration\Connection\SolrTest::testDeweyValues
Failed asserting that 4 matches expected 1.

/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Connection/SolrTest.php:108

FAILURES!
Tests: 2235, Assertions: 11255, Failures: 1.
```

Exactly what you reported in dev! So somehow the change that fixes this for you breaks it for me! I'm very confused.

EreMaijala:
That test failure was a red herring. Real issue was any search hitting an author field.

demiankatz:
Okay, for the record: @EreMaijala and I made sense of this through further discussion on Slack: the test failure we were both seeing was unrelated to the changes here -- it was a bug introduced by new test data from a recent PR, and was corrected by https://github.com/vufind-org/vufind/commit/342fcce6dedb4490c8b90eabff4c3241711ef891.

There WAS a problem related to the Solr 9, but it was not detectable by any of the existing tests. The problem was only triggered by searching text found in author fields, because the root issue here seems to be that Solr 9 highlighting fails on fields that have termVectors set to true without storeOffsetsWithPositions also set to true. We've scaled back this PR to address only that problem, and I've added a simple Mink test to do a couple of author searches to catch regressions in future. This new test is currently failing in dev, but passing in this branch.

demiankatz (approved):
@EreMaijala, I think this is ready to merge now. I'd just like to get your feedback on my test before merging. Once I do merge, I'll take care of the changelog updates.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AuthorSearchTest.php
```diff
@@ -0,0 +1,69 @@
...
+     *
+     * @return void
+     */
+    public function testCorporateAuthorSearch(): void
+    {
+        $page = $this->performSearch('corporate', 'Author');
+        $facets = $this->findCss($page, '#side-collapse-building a')->getText();
+        // We'll check for a known count from a known MARC file to confirm that
+        // results came back.
+        $this->assertTrue(strpos($facets, "author_relators.mrc 10") !== false);
```
EreMaijala: Use [assertStringContainsString](https://phpunit.readthedocs.io/en/9.6/assertions.html#assertstringcontainsstring) in both tests.

EreMaijala:
@demiankatz Thanks! With the forementioned change I think this is ready to be committed.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AuthorSearchTest.php
demiankatz: Good idea -- fix pushed up. I'm going to run the whole test suite one last time, then merge if nothing fails.

demiankatz:
Okay, tests all passing and wiki updated -- merging now. Thanks again for helping figure out the problem here, @EreMaijala!

PR #2696, opened 2023-02-03
Fix validate ids logic

xmorave2:
The method documentation says:

`Validate supplied IDs against remembered IDs. Returns true if all supplied IDs are remembered, otherwise returns false.`

Which is in not true - see the test added

### DIFF
--- a/module/VuFind/src/VuFind/Controller/HoldsController.php
+++ b/module/VuFind/src/VuFind/Controller/HoldsController.php
@@ -209,7 +209,7 @@ class HoldsController extends AbstractBase
         }
         // If the user input contains a value not found in the session
         // legal list, something has been tampered with -- abort the process.
-        if ($this->holds()->validateIds($selectedIds)) {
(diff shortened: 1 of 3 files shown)


EreMaijala (approved):
Thanks, looks good to me. I'd get this fixed but let's see what @demiankatz says about any compatibility issues.

demiankatz (approved):
This bug was introduced in #2206, which has never been released (it was an early change to the dev-9.0 branch). Thus, I think we can safely fix this without introducing backward compatibility problems.

This also reminds me that the last time we talked about this, @EreMaijala opened https://openlibraryfoundation.atlassian.net/browse/VUFIND-1521. Should we try to schedule that for 10.0 so we don't completely forget about it?

EreMaijala:
@demiankatz Of course, no problem then! Indeed, it would make sense to schedule VUFIND-1521 for 10.0.

demiankatz:
Thanks, @EreMaijala, I've scheduled the ticket.

PR #2697, opened 2023-02-06
Semantics for saved search table

stephanieleary:
Adds scoped headers to saved search table for better screen reader support, as well as classes for CSS styling.


themes/bootstrap3/templates/search/history-table.phtml
```diff
@@ -3,25 +3,34 @@
   $saveSupported = $this->accountCapabilities()->getSavedSearchSetting() === 'enabled';
 ?>
 <table class="search-history-table" id="<?=$this->showSaved ? 'saved-searches' : 'recent-searches'?>">
+  <thead>
```
demiankatz: @stephanieleary, I see this is a draft, so maybe I'm reviewing prematurely... but there seem to be some doubled and oddly nested tags here -- e.g. double open theads, double close tbodys, etc. Can you take another look at that?

demiankatz (changes requested):

themes/bootstrap3/templates/search/history-table.phtml
stephanieleary: What an utterly bizarre glitch! I checked my local file and I don't see the duplicated lines, so I'm not sure what happened. I've corrected the file.

demiankatz: I can't explain it, but thanks for fixing it! :-)

demiankatz (approved):
Thanks, @stephanieleary. I've just pushed up a couple of small changes here:

1.) I indented the HTML inside the new thead/tbody tags for consistency with other HTML styling.

2.) I added a 20% width to the time column in the LESS (moving it from being hard-coded in the template). Your previous changes removed this width setting, but I found that this caused the time column to be too wide when search terms were short, and too narrow when search terms were long; I believe that having an explicit width is still helpful for the visual appearance of the table.

Please let me know if you want to make changes or have any objections to these edits. If you're happy with the result, please take the PR out of draft mode, and then it can be merged.

I'm also going to ask @crhallberg for a review in case he has additional comments.

crhallberg (approved):
No further comments! Thank you very much!

demiankatz:
@stephanieleary, I just merged the latest dev branch into this PR, as some conflicts were introduced by recent upstream changes. I believe this is ready to merge now, but just waiting for your thumbs-up.

stephanieleary:
Thanks for following up!

demiankatz:
@stephanieleary, I think this may have gotten a bit lost in the shuffle. I've just merged the latest dev branch in and resolved conflicts. I think it's ready for a final round of testing before being merged.

Does @pasitiis have any suggestions or concerns?

Could @sturkel89 do a quick round of testing to compare this against the dev branch and make sure there are no significant visual anomalies? All changes here impact the search history screen, so it should be a matter of performing a few searches and saving some of them to fully populate the data displayed by the modified template. Turning on the `schedule_searches` setting in config.ini will also reveal some additional controls that are normally hidden.

sturkel89:
I've run some checks on the Search History screen in all three themes, using both the test branch and the dev branch. 

I can't find any visible differences between test and dev in any theme, except for one very minor thing that I would say does not need to be fixed. The one difference is where the alternating row shading starts: in the test branch, it starts with the first item in the list of saved or recent searches. In the dev branch, the shading starts with the header row (Time, Search, Limits, Results, Alert Schedule, Delete). 

**Test branch** - Sandal example:
image

**Dev branch** - Sandal example:
image

The same is seen in every theme. Looks good!

themes/bootstrap3/less/components/search.less
```diff
@@ -547,6 +547,9 @@ table.search-history-table {
     width: 20%;
   }
 }
+.search-history-table .history_time {
```
demiankatz: @EreMaijala, it looks like subsequent to this PR being opened, you eliminated the hard-coded 20% in a different way (see the lines of code directly above this). I don't think we need both solutions to be implemented. Do you have a preference for either keeping your existing solution or adopting the one proposed here?

EreMaijala: @demiankatz I'm fine with either solution, but the other needs to be removed. :) The one proposed here is maybe a bit cleaner. Regarding the even/odd shading, I actually preferred the old one, but I suppose it's more down to a personal preference.

EreMaijala (changes requested):
The 20% column width rule needs to be cleaned up. The rest is ok for me.

demiankatz:
Thanks, @EreMaijala -- I believe I've found a "best-of-both-worlds" approach to the LESS, so I've pushed up that improvement.

Regarding the color alternation, it's simply related to the fact that the alternation applies to `td` and not `th`, so converting the table header changes where the alternation begins. I imagine there's probably a way to make the alternation work across `th` and `td`, but that's beyond my area of expertise. I think we can live with it for now at least. :-)

EreMaijala (approved):

PR #2698, opened 2023-02-06
[VUFIND-1535] Make search history behavior/labeling more consistent.

demiankatz:
This PR makes the breadcrumb and menu labels more consistent for search history. Additionally, it adds a message prompting users to log in for saved searches, rather than forcing the login through less intuitive side menu linking behavior.

TODO
- [x] Run full test suite


demiankatz:
Full integration suite is passing. I think this is ready to merge if it passes review.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SavedSearchesTest.php
```diff
@@ -173,14 +173,14 @@ public function testSearchHistory(): void
         // different problems in a single test.
         $page = $this->performSearch('foo \ bar');
         $this->findAndAssertLink($page, 'Search History')->click();
-        // We should see our "foo \ bar" search in the history, but no saved
-        // searches because we are logged out:
+        // We should see our "foo \ bar" search in the history, but a login link
```
EreMaijala: I can't parse this sentence. Should `but` be `and`?

EreMaijala (changes requested):
This is better than before, but I find it a bit odd that after logging in you don't actually see the "Saved Searches" heading unless you have saved searches. I think it would be better to display it and "You don't have any saved searches" instead of hiding it.

Also a very minor issue: if you log out from the history page after login, you'll be prompted to log in again as the `?require_login` parameter stays in the URL after login. It would be better to set the history page without the query param as followup after login.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SavedSearchesTest.php
demiankatz: Fixed.

demiankatz:
Thanks, @EreMaijala, all good suggestions. I think I've straightened everything out!

demiankatz:
Also, for the record, I re-ran the full integration test suite to be sure my latest changes didn't break anything, and it's still passing.

EreMaijala (approved):
Looks good to me now!

demiankatz:
Thanks, @EreMaijala, merging now!

PR #2699, opened 2023-02-08
Remove unused folder_id field from search table

demiankatz:
The search table in the database has contained a folder_id field since the early days of VuFind 1.x; however, this field has never been used for anything. This PR removes the unused field and associated code/indexes.

TODO
- [x] Add upgrade script logic to optionally remove field/index from MySQL (allow it to be kept in case somebody has found a local use)
- [x] Add migration to remove field/index from PostgreSQL
- [x] Run full test suite
- [x] Update changelog
- [x] Make appropriate updates to #2233 after merging.


demiankatz:
@EreMaijala, I think this is more or less ready to go. The screen for selecting whether or not to remove the deprecated column in the upgrade process could probably be beautified, but at least it's functional. Maybe @crhallberg has some ideas there.

EreMaijala (approved):
Looks good to me and works as well!

demiankatz:
Thanks, @EreMaijala! Here's a screen shot of the new upgrade screen in case @crhallberg wants to suggest any changes to confirmdeprecatedcolumns.phtml before I merge this PR:

image

themes/bootstrap3/templates/upgrade/confirmdeprecatedcolumns.phtml
```diff
@@ -0,0 +1,19 @@
...
+<table class="table table-striped">
+    <tr><th scope="col">Table</th><th scope="col">Column(s)</th></tr>
+    <?php foreach ($deprecated as $table => $columns): ?>
+        <tr><td><?=$this->escapeHtml($table)?></td><td><?=$this->escapeHtml(implode(', ', $columns))?></td></tr>
+    <?php endforeach; ?>
+</table>
+
+<p><b>RECOMMENDED:</b> You can <a href="<?=$this->url('upgrade-confirmdeprecatedcolumns')?>?action=delete">remove the columns</a>.</p>
+
+<p>Alternatively, if you are using these columns for some reason (e.g. in local custom code), you can choose to <a href="<?=$this->url('upgrade-confirmdeprecatedcolumns')?>?action=keep">keep the columns</a>.</p>
```
crhallberg: As regular link, it looks like it will take you to documentation on how to do this. I would instead style these links as buttons by adding the classes `btn btn-default`.

crhallberg (changes requested):

themes/bootstrap3/templates/upgrade/confirmdeprecatedcolumns.phtml
demiankatz: Thanks, @crhallberg. I had to use `btn btn-default btn-link` to avoid having underlines inside the buttons. Is that appropriate, or should I do things differently? (If I leave off `btn-default` they don't look like buttons).

crhallberg: That makes sense, I'm surprised that `.btn-default` doesn't remove the underline. That might be a more consistent option.

crhallberg: Same applies to #2741.

demiankatz: Please take a look at my latest changes to #2741. If you approve of that approach, I can update this to match once it is merged.

demiankatz: Okay, @crhallberg, #2741 has been merged into this, and I've adjusted the template accordingly. I think this is ready to merge now; just needs your approval.

crhallberg (approved):

PR #2700, opened 2023-02-09
Always return true from Redis session handler's destroy method.

EreMaijala:
Since Redis could have already unlinked an expired session, it's not an error if it's not found. But if we return false, PHP will log an error.


demiankatz (commented):
This makes sense to me, @EreMaijala, but one question before I merge: does the return code tell us anything about the nature of the failure? I understand that "does not exist" may not be a problem, but what about "cannot connect to Redis" or something like that? Could that happen here, or would such a failure be captured elsewhere in the code?

EreMaijala:
@demiankatz That should throw an exception.

demiankatz (approved):
Thanks, @EreMaijala!

PR #2701, opened 2023-02-10
Keep current search id in the url

EreMaijala:
The idea is to keep current search id in a query parameter when moving from results to the record page so that we don't need to rely on last search remembered by the search memory. 

Advantages:
- Allows paging through multiple searches and their results without them mixing up each other.
- Simplified handling of hidden filters ($ignoreHiddenFilterMemory, $ignoreHiddenFiltersInRequest)

This is nowhere near complete, but I'm still happy to collect feedback.

TODO:
- [x] Document backward incompatible changes (VuFind\RecordTab\CollectionList constructor, VuFind\Controller\Plugin\ResultScroller constructor, VuFind\Search\Memory constructor and rememberSearch signature, recordLinker view helper now accepts results object).
- [x] Since we inevitably have some backward incompatibilities, remove back-compatibility attemps from constructors etc.?
- [x] Create breadcrumb from sid so that it doesn't get mixed. Done with params not stored with the search retrieved from memory.
- [x] Eliminate duplicate "memories" if possible. Not possible; still needed for breadcrumbs and editing of memory when `retain_filters_by_default = true` 
- [x] Consider turning ignoreHiddenFiltersInRequest off by default and only enabling it in search home pages. Not really in the scope here, so can be considered separately.


module/VuFind/src/VuFind/Controller/AbstractSearch.php
```diff
@@ -389,6 +385,11 @@ protected function getSearchResultsView($setupCallback = null)
                 $this->saveSearchToHistory($results);
             }
 
+            // If a "jumpto" parameter is set, deal with that now:
+            if ($jump = $this->processJumpTo($results)) {
```
demiankatz: I thought if we were jumping to a specific record, we wouldn't want to remember a search in the search history. I imagine you had to refactor this to ensure that the sid was available to the processJumpTo() method, but does this have undesirable side effects? Might there be a better way to do this?

module/VuFind/src/VuFind/Controller/Plugin/ResultScroller.php
```diff
@@ -29,6 +29,9 @@
 
 use Laminas\Mvc\Controller\Plugin\AbstractPlugin;
 use Laminas\Session\Container as SessionContainer;
+use VuFind\RecordDriver\AbstractBase as BaseRecord;
+use VuFind\Search\Base\Results;
+use Vufind\Search\memory as SearchMemory;
```
demiankatz: Lowercase "memory" here.

module/VuFind/src/VuFind/Db/Table/Search.php
```diff
@@ -290,6 +290,7 @@ public function saveSearch(
                 $minified = $normalized->getMinified();
                 $minified->id = $existingRow->getSearchObject()->id;
                 $existingRow->search_object = serialize($minified);
+                $existingRow->session_id = $sessionId;
```
demiankatz, line 293: Is this a bug fix that should be committed separately? I'm not sure I understand the need for it, though -- if it's a saved search, it should have been retrieved by user ID, and if it's part of the current session, the session ID should already be set. Am I missing some other factor?

module/VuFind/src/VuFind/Search/Memory.php
```diff
@@ -179,4 +228,76 @@ public function retrieveSearch()
...
+     *
+     * @param int $id Search ID
+     *
+     * @return ?\VuFind\Search\Base\Results
+     */
+    protected function getSearchById(int $id): ?\VuFind\Search\Base\Results
+    {
+        if (!array_key_exists($id, $this->searchCache)) {
+            $search
+                = $this->searchTable->getOwnedRowById($id, $this->sessionId, null);
```
demiankatz, line 298: Don't we also need to pass the user ID here? Or is this the reason you're rewriting the session ID when an existing search is loaded?

module/VuFind/src/VuFind/View/Helper/Root/SearchMemory.php
```diff
@@ -48,14 +50,25 @@ class SearchMemory extends AbstractHelper
...
      *
      * @param Memory $memory Search memory
      */
-    public function __construct(Memory $memory)
+    public function __construct(Memory $memory, Search $search, string $sessionId,
+        \VuFind\Search\Results\PluginManager $resultsManager)
     {
         $this->memory = $memory;
+        $this->searchTable = $search;
+        $this->sessionId = $sessionId;
```
demiankatz: Is this work in progress? You seem to have added dependencies that aren't used yet.

module/VuFind/tests/unit-tests/src/VuFindTest/Controller/Plugin/ResultScrollerTest.php
```diff
@@ -423,7 +423,7 @@ public function __construct($mockManager, $testResults)
      *
      * @return \VuFind\Search\Base\Results
      */
-    protected function restoreLastSearch()
+    protected function restoreLastSearch(): ?\VuFind\Search\Base\Results
```
demiankatz: Should `@return` above be updated to match?

demiankatz (changes requested):
@EreMaijala, I haven't had time to look at this whole thing in detail yet, but here are a few questions and observations based on a very brief initial skim.

module/VuFind/src/VuFind/Controller/AbstractSearch.php
EreMaijala: I find it odd that a search is not saved if it finds only a single record and jump_to_single_search_result is enabled, but it is if jump_to_single_search_result is disabled. Currently the correct search tab is lost in the former case, and that's why I changed it, but it could be changed back if that's more desirable.
 
In other cases I actually don't quite understand how jumpto works and why it's needed. :)

module/VuFind/src/VuFind/Db/Table/Search.php
EreMaijala: It's not a bug, or that depends on how you take it. I think it should reflect the last session it was executed in, but it not being updated doesn't cause any errors in current code base. 

But if we keep it up to date, we can always retrieve last searches by session id + search id, and won't need to carry user id around.

module/VuFind/src/VuFind/Search/Memory.php
EreMaijala: See above. :)

module/VuFind/src/VuFind/View/Helper/Root/SearchMemory.php
EreMaijala: That was left over after moving the actual functionality to \VuFind\Search\Memory. Now cleaned up!

module/VuFind/tests/unit-tests/src/VuFindTest/Controller/Plugin/ResultScrollerTest.php
EreMaijala: Done.

module/VuFind/src/VuFind/Controller/Plugin/ResultScroller.php
EreMaijala: Fixed

module/VuFind/src/VuFind/Controller/AbstractSearch.php
demiankatz: I agree, I would expect the search to be saved whether or not jump_to_single_search_result is enabled, since a search was performed. I was thinking of the situation where you take an existing search result and manually add jumpto=1 to it to force it to jump to the first record matching the search -- I wouldn't expect that to add a line to search history, since users accessing the link aren't actually performing a search. That use case is the reason for the jumpto parameter, though honestly I can't remember the origins of that use case in the first place! I guess we could add logic to skip the save if there's a jumpto query parameter, or we could just not worry about it because probably nobody uses this or cares. ;-)

demiankatz: Ahh, I did some searching and found the reason for it -- it's for the MARC record linker, which uses jumpto links to link to records that it searches via identifiers. See https://github.com/vufind-org/vufind/blob/61f273c0a2db5f13b840821023ea34740b21522a/module/VuFind/src/VuFind/View/Helper/Root/RecordLinker.php#L75 for an example. I definitely think we do NOT want these links to add entries to search history.

module/VuFind/src/VuFind/View/Helper/Root/RecordLinker.php
```diff
@@ -45,14 +48,60 @@ class RecordLinker extends \Laminas\View\Helper\AbstractHelper
...
+     * @param \Laminas\Http\Request $request Request
      */
-    public function __construct(\VuFind\Record\Router $router)
-    {
+    public function __construct(
+        \VuFind\Record\Router $router,
+        \Laminas\Http\Request $request
+    ) {
         $this->router = $router;
+        $this->request = $request;
```
demiankatz: This also appears to be unused.

module/VuFind/src/VuFind/View/Helper/Root/SearchMemory.php
```diff
@@ -48,14 +50,22 @@ class SearchMemory extends AbstractHelper
...
     /**
      * Constructor
      *
      * @param Memory $memory Search memory
      */
-    public function __construct(Memory $memory)
+    public function __construct(Memory $memory, Search $search)
     {
         $this->memory = $memory;
+        $this->searchTable = $search;
```
demiankatz: I see that you removed some of the unused parameters, but searchTable also still seems to be unused.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/result-list.phtml
```diff
@@ -1,5 +1,5 @@
 <?php
-  $coverDetails = $this->record($this->driver)->getCoverDetails('result-list', 'medium', $this->recordLinker()->getUrl($this->driver));
+  $coverDetails = $this->record($this->driver)->getCoverDetails('result-list', 'medium', $this->recordLinker($this->results)->getUrl($this->driver));
```
demiankatz: Not necessary, but just a thought: would this template be cleaner if we did a `$recordLinker = $this->recordLinker($this->results)` at the top and then abbreviated all the other calls to use the variable?

demiankatz (changes requested):
Thanks for the progress/explanations, @EreMaijala. I've looked through a few more files and found a couple more spots where I think there may be orphaned parameters.

module/VuFind/src/VuFind/View/Helper/Root/RecordLinker.php
EreMaijala: Cleaned up!

module/VuFind/src/VuFind/View/Helper/Root/SearchMemory.php
EreMaijala: Fixed.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/result-list.phtml
EreMaijala: Done while I added the parameter to other templates as well.

module/VuFind/src/VuFind/Controller/AbstractSearch.php
EreMaijala: I revised it so that jumpto doesn't get added to history, and single result jump is handled separately.

module/VuFind/src/VuFind/Controller/AbstractRecord.php
```diff
@@ -342,8 +342,14 @@ public function homeAction()
                 if (true === $driver->tryMethod('isCollection')) {
                     $params = $this->params()->fromQuery()
                         + $this->params()->fromRoute();
+                    $options = [];
+                    if ($sid = $this->getSearchMemory()->getCurrentSearchId()) {
+                        $options = [
```
demiankatz: Since you've already initialized $options above, you could make this a one-liner `$options['query'] = compact($sid);`

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
```diff
@@ -294,17 +294,39 @@ protected function getVuFindUrl($path = '')
...
+    /**
+     * Get current URL without any sid parameter in the query string
+     *
+     * @return string
+     */
+    protected function getCurrentUrlWithoutSid(): string
+    {
+        $this->getMinkSession();
+        $url = $this->getMinkSession()->getCurrentUrl();
+        $url = preg_replace('/\?sid=\d+/', '?', $url);
```
demiankatz: Could these three lines be optimized down to:

`$url = preg_replace('/([&?])sid=\d+/', '$1', $url);`

...or am I missing something?

module/VuFind/src/VuFind/View/Helper/Root/SearchMemory.php
```diff
@@ -28,6 +28,7 @@
 namespace VuFind\View\Helper\Root;
 
 use Laminas\View\Helper\AbstractHelper;
+use VuFind\Db\Table\Search;
```
demiankatz: Is this actually used for anything?

module/VuFind/src/VuFind/View/Helper/Root/RecordLinker.php
```diff
@@ -150,38 +185,56 @@ public function getRequestUrl($url, $includeAnchor = true)
     /**
      * Given a record driver, get a URL for that record.
      *
-     * @param \VuFind\RecordDriver\AbstractBase|string $driver Record driver
+     * @param \VuFind\RecordDriver\AbstractBase|string $driver  Record driver
```
demiankatz: Should we add a use statement for the record driver to make these comments a little less crazy-looking?

module/VuFind/src/VuFind/View/Helper/Root/RecordLinkerFactory.php
```diff
@@ -66,6 +66,9 @@ public function __invoke(
         if (!empty($options)) {
             throw new \Exception('Unexpected options sent to factory.');
         }
-        return new $requestedName($container->get(\VuFind\Record\Router::class));
+        return new $requestedName(
+            $container->get(\VuFind\Record\Router::class),
+            $container->get('Request')
```
demiankatz: This isn't being used in the class any more, so I think we can revert this factory.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/link-collection.phtml
```diff
@@ -1 +1,5 @@
-<?=$this->url('collection', ['id' => $this->lookfor])?>?recordID=<?=urlencode($this->driver->getUniqueID())?>
\ No newline at end of file
+<?=$this->url(
+    'collection',
+    ['id' => $this->lookfor],
+    ['query' => ['recordID' => $this->driver->getUniqueID(), 'sid' => ($this->results ? $this->results->getSearchId() : null) ?? $this->searchMemory()->getCurrentSearchId()]]
```
demiankatz, line 4: I've seen this pattern of `($this->results ? $this->results->getSearchId() : null) ?? $this->searchMemory()->getCurrentSearchId()` in more than once place, if memory serves... It's not the most readable thing. Would it make any sense to either add an optional $results parameter to getCurrentSearchId(), or add a separate getCurrentSearchIdFromResultsOrMemory method to make this a little cleaner? (Feel free to ignore if you don't think it's a worthwhile change).

module/VuFind/tests/unit-tests/src/VuFindTest/Controller/Plugin/ResultScrollerTest.php
```diff
@@ -412,18 +418,20 @@ class ResultScrollerMock extends \VuFind\Controller\Plugin\ResultScroller
      */
     protected $testResults;
 
-    public function __construct($mockManager, $testResults)
+    public function __construct($mockManager, $testResults, $memory)
```
demiankatz: Is there a compelling reason to give the mock class a different constructor signature than its parent? Would it be better to match the signature and hard-code the true up above where we construct this instead?

module/VuFind/tests/unit-tests/src/VuFindTest/Controller/Plugin/ResultScrollerTest.php
```diff
@@ -412,18 +418,20 @@ class ResultScrollerMock extends \VuFind\Controller\Plugin\ResultScroller
...
     /**
      * Stubbed
      *
-     * @return \VuFind\Search\Base\Results
+     * @param int $searchId Search ID
+     *
+     * @return ?\VuFind\Search\Base\Results
      */
-    protected function restoreLastSearch()
+    protected function restoreSearch(int $searchId): ?\VuFind\Search\Base\Results
```
demiankatz, line 495: There's a comment elsewhere in this file that still refers to restoreLastSearch and should be updated.

module/VuFind/src/VuFind/Controller/AbstractSearch.php
```diff
@@ -449,9 +447,49 @@ protected function processJumpTo($results)
 
         // If we got this far, we have a valid parameter so we should redirect
         // and report success:
-        $details = $this->getRecordRouter()
-            ->getTabRouteDetails($recordList[$jumpto - 1]);
-        return $this->redirect()->toRoute($details['route'], $details['params']);
+        return $this->getRedirectForRecord($recordList[$jumpto - 1]);
```
demiankatz: I guess we could now further collapse the end of this function down to:

```
return isset($recordList[$jumpto - 1])
    ? $this->getRedirectForRecord($recordList[$jumpto - 1]) : false;
```

...if we wanted to be a little more concise.

themes/bootstrap3/templates/search/searchbox.phtml
```diff
@@ -1,12 +1,27 @@
 <?php
     // Set default value if necessary:
-    if (!isset($this->searchClassId)) {
-        $config = $this->config()->get('config');
-        $this->searchClassId = $config->Site->defaultSearchBackend ?? 'Solr';
+    $options = null;
```
demiankatz: Since we have an if..else, I don't think it makes sense to initialize these variables before the if -- some are already redefined by both branches, rendering the initialization redundant, and the remainder could just be moved into the else clause.

demiankatz (changes requested):
@EreMaijala, I ran out of time before I could fully review everything here, but I got through most of the files in some detail. I still need to more carefully examine two big ones: ResultScroller.php and searchbox.phtml (and, of course, do some hands-on testing). In any case, in the interest of forward momentum, here are my comments so far, many of them quite minor.

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
EreMaijala: If there's a question mark, we need to keep it, but if we also keep the ampersand, it can lead to a dangling one:
`path?sid=1&foo=2` => `path?&foo=2`
or
`path?foo=2&sid=1` => `path?foo=2&`

While these technically work fine, they don't match the expected results. I trimmed the other preg_replace from the method, but had to keep the rest.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/link-collection.phtml
EreMaijala: I think it's only in one other place and there the alternative is getLastSearchId, so the cases don't quite match. I'd revisit this if we end up needing it in more places.

module/VuFind/tests/unit-tests/src/VuFindTest/Controller/Plugin/ResultScrollerTest.php
EreMaijala: It was already different; the real ResultScroller doesn't have the $testResults parameter. But I dropped the constructor and added a new method for setting `$testResults`.

module/VuFind/src/VuFind/View/Helper/Root/SearchMemory.php
EreMaijala: No, removed.

module/VuFind/src/VuFind/Controller/AbstractRecord.php
EreMaijala: Done.

module/VuFind/src/VuFind/View/Helper/Root/RecordLinkerFactory.php
EreMaijala: Done.

module/VuFind/tests/unit-tests/src/VuFindTest/Controller/Plugin/ResultScrollerTest.php
EreMaijala: Updated.

module/VuFind/src/VuFind/Controller/AbstractSearch.php
EreMaijala: Done.

themes/bootstrap3/templates/search/searchbox.phtml
EreMaijala: Agreed. Now cleaned up.

EreMaijala:
@demiankatz Thanks for the comments. I believe I've addressed them. Mink tests are also passing for me.

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
demiankatz: Good point. How about this instead?

```
$url = preg_replace('/([&?])sid=[^&]*&?/', '$1', $url);
```

That should take care of ampersand trimming in a one-liner, plus it will also remove empty parameters or parameters with unexpected values in them, so I think it's a little more universal than assuming that the value will always be numeric. We could even generalize this into a "remove query parameter" method that takes a parameter name, if we eventually need this for other purposes.

EreMaijala: Yes, that'd work. Now applied!

module/VuFind/src/VuFind/Controller/Plugin/ResultScroller.php
```diff
@@ -63,32 +73,49 @@ class ResultScroller extends AbstractPlugin
...
     public function __construct(
         SessionContainer $session,
         ResultsManager $rm,
-        $enabled = true
+        $enabled = true,
+        SearchMemory $sm = null
     ) {
         $this->enabled = $enabled;
 
         // Set up session namespace for the class.
```
demiankatz: I'm not sure I understand what this comment means. I wonder if this is a relic from some past design.

module/VuFind/src/VuFind/Controller/Plugin/ResultScroller.php
```diff
@@ -63,32 +73,49 @@ class ResultScroller extends AbstractPlugin
...
     /**
      * Initialize this result set scroller. This should only be called
      * prior to displaying the results of a new search.
      *
-     * @param \VuFind\Search\Base\Results $searchObject The search object that was
-     * used to execute the last search.
+     * @param Results $searchObject The search object that was used to execute the
+     * last search.
      *
      * @return bool
```
demiankatz: What does this return value even mean? I wonder if we should explain something in the comments here. (I gather it's true when there are scroll results available, but I wouldn't typically expect an init() method to return a boolean).

module/VuFind/src/VuFind/Controller/Plugin/ResultScroller.php
```diff
@@ -99,27 +126,55 @@ public function init($searchObject)
...
+    /**
+     * Add data to session for a search
+     *
+     * @param Results $searchObject Search object
+     *
+     * @return void
+     */
+    protected function addData(Results $searchObject): void
+    {
+        $data = new \StdClass();
```
demiankatz: Isn't this usually capitalized as `\stdClass`?

module/VuFind/src/VuFind/Controller/Plugin/ResultScroller.php
```diff
@@ -99,27 +126,55 @@ public function init($searchObject)
...
+        // Store last access time for eviction
+        $data->lastAccessTime = time();
 
-        return (bool)$this->data->currIds;
+        if (!isset($this->session->s)) {
+            $this->session->s = [];
+        }
+
+        // Evict oldest entry if storage is full:
+        while (count($this->session->s) > static::LAST_SEARCH_LIMIT) {
```
demiankatz: Maybe this should be factored to a separate method since it's a stand-alone action.

themes/bootstrap3/templates/search/searchbox.phtml
```diff
@@ -1,12 +1,25 @@
...
+      $lastLimit = $params->getLimit();
+      $options = $results->getOptions();
+    } else {
+      if (!isset($this->searchClassId)) {
+          $config = $this->config()->get('config');
+          $this->searchClassId = $config->Site->defaultSearchBackend ?? 'Solr';
+      }
+      $options = $this->searchOptions($this->searchClassId);
+      $hiddenFilters = $this->searchTabs()->getHiddenFilters($this->searchClassId, true, $this->ignoreHiddenFiltersInRequest ?? false);
+      $lastSort = null;
```
demiankatz: Should these be defaulting to the last remembered sort/limit from the searchMemory instead of null? It looks to me like my last sort is no longer being remembered if I go back to the search home page, which is a change in behavior from the dev branch which I assume is not intentional.

themes/bootstrap3/templates/search/searchbox.phtml
```diff
@@ -99,7 +100,14 @@
     <?php endif; ?>
     <button type="submit" class="btn btn-primary"><?=$this->icon('search') ?> <?=$this->transEsc("Find")?></button>
     <?php if ($advSearch): ?>
-      <a href="<?=$this->url($advSearch) . (($this->searchId ?? false) ? '?edit=' . $this->escapeHtmlAttr($this->searchId) : $hiddenFilterParams) ?>" class="btn btn-link" rel="nofollow"><?=$this->transEsc("Advanced")?></a>
+      <?php
+        if ($results) {
```
demiankatz: This could be condensed significantly if you just use a ternary to set the 'query' rather than duplicating the whole line of code...

```
$advSearchQuery = $results ? ['edit' => $results->getSearchId()] : $hiddenFilters;
$advSearchLink = $this->url($advSearch, [], ['query' => $advSearchQuery]);
```

demiankatz (changes requested):
Thanks for all the work here, @EreMaijala, this is coming together very nicely! I did find one issue in hands-on testing, along with a few smaller suggestions.

I'm done reviewing all the code, but I think some more interactive testing is probably a good idea given the scale/scope of the changes here. Are there any areas you'd particularly like me to experiment with? Do you think any new Mink tests might be justified? I'm happy to help with test development if you have anything in mind, since I'll be investing time in testing one way or another.

module/VuFind/src/VuFind/Controller/Plugin/ResultScroller.php
EreMaijala: Likely a relic, yes. I removed it.

module/VuFind/src/VuFind/Controller/Plugin/ResultScroller.php
EreMaijala: I can't see the return value being useful either, but the tests check it so I left it in for now. I added an explanation for it.

module/VuFind/src/VuFind/Controller/Plugin/ResultScroller.php
EreMaijala: Oops, fixed.

module/VuFind/src/VuFind/Controller/Plugin/ResultScroller.php
EreMaijala: Done.

themes/bootstrap3/templates/search/searchbox.phtml
EreMaijala: Yes, fixed.

themes/bootstrap3/templates/search/searchbox.phtml
EreMaijala: Thanks, done!

EreMaijala:
@demiankatz Some more Mink tests might be useful, and if you have a chance to do something about them, please feel free. I haven't yet checked how well the search tabs are tested, but I'd like to ensure that the tabs work properly. Blender tests should check that the correct tab stays active. Also `retain_filters_by_default = true` should be included in tests. Other than that there could still be places where the sid parameter is not added when it should be. I've tried to look for such cases, but I wouldn't be surprised if something's still missing.

EreMaijala:
@demiankatz I'm going to check and/or improve the Blender Mink test now.

EreMaijala:
@demiankatz I've now extended the Mink tests to cover the forementioned cases. All done unless issues are still found!

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/result-list.phtml
```diff
@@ -161,10 +162,15 @@
       <?php if ($this->userlist()->getMode() !== 'disabled'): ?>
         <?php if ($this->permission()->allowDisplay('feature.Favorites')): ?>
           <?php /* Add to favorites */ ?>
+<<<<<<< HEAD
```
demiankatz: There's an unresolved merge conflict here.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/result-list.phtml
```diff
@@ -179,8 +185,13 @@
         <?php foreach ($trees as $hierarchyID => $hierarchyTitle): ?>
           <div class="hierarchyTreeLink">
             <input type="hidden" value="<?=$this->escapeHtmlAttr($hierarchyID)?>" class="hiddenHierarchyId" />
+<<<<<<< HEAD
```
demiankatz: Also here.

demiankatz (changes requested):
@EreMaijala, it looks like some merge conflicts were introduced recently -- see below. I've added some notes to the TODO checkbox about documenting BC breaks -- please review and add anything I may have missed. Once the merge issue is sorted out, I'll do some more testing next week, and hopefully we can wrap this up! (Also, is any further discussion/thought required regarding the last TODO checkbox?)

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/result-list.phtml
EreMaijala: Fixed.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/result-list.phtml
EreMaijala: Fixed.

EreMaijala:
@demiankatz I think the last TODO item can be considered later if needed. The BC list looks ok to me. All unresolved conflicts and a couple of other omissions should be fixed now.

demiankatz (approved):
Thanks, @EreMaijala, this is looking good to me now!

PR #2704, opened 2023-02-13
Add output formatting for arrays on staffviewarray

natecollins:
Replace the print_r() formatting used on staffviewarray with an array renderer that outputs arrays in a YAML-like structure.

Also fixes issue where first-level array keys for field values were not being output.

TODO
- [x] Rename citation CSS class to staff-view in a separate PR after this is merged


natecollins:
Also, here's an example of the updated formatting: https://catalog-prod.lib.msu.edu/EdsRecord/edsagr,edsagr.US19900006166#details

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/PrintArrayHtmlTest.php
```diff
@@ -0,0 +1,199 @@
+<?php
```
demiankatz, line 1: There's an associated .PrintArrayHtmlTest.php.swp above which I suspect was committed inadvertently. Do you mind deleting that?

themes/bootstrap3/templates/RecordTab/staffviewarray.phtml
```diff
@@ -8,9 +8,13 @@
     <tr>
       <th><?=$this->escapeHtml($field)?></th>
       <td>
-        <?php foreach ((array)$values as $value): ?>
-          <?=$this->escapeHtml(is_array($value) ? print_r($value, true) : $value)?><br />
-        <?php endforeach; ?>
+        <?php
+          $values = (array)$values;
+          if (count($values) == 1 && is_int(array_key_first($values))) {
```
demiankatz: What is the purpose of this special case? It seems to me that if $values is ever a non-array value, it will get cast to an array and then shifted back into a non-array. Is this intentional?

module/VuFind/src/VuFind/View/Helper/Root/PrintArrayHtml.php
```diff
@@ -0,0 +1,83 @@
...
+     * Function uses recursion to achieve desired results, so entry can be
+     * either an array or a value to display.
+     *
+     * @param array|string $entry       An array or string to output
+     * @param int          $indentLevel How many spaces to indent output
+     * @param bool         $indentFirst Should first item in array be indented
+     *
+     * @return string                   The formatted HTML for output
+     */
+    public function __invoke($entry, $indentLevel=0, $indentFirst=true)
```
demiankatz: Minor nitpick but I'd put spaces around the = signs here for consistency of style.

demiankatz (changes requested):
Thanks for sharing this, @natecollins, it looks great! Just a few minor comments/questions.

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/PrintArrayHtmlTest.php
natecollins: Ah! Yes, obvious mistake on my part. Will correct it. And unless there is an objection, I'd like to add Vim swap files to the .gitignore (separate from this PR, of course).

themes/bootstrap3/templates/RecordTab/staffviewarray.phtml
natecollins: I wasn't sure why the cast to array was there before, so I left it in assuming it had a purpose.

Previously, all keys for arrays at the top level were not being displayed, whether integer or string. The above still removes the extra indentation level if there is only a single integer key (avoiding the extra indentation for the whole block of array data, like before), but will now show string key values in the output. Also, for arrays with multiple integer keys, it will add the hyphen to show when the next array value begins.

I made an assumption was that top-level keys being hidden previously was to avoid the common case of having single integer top-level key and wanting to skip displaying it. The above code was trying to preserve that.

demiankatz: The previous (array) cast was useful because it allowed us to iterate in a foreach loop over a value whether or not it was an array. Since the loop has been eliminated, I think you could probably now safely remove the cast, and instead just add an `is_array($values)` condition to the beginning of your if statement. That might make the code slightly more straightforward by eliminating the case where we create and then immediately unpack an array.

natecollins: Thanks for explaining, Demian. That makes a lot of sense. I've updated the branch; let me know if there is anything else that could be improved.

demiankatz (approved):
Thanks again, @natecollins! I think this is in good shape now, though I do wonder if we might be able to come up with a better name for the helper than PrintArrayHtml, since the helper can print non-array values too. Maybe FormatDataAsHtml? I'm going to ask for reviews from a couple of others in case they have further ideas! :-)

natecollins:
No problem, Demian! And thank you for your feedback. I also struggled with the name. Whatever the others decide to change it to is fine by me.

module/VuFind/src/VuFind/View/Helper/Root/PrintArrayHtml.php
```diff
@@ -0,0 +1,83 @@
...
+                // Increase indent if entering new array
+                $nextIndentLevel = is_array($value) ? $indentLevel + 2 : 0;
+                // Indent first line in values unless we're continuing from hypen
+                $nextIndentFirst = !is_int($key) || !is_array($value);
+
+                if (is_int($key)) {
+                    // Integer keyed arrays use a hypen list
+                    $html .= "&ndash;&ensp;";
+                } else {
+                    $html .= "<strong>" . $this->view->escapeHtml($key) .
```
EreMaijala: I'd use a span with a class here and format it with css. That'd allow for customized look depending on needs.

module/VuFind/src/VuFind/View/Helper/Root/PrintArrayHtml.php
```diff
@@ -0,0 +1,83 @@
...
+        $html = "";
+        if (is_array($entry)) {
+            foreach ($entry as $key => $value) {
+                if ($indentFirst || $key != array_key_first($entry)) {
+                    $html .= str_repeat("&ensp;", $indentLevel);
+                }
+
+                // Increase indent if entering new array
+                $nextIndentLevel = is_array($value) ? $indentLevel + 2 : 0;
+                // Indent first line in values unless we're continuing from hypen
```
EreMaijala: hyphen?

module/VuFind/src/VuFind/View/Helper/Root/PrintArrayHtml.php
```diff
@@ -0,0 +1,83 @@
...
+                $nextIndentFirst = !is_int($key) || !is_array($value);
+
+                if (is_int($key)) {
+                    // Integer keyed arrays use a hypen list
```
EreMaijala: hyphen?

module/VuFind/src/VuFind/View/Helper/Root/PrintArrayHtml.php
```diff
@@ -0,0 +1,83 @@
...
+                    $html .= "&ndash;&ensp;";
+                } else {
+                    $html .= "<strong>" . $this->view->escapeHtml($key) .
+                             ":</strong> " . (is_array($value) ? "<br/>\n" : "");
+                }
+
+                $html .= $this->__invoke($value, $nextIndentLevel, $nextIndentFirst);
+            }
+        } else {
+            $html = $this->view->escapeHtml($entry) . "<br/>\n";
```
EreMaijala: I'd also enclose the entry in a span so that css could be applied to it.

EreMaijala (changes requested):
@natecollins @demiankatz No doubt this is a huge improvement for EDS, but for non-hierarchical records (like Solr), I'm not sure. All the dashes for multivalued fields make the result a bit cluttered. Also the fact that the current indentation mechanism requires monospace font for good results indicates that this should probably use proper lists (`<ul>`) or css for indentation. I won't hold this up for this, however, since it's not critical in any way. But I flagged a couple of minor things I'd like to see.

demiankatz:
Thanks, @EreMaijala, all good points! I was thinking that as a replacement for a direct `print_r` this was clearly an improvement, but I hadn't considered the impact on flatter structures. One possibility could be to simply introduce the helper less aggressively -- just replace the print_r with the helper call instead of changing the way all data gets displayed. Another could be to make the helper omit hyphens at the top level.

I also agree that using `<ul>` and CSS would be a more elegant approach. @natecollins, let me know if you need any more specific advice about that (i.e. if you haven't worked with our LESS before and need some tips on getting started).

crhallberg (approved):
I am happy to commit this as is, but I have a few thoughts for future improvements.

1. The code might be neater and simpler with the `makeTag` helper. We have plans to apply this helper throughout the code in the near future, so it can be handled then.

2. Semantically, I think this would work better as a [definition list](https://benmyers.dev/blog/on-the-dl/). This would require more styling and wrapping code so it is not a requirement, and can likely be handled with the `makeTag` refactor.

natecollins:
Thanks for all the feedback! I've addressed most of the issues raised in the latest revision, including:
* Flat arrays are no longer hyphenated
* Use of the `makeTag` (this is a handy little helper!)

I have not yet converted the structure to be `ul` or `dl` based. I like the idea, but may need a rewrite to accomplish. I can tinker with that in my spare time, but I'll temporarily be on another dev team starting tomorrow for around a month, so it might be a while before I have something.

themes/bootstrap3/templates/RecordTab/staffviewarray.phtml
```diff
@@ -8,9 +8,12 @@
...
       <th><?=$this->escapeHtml($field)?></th>
       <td>
-        <?php foreach ((array)$values as $value): ?>
-          <?=$this->escapeHtml(is_array($value) ? print_r($value, true) : $value)?><br />
-        <?php endforeach; ?>
+        <?php
+          if (is_array($values) && count($values) == 1 && is_int(array_key_first($values))) {
+            $values = array_shift($values);
+          }
+        ?>
```
demiankatz: Do we need this special array_shift case anymore? It looks to me like with your latest changes, a flat value and a single-valued array actually display the same -- i.e.

```
        <?=$this->printArrayHtml("Foo")?>
```

and

```
        <?=$this->printArrayHtml(["Foo"])?>
```

look exactly the same to me.

demiankatz (changes requested):
Thanks, @natecollins, I think this is looking really good now -- the standard Solr case looks exactly the same before and after your change, and the EDS case is greatly improved, so this seems like it can only be considered a step in the right direction.

As noted earlier, we can probably do this with better, more semantic HTML, but I don't mind merging this as an incremental step, and we can revisit it again in the future when time permits.

One other side note that perhaps @crhallberg or @EreMaijala will have an idea about: why do we call the class used for staff view "citation"? This doesn't seem very appropriate, and strangely, our actual citations don't use the citation class. Maybe we should rename this to something more meaningful (or at least note this as an obvious improvement we can make in our future theme).

For now, there's just one easy simplification I'd like to propose before merging...

themes/bootstrap3/templates/RecordTab/staffviewarray.phtml
natecollins: Quite true! It's not needed anymore as that case is handled within the helper now. It's been corrected; I also added passing it a string value into the unit tests.

demiankatz (approved):
Thanks again, @natecollins! I think this is in a state where it can be merged, though I still wonder if we could improve the helper name. What about something simple like `RenderDataStructure` or `RecursiveRender`?

I'll request a re-review from @EreMaijala so he can clear his pending review (which I believe is now fully resolved) and perhaps chime in with thoughts on naming.

crhallberg:
> why do we call the class used for staff view "citation"?

That doesn't seem right, in fact it's misleading! I think something like `staff-view` would be better.

([the line of code in question](https://github.com/vufind-org/vufind/pull/2704/files#diff-76d27d14e7cecf74dbd7bd124d177ce4eadf4b850564c12fd7e6d2f435951d2eR5))

demiankatz:
Thanks, @crhallberg -- I've added a TODO to the top of this PR to remind us to open a separate PR to address the badly-named class after this gets merged. I think that's better than trying to address two unrelated issues in this single PR. :-)

EreMaijala (approved):
Looks good to me. Still wondering if it would make sense to use a monospaced font in the EDS staff view by default, but it works quite alright as is. I actually prefer the current name `printArrayHtml` over the other options for being less abstract.

demiankatz:
Thanks, @EreMaijala -- if the current naming is good enough for you, I won't waste people's time trying to change it. :-) I've merged this and begun #2714 as a place to complete follow-up work discussed here (and discovered subsequently).

PR #2705, opened 2023-02-13
Add Vim swap files to .gitignore

natecollins:


demiankatz (approved):
Thanks, @natecollins!

PR #2706, opened 2023-02-14
Add retry support to Redis session handler.

EreMaijala:
If Redis is configured with a short timeout, a long-running request (or utility) may be unable to process the session since the Redis client gets disconnected in the middle. Introduces a new generic retry trait to help implementation.

### DIFF
--- a/module/VuFind/src/VuFind/Service/Feature/RetryTrait.php
+++ b/module/VuFind/src/VuFind/Service/Feature/RetryTrait.php
@@ -58,7 +58,8 @@ trait RetryTrait
      *
      * @param callable  $callback       Method to call
      * @param ?callable $statusCallback Status callback called before retry and after
-     * a successful retry. The callback gets the number of
+     * a successful retry. The callback gets the attempt number and either an
+     * exception if an error occurred or null if the request succeeded after retries.
      * @param array     $options        Optional options to override defaults in
      * $this->retryOptions. Options can also include a retryableExceptionCallback
      * for a callback that gets the attempt number and exception as parameters and
--- a/module/VuFind/src/VuFind/Service/Feature/RetryTrait.php
+++ b/module/VuFind/src/VuFind/Service/Feature/RetryTrait.php
@@ -1,0 +1,127 @@
+<?php
+/**
+ * Trait that provides support for calling a method with configurable retries
+ *
+ * PHP version 7
+ *
+ * Copyright (C) The National Library of Finland 2023.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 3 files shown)


module/VuFind/src/VuFind/Service/Feature/RetryTrait.php
```diff
@@ -0,0 +1,127 @@
...
+                $result = $callback();
+                if ($attempt > 1 && $statusCallback) {
+                    $statusCallback($attempt, true);
+                }
+                return $result;
+            } catch (\Exception $e) {
+                if (null === $exception) {
+                    $exception = $e;
+                }
+                if ($checkCallback = $options['retryableExceptionCallback'] ?? '') {
```
demiankatz, line 98: This option is not documented above.

module/VuFind/src/VuFind/Service/Feature/RetryTrait.php
```diff
@@ -0,0 +1,127 @@
...
+        callable $callback,
+        ?callable $statusCallback = null,
+        array $options = []
+    ) {
+        $attempt = 0;
+        $exception = null;
+        $options = array_merge($this->retryOptions, $options);
+        do {
+            if (++$attempt > 1 && $statusCallback) {
+                $statusCallback($attempt, false);
```
demiankatz: What if, instead of a boolean, the status callback received either an exception (failure) or null (success)? This would allow you to pass the exception on each iteration, which might be useful for logging/debugging if you get different exceptions on each retry. As it is, any exception after the first is going to be masked, which might make some issues harder to identify.

module/VuFind/src/VuFind/Service/Feature/RetryTrait.php
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Check if the call needs to be retried and sleep before the retry
+     *
+     * @param int   $attempt Failed attempt number
+     * @param array $options Current options
+     *
+     * @return bool
+     */
+    protected function checkRetryAndSleep(int $attempt, array $options): bool
```
demiankatz: Maybe it would make sense to refactor this so that there's a separate getSleepDurationForAttempt method, and then we just pass that value to usleep. That would probably allow you to make your test a little more straightforward (you could test that sleep durations are calculated correctly without having to actually sleep and measure the duration), and it might slightly improve readability here.

demiankatz (changes requested):
Thanks, @EreMaijala, this looks great! Just a few minor suggestions about the new trait...

EreMaijala:
@demiankatz Thanks for the suggestion. Refactored, should be done now.

module/VuFind/src/VuFind/Service/Feature/RetryTrait.php
```diff
@@ -58,9 +58,11 @@ trait RetryTrait
      *
      * @param callable  $callback       Method to call
      * @param ?callable $statusCallback Status callback called before retry and after
-     * a successful retry
+     * a successful retry. The callback gets the number of
```
demiankatz: This comment appears to be incomplete.

demiankatz (changes requested):
Looks good to me except for one small documentation issue.

module/VuFind/src/VuFind/Service/Feature/RetryTrait.php
EreMaijala: Fixed.

demiankatz (approved):
Thanks, @EreMaijala!

PR #2710, opened 2023-02-16
Fix labeling of blended results and the Blender test.

EreMaijala:
While working on tests for #2701 I noticed that the Blender test wasn't working properly and didn't check the labels of the records. After fixing that I realized that Blender added labels for the initial batch of records only, so on second result screen some of the records were missing the label. This fixes both issues.


demiankatz (approved):
Makes sense to me -- thanks for the fix, @EreMaijala!

PR #2711, opened 2023-02-16
Make Solr cores independently configurable

demiankatz:
This PR makes Solr configuration more flexible -- subsidiary Solr cores (authority, reserves, website) can now be more consistently configured independently of the default biblio core.

TODO
- [x] Consider renaming solrCore property to defaultSolrCore (for better readability but reduced backward compatibility)
- [x] Add more test coverage, if needed
- [x] Run full test suites
- [x] Document backward compatibility breaks when merging ("Search backend factories extended from AbstractSolrBackendFactory may need minor adjustments due to improvements to the design of configuration loading within the factories. The $solrCore property has been renamed to $defaultIndexName, and the getSolrCore() method has been renamed to getIndexName(). The getSolrUrl() method no longer accepts any parameters.")


demiankatz:
@EreMaijala, I ran into a situation where I needed to configure the reserves and biblio cores to point at different Solr indexes and realized there wasn't a way to do it. In solving that problem, I ended up doing a lot of refactoring, which resulted in what I think is a more uniform approach to loading index configuration. I've done quite a bit of hands-on testing but haven't done formal automated testing yet -- I'm interested in your thoughts on this approach before I apply final polish.

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
```diff
@@ -183,6 +192,55 @@ public function __invoke(ContainerInterface $sm, $name, array $options = null)
...
+     *
+     * @param string $setting    Name of setting
+     * @param mixed  $default    Default value if unset
+     * @param string $configName Name of config to look in
+     * @param bool   $fallback   Should we fall back to main config if the
+     * setting is absent from the specified config file?
+     *
+     * @return mixed
+     */
+    protected function getIndexConfig(
```
EreMaijala, line 264: I find this method a bit confusing. It might be clearer to first determine the paths to check according to `$fallback` and then just loop through the path array.

Before that, however, my concern is that this method potentially does a lot of repeated work for each setting retrieved. How would you feel about merging the config sections the first time and caching the results?

EreMaijala (changes requested):
Looks mostly good to me, but now would be a great opportunity to change the term we use from core to collection. Core in Solr has a very specific meaning, and this could cause confusion. As far as I can see we actually mean a collection whenever we mention core. 

Does this cause any issues for the [IndexShards] settings?

demiankatz:
@EreMaijala, regarding core vs. collection, my understanding is that there is no such thing as a collection unless you run Solr in distributed/cloud mode. Since VuFind was built before that mode even existed, and since I've only ever run Solr in single-node mode, it's not surprising that the core terminology is fairly pervasive. The Solr admin panel still has a "core selector" which I use and refer to frequently; does that have a different label if you're running SolrCloud?

In any case, it would definitely be preferable to use up-to-date and clear terminology... but I think most changes will either a) introduce backward compatibility problems or b) require us to have even more complicated logic checking for multiple possible config settings, and c) probably will require some documentation updates as well. We should plan carefully before we leap! :-)

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
demiankatz: Good suggestion, @EreMaijala! My strategy evolved considerably while I was working on this code -- originally I was expecting to override the getIndexConfig method in the various subclasses, and my initial design came from that idea... but when I realized I could instead drive all of this through properties and data-oriented methods for simplicity, I didn't simplify the code as much as I should have!

I've done some significant rewriting, and I think it should be more efficient and easier to understand now.

I realize that my code changes have added even more instances of the word "core" -- I just did that for consistency with the rest of the code for now; we can figure out a strategy for improving it once we have time to discuss in more detail. :-)

EreMaijala:
@demiankatz Core selection is there also in cloud mode, but a core is just a part of a collection. A collection can certainly still consist of just a single core, but typically there are many cores which are either shards or replicas.

I suppose a more generic term would be "index". What do you think? I don't think we need to change all settings, but the comments and function names could reflect this.

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
```diff
@@ -183,6 +206,69 @@ public function __invoke(ContainerInterface $sm, $name, array $options = null)
         return $backend;
     }
 
+    /**
+     * Return an ordered array of configurations to check for index configurations.
+     *
+     * @return string[]
+     */
+    protected function getIndexConfigFallbackPath(): array
```
EreMaijala: I'd rename this to something like `getIndexConfigPaths`. Current name sounds like it just returns a fallback instead of the all.

EreMaijala (changes requested):
@demiankatz Looks good now! Only one minor thing and the terminology (if we decide to do anything about it).

module/VuFind/src/VuFind/Search/Factory/AbstractSolrBackendFactory.php
demiankatz: I changed it to `getPrioritizedConfigsForIndexSettings` -- long, but more accurate.

demiankatz:
Thanks, @EreMaijala -- I've revised comments in the config files to hopefully clarify the core situation, and I've significantly revised comments and property/method names in the factory to reduce use of the term "core." There are references to cores elsewhere in the codebase (e.g. in the VuFindSearch module), but I did not attempt to address those here to avoid scope creep.

demiankatz:
All tests are passing!

EreMaijala (approved):
Thanks, looks good to me!

PR #2713, opened 2023-02-17
[VUFIND-1597] include available ISSNs in OpenURLs for Serial format

gmcharlt:
Records whose format ends up as Serial due to vagaries of the cataloging of the original MARC record should nonetheless have any available ISSNs included in their OpenURL. With this patch, the OpenURL format is set as 'Journal' for such records.

TODO
- [x] Resolve [VUFIND-1597](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1597) when closing

### DIFF
--- a/module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
+++ b/module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
@@ -686,11 +686,24 @@ class DefaultRecord extends AbstractBase
             return 'Article';
         } elseif (in_array('Journal', $formats)) {
             return 'Journal';
+        } elseif (strlen($this->getCleanISSN()) > 0) {
+            // If the record has an ISSN and we have not already
+            // decided it is an Article, we'll treat it as a Book
+            // if it has an ISBN and is therefore likely part of a
+            // monographic series. Otherwise, we'll treat it as a
+            // Journal.
+            // Anecdotally, some link resolvers do not return correct
+            // results when given both ISBN and ISSN for a member of a
+            // monographic series.
+            return strlen($this->getCleanISBN()) > 0 ? 'Book' : 'Journal';
         } elseif (isset($formats[0])) {
             return $formats[0];
-        } elseif (strlen($this->getCleanISSN()) > 0) {
-            return 'Journal';
         } elseif (strlen($this->getCleanISBN()) > 0) {
+            // Last ditch. Note that this is last by intention; if the
+            // record has a format set and also has an ISBN, we don't
+            // necessarily want to send the ISBN, as it may be a game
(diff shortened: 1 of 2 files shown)


gmcharlt:
Noting that the approach I took hews closest to the specific problem I ran into, but an alternative approach would be to simply move the test of `isset($formats[0])` to the end of the if-chain on the theory that getting any available ISSN or ISBN into the OpenURL when the format isn't precisely known to be Article, Book, or Journal will be good enough.

demiankatz:
Thanks, @gmcharlt! I've put a link to the JIRA ticket in the header here, and I've also commented with some more background/context over there. I think some more discussion may be necessary here. :-)

module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
```diff
@@ -686,12 +686,19 @@ protected function getOpenUrlFormat()
...
-        } elseif (strlen($this->getCleanISSN()) > 0) {
-            return 'Journal';
         } elseif (strlen($this->getCleanISBN()) > 0) {
+            // If the record also has an ISSN and we have not already
+            // decided it is an Article, we treat it as a Book on the
+            // assumption that it is part of a monographic series and
+            // its ISBN is more likely to precisely identify it.
+            // Anecdotally, some link resolvers do not return correct
+            // results when given both ISBN and ISSN for a member of a
+            // monographic series.
```
demiankatz, line 697: Should we restructure this to:

```
} elseif (strlen($this->getCleanISSN()) > 0) {
    return strlen($this->getCleanISBN()) > 0 ? 'Book' : 'Journal';
....
```
...rather than having the ISBN check done independently? That way, we're more specifically adding handling for things with ISSNs, rather than very broadly capturing ISBNs. I'm concerned that we might otherwise have some unanticipated side effects for non-Book objects that happen to have ISBNs (DVDs, board games, etc.).

demiankatz (changes requested):
Thanks, @gmcharlt, I think this is moving in the right direction, but I had one more thought...

gmcharlt:
@demiankatz I've updated the PR with the requested change. This is on track to become a very thoroughly documented little function. :)

demiankatz (approved):
Thanks, @gmcharlt, I think this all makes a lot of sense. However, before I merge, I'm going to ask for a couple more reviews just in case somebody else thinks of something I'm overlooking!

lahmann (approved):
Thanks for the contribution @gmcharlt - I have no objections and double checked possible implications with a colleague of mine: this change will very likely improve resolver's handling of proceedings at least in our catalogues so we welcome this pull request.

demiankatz:
Thanks, @gmcharlt and @lahmann -- merging now!

PR #2714, opened 2023-02-20
Improve/simplify staff view code/styles.

demiankatz:
This is a follow-up to #2704.

Changes:
- Rename citation class to staff-view for accuracy
- Simplify Overdrive staff view tab to eliminate redundant logic
- Incorporate work by @natecollins to improve the printArrayHtml helper

TODO
- [x] Fix problem with mixed int/non-int array keys (described below)


demiankatz:
@natecollins, while testing this, I ran into a problem that I think we should try to fix here: when an array has a mix of integer and non-integer keys, the output can look strange. I discovered this when testing the Overdrive staff view. If you render VuFind's internal MARC data structure as an array, you get an odd result, because the "0xx" fields are treated as non-integers with visible keys, but the other MARC fields ("1xx", "2xx", etc.) are treated as integers and the keys are suppressed. This results in a display that looks like this:

image

Note the double hyphen instead of "100" on the author field.

I think we need to be a little more aggressive about displaying keys to avoid this problem -- I think it's better to risk showing too much than to lose important context like this. I think maybe if an array has ANY non-integer keys, we should treat all keys as displayable. Or maybe if an array has non-consecutive keys, we should display all of them. What do you think?

natecollins:
Good catch, @demiankatz. I've created a [PR](https://github.com/demiankatz/vufind/pull/6) into your branch with a change that has it always display mixed key array keys, with a new test added as well. Let me know if that solves the issue.

For future VuFind that supports it, PHP8 has a new function [array_is_list](https://www.php.net/manual/en/function.array-is-list.php) which is slightly logically different, but might serve as a better option than my integer key check function.

demiankatz:
As of this morning, https://github.com/demiankatz/vufind/pull/6 has been merged in here to improve the flexibility of printArrayHtml. Thanks again, @natecollins!

EreMaijala (approved):

PR #2715, opened 2023-02-20
Make sure getCurrentSearchId returns a null instead of an empty string.

EreMaijala:


module/VuFind/src/VuFind/Search/Memory.php
```diff
@@ -243,8 +243,9 @@ public function retrieveSearch()
      */
     public function getCurrentSearchId(): ?int
     {
-        return $this->request->getQuery('sid')
+        $sid = $this->request->getQuery('sid')
             ?? $this->request->getPost('sid');
+        return $sid ?: null;
```
demiankatz: Should we include an `intval` or equivalent here to ensure that we really always return an integer? Otherwise, odd values from the query could get passed through as-is.

demiankatz (changes requested):
Thanks, @EreMaijala, just one question.

module/VuFind/src/VuFind/Search/Memory.php
EreMaijala: Good point, intval added.

EreMaijala: Odd values would have blown up because of the return type, but it's better not to rely on that.

demiankatz (approved):

PR #2716, opened 2023-02-20
SierraRest: Add option to display orders and fix display of itemless …

EreMaijala:
…holdings.

Orders are displayed like items, but only one item per location with further information in the item notes. 

Also changes \DateTime:ISO8601 to \DateTime::ATOM since the former is deprecated.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
@@ -2108,7 +2108,9 @@ class SierraRest extends AbstractBase implements TranslatorAwareInterface,
             $messages[] = $this->translate(
                 [
                     'HoldingStatus',
-                    ($order['copies'] === 1 ? 'copy' : 'copies') . '_ordered_on_date'
+                    1 === $order['copies']
+                        ? 'copy_ordered_on_date'
+                        : 'copies_ordered_on_date'
                 ],
                 [
                     '%%copies%%' => $order['copies'],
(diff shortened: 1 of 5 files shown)


module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -2060,10 +2071,57 @@ protected function getItemStatusesForBib($id, $checkHoldings)
...
+     * @return array
+     */
+    protected function getOrderMessages(array $orders): array
+    {
+        $messages = [];
+        foreach ($orders as $order) {
+            $messages[] = $this->translate(
+                [
+                    'HoldingStatus',
+                    ($order['copies'] === 1 ? 'copy' : 'copies') . '_ordered_on_date'
```
demiankatz: As discussed on a different recent PR, maybe it's better to use full translation strings here to ensure findability if we're searching to see where particular strings are used in the code.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -2511,12 +2571,34 @@ function ($matches) {
...
+            }
+        } else {
+            $cached = [
+                'fields' => [],
+                'data' => [],
+            ];
+        }
+        // Fetch requested fields as well as any cached fields to keep everything in
+        // sync:
+        $allFields = [...$fieldsArray, ...$cached['fields']];
```
demiankatz: Will this automatically deduplicate, or should we do an array_unique?

demiankatz (changes requested):
Thanks, @EreMaijala, just a couple of minor suggestions.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Right back at me :) Fixed!

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: No, it's just like array_merge but faster (and doesn't handle string keys before PHP 8.2). array_unique added.

demiankatz (approved):
Thanks, @EreMaijala!

PR #2717, opened 2023-02-20
Fix problem with request URLs caused by #2701

demiankatz:
TODO
- [x] Add a test somewhere to catch the underlying problem that this fixes

### DIFF
--- a/config/vufind/Demo.ini
+++ b/config/vufind/Demo.ini
@@ -138,7 +138,7 @@ updateHolds = 25
 [Holds]
 ; HMACKeys - A list of hold form element names that will be analyzed for consistency
 ; during hold form processing. Most users should not need to change this setting.
-HMACKeys = "id:item_id:level"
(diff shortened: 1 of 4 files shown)


EreMaijala (approved):
Thanks! It would probably be cleaner if we'd pass the extra query params to getActionUrl instead of appending them manually, but this is good enough for me.

EreMaijala:
@demiankatz I tweaked the testPlaceHold test and Demo driver here so that the request would fail without the fix. I also pushed a commit to pass query parameters and anchor to getActionUrl. What do you think?

EreMaijala:
@demiankatz Note that oddly enough normalize_path needs to be set to false. Otherwise the router will mangle encoded slashes in a record ID, but only when the `fragment` option is passed to it, and even when it's an empty string.

demiankatz:
Thanks, @EreMaijala!

PR #2718, opened 2023-02-21
SierraRest: Use RetryTrait to retry requests that receive an empty re…

EreMaijala:
…ply.

Sierra, depending on server configuration, may sometimes fail with an empty reply. An immediate retry was recommended by the vendor.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
@@ -213,6 +213,14 @@ class SierraRest extends AbstractBase implements TranslatorAwareInterface,
     protected $checkFreezability = false;
 
     /**
+     * Number of retries in case an API request fails with a retryable error (see
+     * $retryableRequestExceptionPatterns below).
+     *
+     * @var int
+     */
+    protected $httpRetryCount = 2;
+
+    /**
      * Exception message regexp patterns for request errors that can be retried
      *
      * @var array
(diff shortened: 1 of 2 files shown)


module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -1504,6 +1517,76 @@ protected function makeRequest(
...
         $patron = false,
         $returnStatus = false
+    ) {
+        // Status logging callback:
+        $statusCallback = function ($attempt, $exception) use (
+            $hierarchy,
+            $params,
+            $method
+        ): void {
+            $apiUrl = $this->config['Catalog']['host'];
```
demiankatz: Would it make sense to have a getUrlFromHierarchy support method rather than having this logic in two places?

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -1504,6 +1517,76 @@ protected function makeRequest(
...
+        };
+
+        $args = func_get_args();
+        return $this->callWithRetry(
+            function () use ($args) {
+                return call_user_func_array([$this, 'requestCallback'], $args);
+            },
+            $statusCallback,
+            [
+                'retryCount' => 2,
```
demiankatz: Would it make sense for this to be configurable via SierraRest.ini?

demiankatz (changes requested):
Thanks, @EreMaijala, just a couple of minor suggestions.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Yes, method added.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Configuration added.

demiankatz (approved):
Thanks, @EreMaijala, looks good to me now!

PR #2720, opened 2023-02-23
Upgrade to PHPUnit 9.6 and fix deprecations.

demiankatz:
This makes progress on [VUFIND-1595](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1595) (by upgrading a key dependency) and lays some groundwork for [VUFIND-1596](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1596) (by fixing some feature usage that won't work when we are ready to upgrade to PHPUnit 10).

TODO:
- [x] Figure out why mocking \Memcache breaks in PHP 7.4.


demiankatz:
@EreMaijala, here's a weird one: PHPUnit 9.6 breaks on one test in PHP 7.4 because it can't mock \Memcache successfully. I'm not sure why this might be, and various simple things I tried (disabling the original constructor, eliminating the "onlyMethods" call from the mock builder) did not make any difference.

If this is related to some low-level problem in PHPUnit or its support library, I doubt we'll get any traction fixing it since PHP 7.4 is EOL anyway. Thus, unless you can think of a better solution, I think we have a couple of options:

1.) Wait to merge this until we raise the minimum PHP version to 8
2.) Skip this test in PHP 7

Any ideas/preferences?

EreMaijala:
@demiankatz I'd prefer skipping the test with PHP 7.4. I doubt there's anything that would be caught that other versions wouldn't catch (apart from not being able to mock the class, obviously :) ).

demiankatz:
@EreMaijala, I'm still puzzled over why this test breaks, but in any case, I've gone ahead and added a skip for it in PHP 7. I noticed that we were only testing the Memcache session handler using the Memcache driver, but not \Memcached, so I added a new test that does the read test with Memcached instead. This one works in PHP 7, so we now have the same coverage in PHP 7 and PHP 8, but with the bonus that PHP 8 is testing both variations of the connection.

EreMaijala (approved):
Excellent! Looks good to me!

PR #2724, opened 2023-02-24
Make EDS icon CSS more specific.

demiankatz:
The icon styles introduced in #2400 conflicted with some existing EDS-specific styles. This PR makes the EDS styles more specific to eliminate the conflict. I'd be interested to hear if @crhallberg or @EreMaijala prefer a different approach, though!

Before the fix:

image

After the fix:

image


EreMaijala (approved):
I'm okay with this, though in the long run I'd prefer normal font icons.

demiankatz:
@EreMaijala, I agree that in the long term it would be nice if the EDS module were a little more standardized; this is one of a few ways in which the EBSCO-contributed code is a bit quirky!

@crhallberg, would it make sense to create a new style of icon template for these EBSCO-provided icons? Then we could use the icon helper and configuration to standardize this while retaining the existing behavior.

crhallberg:
I would make the class name longer and clearer: `icon--eds`. Then, if possible, I would use that class (`.icon--eds`) instead of `.eds.icon` to prevent specificity creep.

demiankatz:
@crhallberg, is this what you had in mind?

crhallberg (approved):

PR #2725, opened 2023-02-24
Add search id to versions and full item status links.

EreMaijala:
This adds the sid parameter to record page links of record versions and the more links from full item status template.

### DIFF
--- a/themes/bootstrap3/templates/RecordDriver/DefaultRecord/versions-link.phtml
+++ b/themes/bootstrap3/templates/RecordDriver/DefaultRecord/versions-link.phtml
@@ -5,7 +5,7 @@
   }
   if (isset($this->tabs['Versions'])) {
     $url = !empty($this->full)
-      ? ($this->recordLinker()->getTabUrl($this->driver, 'Versions', [], ['query' => ['sid' => $this->searchId], 'fragment' => 'tabnav']) . '#tabnav')
+      ? ($this->recordLinker()->getTabUrl($this->driver, 'Versions', [], ['query' => ['sid' => $this->searchId]]) . '#tabnav')
       : '#versions';
     $translationKey = 'other_versions_link';
   } else {
--- a/themes/bootstrap3/templates/RecordDriver/DefaultRecord/versions-link.phtml
+++ b/themes/bootstrap3/templates/RecordDriver/DefaultRecord/versions-link.phtml
@@ -4,7 +4,9 @@
     return;
   }
   if (isset($this->tabs['Versions'])) {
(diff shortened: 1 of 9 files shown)


themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -86,12 +86,14 @@
...
       $dsb = DEFAULT_SEARCH_BACKEND;
       $cspNonce = $this->cspNonce();
+      $searchId = $this->layout()->searchId;
       $appendScripts[] = <<<JS
 VuFind.path = '{$root}';
 VuFind.defaultSearchBackend = '{$dsb}';
 VuFind.addTranslations({$translations});
 VuFind.addIcons({$icons});
 VuFind.setCspNonce('{$cspNonce}');
+VuFind.setCurrentSearchId($searchId);
```
demiankatz, line 96: Should we be more explicit about what happens when searchId is not set? Right now, it doesn't seem to hurt anything, but we end up calling setCurrentSeachId without a parameter, which I don't think is exactly expected behavior. Maybe we should explicitly pass a null here, or else add a default value to the method's parameter in common.js.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/versions-link.phtml
```diff
@@ -4,7 +4,9 @@
     return;
   }
   if (isset($this->tabs['Versions'])) {
-    $url = !empty($this->full) ? $this->recordLinker()->getTabUrl($this->driver, 'Versions') . '#tabnav' : '#versions';
+    $url = !empty($this->full)
+      ? ($this->recordLinker()->getTabUrl($this->driver, 'Versions', [], ['query' => ['sid' => $this->searchId], 'fragment' => 'tabnav']) . '#tabnav')
```
demiankatz: Are we appending the fragment twice here?

demiankatz (changes requested):
Thanks! A couple of minor observations...

themes/bootstrap3/templates/layout/layout.phtml
EreMaijala: Oops, yes. I added json_encode so that it can handle any value just in case.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/versions-link.phtml
EreMaijala: Actually, no. The parameter didn't work. I misread the code thinking that it would, but forgot to remove the extra parameter.

demiankatz (approved):
Looks good to me, and all tests are passing. Thanks!

PR #2728, opened 2023-02-24
Update assorted minor dependencies for VuFind 9.0.

demiankatz:
TODO
- [x] Test everything


demiankatz:
@EreMaijala, some more minor dependency updates. I checked all changelogs and most changes are minor (there's a security fix for PHPcas which may be significant, though I don't think this is commonly used). I also did a visual inspection that QR codes still generate correctly, and I manually tested a harvest to be sure the major VuFindHarvest upgrade didn't impact integration.

demiankatz:
At this point, all Composer dependency updates are done, except for php-cs-fixer (which has some complications that require separate discussion -- I'll review again when time permits) and swagger-ui (which perhaps you could look at if time permits; I suspect it's safe to simply update, but I'm less experienced at testing that the upgrade is successful).

EreMaijala (approved):

PR #2729, opened 2023-02-24
Upgrade to php-cs-fixer 3.14.4, adjust phpcs rules/code to match.

demiankatz:
The latest release of php-cs-fixer indents case statements according to PSR-12, which differs from the PEAR expectations of PHP_CodeSniffer. Since PSR-12 is a more current standard, it seems sensible to comply with it, so this PR adjust the code according to the new php-cs-fixer behavior and turns off the PHP_CodeSniffer rules that are triggered by the changes. It appears that PHP_CodeSniffer does include PSR-12 support, so a future project to investigate the consequences of changing PHP_CodeSniffer settings more comprehensively might be worthwhile.

Note that beyond the case changes, there are a few other minor adjustments -- php-cs-fixer also now de-indents "implements" in multi-line class declarations and removes more empty lines from functions.


EreMaijala (approved):
First I found the de-indentation of `implements` odd, but it makes total sense with the multi-line `implements`in PSR-12. So, looks good to me after digesting a bit. :)

PR #2730, opened 2023-02-28
FOLIO: Refactor getHolding into smaller parts.

demiankatz:
This PR is a follow-up to discussion in #2726.

TODO:
- [x] Add missing test coverage


demiankatz:
@bbusenius, as discussed yesterday, here's a possible approach to refactoring getHolding into smaller parts. There are definitely multiple ways to approach this, but this felt least disruptive.

When I have a little more time, I want to add a test to fully cover the getHolding method so I can be more confident that I haven't broken anything here. In the meantime, though, I'm interested to hear whether you agree with this approach or have any suggestions for improvement.

bbusenius (approved):
I've been going over this and thinking about it. At a quick glance the method names `getHoldingDetailsForItem` and `formatHoldingItem` didn't communicate to me the "separate levels of holdings vs. items" very well, however, knowing how the code works the more I thought about the names, they really started to make a lot of sense. They help communicate the idea that we're returning items with holdings data rather than holdings and items, if that makes sense. That was an idea that I initially struggled with when I started working on VuFind code many years ago. So, in the end, I actually think those names are kind of didactic and they communicate something about the code. I like that.

Beyond that, I can see how this mostly just moves the code around a bit. From what I can tell, I would be able to move our many customizations into the new corresponding locations and they would work just the same. I think you're right about this being a less disruptive approach. I'm supportive of that and supportive of a unit test considering the high stakes this functionality represents.

demiankatz:
Thanks, @bbusenius! I think the best course of action at this point is for me to open a new PR to add test coverage to the existing FOLIO driver's getHolding method. That way, we can ensure that we have 100% coverage of the existing code before running tests against the refactored code. If all goes well, once that PR is ready, we can merge it, merge dev into this, and feel confident that we didn't break anything. :-)

demiankatz:
@bbusenius, all tests are still passing here (though I had to revise one test fixture to reflect the change that due date lookups now take place before location lookups instead of after -- but this makes no functional difference to the output). I'll merge this as soon as the build passes here on GitHub!

PR #2733, opened 2023-03-01
HTTPS support for Wikipedia API Connection

j4lib:
As mentioned on slack, there seems to be some HTTPS-only policy in the Wikipedia API. Here comes the fix with HTTPS support.


demiankatz (approved):
Looks good -- thanks, @j4lib!

At some point, it might make sense to refactor so there's a single support method to return the base API URL, but I don't think it's worth the effort to change that right now. :-)

PR #2734, opened 2023-03-01
[VUFIND-1593] Add an HTML-friendly JSON encode helper.

demiankatz:
TODO:
- [x] Test everything
- [x] Close [VUFIND-1593](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1593) when merging


demiankatz:
@EreMaijala, I have a bit more testing to do, but I'm interested in your thoughts on this approach. A few notes:

1.) I came up with a better fix for the channel escaping than #2687 -- I figured out where the double-escaping was coming from (the URL query helper in the channel provider plugin).

2.) I assume that I don't need to escape JSON differently inside inline scripts, so I just focused on JSON in HTML attributes.

3.) For consistency, a view helper seemed like the best approach... but I'm open to other ideas!

demiankatz:
At this point, all integration tests are passing, though there are some less-used templates (like WorldCatIdentities) that probably need to be manually double-checked. I'll wait to finish all the manual testing until the general approach has been approved, so I only have to do it once. :-)

module/VuFind/src/VuFind/View/Helper/Root/HtmlSafeJsonEncode.php
```diff
@@ -0,0 +1,62 @@
...
+     * @return string
+     */
+    public function __invoke($value, ?string $outerEscaper = 'escapeHtmlAttr')
+    {
+        $json = json_encode(
+            $value,
+            JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP
+        );
+        return $outerEscaper
+            ? $this->getView()->plugin($outerEscaper)->__invoke($json)
```
EreMaijala: Could you use the `($this->getView()->plugin($outerEscaper))($json)` style call here?

module/VuFind/src/VuFind/View/Helper/Root/HtmlSafeJsonEncode.php
```diff
@@ -0,0 +1,62 @@
+<?php
+/**
+ * HTML-safe JSON encoding.
+ *
```
EreMaijala, line 4: This could include a short description on why this is needed.

EreMaijala (changes requested):
Makes sense to me! Just a couple of tiny nitpicks in comments.

demiankatz:
Thanks, @EreMaijala! I've incorporated your feedback, manually tested things, and added a unit test for the new helper. Hopefully this is ready to go now!

EreMaijala (approved):

PR #2737, opened 2023-03-02
Fix collection search links.

EreMaijala:
Fixes Search2Default link-collection template to match that of DefaultRecord and fixes encoding of returned URLs to match other templates and the expected result.

Since Record helper's getLink is expected to return the URL with HTML special characters encoded as entities.


demiankatz (approved):
Thanks, @EreMaijala.

I considered suggesting parameterizing the route name in the DefaultRecord's template, so that the secondary template could wrap around it and simply provide a different route name (to reduce redundancy), but that may be overkill in this case. I'll merge as-is for now. :-)

PR #2738, opened 2023-03-03
Fix deprecation in DbUpgrade helper.

demiankatz:
While reviewing database upgrade code, I noticed that we were using a deprecated feature. This PR fixes it.


EreMaijala (approved):
Good catch!

PR #2735, opened 2023-03-01
Improve FOLIO test suite

demiankatz:
This PR is designed to improve the overall FOLIO ILS driver test suite

TODO
- [x] Make fixtures more robust (check input in addition to simulating output)
- [x] Create a recorder to make fixture generation easier
- [x] Expand test coverage

PR #2739, opened 2023-03-03
Handle empty requiredby date in Folio placeHold function

meganschanz:
We discovered this issue today where if you configure in the `folio.ini` to remove the `requiredByDate` from the `extraHoldFields` setting, it will have no value in the `requiredBy` field that is sent to the `placeHold` function in the Folio driver. Since there currently is no check to see if the value is set or not before it attempts to convert it to a date format, it will result in `$requiredBy` being the current date; effectively expiring holds the day they are received. 

Adding the `isset` check beforehand and defaulting to `null` if it is not set will pass a `null` value to the Folio API when the `requiredByDate` is not configured to be on the holds form, but still through a `hold_date_invalid` exception when the user clears out the date if the `requiredDyDate` is set to display on the form (because the field is set, it's just empty). I figured this would be the desired behavior, and spot checking a few other drivers, like Alma's, this seemed consistent with how they are doing it.

And I figured while I was in there I'd make some unit tests for it, so of course now my couple line functionality change ballooned into many more due to the unit tests!

As always, I'm open to suggestions for improvements :).


meganschanz:
Alright there had to be a snag! I ran the unit tests on my php8 environment, but luckily the [CI is catching something I missed](https://github.com/vufind-org/vufind/actions/runs/4327054064/jobs/7555229621#step:13:78). I can't figure out the correct way to tell the tests that a request should be using the `POST` method. I'm not seeing what the `successful-place-hold` fixture has defined differently that is making that happen since it's not in the file and it's not in the test function unless I'm overlooking it. 

And the other error in the CI is just that it needed a new fixture with different response data (without the expiration date in it). I'm holding off on committing it since I know the pipeline will fail due to the POST issue.

demiankatz:
Sorry, @meganschanz, I pretty significantly revised the FOLIO test suite this week (see #2735) to make it a little more robust. I wasn't expecting anyone else to be writing tests at the same time as me! Apologies for any confusion!

Fortunately, the fix should be easy: just add:

```
"expectedMethod": "POST",
```

into the appropriate step of your JSON fixture.

Let me know if you still run into trouble!

demiankatz:
Also, to elaborate on the changes from #2735, the point of it is to enforce matching not just the outputs from FOLIO but also the inputs from VuFind in our tests. Previously, the test suite didn't really care what requests VuFind was making in most of the tests, which meant the testing wasn't very thorough. Now, through the `expectedMethod`, `expectedPath` and `expectedParams` elements in the test fixtures, we check all the inputs before we provide the sample outputs.

meganschanz:
That makes a lot of sense; such a great improvement :).

demiankatz:
Thanks, @meganschanz. It looks like there is now a conflict between this and the dev branch. You might want to merge the latest dev and sort that out to be sure you're running the latest version of the tests.

Also, are you all set up to run tests locally? I'm not sure how much work you've done with VuFind's test suite before. If I can offer any tips or help, just let me know!

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1248,10 +1248,11 @@ public function placeHold($holdDetails)
     {
         $default_request = $this->config['Holds']['default_request'] ?? 'Hold';
         try {
-            $requiredBy = $this->dateConverter->convertFromDisplayDate(
-                'Y-m-d',
-                $holdDetails['requiredBy']
-            );
+            $requiredBy = isset($holdDetails['requiredBy']) ?
```
EreMaijala: I'd suggest taking advantage of the recently added requiredByTS field that gets automatically added to holdDetails. It'll allow you to simplify the whole try-catch to this: 
`$requiredBy = !empty($holdDetails['requiredByTS']) ? gmdate('Y-m-d', $holdDetails['requiredByTS']) : null;`

EreMaijala (changes requested):

module/VuFind/src/VuFind/ILS/Driver/Folio.php
EreMaijala: Fixed the example above to use gmdate since the the timestamp is always UTC.

meganschanz:
I'm set up to run the unit tests locally. I typically run the `phpunitfaster` (I like dealing with just 1 error at a time) and try to remember to do the `qa-php` to avoid style failures in the CI after I commit. But today's issues were because I'm running PHP 8.1, so I also made a clone to test in a PHP 7.4 environment to reproduce the error in the CI.

But we're nearly there now! The part the tests are stuck on is when an invalid type is passed for `requiredByTS`, in PHP 7.4 an `ILSException` is thrown with the message `hold_date_invalid`, but in PHP 8, it throws a `TypeError` with the message containing `must of be type ?int`. So I'm trying to determine if there is a more generic way test catching either case or if there is a better way to write this test instead of checking for the exception (or possibly remove it if the UI prevents invalid data in the `requiredByTS` field).

demiankatz:
@meganschanz, glad to hear your test setup is working out. And hopefully we won't be supporting PHP 7.4 much longer, which will make life easier (I'm expecting to discuss that in more detail on tomorrow's Community Call).

Regarding the exception situation, I think the solution may be to do better type checking up front to avoid passing an invalid value to the method. Maybe rather than doing a `!empty` check, we should do an `!empty && is_int` check. Or maybe something like this would be more robust:

```
            $requiredByTS = intval($holdDetails['requiredByTS'] ?? 0);
            $requiredBy = !empty($requiredByTS) ? gmdate('Y-m-d', $requiredByTS) : null;
```

(I imagine some weird values of requiredByTS might result in intval returning 1 instead of 0... so maybe checking for the timestamp being greater than 1 would be better than checking for a non-empty value... but not sure how likely that is).

In any case, details aside, my thought is that if we ensure correct types up front (which is best practice anyway), then it should work the same in PHP 7 and PHP 8 and make the rest of the work easier.

EreMaijala:
requiredByTS should always be an int. If you have a case where it isn't, let's fix it!

meganschanz:
My hesitation with having it either default to `0` if it can't do an `intval` conversion of if we do an `is_int` check on the initial `if` statement is both of those result in values for the `requiredBy` field (`null` or `0`) instead of throwing an error about an invalid hold date, leading to a hold being potentially placed successfully (Folio is ok with a `null` `requiredBy` value but I'm not sure about giving it `1970-01-01`). I like the idea of doing a check prior to the conversion, but maybe we'd still want to keep the throwing of the exception. 

```
if (!empty($holdDetails['requiredByTS']) && !is_int($holdDetails['requiredByTS'])) {
    $this->throwAsIlsException($e, 'hold_date_invalid');
}
$requiredBy = !empty($holdDetails['requiredByTS']) ? gmdate('Y-m-d', $holdDetails['requiredByTS']) : null;
```

But as @EreMaijala mentioned, it should always be an `int. And scanning the code, I can't see anywhere in the code where `requiredByTS` wouldn't be an `int` so maybe this test isn't even necessary at the moment.

EreMaijala:
@meganschanz Yes, the `!empty()` should always be enough. That'll ensure that the value is present and isn't used if zero. Incidentally (or not!) I've done somewhat similar work with other drivers in PR #2740.

demiankatz:
Okay, that makes sense to me -- if the value is guaranteed to be an int, then it would make sense to me for any test passing in an invalid non-int value to expect some kind of exception. If you don't think the test is necessary at all, feel free to remove it. If you'd like to cover that base just for completeness, you could set up the test with PHP 8-based expectations and just add code to skip the test when running in PHP 7. Here's an example of how to do that: https://github.com/vufind-org/vufind/blob/64f045163d689221e4b9145a8944b57e6b857c47/module/VuFind/tests/unit-tests/src/VuFindTest/Session/MemcacheTest.php#L53-L57

demiankatz:
I've just merged #2740, so it might make sense to incorporate some of that recent work into this PR too, if @meganschanz thinks it's useful/appropriate.

meganschanz:
I merged in the `dev` branch again to make sure none of my changes were impacted by PR #2740, and there were no issues. Luckily the newly refactored method of checking for the datatype first meant that we don't have to skip the test for PHP 7.4 (because it wasn't re-raising an Exception of a different type). 

Even though there is no current code scenario when a non-int would be passed to `placeHold` I figured a test for it wouldn't hurt so we can make sure that never gets introduced to the code.

demiankatz (approved):
This looks great, @meganschanz. Thanks for all the work on it! I made one tiny style tweak (moving the `?` operator to the front of line 1379 and reducing indentation for consistency with style elsewhere in the project) rather than bothering you with another trivial review. :-)

demiankatz:
(@EreMaijala still has an outstanding review on this one, so I've requested his approval in order to be able to merge this).

EreMaijala (approved):
Looks good to me!

demiankatz:

PR #2740, opened 2023-03-06
Add support for optional "required by" date for holds.

EreMaijala:
"requiredByDateOptional" can be used when the user needs to be able to choose a last effective date for a hold, but is not required to do so. Additionally "defaultRequiredDate" can now be set to no value or empty string for no default. Support for unspecified "required by" date is included for Demo, KohaRest and SierraRest drivers.

Note that Sierra likely supports fetching `notNeededAfterDate` also before API v6, but I've been unable to verify this, so decided to play safe.


demiankatz (changes requested):
Thanks, @EreMaijala, this looks good to me, but a couple of questions/suggestions before we merge:

1.) Should we add a test for placing an optional hold, so we can differentiate/confirm the behavior in the Mink tests?

2.) Should we create a wiki page to document all the extra hold fields and their meanings? I think we now have enough that it's getting confusing, and it might be helpful to improve documentation. Maybe each ILS driver .ini file can list the supported values and include a link to a wiki page elaborating on the purpose of those values... and then the wiki page can include a note explaining that not every driver supports every option. I'm happy to work on that as a follow-up to this PR, but thought it worth discussing...

themes/bootstrap3/templates/record/hold.phtml
```diff
@@ -31,7 +31,7 @@
     </div>
   <?php endif; ?>
 
-  <?php if (in_array("requiredByDate", $this->extraHoldFields)): ?>
+  <?php if (in_array("requiredByDate", $this->extraHoldFields) || in_array("requiredByDateOptional", $this->extraHoldFields)): ?>
```
demiankatz, line 34: Do we need to display a different label, or some other visual indicator, to differentiate between optional and non-optional versions of the behavior?

EreMaijala: Added!

languages/sv.ini
```diff
@@ -583,6 +583,7 @@ hold_record_already_on_loan = "Du har detta material redan utlånad"
 hold_request_group = "Reservera från"
 hold_requested_group = "Reserverad från"
 hold_required_by = "Sista giltighetsdagen"
+hold_required_by = "Sista giltighetsdagen (valfri)"
```
demiankatz: Key is missing _optional here.

demiankatz (changes requested):
Thanks for adding the label -- just one minor problem!

languages/sv.ini
EreMaijala: Oops, fixed.

demiankatz (approved):
Thanks, @EreMaijala, this looks good to me!

PR #2741, opened 2023-03-06
Improve button styling in critical fix page.

demiankatz:
Related to recent discussion on #2699, here's some added button styling for another link that needs it.


demiankatz:
@crhallberg, I just turned off text decoration on buttons by default and that seems to allow this to look correct without adding the extra btn-link class. I looked at it in all three themes and it looks fine everywhere (though for some reason there's no mouseover reaction in bootprint3, even though bootstrap3/sandal cause a negative highlighting effect... but I don't think it's a big deal). Does this look reasonable to you, or do you think there might be unwanted side effects somewhere?

themes/bootstrap3/less/bootstrap.less
```diff
@@ -58,6 +58,9 @@ label { cursor: pointer; }
 .icon-bar { background-color: #888; }
 
 // Icons vs. Underline fix
+.btn {
+  text-decoration: none;
+}
 .btn-link {
   &:extend(.icon-link);
```
crhallberg, line 65: As long as `.btn-link` still has the underline, this is exactly the change I was hoping for!

crhallberg (approved):

themes/bootstrap3/less/bootstrap.less
demiankatz: It does indeed, so I think we're good to go. Thanks for the review!

PR #2742, opened 2023-03-07
Fix Feedback admin pagination.

demiankatz:
I just discovered that if you have more than 20 feedback items, you can't page through them. This PR should fix the problem.


xmorave2 (approved):
Thanks @demiankatz it looks good to me

PR #2743, opened 2023-03-07
Fix redirect after updating feedback status.

demiankatz:
This fixes another small bug I found while testing the Feedback admin -- after changing a status, invalid filters get applied to the URL. This is caused by passing null values as part of the query option when building the URL; it can be fixed with a simple array_filter.


xmorave2 (approved):
Thanks for the fix @demiankatz. It works like a charm.

PR #2745, opened 2023-03-07
Modernize code styles to use PSR-12 instead of PEAR

demiankatz:
This PR modernizes VuFind's code style tools to use PSR-12 instead of PEAR. It also changes line length warnings from 80 to 120 characters.

TODO
- [x] Update PHP_CodeSniffer configuration (add back PEAR-related commenting rules)
- [x] Update php-cs-fixer core code configuration
- [ ] Update php-cs-fixer template configuration
- [x] Manually fix any outstanding problems
- [x] Update wiki documentation to reflect new standards
- [x] Apply equivalent styles to VuFind sub-projects (code, http, etc.) after merging


demiankatz:
@EreMaijala, I think it would be best to merge this work now. Everything is done except for revising the php-cs-fixer template configuration, and I think it would be best to handle that separately to limit the number of files changed in a single commit. Once this is merged, I'll begin work on that (as well as updating the subprojects, since they don't have templates in them). I'll also update the wiki now, since this is clearly the direction we're moving toward.

demiankatz:
Okay, the [wiki page](https://vufind.org/wiki/development:coding_standards) is now updated.

EreMaijala (approved):

PR #2749, opened 2023-03-08
PSR-12: Standardize order/spacing of declare statements.

demiankatz:
This is more work to pave the way for #2745.


demiankatz:
This build was failing because php-cs-fixer's no_blank_lines_before_namespace rule conflicts with PSR-12. I have removed the rule here in addition to modifying the order of declare statements, in further preparation for the other changes in #2745.

EreMaijala (approved):

PR #2751, opened 2023-03-08
Remove form-attr-polyfill.js

demiankatz:
This is progress on cleaning up the js/lib directory as discussed on [VUFIND-1566](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1566).

I believe that this polyfill is no longer needed now that Internet Explorer is long past end of life, but please correct me if I am mistaken!

TODO
- [x] Run full test suite

PR #2755, opened 2023-03-08
Upgrade autocomplete.js and move it from lib/ to vendor/

demiankatz:
In combination with #2751, this resolves  [VUFIND-1566](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1566) by eliminating the need for a js/lib directory.

I'm starting the ball rolling by simply moving the file to a more appropriate directory, but I'm hoping that @crhallberg can help with some of the below items.

TODO
- [x] Should we update to a newer version of autocomplete.js? (The [separate project](https://github.com/vufind-org/autocomplete.js) seems to be well ahead of the version used here -- is it still compatible?)
- [x] Can we load this via package.json and copy it into place via build.xml like we have started doing for some other JS dependencies?
- [x] Release new version of upstream package to reflect changes required here; update package.json to match.


demiankatz:
Thanks for the progress on this, @crhallberg! Looks like we need to run lessToSass to get the build passing. I'll do some more thorough testing tomorrow. Please feel free to approve this PR when you feel this is ready to go; I won't actually merge it until I've approved your changes on top of my initial work. :-)

demiankatz:
@crhallberg, I went ahead and ran LessToSass to get the build to pass. I also ran the full test suite, which is passing. However, when I tried hands-on testing, this does not appear to be working correctly. Some problems I noticed:

  * If I search for something that does not exist in the index, the "Loading..." option never disappears
  * Sometimes, if I click on an autocomplete result, it submits the form with the text I was typing, instead of changing the form contents to the text I clicked on
  * I'm not sure how/why but at least once I clicked on something and ended up performing a search for the word "undefined".

Bottom line: it seems like it may be a good idea to develop some autocomplete tests on the dev branch so we can then verify the functionality post-upgrade.

demiankatz:
Also, if you think it's going to be complex/time-consuming to fix these problems, perhaps we should just settle for moving the existing file in the 9.0 release and then move the remaining work to 9.1 or 10.0 as appropriate. Let me know what you think!

crhallberg (approved):
Just needs the full test suite.

demiankatz:
Thanks, @crhallberg. I've just resolved the various merge conflicts that accumulated here due to recent activity. I believe this is now ready for @sturkel89 to test when time permits.

themes/bootstrap3/less/bootstrap.less
```diff
@@ -1,19 +1,20 @@
 /* #LESS> */
 @import "vendor/bootstrap/bootstrap";
 @import "vendor/font-awesome/font-awesome";
+@import "vendor/autocomplete";
 /* <#LESS */
 /* #SCSS>
 $fa-font-path: "../../bootstrap3/css/fonts";
 @import "vendor/bootstrap";
 @import "vendor/font-awesome/font-awesome";
+@import "vendor/autocomplete";
```
demiankatz: @crhallberg, since the font-awesome and autocomplete imports appear identical for LESS and SCSS, is there any reason to keep them inside these conditional blocks? It looks like the conditional part could end after the bootstrap import, since that works differently in each case.

themes/bootstrap3/scss/components/autocomplete.scss
```diff
@@ -1,4 +1,5 @@
-/* https://github.com/vufind-org/autocomplete.js 1.0b */
+/* https://github.com/vufind-org/autocomplete.js (v2.1.3) */
```
demiankatz, line 1: Should this file move to vendor for consistency with the LESS version? Otherwise, I don't think the imports are going to work right.

demiankatz:
@crhallberg, I spotted a couple of new minor issues related to LESS vs. SCSS (though these don't need to delay @sturkel89's testing). I've re-requested a review from you to flag the fact that work is needed, since I can't put a review on my own PR. :-)

sturkel89:
I’ve done some testing of this branch and found problems when auto-submit is turned on. I’ve sent the details directly to @crhallberg for review.

demiankatz:
@crhallberg, I took a moment to merge this with dev and get everything up to date for you. Hopefully that will save you a little time!

crhallberg:
I've tackled @sturkel89's feedback! Thanks again!

sturkel89:
@crhallberg Looks great! Almost everything is now perfect in autocomplete-update. The only remaining weirdness is this: 

When you type a few characters into the search box, **Loading…**. appears for a moment and then disappears; this is good behavior. As you add additional letters to the word (or string of nonsense letters), **Loading…** flashes by VERY quickly with each added letter -- for a tiny fraction of a second. It flashes by so quickly that it isn’t readable. This is probably not a huge problem, but I wanted to note it since the behavior isn't consistent during the person's interaction with the system. @demiankatz - what's your take on this?

demiankatz:
Thanks, @sturkel89! I'm seeing the same thing -- it seems like in the dev branch, there is some kind of threshold that prevents "Loading..." from displaying instantly, which prevents it from blipping by in cases where the data is returned very quickly... but maybe the updated code has a lower or different threshold, and so I'm seeing flashes of "Loading" more frequently. This is certainly not a major problem, but if there's an easy fix, it might be worth applying.

I also noticed another discrepancy -- the old code bolded the letters matching the typed keywords, but the new code does not. Is this an intentional change, or a bug?

New code:

image

Old code:

image

I won't merge anything until @crhallberg has a chance to offer feedback on these observations. In the meantime, I'll work on adding an automated autocomplete test, since that does seem to be missing from our current test suite.

demiankatz:
Okay, as promised, I've pushed up some basic Mink tests. I'm sure there's room to do more, but this at least covers the very basics.

crhallberg:
I've added the highlighting back in. The delay on Loading... is the same as the previous version (300ms), so I'm not sure why it's being more intrusive/distracting.

demiankatz:
@crhallberg, I think the difference between this branch and the dev branch is that keystrokes reset the timer -- if you keep typing, the "Loading..." text remains visible until you stop typing, and then it disappears after a moment. But I think in this branch, new keystrokes don't reset the timer (or something like that) so the loading message keeps blipping in and out. Does that make any sense?

demiankatz:
In any case, all tests are passing, but I'll wait to see what you think about my keystroke theory before taking further steps. :-)

crhallberg:
Okay, I see the difference: the Loading in the old version was showing before the search delay, and in the new version it is after. I'd make this change (563a242ca5922b12beb32b1db76e05fc80c1841f) to make the behavior the same if it's preferred.

demiankatz:
Thanks, @crhallberg, I think it's worth making the change, as it feels a little smoother that way.

crhallberg:
Then assuming all tests pass, I think this is ready to go!

demiankatz:
Thanks, @crhallberg -- this is looking good to me now, and I'm running tests to be on the safe side (though I don't expect any failures).

demiankatz:
Unfortunately, I'm getting intermittent failures on my AutocompleteTest, but I haven't determined yet whether this is due to a bug in the code or a bug in my test. Still investigating...

demiankatz:
Well, this is frustrating. I tried a couple of different environments, and while the tests never failed, I saw occasional retries in places I wouldn't have expected. I can ONLY do headless testing, because Chrome test mode no longer seems to work in any of my test environments, so I can't really see what's going wrong exactly. It always works right when I test by hand. I suspect this is actually okay to merge and the problems are somehow tied to my test setup, but I'm going to wait and see if @EreMaijala has time to try this in his environment before I merge it, just to be on the safe side.

EreMaijala (changes requested):
I can't see anyt retries with a couple of attempts. What I did notice, though, is that with bootprint3 the "Loading..." indicator color doesn't seem to fit in the theme anymore. I think a more neutral color like the old black-on-white box or bootprint3-specific tweak is needed.

demiankatz:
Thanks, @EreMaijala -- @crhallberg, can you look at the color issue today? I'd love to get this wrapped up before vacation if possible (though if not, life goes on... :-) ). A related question: I also notice that LOADING is now all-caps? I don't really have a preference one way or the other, but it's another obvious change from the old code that I noticed. Was there a reason for it?

themes/bootstrap3/less/vendor/autocomplete.less
```diff
@@ -1,10 +1,11 @@
 /* https://github.com/vufind-org/autocomplete.js (v2.1.3) */
 
-@autocomplete-item-bg   : #fff;
-@autocomplete-active-bg : @brand-primary;
-@autocomplete-hover-bg  : lighten(@brand-primary, 40%);
-@autocomplete-border    : lightgray;
-@autocomplete-secondary : darkgray; // item description
+@autocomplete-item-bg     : #fff;
```
demiankatz: This file does not appear to be coming from https://github.com/vufind-org/autocomplete.js, so I don't think it makes sense to include it in the vendor directory. Am I missing something?

demiankatz (commented):
@crhallberg, I noticed that you edited a vendor file in the latest commit, which led to the comment below -- if these files aren't coming from an external source, they should not be kept in vendor directories.

Also, I've found a new bug: I don't think the result cache is getting cleared when the user changes the search type. Steps to reproduce:

1.) Type a word in the search box -- you'll see the "all fields" suggestions.
2.) Now change the search type to "author"
3.) Type the same word -- you'll see the same results, which is incorrect.
4.) Now refresh the page
5.) Change the search type to "author" without typing anything first
6.) Type the same word from above -- you'll see different results from what you saw in step 3

demiankatz:
@crhallberg, I've just added a new Mink test that catches the latest bug I reported. I believe this problem also explains the intermittent failures I was seeing last week -- I think a timing issue in the browser was sometimes causing this caching problem to trigger.

themes/bootstrap3/less/vendor/autocomplete.less
crhallberg: This file is an adaptation of the css provided in the repository. Since it only applies to the autocomplete, I thought this was a good place for it. If you'd rather not have it hear (since it's customized and would be edited outside of updates) we can put it in `components/`.

demiankatz: Yes, please move it to components. Putting something in a vendor folder should be taken to mean "this file came from somewhere else and thus cannot be edited." Since these files were locally created, I think components is a more appropriate home for them.

demiankatz:
@crhallberg, thanks for moving autocomplete.less. Don't forget to move autocomplete.scss to match! After that, I think it's just one more bug fix (as described above) and this is finally done. :-)

demiankatz:
Thanks, @crhallberg! The full test suite is passing now, and this also looks good with regard to my manual testing. Since the latest fix required a change to the vendor file, we'll need to issue an upstream release before we can merge this. I've added a TODO checkbox for that final step. I'm also asking @EreMaijala to take one more look, since he has an outstanding review.

EreMaijala (changes requested):
Everything else looks good to me, but I just noticed that you get autocomplete suggestions for an empty search if you click on the field with mouse cursor. Not what I expected and not the way autocomplete used to behave.

demiankatz:
Good catch, @EreMaijala. I think there used to be a string length minimum before suggestions would appear. Can you look into restoring that, @crhallberg?

EreMaijala:
The length limit is there when you start typing, so I suspect clicking the field triggers something else.

demiankatz:
@EreMaijala, looks like @crhallberg's latest update has fixed the problem you identified. I'm running the full test suite now, but I think we're probably in good shape now.

demiankatz:
Update: all Mink tests are passing, and I can't break it with manual testing either. :-) I'll merge if @EreMaijala gives final approval.

EreMaijala (approved):

PR #2757, opened 2023-03-08
PSR-12: One class per file in tests

demiankatz:
This PR eliminates multiple class definitions per-file in tests, in order to comply with PSR-12 (and in most cases, for cleaner/simpler code).

TODO
- [x] Finish refactoring tests (MultiBackendTest still needs to be done).


demiankatz:
@EreMaijala, this just refactors tests so we don't have extra classes inside test files. The refactoring of MultiBackendTest is admittedly pretty ugly/clumsy, but I spent more than an hour trying to do it in a smart way and ran out of time, so this was the best I could manage.

I'm pretty sure we can totally replace all these extra classes and traits with mocking, but I couldn't get it to work. If you're interested in seeing my work-in-progress on a smarter refactor, you can find it here: https://github.com/vufind-org/vufind/commit/5b66857372146ec7282c453e41e211570be40679 -- but it doesn't work! I think I need to tear it all down and try again from a slightly different angle, but I unfortunately don't have the time. (It's probably broken due to some silly error on my part, but identifying the source of the error is a lot of work since the error messages aren't very helpful).

EreMaijala (approved):
I actually find the ILSMockTrait & co. clearer and easier to understand than trying to mock a pretty extensive driver properly, so I wouldn't necessarily want to change this approach even if mocking worked!

PR #2761, opened 2023-03-09
PSR-12: refactor Aleph driver to one-class-per-file.

demiankatz:
The PSR-12 standard requires only one class per file; this PR refactors the Aleph driver to separate out its support classes.

I'd be interested in hearing from @xrosecky and @welblaud as well if they have any concerns.


xmorave2 (approved):
I do think it's the step in the right direction.

With the exception we could go even further in the future and make it more general for API based drivers and use it in more drivers. But it is out of the scope of this PR.

demiankatz:
Thanks @xmorave2 (and @welblaud). I'll merge this now, and if you want to follow up with improvements to the exception logic, please feel free.

PR #2762, opened 2023-03-09
Improved semantic markup for navigation

jpkanter:
As part of an accessibility checkup it was noted that one of our VF6 instances could be easier for screenreaders if sidebars became `<aside>` elements so they can be quickly accessed with those programs.

My own quick research and testing yielded no ill effects when swapping `<div>` for `<aside>` as long as no `div.` specific css rules were created

This is a quick and dirty pull request, i went with a simple grep through the templates and looked for elements that qualified as "sidebar-esque".

This pull request is only but half of what our check up actually noted, further down they also mentioned that an `aria-label` for each sidebar would be most appropriate. I am not familiar with the newest code version of VuFind so i refrained from proposing `aria-label`s, having those is probably desirable.


demiankatz:
Thanks, @jpkanter! @crhallberg, could you take a closer look at this?

demiankatz:
I imagine @ckaz might also have thoughts on this.

jpkanter:
Well @ckaz is in the same team as I, so his thoughts were "go ahead and do that"

demiankatz:
Ha, in that case I'll withdraw that review request and wait for @crhallberg. :-)

crhallberg:
I've thought about this and what gives me pause is putting search facet controls and the MyResearch navigation into an `<aside>` element. I don't think that the [screen reader announcement of "Complementary"](https://russmaxdesign.github.io/accessible-forms/landmarks-aside01.html) is beneficial for those uses - but is great for related works, etc.

For the search facets, I think [role="search"](https://russmaxdesign.github.io/accessible-forms/landmarks-search.html) might be better. For the MyResearch navigation, [<nav>](https://russmaxdesign.github.io/accessible-forms/landmarks-nav01.html) would be better.

Either way, adding an [aria-label](https://russmaxdesign.github.io/accessible-forms/landmarks-aside04.html) or an [aria-labelledby](https://russmaxdesign.github.io/accessible-forms/landmarks-aside03.html) (where headers exist) would improve use in all cases.

crhallberg (changes requested):
I've thought about this and what gives me pause is putting search facet controls and the MyResearch navigation into an <aside> element. I don't think that the [screen reader announcement of "Complementary"](https://russmaxdesign.github.io/accessible-forms/landmarks-aside01.html) is beneficial for those uses - but is great for related works, etc.

For the search facets, I think [role="search"](https://russmaxdesign.github.io/accessible-forms/landmarks-search.html) might be better. For the MyResearch navigation, [`<nav>`](https://russmaxdesign.github.io/accessible-forms/landmarks-nav01.html) would be better.

Either way, adding an [aria-label](https://russmaxdesign.github.io/accessible-forms/landmarks-aside04.html) or an [aria-labelledby](https://russmaxdesign.github.io/accessible-forms/landmarks-aside03.html) (where headers exist) would improve use in all cases.

demiankatz:
@crhallberg, thanks for the feedback. I just edited your comment slightly because the `<nav>` tag needed to be escaped in order to be readable. :-)

ckaz:
> I've thought about this and what gives me pause is putting search facet controls and the MyResearch navigation into an
> element. I don't think that the [screen reader announcement of "Complementary"](https://russmaxdesign.github.io/accessible-forms/landmarks-aside01.html) is beneficial for those uses - but is great for related works, etc.
> 
> For the search facets, I think [role="search"](https://russmaxdesign.github.io/accessible-forms/landmarks-search.html) might be better. For the MyResearch navigation, [`<nav>`](https://russmaxdesign.github.io/accessible-forms/landmarks-nav01.html) would be better.
> 
> Either way, adding an [aria-label](https://russmaxdesign.github.io/accessible-forms/landmarks-aside04.html) or an [aria-labelledby](https://russmaxdesign.github.io/accessible-forms/landmarks-aside03.html) (where headers exist) would improve use in all cases.

Sorry for the delay, yes, we are on the same team. 

There was a slight misunderstanding as the accessibility report has its cryptic moments. 
I believe, the accessibility testers asked us to use 
`<section id="search-sidebar" aria-label="">`
or 
`<section id="search-sidebar" aria-labeledby="">`
or
` <section id="search-sidebar" title="">`
rather than a simple `<div id="search-sidebar" ...>` that gives no explanation. The view they tested was the result list even though a labelled `<section>` would make sense to me for all sidebars except recommendations, where my gut feeling is that `<aside>` could make more sense there.

crhallberg:
> The view they tested was the result list even though a labelled `<section>` would make sense to me for all sidebars except recommendations, where my gut feeling is that `<aside>` could make more sense there.

This works for me! We are on the same page, indeed. What are your thoughts on the best way to use aria-label/labelledby?

ckaz:
> What are your thoughts on the best way to use aria-label/labelledby?

I agree with you this might be the most helpful aspect of the entire change. If we have an ID which we could use for labeledby, I'd prefer this because the "labeledby" will change along any possible changes made to the ID the labelledby refers to.

As search facets go, I also fully agree that `<nav role="search">` would be best. In My Account, I'd also prefer `<nav>`  while
`<aside>` seems more suited to the related items carousel and such. `<section>` would catch them all while being less specific, which may be a down point.

ckaz:
Sorry, I just checked my own comments and they weren't clear enough, I' afraid. I should have looked at the files in more detail: 

In 

- record/view.phtml, I'd add 

`<section class="<?= $this->layoutClass('sidebar') ?>" ...`

RecordDriver/DefaultRecord/toolbar.phtml already has a `<nav>` wrapper which I'd add the aria-label to:

`<nav class="record-nav hidden-print" aria-label="<?= $this->transEscAttr('Toolbar') ?>`

Then I'd wrap the entire  Related/Similar.phtml in an  `<aside>`  .

In search/results.phtml, I'd either use <nav> like so: `<nav class="<?=$this->layoutClass('sidebar')?>" id="myresearch-sidebar">` without `role="search"` as this might be misleading (just read up on it again). OR I'd add `role="navigation"` to the existing `<div>` (Pls. note that the lists of facet values themselves should be structured as lists `<ul>` -- should we make another PR just for the lists?)

I **wouldn't use aria-label or aria-labeledby in either context** as the containers have headings right inside the container 

In collection/view.phtml, I'd also use `<nav>` or add `role="navigation"` 

This would leave My Account Menu: Here we could add a surrounding container `<section>` in files such as myresearch/profile.phtml, myresearch/editlist.phtml ...  or, my current preference, wrap the menus in myresearch/menu.phtml in  `<nav>`  containers 

Hope this is clearer! My best, Claas

crhallberg:
Sounds good! Thank you!

jpkanter:
Quick Update: I am still intending to do the desired changes but it got a bit more complex now so I wanted it to do properly so I have to find time in my schedule.

demiankatz:
No problem, @jpkanter, as long as you don't mind this waiting until release 9.1 or 10.0, since the 9.0 window is closing rapidly. :-)

crhallberg (approved):
Looks good to me after the merge conflict is resolved! Probably the void element merge from this week.

EreMaijala:
I discussed this with a local usability expert, and our conclusion was that using `<aside>` for non-standalone content is problematic. This is from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside):

> The &lt;aside&gt; [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents a portion of a document whose content is only indirectly related to the document's main content.

At least side facets and similar records are directly related to the main content, so they could be something like `<div role="region" aria-label="Something">`.

jpkanter:
And that is why this ticket changed a bit, the actual `<aside>` is only used for recommendations now, originally i had changed the navigation menus as well, but those now became `<nav>` i thought that that is most fitting for the definition mentioned above. The question really is..what defines related and how related are recommendations really to the main content as they are *not* the main content.

EreMaijala:
@jpkanter Granted, for related records it's not so clear cut. However, if you take a list of related records and display it standalone, it becomes just a list of records with no relevance to anything obvious, which to me implies that it's not a useful set of information as a standalone element. The more problematic cases are the search sidebar (last changed file) which contains facets, and the mylist sidebar that contains tags for favorites. They're related to the main content and useless outside of the context.

jpkanter:
@EreMaijala I see. So what are *we* going to do now? I can of course make the necessary changes or we bury this whole PR as its benefit is to minor.

EreMaijala:
@jpkanter As far as I can see the nav changes make sense, so let's not bury them. Also since the sidebars were reported as being problematic as they are, something should be done to them. My suggestion would be to try the forementioned `<div role="region" aria-label="Something">`, but that needs some further validation. I'd also welcome feedback from others.

themes/bootstrap3/templates/search/history.phtml
```diff
@@ -55,12 +55,7 @@
 
 <?php if ($saveSupported): ?>
   <div class="<?=$this->layoutClass('sidebar')?>" id="myresearch-sidebar">
-    <?=
-      $this->context($this)->renderInContext(
-          "myresearch/menu.phtml",
-          ['active' => 'history']
-      );
-    ?>
+    <?=$this->context($this)->renderInContext("myresearch/menu.phtml", ['active' => 'history'])?>
```
demiankatz, line 58: I see that in a couple of places, the code has been collapsed like this. I don't object to this change since it makes things more concise while (in my opinion, at least) remaining reasonably readable, but I'm curious if it has anything to do with other changes, or whether it somehow happened unintentionally.

themes/bootstrap3/templates/record/view.phtml
```diff
@@ -81,10 +81,10 @@
     <?=$this->driver->supportsCoinsOpenURL() ? '<span class="Z3988" title="' . $this->escapeHtmlAttr($this->driver->getCoinsOpenURL()) . '"></span>' : ''?>
   </div>
 
-  <div class="<?=$this->layoutClass('sidebar')?>">
+  <nav class="<?=$this->layoutClass('sidebar')?>">
```
demiankatz: Related record content isn't necessarily navigation related...

themes/bootstrap3/templates/primo/advanced.phtml
```diff
@@ -77,7 +77,7 @@
     <input type="submit" class="btn btn-primary" name="submit" value="<?=$this->transEscAttr("Find")?>">
   </div>
 
-  <div class="<?=$this->layoutClass('sidebar')?>">
+  <nav class="<?=$this->layoutClass('sidebar')?>">
```
demiankatz: This also feels like something other than navigation -- it covers filters, etc.

themes/bootstrap3/templates/search/advanced/layout.phtml
```diff
@@ -166,7 +166,7 @@
       <?php endif; ?>
     </div>
 
-    <div class="<?=$this->layoutClass('sidebar')?>">
+    <nav class="<?=$this->layoutClass('sidebar')?>">
```
demiankatz: As noted in the other advanced search context, I don't think this is navigation-related.

demiankatz (changes requested):
I agree with @EreMaijala that there's some clear improvement here that we don't want to lose. Maybe we just need to scale this back a little bit (e.g. revert aside-related changes) so we can merge the uncontroversial parts, and open one or more follow-up PRs to discuss the bits that are less clear-cut. Also, see below for a few more detailed comments.

themes/bootstrap3/templates/search/history.phtml
jpkanter: This is a totally valid question. Normally I tend to break up long lines, especially parameter-fests like those types, but it seems unnecessary in this specific instance. But I actually don't posses any memory of having done this, its a bit weird, for a pull request i don't want to change things without explicit reason. I take it its fine as it is?

jpkanter:
I looked more in-depth into this and changed some things around

* `/RecordTab/collectionlist.phtml`- changed back to div, i would like to make it `role="region"` but i dont know what it is exactly so i rather leave it alone
* `/Related/Similar.phtml` there was a lone `aside` there that would be doubled by the element above it that calls this one
* `/myresearch/mylist.phtml` - another double `aside` but i did change the `<div>` above to `role="region")`, i hope the translation will always fit, but as its part of the my account navigation that should fit(?)
* `/primo/advanced.phtml` - tbf, i don't know what primo is, i just changed this back
* `/record/view.phtml` - changes according to suggestions
* `/search/advanced/layout.phtml` - i tested this locally and the translation made the most sense, it might even be a true `<aside>` as the search tips could be context-free but are still in some context
*  `/search/reservessearch.phtml` - i don't know what "reserversearch" is, therefore i cannot test it 
* `/search/results.phtml` - now also a `role="region"` should be more appropriate

themes/bootstrap3/templates/myresearch/mylist.phtml
```diff
@@ -101,7 +101,7 @@
   <?php endif; ?>
 </div>
 
-<div class="<?=$this->layoutClass('sidebar')?>" id="myresearch-sidebar">
+<div class="<?=$this->layoutClass('sidebar')?>" id="myresearch-sidebar" role="navigation" aria-label="<?=$this->transEsc('Your Account')?>">
```
demiankatz: If you're going to do this here, I would think that you would also want to make matching changes to all of the templates that make a renderInContext call to import the `myresearch/menu.phtml` template. (See, for example, the schedulesearch.phtml template below where we talked about the unexpected/unexplained whitespace changes... and there are quite a few others too).

themes/bootstrap3/templates/search/advanced/layout.phtml
```diff
@@ -166,7 +166,7 @@
       <?php endif; ?>
     </div>
 
-    <div class="<?=$this->layoutClass('sidebar')?>">
+    <div class="<?=$this->layoutClass('sidebar')?>" role="region" aria-label="<?=$this->transEsc("Search Tips")?>">
```
demiankatz: Is "Search Tips" an appropriate label for this whole region? While there are search tips at the bottom, there's also potentially a set of filter-related controls which are functional. I'm concerned this label might cause people to totally skip over those controls.

themes/bootstrap3/templates/search/results.phtml
```diff
@@ -115,7 +115,7 @@
 <?php /* End Main Listing */ ?>
 
 <?php /* Refine Search Options */ ?>
-<div class="<?=$this->layoutClass('sidebar')?>" id="search-sidebar">
+<div class="<?=$this->layoutClass('sidebar')?>" id="search-sidebar" role="region" aria-label="<?=$this->transEsc($this->slot('side-facet-caption')->get('Refine Results')) ?>">
```
demiankatz: The side recommendation area is entirely configurable. It _usually_ includes the "refine results" facet controls, but in theory it could be configured to contain entirely different types of controls. In the scenario where the region contains "refine results" controls, it seems this label will be redundant. In the case where it contains something completely different, it seems that this label will be misleading. Do we really need a label for this at all? If each of the controls within the sidebar has an appropriate heading on it, won't that be sufficiently useful navigation without wrapping it in an extra layer? If the extra layer really is needed, maybe it needs a much broader label.

demiankatz (changes requested):
Thanks for the progress, @jpkanter! I've taken a closer look at the latest version and have a few more thoughts/suggestions.

There are a number of files here that I think are perfectly ready to get merged, so if at any point you'd like me to do a partial merge in order to reduce the number of files under consideration here, I'm happy to do this. But I also don't mind waiting until our conversation is complete -- let me know if you have a preference.

Also, to provide context for some of the templates you were unsure about in your latest comment:

`/RecordTab/collectionlist.phtml` - this is used by VuFind's "collection mode" where a record can be the parent of a set of other records. This template embeds a full set of search controls/results inside a tab of the record view, allowing the user to search within the contents of a collection. You can see it in action here, for example: https://digital.library.villanova.edu/Item/vudl:313596

`/primo/advanced.phtml` - Primo is a commercial discovery service, similar to EDS or Summon. This template provides the advanced search options for that particular service; it is parallel to /search/advanced.phtml in functionality but customized to account for the quirks/limitations of the Primo service.

`/search/reservessearch.phtml` - this is used for searching course reserves if they are stored in a local Solr index; see the [course reserves wiki page](https://vufind.org/wiki/configuration:course_reserves) for more details. If you want to test this, I can provide instructions on how to set up some fake data in a test environment to see the interface. If not, don't worry -- I'm not trying to make extra work for you. :-)

themes/bootstrap3/templates/search/results.phtml
jpkanter: `role="region"` demands an label according to the documents i have read. I already suspected that the content is to volatile for this kind of change, in that case, this cannot be done and its probably best removed.

themes/bootstrap3/templates/search/advanced/layout.phtml
jpkanter: I am hesitant to introduce new language tokens as I only know 5 languages and barely speak 2 of them, so translating seems like a steep cost for me, this one fit the best but your concerns are more than valid. If no proper language token can be found the `role` attribute cannot be used

themes/bootstrap3/templates/myresearch/mylist.phtml
jpkanter: For this I have to deeply apologize, I did those changes in a hurry, you are right. I shall set the whitespace/formatting right for all calls and add the proper changes

themes/bootstrap3/templates/search/advanced/layout.phtml
EreMaijala: Just a clarification that fortunately you don't have to manage all the languages yourself. :) It would be enough to introduce a new translation in English. You can of course include translations for other languages if you have them. I appreciate your thoughtfulness and agree that new translation strings should only be added when really needed, and I'm not implying that you can find a good label for this case.

themes/bootstrap3/templates/myresearch/mylist.phtml
demiankatz: No need to apologize -- I always appreciate contributions, and this is what the review process is for. :-)

themes/bootstrap3/templates/search/results.phtml
demiankatz: Yes, probably easiest to remove this for now. Maybe it's worth reviewing all the templates in the "Recommend" directory to make sure they all have appropriate headings. I think as long as all of those are well-labeled, the sidebar contents should be perfectly accessible.

themes/bootstrap3/templates/search/advanced/layout.phtml
demiankatz: Just to second @EreMaijala's comments -- we'll accept language strings in any language(s) you're willing and able to provide, and we have volunteers who help translate into all the other languages. I definitely don't expect everyone to speak every language, and of course I'd prefer that people only submit translation strings they are very confident in. The experts will help with the rest! :-)

jpkanter:
I hope i did nothing wrong with my git fu, on a first glance everything seems to be in order, i merged the current dev branch back into this PR.

I made the changes requested / reverted one instance, furthermore i *created* a new language token to describe the kind of menu more precise we find in advanced search, although, its not always filter AND tips so i am wavering here.

demiankatz:
Thanks, @jpkanter, this is looking better and better! I have made a couple of small changes:

1.) One file had its end-of-file line break removed but no other changes. I restored the line break to eliminate the file from this PR.

2.) I have renamed the PR to better reflect the scope of work (since it has evolved considerably since it was first opened).

I think it's nearly ready to merge, but I'm still not convinced that the newly-added language string and navigation heading is useful. On the advanced search screen, we use a sidebar for filters and search tips because it serves a visual purpose. But for a screen reader user, the fact that the filters and search tips already have h2 headings on them should make them appropriately accessible. Wrapping those existing headings in a higher-level heading that adds no extra information (and which could be misleading in situations where no filters are applied) seems like it would reduce rather than improve accessibility. Should we just leave this one as a div and not aria-label it at all, or is there really a reason that we MUST label it? I would argue that the role of this particular div is NOT really navigation; is there a better role we could assign to it?

(Interested in @crhallberg and @EreMaijala's updated thoughts on this as well).

demiankatz:
I should also note that the GitHub Actions build is failing, but I think this is due to technical difficulties at GitHub rather than an actual problem with this PR. I'll keep an eye on things and retry it later once it looks like the problem is fixed.

demiankatz:
@jpkanter, the GitHub Actions problem has resolved itself, and now the build is failing because of a real (but very minor) problem: the language files are not alphabetized correctly. You can run `php $VUFIND_HOME/public/index.php language normalize languages` and then commit the resulting changes to fix the problem... but since I'm proposing that we don't need the new language file at all, maybe that won't be necessary anyway. :-)

demiankatz (commented):
Thanks, @jpkanter, I think this is very nearly ready to go. I fixed a couple of small typos in the PR, so I believe everything is now technically correct. I also ran the full test suite, which is passing, so nothing is broken by these changes, and I did a visual inspection to confirm that there are no layout problems caused by the changes (the only visible change I could see was in the OAI-PMH menu, which shifted around by a pixel or so... but this is entirely harmless and no one will notice it).

However, before I approve and merge, I do have one last question regarding the myresearch-sidebar: we're setting an aria-label on the sidebar of "Your Account," but this region also includes an h3 element containing exactly the same text. Has anybody tested this with a real screen reader to be sure that it's actually helpful and isn't just causing duplicate navigation options? I'm happy to merge if this is confirmed to work better in real-world scenarios, I just thought this might be an issue and have no easy way of testing it myself.

I'm also interested in whether @EreMaijala and his team have any further thoughts.

In any case, thanks for all the work on this, and I think we are very nearly done one way or another! :-)

EreMaijala:
@demiankatz Looks good to me but I'll ask for another set of eyes to check this out as well.

EreMaijala (approved):
All eyes approved!

EreMaijala:
@demiankatz Approved, though I'm not sure about the "Your Account" aria-label.

demiankatz:
Thanks, @EreMaijala! I'll wait and see if @jpkanter has any more to share on the aria-label concerns before merging. I'd also be interested in input on this issue from @ckaz if he has time!

demiankatz:
Thanks, @EreMaijala! I'll wait and see if @jpkanter has any more to share on the aria-label concerns before merging. I'd also be interested in input on this issue from @ckaz if he has time!

jpkanter:
I cannot say much about the actual screenreader situation, I usually use orca to confirm that all things are in order but my knowledge about how to use screenreaders is limited so I cannot really say how jumping around works properly only that I *imagine* that it should be helpful regardless but wishful thinking is not facts.

ckaz:
> However, before I approve and merge, I do have one last question regarding the myresearch-sidebar: we're setting an aria-label on the sidebar of "Your Account," but this region also includes an h3 element containing exactly the same text. Has anybody tested this with a real screen reader to be sure that it's actually helpful and isn't just causing duplicate navigation options? I'm happy to merge if this is confirmed to work better in real-world scenarios, I just thought this might be an issue and have no easy way of testing it myself.

Dear all,

I agree with you, Demian, that 

`<div class="<?=$this->layoutClass('sidebar')?>" id="myresearch-sidebar" role="navigation" aria-label="<?=$this->transEsc('Your Account')?>">`
`   <h3><?=$this->transEsc('Your Account')?></h3>`

makes little sense, at first glance: The text inside `<h3>` seems to announce the purpose correctly and sufficiently. However, it doesn't.

When I tested this using NVDA+Chrome on Windows 10, NVDA would announce:

"Your Account  {short pause} Navigation {short pause} Landmark {longer pause}
Heading, Level 3 {short pause} Your Account"

So, the screen reader will read both, the role and the label, as expected.


When I tested

`<div class="<?=$this->layoutClass('sidebar')?>" id="myresearch-sidebar" role="navigation">`
`   <h3><?=$this->transEsc('Your Account')?></h3>`

NVDA would read:

"Navigation {short pause} Landmark {longer pause}
Heading, Level 3 {short pause} Your Account"

So, the screen reader will read the role but the _purpose_ of the section is lost and the heading only refers to the first part of the sidebar. So **the aria-label text is misleading**.

Therefore, my verdict would be this:
If _all_ sidebars follow the pattern of 
EITHER  
  nav + aria-label 
OR 
div + role="navigation" + aria-label

I'd **keep the aria-label on the My Account sidebar** but would consider **naming it more aptly**, to wrap up the multi-functionality of this navigation section in a good reference, such as

"My Account, Saved Lists, Loans, Fines and more"

This is just my crappy first shot at this, I'm sure @demiankatz will be able to come up with a better label.

Cheers!

themes/bootstrap3/templates/checkouts/history.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/holds/list.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/librarycards/home.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/checkedout.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/edit.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/editlist.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/fines.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/illrequests.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/mylist.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/newpassword.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/odmycontent.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/profile.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/schedulesearch.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/myresearch/storageretrievalrequests.phtml
ckaz, line 1: aria-label requires more apt descriptive content

themes/bootstrap3/templates/search/history.phtml
ckaz, line 1: aria-label requires more apt descriptive content

demiankatz:
@ckaz, coming up with descriptive aria-labels for VuFind's sidebars is a challenging problem, because the content of these sidebars is extremely dynamic. The pattern is often that a high-level template creates a sidebar area, and then a sub-template or helper populates it based on configuration and capabilities. At the level of code where the sidebar is created, we don't know what will be inside it -- and this is by design, because it allows VuFind to be very flexible.

In the case of the "Your Account" menu, depending on which ILS driver is used and which options are turned on, the contents of the menu could vary a great deal. We really can't assume anything specific -- the whole purpose of having a menu sub-template is to analyze the settings and capabilities to build a correct menu, and duplicating that logic just to create an aria-label seems excessive. But maybe we can come up with a more general aria-label that encompasses all of this, like "Account information and options" or "Contents of your account"?

In other types of sidebars, like in search results and record views, there's simply no way to know what will be in there, except that it's additional information somehow related to the search results or record being viewed. It's very hard to accurately label this. But as I've said before, I'm not sure that it's necessary to group those things together conceptually in the screen reader representation, because they don't necessarily have a grouped semantic meaning. If they're headed correctly and can be found in navigation, isn't that enough? If there really is a high value in this, we can come up with some kind of subsystem to build up a detailed label for the sidebars, but that's a potentially a very large and complex job, and it seems to me to offer minimal benefit at best.

ckaz:
> ... maybe we can come up with a more general aria-label that encompasses all of this, like "Account information and options" or "Contents of your account"?

That, in my opinion would be wholly sufficient. It's just that duplicating the first heading of the Account menu in the navigation landmark is misleading. Navigation elements should annouce themselves, even if only under more general terms.

> In other types of sidebars, like in search results and record views, there's simply no way to know what will be in there [...]. If they're headed correctly and can be found in navigation, isn't that enough?

I'm quite happy with the other sidebars, which I'm aware of, such as the one in search results: It is headed "Refine Results" and that is perfectly fine, I think.

demiankatz:
Thanks, @ckaz. I would propose then that @elsenhans just do a global search-and-replace to change `aria-label="<?=$this->transEsc('Your Account')?>"` to `aria-label="<?=$this->transEsc('account_menu_label')?>"` and then add an account_menu_label string to whichever language file(s) can be easily changed (I'm expecting en.ini and de.ini). Feel free to use my suggestions or pick something else -- we can discuss further as needed, but as long as it's centralized in a single new language string, it will be easy to adjust.

jpkanter:
I don't know what it is with the `similar.phtml` but for some reason the white space keeps reappearing despite it not being present on my side

Anyway, I did as was asked.

demiankatz (approved):
Thanks, @jpkanter, this looks good to me now!

I fixed the recurring whitespace issue again. I'm not sure why that would be happening. Is it possible that you have an editor setting that's stripping trailing whitespace off the end of files? Normally we try to include a line break at the end of the final line of every file, and the issue was caused by that final line break being deleted in that one template... but of course if it was an editor setting I would expect the same thing to happen to every file, not just that one! Maybe it's some kind of Git merging quirk. Anyway, not a big problem. :-)

PR #2763, opened 2023-03-09
SetupThemeResources: Ensure 'Content-Type' is never escaped

mtrojan-ub:
As mentioned in this week's community call:

According to Google Scholar, they can only respect the attribute encoding after the content type detection has been  finished successfully, so the Content Type declaration itself should not be encoded.


demiankatz:
Thanks, @mtrojan-ub! I'm curious if @EreMaijala has any thoughts on this. In particular, I wonder if there's any point to having the http-equiv meta tag at all if we're always including a real HTTP Content-Type header in our responses (which I'm pretty sure we are doing anyway).

EreMaijala (changes requested):
Demian is right, we already use a proper header. I think it would be cleanest to just remove this from page meta tags.

demiankatz:
Thanks, @EreMaijala, I went through the commit history just to confirm that this http-equiv stuff has been around since the very beginning (it was first committed as part of 2.0RC1, which means it was ported directly from 1.x) and so there's no recent history indicating that we need it for some specific reason. Unless @mtrojan-ub thinks that Google Scholar explicitly needs it to interpret data correctly, I think we're in pretty good shape to get rid of it.

mtrojan-ub:
OK, so i will ask them again just to be sure.

demiankatz:
@mtrojan-ub, since it's been a little over a month, I just wanted to check in: did you ever get a response to your inquiry?

mtrojan-ub:
They suggested that we should still keep it, however i'm not entirely sure why (maybe for compatibility reasons).

demiankatz:
Thanks for the update, @mtrojan-ub. In that case, I'd suggest this approach:

1.) Refactor the head meta code to a separate method (e.g. `addHttpEquivMeta($headMeta)`).

2.) Add a detailed comment in the function or its header to the effect of "Adding this meta tag is redundant, since we already include appropriate values in the real HTTP headers. However, Google Scholar support has requested that we retain this tag to help with their crawling process."

This way, we've documented in the code WHY this is here, and we've also created it in a separate method so that it can be easily overridden or disabled in scenarios where different behavior is desired.

@EreMaijala, would you agree?

EreMaijala:
@demiankatz I suppose the meta tag could still be useful if the page is saved without headers. Your suggestion makes sense to me.

demiankatz:
Thanks, @EreMaijala, so now we just need to see what @mtrojan-ub thinks once he has a moment free. :-)

mtrojan-ub:
Thanks for the feedback! I will take care of the change when time permits.

module/VuFindTheme/src/VuFindTheme/View/Helper/SetupThemeResources.php
```diff
@@ -74,20 +76,39 @@ public function __invoke()
      */
     protected function addMetaTags()
     {
-        // Set up encoding:
         $headMeta = $this->getView()->plugin('headMeta');
-        $headMeta()->prependHttpEquiv(
-            'Content-Type',
-            'text/html; charset=' . $this->container->getEncoding()
-        );
+        $headMeta->setAutoEscape(false);
```
demiankatz, line 86: Shouldn't this line move into the new addHttpEquivMeta? And shouldn't we restore the previous state of setAutoEscape after we're done?

demiankatz (changes requested):
Thanks for the progress, @mtrojan-ub -- just a couple of questions!

module/VuFindTheme/src/VuFindTheme/View/Helper/SetupThemeResources.php
demiankatz: (I see that you've bypassed the autoescaping by explicitly escaping the generator below... but I wonder if the previous behavior of simply switching the autoescaping off and back on again to permit the meta tag to be escaped differently is a little more backward-compatible and intuitive).

mtrojan-ub: After some more testing i noticed that `setAutoEscape` does not seem to behave the way i expected it to. In my first attempt, i thought it would be stateful so we could just disable it before adding the Content-Type header and re-enable it again afterwards, but it seems like in the end this will affect all settings no matter where it is called and only the last setting will matter when the whole page is generated. 
Seems like the inspector in the browser always shows the rendered value (after escaping has been removed again), and you can only see the difference if you actually look at the HTML source code.

demiankatz: Thanks, @mtrojan-ub, that does make sense (though it's a bit unfortunate/annoying). Under the circumstances, I agree that this may be the best that we can do! I have a few more comments/suggestions as a result:

1.) We should add a comment above the `setAutoEscape` call like `// Disable automatic escaping of meta tag content; we need some tags to skip escaping.

2.) We should make sure all other headMeta calls are explicitly escaped (I think there's only one that got missed, in shortlink/redirect.phtml.

3.) We'll have to be sure to note this in the changelog... and I even wonder if we should make this a 10.0 PR instead of a 9.1 PR since this is a potentially backward breaking change (though in a very small, subtle way).

... though the more I think about this, the less I like it. I feel that we may be potentially creating more problems than we are solving (though the severity of this depends on how likely it is that people are using headMeta in their local custom themes).

Some potential different approaches:

1.) We could create a "headMetaUnescaped" view helper to explicitly differentiate the behaviors. This could just be a subclass of headMeta that turns off escaping. Then we need to add an extra line to the layout template to render the contents of this container, but we can keep things separated without breaking anything unexpectedly.

2.) We could move the encoding meta tag entirely out of the headMeta helper (there are other hard-coded meta tags in our current layout template for some reason, so there's already precedent for this). Of course, we would then need a mechanism for accessing the encoding type from the template, which might not be possible at present.

3.) We could change the behavior of setupThemeResources so that it has additional public methods, and then we could put a `getMetaTags` method on it which actually renders the meta tags instead of merely setting up the meta tag container. This would give us more fine-grained control over the escaping, and the original behavior we had wanted would become possible. I think it could probably even be achieved in a backward-compatible way -- I could put together a proof of concept if you like the idea but aren't entirely following what I'm suggesting. :-)

It's possible that some of these ideas might be complementary (i.e. it would be possible to do both 1 and 3).

What do you think?

demiankatz:
@mtrojan-ub, I've merged the latest dev into this branch to keep things up to date. We should resume the conversation here when time permits!

demiankatz:
@mtrojan-ub, I noticed that some conflicts had appeared here. I just resolved them to keep things moving forward, though I think the next step here is to continue our conversation above. Please let me know your thoughts when time permits!

mtrojan-ub:
As discussed on the community call => since the contact with Google Scholar is no longer active, and this PR is already open for quite a while and has no real progress, i think we should close it.

PR #2767, opened 2023-03-09
PSR-12: adjust whitespace in headers.

demiankatz:
One of the most consequential, yet also insignificant, impacts of changing to PSR-12 is that it mandates extra blank lines in file headers -- one after the open PHP tag, and one before the namespace. This PR simply applies all of those new blank lines without any other changes, since this is compatible with our other existing style rules. By merging this separately, we can greatly reduce the size of #2745, and be less likely to miss any more problematic changes introduced by new auto-fixer rules.


demiankatz:
Hmm, for some reason the php-cs-fixer rules I chose to create this PR don't seem to have consistently added the top blank lines as expected. It wouldn't hurt to merge this as incremental progress, but I'm going to see if I can make it better first...

demiankatz:
I'm not sure why php-cs-fixer wasn't entirely cooperating, but I added the remaining missing whitespace using a regular expression.

EreMaijala (approved):
Looks good to me, and I actually checked that the change that wasn't just a new line (in GoogleTagManager.php) made sense.

demiankatz:
Thanks, @EreMaijala, and I know where that odd GoogleTagManager change came from: I have VSCode set to "remove trailing whitespace" and it must be that the php-cs-fixer rule for that retains trailing space inside heredoc blocks.

PR #2771, opened 2023-03-10
PSR-12: avoid case clauses without break

demiankatz:
Another PSR-12 rule that can be safely applied ahead of merging #2745 -- separating this out to make the remaining review of #2745 a little easier.


demiankatz:
@EreMaijala, I've reviewed the first two files and feel pretty confident that the missing breaks are intentional. I'm less sure about the Oauth2 file, which I think is one of yours. What do you think?

module/VuFind/src/VuFind/OAuth2/Entity/UserEntity.php
```diff
@@ -191,6 +191,7 @@ public function getClaims()
                             . $this->oauth2Config['Server']['hashSalt']
                         );
                     }
+                    // no break
```
EreMaijala: This one seems like a mistake, and should actually have a break.

module/VuFind/src/VuFind/AjaxHandler/GetResolverLinks.php
```diff
@@ -136,6 +136,7 @@ public function handleRequest(Params $params)
                     // Special case -- modify DOI text for special display:
                     $link['title'] = $this->translate('Get full text');
                     $link['coverage'] = '';
+                    // no break
```
EreMaijala: I'd prefer repeating the code from the `default` case and adding a break here, but either way works for me.

module/VuFind/src/VuFind/Net/UserIpReader.php
```diff
@@ -119,10 +119,12 @@ public function getUserIp()
                         if (!empty($parts)) {
```
EreMaijala: We could check for `empty($parts)` before the switch and just have return statements for `'first'` and `'last'`.

EreMaijala (changes requested):
I find case statements flowing from one to another generally confusing and error-prone, so I'd rather get rid of them, but if you disagree, the only real issue is the UserEntity case.

EreMaijala:
> @EreMaijala, I've reviewed the first two files and feel pretty confident that the missing breaks are intentional. I'm less sure about the Oauth2 file, which I think is one of yours. What do you think?

I believe it's a mistake, and should have a break. Functionally it doesn't change anything in any normal situation.

demiankatz:
Thanks, @EreMaijala, I agree that a little more work was helpful here. I've rewritten the first two files to avoid the need to drop through cases (and in the second file, to eliminate case entirely, as it made things much harder to understand). I also added the missing break. Two of the three files already had test coverage, so I'm confident the refactoring is okay; the third file was missing coverage, so I wrote a test and confirmed the same behavior before and after my changes.

EreMaijala (approved):

PR #2773, opened 2023-03-11
Fix ltrim paramaters order in record linker view helper

xmorave2:


EreMaijala (approved):
Drat! I think that was my mistake. Thanks for the fix!

demiankatz:
Thanks, @xmorave2. I'm merging this since it's an obvious bug fix, but the fact that this slipped through seems to suggest that we have a gap in test coverage somewhere! It would be good to address that at some point, though at the moment I'm a bit too focused on finishing up other open PRs to have the time....

xmorave2:
@demiankatz I tried to create few basic test for getActionUrl here: #2777

PR #2760, opened 2023-03-09
Fix HTML validation and JS deprecation issues.

EreMaijala:
Notes:
- role=menuitem has been removed from hierarchical facet entries. It was invalid and likely to be ignored anyway since it wasn't contained in a menu or menubar.
- The cookie consent module creates duplicate id's for table headers. This has been reported as https://github.com/orestbida/cookieconsent/issues/483.
- embedded_record.js is currently using collapse to manage switching between short and long view. I don't think collapse can manager aria attributes properly in this case since the heading toggles two elements, but I'm not sure what would be the best remedy either.

PR #2774, opened 2023-03-13
Fix hierarchical facet markup and improve accessibility.

EreMaijala:
- Adds a jsTree plugin to render the badge and exclude link so that they are outside of the tree item's anchor element and the markup resembles that of normal facets as accurately as possible.
- Uses jsTree's selected state to indicate that a facet item is applied so that it gets the aria-selected="true" attribute.


EreMaijala:
While checking HTML validation of complete pages (#2760) I noticed that the markup that the hierarchical facets contained was not valid since an &lt;a&gt; element cannot contain another one. This fixed the markup and also uses jsTree's state.selected to indicate that a facet item is applied for accessibility. Includes also a bit of cleanup for the styles previously needed to work around differences between normal and hierarchical facets and to reset styles from applied parent facet overflowing to its children.

demiankatz:
Thanks, @EreMaijala, the code looks good and all tests are passing. I did notice that the highlight color has changed, in the bootprint3 theme, at least...

Before:

image

After:

image

Is this known/intentional? I don't think the new version looks bad, but it is less consistent with how the other (non-hierarchical) fields highlight.

demiankatz (changes requested):
I just left a comment about a cosmetic change that looks like it might not be intentional. I'm going to hold off on further testing until you have a chance to respond to that.

One other thought: should we take an opportunity (either here, or perhaps more appropriately, in a separate PR) to adjust jsTree so that it gets loaded via npm like some other dependencies? Since this is such an important dependency for us, it might be good to manage it more explicitly now that we have a mechanism for doing so.

EreMaijala:
@demiankatz Oops, I didn't test properly with bootprint3. Will need to fix. @pasitiis is working on jsTree update. I think that would be a good place to move it to an npm dependency.

EreMaijala:
@demiankatz Should be better now.

demiankatz (approved):
Thanks, @EreMaijala, this looks much better now -- in fact, it seems to fix a bug with underlining styles that's currently visible in dev. I'm going to leave this open a little longer in case @crhallberg has anything to add before it gets merged.

demiankatz:
I've merged the latest dev branch into this PR so that we can take advantage of the recent jstree upgrade here. All tests are still passing.

themes/bootstrap3/templates/Recommend/SideFacetsDeferred.phtml
```diff
@@ -1,5 +1,6 @@
 <?php
   $this->headScript()->appendFile('vendor/jsTree/jstree.min.js');
+  $this->headScript()->appendFile('jstree-vufind-facet-plugin.js');
```
crhallberg, line 3: Is this a good candidate for the `js/lib/` folder?

crhallberg (approved):

themes/bootstrap3/templates/Recommend/SideFacetsDeferred.phtml
EreMaijala: @crhallberg Not sure. If js/lib is for libraries that are somewhat standalone, this is probably an edge case. It's pretty much useless without the custom driving code in facets.js, but it's sort of a standalone plugin.

demiankatz: I thought we were working to eliminate the js/lib folder... I don't mind having it if its purpose is clearly defined, but maybe we're better off with more descriptive names if we partition our JS code into directories.

demiankatz:
I'm going to go ahead and merge this as-is. I think there could be arguments made for putting the jstree plugin in a subdirectory (i.e. jstree/vufind-plugin.js instead of jstree-vufind-plugin.js, or use the lib folder as suggested), but I don't think that conversation needs to hold up the rest of the work. We can reorganize as a separate pass if there's a strong reason to (or we can just wait until the new theme overhauls everything).

Thanks again, everyone!

PR #2775, opened 2023-03-13
Fix checkbox underlining for normal OR facets.

EreMaijala:


themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
```diff
@@ -22,42 +23,38 @@
...
+<?php endif; ?>
 
-  <?php if (!$this->facet['isApplied'] && null !== $this->facet['count']): ?>
-    <span class="badge">
-      <?php $currentCount = $this->localizedNumber($this->facet['count']); ?>
-      <span class="sr-only"><?=$this->transEsc('result_count', ['%%count%%' => $currentCount]) ?></span>
-      <span aria-hidden="true"><?=$currentCount ?></span>
-    </span>
-  <?php endif; ?>
+<?php if (!$this->facet['isApplied'] && null !== $this->facet['count']): ?>
```
demiankatz: I think the extra indenting here was intentional, because the if..else above creates an open tag, and then this markup is indented inside that markup. I can think of some ways we could refactor to perhaps make this more clear, but I'm not sure if it's worth it. Just raising the issue in case it's worth discussing further.

demiankatz (commented):
Thanks, @EreMaijala, this looks good to me -- I appreciate the consolidation of redundant if/else logic, plus it looks visually better, and all tests are passing. Just one minor style question below.

I'm also asking for a review from @crhallberg in case he has any other insight -- I know this was on his to-do list somewhere, so if nothing else I want to alert him that it's now done. :-)

crhallberg (approved):

themes/bootstrap3/templates/Recommend/SideFacets/single-facet.phtml
EreMaijala: Right, I got confused about it. And now I couldn't resist refactoring it a bit. Better?

demiankatz (approved):

PR #2777, opened 2023-03-14
Add tests for RecordLinker getActionUrl

xmorave2:
Just few basic test as a start


demiankatz (approved):
Thanks, @xmorave2! I just added one more assertion to test the case where there is both a non-empty query AND an anchor value.

PR #2780, opened 2023-03-14
eslint: apply to js/lib/ and more

crhallberg:
Include more files in eslint by using ignorePatterns.

In the future, [eslint will use a flat configuration](https://eslint.org/docs/latest/use/configure/configuration-files-new), allowing us to specify all of this in eslint configuration, rather than between that and build.xml.

TODO
- [x] Finish #2755 before merging - we need to get autocomplete.js out of the lib directory before the build will pass here.


demiankatz (approved):
Thanks, @crhallberg! Now that #2755 is merged, this revealed a couple of very minor comment-related issues in files previously missed by eslint. I have fixed them and confirmed that things still work as expected. Merging this now and closing VUFIND-1566.

PR #2789, opened 2023-03-16
Clean up non-standard object construction.

demiankatz:
This PR cleans up some non-standard use of `new` -- adding more missing parentheses in one place, and eliminating an inexplicable unnecessary `new` in a test.


EreMaijala (approved):
I wasn't even aware of that you can use `new` like that without issues.

demiankatz:
Yes, @EreMaijala, I found that surprising too!

PR #2792, opened 2023-03-16
PSR-12: assorted style fixes.

demiankatz:
This is the final spin-off from #2745, containing all remaining changes that can be safely committed before we fully change style standards. Once this is merged, I think #2745 itself can be reviewed -- I'll share more notes there when the time comes.


demiankatz:
php-cs-fixer was failing here due to a PSR-12-incompatible rule in the configuration. I've removed that to get the build to pass -- one more detail of #2745 that's been refactored here to make the final commit smaller.

EreMaijala (approved):

PR #2794, opened 2023-03-17
PSR-12: update template configuration for php-cs-fixer.

demiankatz:
This PR adjusts the php-cs-fixer template configuration to use PSR-12 as much as possible, while overriding a couple of settings that don't play well with the template format. It greatly reduces the size of the configuration and better aligns it with the primary config.

I'm sure if we could find a way to make the 'braces' fixer work better with templates, we could significantly improve some of our code. Unfortunately, I still can't find a way to use it without severely messing up indentation.


tests/vufind_templates.php-cs-fixer.php
```diff
@@ -5,34 +5,26 @@
...
-    'line_ending' => true,
     'list_syntax' => ['syntax' => 'short'],
     'lowercase_cast' => true,
-    'lowercase_keywords' => true,
     'magic_constant_casing' => true,
-    'method_argument_space' => true,
     'native_function_casing' => true,
     'no_blank_lines_after_class_opening' => true,
-    'no_blank_lines_after_phpdoc' => true,
+    'no_closing_tag' => false,
```
EreMaijala: Do we really need this? I believe it's useful in ensuring that a file doesn't output unexpected line feeds etc.

tests/vufind_templates.php-cs-fixer.php
```diff
@@ -41,10 +33,8 @@
     'no_leading_namespace_whitespace' => true,
     'no_mixed_echo_print' => true,
     'no_singleline_whitespace_before_semicolons' => true,
-    'no_spaces_after_function_name' => true,
     'no_spaces_around_offset' => true,
-    'no_spaces_inside_parenthesis' => true,
-    'no_trailing_whitespace' => true,
+    'no_trailing_whitespace_in_comment' => false,
```
EreMaijala: Why do we need this?

EreMaijala (changes requested):

tests/vufind_templates.php-cs-fixer.php
demiankatz: In templates, we might want to output whitespace, and it makes things look unbalanced when, e.g. a template ends with `<?php endif`.

tests/vufind_templates.php-cs-fixer.php
demiankatz: If we have a comment inside a single line `<?php` tag, disabling this rule retains the space between the comment and the `?>`, which improves readability.

tests/vufind_templates.php-cs-fixer.php
EreMaijala: Understood, but the rule only applies to files that contain only PHP, so you'd need to have something like `echo "\n";` in place anyway. Otherwise you'd depend on something very unobvious.

demiankatz: Do you have time to change the rule and run the fixer? I don't think php-cs-fixer is smart enough to restrict the rule to PHP-only files, and I'm pretty sure it messed up a lot of files, but maybe I'm wrong. I can do it in a couple hours to double check, but I'm not in the office yet.

EreMaijala: Tested, and to me it looks like all the changes make sense. Removing the closing tags also allows adding of ending line feeds, which is a good thing.

demiankatz: You're right, I think I was blaming the `no_closing_tag` rule for problems that were actually being caused by the `braces` rule. Thanks for looking into it more carefully than I did! :-)

I've also put a comment on `no_trailing_whitespace_in_comment` to document the reason we're disabling that rule.

EreMaijala (approved):

PR #2795, opened 2023-03-17
php-cs-fixer: apply PHP 7.4 compatibility rules.

demiankatz:
While simplifying and consolidating php-cs-fixer rules, it seemed sensible to simplify the configuration around PHP version compatibility rules. This brings things up to the recommended PHP 7.4 settings, though I think we can safely raise this all the way to 8.2 without breaking backward compatibility. Since there are already a large number of changes (all related to trailing commas inside multi-line arrays), I thought it best to get this merged first and then advance the bar.

TODO
- [x] Run full test suite

### DIFF
--- a/module/VuFind/Module.php
+++ b/module/VuFind/Module.php
@@ -62,8 +62,8 @@ class Module
         return [
             'Laminas\Loader\ClassMapAutoloader' => [
                 'classes' => [
-                    'minSO' => __DIR__ . '/src/VuFind/Search/minSO.php'
-                ]
+                    'minSO' => __DIR__ . '/src/VuFind/Search/minSO.php',
+                ],
             ],
             'Laminas\Loader\StandardAutoloader' => [
                 'namespaces' => [
--- a/module/VuFind/config/module.config.php
+++ b/module/VuFind/config/module.config.php
@@ -43,7 +43,7 @@ $config = [
                     'defaults' => [
                         'controller' => 'Content',
                         'action'     => 'Content',
-                    ]
+                    ],
                 ],
             ],
             'shortlink' => [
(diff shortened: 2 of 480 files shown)


demiankatz:
I ran out of time before I could thoroughly review this, but I think it might be worth splitting off separate PRs that apply the assign_null_coalescing_to_coalesce_equal and heredoc_indentation rules so that all that's left here is changes related to trailing_comma_in_multiline. We can also consolidate work with #2794 after that PR is merged.

demiankatz:
@EreMaijala, this PR simplifies/consolidates our php-cs-fixer configuration a bit more. After the merge of #2796 and #2797, all changes here are related to the trailing_comma_in_multiline rule, which puts a trailing comma on the last entry in a multi-line array. It affects a lot of files, but I think it's a worthwhile improvement. Any concerns/objections? I've confirmed that all tests are still passing.

EreMaijala:
@demiankatz I think it makes sense. I'll check the changes a bit later.

demiankatz:
Thanks, @EreMaijala -- just to share what I've done so far, in case it influences your review strategy:

1.) I rebuilt this PR from the latest dev branch, by applying the modified php-cs-fixer rules here to clean code. (I didn't want to risk weird things happening by merging the previous version of the PR with hand-applied changes in dev, so it was safer to just do the processing over again from a clean copy).

2.) I saved the php-cs-fixer output to a file and ran it through `| grep -v \(trailing_comma_in_multiline\)` to be absolutely confident that ONLY the one rule was applied to any of the changed files.

3.) I reviewed all of the changed files under themes/ by hand.

4.) I reviewed the first ~150 changed files under module/ by hand (but I stopped when it became painful to do that in GitHub due to the extra clicking required). That leaves around 300 files that I confess I haven't studied directly.

I didn't see any changes that were anything other than added end-of-line commas in all of my reviewing.

EreMaijala (approved):
Everything looking good. I also checked the later changes from the raw diff.

demiankatz:
Thanks, @EreMaijala!

PR #2798, opened 2023-03-20
php-cs-fixer: consolidate and update PHPUnit rules.

demiankatz:
Another thread of php-cs-fixer cleanup: using the built-in ruleset for PHPUnit to reduce the number of separate rules configured to fix tests, and also bringing that ruleset up to date for the version of PHPUnit we are using. We'll want to raise this to support 10.0 as well in the near future, but I'll address that separately to keep the number of diffs per PR under control.

TODO
- [x] Run full test suite


demiankatz:
@EreMaijala, this is ready for review now -- all tests are passing. I had to make a couple changes on top of what php-cs-fixer did automatically: one to remove a completely unnecessary assertion that was causing problems, and one to use a newer assertion method name to get ahead of a PHPUnit 10 deprecation.

EreMaijala (approved):

PR #2800, opened 2023-03-21
Fix CoverController::getImageParams isbn array check

LuomaJuha:
I think this was forgotten in https://github.com/vufind-org/vufind/pull/2633 but the parameters do not work properly when trying to fetch an image to show.

### DIFF
--- a/module/VuFind/src/VuFind/Controller/CoverController.php
+++ b/module/VuFind/src/VuFind/Controller/CoverController.php
@@ -88,20 +88,14 @@ class CoverController extends \Laminas\Mvc\Controller\AbstractActionController
     protected function getImageParams()
     {
         $params = $this->params();  // shortcut for readability
-        $isbn = null;
+        // Legacy support for "isn", "isbn" param which has been superseded by isbns:
         foreach (['isbns', 'isbn', 'isn'] as $identification) {
-            $isbn = $params()->fromQuery($identification);
-            if (empty($isbn)) {
-                continue;
+            if ($isbns = $params()->fromQuery($identification)) {
+                break;
             }
-            if (is_array($isbn)) {
-                $isbn = reset($isbn);
-            }
-            break;
         }
         return [
-            // Legacy support for "isn" param which has been superseded by isbn:
-            'isbn' => $isbn,
+            'isbns' => (array)$isbns,
             'size' => $params()->fromQuery('size'),
             'type' => $params()->fromQuery('contenttype'),
             'title' => $params()->fromQuery('title'),
--- a/module/VuFind/src/VuFind/Controller/CoverController.php
+++ b/module/VuFind/src/VuFind/Controller/CoverController.php
@@ -88,14 +88,16 @@ class CoverController extends \Laminas\Mvc\Controller\AbstractActionController
     protected function getImageParams()
     {
         $params = $this->params();  // shortcut for readability
+        $isbns = null;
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/Controller/CoverController.php
```diff
@@ -88,9 +88,20 @@ public function __construct(
...
     {
         $params = $this->params();  // shortcut for readability
+        $isbn = null;
+        foreach (['isbns', 'isbn', 'isn'] as $identification) {
+            $isbn = $params()->fromQuery($identification);
+            if (empty($isbn)) {
+                continue;
+            }
+            if (is_array($isbn)) {
+                $isbn = reset($isbn);
```
demiankatz: Why is this reset necessary?

module/VuFind/src/VuFind/Controller/CoverController.php
```diff
@@ -88,9 +88,20 @@ public function __construct(
...
+            }
+            break;
+        }
         return [
             // Legacy support for "isn" param which has been superseded by isbn:
-            'isbn' => $params()->fromQuery('isbn') ?: $params()->fromQuery('isn'),
+            'isbn' => $isbn,
```
demiankatz: I think the VuFind\Cover\Loader is now expecting an 'isbns' key instead of singular 'isbn' in order to take advantage of multiple values. I wonder if `'isbns' => (array)$isbn` would be the best way to approach this.

demiankatz (changes requested):
Thanks, @LuomaJuha! I'm copying @xmorave2 into this thread in case he has any insights. I thought I had done end-to-end testing of this, but maybe I did something incorrectly or overlooked a detail. It certainly seems like you've found a gap in the code. I also have a couple of questions about implementation details (see below).

module/VuFind/src/VuFind/Controller/CoverController.php
xmorave2: Why do you reduce the ISBNs to first one only? We probably want to pass all available ISBNs to cover loader. 
So we want something like this instead: 

```
$isbns = $params()->fromQuery($identification);
if (empty($isbns)) {
    continue;
}
if (!is_array($isbns)) {
    $isbns = [$isbns];
}
```

module/VuFind/src/VuFind/Controller/CoverController.php
xmorave2: Yes we need `isbns` param instead of `isbn`.

xmorave2 (changes requested):
Thanks @LuomaJuha for catching this! It's definitely a bug and my oversight from #2633.

module/VuFind/src/VuFind/Controller/CoverController.php
LuomaJuha: Now it should convert them into an array as default. And keep all the array values.

module/VuFind/src/VuFind/Controller/CoverController.php
LuomaJuha: isbns it is now.

xmorave2 (approved):
Thanks for quick update @LuomaJuha. I do some hand on testing using ObalkyKnih driver with ajax covers disabled and it is working fine.

demiankatz (approved):
Looks good! Thanks, @LuomaJuha and @xmorave2!

PR #2801, opened 2023-03-22
Fail with exception if a MARC file fails to load during merge operation.

demiankatz:
TODO
- [ ] Should we add a command line switch to ignore errors?


demiankatz:
@EreMaijala, it seems based on discussion in vufind-tech that we can just merge this as-is, unless you have concerns!

EreMaijala (approved):
@demiankatz Agreed, and no concerns.

demiankatz:
Thanks, @EreMaijala! I decided to add test coverage to this, which ultimately prompted me to make the code a little more sophisticated. Since tests are passing, I feel pretty safe merging this version, but if you take another look and have concerns, please let me know.

EreMaijala (approved):
Even better!

PR #2803, opened 2023-03-22
npm script wrappers for Grunt

crhallberg:
Step one of grunt replacement.

- [x] ~~Do we want to do Phing theme actions (eslint, jshint)?~~
- [x] ~~Do we want to do other CLI theme actions (compress, makeTheme)?~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,16 @@
   "bugs": {
     "url": "https://vufind.org/jira"
   },
+  "scripts": {
+    "less": "grunt less",
+    "lessdev": "grunt lessdev",
+    "scss": "grunt scss",
+    "lessToSass": "grunt lessToSass",
+    "watch:less": "grunt watch:less",
+    "watch:lessdev": "grunt watch:lessdev",
+    "watch:scss": "grunt watch:scss",
+    "watch:lessToSass": "grunt watch:lessToSass"
+  },
   "dependencies": {
     "grunt": "^1.0.4",
     "grunt-contrib-less": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -10,10 +10,14 @@
     "url": "https://vufind.org/jira"
   },
   "scripts": {
+    "css": "npm run less",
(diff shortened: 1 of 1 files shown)


demiankatz (changes requested):
Thanks for getting this started. I've marked this as a 9.0 milestone and moved the more comprehensive work in #2624 to the 10.0 milestone.

We had discussed making things more general -- so maybe it's worth having a meta-script like `compilecss` that's an alias for the less task, and then we can recommend that people use that task in documentation, so if we change our process in future, the basic command remains the same. I think that's the only change I'd like to see in order to get this merged.

I don't think we need to add anything extra for now -- I think the main goal of this first step is to move away from Grunt. It may be worth having a broader discussion about which tasks should be initiated by npm, which by phing, and which by composer, but I don't think we need to decide on that to move this PR forward.

demiankatz (approved):
Thanks, @crhallberg, I think this is a step in the right direction! I've revised the wiki documentation to redirect most Grunt references to NPM now. See, for example: https://vufind.org/wiki/development:npm

We should be sure to keep this wiki page up to date as things develop further!

aleksip:
@demiankatz @crhallberg I'm a bit slow to comment, but this is a great direction. We use npm scripts exclusively in our Admin Interface and UI projects.

We have settled with a `verb` and `verb:object` pair for naming, which has worked quite well so far. So we have for example:

```
build
build:css
build:sass
build:pcss
lint
lint:css
lint:js
```
And so on. Here `build` calls `build:css` which in turn calls both `build:sass` and `build:pcss`.

We use a colon to separate the words because there are npm tools that support it for running scripts, so for example the `lint` script can just call `lint:*`.

demiankatz:
Thanks, @aleksip, this makes sense. If @crhallberg wants to revise the names, I'm happy to revise the references in the wiki. And of course there's nothing stopping us from keeping the old, shorter names as aliases if that would help ease the transition from grunt.

PR #2804, opened 2023-03-22
Add support for loading results with JS

EreMaijala:
This turns on, by default, loading of results with JS when paging or changing sort order or result limit. The aim was to do this with as few template changes as possible and retain back-compatibility. It's also possible to turn off with a setting, but I believe defaulting to on due to accessibility (see below) and performance would make sense (though please let me know if this is a bad idea!).

This is related to the accessibility issue discussed in #2492, and I've spent some time trying to ensure that the controls work without issues here. @pasitiis has provided some feedback to ensure that e.g. the aria-live attribute is used properly.

Goals:
- Should still look and feel like VuFind normally does
- Page URL must reflect current state (e.g. reloading the page must lead to same results)
- Navigating history in browser must work (e.g. browser's back and forward buttons)

TODO:
- [X] The `base_page_max_age` needs to be implemented if deemed useful. Not worth it. It's 
- [X] Testing functionality (all search backends, history handling, error handling, page address handling etc.)
- [ ] Testing with different browsers (now tested with Firefox, Chrome and Safari on MacOS)
- [ ] Documentation?
- [x] Some refactoring could be separated from this
- [ ] openurl.js is a bit troublesome
- [ ] Additional Mink tests?


EreMaijala:
Mink tests are passing for me with the included changes to not expect query time.

themes/bootstrap3/templates/search/list-list.phtml
```diff
@@ -1,7 +1,20 @@
...
+  $listClasses = ['record-list'];
+  if ($this->results->getOptions()->loadResultsWithJsEnabled()) {
+    $this->headScript()->appendFile('search.js');
+    // Ensure OpenURL support is available even if none of the results on the initial
+    // page has one:
+    $this->headScript()->appendFile('openurl.js');
+    $listClasses[] = 'js-record-list';
+  }
+  if ($class = $this->resultsClass ?? '') {
+    $listClasses[] = $this->escapeHtmlAttr($class);
```
demiankatz: Would it be more clear to escape the imploded list below, or else to add a variable suffix to clarify that these are already escaped?

themes/bootstrap3/templates/search/list-grid.phtml
```diff
@@ -1,4 +1,17 @@
...
+  $listClasses = ['search-grid'];
+  if ($this->results->getOptions()->loadResultsWithJsEnabled()) {
+    $this->headScript()->appendFile('search.js');
+    // Ensure OpenURL support is available even if none of the results on the initial
+    // page has one:
+    $this->headScript()->appendFile('openurl.js');
+    $listClasses[] = 'js-record-list';
+  }
+  if ($class = $this->resultsClass ?? '') {
+    $listClasses[] = $this->escapeHtmlAttr($class);
```
demiankatz: If changes are made to list-list.phtml, they'll need to be matched here. I wonder if there's a way to avoid redundancy between these two templates (though that doesn't seem likely without a view helper, which is probably overkill).

demiankatz (changes requested):
I've taken only a very cursory initial glance at this, but I left a couple of minor comments. I also noticed a couple of stand-alone changes that I've gone ahead and merged separately to dev to reduce the size of this PR. I did a quick hands-on test and think this is off to a great start. Thanks for getting it underway, @EreMaijala!

themes/bootstrap3/templates/search/list-grid.phtml
EreMaijala: This was already the case before, just not as prominent. Maybe cross-linked comments would help?

demiankatz: True, I just brought it up because a significant amount of parallel logic was being added to both templates. I agree that adding cross-linked comments certainly can't hurt.

themes/bootstrap3/templates/search/list-list.phtml
EreMaijala: I refactored this out. :)

themes/bootstrap3/templates/search/list-grid.phtml
EreMaijala: I did some refactoring to reduce duplication here as well as in GetSearchResults AJAX handler by adding results-list.phtml template between results.phtml and list-*.phtml (kind of like what combined search has, but just for setting classes and other attributes).

config/vufind/searches.ini
```diff
@@ -154,6 +154,15 @@ retain_filters_by_default = true
 ; JavaScript is not supported by the browser.
 load_results_with_js = true
 
+; This setting can be used to configure pagination control on top of results.
+; Possible options are:
+; empty string or false  No top pagination (default when load_results_with_js is
```
demiankatz, line 159: I like the top paginator, though I'd definitely want to hear more input from people more invested in UI design than me. 

However, I don't think the comments about how it behaves match the code -- I don't get different defaults when I turn off AJAX.

demiankatz (changes requested):
I haven't had time to delve into this in much more depth, but I did look at the top navigation options you recently added. See below!

xmorave2:
Nice work @EreMaijala! I tried it an it worked fine, but I found one bug - the search result pages loaded by javascript do not have covers when ajaxcovers is enabled: there is no ajax call for covers intitiated.

EreMaijala:
@xmorave2 Thanks! I'll fix the issue. I originally included all functionality that was initialized by initResultScripts, but I've already found several others that were missing from it.

config/vufind/searches.ini
EreMaijala: Default behavior fixed.

EreMaijala:
@xmorave2 AJAX covers should work now. I added an extra check to ensure that a cover is never loaded twice just to be sure.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
```diff
@@ -0,0 +1,389 @@
...
+     *
+     * @return ?Results
+     */
+    protected function getSearchResults(Params $params): Results
+    {
+        parse_str($params->fromQuery('querystring', ''), $searchParams);
+
+        $backend = $params->fromQuery('source', DEFAULT_SEARCH_BACKEND);
+        $searchType = $params->fromQuery('searchType', '');
+        if ('versions' === $searchType) {
```
demiankatz, line 217: Is there a way we can refactor code to reduce the redundancy between this and the RecordVersionsSearchTrait? I realize that things are slightly different in each place, but there's so much overlap that this seems like a potential maintenance problem if we can't find a way to better centralize it. I also don't particularly like having so much versions-specific logic in what should be a more generic AJAX helper -- it would be great if we could find a way to better decouple these things.

themes/bootstrap3/js/cart.js
```diff
@@ -161,7 +161,7 @@ VuFind.register('cart', function Cart() {
     var $form = typeof _form === 'undefined'
       ? $('form[name="bulkActionForm"]')
       : $(_form);
-    $("#updateCart, #bottom_updateCart").unbind('click').click(function cartUpdate() {
+    $("#updateCart, #bottom_updateCart").off('click').on('click', function cartUpdate() {
```
demiankatz: Should we have a separate PR to fix use of deprecated `unbind`? This seems to happen in quite a few places, and maybe it's something we can/should clean up separately from the other work here.

themes/bootstrap3/js/covers.js
```diff
@@ -57,14 +57,19 @@ function loadCoverByElement(data, element) {
 
 function loadCovers() {
   $('.ajaxcover').each(function getDataAndLoadCovers() {
-    var img = $(this).find('img');
+    let $cover = $(this);
```
demiankatz, line 60: Would it be worth merging the changes to this file separately, since this seems like a generic improvement to prevent double-loading? (Anything to shave files off a PR!)

themes/bootstrap3/templates/search/results.phtml
```diff
@@ -30,26 +30,22 @@
...
-  // Create a css class for results from the search class:
-  $searchClass = $this->params->getSearchClassId();
-  if (!ctype_upper($searchClass)) {
-    $searchClass = preg_replace('/([a-zA-Z])(?=[A-Z])/', '$1-', $searchClass);
-  }
-  $this->resultsClass = 'search-results-' . strtolower($searchClass);
 
-  $this->render('search/results-scripts.phtml', ['displayVersions' => !empty($this->params->getOptions()->getVersionsAction())]);
-
-  // Load only if list view parameter is NOT full:
```
demiankatz, line 48: Is refactoring this code into results-scripts.phtml actually a bug fix? Are there current scenarios where combined search or versions break if we use non-full list options? Or is such a configuration not possible? (I didn't test this, but I can look into it if you think this might be something we should worry about, and potentially merge early).

demiankatz (changes requested):
@EreMaijala, I still haven't found time to give this a really thorough review, but I had a few minutes and used them to randomly browse around and raise a few issues as I saw them. (As usual, I'm looking at the files with fewer/shorter diffs first, before delving into the big complicated ones :-) ).

themes/bootstrap3/templates/search/results.phtml
EreMaijala: Well, it's a bug fix, but apparently nobody has noticed it, so not a high priority in itself. But the refactoring and fixes to result page script initialization could be extracted to a separate PR. I'll see what I can do!

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
EreMaijala: I extracted a tiny bit to a new trait, but it's not much. The special case is needed because versions search doesn't have a lookfor parameter, to ensure that the work keys are current and to make sure it's not saved in history.

It would be neat if we could do most of this in initFromRequest, but then the Params object would need a record loader, and that feels wrong too.

themes/bootstrap3/js/cart.js
EreMaijala: Yes. I'll take care of it.

themes/bootstrap3/js/covers.js
EreMaijala: I'll take a look.

themes/bootstrap3/js/cart.js
EreMaijala: See PR #2876.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
demiankatz: What if, instead of a trait, we had a `\VuFind\Record\VersionsHelper` service, which could in turn have a dependency on the RecordLoader? Then we could fetch the VersionsHelper instead of the RecordLoader in the RecordVersionsSearchTrait, and we could make the VersionsHelper a direct dependency of the AJAX handler instead of the RecordLoader. I think this would allow further consolidation of logic (e.g. retrieval of keys).

EreMaijala: An excellent idea, thanks! I was so concentrated on the trait approach that I couldn't think that far. :)

demiankatz: No problem! Traits are very inviting because they're so easy, but I think they're actually only really good in a limited set of situations because they are hard to override/customize. I'm sure I haven't perfectly lived up to it, and I'm sure there are valid exceptions to the rule, but I'm trying to make this my rule of thumb: if it's behavior shared by a whole bunch of different things (i.e. a reusable pattern), it should be a trait. If it's centralized logic (i.e. complex behavior that does exactly the same thing in different parts of the code), it should be a service so that it can be overridden/extended.

EreMaijala: The change has been done, but I'll hold off any review requests until after #2876 and #2878 are done.

EreMaijala:
I'll open a new pull request targeting dev-10.0 branch, so closing this one.

EreMaijala:
Continued in #2929.

PR #2805, opened 2023-03-23
Avoid JS error when there is no callnumAndLocation element.

EreMaijala:


demiankatz (approved):
Thanks, @EreMaijala! I've added a test to cover this scenario, so I'll merge this once CI passes and you approve my test.

PR #2806, opened 2023-03-23
Fix double-encoding of hidden filters in advanced search tabs.

EreMaijala:
This follows the convention of basic search tabs where the remapped url is not encoded by the searchTabs helper.


EreMaijala:
Mink tests are still passing for me.

demiankatz (approved):

demiankatz:
Thanks, @EreMaijala. Merging now. As discussed on Slack, we should add a test to cover this when time permits.

PR #2807, opened 2023-03-24
Fix export links not working

LuomaJuha:
The export buttons in toolbar are not working due to links having extra data. This should fix that problem.

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/RecordLinker.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/RecordLinker.php
@@ -138,6 +138,7 @@ class RecordLinker extends \Laminas\View\Helper\AbstractHelper
      * to, or source|id pipe-delimited string
      * @param string                $action  Record action to access
      * @param array                 $query   Optional query parameters
+     * @param string                $anchor  Optional anchor
      * @param array                 $options Optional options
      *
      * @return string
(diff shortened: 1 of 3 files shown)


demiankatz:
Thanks, @LuomaJuha! We should probably add a Mink test to cover this situation. I should have some time later today to work on that, unless you'd like to do it sooner. :-)

LuomaJuha:
@demiankatz I could try and see if I get the mink tests running, never done that

demiankatz:
@LuomaJuha, if you want to try it, you can probably model it on one of the simpler tests, like my recently-built https://github.com/vufind-org/vufind/blob/dev/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SearchViewsTest.php -- but if you don't have time, just let me know and I'll work on it later. :-)

LuomaJuha:
@demiankatz I made a very small test, if that would work somehow!

demiankatz (approved):
Thanks, @LuomaJuha, this is great!

I just made one small simplification to the test: the changeConfigs() call was not needed because the values you were setting are already present by default.

LuomaJuha:
@demiankatz Yeah, I added it for just to be sure! I should try to check how I can run those mink tests on Linux as they are very neat and easy to write.

demiankatz:
@LuomaJuha, if you want more background on setting up the Mink tests, the [video on the subject](https://vufind.org/wiki/videos:testing_3) might be interesting to you! (The [setting up a test environment](https://vufind.org/wiki/videos:testing_1) video is a prerequisite, if you haven't already seen that one).

PR #2808, opened 2023-03-27
Fix visual view link and jump to record handling.

EreMaijala:
If user jumps to visual view from e.g. the second page, we need to ensure the view url doesn't have a page number. This avoids drilling into visual results from displaying e.g. results 21 to 2 of 2. Also adds a check that a record was returned before trying to jump to it.

I'm not sure how much the visual view is used since it seems a bit raw to me, but at least this prevents an error that could perhaps happen in other circumstances as well, at least with a hand-crafted url if nothing else.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2809, opened 2023-03-27
Use standard dropdown control for QR codes in results list.

EreMaijala:
Adds a template for the QR code to avoid duplication of code.


themes/bootstrap3/templates/RecordDriver/DefaultRecord/controls/qrcode.phtml
```diff
@@ -0,0 +1,15 @@
+<?php if ($QRCode = $this->record($this->driver)->getQRCode($this->context)): ?>
+  <div class="dropdown hidden-xs">
+    <a href="#" class="dropdown-toggle qrcodeLink result-link icon-link" data-toggle="dropdown" aria-expanded="false" rel="nofollow">
+      <?=$this->icon('qrcode', 'icon-link__icon') ?>
+      <span class="icon-link__label"><?=$this->transEsc('qrcode_show')?></span>
+    </a>
+    <ul class="dropdown-menu qrcode">
+      <li>
+        <script type="text/template" class="qrCodeImgTag">
+          <img alt="" src="<?=$this->escapeHtmlAttr($QRCode);?>">
```
demiankatz: Are you intentionally omitting the alt here? (I realize that an alt of "QR Code" isn't especially useful... but I wonder if it's better than nothing... also, we should be consistent across different QR Code displays, and the alt tag is still present in core.phtml and collection-info.phtml).

themes/bootstrap3/js/common.js
```diff
@@ -619,22 +619,12 @@ function setupQRCodeLinks(_container) {
   var container = _container || $('body');
 
   container.find('a.qrcodeLink').click(function qrcodeToggle() {
-    if ($(this).hasClass("active")) {
-      $(".result-link-label", this).html(VuFind.translate('qrcode_show'));
-      $(this).removeClass("active");
-    } else {
-      $(".result-link-label", this).html(VuFind.translate('qrcode_hide'));
```
demiankatz, line 626: The reason this wasn't working is that the qrcode translation strings were not being passed through in layout/js-translations.phtml. Would it be better to fix that rather than to remove the hide translation? (I don't have strong feelings one way or another, but just raising the question).

demiankatz (changes requested):
Thanks! This looks pretty good, but I had a couple of questions before merging...

themes/bootstrap3/js/common.js
EreMaijala: I decided to remove it for consistency with all other dropdowns that don't change the text. But it could be added back if that's better.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/controls/qrcode.phtml
EreMaijala: It was intentional, but you're right in that it was also inconsistent. I restored the alt text here.

themes/bootstrap3/js/common.js
demiankatz: Makes sense to me, but I'm going to ask @crhallberg for an additional review in case he has any further thoughts or opinions. :-)

demiankatz (approved):

themes/bootstrap3/js/common.js
crhallberg: I'm of the opinion that the text should change, especially since we already have the translation strings. In the future, we can make the other toggles consistent.

crhallberg (approved):
I'll leave it up to you and DK about the translation strings, but this is definitely a huge improvement. Thank you!

themes/bootstrap3/js/common.js
demiankatz: Thanks, @crhallberg. If @EreMaijala doesn't mind putting back the text switching, it would greatly reduce the number of files touched by this PR and reduce the behavior changes as well. But if he feels strongly that it's better this way, he already has approvals from both of us and is free to hit the merge button. I'll leave the final decision up to him! :-)

EreMaijala: The problem with text switching, as I see it, is that it makes the aria attributes a bit weird. If you have `aria-expanded="true"` on a toggle that says "Hide QR Code", it just doesn't make a whole lot of sense. Maybe it would be best to remove the verb and just display e.g. "QR Code ▼"? We already have a translation for "QR Code", so no new language strings would be needed.

demiankatz: @EreMaijala, that sounds like a great compromise to me, and I think it probably lets us get rid of a second translation string. :-)

EreMaijala: Done. :)

demiankatz:
Looks good to me! I'm running the test suite to be safe, and I'll merge if it passes.

PR #2810, opened 2023-03-29
Add ability to retrieve record links from ILS driver.

demiankatz:
TODO
- [x] Test KohaRest driver


demiankatz:
@EreMaijala, it has been an occasional request to add links directly to the Koha OPAC from VuFind, and this PR is a proof of concept to create a mechanism that allows ILS drivers to provide links to record drivers. I'm interested in your thoughts. In particular:

1.) Can you test the KohaRest changes? I wrote the code but can't easily test it.

2.) Do you think it's sufficient to have a view_in_opac translation string that we can override, or should we make the description configurable via KohaRest.ini?

3.) Any other thoughts on the general architecture here?

Since this adds a new language string, I'm targeting it for the 9.1 release.

config/vufind/KohaRest.ini
```diff
@@ -57,6 +57,9 @@ host = "http://koha-server/api"
 clientId = ""
 ; Koha API Client Secret for OAuth2 authentication
 clientSecret = ""
+; Your Koha OPAC record display URL, if you want to provide direct links to your Koha
+; OPAC. Leave this commented out to omit OPAC links.
+;opacURL = "http://koha-server/cgi-bin/koha/opac-detail.pl"
```
EreMaijala: This works right now, but I'd like the configuration to specify a complete URL with a placeholder for the ID so that the URL can be formulated freely. Something like `opacURL = "http://koha-server/cgi-bin/koha/opac-detail.pl?biblionumber=%%id%%"` would be great.

EreMaijala (changes requested):

config/vufind/KohaRest.ini
demiankatz: Thanks, I've added the suggested flexibility, as well as the beginning of a unit test for the KohaRest driver (currently covering only the newly-added method).

EreMaijala (approved):

PR #2811, opened 2023-03-31
Improve author deduplication

mtrojan-ub:
Using array_merge for deduplication is causing side effects. You can see that especially when you have e.g. the same author with different roles. Here's an example related to ubtue/tuefind#2164:

primary:
```
Array
(
    [role] => Array
        (
            [0] => aut
        )
    [id] => Array
        (
            [0] => 081889968
        )
)
```
secondary:
```
Array
(
    [role] => Array
        (
            [0] => trl
        )
    [id] => Array
        (
            [0] => 081889968
        )
)
```

With array_merge, it will be merged into:
```
Array
(
    [role] => Array
        (
            [0] => trl
        )
    [id] => Array
        (
            [0] => 081889968
        )
)
```

This means that all existing fields from primary will be completely overridden by secondary (if they also exist there), losing the primary role. 

So i suggest that we use array_merge_recursive instead, which might create some duplicate values first:
```
(
   [role] => Array
        (
            [0] => aut
            [1] => trl
        )
    [id] => Array
        (
            [0] => 081889968
            [1] => 081889968
        )
)
```

But after the following dedup_data step, it will finally be cleaned up and look like this:
```
(
    [role] => Array
        (
            [0] => aut
            [1] => trl
        )

    [id] => Array
        (
            [0] => 081889968
        )
)
```


demiankatz:
Thanks, @mtrojan-ub, do you think it would be feasible to add some unit tests around this case?

demiankatz:
(I think it shouldn't be too hard to add something to the [existing test](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/tests/unit-tests/src/VuFindTest/RecordDriver/DefaultRecordTest.php).)

demiankatz (changes requested):
Just putting this into "request changes" mode to reflect my test suggestion above. :-) Let me know if I can help with anything!

mtrojan-ub:
Good idea! I added a unit test, can you please take another look?

demiankatz (approved):
Makes sense. Thanks for adding the test, @mtrojan-ub!

PR #2812, opened 2023-03-31
Record Title labels for checkboxes

jpkanter:
The template `checkbox.phtml`gets called by the Record View Helper and by default only adds some rudimentary info to the template itself, there is also the option to provide a counter for a translated helper text á la "Select entry %X%".

In a recent accessibility report the reporting body noted that it would be better if the checkbox had a label actually referring to the title of the record. In any given fork this can be achieved pretty easily by overwriting the template and creating a variable prior to calling the controller.

I changed the record view helper so it always includes the record `TitleHtml()` (the function is a bit weirdly named as it actually does not give an HTML title but an escaped one. The previous way it worked is preserved and takes precedence over the record title for backwards compatibility. This especially helps catalogues that do not include the number of the record in their number (as many FINC library do for instance). 

There was some discussion regarding this change internally. The title is basically doubled for a screen reader so it bloats the whole thing a bit. Therefore I wanted to broaden the audience for this specific matter.

_side note: why is the checkbox.phtml so horribly compressed. I kept the style as it, but its very _unfun_ to edit_

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/Record.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/Record.php
@@ -414,7 +414,7 @@ class Record extends \Laminas\View\Helper\AbstractHelper
             . $this->driver->getUniqueId();
         $context
             = ['id' => $id, 'number' => $number, 'prefix' => $idPrefix,
-               'title' => $this->getTitleHtml()];
+               'titleHtml' => $this->getTitleHtml()];
         if ($formAttr) {
             $context['formAttr'] = $formAttr;
         }
--- a/module/VuFind/src/VuFind/View/Helper/Root/Record.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/Record.php
@@ -413,7 +413,8 @@ class Record extends \Laminas\View\Helper\AbstractHelper
         $id = $this->driver->getSourceIdentifier() . '|'
             . $this->driver->getUniqueId();
         $context
-            = ['id' => $id, 'number' => $number, 'prefix' => $idPrefix];
+            = ['id' => $id, 'number' => $number, 'prefix' => $idPrefix,
+               'title' => $this->getTitleHtml()];
         if ($formAttr) {
             $context['formAttr'] = $formAttr;
         }
--- a/themes/bootstrap3/templates/record/checkbox.phtml
+++ b/themes/bootstrap3/templates/record/checkbox.phtml
@@ -8,8 +8,8 @@
   <span class="checkbox-icon"></span>
   <?php if (strlen($this->number ?? '') > 0): ?>
     <span class="sr-only"><?=$this->transEsc('result_checkbox_label', ['%%number%%' => $this->number]) ?></span>
-  <?php elseif (strlen($this->title ?? '') > 0): ?>
-    <span class="sr-only"><?=$this->title?></span>
+  <?php elseif (strlen($this->titleHtml ?? '') > 0): ?>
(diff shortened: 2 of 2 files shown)


demiankatz (changes requested):
Thanks, @jpkanter! A few notes:

1. Regarding `getTitleHtml()`, it does indeed return HTML -- I believe that HTML-escaped text counts as HTML, and if highlighting is enabled and applicable, the value may include tags as well. (The point here is that this text can be output as part of an HTML document without the need for additional escaping).

2.) Regarding the formatting of checkbox.phtml, I agree that it is quite ugly and reformatting would be helpful. Please feel free to make changes if you wish, while we are modifying the file anyway. The only reason I can think of for having such long lines is if we are intentionally avoiding the insertion of line breaks into the HTML, but I don't see an obvious reason why that would cause a problem here (of course, maybe reformatting it will reveal a problem).

3.) I see that some tests are failing -- let me know if you need help figuring out why.

jpkanter:
It seems the test fails because there is no mockup method for escaping HTML inside the getTitleHTML() function 

```
1) VuFindTest\View\Helper\Root\RecordTest::testGetCheckbox
Error: Object of type Mock_escapeHtml_739ea212 is not callable
``

File: module/VuFind/src/VuFind/View/Helper/Root/Record.php [hunk 1]
```diff
@@ -413,7 +413,8 @@ public function getCheckbox($idPrefix = '', $formAttr = false, $number = null)
         $id = $this->driver->getSourceIdentifier() . '|'
             . $this->driver->getUniqueId();
         $context
-            = ['id' => $id, 'number' => $number, 'prefix' => $idPrefix];
+            = ['id' => $id, 'number' => $number, 'prefix' => $idPrefix,
+               'title' => $this->getTitleHtml()];
```
demiankatz: I would recommend calling this value `titleHtml` instead of just `title`. This makes it more clear that it does not need additional escaping applied in the checkbox.phtml template -- otherwise I think there's a chance that in future reviews this might trigger people to reinvestigate code unnecessarily.

demiankatz (changes requested):
Thanks for the progress on this, @jpkanter. One small suggestion.

Thanks also for identifying the cause of the test failure. Do you need help fixing that? If so, I might be able to find some time next week when I'm back from vacation.

jpkanter:
> Thanks also for identifying the cause of the test failure. Do you need help fixing that?

I fear that I do indeed need assistance in resolving the Code Check error, as of now I simply lack understanding. 

I understood it so that the tests are part of `VuFind/dev`, but if I would extend the mock methods (or rather make it use the actual one because that is not what is being tested here) I could not do that in this pull request. The fact that you indicate that I could fix that myself, indicates I am either not seeing something or have a grave miss understanding on how this works.

jpkanter:
It  turns out that *we* got an actually bigger and more in-depth ticket internally that has the content of this pull request as a subset of covered features. Therefore I am closing this pull request in favor for a later, more through submission by one of my colleagues or me.

demiankatz:
Thanks, @jpkanter, though if you'd like to merge some or all of this to make the other job smaller, I'm never opposed to adding functionality in small steps. If nothing else, the style improvements to checkbox.phtml seem like they could be beneficial right away. Feel free to reopen if you wish!

demiankatz:
Also, regarding the test issue from above, there's code in the test suite that defines mock helpers for the view-related tests. Most likely, we just need to define another mock helper now that there are new dependencies in this template. If you decide to pursue this further, I'm happy to help with that as needed.

PR #2815, opened 2023-04-04
Aleph driver: fix return value from cancelHolds

xrosecky:
Aleph driver: return value from cancelHolds fixed according to ILS driver documentation


demiankatz (approved):
Thanks, @xrosecky!

PR #2818, opened 2023-04-04
Prevent endless redirect loop if library card selection was referred …

EreMaijala:
…to by itself.

Steps to reproduce:

1. Open two browser tabs
2. Log in in the first tab
3. Open same address in second tab
4. Switch back to the first tab
5. Make sure the account has at least two library cards
6. Go to e.g. holds page where the library card selection is enabled
7. Log out in the second tab
8. Select another library card in the first tab
9. Log in again


demiankatz (approved):
Thanks, @EreMaijala!

PR #2819, opened 2023-04-04
Fix active library card check if user has no card selected.

EreMaijala:
This probably only happens if storedCatalogLogin fails.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2820, opened 2023-04-04
Reset focus to bookbag toggle buttons after action.

charbugs:
Hi VuFinders,

this PR is an small accessibility improvement. It makes the keyboard focus stay on the bookbag toggle button when hitting the enter key.

To test this feature you need to set `showBookBag = true` in config.ini.


demiankatz (approved):
This makes sense to me, but I'm also going to make sure @crhallberg agrees before merging.

demiankatz:
(For the record: I just ran the full Mink test suite, and all tests are passing).

demiankatz:
I also did some hands-on testing and see how this is clearly a functional improvement... but I'll wait and make sure @crhallberg has no technical suggestions.

crhallberg (approved):
Looks good to me!

PR #2821, opened 2023-04-04
Guard against unbalanced hierarchy fields.

charbugs:
From the VuFind documentation:

> **hierarchy_sequence** - A value indicating the position of the current record relative to its siblings, used for sorting records when rendering trees. This may be multi-valued if a record belongs to multiple collections; **the order of multiple values in this field must correspond with values in hierarchy_parent_id**.

For some records we have trouble to met this requirements, i.e. we have different lengths for `hierarchy_sequence` and `hierarchy_parent_id`. In https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/Hierarchy/TreeDataFormatter/AbstractBase.php#L131 this results in an error trying to read a non-existing array element. In the end the hierarchy tree will not show up on the page.

This PR adds an check to ensure the two fields are balanced (similar to the check in https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/Hierarchy/TreeDataFormatter/AbstractBase.php#L154)

In case the check fails the method returns an empty result array. In consequence the record in question will be sorted to the top of the list or tree (see https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/Hierarchy/TreeDataFormatter/Json.php#L100)

Of course, it would be better to met the requirements mentioned above but it seems to be not easy for some records. This PR would enable us (and probably others?) to show the hierarchy tree for this records in a slightly random order rather than not at all.


demiankatz (commented):
Thanks, @charbugs. I agree that some more flexibility could be helpful here, but I have a few ideas/questions:

1.) Would it make sense to add a log message/warning, and/or to have a config setting that controls how count mismatches are handled, so that users can allow these inconsistencies when they are necessary, but do not silently allow them through when they are unexpected? It might be best if we could allow your use case without increasing the possibility of allowing subtle bugs to go unnoticed.

2.) Rather than a strict `===` check, would a `<=` check be better, since then if we have EXTRA position data for some reason, the existing position data is not totally ignored? Or what if we just say `$fields->hierarchy_sequence[$key] ?? 0` inside the loop to default to position 0 when data is missing?

In any case, I'm open to potentially accepting this as-is, but I thought these points might be worth discussing first.

Thanks again!

charbugs:
Hi Demian,

> It might be best if we could allow your use case without increasing the possibility of allowing subtle bugs to go unnoticed.

I see your point and I agree.

> Would it make sense to add a log message/warning, and/or to have a config setting that controls how count mismatches are handled ...

This would probably the best solution to allow this kind of flexibility. That said, it seems difficult to me how such a config setting may look like. `skip_if_hierarchy_parent_id_and_hierarchy_sequence_is_not_balanced` is to narrow and would not add much value to VuFind. On the other hand, `skip_if_something_weird_with_hierarchy_data` would presumably require a lot more complex checks in a bunch of files.

Not sure how to tackle this at the moment ...

demiankatz:
Thanks, @charbugs. I think there are  a few questions to answer:

1.) As you say, what should the configuration be called, and where should it live? I'd suggest something like `validate_hierarchy_sequences` with a comment like `; If true, throw an exception if hierarchy parent and sequence data is out of sync`. It probably makes sense to put the new setting in the HierarchyDefault.ini / HierarchyFlat.ini files, since that's where hierarchy sorting is configured, and I think there may be some parallels between these settings.

2.) Once the configuration is defined, how should it be provided to the TreeDataFormatter, since at present, those classes do not have direct access to configuration? We would either need to modify the factory to pass in a configuration, or else we might need to add a new parameter to the setRawData method and then pass the value down from higher-level code. I don't have a strong preference, but I can spend more time looking closely at the code if you want my opinion!

charbugs:
I updated the PR arccording to your last comment. 

It works as intended but I'm not sure if simply throwing an exception is sufficient to let the developer know the fields are unbalanced because effectivily the exception results in a Javascript error in der browser console. But the Javascript error message (something like "can not read 'top' from undefined")  does not give a hint to the root of the problem.

What do you guys think?

module/VuFind/src/VuFind/Hierarchy/TreeDataFormatter/AbstractBaseFactory.php
```diff
@@ -69,7 +69,8 @@ public function __invoke(
         if ($options !== null) {
             throw new \Exception('Unexpected options sent to factory!');
         }
-        $treeDataFormatter = new $requestedName();
+        $config = $container->get(\VuFind\Config\PluginManager::class)->get('HierarchyDefault')->get('HierarchyTree');
```
demiankatz: It's been a long time since I've looked at this code, but I can't remember under what circumstances HierarchyDefault gets used, and under what circumstances HierarchyFlat gets used. I think it may be driven by the record driver, in which case it may be impossible to know which configuration is in play at the time this factory executes. If that's the case, maybe moving this setting to the `[Hierarchy]` section of config.ini would be safer.

module/VuFind/src/VuFind/Hierarchy/TreeDataFormatter/AbstractBase.php
```diff
@@ -77,6 +77,23 @@ abstract class AbstractBase implements \VuFind\I18n\HasSorterInterface
...
+     * @var bool
+     */
+    protected $validateHierarchySequences;
+
+    /**
+     * Constructor
+     *
+     * @param bool $validateHierarchySequences Throw an exception if hierarchy parent and sequence data is out of sync?
+     */
+    public function __construct($validateHierarchySequences)
```
demiankatz: Might be good to provide a default value here for greater backward compatibility.

demiankatz (changes requested):
Thanks for the progress here, I think this is coming along well. In addition to the notes below, a couple of thoughts:

1.) A unit test would probably be helpful (let me know if you'd like any help developing that).

2.) I wonder if improved error handling in the Javascript layer might be a valuable complement to this (based on what you said about the behavior).

demiankatz:
Thanks, @charbugs, this looks pretty reasonable to me. So two questions before I merge this:

1.) Do you want to work on improving error handling on the Javascript side now, or is that something better addressed in a follow-up PR?

2.) Do you have any interest in adding unit test coverage for the new functionality?

If you're not able to do either of those things, I'm willing to help when I have a bit more time (after the 9.0 release). But if you're interested in helping with either or both, that's great too! :-)

charbugs:
Hi Demian,

I added an unit test for the `getHierarchyPositionsInParents` method.

Regarding the error handling, I fixed a bug in the `HierarchyController`. With that fix the error message from `getHierarchyPositionsInParents` will be visible in the browser console in development mode. I think this sufficient to hint the developer what went wrong.

charbugs:
One note about the test. The method under test is `getHierarchyPositionsInParents` and is defined in `AbstractBase`. But I tested the `Json` class because I could not manage to instanciate the `AbstractBase` class even with the help of the reflection api. Let me know if there is a better way to test abstract classes.

module/VuFind/src/VuFind/Hierarchy/TreeDataFormatter/AbstractBaseFactory.php
```diff
@@ -69,7 +69,8 @@ public function __invoke(
         if ($options !== null) {
             throw new \Exception('Unexpected options sent to factory!');
         }
-        $treeDataFormatter = new $requestedName();
+        $config = $container->get(\VuFind\Config\PluginManager::class)->get('config')->get('Hierarchy');
+        $treeDataFormatter = new $requestedName($config->get('validateHierarchySequences'));
```
demiankatz: I'd refactor this to:

```
        $config = $container->get(\VuFind\Config\PluginManager::class)->get('config');
        $treeDataFormatter = new $requestedName($config->Hierarchy->validateHierarchySequences ?? true);
```

This is a little more consistent with the config access syntax we use in the rest of the code (we don't typically use the getter on config objects) and it provides a more explicit default value if the setting is absent.

module/VuFind/tests/unit-tests/src/VuFindTest/Hierarchy/TreeDataFormatter/JsonTest.php
```diff
@@ -0,0 +1,133 @@
+<?php
+
+/**
+ * Hierarchy TreeDataFormatter Json Test Class
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2010.
```
demiankatz: Should probably make this date 2023.

module/VuFind/tests/unit-tests/src/VuFindTest/Hierarchy/TreeDataFormatter/JsonTest.php
```diff
@@ -0,0 +1,133 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Tests
+ * @author   Demian Katz <demian.katz@villanova.edu>
```
demiankatz: Please feel free to give yourself credit for doing this work. :-)

module/VuFind/tests/unit-tests/src/VuFindTest/Hierarchy/TreeDataFormatter/JsonTest.php
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Invokes a private or protected method and returns its return value.
+     *
+     * @param object $obj     Object
+     * @param string $method  Method name
+     * @param mixed  ...$args Arguments passed to method
+     *
+     * @return mixed
+     */
+    private function invokePrivateMethod($obj, $method, ...$args)
```
demiankatz: You can mix in VuFind's ReflectionTrait to get this functionality without having to redefine it in your test. See:

https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFindTest/Feature/ReflectionTrait.php#L56

demiankatz (changes requested):
Thanks for the progress here, this is looking great! Just a few minor suggestions.

Regarding your question about testing AbstractBase vs. Json, you could likely test AbstractBase by creating a mock using getMockForAbstractClass -- see https://docs.phpunit.de/en/9.5/test-doubles.html#mocking-traits-and-abstract-classes for details.

demiankatz (approved):
Thanks, @charbugs! I've backported your fix to HierarchyController.php into the release-9.0 branch so that fix can be incorporated into release 9.0.2. I'm merging the rest to dev so it will be part of release 9.1. Thanks again for your patience through all my feedback. :-)

PR #2825, opened 2023-04-11
Add example import config for SubjectsPlus.

demiankatz:
This adds SubjectsPlus import support, courtesy of @livedvd. I've refined it slightly and tested that it works correctly using data from a publicly-available SubjectsPlus installation: https://www.hksyu.edu.hk/lib/SubjectsPlus/oai/oai.php

### DIFF
--- a/import/xsl/subjectsplus.xsl
+++ b/import/xsl/subjectsplus.xsl
@@ -1,0 +1,183 @@
+<!-- available fields are defined in solr/biblio/conf/schema.xml -->
+<xsl:stylesheet version="1.0"
+    xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+    xmlns:oai_dc="http://www.openarchives.org/OAI/2.0/oai_dc/"
+    xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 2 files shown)


import/xsl/subjectsplus.xsl
```diff
@@ -0,0 +1,183 @@
...
+                    <xsl:value-of select="$institution" />
+                </field>
+
+                <!-- COLLECTION -->
+                <field name="collection">
+                    <xsl:value-of select="$collection" />
+                </field>
+
+                <!-- LANGUAGE -->
+                <xsl:if test="dc:language">
```
EreMaijala: I'd leave out the `xsl:if` surrounding `xsl:for-each` in each such case.

EreMaijala (changes requested):

import/xsl/subjectsplus.xsl
demiankatz: I've taken care of this. I didn't mess with the top-level if statements that control the multi-part vs. single-part indexing. I suspect that can be simplified too, but it somehow felt clearer to leave the parallel if's in place.

EreMaijala (approved):

PR #2826, opened 2023-04-11
Fix scoping of dc:subject in OJS import.

demiankatz:
I discovered this bug while working on #2825 -- when importing combined records, each record would include every subject heading from every record, instead of just its own subjects. I've fixed the scoping here and confirmed that it works correctly.

I think in practice this hasn't caused problems because OJS metadata typically doesn't include dc:subject anyway... but now, if it ever does, it will be handled correctly.

### DIFF
--- a/import/xsl/ojs.xsl
+++ b/import/xsl/ojs.xsl
@@ -73,8 +73,8 @@
                     </xsl:for-each>
                 </xsl:if>
                 <!-- SUBJECT -->
-                <xsl:if test="//dc:subject">
-                    <xsl:for-each select="//dc:subject">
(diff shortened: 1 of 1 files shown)


EreMaijala (approved):
Looks good to me! I believe the surrounding `xsl:if` isn't needed since there's nothing surrounding the set of topic fields, but since it's the prevalent style in the file, no need to change that here.

demiankatz:
Sounds good. Maybe it's worth a separate PR to simplify everything if you are indeed correct, but I'll defer that to another time. :-) Thanks!

PR #2827, opened 2023-04-11
Remove unnecessary if's from XSLT.

demiankatz:
This should clean up outstanding unnecessary XSLT if statements.


EreMaijala (approved):
Very nice, thanks!

PR #2824, opened 2023-04-11
Fix compact in AbstractRecord::homeAction

LuomaJuha:
compact(): Argument #1 must be string or array of strings, int given in .../Controller/AbstractRecord.php on line...

PR #2829, opened 2023-04-12
Add Mink tests for simple container links.

EreMaijala:
These tests also ensure that checkRoute and sid parameters are handled properly (see #2824).

### DIFF
--- a/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ContainerLinksTest.php
+++ b/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ContainerLinksTest.php
@@ -56,11 +56,11 @@ class ContainerLinksTest extends \VuFindTest\Integration\MinkTestCase
     }
 
     /**
-     * Test simple container linking without ID.
+     * Test default behavior of container links.
      *
      * @return void
      */
-    public function testSimpleContainerLinksWithoutID(): void
+    public function testDefaultContainerLinks(): void
     {
         $page = $this->performSearch('id:jnl1-1');
         $this->assertMatchesRegularExpression(
--- a/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ContainerLinksTest.php
+++ b/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ContainerLinksTest.php
@@ -1,0 +1,111 @@
+<?php
+
+/**
+ * Mink test class for simple container links functionality (including collection routing).
+ *
+ * PHP version 7
+ *
+ * Copyright (C) The National Library of Finland 2023.
+ *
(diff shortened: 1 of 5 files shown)


tests/data/journal_article.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<collection xmlns="http://www.loc.gov/MARC21/slim">
+  <record>
+    <leader>     nas  2200301 a 4500</leader>
+    <controlfield tag="001">jnl1</controlfield>
+    <controlfield tag="008">231104c2023    xx  x goo     0   a|eng||</controlfield>
+    <datafield tag="024" ind1=" " ind2=" ">
+      <subfield code="2">test</subfield>
+      <subfield code="a">journal1</subfield>
```
demiankatz: Should this be jnl1 rather than journal1? I'm not sure why the hierarchy ID would be different from the record ID in this context.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ContainerLinksTest.php
```diff
@@ -0,0 +1,111 @@
...
+        $session->visit($this->getVuFindUrl() . $path);
+        return $session->getPage();
+    }
+
+    /**
+     * Test simple container linking without ID.
+     *
+     * @return void
+     */
+    public function testSimpleContainerLinksWithoutID(): void
```
demiankatz: You don't have simple container links turned on in this test -- I'm wondering if the test name is misleading. What about testDefaultContainmentBehavior, with a comment like "Test handling of container fields when simpleContainerLinks is false (default behavior)."?

demiankatz (changes requested):
Thanks, @EreMaijala -- one minor suggestion, as well as a question about the test data.

tests/data/journal_article.xml
EreMaijala: It doesn't make a difference for this case, but I changed it for consistency.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ContainerLinksTest.php
EreMaijala: I changed it to testDefaultContainerLinks with the suggested comment. Does it work for you?

demiankatz (approved):
Thanks, @EreMaijala! I'll merge this as soon as CI passes.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ContainerLinksTest.php
demiankatz: Yep, that's great!

PR #2830, opened 2023-04-13
Remove trailing slashes from void elements.

EreMaijala:
This fixes HTML validation warning "Trailing slash on void elements has no effect and interacts badly with unquoted attribute values."

This also improves consistency.


demiankatz (approved):
Thanks, @EreMaijala, I'm happy to see our HTML markup made more consistent. I've carefully reviewed everything, and in an amazing turn of fate, I only found one problem -- in the VERY LAST FILE IN THE PR!

The themes/root/templates/search/opensearch-describe.phtml template is used to generate XML, not HTML, so the void element slashes ARE NEEDED there to produce valid output. I've simply reverted your changes to that one file, and everything else looks good.

EreMaijala:
@demiankatz Oh my! Thanks for checking so carefully!

PR #2832, opened 2023-04-13
Always overwrite existing files when copying node modules.

EreMaijala:
Since git doesn't preserve timestamps, the target may be never than the file downloaded from npm, and phing would skip copying over newer files without overwrite="true".


demiankatz (approved):
Makes sense -- thanks, @EreMaijala!

PR #2833, opened 2023-04-13
Fix cookie consent dialog z-index to be below dropdowns.

EreMaijala:
This ensures dropdowns and modals stay above and e.g. language can be changed.


EreMaijala:
I've added a feature request to make it possible to override the z-index without having to resort to `!important`here: https://github.com/orestbida/cookieconsent/issues/501

demiankatz (approved):
Thanks, @EreMaijala! Hopefully we can simplify this after a future upgrade. :-)

PR #2834, opened 2023-04-13
Enable NFC normalization of test records to avoid warnings.

EreMaijala:
Non-NFC UNICODE results in the following message from HTML validator in many tests: "Text run is not in Unicode Normalization Form C"


demiankatz (approved):
Seems safe enough to me, and all tests pass. I just added a comment to the file to explain why we're doing it, in case somebody wonders in future.

PR #2835, opened 2023-04-13
fix(truncate): prevent focus on hidden elements.

crhallberg:
- Hidden elements now have `tabindex="-1"` applied to them.
- Using IntersectionObservers, we add or remove the property as the truncation expands.

- [x] Discovered edge case where -1 is applied but not removed.


themes/bootstrap3/js/truncate.js
```diff
@@ -184,6 +184,24 @@ VuFind.register('truncate', function Truncate() {
...
       container.addClass('truncate-done');
 
+      // Make hidden elements unfocusable
+      // - Create IntersectionObserver
+      const root = container.get(0);
+      const observer = new IntersectionObserver(
+        (entries) => {
+          entries.forEach((entry) => {
+            if (entry.intersectionRatio > 0) {
+              entry.target.removeAttribute("tabindex");
```
demiankatz, line 194: Is this going to create problems if you truncate a region that contains explicit tab indexes? It seems to me like this will cause them all to get reset. Should we only remove the attribute if it equals -1? Or should we move any existing value to a data element so we can restore it exactly as it was?

themes/bootstrap3/js/common.js
```diff
@@ -294,6 +294,17 @@ function htmlEncode(value) {
   }
 }
 
+/**
+ * Keyboard and focus controllers
+ * Adapted from Micromodal
+ * - https://github.com/ghosh/Micromodal/blob/master/lib/src/index.js
+ */
+const FOCUSABLE_ELEMENTS = ['a[href]', 'area[href]', 'input:not([disabled]):not([type="hidden"]):not([aria-hidden])', 'select:not([disabled]):not([aria-hidden])', 'textarea:not([disabled]):not([aria-hidden])', 'button:not([disabled]):not([aria-hidden])', 'iframe', 'object', 'embed', '[contenteditable]', '[tabindex]:not([tabindex^="-"])'];
+function getFocusableNodes(container) {
```
demiankatz, line 303: Would it be better to plug this into the VuFind namespace container rather than creating a new global function? I'm okay with this as-is if there's no more logical place to put it, but I thought we were trying to avoid globals as much as possible.... maybe this is an opportunity to start a VuFind.util module or something.

demiankatz (changes requested):
Thanks, @crhallberg -- a couple of thoughts/suggestions. I'm also running the Mink tests now, so I'll let you know how that turns out once it's done! :-)

demiankatz:
All tests are passing, so that's a good start.

themes/bootstrap3/js/common.js
crhallberg: A VuFind.util would be good, but when I looked at the VuFind object yesterday, it seemed to be solely system things and not utility. So I put this here. This won't be a problem in the future of JS modules.

demiankatz:
Thanks for the progress on this, @crhallberg. Since the dev branch has officially gotten ahead of release 9.0 now (due to a mistake on my part -- oops), I'm rebasing this PR against the newly-created release-9.0 branch to keep it on track for the release. I'm also requesting a review from @EreMaijala in case he has any additional thoughts.

themes/bootstrap3/js/common.js
crhallberg: My comment wasn't clear: I don't think this function fits with the methods in the VuFind object, so I don't agree with moving it there. I do look forward to a future where we can get all of this out of the global scope!

EreMaijala (changes requested):
I might be missing something, but as far as I can see we already control all the elements we hide and show, so there should be no need for intersection observer. Instead we could just store tabindex during initial hide and then process it when toggling the elements. This would also ensure that we don't toggle tabindex for non-truncated elements even if they aren't currently visible for some reason.

EreMaijala:
I **was** missing something, so please disregard the above. I completely missed the height-based truncation not handling the individual elements.

EreMaijala (approved):
I still think that a VuFind.util class could be added, but I wouldn't block this based on that.

themes/bootstrap3/js/common.js
demiankatz: The VuFind object can register components -- I was just proposing creating a util component to hold this type of thing. I think there's existing precedent for that. But as @EreMaijala says, we don't have to do that right now. Maybe further JS reorganization can be a follow-on task for 10.0.

demiankatz (approved):
Just confirmed that the full test suite is passing -- this should be ready to merge now!

PR #2836, opened 2023-04-14
Update Vagrantfile for release 9.0.

demiankatz:
This PR updates the Ubuntu version in the Vagrantfile for compatibility with VuFind 9 dependencies. It also reorganizes some package names for consistency with the .deb file configuration, to make future synchronization/maintenance easier.

TODO
- [x] Test that this actually works

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -9,7 +9,7 @@ Vagrant.configure("2") do |config|
 
   # Every Vagrant development environment requires a box. You can search for
   # boxes at https://app.vagrantup.com/boxes/search
-  config.vm.box = "ubuntu/focal64"
+  config.vm.box = "ubuntu/jammy64"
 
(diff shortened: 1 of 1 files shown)


demiankatz:
I was able to test that this works correctly in Windows, though I had to add a note to the [Vagrant wiki page](https://vufind.org/wiki/installation:vagrant) indicating a permission-related problem where you need to run Vagrant as admin in order to be able to create symlinks (which is a prerequisite for running SolrMarc). There may be a more elegant workaround, but running Vagrant under Windows probably isn't the best idea to begin with, so I didn't waste a lot of time searching for nuance -- just wanted to confirm that it is theoretically possible to get it to work, which it is. :-)

EreMaijala (approved):
Looks good to me!

PR #2838, opened 2023-04-17
Final changes to make Mink tests pass HTML validation

EreMaijala:
This is the final set of changes required to make Mink tests pass HTML validation. Last commit is the slightly controversial one; The bootstrap3 popover adds a title element to popovers even if it's empty, and that causes a validation warning. We could of course add styling to hide the heading to avoid visible changes, but I couldn't decide what's the best course of action.


module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ListViewsTest.php
```diff
@@ -94,6 +94,9 @@ protected function gotoRecord()
      *
      * @retryCallback tearDownAfterClass
      *
+     * @skip_html_validation true
+     * @todo                 Enable HTML validation when the issues are fixed in the upstream code
```
demiankatz: Do you plan to open a JIRA ticket to cover this TODO?

themes/bootstrap3/js/cart.js
```diff
@@ -242,6 +245,7 @@ VuFind.register('cart', function Cart() {
     // Search results
     _registerUpdate();
     $("#updateCart, #bottom_updateCart").popover({
+      title: VuFind.translate('Information'),
```
demiankatz: I haven't had a chance to actually try this and see what it looks like yet, but might a more specific message be in order? Is it possible that "Book Bag" would be better than "Information"? (I guess it depends on whether we're trying to communicate the priority of the message, or the context of the message, or both...)

demiankatz (commented):
A question and a comment; I'll try to give this more attention tomorrow if time permits!

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ListViewsTest.php
EreMaijala: ~~Yes, I opened https://openlibraryfoundation.atlassian.net/browse/VUFIND-1606 before posting the PR. :)~~

themes/bootstrap3/js/cart.js
EreMaijala: Perhaps, I thought about it too, but somehow ended up with the vague "Information". I think it might make sense to also turn the messages into bullets if there are more than one. I'll test a bit.

EreMaijala: Title changed and bullets added when there are more than one message.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ListViewsTest.php
EreMaijala: No, sorry, that was for the other, and for this I failed to click the Create button. Now done as https://openlibraryfoundation.atlassian.net/browse/VUFIND-1607.

themes/bootstrap3/js/cart.js
```diff
@@ -201,7 +214,7 @@ VuFind.register('cart', function Cart() {
       }
       _cartNotificationTimeout = setTimeout(function notificationHide() {
         $('#' + elId).popover('hide');
-      }, 5000);
+      }, 50000);
```
demiankatz: I understand why you would increase this value, to ensure that the user has time to read the message even if it is very long. However, it would be better if the popover was dismissable. As far as I can tell, there is no way to make it go away after reading it, and having it floating there for 50 seconds potentially blocking other controls is annoying. Do you think we could easily add the ability to make the popover go away when the user clicks, or something similar?

demiankatz (changes requested):
I'm pretty happy with how this turned out, but I have one question...

themes/bootstrap3/js/cart.js
EreMaijala: Oops, that wasn't an intentional change, though 5 secs is low when there's a lot going on. Reverted now since that's something to think about another time.

demiankatz (approved):
Thanks, @EreMaijala, this looks good to me now. I'm going to ask for a review from @crhallberg as well in case he has any concerns I'm not thinking of.

I assume you don't mind letting this go until the 9.1 release, but let me know if you feel greater urgency about getting it merged into 9.0.

EreMaijala:
@demiankatz 9.1 is fine for me, no hurry!

themes/bootstrap3/js/cart.js
```diff
@@ -183,12 +182,26 @@ VuFind.register('cart', function Cart() {
...
+        if (msgs.length > 1) {
+          var ul = document.createElement('ul');
+          msgs.forEach((current) => {
+            var li = document.createElement('li');
+            li.innerHTML = current;
+            ul.appendChild(li);
+          });
+          msg = ul.outerHTML;
+        } else {
+          msg = msgs.pop();
```
crhallberg, line 204: This is a massive improvement. Thank you!

themes/bootstrap3/scss/components/search.scss
```diff
@@ -88,9 +88,19 @@ header .container.navbar { margin-bottom: 0; }
...
+    padding-left: 1em;
+  }
 }
 .action-toolbar li {
   display: inline-block;
+
+  .popover li {
+    display: list-item;
+    padding-left: 0.5em;
+  }
```
crhallberg, line 103: Do we want to apply this style to ul and li globally?

crhallberg (approved):

themes/bootstrap3/scss/components/search.scss
EreMaijala: I was afraid it could have unexpected consequences, and in other occasions like larger content areas the normal indentations could be fine. Part of these settings are these just to reverse the unstyled lists inherited from the toolbar, and part makes the indentation suit the small popover area better.

demiankatz: Yeah, seems safer to keep this more narrowly scoped, at least for now. In the long term, I think a broader CSS refactor is probably in order. :-)

demiankatz:
I've cherry-picked the first two commits from here back to release-9.0 since both are trivial fixes/improvements and might as well get released ASAP. I'll merge the remaining cart work to the dev branch to be on the safe side. Thanks again, @EreMaijala!

PR #2840, opened 2023-04-18
Improve npm script verbage

crhallberg:
Thanks, @aleksip! I hope this is a step in the right direction. I'd love your feedback on the `watch` verbage. And if there is a good prefix for tools like `lessToSass`. Also `lessdev`, I wanted to go with `dev:less` but that has connotations of watching and serving from other frameworks, so I went with `build:lessdev` for now. Maybe `dev:less` would be better, but how do we explicitly build or watch?

TODO
- [x] Update wiki instructions after merging

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,17 +12,18 @@
   "scripts": {
     "build": "npm run build:css",
     "watch": "npm run watch:css",
-    "dev": "npm run dev:css",
+    "build-dev": "npm run build-dev:css",
+    "watch-dev": "npm run watch-dev:css",
 
     "build:css": "npm run build:less",
     "watch:css": "npm run watch:less",
-    "dev:css": "npm run dev:less",
+    "build-dev:css": "npm run build-dev:less",
+    "watch-dev:css": "npm run watch-dev:less",
 
     "build:less": "grunt less && npm run lessToSass",
-    "build:lessdev": "grunt lessdev",
-    "dev:less": "grunt watch:lessdev && npm run watch:lessToSass",
-    "watch:less": "grunt watch:less",
-    "watch:lessdev": "grunt watch:lessdev",
+    "watch:less": "grunt watch:less && npm run watch:lessToSass",
+    "build-dev:less": "grunt lessdev && npm run lessToSass",
+    "watch-dev:less": "grunt watch:lessdev && npm run watch:lessToSass",
 
     "build:scss": "grunt scss",
     "watch:scss": "grunt watch:scss",
--- a/package.json
+++ b/package.json
@@ -12,12 +12,15 @@
   "scripts": {
     "build": "npm run build:css",
     "watch": "npm run watch:css",
+    "dev": "npm run dev:css",
 
     "build:css": "npm run build:less",
     "watch:css": "npm run watch:less",
+    "dev:css": "npm run dev:less",
 
     "build:less": "grunt less",
     "build:lessdev": "grunt lessdev",
+    "dev:less": "grunt watch:lessdev && npm run watch:lessToSass",
     "watch:less": "grunt watch:less",
     "watch:lessdev": "grunt watch:lessdev",
 
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "watch:css": "npm run watch:less",
     "dev:css": "npm run dev:less",
 
-    "build:less": "grunt less",
+    "build:less": "grunt less && npm run lessToSass",
     "build:lessdev": "grunt lessdev",
     "dev:less": "grunt watch:lessdev && npm run watch:lessToSass",
     "watch:less": "grunt watch:less",
--- a/package.json
+++ b/package.json
@@ -11,12 +11,13 @@
   },
   "scripts": {
     "build": "npm run build:css",
+    "watch": "npm run watch:css",
 
     "build:css": "npm run build:less",
     "watch:css": "npm run watch:less",
 
     "build:less": "grunt less",
-    "lessdev": "grunt lessdev",
+    "build:lessdev": "grunt lessdev",
     "watch:less": "grunt watch:less",
     "watch:lessdev": "grunt watch:lessdev",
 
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks, @crhallberg -- I've asked for reviews from @aleksip and @EreMaijala to make sure we get their attention. :-)

Regarding the dev question, would "builddev:less" be a possibility? It's more specific than just "dev" as a verb. Not sure if that's the right convention, though; just brainstorming!

I also wonder if lessToSass is a different type of "build:" action, perhaps? And maybe should the build:lessToSass action happen automatically somewhere in the stack (either as a second action of the default build action, or somewhere deeper in the chain)? Similarly, would it make sense for the "watch" action to run both "watch:css" and "watch:lessToSass" since right now we would really want to do both of those things to keep everything in sync?

In any case, I don't have particularly strong feelings about any of this, I just want to be sure we have a common understanding of our conventions going forward, whatever they should end up being.

Once this is approved and ready to merge, I'll update the wiki documentation to replace "npm run css" with "npm run build:css" (unless somebody feels we should be promoting a different action, or we land on a different final name).

demiankatz:
Small update: this PR was based against dev, but it needs to be based against release-9.0. I just rebased it for you.

EreMaijala (approved):
Looks good to me, but I'd defer to @aleksip for proper comments. :)

@demiankatz's idea about including lessToSass in the normal less build makes sense to me, but that could be a separate change.

aleksip:
Ah, naming things, hardest thing in the business!

I don't know if there are any 'right' ways to name these unless there is a technical reason, like there is for the colon, but I'm seconding all of @demiankatz's suggestions, which make sense to me. Or maybe `buildDev` would be slightly clearer and consistent with the camelCased `lessToSass`?

I guess this might be obvious, but just wanted to explicitly mention that there probably is not a real life use case for a separate `watchDev`, since watching is only needed in development...

demiankatz:
Thanks, @EreMaijala and @aleksip!

I agree that `buildDev` is more readable than `builddev` so I'll second @aleksip's suggestion there.

I imagine there _could_ be a use case for a `watchDev` -- in some situations you might want to be building to the production standards for testing purposes, to be sure things are not working in dev builds and broken in production builds, while most of the time I would expect you'd want to be building in dev mode for easier debugging. But that's probably a rarely-used distinction, and I'll leave it to @crhallberg's discretion whether or not to bother differentiating. We can always do it one way for now and revise later. :-)

In any case, I'll be happy to merge whatever @crhallberg ends up with by end of day today based on all this feedback! :-)

demiankatz:
I've gone ahead and merged some changes into the release-9.0 branch as a baseline for the 9.0 release -- mainly the changes to add the "build" script and change "css" to "build:css", etc.

This way, I can start to update the wiki documentation and be sure it will be accurate relative to the 9.0 release. But I think there are still some points related to the dev scripts and watch scripts, etc., that might benefit from further discussion here, so I'm leaving this open to continue those parts of the conversation and make further refinements.

Once we settle on a final strategy, we can decide whether to release it for 9.0.1, 9.1, etc.

crhallberg:
I think perhaps the best course of action is to add `dev:less` and have it be an alias for `watch:lessdev`. I'm not partial to the camelCase solution but this may be the best for both functionality and expectations.

Another change that I will likely make is to have `build:less` also run `lessToSass`. Should `dev:less` do this as well? It's a relatively quick operation.

demiankatz:
@crhallberg, I think if we're going to pair up the CSS building and SASS conversion, we might as well do it consistently across actions.

Regarding camel case, I don't have a strong feeling -- `buildDev` is more readable, while `builddev` is easier to type. I can live with either one. I think the more important detail here is that there may be value in having a different "verb" for a normal build vs. a dev build, so if in future we use the pattern of `npm run build:*` we don't end up building the dev and prod versions at the same time. Thus, I think having separate prefixes for separate categories of actions could be valuable.

EreMaijala:
How about `build-dev`? Different verbs alleviate the issue, but if we keep build for all..

crhallberg:
I've switched to `build-dev` and `watch-dev`. I think that makes the most sense. A future `dev` might be good to run all scripts but since it's just CSS, there's no need here.

I also have all of the less commands automatically run `lessToSass`. This is a very quick command so this should be fine.

package.json
```diff
@@ -11,14 +11,19 @@
...
     "watch:css": "npm run watch:less",
+    "build-dev:css": "npm run build-dev:less",
+    "watch-dev:css": "npm run watch-dev:less",
 
-    "build:less": "grunt less",
-    "lessdev": "grunt lessdev",
-    "watch:less": "grunt watch:less",
-    "watch:lessdev": "grunt watch:lessdev",
+    "build:less": "grunt less && npm run lessToSass",
+    "watch:less": "grunt watch:less && npm run watch:lessToSass",
```
demiankatz: I don't think this works -- don't you need to use something like concurrently to get multiple watches working simultaneously? I did a test with this and when I changed a file, the code compiled, but the lessToSass step did not occur.

demiankatz (changes requested):
Thanks, @crhallberg, I like the naming changes, but I think there's a problem with the implementation of watches.

package.json
crhallberg: That's a good point. Maybe I was hoping that the lessToSass would happen after the watch but that doesn't make sense. I'll skip the lessToSass here.

demiankatz (approved):
Thanks, @crhallberg! I think this is good forward progress. Ideally, it would be nice if the generic "npm run watch" ran all the watches, but since there's not a convenient way to do that without building more infrastructure, I think we can call it good enough for the moment.

demiankatz:
I've updated the [wiki page](https://vufind.org/wiki/development:npm) to match! Next step is to merge this into #2624 so we can resume work on further improvements when 10.0 development ramps up.

PR #2841, opened 2023-04-19
Avoid broken links being generated by Record View Helper

mtrojan-ub:


demiankatz (approved):
Thanks, @mtrojan-ub, this looks reasonable to me, and I've added a unit test to confirm the expected behavior. I'll cherry-pick this back to the release-9.0 branch tomorrow unless @EreMaijala has any concerns in the meantime.

EreMaijala (approved):
Looks good to me.

If we were to start from scratch, I'd avoid prepending anything in getCurrentHiddenFilterParams since its caller should know best what to prepend, and not doing it automatically would make it harder to make these mistakes. But this is good in keeping things back-compatible.

demiankatz:
Thanks, everyone -- as planned, I've backported this to release-9.0 so we can take advantage of the fix in Monday's release.

demiankatz:
To @EreMaijala's point, we could consider deprecating getCurrentHiddenFilterParams  in future in favor of a new method that eliminates the prepending functionality... but that's a conversation for another day. :-)

PR #2842, opened 2023-04-19
use headline for title in sidefacets

elsenhans:
This pull request is adding a h3-headline around the tiles in the sidebar with the layout remaining unchanged.
The titles do already act visually as headlines and should therefore be labeled as such.


demiankatz:
Thanks, @elsenhans -- I'm requesting a review from @crhallberg to see if he has any concerns or suggestions related to this change.

crhallberg (approved):

demiankatz (changes requested):
Thanks, @elsenhans.

The full test suite passes with these changes, which is good news.

However, this change has a significant impact on font size in the sidebar.

Before:
image

After:
image

Based on the style adjustments, this doesn't look intentional to me, and it definitely feels a bit excessive. Can/should this be changed further?

demiankatz:
I also wonder if perhaps a class (like `facet-title`) should be added to the new `<h3>` to allow more specific styling, just in case there is a different recommendation module that uses an `<h3>` inside the sidebar. That will reduce the odds of unwanted/unanticipated side effects.

elsenhans:
Thank you for your feedback @demiankatz !
Well, my adjustments in [sidebar.less](https://github.com/vufind-org/vufind/pull/2842/files#diff-843ae86e8e3a6096449acf6ecbb1eac467573ee355c73a66a5b5a72065d8efb0) and [sidebar.scss](https://github.com/vufind-org/vufind/pull/2842/files#diff-bd9097686a9a6d6cccead3168405d0203513aed91527d7ae165da42271a1a876) should prevent exactly that the h3-title is displayed larger than before. 
Did you run the scss- or lessdev-Grunt-task? 
If I do it is already looking with the h3 like this:
image

elsenhans:
> I also wonder if perhaps a class (like facet-title) should be added 

Sure, I can do that. But does it make sense to assign class names that are not in use yet? 
Or I should use the class name to add the styling just to that instead of all h3-elements in the sidebar?

demiankatz:
Thanks, @elsenhans -- I hadn't noticed that you hadn't committed the "compiled.css" file along with the LESS changes -- that definitely explains why I wasn't seeing what I expected. When changing LESS files in future, please commit the compiled assets along with the source changes to make the PR's complete. (I realize that it's somewhat redundant to keep these files in Git, but we do this so that developers can easily work with VuFind without having to set up Node.js if they don't need/want to).

Regarding the class name, I was suggesting that it might be safer to target something like `h3.facet-title` instead of `.sidebar h3` for this change, to avoid unwanted side effects. I see at least one existing recommendation module (SummonResultsDeferred) which can potentially put an h3 into the sidebar, and I don't think we want to impact that one.

elsenhans:
Sorry, did not know about commiting the `compiled.css`. Thanks for the hint!
Sure, I will do that and add the class name.

demiankatz:
No problem, @elsenhans -- live and learn. :-)

I'll give this another review once you've had a chance to push up your updates. Thanks again for the work on this!

demiankatz:
Thanks, @elsenhans! I'm focused on finishing up the 9.0 release packaging today, so I probably won't get to testing and merging this until early next week, but I appreciate the progress and I'll keep you posted.

demiankatz (approved):
Thanks again, @elsenhans! For some reason, it looks like only the bootstrap3 theme's compiled.css got pushed up here. I just went ahead and committed the other compiled files so that we can get this merged now. For future reference, please keep in mind that all the compiled.css files should be committed -- since most of the themes inherit from bootstrap3, when you change the less in that theme, it impacts the others. :-)

Thanks again for your ongoing work on accessibility -- it is greatly appreciated!

elsenhans:
Ok, thank you very much for your feedback @demiankatz !
I will add all the css files in the future.

PR #2843, opened 2023-04-20
Fix Lucene query string control character check

EreMaijala:
A lone `&` or `|` should not be considered a control sequence in Lucene, while `&&` or `||` should. Adds tests to verify correct behavior.


demiankatz (approved):
Thanks, @EreMaijala -- makes sense to me. Thanks for catching this _very_ old bug.

PR #2844, opened 2023-04-20
Restore access to item status methods and add save status methods.

EreMaijala:
Access to methods for checking item and save statuses for a container or a record were lost during refactoring. This restores the pre-existing methods for item statuses and adds similarly named ones for save statuses (not back-compatible but accessible).


EreMaijala:
I suppose that in most cases calling `VuFind.itemStatuses.init(container);` or `VuFind.saveStatuses.init(container);` works fine, but I think we should keep the methods that existed in 8.1 at least for item statuses.

demiankatz (approved):
Thanks for the catch, @EreMaijala! I'll update the changelog.

PR #2845, opened 2023-04-20
Add aria labels for control elements

elsenhans:
This pull request adds some aria-labels to buttons and input-fields. 
This is increasing the accessibility because a screen reader can read these aria-labels. Without them the control elements are not accessible by a screen reader.


themes/bootstrap3/js/channels.js
```diff
@@ -13,7 +13,7 @@ VuFind.register('channels', function Channels() {
     }
     var $cont = $(
       '<div class="dropdown">' +
-        '<button class="btn btn-link" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">' +
+        '<button class="btn btn-link" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true" aria-label="' + VuFind.translate('toggle_dropdown') + '">' +
```
demiankatz, line 16: Did you test that this translates correctly? I'm pretty sure it won't work unless the string is registered in the jsTranslations helper (typically done either in [js-translations.phtml](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/layout/js-translations.phtml) or in feature-specific code for more narrowly-scoped/rarely-used translations.

demiankatz (changes requested):
Thanks, @elsenhans! I just found one area where I think there may be a problem here.

I'm also curious: does aria-label really add value to submit inputs when it contains a value which is identical to the value of the input? I would think the screen reader would use the value if there was no aria-label defined. In the past, I've heard the advice that aria-label should be used sparingly if native HTML features do the same job equally well, but I'm not an expert, so maybe there is a strong reason to do this. Just curious!

I'm also requesting a review from @crhallberg in case he has further input.

themes/bootstrap3/js/channels.js
elsenhans: Thanks @demiankatz ! 
I missed that

elsenhans:
Seems like the screen reader reads both attributes, `value` and `aria-label` and therefore this is a duplication.
Sorry, seems we have not worked out a good solution here.

I would suggest to take out the `aria-label` for the elements having already a `value`-attribute again.
So there would just be two buttons affected where to add/keep the `aria-label`, one in the `header.phtml` and the other one in the `channelList.phtml`.

What do you think @demiankatz and @crhallberg ?

demiankatz:
Thanks, @elsenhans, I agree, we need to label cases where there is an icon inside a button with no text. The only question is how best to label these scenarios. I think there are at least three options:

1.) Put an aria-label on the button, as you have done here.

2.) Put an aria-label or alt text on the icon.

3.) Put "sr-only" text inside the button alongside the icon.

I have no idea which of these approaches is the "best practice," but we should probably be consistent in our approach. @crhallberg, do you have any thoughts/opinions on this?

elsenhans:
The `aria-label` attribute is strongly supported on interactive elements like links and buttons, the `alt` attribute is only meant for images.

If a screen reader were to vocalize visual indicators, like icons, it could cause a confusing experience for users. It is better to keep the icon for the screen reader hidden.

Thats why I would suggest something like this:
`<button aria-label='Short description'><i class="fa fa-icon" aria-hidden="true"></i></button>`

demiankatz:
@elsenhans, I agree that the alt approach is least desirable since icons are not always implemented as images. However, the other pattern we've used in some other parts of the code is "sr-only" text inside links, and I imagine a button could be implemented similarly. Here's an example:

https://github.com/vufind-org/vufind/blob/54b8e73767de526823d8851e0161fdabb0bcad8a/themes/bootstrap3/templates/eds/advanced.phtml#L32-L35

Again, I'm not saying that this approach is better than what you propose, I'm just trying to figure out if we can establish a pattern to make all of our code more consistent.

Also, a different thought: one drawback to both of these proposed approaches is that they _only_ benefit screen reader users. However, even sighted users might benefit from something like a mouseover effect to clarify what icons actually mean. Is there a strategy (like a well-placed title attribute) that might enhance the experience for ALL users, improving accessibility even more broadly?

elsenhans:
It would be great to establish a pattern and having more consistent code. I am looking forward to it :)
Well, I just learned today that `sr-only` may actually be visible or made be visible on some screens. That is why we prefer `aria-label`.

Beyond that, even with a icon having a description, the gain in knowledge for the visually impaired would not be higher as a button with clear function.

Adding a description for all users would be possible with a `title` attribute I guess. Like it is at the button for jumping to the sidebar https://github.com/vufind-org/vufind/blob/c2aefdbe84f80618f473b0dd712ef64d7384cd37/themes/bootstrap3/templates/myresearch/checkedout.phtml#L14

What do you think @ckaz , @crhallberg ?

demiankatz:
Thanks, @elsenhans, I do rather like the idea of a title on the button, but if @ckaz or @crhallberg thinks that's a bad idea, I think you're right that aria-label is the next best thing.

I also notice that the line of code you reference is missing a translate call on the title attribute! I've opened #2853 to get that fixed. Thanks for indirectly bringing this to my attention. :-)

ckaz:
> I do rather like the idea of a title on the button, but if @ckaz or @crhallberg thinks that's a bad idea, I think you're right that aria-label is the next best thing.

Dear all,

I'm not sure if I'm fully getting all the issues here but here goes:

I think, we won't need either title attributes or tooltips for buttons that announce themselves correctly.

We will need an accessible solution for icon buttons or pure icon interactive elements. The easiest, it seems, would be the title attribute, but this is widely discouraged. Apparently, support for the title attribute is patchy at best. [Heydon Pickering](https://inclusive-components.design/tooltips-toggletips/) argues that "the title attribute works reliably in screen reader software [ONLY] on form elements like <input>s. Even then, touch and keyboard users won't get to see the title message appear ..." 

For **icon buttons**, he therefore **suggests including textual labels**, which would be ideal but might cause issues with available space. That would be my favourite, too. A very charming solution -- if there is enough space. If there isn't, as in the facets sidebar, we'd need to come up with something else.

Buttons with clear, established icons, such as the "X" on the modal close button, won't need visible textual labels and could be equipped with an sr-only/visually hidden text or, as is currently the case, an aria-label. 

However, we would need to differentiate this "X" from the "X" for removing search elements in advanced search and from the "X" for excluding facets, or, if a different icon is used, announce what it does. Here, a tooltip or "toggletip" (see [Pickering](https://inclusive-components.design/tooltips-toggletips/#inclusivetoggletips) again) could be a workable solution because of the limited space.

This concept could then be extended to augment control elements such as the off-canvas sidebar button. Before we do, however, we should reconsider whether "Expand sidebar" does really provide much value to non-sighted or visually impaired users or whether a better wording would allow us to do without further explanation.

[Sara Soueidan](https://www.sarasoueidan.com/blog/accessible-icon-buttons/) also has an interesting take on accessible icon buttons.

So, to pick up Demian's notion that a default solution would be nice, I think we might need two defaults, depending on whether there is enough space for the element or not.

Sorry, if this makes it even more confusing.

I'd be keen on hearing @crhallberg 's opinion on that.

demiankatz:
Thanks, @ckaz -- you make a very good point that title's value is limited by the lack of touch/keyboard support, so I agree that investing in using it doesn't really make much sense. We need something better.

Also a good point that "expand sidebar" doesn't really mean anything to a user with a screen reader, since the information on the page is likely exposed independently from the visual layout of the screen. Is it possible that those elements should be aria-hidden rather than aria-labeled, if we can guarantee that they're not actually needed to expose anything in a screen reader scenario?

I also think, since there are some easy and obvious improvements in this PR mixed with some changes that clearly require more discussion/brainstorming, that I should spend some time cherry-picking the easy parts out of this, so we can get some benefit immediately and then spend more time discussing the rest. I'll let you know when I've had a chance to do that!

I'm also definitely interested in hearing from @crhallberg on all this!

demiankatz:
Actually, I think I misspoke -- when I spoke of cherry-picking bits and pieces, I was actually thinking of #2762. This PR should probably be polished up and merged all at once -- I think there are still some changes here that need to be reverted and/or discussed further, so I'll just wait and see how things develop. :-)

elsenhans:
OK, so I am going to revert first all the aria-label for the elements having already a value-attribute. 
Will do it tomorrow!

themes/bootstrap3/templates/channels/channelList.phtml
```diff
@@ -26,8 +26,8 @@
 <?php if (empty($token)): ?>
   <form action="<?=$this->url('channels-search')?>" class="channel-search form-inline">
     <?=$this->transEsc('channel_searchbox_label')?>
-    <input type="text" name="lookfor" class="form-control" value="<?=$this->escapeHtmlAttr($this->lookfor) ?>">
-    <input type="submit" value="<?=$this->transEscAttr('Submit')?>" class="btn btn-default">
+    <input type="text" name="lookfor" class="form-control" value="<?=$this->escapeHtmlAttr($this->lookfor) ?>" aria-label="<?=$this->transEscAttr('search_terms') ?>">
+    <input type="submit" value="<?=$this->transEscAttr('Submit')?>" class="btn btn-default" aria-label="<?=$this->transEscAttr('Submit') ?>">
```
crhallberg: This label is redundant since `<input type="submit">` elements already [use their value attribute for screen readers](https://russmaxdesign.github.io/accessible-forms/01-input-submit.html).

themes/bootstrap3/templates/channels/channelList.phtml
```diff
@@ -26,8 +26,8 @@
 <?php if (empty($token)): ?>
   <form action="<?=$this->url('channels-search')?>" class="channel-search form-inline">
     <?=$this->transEsc('channel_searchbox_label')?>
-    <input type="text" name="lookfor" class="form-control" value="<?=$this->escapeHtmlAttr($this->lookfor) ?>">
-    <input type="submit" value="<?=$this->transEscAttr('Submit')?>" class="btn btn-default">
+    <input type="text" name="lookfor" class="form-control" value="<?=$this->escapeHtmlAttr($this->lookfor) ?>" aria-label="<?=$this->transEscAttr('search_terms') ?>">
```
crhallberg, line 29: Would it be a better enhancement to use `placeholder` instead of `aria-label`? This would add a visual clue and a screen reader hint.

https://russmaxdesign.github.io/accessible-forms/accessible-name-input05.html

themes/bootstrap3/templates/channels/channelList.phtml
crhallberg: There's a few more `type="submit"` and `type="button"` in the myresearch forms farther down.

crhallberg (changes requested):
Thank you for this! There are a few points of redundancy (to the best of my knowledge). Gosh, I wish ARIA was clearer and consistent.

themes/bootstrap3/templates/channels/channelList.phtml
elsenhans: I think `placeholder` is not an equivalent alternative to `aria-label`, 
["placeholder text is not broadly supported across assistive technologies and not displayed in older web browsers" ](https://www.w3.org/WAI/tutorials/forms/instructions/)
But it seems there are different views on `placeholder` and accessebility, do you have more information about that @ckaz ?

crhallberg: I'd also like a third opinion. I'm disappointed to see that W3 haven't updated that page in three years! The page I linked led me to believe that placeholder was supported by screen readers [and browsers](https://caniuse.com/css-placeholder) but I don't have a lot of hands on experience, nor a setup that would let me test options myself.

ckaz: > I think `placeholder` is not an equivalent alternative to `aria-label`,
> ["placeholder text is not broadly supported across assistive technologies and not displayed in older web browsers" ](https://www.w3.org/WAI/tutorials/forms/instructions/)
> But it seems there are different views on `placeholder` and accessebility, do you have more information about that @ckaz ?

I'd stick with labels/aria-labels -- placeholders could be used as an option for the libraries who want them. 

Apart from what you both wrote, another case against placeholders is that placeholders are for people who _get_ them, only. To those who don't, the form may look as if it had already been filled in, which may cause confusion, which may cause distrust in the site/product etc. etc. ... In addition, if there are _required_ fields in the form, reviewing errors may also become more difficult if placeholders are there. They pose an UX issue while improving usability for some users at the same time -- in my opinion at least.
Cheers!

elsenhans: When I add a `placeholder` and start the Screenreader **Orca** with **Chrome** it is just reading the aria-label but not the placeholder. 
With **Firefox** and **Orca** it is reading both, aria-label and placeholder. 
It is just one screenreader, but for Orca it seems that aria-label works better and more reliable across different browsers.

demiankatz: If placeholder is being completely ignored in some situations, I agree that we probably need to stick with aria-label. What do you think, @crhallberg? Are any further adjustments needed, or should we merge this one?

elsenhans:
Are there further comments on this PR? Are any further adjustments needed? Please let me know if I can do anything else @crhallberg @demiankatz

demiankatz:
Thanks, @elsenhans -- just waiting on final review by @crhallberg.

crhallberg (approved):

demiankatz (approved):
Thanks, @elsenhans and @crhallberg -- merging now!

PR #2846, opened 2023-04-20
Fix Cookie SASS

crhallberg:
An improper regex and a misplaced @import caused issues when compiling SASS converted from LESS.


crhallberg:
Will reopen with proper branch basing.

PR #2847, opened 2023-04-20
Fix Cookie SASS

crhallberg:
A poorly constructed regex and a misplaced @import caused SASS errors when converting from LESS.


demiankatz (approved):
Thanks for the fixes, @crhallberg!

PR #2848, opened 2023-04-21
Add a phing task reset_setup for resetting CI environment.

EreMaijala:
reset_setup resets configuration, cleans caches and recreates the database.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -360,14 +360,7 @@
       </then>
     </if>
 
-    <!-- fail if a local config.ini already exists -->
-    <property name="local_config_ini" value="${localdir}/config/vufind/config.ini" />
-    <if>
-      <available file="${local_config_ini}" />
-      <then>
-        <fail>${local_config_ini} already exists</fail>
-      </then>
-    </if>
+    <phingcall target="fail_if_local_config_exists" />
 
     <!-- run extra startup action, if any -->
     <if>
@@ -404,7 +397,7 @@
     <!-- build and configure the requested database type -->
     <phingcall target="setup_database" />
 
-    <!-- start Solr (use restart in case of old PID files) -->
+    <!-- start Solr -->
     <phingcall target="setup_solr" />
 
     <phingcall target="import_biblios" />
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -402,38 +402,42 @@
     </if>
 
     <!-- build and configure the requested database type -->
+    <phingcall target="setup_database" />
+
+    <!-- start Solr (use restart in case of old PID files) -->
```
demiankatz: This comment should probably move inside the setup_solr task since it refers to an implementation detail.

build.xml
```diff
@@ -402,38 +402,42 @@
...
+
+    <phingcall target="import_biblios" />
+    <phingcall target="import_authorities" />
+    <phingcall target="build_alphabrowse" />
+  </target>
+
+  <!-- Setup basic configuration and run the install command -->
+  <target name="configure_install" description="configure and run installation">
+    <!-- fail if a local config.ini already exists -->
+    <property name="local_config_ini" value="${localdir}/config/vufind/config.ini" />
```
demiankatz, line 411: We're doing the local_config_ini check in two places. Should we refactor to a separate task, or is there a reason to repeat it?

build.xml
```diff
@@ -611,6 +637,36 @@ ${git_status}
     <exec command="git reset --hard" passthru="true" />
   </target>
 
+  <!-- Reset database and configuration of the demo setup -->
+  <target name="reset_setup" description="reset configuration and database of demo">
+    <property name="solr_pid_file" value="${localdir}/solr-8983.pid" />
+    <!-- fail if the system is already running -->
+    <if>
+      <not><available file="${solr_pid_file}" /></not>
```
demiankatz, line 645: This logic has been reversed. Do the comment and message need to be changed to match?

demiankatz (changes requested):
Thanks, @EreMaijala, this is going to save a lot of time! Just a few minor comments/questions...

build.xml
EreMaijala: Moved.

build.xml
EreMaijala: Refactored. This makes the output a bit less clean as it seems as if the subsequent output belongs to the check task, so I named it so that the output doesn't look so odd.

In any case both are needed as I want to ensure that any extra_startup_command doesn't run if config exists, and it's done before configuration and install, which also needs to be prevented if config already exists.

build.xml
EreMaijala: Yes, fixed.

demiankatz:
Thanks, @EreMaijala, this looks good -- I'll just do some hands-on testing before merging, which I might not get around to until next week due to my focus on 9.0 release preparation today. :-)

EreMaijala:
@demiankatz Thanks, and no problem.

demiankatz (approved):
This looks good to me. I even confirmed that you can use `reset_setup` in combination with the `dbtype` parameter to flip back and forth between MySQL and PostgreSQL setups.

PR #2849, opened 2023-04-21
Improve Mink FeedbackTest

EreMaijala:
…ies.

This fixes intermittent retries from the interval captcha Mink test. And a stray comment.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2850, opened 2023-04-24
Improve routing configuration to simplify URLs for help pages

mtrojan-ub:
I wonder whether we should change this route so the url() helper could be used in a cleaner way in many cases.


demiankatz:
@mtrojan-ub, could you tell me more about your motivation for this change? I'm certainly open to cleaning up our help links, but I think some adjustments are needed, and understanding your goals might influence my suggestions. :-)

If you want to avoid query parameters on help links, I would suggest leaving the existing /Help/Home route for backward compatibility, and instead adding a new Help/:page route. I don't think we want URL's looking like /Help/Home/Topic when we could just have /Help/Topic. Then the existing /Help/Home route could redirect to the new route for backward-compatibility (we don't want bookmarked help pages to break).

Alternatively, if you don't like having hard-coded query parameters in the templates, we could also simply clean up the calls to use the mechanism that sends an array of query parameters through the options argument, instead of using hard-coded values in the template. This might be more technically correct at the template level while not changing anything at a lower level.

I have no strong preference for either of these approaches, but I think if you want to make changes to help pages, we should choose one or the other. Let me know what you think!

demiankatz:
@mtrojan-ub, this is another one that I've merged dev into to bring us up to date. I think further discussion (see above) is still required when time permits.

mtrojan-ub:
@demiankatz: I think the initial reason was that we ran into some general problems a while ago where the url helper would output a route including a `?`, so appending `?` by hand seems to be dangerous, so it might be a good thing to avoid that. Another thing that we wondered in at this point => there are certain cases where a `_` parameter is added including a timestamp, and we didn't quite understand what this is for or whether this maybe could be added automatically if necessary.

demiankatz:
@mtrojan-ub, I agree that it's a good idea to reduce hard-coded query strings in templates, in favor of passing an array of parameters to the appropriate helper argument/option. If that's your main concern, maybe the quickest fix here, as suggested above, is simply to refactor the links to pass appropriate options instead of hard-coding parameters. Or we could make a cleaner-looking route, but in that case, we'd still need to maintain backward URL compatibility, etc., also as suggested above.

Regarding the `?_=[timestamp]` parameters, those are used to ensure that outdated assets never get cached in end user browsers -- the timestamp is the modification date of the file (I believe this is used on CSS and JS files). Thus, when a file changes, the URL also changes, and the browser automatically fetches a new version. I can't remember the implementation details behind that mechanism, but if it could be improved, I'm certainly open to refactoring those templates as well, time permitting.

Please let me know how you'd like to proceed, and if you need any more information or input from my end!

demiankatz:
Moving to the 10.1 milestone as discussed at today's Community Call.

mtrojan-ub:
@demiankatz, i think you're right with your suggestions. There should be a shorter route like e.g. /Help/:topic and we still need to maintain backwards compatibility, combined with reduced hard-coded query strings.

Regarding the ?_= timestamps i think this is dangerous since crawlers might try to crawl these pages over and over again with different query parameters, though the content will always be the same. Treating them as `rel="nofollow"` is also not the best solution since we might want them to be crawled once anyway. Maybe it would be better to use a meta html tag on the help page itself like e.g. `<meta http-equiv="cache-control" content="no-cache" />`?

demiankatz:
@mtrojan-ub, I haven't tried running this yet, but it looks like work still needs to be done to maintain backward compatibility to old help links. Am I missing something, or is that still on the "to do" list? Let me know if you need help with it.

Regarding the ?_ timestamps, these should only be getting applied to assets like Javascript and CSS. They should only change when the underlying files change. Since this only applies to Javascript and CSS files, adding meta tags to them would not be an option. I'm not sure I understand where/why/how this would apply to help pages anywhere. Are you actually seeing this somewhere, or is it a theoretical concern? (Sorry if I've lost track of part of this conversation -- it's been a while, so maybe I'm just confused and forgetful ;-) ).

mtrojan-ub:
@demiankatz, i just took some time yesterday to solve the latest merge conflicts, but wasn't able to apply the latest changes yet (still got some stuff on my local machine).

Regarding `?_=` there is at least 1 URL pointing to a help page (not a JS/CSS file!) which contains a timestamp (for topic = search). So maybe we should just remove the timestamp there to avoid further confusion.

mtrojan-ub:
The discussed changes have been implemented now, @demiankatz. What do you think?

themes/root/templates/HelpTranslations/de/blender_advsearch.phtml
```diff
@@ -5,7 +5,7 @@
   <dd>
     <p>Wenn Sie das erste Mal die erweiterte Suche aufrufen, erscheinen
        verschiedene Suchfelder. In jedem Feld können Sie einen Suchbegriff
-       eingeben. Folgende <a href="<?=$this->url('help-home', [], ['query' => ['topic' => 'blender']])?>">Suchoperatoren</a>
+       eingeben. Folgende <a href="<?=$this->url('help', [], ['query' => ['topic' => 'blender']])?>">Suchoperatoren</a>
```
demiankatz: This still seems to be using the query approach. Might be worth searching the templates for `['query' => ['topic'` to see if similar things were missed anywhere else.

module/VuFind/src/VuFind/Controller/HelpController.php
```diff
@@ -52,9 +52,14 @@ class HelpController extends AbstractBase
      */
     public function homeAction()
     {
+        $topic = $this->params()->fromRoute('topic');
+        if (empty($topic)) {
```
demiankatz: It might be a good idea to add a special case for `$topic === 'Home'` -- right now, things are working because the help-home route gets matched for /Help/Home, but if we eventually remove that legacy route but somebody continues linking to a legacy URL, I think it will end up coming here and resolving to a topic of 'Home' which will cause problems. I don't think it hurts to cover that scenario, just in case!

demiankatz (changes requested):
Thanks, @mtrojan-ub, this looks great and is nearly done -- just see below for a couple of comments. If you need help cleaning up links in the help files, let me know; I don't mind assisting if you're busy. :-)

module/VuFind/src/VuFind/Controller/HelpController.php
mtrojan-ub: I added a minor change, however i'm not sure whether that covers the case that you meant correctly. Please have another look.

demiankatz: What I had in mind was more like:

```
if (empty($topic) || $topic === 'Home') {
    $topic = $this->params()->fromQuery('topic');
}
```

If the route topic resolves to Home, then the real topic is in the query and we need to retrieve it.

You can test this scenario by accessing an old-style URL after removing the 'Help/Home' static route from the configuration.

mtrojan-ub: Understood, thanks! The code has been adjusted accordingly.

demiankatz (approved):
Thanks, @mtrojan-ub, this all looks good to me now!

PR #2779, opened 2023-03-14
Add support for purging loans from history.

EreMaijala:
- Moves loan history functionality to a newly introduced CheckoutsController.
- Adds purge support to Demo and SierraRest drivers.

Note: New translation strings!

TODO:
- [x] Add mink test
- [x] Update driver spec before merging
- [ ] Update TODO list on #2723 if this gets merged first (or update this code if #2723 merges first)

PR #2852, opened 2023-04-25
Add header and h2 for title and author

elsenhans:
This PR is adding a `h2` element for the titles in all lists (result-list, saved items and all other lists available in the account). 
Also the _author_ is connected to the _title_ by an `header` element.

Since we only do have data for the _Default-RecordDriver_, I am not able to access the result-lists in the other RecordDrivers.


themes/bootstrap3/templates/RecordDriver/DefaultRecord/result-list.phtml
```diff
@@ -20,19 +20,21 @@
...
+          <?php foreach ($this->driver->tryMethod('getTitlesAltScript', [], []) as $altTitle): ?>
+            <div class="title-alt">
+              <?=$this->escapeHtml($altTitle)?>
+            </div>
+          <?php endforeach; ?>
+        </h2>
 
-      <div>
         <?php if ($this->driver->isCollection()): ?>
+          </header>
```
demiankatz: This close tag is inside a condition, but there's also a similar close tag AFTER the end of the if statement. I don't think this is correct, but maybe I'm misreading something.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
```diff
@@ -33,36 +33,37 @@
...
+                <?=$this->escapeHtml($altTitle)?>
+              </div>
+            <?php endforeach; ?>
+          </h2>
 
-        <div class="resultItemLine2">
           <?php if ($this->driver->isCollection()): ?>
-            <?=implode('<br>', array_map([$this, 'escapeHtml'], $this->driver->getSummary())); ?>
+            </header>
+            <div class="resultItemLine1">
```
demiankatz: We seem to be somewhat redefining the use of the resultItemLine1 and resultItemLine2 classes by moving things around here. Did you investigate whether these are used anywhere in existing styles/code? If not, and if we're willing to break backward compatibility in cases where people have introduced local custom styling against these classes, I wonder if we should revisit the whole naming concept behind them.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
```diff
@@ -33,36 +33,37 @@
...
               <?php $authorCount = count($summAuthors); foreach ($summAuthors as $i => $summAuthor): ?>
                 <a href="<?=$this->record($this->driver)->getLink('author', $summAuthor)?>"><?=$this->escapeHtml($summAuthor)?></a><?=($i + 1 < $authorCount ? ';' : '') ?>
               <?php endforeach; ?>
             <?php endif; ?>
-
+            </header>
+            <div class="resultItemLine1">
             <?php $journalTitle = $this->driver->getContainerTitle(); $summDate = $this->driver->getPublicationDates(); ?>
             <?php if (!empty($journalTitle)): ?>
-              <?=!empty($summAuthor) ? '<br>' : ''?>
```
demiankatz, line 61: I notice that some line breaks are getting removed here (and in some other places as well). Is this intentional? Have its possible consequences been tested/inspected?

demiankatz (changes requested):
Thanks, @elsenhans. I defer to @crhallberg and others should they have opinions about the general strategy here (use of header/h2 elements). However, I just wanted to start the conversation off by highlighting a couple of specific details I had thoughts or questions about.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/result-list.phtml
elsenhans: I think that closing tag belongs to another `if`. The closing `header` is inside the `if` and the `else` condition, like:
```
<?php if ($this->driver->isCollection()): ?>
  </header>
  <div>
<?php else: ?>
  <?php ... if (!empty($summAuthors)): ?>
    <?php foreach ($summAuthors as $i => $summAuthor): ?>
      ...
    <?php endforeach; ?>
  <?php endif; ?>
  </header>
  <div>
```

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
elsenhans: Well, I didn't find the class names in any styling files or visual differences. Seems like only the lines are numbered by it.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
elsenhans: Yes, that was intentional, because a line break is already done by the header tag. Keeping the breaks would create empty lines.

ckaz:
Dear all,
our accessibility testers (who are based at the German Library for the Visually Impaired) did indeed make a strong case in favour of using `<header>` elements to indicate important content belonging together and "heading" the information that follows. In effect, all lists would need the `<header>` element wrapping titles and imo. at least authors (one could make a case that even dates would be relevant header content).
Cheers, Claas

themes/bootstrap3/templates/RecordDriver/DefaultRecord/result-list.phtml
demiankatz: Sorry, I see that you're right! However, I wonder if it would be more readable and easier to understand/maintain if we eliminated the duplicate close tag by refactoring so it's more like:

```
if (!$this->driver->isCollection()): ?>
  <?php ... if (!empty($summAuthors)): ?>
  ...
  <?php endif; ?>
<?php endif; ?>
</header>
<div>

<?php if ($this->driver->isCollection()): ?>
  ...
<?php else: ?>
  ...
<?php endif; ?>
```

Do you see what I mean? It adds an additional if statement to the template, but it makes the overall flow of the logic a little easier to follow.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
demiankatz: I'm pretty sure you're correct that these classes are not used in the core code, but I think some institutions use them to apply local customizations... so I'm concerned that if we change how they are used, it will create weird side effects, and we might be better off eliminating them or replacing them with more descriptive names rather than moving them around. Maybe @crhallberg can weigh in on this issue.

demiankatz:
Thanks, @ckaz. Just a note that I made a small edit to your comment to put backticks around `<header>` since GitHub was not displaying the tag otherwise. :-)

themes/bootstrap3/templates/RecordDriver/DefaultRecord/result-list.phtml
elsenhans: Yes, I agree, that would be easier to read! 
I will adjust it.
Will do it tomorrow!

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
```diff
@@ -33,36 +33,37 @@
...
+          <h2>
+            <?php $missing = $this->driver instanceof \VuFind\RecordDriver\Missing; ?>
+            <?php if (!$missing): ?><a href="<?=$this->escapeHtmlAttr($this->recordLinker()->getUrl($this->driver))?>" class="getFull" data-view="<?=$this->params->getOptions()->getListViewOption() ?>"><?php endif; ?>
+              <span class="title"><?=$this->record($this->driver)->getTitleHtml()?></span>
+            <?php if (!$missing): ?></a><?php endif; ?>
+            <?php foreach ($this->driver->tryMethod('getTitlesAltScript', [], []) as $altTitle): ?>
+              <div class="title-alt">
+                <?=$this->escapeHtml($altTitle)?>
+              </div>
+            <?php endforeach; ?>
```
crhallberg, line 46: Do we want all of the alts inside of the h2?

elsenhans: Well, the title should be in the h2 and these conditions belong to it, right?

themes/bootstrap3/templates/RecordDriver/DefaultRecord/result-list.phtml
elsenhans: Sorry, but it is not possible to eliminated the duplicate close tag `</header>`. 
The point is, that we can't move the closing tag outside of that condition:
`<?php if (!$this->driver->isCollection() ?>`
The closing tag has to be after `$summAuthors` but there are following $journalTitle, $summDate abd other data which is rendered in the same condition.
We would need something like this, using the same if-query twice:
```
<? if (!$this->driver->isCollection()): ?>
  <?php ... if (!empty($summAuthors)): ?>
  ...
  <?php endif; ?>
<?php endif; ?>
</header>
<div>
<? if (!$this->driver->isCollection()): ?>
<?php if (!empty($journalTitle)): ?>
...
<?php elseif (!empty($summDate)): ?>
...
<?php endif; ?>

<?php if ($this->driver->isCollection()): ?>
  ...
<?php endif; ?>

```

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
demiankatz: The variable/class naming may be misleading here. Based on a quick glance at the code, I thought these represented "alternate titles" in which case I would agree with @crhallberg that they do not belong inside the h2. However, I see on closer examination that these are actually alternate script forms for the primary title -- e.g. Chinese characters to accompany a Romanized version of the same text. In that scenario, I agree with @elsenhans that it is appropriate to include them in the heading. So I think this is fine as it is.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/result-list.phtml
demiankatz: @elsenhans, I looked at this again, and I still think my proposed solution will work. Maybe I just abbreviated it too much to be clear. Here is the refactored code I'm suggesting:

```
      <header>
        <h2>
          <a href="<?=$this->escapeHtmlAttr($recordLinker->getUrl($this->driver))?>" class="title getFull" data-view="<?=isset($this->params) ? $this->params->getOptions()->getListViewOption() : 'list' ?>">
            <?=$this->record($this->driver)->getTitleHtml()?>
          </a>
          <?php foreach ($this->driver->tryMethod('getTitlesAltScript', [], []) as $altTitle): ?>
            <div class="title-alt">
              <?=$this->escapeHtml($altTitle)?>
            </div>
          <?php endforeach; ?>
        </h2>

        <?php if (!$this->driver->isCollection()): ?>
          <?php $summAuthors = $this->driver->getPrimaryAuthorsWithHighlighting(); if (!empty($summAuthors)): ?>
            <?=$this->transEsc('by')?>
            <?php $authorCount = count($summAuthors); foreach ($summAuthors as $i => $summAuthor): ?>
              <a href="<?=$this->record($this->driver)->getLink('author', $this->highlight($summAuthor, null, true, false))?>" class="result-author"><?=$this->highlight(rtrim($summAuthor, ','))?></a><?=$i + 1 < $authorCount ? ',' : ''?>
            <?php endforeach; ?>
          <?php endif; ?>
        <?php endif; ?>
      </header>
      <div>
        <?php if ($this->driver->isCollection()): ?>
          <?=implode('<br>', array_map([$this, 'escapeHtml'], $this->driver->getSummary())); ?>
        <?php else: ?>
          <?php $journalTitle = $this->driver->getContainerTitle(); $summDate = $this->driver->getPublicationDates(); ?>
```

There is only one bit of variant logic before the close header tag in both branches of the existing if..else, so this just creates a new if statement to move that piece higher in the code, introduces the close header and open div tags in between so we only define those once, and then keeps the rest of the existing if..else. Does that make sense?

elsenhans: Sure, I can do that. Meaning that `else` (from "keeps the rest of the existing if..else")  will be the same case like `<?php if (!$this->driver->isCollection()): ?> `
I will change it right now.

demiankatz: Yes, exactly! Thanks!

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
crhallberg: It seems that these two classes and `.title-alt` are here for customization, but I am surprised none of the three have any core styles. I'm not too worried about customizations, since they would be obvious and easy to fix, but it's not a non-factor.

demiankatz: My thought is mainly that I don't really like the resultItemLine1 / resultItemLine2 class names, as they're not very descriptive... and I'm not sure if they're even true after this refactoring, since resultItemLine1 now comes after the first line of the entry. Maybe we're better off eliminating them or thinking of better names if we're breaking backward compatibility anyway.

crhallberg:
In order to get the accessibility benefits, do we need to [wrap each section with a new header](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header#accessibility) with a particular semantic container like `<section>` or `<article>`? 

Otherwise, they will be considered global `banner`s (ARIA role), which [it's recommended to only have one of](https://webaim.org/techniques/semanticstructure/#regions).

ckaz:
> In order to get the accessibility benefits, do we need to [wrap each section with a new header](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header#accessibility) with a particular semantic container like `<section>` or `<article>`?
> 
> Otherwise, they will be considered global `banner`s (ARIA role), which [it's recommended to only have one of](https://webaim.org/techniques/semanticstructure/#regions).

Not as far as I know. We use them according to W3C which quotes:

Definition and Usage
The `<header>` element represents a container for introductory content or a set of navigational links.

A `<header>` element typically contains:
    one or more heading elements (`<h1>` - `<h6>`)
    logo or icon
    authorship information

Note: You can have several `<header>` elements in one HTML document. However, `<header>` cannot be placed within a `<footer>`, `<address>` or another `<header>` element.

https://www.w3schools.com/tags/tag_header.asp

Cheers!

crhallberg:
It looks like we do need to wrap the results in a semantic container, otherwise they all get flagged as "banner" which is "Semantically equivalent to `<header>`". Now I know we're using header but we want to use it locally, not globally.

image

elsenhans:
Here it says: "If a page includes more than one banner landmark, each should have a unique label."
https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/banner.html

This should be fulfilled by the record title or do we need a further addition to this?

demiankatz:
According to @elsenhans link, the header should only be treated as a banner if it is a direct descendent of body... but in VuFind, it should be contained inside a `<div role="main">` which according to the documentation should prevent the banner behavior. I'm not sure why @crhallberg is seeing different behavior. Could this be a problem with the tool being used rather than the markup? If not, we should investigate further to figure out the discrepancy.

elsenhans:
When I change the `<div role="main">` to  `<main role="main">` the landmark navigator does not mark the headers in the reslut-list as banners no longer. But I am not sure why, because in both cases it is recognizing the main landmark.
Was there a reason to use the role attribute instead of the HTML tag?
But role and tag should actually do the same thing.

image

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
elsenhans: So I understand that I should remove the `class="resultItemLine1"` here, right @demiankatz ?

demiankatz: I just did a search of the code to see where resultItemLine1 is used, and it is present in a number of templates... It is also used in some of our integration tests to identify links. I'm pretty sure that moving or removing it in this template will actually cause tests to fail.

In the long term, I would like to replace this with something more semantically meaningful, and make sure it is applied consistently across different templates. However, that's probably out of scope for this pull request.

Would a better solution be to move the resultItemLine1 class up to the `<header>` element, since that IS now the first line of the result? I believe doing that will at least prevent tests from failing (since the tests are looking for links inside the resultItemLine1 area in order to click on favorites).

demiankatz:
@elsenhans, do we need `<main role="main">` or could we just use `<main>`?

@crhallberg knows the history better than I do, but I wonder if perhaps we used div with role due to lack of support for main many years ago. I don't think that code has changed in a long time, and maybe using more precise semantic tags would be an improvement at this stage. Maybe that's worth doing in a separate PR to make incremental forward progress and enable us to proceed with this one.

elsenhans:
@demiankatz `<main>` seems to be enough for not marking the result-list headers as banners anymore.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
elsenhans: @demiankatz I did add `class="resultItemLine1"` to header and `class="resultItemLine2"` to the following line

ckaz:
Dear all, I was certain that when `<header>` elements are used within other landmarked sections such as `<main>` _or_ container with `role="main"` (not both) or `<article>`, they _should not_ be recognized as banners, so I think we're good here.

demiankatz:
@ckaz, it seems that both @elsenhans and @crhallberg have found evidence that `<div role="main">` is not being recognized correctly in some circumstances, though I agree with you that it _should be_. As noted earlier, this suggests to me that it might be worth building a separate PR to use more semantic tags (at very least, `<main>`, but maybe there are some others that could be streamlined as well). Is anyone interested in doing this?

As for this PR, I see that some conflicts have accumulated. @elsenhans, can you resolve them? Please let me know if you need help with that. I suspect the historicloans.phtml conflict is related to the recent renaming of that file in #2779. All the compiled.css conflicts can be resolved by just running `npm run build:css` after the `git merge` command and then adding/committing the results.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/list-entry.phtml
demiankatz: Thanks, @elsenhans, that's great! I'm going to be working with @sturkel89 later today to run tests and do some visual comparisons of the new and old code, so we'll let you know if any feedback or problems emerge from that process.

sturkel89:
I just ran the full test suite and had these failures:
```
There were 2 failures:

1) VuFindTest\Mink\ContainerLinksTest::testDefaultContainerLinks
Element not found: .result-body div a index 1
Failed asserting that null is of type "object".

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:387
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ContainerLinksTest.php:71
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:100

2) VuFindTest\Mink\ContainerLinksTest::testSimpleContainerLinksWithID
Element not found: .result-body div a index 1
Failed asserting that null is of type "object".

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:387
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ContainerLinksTest.php:101
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:100

FAILURES!
Tests: 2339, Assertions: 11939, Failures: 2.
```

demiankatz:
Thanks, @sturkel89! It looks like we'll need to fix the CSS selectors in `module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ContainerLinksTest.php` to reflect the new markup. @elsenhans, let me know if you'd like help with that! If you're not familiar with the test suite, I'm willing to assist. :-)

elsenhans:
Sorry, it seems i have accidentally merged the dev-branch in this branch :/
I will try to solve that as soon as possible.

demiankatz:
@elsenhans, this PR is targeted against the dev branch now, so it should be fine to merge upstream/dev into it. I notice there are some changes here that seem unrelated to your changes, but hopefully a clean merge will sort it all out. Let me know if I can help!

elsenhans:
Thank you @demiankatz and @sturkel89 !
Seems the conflicts are solved. 
Are there still tests failing?
Anything else I should adapt for the PR itself?

demiankatz:
Thanks, @elsenhans! This looks much better.

I was able to fix the broken tests that @sturkel89 identified in commit efc0f6c61124c6b68eca4b02cb8d2c4d9f89f429.

@sturkel89, do you mind doing a little bit of hands-on testing to compare the dev branch against this PR in these specific areas:

  * Search results
  * Favorites
  * Checked out items
  * Holds
  * ILL requests
  * Storage retrieval requests
  * Transaction history (see [this wiki page](https://vufind.org/wiki/development:testing:manual_testing#transaction_history) for instructions on how to access this feature)

All of the changes here have to do with how records are displayed in lists, so it should just be a matter of checking the various lists in the various themes and reporting any anomalies you notice. (I don't expect that you should see anything, but just hoping for a final double-check).

crhallberg:
Sorry to not be part of the conversation yesterday but as far as I know, [`<main>` should also prevent `<header>` from being a banner](1). And it shouldn't need a `role="main"`.

It occurs to me now that we can set the role to whatever we want. A header is either a `banner` or `generic`, if encased as we've been talking about. Which makes me confused about what the benefits of `<header>` are outside of being a page banner. [Like, why does HTML have this special case if it's equivalent to any other generic?](1)

[1] https://www.w3.org/TR/html-aria/#el-header

elsenhans:
@crhallberg The benefit is in this case, is that the `<header>` element represents a container for introductory content for the listed records. The `<header>` elements indicate important content belonging together (title and author) and "heading" the information that follows.

Thanks @demiankatz for fixing the test!

sturkel89:
I have tested this branch against the dev branch in all three themes, looking at all of these areas:

- Search results
- Favorites
- Checked out items
- Holds
- ILL requests
- Storage retrieval requests. 

Results lists in each of these areas in bootstrap3 and booprint3 look the same in dev and the PR branch. 

However, in the sandal theme, the PR branch shows shading behind the text in the author and title areas. This happens in all of the areas in VuFind listed above.

image

image

image

demiankatz:
Thanks, @sturkel89! There must be some `<header>` specific styling in the sandal theme that is having unintended consequences here. @crhallberg, can you take a look or make some suggestions?

crhallberg:
I don't have permission to edit this branch directly, but here's the changes I made. I can open a PR on their repository if that's easiest. [banner.patch](https://github.com/vufind-org/vufind/files/11480472/banner.patch)

demiankatz:
Thanks, @crhallberg! If @elsenhans doesn't want to directly incorporate this into this PR, another option might be for you to open this as a PR directly against dev, since using a more specific class for the Sandal styling could be viewed as an independent improvement.

elsenhans:
Sorry, i didn't understand it that way, please link and ask me directly in the future ;)
I did commit the banner.patch. @crhallberg would you please check if I did not miss anything? Thanks!

demiankatz:
Thanks, @elsenhans!

@sturkel89, can you give this another look when you have a chance? The only change should be a fix to the sandal issue you observed.

sturkel89:
Note: the CSS for the entire Sandal theme wasn't right, so with the help of @demiankatz I used _grunt less_ which fixed it. Sandal is now displaying VuFind properly on the test branch.

I went through and checked search results, list items, favorites, held items, etc. on the test branch in Sandal, and everything displays properly, with no extra shading behind author and title.

demiankatz:
Thanks, @sturkel89, I've pushed up a fixed version of the Sandal CSS based on your comments. @elsenhans, any idea why the Sandal theme's compiled.css was garbled? Did you run `grunt less` yourself or did it perhaps get messed up by the patch file somehow? In any case, I think we're in good shape now. I'll wait for your reply and @crhallberg's approval before merging.

elsenhans:
Thank you for fixing that @demiankatz ! I was just using the patch. Next time I will generate the compiled.css myself.
Also thanks a lot for your help @sturkel89 and @crhallberg !

demiankatz:
I just spoke with @crhallberg, and I think he's still uncertain of the benefit of the header tags. I'm wondering if perhaps @pasitiis could take a look at this and offer an opinion. I'm happy to merge if we have a consensus that it's a positive change, but I want to be sure there's general agreement on this.

ckaz:
Dear all,
I'm still convinced it is a good thing, because it brings important information on each item together semantically by wrapping them in `<header>`. That for me is the key argument, and it is in line with what our accessibility testers bemoaned.
Cheers! Claas

Just had another thought on this.

For me the key argument is that we could improve accessibility by grouping the most important information in those lists together, and marking them as headers for the respective items. Considering the different options of result list presentations, e.g. as accordeons, we might even consider wrapping each result in `<article>`, no?

ckaz:
Dear all, we are currently getting back to the accessibility testing group who made the original recommendation. We will ask them for feedback whether their recommendation would still be valid in the light of the issues voiced in this PR.  Cheers!

demiankatz:
Thanks, @ckaz! I'm also hoping to discuss this on tomorrow's Community Call to see if we can get any further input.

Also please note that it appears that there are quite a few conflicts here as a result of other recent changes. Let me know if any help is needed to get them resolved!

crhallberg (approved):

crhallberg:
Happy to go with @ckaz's accessibility testing group's recommendations!

ckaz:
Sorry, getting this checked took longer than we expected:

@elsenhans is on leave right now, so let me summarize the feedback we got.

Here's the testing authority's verdict: The **recommendation to wrap author and title into a `<header>` was originally made for a catalog where the author name preceded the medium's title**, like so:

```
<div class="result-body">
  <div class="author-names">
    <a href="..." class="author">Stordeur, Danielle</a> (VerfasserIn),  
    <a href="..." class="author">Helmer, Daniel</a> (VerfasserIn) 
  </div>

  <h2>
    <a id="..." href="..." class="title getFull" data-view="full" lang="">Jerf el Ahmar : un nouveau site de l'horizon PPNA sur le moyen Euphrate syrien</a>
  </h2>
...
</div>
```

The testing team reviewed their **earlier recommendation** **adding** the wrapping **`<header>` or `<hgroup>`** elements with and without roles and **they _could not_ find any improvement in the screen reader output**. 

The **new recommendation** for such cases is:

1. wrap authors and title in the same `<h2>` element and use `<span>` for styling OR
2. move author names below the `<h2>` title.

So **for this PR and the default VuFind**, I think we can **remove the `<headers>` from the lists again**. Sorry for the hassle and all the extra work.

demiankatz:
No worries, @ckaz, thanks for taking the time to follow up (and thanks to @crhallberg for observing the opportunity to simplify).

Is @elsenhans able to work on the simplification, or do you need more help from our end?

elsenhans:
Thanks @ckaz for taking care!
@demiankatz  I would like to create a new PR with the new requirements. I hope this is okay for everyone?

demiankatz:
@elsenhans, if you want to redo things, you can either close this PR and open a new one, or you can force-push a fresh set of changes to this existing branch; whatever is easier for you!

elsenhans:
Thanks @demiankatz !
I will close here and open a new PR as soon as possible ;)

elsenhans:
Like Claas did mention above, do we have this two possibilities for the new PR:

1. wrap authors and title in the same `<h2>` element and use `<span>` for styling OR
2. move author names below the `<h2>` title.

Do you have any preferences @demiankatz , @crhallberg ?
For me, the solution with the `<span>`-element seems to be the better solution, without changing the ordering of elements.

demiankatz:
@elsenhans, I don't have a strong preference, but I do agree that if we can find a suitable solution that does not require reordering of elements, that's likely to be less potentially disruptive.

elsenhans:
Thanks @demiankatz !

crhallberg:
@elsenhans wrap authors and title in the same `<h2>` element and use `<span>` for styling

I think this would be the more accessible option, since a list of all the headers would include authors as helpful and not overly verbose data. As long as h2 is the appropriate level for search results.

elsenhans:
Thanks for your feedback @crhallberg !
I will open a new PR with that solution.

PR #2853, opened 2023-04-25
Add missing title translations.

demiankatz:


crhallberg (commented):
This is a good change - if we want to keep these labels. Without them the link will just read "Your Account" which is accurate. I don't know if "Expand Sidebar" is better, since it doesn't say where we're going and screen readers don't really mind if content is just off the screen.

A more important change might be to make sure the offcanvas bar expands if focus moves into it.

demiankatz:
Good point, @crhallberg -- based on other recent discussion, title attributes aren't necessarily desirable, and this one doesn't necessarily convey any useful information since as you say, the link text describes what the link actually leads to.

If you search the code for the correct translation (sidebar_expand) you'll find that there are also three other templates that have variations on these theme, though they're implemented inconsistently with one another. We should really normalize all of this!

I think my inclination would be to merge this PR into the release-9.0 branch, since what I'm doing here is clearly a bug fix -- I'm translating text that is supposed to be translated but which was overlooked, to prevent people from encountering the wrong language. Using the correct translation string will also make it easier to find all instances of this pattern in a single search. Then we can figure out how to change this in the dev branch, since that's going to be more than a bug fix, and probably should wait until 9.1 rather than 9.0.2.

Once this one is merged, maybe you (or one of our recent accessibility contributors like @elsenhans, @jpkanter or @stephanieleary) would be interested in starting a follow-up PR to propose an improvement to the situation.

crhallberg (approved):
Makes sense to me.

PR #2854, opened 2023-04-26
Remove deprecated CSP directive prefetch-src

sipek-mzk:
Remove deprecated directive `prefetch-src` form `config/vufind/contentsecuritypolicy.ini`.
For more info see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/prefetch-src


demiankatz (approved):
Thanks, @sipek-mzk!

PR #2855, opened 2023-04-27
Fix translation of patron blocks in holdings tab.

EreMaijala:
Blocks were translated when displayed in e.g. the profile page but not when they were displayed as part of the message in the holdings tab.


demiankatz (approved):
Thanks, @EreMaijala -- I'll cherry-pick this to the release-9.0 branch.

PR #2856, opened 2023-04-27
Sierra: Add mapping of patron block codes to translatable strings.

EreMaijala:
The blocks are returned as single character codes. There are no defaults, so every library needs to define their own mappings.

### DIFF
--- a/config/vufind/SierraRest.ini
+++ b/config/vufind/SierraRest.ini
@@ -166,6 +166,11 @@ title_hold_bib_levels = a:b:m:d
 ; of defining the "checked out" message text:
 ;Charged = "Charged"
 
+; This section allows mapping of patron block codes to VuFind translation strings
+[PatronBlockMappings]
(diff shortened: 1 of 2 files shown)


demiankatz (commented):
Thanks, @EreMaijala. Looks reasonable, but one question before I merge this: are block codes instance-specific, or are these universal? If they're universal, maybe we should consider including a default set of values. This will be targeted to release 9.1, so we can add new translation strings if necessary.

EreMaijala:
@demiankatz As far as I know they're instance-specific. The [Sierra manual](https://documentation.iii.com/sierrahelp/Default.htm#sril/sril_records_fixed_field_types_patron.html#MBLOCK) gives `a` and `b` as an example, but doesn't imply those are preinstalled, and they don't match the ones used in Finland.

demiankatz (approved):
In that case, I'll merge as-is and add a note to the changelog. Thanks!

PR #2857, opened 2023-04-27
Add support for "uncertain" item availability status

EreMaijala:
- Extends the current availability field with a tri-state option while retaining compatibility.
- Changes the status message handling so that if the tri-state field is used and a status message is provided, it's also displayed for available items.
- Adds Mink tests to cover the new functionality. I couldn't find any that would directly test holdings apart from SearchViewsTest, so I added a new HoldingsTest class.
- Add support to Alma driver for mapping item location types to statuses e.g. to be able to mark items intended only for staff use.

TODO
- [ ] Add relevant changelog notes when merging (Alma driver constructor signature change)


EreMaijala:
Note that the added tests don't even try to cover all aspects of item status information, but at least they provide a start.

themes/bootstrap3/templates/ajax/status-uncertain.phtml
```diff
@@ -0,0 +1 @@
+<span class="label label-warning"><?=$this->transEsc('Uncertain')?></span>
```
demiankatz: We should be sure to add this new translation string to the language files.

module/VuFind/src/VuFind/ILS/Connection.php
```diff
@@ -59,6 +59,27 @@ class Connection implements TranslatorAwareInterface, LoggerAwareInterface
     use \VuFind\I18n\Translator\TranslatorAwareTrait;
     use \VuFind\Log\LoggerAwareTrait;
 
+    /**
+     * Status code for unavailable items
+     *
+     * @var int
+     */
+    public const ITEM_STATUS_UNAVAILABLE = 0;
```
demiankatz, line 67: Is there a compelling reason to put these in the Connection class? Might there be an argument for having a separate class for containing status-specific constants (and maybe, in the future, logic)? I don't mind having them here, but it feels like it might not be the best fit since they're not actually used directly as part of this class.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
```diff
@@ -460,6 +463,58 @@ protected function getItemStatusError($record, $msg = '')
...
 
+    /**
+     * Convert availability boolean or int to a string
+     *
+     * @param mixed $available Boolean value or one of the Connection::ITEM_STATUS_*
+     *                         constants
+     *
+     * @return string
+     */
+    protected function availabilityToString($available): string
```
demiankatz, line 480: I guess in theory, if we create a status-specific class in the ILS namespace, logic like this could move there too... though then we'd have to think about extensibility (does it become a registered service?), etc. Doesn't have to change, just brainstorming.

demiankatz (changes requested):
I haven't had time for an in-depth analysis yet, but a few high-level comments for starters.

In general, this is making me wonder if this is an opportunity to split apart the AJAX handler so that it is more focused on the AJAX request/response handling, and the actual status-related logic is in its own class for easier testing and reuse. But maybe that's not the right path, just something the work in progress suggested to me.

themes/bootstrap3/templates/ajax/status-uncertain.phtml
EreMaijala: Added.

module/VuFind/src/VuFind/ILS/Connection.php
EreMaijala: It was the only existing place that I could think of. I've now added a logic class to contain the constants.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
EreMaijala: The problem, as I see it, is that the current methods and their expected return values are pretty specific to this handler and what the caller expects to receive. For instance the availability field is a string that used to contain 'true' or 'false', which is not something I'd spread any further, though it works fine here as adding a new option is easy. That said, moving logic to the logic class would make sense, but it should be generalized. Then the AJAX handler would just convert it to what it needs to return. But this is way beyond the scope of this change set. :)

demiankatz: I'm fine with waiting -- now that we've established a new class, we can always refactor things there in future if it seems worth the effort.

languages/en.ini
```diff
@@ -1351,6 +1351,7 @@ total_users = "Total Users"
 Transliterated Title = "Transliterated Title"
 tree_search_limit_reached_html = "Your search returned too many results to display in the tree. Showing only the first <b>%%limit%%</b> items. For a full search click <a id="fullSearchLink" href="%%url%%" target="_blank">here.</a>"
 trending_items_channel_title = "Trending Items"
+Uncertain = "Uncertain"
```
demiankatz: I wonder if, while we're changing all this, it's the time to either establish a new text domain, or at least use a prefix, for item availability statuses. As a translation string, "Uncertain" (just like "Unknown" and "Available") is pretty ambiguous. I wouldn't remove any existing strings, but it might be worth copying them to prefixed/text-domained versions.

Again, if you'd rather keep things simple here, we don't have to do it right away... but since we're potentially adding open-ended status codes, now might be a good time to clean this up.

module/VuFind/src/VuFind/ILS/Driver/Alma.php
```diff
@@ -334,6 +348,38 @@ public function getHolding($id, $patron = null, array $options = [])
                     );
                 }
 
+                $available = null;
```
demiankatz: Wouldn't it make more sense to add all this new logic to getAvailabilityFromItem rather than getHolding, since this is really just an extension of that functionality? Maybe getAvailabilityFromItem needs to be broken up so that it utilizes one or two support methods. But unless I'm missing something, none of this new code requires new values that aren't already available inside getAvailabilityFromItem.

module/VuFind/src/VuFind/ILS/Driver/Alma.php
```diff
@@ -1878,21 +1924,87 @@ protected function getTranslatableStatusString($element)
...
     }
 
+    /**
+     * Get location type for an item
+     *
+     * @param SimpleXMLElement $item Item
+     *
+     * @return string
+     */
+    protected function getItemLocationType($item)
```
demiankatz, line 1965: I also wonder if some of these new methods might justify additions to the existing unit test. Even if it's a pain to get test fixtures, etc., some quick tests that use the ReflectionTrait to directly call protected methods could allow checks that all the code paths work as expected.

themes/bootstrap3/templates/RecordTab/holdingsils/extended.phtml
```diff
@@ -29,7 +29,16 @@
...
+              <?php
+                if ((int)$holding['availability'] === \VuFind\ILS\Logic\ItemStatus::STATUS_UNCERTAIN) {
+                  $statusClassEsc = 'text-warning';
+                  $status = $holding['status'];
+                } else {
+                  $statusClassEsc = 'text-success';
+                  $status = is_int($holding['availability']) && !empty($holding['status']) ? $holding['status'] : 'Available';
+                }
+              ?>
+              <span class="<?=$statusClassEsc?>"><?=$this->transEsc($status)?><link property="availability" href="http://schema.org/InStock"></span>
```
demiankatz: Does uncertain status require a different schema.org property?

themes/bootstrap3/templates/RecordTab/holdingsils/standard.phtml
```diff
@@ -20,7 +20,16 @@
...
+            <?php
+              if ((int)$holding['availability'] === \VuFind\ILS\Logic\ItemStatus::STATUS_UNCERTAIN) {
+                $statusClassEsc = 'text-warning';
+                $status = $holding['status'];
+              } else {
+                $statusClassEsc = 'text-success';
+                $status = is_int($holding['availability']) && !empty($holding['status']) ? $holding['status'] : 'Available';
+              }
+            ?>
+            <span class="<?=$statusClassEsc?>"><?=$this->transEsc($status)?><link property="availability" href="http://schema.org/InStock"></span>
```
demiankatz: If we change this above, we also need to change it here.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
```diff
@@ -276,20 +277,23 @@ protected function getItemStatus(
...
-                isset($info['use_unknown_message'])
-                && $info['use_unknown_message'] == true
-            ) {
-                $use_unknown_status = true;
+            if ($info['use_unknown_message'] ?? false) {
+                $useUnknownStatus = true;
+            }
+            // Check for non-reserves:
+            if ($info['reserve'] === 'N') {
+                $reserve = false;
```
demiankatz: Is there a reason we assume things are on reserve unless otherwise specified? Might this lead to false positives?

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
```diff
@@ -460,6 +464,58 @@ protected function getItemStatusError($record, $msg = '')
...
+     */
+    protected function getAvailabilityMessage(
+        array $messages,
+        $available,
+        bool $useUnknownStatus
+    ): string {
+        if ($useUnknownStatus) {
+            $key = 'unknown';
+        } else {
+            switch ($available) {
```
demiankatz: Is it really worth using a switch here? Would an if..elseif..else be more readable? (I'm okay with keeping it, especially if you expect this to grow significantly in future, but a switch feels like overkill for three options).

demiankatz (changes requested):
Thanks, @EreMaijala! I've given this a more thorough review. Once you've had a chance to address this feedback, I'll do some hands-on testing next.

languages/en.ini
EreMaijala: We already have HoldingStatus text domain, I just forgot. Fixed.

module/VuFind/src/VuFind/ILS/Driver/Alma.php
EreMaijala: Now refactored.

module/VuFind/src/VuFind/ILS/Driver/Alma.php
EreMaijala: There were existing unit tests only for Alma resolver driver, but since adding new tests made sense I went through the effort here. Now all of getHolding and the new methods should be covered. While working on it I noticed that the resulting URLs were a bit funny when paging wasn't enabled, so that's fixed as well. And there was a useless constructor param (config loader), which was removed.

themes/bootstrap3/templates/RecordTab/holdingsils/extended.phtml
EreMaijala: I guess it depends on how you think about it, but I changed it to `LimitedAvailability` since that should cover most cases.

themes/bootstrap3/templates/RecordTab/holdingsils/standard.phtml
EreMaijala: Done.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
EreMaijala: I've no idea what I was thinking. Now inverted.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
EreMaijala: I agree it's cleaner with an `if` chain, changed.

module/VuFind/src/VuFind/ILS/Driver/Alma.php
demiankatz: Thanks for taking the time!

themes/bootstrap3/templates/RecordTab/holdingsils/standard.phtml
```diff
@@ -40,7 +51,7 @@
             </a>
           <?php endif; ?>
         <?php else: ?>
-          <?php /* Begin Unavailable Items (Recalls) */ ?>
+          <?php /* Begin Unavailable Items (Holds/Recalls) */ ?>
```
demiankatz: Why the changed comment here? I think in VuFind's simplistic view, available items get holds and unavailable items get recalls... and that's why "holds" wasn't part of this comment. Has something changed?

I also notice that the equivalent line in extended.phtml was NOT changed, so if we do keep this change, we should make the parallel edit there.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/AlmaTest.php
```diff
@@ -0,0 +1,243 @@
...
+     * @param array  $config Driver configuration (null to use default)
+     *
+     * @return void
+     */
+    protected function createConnector(string $test, array $config = null): void
+    {
+        // Setup test responses
+        $this->fixtureSteps = $this->getJsonFixture("alma/responses/$test.json");
+        $this->currentFixture = $test;
+        $this->currentFixtureStep = 0;
```
demiankatz, line 161: Now that we're using this pattern in three drivers, I wonder if we should consider making it a trait somehow. Not something we have to do right now, just food for thought...

module/VuFind/src/VuFind/Cache/KeyGeneratorTrait.php
```diff
@@ -58,7 +58,8 @@ protected function getCacheKey($suffix = null)
         // Test the build key
         if (
             $this->cache
-            && !preg_match($this->cache->getOptions()->getKeyPattern(), $key)
+            && ($keyPattern = $this->cache->getOptions()->getKeyPattern())
```
demiankatz: Is this a bug fix worth merging separately?

demiankatz (changes requested):
Thanks, @EreMaijala -- a few more comments.

I also wonder if creating a test for AjaxHandler/GetItemStatuses would be a good idea to ensure that we're not missing any code paths. I'm willing to help with that if you don't have time (not that I have a lot of time myself... :-) ).

themes/bootstrap3/templates/RecordTab/holdingsils/standard.phtml
EreMaijala: That was a mistake, now reverted (though it bugs me a bit that it really is the hold function everywhere and it's up to the ILS to decide if it's handled as a recall).

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/AlmaTest.php
EreMaijala: I tried to think of something clear and useful, but since the actual implementations vary a bit, I didn't come up with anything.

module/VuFind/src/VuFind/Cache/KeyGeneratorTrait.php
EreMaijala: Yes, I'll open a PR.

EreMaijala: PR #2892.

EreMaijala:
@demiankatz The newly added HoldingsTest class should test getItemStatuses. I'll need to finish remote code coverage support to ensure that it does. :)

themes/bootstrap3/templates/RecordTab/holdingsils/standard.phtml
demiankatz: Agreed; adding more granularity to that might make sense at some point, but somehow we've gotten this far without anyone raising a use case where the distinction was crucial.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/AlmaTest.php
demiankatz: If I find myself adding it to a fourth test class in the future, I'll give this some more thought. Not a big deal in the meantime. :-)

module/VuFind/src/VuFind/Cache/KeyGeneratorTrait.php
demiankatz: Thanks! I've merged that back into this PR to reduce the file count.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
```diff
@@ -276,20 +280,23 @@ protected function getItemStatus(
...
-            if (
-                isset($info['use_unknown_message'])
-                && $info['use_unknown_message'] == true
-            ) {
-                $use_unknown_status = true;
+            if ($info['use_unknown_message'] ?? false) {
+                $useUnknownStatus = true;
+            }
+            // Check for non-reserves:
+            if ($info['reserve'] !== 'Y') {
```
demiankatz: Is it worth using `$info['reserve'] ?? 'N'` to be more tolerant of missing values?

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
```diff
@@ -366,18 +373,17 @@ protected function getItemStatusGroup($record, $messages, $callnumberSetting)
...
         // Summarize call number, location and availability info across all items:
         $locations = [];
-        $use_unknown_status = $available = false;
+        $useUnknownStatus = false;
+        $available = 'false';
         foreach ($record as $info) {
             // Find an available copy
-            if ($info['availability']) {
-                $available = $locations[$info['location']]['available'] = true;
+            if ($info['availability'] && 'true' !== $available) {
```
demiankatz: Might this check cause a short-circuit that prevents $locations[$info['location']] from getting populated? Looks to me like if the first location is available, then we won't fill in the details for subsequent locations, but maybe I'm missing something.

demiankatz (changes requested):
I finally did a very close review and found a couple of remaining issues. I wonder if some of this indicates that we need a better test case somewhere to cover the grouped scenario!

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
EreMaijala: I reverted the check to what it was originally, but added null coalescing.

module/VuFind/src/VuFind/AjaxHandler/GetItemStatuses.php
EreMaijala: Oops, fixed.

EreMaijala:
@demiankatz HoldingsTest Mink test now covers this pretty well. I'm sure there are still cases that could be added to ensure that all different permutations are covered, but this should cover all cases here.

demiankatz (commented):
Thanks, @EreMaijala! Just two questions before I do a final test run and hopefully merge this:

1.) At this point, I feel pretty comfortable merging this into release 9.1 due to minimal backward-breaking changes. Do you agree, or should we make this dev-10.0?

2.) What led you to revert your reserves changes? Your previous logic seemed like it was addressing a longstanding bug in the code, but maybe I misunderstood something. In any case, I think it's fine to merge as-is on the principle of avoiding making too many changes at once, but I'm wondering if a more nuanced reserves check is worth working on as a follow-up.

EreMaijala:
@demiankatz 

1. Works for me!
2. I chickened out. I started to think that perhaps, if someone is relying on the quirks of the current code, something could go wrong. But if you feel like it's better to fix things for good now, I'd be happy to re-add it here.

demiankatz:
@EreMaijala, since it's a long-standing behavior that we're considering changing, maybe the most cautious thing would be to introduce the reserves changes in a dev-10.0 PR after this one gets merged. I'll try to run final tests and get this merged before the end of the week!

module/VuFind/src/VuFind/ILS/Driver/AlmaFactory.php
```diff
@@ -69,7 +69,6 @@ public function __invoke(
         // passed in as options):
         $driver = new $requestedName(
             $container->get(\VuFind\Date\Converter::class),
-            $container->get(\VuFind\Config\PluginManager::class),
```
demiankatz, line 72: Making this change turns AlmaFactory into a copy-and-paste of DriverWithDateConverterFactory. I think we can delete this factory and change the configuration in the PluginManager.

demiankatz (changes requested):
@EreMaijala, I think this is nearly ready to merge, but I noticed one small issue (see below) and also had a couple of questions:

1.) Can you update the driver spec in the wiki to reflect the new options in the available field?

2.) For changelog entries, I'm planning to add:

`- ILS drivers can now report an "uncertain" status in addition to the traditional true/false availability status values. This new feature is currently supported by the Alma and Demo drivers. See the locationTypeItemStatus setting in Alma.ini for related configuration.`

with this in the "BC breaks" section:

`- The Alma driver's constructor signature and factory configuration have changed, and its internal availability status logic has been significantly revised. See [[https://github.com/vufind-org/vufind/pull/2857|pull request #2857]] for details.`

`- The GetItemStatuses AjaxHandler has been significantly revised to support a broader range of statuses. See [[https://github.com/vufind-org/vufind/pull/2857|pull request #2857]] for details.`

Does that sound reasonable? Is there anything else I should add?

module/VuFind/src/VuFind/ILS/Driver/AlmaFactory.php
EreMaijala: Awesome, one _less_ file! Done.

EreMaijala:
@demiankatz I'll update the driver spec. The change log texts look good to me, thanks. I fixed just one typo in there.

demiankatz (approved):
Looking good!

PR #2858, opened 2023-04-27
Fix check for getUrlsForRecord to include the record id for MultiBack…

EreMaijala:
…end support.


demiankatz (approved):
Thanks for the fix. Looks like I introduced this problem in #2810. :-)

PR #2860, opened 2023-04-28
Add LibGuidesAZ backend for Databases A-Z list

maccabeelevine:
Purpose:
- Display results from a search of the LibGuides Databases A-Z list, which is a separate data source from the standard LibGuides search (of research guides).
- In particular, there should be the ability to include these results within a Combined search page.  Currently that requires the AZ list be a separate backend, since there is no way otherwise to distinguish it from the standard LibGuides data in combined.ini.

Approach:
- Like the existing LibGuides integration, we can't use the LibGuides actual API because it doesn't support search.  The [API doc itself](https://ask.springshare.com/libguides/faq/873) makes clear that the "widget" API should be used instead.
- In most cases just extend the existing LibGuides classes.


maccabeelevine:
@demiankatz I haven't yet written unit tests for this, wanted your eyes on the main code first.  Also I don't think the same "extend the base LibGuides class" approach will work for test files, or at least I've never done that myself, sit would likely just be a copy/paste of those four files.

module/VuFind/src/VuFind/Controller/LibGuidesAZController.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+/**
+ * LibGuides A-Z Databases Controller
+ *
+ * PHP version 7
+ *
+ * Copyright (C) Villanova University 2010.
```
demiankatz: Might want to update dates to reflect when the files were created.

module/VuFind/src/VuFind/Controller/LibGuidesAZController.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Is the result scroller active?
+     *
+     * @return bool
+     */
+    protected function resultScrollerActive()
+    {
+        $config = $this->serviceLocator->get(\VuFind\Config\PluginManager::class)
+            ->get('LibGuidesAZ');
```
demiankatz, line 66: You should probably include an example LibGuidesAZ.ini file (did you perhaps forget to commit this?).

demiankatz: Note also that you could theoretically use our built in [Parent_Config] support to make LibGuidesAZ.ini inherit all or most settings from LibGuides.ini if you think that would improve maintainability.

module/VuFind/src/VuFind/Controller/LibGuidesAZController.php
```diff
@@ -0,0 +1,70 @@
...
+        return isset($config->Record->next_prev_navigation)
+            && $config->Record->next_prev_navigation;
```
demiankatz: This might be cleaner as `return $config->Record->next_prev_navigation ?? false;`

(And I realize you copied this from the existing code... and I also realize that this probably serves no real purpose at all since I don't think there's a record view where this would even apply... but I just can't resist advocating for null coalescing at every opportunity ;-) ).

themes/bootstrap3/templates/libguidesaz/home.phtml
```diff
@@ -0,0 +1,2 @@
+<?php
+    include $_SERVER['CONTEXT_DOCUMENT_ROOT'] . '/../themes/bootstrap3/templates/libguides/home.phtml';
```
demiankatz: Is there a reason you can't `$this->render('libguides/home.phtml')` here for cleaner code?

themes/bootstrap3/templates/libguidesaz/results.phtml
```diff
@@ -0,0 +1,2 @@
+<?php
+    include $_SERVER['CONTEXT_DOCUMENT_ROOT'] . '/../themes/bootstrap3/templates/libguides/results.phtml';
```
demiankatz: I'd also use render here if possible. See, for example, how we do this in [search2](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/search2/results.phtml).

module/VuFind/src/VuFind/Search/Factory/LibGuidesAZBackendFactory.php
```diff
@@ -0,0 +1,112 @@
...
+ * @package  Search
+ * @author   David Maus <maus@hab.de>
+ * @author   Maccabee Levine <msl321@lehigh.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Site
+ */
+
+namespace VuFind\Search\Factory;
+
+use VuFindSearch\Backend\LibGuidesAZ\Backend;
```
demiankatz: If you want to reduce the number of spare classes you're creating, I suspect that you can eliminate all of the LibGuidesAZ classes, and just reuse equivalent LibGuides classes. Having separate classes is important inside the VuFind module since it relates to routing and parameter processing, but once you get to the search backend code, there's no consequence to reusing the same backend code multiple times (we already do this for Solr, etc.).

It looks to me like the only code that's unique to the Backend\LibGuidesAZ classes is the widget_type used in the query builder... and you could address that by adding a `defaultWidgetType` property to the class, and a `setDefaultWidgetType` method, and then have the factory set up the query builder appropriately.

There's nothing wrong with your current approach if you think there might be a reason to more explicitly differentiate functionality in the future, but if you want to reduce the number of files in the PR, I think this could be an easy win.

demiankatz (changes requested):
Thanks again, @maccabeelevine! This looks like a great start. I just have a few comments (most significantly, about possibly reducing the number of files added to the VuFindSearch module). I haven't done a really close review of your factory refactoring yet, since I imagine it may change again in response to my comments. I'll look there more carefully once you've had a chance to respond to this current round of review.

Also note that I unfortunately can't test this myself hands-on since I don't have a LibGuides account. It might be worth reaching out to the community to see if anyone else is interested in getting involved in this review process.

demiankatz:
@maccabeelevine, I've merged the latest dev branch into this PR in an effort to get our green checkmark back (since we recently added PHP 8.2 testing requirements). I think this ball is still in your court, so please let me know when it's convenient to discuss the above review!

maccabeelevine:
Hi @demiankatz yes this is on my plate, hopefully next week!  Thanks for the ping.

module/VuFind/src/VuFind/Controller/LibGuidesAZController.php
maccabeelevine: Yeah this was a copy/paste and yeah this will probably never apply, but I agree with the principle.  Fixed.

themes/bootstrap3/templates/libguidesaz/home.phtml
maccabeelevine: The only reason was "laminas newbie".  Fixed.

themes/bootstrap3/templates/libguidesaz/results.phtml
maccabeelevine: Done.

module/VuFind/src/VuFind/Search/Factory/LibGuidesAZBackendFactory.php
maccabeelevine: I took this advice.  I'm not sure about whether more differentiation will be necessary in the future, but I think better to cross that bridge later than add five trivial files at this point.

module/VuFind/src/VuFind/Controller/LibGuidesAZController.php
maccabeelevine: Oops.  Done.

module/VuFind/src/VuFind/Controller/LibGuidesAZController.php
maccabeelevine: I didn't even create a separate LibGuidesAZ.ini, that line 66 was a mistake, and somehow it still picks up the configuration from LibGuides.ini.  Must be the fact I had LibGuidesAZ/Options extending LibGuides/Options.  Fixed.

I decided to take your [Parent_Config] suggestion.  Of the seven properties in LibGuides.ini, the first four (timeout, institution ID, version, baseUrl) probably have to be identical for LibGuidesAZ.  The other three (default_limit, limit_options, defaultSearch) theoretically could be different, i.e. perhaps defaultLimit if both LibGuides types are used in a Combined Search solution.

I wasn't sure whether to copy / paste and comment out all the properties into LibGuidesAZ.ini.  On the plus side it's easier to override then, on the minus you have to maintain them all in both places.  I decided not to copy them but can do so if it's VuFind practice.

maccabeelevine:
@demiankatz Ok I addressed all the issues above.  The PHP 8 build is broken again, looks like an issue with testMergedParamBag, but I'm not clear how to resolve it and I don't write to write any AZ-specific tests until the existing ones work again.

module/VuFind/src/VuFind/Controller/LibGuidesAZController.php
demiankatz: Usually our practice is to copy the settings but not the associated comments, instead putting in a note explaining where the detailed information can be found. This helps prevent things from getting out of sync over time while still providing helpful reminders of available settings. See [Search2.ini](https://github.com/vufind-org/vufind/blob/dev/config/vufind/Search2.ini) for an example of this.

demiankatz (changes requested):
@maccabeelevine, thanks for all the progress on this. I'm pretty sure that all the failing tests are related to the fact that you've added a new parameter to the ParamBag used in the backend, so test expectations need to be adjusted to reflect the fact that there's an extra value now. Let me know if you need more detailed support sorting it out!

demiankatz:
To be a little more precise, @maccabeelevine, I think the issue is that all the expectations like this one:

https://github.com/vufind-org/vufind/blob/dev/module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/LibGuides/BackendTest.php#L169

Need to have the widget_type value added to them.

module/VuFind/src/VuFind/Controller/LibGuidesAZController.php
maccabeelevine: Ok, I copied the settings.  I think the [Parent_Config] section being present, plus its comment, makes clear where the details can be found.

config/vufind/LibGuidesAZ.ini
```diff
@@ -0,0 +1,16 @@
+; Settings for the LibGuides AZ Databases List are likely to be the same as those 
+; for the canonical "guides" part of LibGuides, defined in LibGuides.ini.
+; But override here if needed.
+[Parent_Config]
+relative_path = ./LibGuides.ini
+
+; ---------- config.ini settings ----------
```
demiankatz: I don't think you need this comment; these settings don't come from config.ini. I'd either change this to say "LibGuides.ini" or remove it entirely. Given the Parent_Config settings right above, removing it is probably fine.

config/vufind/LibGuidesAZ.ini
```diff
@@ -0,0 +1,16 @@
...
+
+;[General]
+;timeout = 30
+;iid = my-id
+;version = 2
+;baseUrl = https://lgapi.libapps.com/widgets.php
+;default_limit        = 20
+;limit_options        = 10,20,40,60,80,100
+;defaultSearch = "general"
```
demiankatz: You might want to add an end-of-file line break here for consistency with most other files in the project.

demiankatz (changes requested):
Thanks for the .ini update -- a couple of very minor nitpicks for you. :-)

I'll give this some hands-on testing once the test suite is passing. Let me know if I can help with anything else in the meantime!

config/vufind/LibGuidesAZ.ini
maccabeelevine: Yeah typo, I meant Libguides.ini obviously.  But you're right it's redundant, I removed it.

config/vufind/LibGuidesAZ.ini
maccabeelevine: Done!

config/vufind/LibGuidesAZ.ini
demiankatz: Thanks!

module/VuFind/src/VuFind/Search/LibGuides/Options.php
```diff
@@ -40,14 +40,16 @@
  */
 class Options extends \VuFind\Search\Base\Options
 {
+    protected string $iniName = 'LibGuides';
```
demiankatz, line 48: This should get a comment on it, like:

```
/**
 * Name of .ini file to use for search and facet settings.
 *
 * @var string
 */
```

module/VuFind/src/VuFind/Search/LibGuidesAZ/Options.php
```diff
@@ -0,0 +1,56 @@
...
+ * @category VuFind
+ * @package  Search_LibGuides
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Maccabee Levine <msl321@lehigh.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Page
+ */
+class Options extends \VuFind\Search\LibGuides\Options
+{
+    protected string $iniName = 'LibGuidesAZ';
```
demiankatz, line 50: This needs the same comment as the other Options class.

demiankatz (changes requested):
I did one last review of this before testing, and I noticed one trivial thing in need of fixing: a missing doc comment on a couple of class properties.

I'll go ahead with my testing now since fixing this problem has no functional effect on the code. I'll report on my findings shortly! :-)

module/VuFind/src/VuFind/Search/LibGuides/Options.php
maccabeelevine: php-cs-fixer and phpcs-console didn't catch this one!  Fixed.

module/VuFind/src/VuFind/Search/LibGuidesAZ/Options.php
maccabeelevine: Fixed!

module/VuFind/src/VuFind/Search/LibGuides/Options.php
demiankatz: Yes, for some reason, the style checkers don't complain about this -- it just creates a warning in the PHPDocumentor output. In any case, thanks for the fix!

demiankatz (approved):
Looks good, @maccabeelevine! Merging now.

dltj:
@maccabeelevine : Do you have an example of how this looks in the Lehigh VuFind interface? I looked for [ABI/INFORM](https://asa.lib.lehigh.edu/Search/Results?lookfor=ABI%2FINFORM&type=AllFields&submit=Search) and [Access Science](https://asa.lib.lehigh.edu/Search/Results?lookfor=Access+Science&type=AllFields&limit=20), for instance, and those look like FOLIO records.

maccabeelevine:
> @maccabeelevine : Do you have an example of how this looks in the Lehigh VuFind interface? I looked for [ABI/INFORM](https://asa.lib.lehigh.edu/Search/Results?lookfor=ABI%2FINFORM&type=AllFields&submit=Search) and [Access Science](https://asa.lib.lehigh.edu/Search/Results?lookfor=Access+Science&type=AllFields&limit=20), for instance, and those look like FOLIO records.

@dltj Sorry no it's not in production yet at Lehigh.  Happy to screenshare if you like.

PR #2862, opened 2023-05-02
enable multiple favicons through configuration

charbugs:
Hi,

This PR proposes a solution to enable multiple favicons: The `favicon` config value in `themes.config.php` can be an array of attribute sets for link elements.

In order for you to test it I created an example set of 4 favicons of different sizes an of different background colors:

[favicons.zip](https://github.com/vufind-org/vufind/files/11373863/favicons.zip)

You must change the `theme.config.php` to

```php
'favicon' => [
        [
            'href' => 'favicons/favicon-16x16-blue.png',
            'rel' => 'icon',
            'sizes' => '16x16',
            'type' => 'image/png'
        ],
        [
            'href' => 'favicons/favicon-32x32-green.png',
            'rel' => 'icon',
            'sizes' => '32x32',
            'type' => 'image/png'
        ],
        [
            'href' => 'favicons/favicon-192x192-red.png',
            'rel' => 'icon',
            'sizes' => '192x192',
            'type' => 'image/png'
        ],
        [
            'href' => 'favicons/favicon-512x512-orange.png',
            'rel' => 'icon',
            'sizes' => '512x512',
            'type' => 'image/png'
        ],
    ],
```

Depending on your browser, device and operation system different icons may be used in different contexts, e.g. browser tab, browser history, app icon on you start screen.

What do you guys think about this?


module/VuFindTheme/src/VuFindTheme/View/Helper/SetupThemeResources.php
```diff
@@ -126,16 +126,33 @@ protected function addLinks()
...
-                    'href' => $imageLink($favicon),
-                    'type' => 'image/x-icon', 'rel' => 'shortcut icon'
-                ]
-            );
+            if (is_array($favicon)) {
+                foreach($favicon as $attrs) {
+                    $headLink(
+                        [
+                            'href' => $imageLink($attrs['href']),
+                            'rel' => $attrs['rel'],
```
demiankatz: It would probably be a good idea to address the possibility of missing values here -- i.e. either default with null coalescing, skip attributes if no reasonable default can be determined, or skip/throw an exception if a required value like href is missing.

One question is whether we really need to "curate" which attributes pass through. Would it be better to do something like:

```
foreach ($favicon as $attrs) {
    // Convert string into href array
    $attrs = is_array($attrs) ? $attrs : ['href' => $attrs];
    if (!isset($attrs['href'])) {
        continue; // we can't display without an href!
    }
    $attrs['href'] = $imageLink($attrs['href']);
    $attrs['rel'] = $attrs['rel'] ?? 'shortcut icon';
    $attrs['type'] = $attrs['type'] ?? 'image/x-icon';
    $headLink($attrs);
}
```

This is more open-ended if we want to add support for additional attributes via the config file without changing code.

(Note that since this tolerates strings in place of arrays, it could also be refactored to a support method that you could potentially share between both the single-value and multi-value scenarios).

demiankatz (changes requested):
Thanks, @charbugs, this seems like a good idea to me! I made some suggestions below regarding the existing code. On top of that, we should probably add some additional comments to the theme configuration so that available options are more fully documented. It might also be nice to add some different icon sizes by default, though I'm not sure what image(s) we would want to use in that case -- I'm not sure where our current favicon graphic originated.

charbugs:
Hi @demiankatz, thanks for the review.

> It might also be nice to add some different icon sizes by default, though I'm not sure what image(s) we would want to use in that case -- I'm not sure where our current favicon graphic originated.

Not to be misunderstood, I don't recommend a certain set of icons or icon properties with this PR. It seems to me like best practices around favicons is changing from time to time. So I think we should leave it to the developers to decide what is best for their use cases. The icons I attached to this PR where created with an [favicon generator](https://favicon.io/) just for demoing.

> This is more open-ended if we want to add support for additional attributes via the config file without changing code.

I agree and I updated the code to allow for arbitrary attributes.

> It would probably be a good idea to address the possibility of missing values here -- i.e. either default with null coalescing, skip attributes if no reasonable default can be determined, or skip/throw an exception if a required value like href is missing.

In general I think we should count on the browser to find default values for missing attributes. Having said that:

* If the `favicon` config value is a string we take it as an path to an .ico file and set `type=image/x-ico` and `rel=icon`.
* We need to set the `rel` attribute in any case because for some reason I could not figure out `HeadLink::isDuplicateStylesheet` throws an error if it is not set. I set it to `icon` by default.
* There are 3 options when it comes to the `href` attribute: 1. don't validate at all, 2. throw if not set, 3. throw if file not found. I tend to 1.

>  On top of that, we should probably add some additional comments to the theme configuration so that available options are more fully documented.

Where should the comments go? Probably in `themes/root/theme.config.php`?

demiankatz:
Thanks, @charbugs! I understand that you're not recommending any specific practice for favicons here... but your improvements suggest to me that we might want to offer some higher-resolution alternatives at some point, since our current icon is very 2007. ;-) Not something that has to be addressed as part of this work, of course, but something to keep in mind for the future!

Regarding the location of comments, I think for now, [bootstrap/theme.config.php](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/theme.config.php) is probably the best place, since that's where we actually use all the settings. (In principle, I agree that root would probably be better, but we already have most of our comments in bootstrap3 so we might as well follow that precedent for now, and revisit when we refactor to create new themes in the next major release).

demiankatz (changes requested):
@charbugs, this PR is reporting a conflict in SetupThemeResources.php. Do you have any idea what is going on there? GitHub is acting strangely today, so maybe it's a glitch... but if it's a real conflict, could you resolve it when time permits? Please let me know if I can be of any further assistance!

charbugs:
I added comments to theme.config.php and resolved the merge conflict. Not sure what it means that the task "CI / Tests with PHP 7.4 (pull_request)" has been "Cancelled after 2m".

demiankatz:
Thanks, @charbugs! When a build gets canceled in GitHub Actions, it usually means that a different build has failed. In this case, the 8.0 build failed because you need to run `vendor/bin/phing php-cs-fixer` to apply some automatic style fixes. When one build fails, GitHub Actions cancels any that are still running to save on resources. When this happens, it's just a good idea to find which build actually failed and then click on it to see the detailed output.

Let me know if you need any more help with this! :-)

charbugs:
I executed the vendor/bin/phing php-cs-fixer command but the PHP 8.0 task still fails.

demiankatz:
@charbugs, is it possible you haven't run `composer install` recently? It looks like an outdated version of php-cs-fixer ran, which is why a bunch of weird changes occurred and the build is still failing. I'd recommend reverting the last commit, running `composer install` and then trying again. Sorry for the nuisance!

charbugs:
You were right, I used an outdated version of php-cs-fixer! Now all checks are :heavy_check_mark:

demiankatz (approved):
This looks good to me. Thanks for all the work on it, @charbugs.

PR #2864, opened 2023-05-02
Fix formatting error in Welsh translation.

demiankatz:
This PR fixes a problem that @lahmann noticed and reported on today's Community Call: a later line of the translation file somehow got duplicated as part of an earlier translation.


lahmann (approved):
Thanks for fixing this!

PR #2865, opened 2023-05-02
Fix login followup URL for off-site auth

natecollins:
When using a link to an off-site auth system (single-sign-on) in the login lightbox, the followup URL was not being preserved. Lightbox handled reloading the page for direct logins, but that doesn't help if the user leaves the site to authenticate.

This patch makes it so the login followup URL is always set, but then clears that URL on login if the user is inside the lightbox.

This change resolved our issues with login followups. I welcome any insight if there are other considerations that need to be taken into account. Thanks!

### DIFF
--- a/module/VuFind/src/VuFind/Controller/MyResearchController.php
+++ b/module/VuFind/src/VuFind/Controller/MyResearchController.php
@@ -155,8 +155,9 @@ class MyResearchController extends AbstractBase
                     if (
                         $this->params()->fromPost('processLogin')
                         && $this->inLightbox()
-                        && empty($this->getFollowupUrl())
                     ) {
+                        // Clear followup and let lightbox handle redirect
+                        $this->clearFollowupUrl();
                         return $this->getRefreshResponse();
                     }
                 }
@@ -298,11 +299,10 @@ class MyResearchController extends AbstractBase
                 : $this->redirect()->toRoute('home');
         }
         $this->clearFollowupUrl();
-        // Set followup only if we're not in lightbox since it has the short-circuit
-        // for reloading current page:
-        if (!$this->inLightbox()) {
-            $this->setFollowupUrlToReferer();
-        }
+        // Lightbox handles reloading current page without a followup, however,
+        // always set a followup as lightbox won't handle page reload if auth
+        // is external. We clear the followup in homeAction if lightbox detected.
+        $this->setFollowupUrlToReferer();
         if ($si = $this->getSessionInitiator()) {
             return $this->redirect()->toUrl($si);
         }


demiankatz:
Thanks, @natecollins! Could you share a specific scenario where you are seeing this problem? My institution also uses SSO, and I haven't encountered the problem you describe, so I'm wondering if it's related to some narrower subset of functionality. I'd like to find a way to reproduce the problem locally before I review the code, just to be sure I have the full context in my mind. :-)

I'm also copying @crhallberg into this thread in case he has any additional insights.

natecollins:
For our situation, we have ChoiceAuth using both SAML (via Okta) through SimpleSAMLphp for SSO, along with FOLIO Okapi to auth community users. Both options are listed in the login lightbox (link to SAML auth, and login boxes for Okapi, example at https://catalog-beta.lib.msu.edu/). Community user logins via Okapi work as expected. However, we were missing the followup URL after users came back from the offsite Okta SAML auth, so users were always being returned to MyResearch/Home regardless of where they attempted to login from.

For reference, our [auth classes](https://github.com/MSU-Libraries/catalog/tree/main/vufind/module/Catalog/src/Catalog/Auth) and [configs](https://github.com/MSU-Libraries/catalog/tree/main/vufind/local/config/vufind) are available in our public Github, if they can better explain it than my description.

If there is a more narrow way to approach this, I'm all ears! I'm always get a bit paranoid about causing unexpected changes, but it seemed unavoidable to have to to set the followup URL to get people back to their original page. And it seemed others might run into the same situation should they have a similar setup.

demiankatz:
Thanks, @natecollins. I think the problem may actually be related to your SimpleSAML implementation, since I'm not having the problem with my Shibboleth-based SSO. When VuFind generates a URL to redirect to SSO, it's supposed to pass a callback URL along telling the SSO where to return the user to. This functionality should take the place of storing a redirect URL in the session. I'm wondering if perhaps this is getting lost somewhere, and so SimpleSAML is returning to a default location instead. Does that seem like a plausible theory? If so, let me know if you need any more details of the implementation, or if you'd like me to take a closer look at your SimpleSAML code.

(On a side note, I think it would be very interesting to offer SimpleSAML support as a core VuFind feature, so people have an out-of-the-box alternative to setting up Shibboleth for SSO; I just haven't had time to work on an implementation. If you think your current code might make sense as a community contribution, I'd certainly be happy to work with you on that as well).

natecollins:
Ah, thank you, @demiankatz, I wish I had caught that earlier! I didn't write our SAML auth class, but looking through the SimpleSAML docs, it does seem like if I should be able to get the return URL working properly without bothering VuFind's code. I'll direct my attention there. My apologies for going down the wrong path here.

As evidenced by my erroneous PR, I think our SAML code needs some additional review, but we'd be glad contribute anything we can. We're intent on ensuring the [entirety of our public catalog implementation](https://github.com/MSU-Libraries/catalog) is open source. Assuming we can get our SAML into proper shape, we are more than willing to work with you to add it to core VuFind. Thanks again!

demiankatz:
Thanks, @natecollins, I'm glad I could help! I'll look forward to working with you on SAML code at some point in the future. It's great that you're already sharing all of your code, and hopefully if we can work over time to get any community-reusable parts merged upstream, it will reduce your maintenance burden in the long term while also benefiting other users. :-)

Let me know if I can assist with anything else in the meantime!

PR #2866, opened 2023-05-02
Allow scripts to run via the Admin module.

demiankatz:
This is a proof of concept for a feature requested by @livedvd. For situations where VuFind is hosted for a group of users, and those users do not have command-line access, it may be valuable to allow curated access to certain commands via the admin interface. This PR adds a mechanism to make this possible.

I want to proceed very cautiously with this functionality, since allowing any gaps in security could have serious consequences. I think that as implemented here, it's pretty safe, however, as there is no opportunity to inject anything unexpected to the command line. The only way to define commands is via the .ini file, and in the current setup, there is no way to change that without already having command line access anyway. Obviously if we combined this with web-based .ini editing in future, we would be opening the door to all kinds of abuse -- that's a scenario we will have to avoid (but this is not the only reason to be very careful about web-based .ini editing as a future idea).

I'm curious if others find this feature interesting, or if there are strong feelings that we should NOT do this. (It could be implemented as an optional external module if that would make people more comfortable). I'm also interested in whether this would be better served as a separate controller rather than part of the Maintenance controller. I did it this way simply to reduce overhead during the proof-of-concept process, but it's easy enough to split apart if there's any desire.


module/VuFindAdmin/src/VuFindAdmin/Controller/MaintenanceController.php
```diff
@@ -50,10 +50,62 @@ public function homeAction()
...
+    protected function getScripts(): array
+    {
+        // Load the AdminScripts.ini settings
+        $config = $this->serviceLocator->get(\VuFind\Config\PluginManager::class)
+            ->get('AdminScripts')->toArray();
+
+        // Filter out any commands that the current user does not have permission to run:
+        $permission = $this->permission();
+        $filter = function ($script) use ($permission) {
+            return !isset($script['permission'])
```
EreMaijala: I'd make permission mandatory for an added layer of security.

module/VuFindAdmin/src/VuFindAdmin/Controller/MaintenanceController.php
```diff
@@ -50,10 +50,62 @@ public function homeAction()
...
+            $code = $output = null;
+            exec($details['command'], $output, $code);
+            $successCode = intval($details['successCode'] ?? 0);
+            if ($code !== $successCode) {
+                $this->flashMessenger()->addMessage(
+                    "Command failed; expected $successCode but received $code",
+                    'error'
+                );
+            } else {
+                $this->flashMessenger()->addMessage(
```
EreMaijala: I'd prefer addSuccessMessage (and addErrorMessage), but this is the common way..

EreMaijala (changes requested):
I think it's fine to have something like this when there are enough safeguards.

Some potential issues:

1. Do we want to define the default working directory for scripts?
2. This obviously won't work if the command needs permissions different than the web server.
3. Long-running commands would be an issue. You can't really run one of the longer-running utils synchronously without causing a timeout somewhere, or huge output. For this we'd need a cron'able command queue runner or something.

module/VuFindAdmin/src/VuFindAdmin/Controller/MaintenanceController.php
demiankatz: I always forget about those options, but I agree that they're more concise. Fixed! I'll open a separate PR to fix the other flash messenger calls in this controller.

module/VuFindAdmin/src/VuFindAdmin/Controller/MaintenanceController.php
demiankatz: My original thought was that you already need permission to access the admin module, so in some scenarios, that may be enough without being forced to add extra layers. However, it is relatively easy to just turn on "admin_enabled" in config.ini without setting proper permissions in permissions.ini, so I agree that we should probably be more strict. However, I see two possible approaches:

1.) If permission is undefined, default to "Access.AdminModuleScriptRunner" so there's a single permission to configure if granularity is not important (which makes the configuration more concise).

2.) Force a permission to be explicitly defined for every block in the configuration.

I somewhat prefer option 1, but option 2 is easy enough if you think it's better!

demiankatz:
Thanks, @EreMaijala! I agree with all of the concerns you raise, but I don't want to over-engineer anything without a more fleshed-out use case, so for now, I've just added some more comments to the .ini file addressing these issues. Let me know if you think those are sufficient for now or need to be expanded upon. I'm certainly open to adding more tools to VuFind to enhance this feature in future, but this also feels like something where we might be able to recommend existing external tools and best practices that can be triggered via scripts, and then we don't have to reinvent any wheels.

module/VuFindAdmin/src/VuFindAdmin/Controller/MaintenanceController.php
EreMaijala: Option 1 would be fine for me.

EreMaijala: Though on a second thought, I think not requiring particular permission when you can already access the admin UI is quite fine as well.

EreMaijala (approved):

module/VuFindAdmin/src/VuFindAdmin/Controller/MaintenanceController.php
demiankatz: Thanks, @EreMaijala. I suppose we could also potentially add a "default_script_runner_permission" setting to config.ini (or something like that) to establish a configurable default permission... Or we could create a [General] or [Global] section of AdminScripts.ini for global configuration settings like that. Do any of those ideas particularly appeal (or not appeal) to you?

EreMaijala: Nah, I think it's better to keep things compact so that any effective permissions are easy to see. If I had to choose, the section in the same ini file would be better for me, but I'm not sure why it would need to be configurable.

demiankatz: How does [ca260f0](https://github.com/vufind-org/vufind/pull/2866/commits/ca260f0df8b012f9c77e59167d5c50695035956a) look to you?

EreMaijala (approved):

PR #2867, opened 2023-05-03
KohaRest: Add support for purging loan history.

EreMaijala:
Selective deletion is not currently supported by the Koha API.

### DIFF
--- a/config/vufind/KohaRest.ini
+++ b/config/vufind/KohaRest.ini
@@ -165,4 +165,7 @@ extraFields = item-issue:acceptTerms:pickUpLocation
 [TransactionHistory]
 ; By default the loan history is disabled. Uncomment the following line to enable it.
 ;enabled = true
+; By default purging of all loan history is enabled. Uncomment the following line
+; to disable it:
+;purgeAll = false
 
--- a/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
@@ -1649,7 +1649,6 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
                 ],
                 'default_sort' => '-checkout_date',
                 'purge_all' => $this->config['TransactionHistory']['purgeAll'] ?? true,
-
             ];
         } elseif ('getMyTransactions' === $function) {
             $limit = $this->config['Loans']['max_page_size'] ?? 100;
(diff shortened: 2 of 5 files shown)


module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -1609,6 +1648,8 @@ public function getConfig($function, $params = [])
                     '+title' => 'sort_title',
                 ],
                 'default_sort' => '-checkout_date',
+                'purge_all' => $this->config['TransactionHistory']['purgeAll'] ?? true,
+
```
demiankatz: Unnecessary extra blank line?

config/vufind/KohaRest.ini
```diff
@@ -165,4 +165,7 @@ extraFields = item-issue:acceptTerms:pickUpLocation
 [TransactionHistory]
 ; By default the loan history is disabled. Uncomment the following line to enable it.
 ;enabled = true
+; By default purging of all loan history is enabled. Uncomment the following line
+; to disable it:
+;purgeAll = false
```
demiankatz, line 170: It seems a little dangerous to have this on by default -- libraries that rely on this data for statistics or other reasons might unexpectedly lose it. Would it be better for this to be an opt-in setting? Or at very least should we put a warning on the changelog?

(Practically speaking, there probably aren't too many libraries that have turned on the transaction history setting, which thus means that purging is not a risk... but it still seems like a situation that could potentially cause problems).

demiankatz (changes requested):
A couple of comments before we merge this; also, is it worth adding a unit test for the new method? I realize that the KohaRestTest is very incomplete, but this one doesn't look like it would be too hard to test if we mock out makeRequest, so maybe it's worth covering to start moving the needle in the right direction. :-)

config/vufind/KohaRest.ini
EreMaijala: For any normal use case the user should be able to clear their history since it's something they opted in for in the first place. This should not cause issues with statistics that would be still available, albeit maybe with less precision after anonymization. Also, this is the same setup as with Sierra, so if we end up turning purging off by default, we should do it for both.

demiankatz: If it's an opt-in user feature, then I definitely don't object to keeping this as-is. Thanks for the clarification!

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
EreMaijala: Fixed.

EreMaijala:
@demiankatz Good idea, tests added.

demiankatz (approved):
Looks great! Thanks, @EreMaijala.

PR #2898, opened 2023-05-03
Fix Admin module expiration to use correct date value.

demiankatz:
The refactoring in #2450 did not account for calls made by the Admin module. This PR corrects the oversight.


demiankatz:
Duplicates #2871 -- this copy was temporarily inaccessible due to a GitHub bug which has now been corrected. Closing now since the other PR is already merged.

PR #2870, opened 2023-05-04
Raise version requirement to PHP 8.

demiankatz:
This PR raises our minimum version requirement to PHP 8, since PHP 7 has reached end of life.

Next steps (to be done in separate PRs):

- Raise eligible dependency versions (and test)
- Adjust the PHP version in all file headers (probably best done as a separate PR to reduce confusion)
- Add PHP 8.2 to the build matrix and to branch protection rules (it doesn't work as of this writing, but updating the dependencies may help).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,16 +11,16 @@ jobs:
       VUFIND_LOCAL_DIR: $GITHUB_WORKSPACE/local
     strategy:
       matrix:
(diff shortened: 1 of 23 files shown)


demiankatz:
Note that I've removed the PHP 7.4 build from our branch protection rules, since we won't be able to merge this if a successful 7.4 build is a prerequisite for merges.

EreMaijala (approved):

PR #2875, opened 2023-05-10
KohaRestTest: Fix failure message and setting of cache storage.

EreMaijala:


demiankatz (approved):
Thanks for the fixes!

PR #2877, opened 2023-05-11
Fix numbering of combined results.

EreMaijala:
I made a mistake when fixing HTML id's causing the record number to not be incremented for combined results. This fixes the issue.


demiankatz (approved):
Thanks for the fix, @EreMaijala! I've cherry-picked this back to the release-9.0 branch since it fixes a bug introduced in release 9.0.

PR #2876, opened 2023-05-10
Fix jQuery deprecations

EreMaijala:
Adds an eslint plugin for detecting deprecated functionality and applies automatic and manual fixes.

One remaining issue is that facets use Bootstrap's collapse and `$.support.transition` required to disable transitions when restoring the state is also deprecated. The rule has been disabled for now to allow this.

PR #2878, opened 2023-05-11
Fix and clean up results scripts.

EreMaijala:
- This consolidates initialization of scripts for search results to VuFind.initResultScripts function and adds initializations that were missing from the previous implementation in combined-search.
- results-scripts.phtml is now used for loading all results-related scripts.
- Some additional checks are added to ensure re-triggering initialization doesn't load e.g. covers again.

Extracted from PR #2804

TODO


EreMaijala:
I've tested the changes and run Mink tests and everything seems to work properly.

themes/bootstrap3/templates/search/results.phtml
```diff
@@ -30,26 +30,21 @@
...
+  $this->showCartControls = $options->supportsCart() && $this->cart()->isActive()
     && ($this->showBulkOptions || !$this->cart()->isActiveInSearch());
   // Enable bulk options if appropriate:
   $this->showCheckboxes = $this->showCartControls || $this->showBulkOptions;
-  // Create a css class for results from the search class:
-  $searchClass = $this->params->getSearchClassId();
-  if (!ctype_upper($searchClass)) {
-    $searchClass = preg_replace('/([a-zA-Z])(?=[A-Z])/', '$1-', $searchClass);
-  }
-  $this->resultsClass = 'search-results-' . strtolower($searchClass);
```
demiankatz, line 44: Why are you removing resultsClass here? I think this value is still used in subtemplates like list-list.phtml and list-grid.phtml. Am I missing something?

demiankatz (changes requested):
Thanks, @EreMaijala -- didn't have time for an exhaustive test yet, but I did spot one thing that I wanted to ask about...

themes/bootstrap3/templates/search/results.phtml
EreMaijala: Oops, thanks for spotting that. It was a bad pick from #2804.

demiankatz (changes requested):
@EreMaijala, it looks like merging #2876 has introduced some conflicts here. Can you take a look?

EreMaijala:
@demiankatz Resolved!

themes/bootstrap3/js/common.js
```diff
@@ -258,6 +246,65 @@ var VuFind = (function VuFind() {
     _searchId = searchId;
   };
 
+  function setupQRCodeLinks(_container) {
+    var container = _container || $('body');
+
+    container.find('a.qrcodeLink').click(function qrcodeToggle() {
```
demiankatz: Should this be `.on("click", ...`?

demiankatz (changes requested):
One small deprecation-related comment.

Beyond that, this kind of feels to me like a 10.0 change, because of the change in the behavior of the results-scripts.phtml (which is a file likely to be locally overridden), significant refactoring of Javascript, etc. Would you agree? Is this perhaps the one that prompts us to start dev-10.0, so that we can then pile the AJAX results on top?

themes/bootstrap3/js/common.js
EreMaijala: Yes, fixed.

EreMaijala:
@demiankatz I agree, dev-10.0 would be in order.

demiankatz (approved):
This looks good to me, and all tests are passing. I plan to merge it next week after the `dev-10.0` branch is established.

I propose this text for the changelog:

`- Javascript initialization has been revised and refactored for reduced redundancy and greater flexibility. Significant changes that could impact backward compatibility include refactoring additional logic from search/results.phtml into search/results-scripts.phtml and moving the initResultScripts Javascript function from combined-search.js into common.js. See [[https://github.com/vufind-org/vufind/pull/2878|pull request #2878]] for details.`

PR #2882, opened 2023-05-11
Upgrade php-cs-fixer to version 3.15.1. 

EreMaijala:
This version supports PHP 8.2 and seems to be a drop-in replacement. Newer versions will require more effort.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2883, opened 2023-05-11
Update dependencies for PHP 8.

demiankatz:
This PR is intended to bring all dependencies up to date now that we've raised our minimum PHP requirement to version 8. It may spawn some smaller PRs if any individual dependencies require complex changes -- but I intend to leave this one open until everything is updated, so it can be the catch-all for the "easy stuff."

TODO
- [x] Finish updating dependencies
- [x] Ensure that all tests pass


demiankatz:
Note that we cannot upgrade to diactoros v3 until we raise our PHP requirement to 8.1 due to a conflict with laminas-psr7bridge. I've upgraded diactoros as far as it is currently able to go.

demiankatz:
laminas-recaptcha cannot be updated due to its introduction of a "final" keyword to a class that VuFind extends. See #2911 for work-in-progress on a solution.

demiankatz:
@EreMaijala, I think this is at a point where it's ready to merge. There are a couple of dependencies that I'm not updating here that could theoretically be updated (e.g. swagger-ui, which does not directly relate to PHP at all, and league/commonmark, which @xmorave2 usually updates separately), and there's some further related work-in-progress in #2911, which I plan to finish up soon. But if we merge this, we'll get all the relevant PHP 8 compatibility updates unlocked by raising our minimum version. I see that quite a few packages are already releasing versions with 8.1 as a minimum, so we'll almost certainly want to raise our own requirement further as we get closer to release 10.0... but this is as far as we can get for VuFind 9.1.

EreMaijala (approved):
Looks good, and all tests are passing for me as well!

PR #2886, opened 2023-05-11
Display a message in record view for cached records.

EreMaijala:
The message is displayed for cached records when the cache is used in fallback mode to indicate that the record is not available in the search index. This helps avoid any confusion about missing services and for staff to see a deleted record still remain visible.


demiankatz (approved):
Makes sense to me! Thanks, @EreMaijala.

PR #2887, opened 2023-05-11
Eliminate duplicate list object retrieval logic.

demiankatz:
I happened to notice that we had nearly identical logic for obtaining list objects in two different places, and it was easy to consolidate. However, it requires some signature and visibility changes, so this will have to be done in a major release. Perhaps this is our first dev-10.0 branch PR, when we're ready to begin that work?

I also wonder if at some point we should further consolidate so that the Favorites controller plugin is eliminated and all of that logic goes into the FavoritesService... but this seems like a reasonable first step toward consolidation.

TODO
- [x] Update changelog to document backward-breaking changes when merging

### DIFF
--- a/module/VuFind/src/VuFind/Controller/Plugin/Favorites.php
+++ b/module/VuFind/src/VuFind/Controller/Plugin/Favorites.php
@@ -31,6 +31,7 @@ namespace VuFind\Controller\Plugin;
 
 use VuFind\Db\Row\User;
 use VuFind\Exception\LoginRequired as LoginRequiredException;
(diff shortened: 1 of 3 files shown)


EreMaijala (approved):
Looks good to me! And yes, dev-10.0 would make sense.

demiankatz:
Thanks, @EreMaijala! I'll leave this open for a little while (I don't want to create extra work by opening dev-10.0 prematurely), but as soon as there's a more compelling reason to start the branch, I'll be sure that this one gets merged in.

PR #2888, opened 2023-05-12
Add debug logging of OAuth2 server exceptions.

EreMaijala:
This will help with troubleshooting setup issues. These exceptions are only logged when debug logging is enabled as they are typically caused by a bad request from a client.


demiankatz (approved):
Makes sense to me! Thanks, @EreMaijala!

PR #2889, opened 2023-05-12
Update php-cs-fixer rules for PHP 8

demiankatz:
This PR raises the migration ruleset version to 8.0 for php-cs-fixer and turns on some "risky" PHP 8 rules that should be safe in the context of VuFind.

TODO
- [x] Finish applying rules
- [x] Run full test suite


demiankatz:
@EreMaijala, I think all of these changes improve code clarity, and the test suite is still passing!

I notice that there are still some get_class calls in the code that php-cs-fixer is missing for some reason, but I imagine they'll get picked up in the future when bugs are fixed. For now, at least the stuff it's fixing is an improvement!

EreMaijala (approved):
Awesome! :)

PR #2890, opened 2023-05-12
Fix more jQuery deprecations.

demiankatz:
This is a follow-up to #2876 which catches a few more deprecated calls that were not adjusted there.

In preparing this, I also noticed that we currently depend on a jQuery plugin that uses many deprecated calls; I have opened [VUFIND-1614](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1614) to remind us to deal with that in the next major release.

Would it also be a good idea to open a ticket about eliminating use of Bootstrap 3 collapsing? @EreMaijala / @crhallberg, do you have any suggested alternatives if we go that route?

TODO
- [x] Run full test suite

### DIFF
--- a/themes/bootstrap3/js/embedded_record.js
+++ b/themes/bootstrap3/js/embedded_record.js
@@ -155,7 +155,8 @@ VuFind.register('embedded', function embedded() {
               }
               return ajaxLoadTab(this.id);
             });
-            longNode.find('[id^=usercomment]').find('input[type=submit]').unbind('click').click(
+            longNode.find('[id^=usercomment]').find('input[type=submit]').off("click").on(
+              "click",
               function embeddedComments() {
                 return registerAjaxCommentRecord(longNode);
               }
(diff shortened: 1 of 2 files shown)


demiankatz:
All tests passing, plus I manually double-checked all impacted functionality. I think this should be safe to merge!

EreMaijala (approved):

EreMaijala:
@demiankatz Odd that these were not catched. 

Regarding collapse, I think we the minimal impact approach would be to roll our own jQuery plugin, but that doesn't sound inviting to me. And since we can't keep the data tags quite the same unless we remove Bootstrap's collapse from its package, we will necessarily need some backward-incompatible changes, so going pure JS while making it easy to adjust locally would probably be my choice. Or moving to Bootstrap 5, but that's a bit more intrusive. :) I'm just afraid that these kind of issues will become more unavoidable with bs3.

demiankatz:
The ways of automatic code-fixing tools are mysterious.

I also agree that a pure JS solution for collapsing seems preferable. I'd really like to work on paring back our reliance on Bootstrap as much as possible, and clearly this is one piece of that puzzle.

PR #2891, opened 2023-05-15
Add support for specifying fallback languages independently of defaul…

EreMaijala:
…t language.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2892, opened 2023-05-15
[release-9.0] Check that key pattern is not empty before using it.

EreMaijala:
Lifted from #2857.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2893, opened 2023-05-15
Add getAccountBlocks to PAIA

charbugs:
This PR adds the `getAccountBlocks` API methode to the PAIA driver according to https://vufind.org/wiki/development:plugins:ils_drivers?s[]=getaccountblocks#getaccountblocks


languages/en.ini
```diff
@@ -832,10 +832,21 @@ No pickup locations available = "No pickup locations available"
 No Preference = "No Preference"
 No reviews were found for this record = "No reviews were found for this record"
 No Tags = "No Tags"
+no_change_password_scope = "You are not entitled to change password."
+no_delete_notifications_scope = "You are not entitled to delete notifications."
 no_description = "Description not available."
 no_email_address = "Email address missing."
 no_items_selected = "No Items were Selected"
 no_proxied_user = "No proxied user (request for yourself)"
+no_read_fees_scope = "You are not entitled to read fees."
```
demiankatz: I would recommend moving these messages into the ILSMessages text domain, since I think that's the most appropriate location for these types of messages. This should just be a matter of moving the lines into the appropriate subdirectory of the languages directory, and adding the `ILSMessages::` prefix to the strings when referred to in the code.

I also wonder if the language should be made more consistent. Is there a reason some use "not entitled" while others use "not allowed"? Some of the existing strings simply use "cannot".

config/vufind/PAIA.ini
```diff
@@ -18,6 +18,25 @@ baseUrl = ""
...
 ;paiaCache = false
 
+; VuFind allows to report account blocks to the user. As PAIA blocks implicitly by
+; not issuing PAIA scopes for a patron VuFind would report any possibly missing
+; scope. Therefore with the following setting you can define for which missing scope
+; the user will get a notification in MyResearch views. By default any missing scope
+; results in a notification.
+; For a comprehensive list of PAIA scopes refer to
+; http://gbv.github.io/paia/paia.html#access-tokens-and-scopes .
+;muteAccountBlocksNotificationForScopes[] = "read_patron"
```
demiankatz: Would it be better to make this "opt-in" rather than "opt-out" -- that is, rather than alerting for missing scopes unless they are muted, instead alert only for scopes where alerts are requested? This seems a little safer, since otherwise depending on usage of a system, users might end up getting a lot of false positive block requests. Also, making it opt-in might make it easier to add custom scopes (or newly added ones in the future).

In fact, you could even customize the text of the block messages directly in the config file... what if you used a format like:

`addBlockNotificationForMissingScope[update_patron] = "ILSMessages::no_update_patron_scope"`

This approach might enable you to reduce the amount of code in the driver as well! There seem to be some special cases -- e.g. around patron scopes -- but I still imagine some reduction may be possible.

(Just brainstorming, though -- feel free to ignore the suggestion if you disagree).

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
```diff
@@ -224,6 +232,12 @@ public function init()
         } else {
             $this->debug('Caching not enabled, disabling it by default.');
         }
+
+        if (isset($this->config['PAIA']['muteAccountBlocksNotificationForScopes'])) {
```
demiankatz: This could be reduced to:

`$this->mutedScopes = $this->config['PAIA']['muteAccountBlocksNotificationForScopes'] ?? [];`

module/VuFind/src/VuFind/ILS/Driver/PAIA.php
```diff
@@ -2171,4 +2185,120 @@ protected function paiaDeleteRequest($file, $access_token = null)
...
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     */
+    public function getAccountBlocks($patron)
+    {
+        $blocks = [];
+        if (
+            !$this->paiaCheckScope(self::SCOPE_UPDATE_PATRON)
+            && !in_array(self::SCOPE_UPDATE_PATRON, $this->mutedScopes)
+        ) {
+            array_push(
```
demiankatz: The syntax `$blocks[] = 'no_update_patron_scope';` might make this a bit more concise.

demiankatz (changes requested):
Thanks, @charbugs! See below for some comments and suggestions. I would also like to get a review or two from other PAIA users, but I think it makes sense for us to finish this initial general review before soliciting additional opinions.

lahmann:
@charbugs and me worked on this together, so I fully approve this pull request. Maybe @olli-gold should take a look at it?

charbugs:
Hi Demian,

thank you for the review so far. I have updated the PR according to your suggestions:

* account block configuration where switched to opt-in
* translation has been moved to ILSMessages
* made english translations more concise
* some refactorings

> This approach might enable you to reduce the amount of code in the driver as well! There seem to be some special cases -- e.g. around patron scopes -- but I still imagine some reduction may be possible.

I know, without the special case the code could be a simple loop over a lookup table. But implementing that lookup/loop thing AND handling the special cases might read not so well, I guess.  To reduce the code I added a helper function that does that verbose scope checks. Do you think this is sufficient?

demiankatz:
Thanks for the improvements, @charbugs! How would you feel about this simplification of the getAccountBlocks function?

```
    public function getAccountBlocks($patron)
    {
        $defaultMessages = [ /* this is where you map scope => block message */ ];
        $blocks = [];

        foreach ($this->accountBlockNotificationsForMissingScopes as $scope) {
            if (!$this->paiaCheckScope($scope)) {
                $blocks[$scope] = $defaultMessages[$scope];
            }
        }

        // Special case: if update patron is missing, we don't need to also add
        // more specific messages.
        if (isset($blocks[self::SCOPE_UPDATE_PATRON])) {
            unset($blocks[self::SCOPE_UPDATE_PATRON_NAME]);
            unset($blocks[self::SCOPE_UPDATE_PATRON_EMAIL]);
            unset($blocks[self::SCOPE_UPDATE_PATRON_ADDRESS]);
        }

        return count($blocks) ? array_values($blocks) : false;
    }
```

This uses a keyed array so we can identify the special patron case and clean it up, and then it eliminates the array keys at the very end to return the expected format. I didn't fill in the $defaultMessages array, since I thought that would make my example harder to read... ideally it might make more sense for that to be a class property instead of a local variable, but I just made it local for the sake of simplicity in the example.

Also note that I had suggested using a slightly different configuration syntax... if instead of:

```
accountBlockNotificationsForMissingScopes[] = "read_patron"
```

you used:

```
accountBlockNotificationsForMissingScopes[read_patron] = "ILSMessages:no_read_patron_scope"
```

then it would be easier for people to customize messages through the configuration, and the code could be reduced to:

```
    public function getAccountBlocks($patron)
    {
        $blocks = [];

        foreach ($this->accountBlockNotificationsForMissingScopes as $scope => $message) {
            if (!$this->paiaCheckScope($scope)) {
                $blocks[$scope] = $message;
            }
        }

        // Special case: if update patron is missing, we don't need to also add
        // more specific messages.
        if (isset($blocks[self::SCOPE_UPDATE_PATRON])) {
            unset($blocks[self::SCOPE_UPDATE_PATRON_NAME]);
            unset($blocks[self::SCOPE_UPDATE_PATRON_EMAIL]);
            unset($blocks[self::SCOPE_UPDATE_PATRON_ADDRESS]);
        }

        return count($blocks) ? array_values($blocks) : false;
    }
```

The cost is a perhaps less simple/obvious configuration format, but the result is more flexibility and simpler code with fewer hard-coded messages.

config/vufind/PAIA.ini
```diff
@@ -18,6 +18,21 @@ baseUrl = ""
 ; same as for DAIA cache (see cacheLifetime setting in DAIA.ini).
 ;paiaCache = false
 
+; Specify what missing PAIA scopes should be reported to the user as account blocks.
+; For a comprehensive list of PAIA scopes refer to
+; http://gbv.github.io/paia/paia.html#access-tokens-and-scopes .
+;accountBlockNotificationsForMissingScopes[read_patron] = "no_read_patron_scope"
```
demiankatz: I think all of these values need to have the `ILSMessages::` prefix added. I'd also add a comment explaining the format, like:

```
; Configuration format is accountBlockNotificationsForMissingScopes[scope] = message
; where scope is a PAIA scope name and message is the message to display to the user
; when that scope is missing. Messages will be translated.
```

demiankatz (changes requested):
Thanks for the latest updates. This looks great, but I have a couple of corrections/suggestions related to the configuration file.

demiankatz (approved):
Thanks, @charbugs! This looks great to me now. I'll just wait to merge until @olli-gold has a chance to review (as @lahmann suggested), so we get one more PAIA perspective.

olli-gold (approved):
Nothing to gripe about here - Thank you for the contribution!

demiankatz:

PR #2895, opened 2023-05-17
Switch SolrMarc from log4j to reload4j.

demiankatz:
As discussed on https://github.com/solrmarc/solrmarc/issues/96, this gets rid of the outdated Log4j jar from SolrMarc by switching to the more up-to-date Reload4j fork of Log4j 1.x. In the future, it would be best to switch to a current Log4j 2.x, but until then, this will stop SolrMarc from being flagged by vulnerability detectors (even though the Log4j vulnerabilities are not relevant to a command-line tool like SolrMarc).


EreMaijala (approved):
Looks good to me.

PR #2896, opened 2023-05-17
Make Mink test for record rating more robust.

EreMaijala:
Uses assertEqualsWithTimeout to check for updated rating since it gets updated asynchronously.


demiankatz (approved):
Thanks, @EreMaijala, looks good to me! I have noticed this test failing on rare occasions (though retries always sorted it out), so I'm glad you had time to investigate the cause.

PR #2897, opened 2023-05-17
Add code coverage support to Mink tests.

EreMaijala:
Adds code coverage reporting support to Mink tests for a locally running web server.

This is designed so that it's easy to add support for fetching code coverage results from a remote server later if the need arises, but for now the test runner just tells VuFind where to put the files. An important detail is that any single visit from a Mink test may trigger a number of requests via JavaScript handlers, so it's not only the result of the request triggered by the test's visit to an URL that will need to be recorded.

Prereqs:
- pcov extension with `pcov.enabled=1` or XDebug with `mode=coverage` in PHP ini.
- remote profiling enabled with `SetEnv VUFIND_CODE_COVERAGE 1` in Apache config.

Running:
`./phing.sh phpunit -Dphpunit_remote_coverage=1`

TODO:
- [X] Update https://vufind.org/wiki/development:testing:unit_tests with a section about setting up remote code coverage. Done: https://vufind.org/wiki/development:testing:unit_tests#code_coverage


EreMaijala:
Note that this also contains the fix from #2896.

demiankatz:
Thanks, @EreMaijala! I probably won't have time to review this until tomorrow, but I at least got #2896 merged so if you want to merge dev and reduce diffs here, you're free to do so. :-)

public/index.php
```diff
@@ -13,5 +13,10 @@
     return $app->getServiceManager()
         ->get(\VuFindConsole\ConsoleRunner::class)->run();
 } else {
+    // Setup remote code coverage if enabled:
+    if (getenv('VUFIND_CODE_COVERAGE')) {
+        include __DIR__ . '/../module/VuFind/functions/codecoverage.php';
+        setupRemoteCodeCoverage();
```
demiankatz: Should we include "VuFind" in this function name to better ensure uniqueness? I think that would be consistent with other local function calls made from this file. (In the long term, namespacing the functions might be a cleaner approach, but with a function autoloader coming in a future release of PHP, maybe it's better to wait for that before doing anything fancy).

demiankatz (changes requested):
Thanks, @EreMaijala, this looks great to me! In addition to the comment below, a couple of other notes:

1.) Should we adjust the startup action in build.xml so that it uncomments the relevant httpd-vufind.conf setting when phpunit_remote_coverage is enabled? I think we need something like that in order to easily integrate this into Jenkins.

2.) I notice that you indicated this will be added in VuFind 10 in your wiki page updates... but I don't see a reason why we couldn't add this for 9.1. Am I missing something?

public/index.php
EreMaijala: Added.

EreMaijala:
@demiankatz Latest dev has now been merged. Regarding the other notes:

1. I think I managed to do it.
2. I was perhaps overly cautious. I've now updated the version on the wiki page.

demiankatz (approved):
This looks good to me! I've configured Jenkins to turn on the new coverage option, so once this is merged (which I plan to do as soon as the build passes), we'll see if it works correctly in that environment. Fingers crossed!

PR #2900, opened 2023-05-22
Extra request details from search backend

ThoWagen:
Sometimes it is useful to get the url that was used to perform a search in a controller or template. Therefore, with this PR the extra request details are introduced. Additional information about requests can be acquired via the CallMethodCommand and child classes from the backends. For now, the only implemented use case is getting the url which was used to perform a search in Solr. It can be acquired from Solr/Results. E.g. in the search/resulst.phtml template, adding `<?= $this->results->getExtraSearchDetails()['solrRequestUrl'] ?>`will simply display it.


demiankatz (changes requested):
Thanks for sharing this, @ThoWagen! A couple of quick thoughts:

1.) It might be worth moving the getExtraSearchDetails logic from \VuFind\Search\Solr\Results into \VuFind\Search\Base\Results, since the general functionality is not tied to Solr even though it's currently only implemented there.

2.) In the Results object, I wonder if it's worth calling the details extraBackendDetails or extraDetailsFromBackend or extraSearchResultDetails instead of extraSearchDetails to make it more clear where they're coming from. Just saying "Search" is a little ambiguous in this context.

3.) I wonder if it might be better to make getLastUrl into getLastUrls and have it return an array instead of a string. In at least some situations, issuing a Solr command can cause multiple Solr calls to be made (for example, when multiple spell check indexes are enabled), and as currently implemented, only the last URL called will be retrievable. Maybe this doesn't matter and the extra complexity is undesirable, but it at least seems worth considering.

ThoWagen:
1.) & 2.) I can definitely do this.

3.) That is something I considered, but decided against it. "getLastUrl" is not called after every request and that might lead to an accumulation of multiple urls that have nothing to do with each other. However, the last url in the array could always be used in our use case. I'm not sure about this one.

demiankatz:
Thanks, @ThoWagen. I see what you mean regarding accumulating URLs. I'm fine with sticking with "last URL" for now, and working on "last URLs" when a use case arises. That being said, one possible approach could be to revise the ExtraRequestDetailsInterface to include a `resetExtraRequestDetails` method... then the command could clear out any accumulated extra details before running, ensuring that the details retrieved only apply to the current command. Maybe that's worth implementing as part of the interface to allow for future flexibility, even if we don't need it right away. I'll defer to your preferences on that decision. :-)

ThoWagen:
@demiankatz Thank you for that suggestion. I think this is a good compromise. 

The requested changes are implemented and pushed.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
```diff
@@ -157,6 +164,26 @@ public function getUniqueKey()
...
+    }
+
+    /**
+     * Clears the last url
+     *
+     * @return void
+     */
+    public function resetLastUrl()
+    {
+        $this->lastUrl = null;
```
demiankatz, line 185: I notice that you initialize to `""` in the property (which states a string type) but reset to null here. It would probably be a good idea to be consistent in both places -- either always reset to empty string, or else use the `?string` type to indicate a nullable string, and reset/initialize to null.

module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
```diff
@@ -110,8 +128,12 @@ public function execute(BackendInterface $backend): CommandInterface
             );
         }
         $args = $this->getArguments();
-        return $this->finalizeExecution(
+        $this->finalizeExecution(
```
demiankatz, line 134: Should you add a call to `resetExtraRequestDetails` prior to finalizing execution, to be sure that no values from a previous command bleed into this one?

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php
```diff
@@ -609,6 +609,23 @@ function () use ($connector) {
         );
     }
 
+    /**
+     * Test extra request details
+     *
+     * @return void
+     */
+    public function testExtraRequestDetails()
```
demiankatz, line 618: Worth adding a test for resetExtraRequestDetails as well?

demiankatz (changes requested):
Thanks, @ThoWagen, looking great! Just a few minor details worth polishing...

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
ThoWagen: You are right. I will fix it.

module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
ThoWagen: For now it would not do much, but i guess it makes sense for consistency reasons.

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php
ThoWagen: I am not sure how useful such a test would be, but I can certainly add one.

demiankatz: I agree, it's not incredibly useful at this stage, but if we're adding test coverage for new functionality, we might as well cover it all as long as it's not very difficult. :-)

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Connector.php
ThoWagen: Done. Also changed the type from string to Laminas\Uri\Http since this is the actual type.

module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
ThoWagen: Done

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php
ThoWagen: Done

module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
```diff
@@ -58,6 +59,13 @@ abstract class CallMethodCommand extends AbstractBase
      */
     protected $method;
 
+    /**
+     * Optional search details.
+     *
+     * @var null|array
```
demiankatz: Would `?array` be more concise than `null|array` here and in the associated function?

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php
```diff
@@ -609,6 +609,49 @@ function () use ($connector) {
...
+    {
+        $connector = $this->getConnectorMock(['getLastUrl']);
+        $connector->expects($this->once())->method('getLastUrl')
+            ->will($this->returnValue('someUrl'));
```
demiankatz: Should we return a `\Laminas\Uri\Http` object in these tests to stay in alignment with the types? Or should we convert the Http object to a string when we save it in the connector in order to keep things simpler with basic types?

demiankatz (changes requested):
I think this is nearly ready to go -- just a couple of final thoughts on types.

module/VuFindSearch/src/VuFindSearch/Command/CallMethodCommand.php
ThoWagen: Done

module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/Solr/BackendTest.php
ThoWagen: I think using \Laminas\Uri\Http is nice for the parsing functionality and the Uris can be converted to string implicitly anyway . Changed the tests.

demiankatz (approved):
Looks good to me -- thanks again, @ThoWagen!

PR #2902, opened 2023-05-22
FOLIO: Use shib_cql if Shib is used for login

dltj:
Fixes [VUFIND-1615](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1615).

### DIFF
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -98,7 +98,7 @@ password_field = false
 ; %%password_field%% = The password_field config setting (above)
 ;cql = '%%username_field%% == "%%username%%" and %%password_field%% == "%%password%%"'
 ; If this CQL statement is uncommented, it will be used when a user is logged in
-; via Shibboleth (e.g., the $_SERVER['Shib-Session-ID'] array value exists).
+; via Shibboleth.
 ;shib_cql = 'externalSystemId == "%%username%%"'
 
 ; Should we try to log the user into the Okapi API (true) or just look them
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -97,9 +97,12 @@ password_field = false
 ; %%username_field%% = The username_field config setting (above)
 ; %%password_field%% = The password_field config setting (above)
 ;cql = '%%username_field%% == "%%username%%" and %%password_field%% == "%%password%%"'
-; If this CQL statement is uncommented, it will be used when a user is logged in
-; via Shibboleth.
-;shib_cql = 'externalSystemId == "%%username%%"'
+; The `cql_by_auth_method` setting allows you to specify different CQL queries
+; for different VuFind authentication methods. The keys should be the names
+; of the authentication methods as defined in your VuFind configuration, and
+; the values should be CQL queries that can use the same placeholders as
+; described above for the general `cql` setting. For instance, for Shibboleth:
+;cql_by_auth_method[Shibboleth] = 'externalSystemId == "%%username%%"'
 
 ; Should we try to log the user into the Okapi API (true) or just look them
 ; up in the database using [API] credentials above (false). If set to true,
@@ -108,6 +111,10 @@ okapi_login = true
 ; Should we override the Okapi token created using [API] credentials with the
 ; user's credentials after they log in? (Only valid if okapi_login = true)
 use_user_token = false
+; If set to true, `debug_login` will log the CQL query (with password redated),
+; the environment variables, and the response from the FOLIO `/users` API call.
+;debug_login = false
+
 
 ; This section controls hold behavior; note that you must also ensure that Holds are
 ; enabled in the [Catalog] section of config.ini in order to take advantage of these
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -97,6 +97,9 @@ password_field = false
 ; %%username_field%% = The username_field config setting (above)
 ; %%password_field%% = The password_field config setting (above)
 ;cql = '%%username_field%% == "%%username%%" and %%password_field%% == "%%password%%"'
+; If this CQL statement is uncommented, it will be used when a user is logged in
+; via Shibboleth (e.g., the $_SERVER['Shib-Session-ID'] array value exists).
+;shib_cql = 'externalSystemId == "%%username%%"'
 
 ; Should we try to log the user into the Okapi API (true) or just look them
 ; up in the database using [API] credentials above (false). If set to true,
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -1273,6 +1273,10 @@ class Folio extends AbstractAPI implements
     /**
      * Support method for patronLogin(): authenticate the patron with a CQL looup.
      * Returns the CQL query for retrieving more information about the user.
+     * 
+     * NOTE: this method looks for the existence of a $SERVER['Shib-Session-ID'] variable 
+     * and, if found, looks for a `shib-cql` configuration stanza to use instead of the
+     * standard `cql` stanza.
      *
      * @param string $username The patron username
      * @param string $password The patron password
@@ -1284,9 +1288,16 @@ class Folio extends AbstractAPI implements
         // Construct user query using barcode, username, etc.
         $usernameField = $this->config['User']['username_field'] ?? 'username';
         $passwordField = $this->config['User']['password_field'] ?? false;
-        $cql = $this->config['User']['cql']
-            ?? '%%username_field%% == "%%username%%"'
-            . ($passwordField ? ' and %%password_field%% == "%%password%%"' : '');
+        if (
+            isset($this->config['User']['shib_cql'])
+            && array_key_exists('Shib-Session-ID', $_SERVER)
+        ) {
+            $cql = $this->config['User']['shib_cql'];
+        } else {
+            $cql = $this->config['User']['cql']
+                ?? '%%username_field%% == "%%username%%"'
+                . ($passwordField ? ' and %%password_field%% == "%%password%%"' : '');
+        }
         $placeholders = [
             '%%username_field%%',
             '%%password_field%%',
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -1340,10 +1340,10 @@ class Folio extends AbstractAPI implements
         ];
         $cql_expanded = str_replace($placeholders, $values, $cql);
 
-        $cql_redacted = str_replace($this->escapeCql($password), "XXXX", $cql_expanded);
+        $cql_redacted = str_replace($this->escapeCql($password), 'XXXX', $cql_expanded);
         if ($this->config['User']['debug_login'] ?? false) {
             $this->debug('FOLIO patron login CQL: ' . $cql_redacted);
-            $this->debug("Environment: " . json_encode($_SERVER));
+            $this->debug('Environment: ' . json_encode($_SERVER));
         }
         return $cql_expanded;
     }
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -116,11 +116,11 @@ class Folio extends AbstractAPI implements
     protected $dateConverter;
 
     /**
-     * Auth Manager
+     * Auth method
      *
-     * @var \VuFind\Auth\Manager
+     * @var string
      */
-    protected $authManager = null;
+    protected $authMethod = null;
 
     /**
      * Default availability messages, in case they are not defined in Folio.ini
@@ -157,11 +157,11 @@ class Folio extends AbstractAPI implements
     public function __construct(
         \VuFind\Date\Converter $dateConverter,
         $sessionFactory,
-        \VuFind\Auth\Manager $authManager
+        string $authMethod
     ) {
         $this->dateConverter = $dateConverter;
         $this->sessionFactory = $sessionFactory;
-        $this->authManager = $authManager;
+        $this->authMethod = $authMethod;
     }
 
     /**
@@ -1283,9 +1283,15 @@ class Folio extends AbstractAPI implements
     /**
      * Get the CQL query template for retrieving the patron's information.
      *
-     * It checks if Shibboleth authentication is being used and returns the
-     * corresponding CQL query template. Otherwise, it uses the configured CQL query
-     * to retrieve the patron's information.
+     * This supports both a single CQL query configured for all auth methods, and 
+     * method-specific CQL queries, which can be configured by including the auth 
+     * method name as a key in the `cql_by_auth_method` array in the User section of 
+     * the config file. If method-specific CQL is configured, the driver will use 
+     * the query for the selected auth method if it exists, and fall back to the 
+     * general CQL query if not. If neither is configured, it will fall back to a 
+     * default query that looks for the username and password in the fields 
+     * specified by username_field and password_field in the config file, or username 
+     * and password if those fields are not specified.
      *
      * @param string $usernameField The field to use for the username in the CQL query
      * @param string $passwordField The field to use for password in the CQL query
@@ -1294,16 +1300,9 @@ class Folio extends AbstractAPI implements
      */
     protected function getLoginCql($usernameField, $passwordField)
     {
-        if (
-            isset($this->config['User']['shib_cql'])
-            && $this->authManager->getSelectedAuthMethod() === 'Shibboleth'
-        ) {
-            $cql = $this->config['User']['shib_cql'];
-        } else {
-            $cql = $this->config['User']['cql']
-                ?? '%%username_field%% == "%%username%%"'
-                . ($passwordField ? ' and %%password_field%% == "%%password%%"' : '');
-        }
+        $cql = $this->config['User']['cql_by_auth_method'][$this->authMethod]
+            ?? $this->config['User']['cql']
+            ?? '%%username_field%% == "%%username%%"' . ($passwordField ? ' and %%password_field%% == "%%password%%"' : '');
         $placeholders = [
             '%%username_field%%',
             '%%password_field%%',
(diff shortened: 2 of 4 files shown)


module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -903,9 +907,16 @@ protected function getUserWithCql($username, $password)
         // Construct user query using barcode, username, etc.
         $usernameField = $this->config['User']['username_field'] ?? 'username';
         $passwordField = $this->config['User']['password_field'] ?? false;
-        $cql = $this->config['User']['cql']
-            ?? '%%username_field%% == "%%username%%"'
-            . ($passwordField ? ' and %%password_field%% == "%%password%%"' : '');
+        if (
+            array_key_exists('shib_cql', $this->config['User'])
+            && array_key_exists('Shib-Session-ID', $_SERVER)
```
demiankatz: The Shibboleth session ID is configurable here: https://github.com/vufind-org/vufind/blob/dev/config/vufind/config.ini#L835 -- I wonder if it's worth making config.ini available to the FOLIO driver so it can check for this configuration (or if it's at least worth adding a setting to Folio.ini to allow override of the name of the parameter).

We also try to avoid directly referencing $_SERVER in our code, since Laminas wraps this up in the request object. However, that raises the question of whether it's worth creating a new dependency on the request object for this one feature. Something to think about, at least.

And just as a matter of style, I think `isset($this->config['User']['shib_cql'])` might be more typical of VuFind style than using `array_key_exists`.

demiankatz (commented):
Thanks, @dltj! I've somehow managed to have three different thoughts related to the same if statement -- see below.

Beyond that, I guess the main question is whether this is too specific of a customization for the core distribution. I definitely see the use case, but it relies on a specific combination of authentication and ILS technology which may or may not be commonly found. I'm definitely not automatically opposed to merging this, but that's a factor to consider.

Perhaps a good first step would be to refactor the code so there's a `protected function getLoginCQL` that returns the login CQL statement, since this would make it easier to implement these types of customizations in future, whether or not we decide to adopt this one now.

What do you think?

dltj:
> Beyond that, I guess the main question is whether this is too specific of a customization for the core distribution. I definitely see the use case, but it relies on a specific combination of authentication and ILS technology which may or may not be commonly found. I'm definitely not automatically opposed to merging this, but that's a factor to consider.

I can't speak to the other VuFind drivers, but FOLIO's use of CQL and the customizable capabilities it brings to connecting VuFind to the ILS are quite useful. I know, at least in the NCIP space, it is quite difficult to use combinations of matching identifiers (with one unnamed system allowing for _only_ barcode to be used). This might be a bit of an edge case on one specific driver—and the work-around is not all that hard—but I thought it useful to add it here so someone else doesn't need to work out something equivalent.

> Perhaps a good first step would be to refactor the code so there's a `protected function getLoginCQL` that returns the login CQL statement, since this would make it easier to implement these types of customizations in future, whether or not we decide to adopt this one now.

Ah, now _that_ is a level of abstraction I'm not feeling qualified to comment on. Related to your question above, I'm trying to imagine other combinations of ChoiceAuth authentication selections other than "Shibboleth-ILS" where this would be useful. Maybe it is just that I haven't used the other options.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
dltj: > The Shibboleth session ID is configurable here: https://github.com/vufind-org/vufind/blob/dev/config/vufind/config.ini#L835 -- I wonder if it's worth making config.ini available to the FOLIO driver so it can check for this configuration (or if it's at least worth adding a setting to Folio.ini to allow override of the name of the parameter).

Ah, good point...I thought I had seen that value somewhere. If you can point me in the right direction to "make config.ini available to the FOLIO driver" I'm happy to add that. I'll push a change for the `isset()` versus `array_key_exists()`...this is my lack of PHP experience showing through...I don't know the common idioms; I just know I need something done and I go looking for the language-specific way of doing it.

demiankatz: The easiest way to make the setting available to the FOLIO driver is through the factory... so in FolioFactory.php, you can add something like:

```
$config = $container->get(\VuFind\Config\PluginManager::class)->get('config');
```

and then $config will be an object representation of config.ini that you can either pass directly into the Folio constructor, provide via a setter, etc. The best approach to getting the data into the driver is probably a matter of taste. :-)

And no worries about the isset/array_key_exists thing. In PHP, the general rule is that if you can do something with either a function or a built-in keyword, the keyword approach is usually faster and more concise. But of course that requires you to know all the options available to you. :-)

demiankatz:
I guess what I'm thinking about here is that this customization is specialized for two different reasons: first of all, it behaves differently depending on the authentication method, and secondly, it refers to implementation details of a specific authentication method. Ideally, I'd love to set things up so that we could achieve both of these things in a more generic way, so that this behavior is one possible configuration, but more things are also possible.

One possible improvement could be to fetch the \VuFind\Authentication\Manager service out of the container and use it to determine the active authentication method so we don't have to "sniff" for Shibboleth at all.

I also wonder if there's some advantage to be gained from refactoring code that extracts attributes from authentication methods so that we could fetch those attributes from the authentication manager and use them in situations like this. There are at least two other authentication methods beyond Shibboleth (LDAP and CAS) that can provide user attributes that could be useful in situations like this, so having a generic way to obtain them without caring about the authentication implementation details would be valuable. But I think that's a trickier problem to solve since in different scenarios, the access to the attributes may be more limited, so we might need to cache them in the session or database to ensure that we have them when we need them. I can give this some deeper thought if you think it's a road potentially worth following!

module/VuFind/src/VuFind/ILS/Driver/Folio.php
demiankatz: I also notice that there were two array_key_exists calls here, but only one has been changed to isset. Though perhaps you're holding off on changing the other one since other parts of our discussion may impact the same line of code. :-)

demiankatz:
@dltj, I just noticed that this PR hasn't gone anywhere in quite a while, so I thought I'd follow up and see if this is something you still want to pursue. I'm happy to collaborate as needed -- just checking if this is something we need to revive, or if it is no longer needed.

demiankatz:
@dltj, just doing another review of pull requests without milestones, and I see that this hasn't moved in a while. Is it still needed?

dltj:
Yup —still needed...we have this in place for a number of our libraries, so
I think it is generally useful. It is on my development list.

demiankatz:
Thanks for the confirmation, @dltj! I'm putting an 11.0 milestone on this so we don't lose track of it indefinitely, but also so you don't feel huge pressure to complete it right away. Obviously, if you get to it sooner, we can move the milestone earlier, and if you need more time, we can move it later... but I feel better having it on the schedule somewhere so it doesn't slip through the cracks.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
dltj: I'm working on this because it is tangentially related to something else. (I'm setting a feature flag in the main VuFind configuration in preparation for isolating code to propose as pull requests to the upstream VuFind.) Is this a legitimate way to inject the main VuFind configuration into the ILS driver?

## FolioFactory.php
```php
<?php

namespace IndexDataShared\ILS\Driver;

use Psr\Container\ContainerInterface;

class FolioFactory extends \VuFind\ILS\Driver\FolioFactory
{
    /**
     * Create an object
     *
     * Adds the main VuFind configuration object to the ILS driver.
     *
     * @param ContainerInterface $container     Service manager
     * @param string             $requestedName Service being created
     * @param null|array         $options       Extra options (optional)
     *
     * @return object
     *
     * @throws ServiceNotFoundException if unable to resolve the service.
     * @throws ServiceNotCreatedException if an exception is raised when
     * creating a service.
     * @throws ContainerException&\Throwable if any other error occurs
     */
    public function __invoke(
        ContainerInterface $container,
        $requestedName,
        array $options = null
    ) {
        $mainConfig = $container->get(\VuFind\Config\PluginManager::class)
            ->get('config');
        $driver = parent::__invoke($container, $requestedName, $options);
        if (method_exists($requestedName, 'setMainConfig')) {
            $driver->setMainConfig($mainConfig);
        }
        return $driver;
    }
}
```

## FOLIO driver (in part)
```php
<?php

namespace IndexDataShared\ILS\Driver;

class Folio extends \VuFind\ILS\Driver\Folio
{
    /**
     * VuFind configuration
     *
     * @var array
     */
    protected $mainConfig = null;

    /**
     * Set main configuration
     *
     * Store a reference to the main VuFind configuration
     *
     * @param array $mainConfigArray Configuration array loaded from the main VuFind configuration
     */
    public function setMainConfig($mainConfig)
    {
        $this->mainConfig = $mainConfig;
    }
```

demiankatz: Yes, @dltj, that is an appropriate way to inject the main configuration! A few random thoughts and comments:

1.) In the comments, I would suggest just referring to config.ini instead of saying "main VuFind configuration" -- it's both shorter and more explicit.

2.) It looks like you may still be enforcing an 80-character wrap in some places (or maybe you just copied and pasted some old wrapped code). If the tools are still forcing this on you, you should merge the latest dev branch here, since we raised the limit to 120 characters quite a while ago... but this PR has been sitting around long enough that it might predate that change.

3.) Since you're setting the configuration with a setter instead of via the constructor, you'll need to be careful to account for the possibility of a null value anywhere you read the property. That's not a problem, and you probably already know that, but just mentioning it because it's a detail that can get easily overlooked in the heat of coding. ;-)

4.) When working with ILS drivers, it's always important to remember that there's the possibility of having multiple instances of the same driver via the MultiBackend driver (e.g. in case a consortium needs to connect to multiple FOLIO instances). This is why it's generally preferable to inject settings via the ILS driver configuration instead of the main configuration, since that allows separate instances to be customized individually. In this instance, since you're looking for a global Shibboleth setting, and that seems to me like a higher-level concern than the ILS driver, I think the config.ini injection is probably appropriate (which is why I proposed it in the first place). There's also the option in the future of adding an override in Folio.ini that falls back to config.ini if some crazy use case comes up where we need the ability to differentiate further. But in any case, I'm just mentioning this as an important consideration because once config.ini is injected, it can get increasingly tempting to use it for things, and some of those things may not be the best way of achieving whatever goals are being worked on.

I hope this helps (and sorry for wasting time with anything that didn't need to be said... but I figured better to over-warn than under-warn). :-)

dltj: Ah, very good discussion discussion...

1. Changed the wording in the setter.
2. I've been fighting getting the right PHP formatter to work in VS-Code. I've given up for now and am running `vendor/bin/phping phpcbf` from the command line at the moment.
3. Frankly, I couldn't figure out how to do it via a constructor, but that is my ignorance of the PHP OOP style. Thank you for the warning!
4. You saved the most important for last. I hadn't considered the impact of the MultiBackend driver, and I was already considering using settings in the config.ini for feature flags. To be safe, I guess I'll need to put feature flags in two places...in `Folio.ini` for ILS driver flags and `config.ini` for everything else. It'll probably be rare that we need a MultiBackend installation, but just as I say that the need will probably come up right around the corner.

THANKYOU THANKYOU for all of the comments. I'm about to push a new version of the branch that is rebased on the current dev branch.

demiankatz: Glad to help! Regarding style fixes, note that now that you're up to date with the latest dev branch, you can run `composer fix` to do all of the automatic fixes (php-cs-fixer and phpcbf), and phpcbf should be much faster due to some adjustments to its caching configuration. Hopefully that will make everything more convenient!

If you're interested in learning how to do constructor injection, we can certainly discuss. (I don't think there's a need to change anything, but if you have any questions about the PHP syntax, etc., I'm happy to answer them for your future reference).

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -120,6 +120,13 @@ class Folio extends AbstractAPI implements
         'Open - Awaiting delivery',
     ];
 
+    /**
+     * VuFind configuration, config.ini
```
demiankatz: I'd change this to "Configuration loaded from config.ini"

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -150,6 +157,20 @@ public function setConfig($config)
         $this->tenant = $this->config['API']['tenant'];
     }
 
+    /**
+     * Set main configuration
+     *
+     * Store a reference to the config.ini configuration
+     *
+     * @param array $mainConfig Configuration array loaded from the main VuFind configuration
```
demiankatz: I don't think you need both of the lines above, and I'd adjust the `@param` to reference config.ini.

This is obviously not a big deal, but I'm trying to avoid unnecessary references to VuFind® since if we mention it, we should include the ®, and it's less of a hassle to just avoid mentioning the name. ;-)

demiankatz (changes requested):
A couple more suggested tweaks to comments. I see that you also still need to finish implementing code to use the config.ini settings now that they are accessible. Please request a review from me once you're ready for me to take another look -- I don't want to bug you with any further suggestions before you're actually ready for them. ;-)

demiankatz:
Reassigning milestone as per discussion on the 1 July Community Call. I'm assuming that @dltj does not have time to get this done sooner, but of course we can complete it ahead of the milestone if the opportunity arises -- I just don't want to treat it as a critical requirement for the 11.0 release to be considered "done."

dltj:
This is still a draft. I've tested it against an actual FOLIO server for ILS authentication, but I haven't been able to hook it up to our local Shib IdP (yet).

config/vufind/Folio.ini
```diff
@@ -97,6 +97,9 @@ password_field = false
 ; %%username_field%% = The username_field config setting (above)
 ; %%password_field%% = The password_field config setting (above)
 ;cql = '%%username_field%% == "%%username%%" and %%password_field%% == "%%password%%"'
+; If this CQL statement is uncommented, it will be used when a user is logged in
+; via Shibboleth.
+;shib_cql = 'externalSystemId == "%%username%%"'
```
demiankatz: What if we made this `cql_by_auth_method[Shibboleth]` instead of hard-coding things to Shibboleth? That would make the feature more flexible and avoid hard-coded assumptions about one specific Auth method.

module/VuFind/src/VuFind/ILS/Driver/FolioFactory.php
```diff
@@ -71,6 +71,7 @@ public function __invoke(
             $manager = $container->get(\Laminas\Session\SessionManager::class);
             return new \Laminas\Session\Container("Folio_$namespace", $manager);
         };
-        return parent::__invoke($container, $requestedName, [$sessionFactory]);
+        $authManager = $container->get(\VuFind\Auth\Manager::class);
```
demiankatz, line 74: If the only thing we're using the $authManager for in the driver is to get the current auth method, is it worth considering passing only that value, rather than the whole object? That would make the dependency coupling a little looser.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/FolioTest.php
```diff
@@ -170,9 +170,13 @@ protected function createConnector(string $test, ?array $config = null): void
             $manager = new \Laminas\Session\SessionManager();
             return new \Laminas\Session\Container("Folio_$namespace", $manager);
         };
+        // Create a mock Auth Manager
+        $authManager = $this->getMockBuilder(\VuFind\Auth\Manager::class)
+            ->disableOriginalConstructor()
+            ->getMock();
```
demiankatz: If you take my above suggestion, you won't need this at all, but if you decide to keep it you can simplify this to:

```suggestion
        $authManager = $this->createMock(\VuFind\Auth\Manager::class);
```

demiankatz (changes requested):
Thanks, @dltj -- see below for a few suggestions based on a quick review!

config/vufind/Folio.ini
demiankatz: And just to be clear, this would let you refactor the config logic down to something like:

```
        $cql = $this->config['User']['cql_by_auth_method'][$this->authManager->getSelectedAuthMethod()]
            ?? $this->config['User']['cql']
            ?? '%%username_field%% == "%%username%%"' . ($passwordField ? ' and %%password_field%% == "%%password%%"' : '');
```

dltj:
Hey, Demian — this work is ready for review again. I had an opportunity to test it against a live Shibboleth identity provider, and it checks out. And in the course of working on that problem, I found I needed some information about the process, so I added a `debug_login` flag to the config. With Shibboleth, it is so hard to get everything aligned to debug issues, so having a little extra in the logs has been helpful.

config/vufind/Folio.ini
```diff
@@ -105,6 +111,10 @@ okapi_login = true
 ; Should we override the Okapi token created using [API] credentials with the
 ; user's credentials after they log in? (Only valid if okapi_login = true)
 use_user_token = false
+; If set to true, `debug_login` will log the CQL query (with password redated),
```
demiankatz, line 114: Typo?

```suggestion
; If set to true, `debug_login` will log the CQL query (with password redacted),
```

config/vufind/Folio.ini
```diff
@@ -97,6 +97,12 @@ password_field = false
 ; %%username_field%% = The username_field config setting (above)
 ; %%password_field%% = The password_field config setting (above)
 ;cql = '%%username_field%% == "%%username%%" and %%password_field%% == "%%password%%"'
+; The `cql_by_auth_method` setting allows you to specify different CQL queries
+; for different VuFind authentication methods. The keys should be the names
+; of the authentication methods as defined in your VuFind configuration, and
+; the values should be CQL queries that can use the same placeholders as
+; described above for the general `cql` setting. For instance, for Shibboleth:
```
demiankatz, line 104: I don't think we have a history of using backticks, so maybe we don't need them here; I also wonder if we should avoid saying "VuFind" repeatedly, since it may be implicit. Feel free to disagree if you think my suggested edits reduce clarity, though!

```suggestion
; The cql_by_auth_method setting allows you to specify different CQL queries
; for different authentication methods. The keys should be the names
; of the authentication methods as defined in config.ini, and
; the values should be CQL queries that can use the same placeholders as
; described above for the general cql setting. For instance, for Shibboleth:
```

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -145,13 +152,16 @@ class Folio extends AbstractAPI implements
      * @param \VuFind\Date\Converter $dateConverter  Date converter object
      * @param callable               $sessionFactory Factory function returning
      * SessionContainer object
+     * @param str                    $authMethod     Authentication method
      */
     public function __construct(
         \VuFind\Date\Converter $dateConverter,
-        $sessionFactory
+        $sessionFactory,
+        string $authMethod
```
demiankatz, line 160: Maybe worth using constructor property promotion to simplify code. Also maybe worth defaulting to null for increased back-compatibility (if we want to make it optional). What do you think?

```suggestion
        protected ?string $authMethod = null
```

If you want to make the value mandatory, we should probably push this to dev-12.0 as a breaking change.

demiankatz (changes requested):
Thanks, @dltj, I confess I didn't have time for a super-thorough review right before today's Community Call, but I wanted to take a quick look. See below for some thoughts; I'll review further after you have a chance to respond.

demiankatz:
Moving milestone following discussion with @dltj.

PR #2904, opened 2023-05-23
Improve FOLIO holdings display.

demiankatz:
I noticed that the FOLIO driver is not currently displaying holdings summaries. It looks like this is because when holdings statement retrieval was introduced in #1708, the data was sent to a non-standard `issues` key instead of the standard `summary` key in the holdings array.

This PR adds summary data and does some work to deduplicate it for better displays, since at least in my test instance, duplicate data seems common. I'm leaving the issues key in case somebody relies on it, but perhaps @bbusenius can comment on whether or not that's actually useful/necessary.


bbusenius (approved):
We're definitely using the issues key. Our textual holdings config looks like this:

```
holdings_text_fields[] = 'notes'
holdings_text_fields[] = 'summary'
holdings_text_fields[] = 'holdings notes'
holdings_text_fields[] = 'issues'
holdings_text_fields[] = 'unbound issues'
holdings_text_fields[] = 'unbound supplements'
holdings_text_fields[] = 'supplements'
holdings_text_fields[] = 'indexes'
```
I'm assuming that `summary` just isn't doing anything and we never noticed because issues did what we wanted. I think our need for "issues" rather than "summary" is a matter of nomenclature. That's how our librarians and users expect to see it labeled. I see there's an entry for "Summary" in en.ini. Maybe we just needed to update that? That might be something we missed. In regards to keeping the issues key, we have that logic encapsulated in our local Folio driver override so I don't think we would be affected by the change. Regarding deuping, I don't think we've had any problems with duplicate data here but I don't see any harm in it.

EreMaijala:
Would it make sense to standardize the `issues` key? Our default is `['holdings_notes', 'summary', 'supplements', 'indexes']` and I don't see any problem adding issues to it.

demiankatz:
@EreMaijala, in this situation, "issues" doesn't add anything different from "summary." If we could differentiate between issue-related notes and summary notes, I agree that adding "issues" would help... but in this case we're just putting the same data in two different places. As a result, I'd be inclined to rename "issues" to "summary" in this driver for consistency with the driver spec, since @bbusenius doesn't seem to mind. If something more granular can be achieved in future, I think that's a whole separate project.

EreMaijala:
@demiankatz Okay, understood. :)

demiankatz:
Okay, as discussed yesterday, I've simplified the code to just rename "issues" to "summary" for consistency with the spec, and I've added some additional data to one of our text fixtures to exercise the revised code. I'll merge this as soon as CI passes.

PR #2905, opened 2023-05-23
Clean up coverage tmp directory without recreating it.

EreMaijala:
This allows any previously set permissions to stay the same.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2906, opened 2023-05-25
Add codes to MailExceptions and prevent unnecessary exceptions from being printed

LuomaJuha:
The idea of this pr is to hide too technical errors from the user when i.e a connection can not be established to smtp server.

### DIFF
--- a/module/VuFind/src/VuFind/Controller/AbstractRecord.php
+++ b/module/VuFind/src/VuFind/Controller/AbstractRecord.php
@@ -32,7 +32,6 @@ namespace VuFind\Controller;
 use VuFind\Exception\BadRequest as BadRequestException;
 use VuFind\Exception\Forbidden as ForbiddenException;
 use VuFind\Exception\Mail as MailException;
-use VuFind\Exception\SMS as SMSException;
 use VuFind\RecordDriver\AbstractBase as AbstractRecordDriver;
 use VuFindSearch\ParamBag;
 
@@ -615,7 +614,7 @@ class AbstractRecord extends AbstractBase
                 $sms->text($view->provider, $view->to, null, $body);
                 $this->flashMessenger()->addMessage('sms_success', 'success');
                 return $this->redirectToRecord();
-            } catch (SMSException $e) {
+            } catch (MailException $e) {
                 $this->flashMessenger()->addMessage($e->getDisplayMessage(), 'error');
             }
         }
--- a/module/VuFind/src/VuFind/Controller/AbstractRecord.php
+++ b/module/VuFind/src/VuFind/Controller/AbstractRecord.php
@@ -549,7 +549,7 @@ class AbstractRecord extends AbstractBase
                 $this->flashMessenger()->addMessage('email_success', 'success');
                 return $this->redirectToRecord();
             } catch (MailException $e) {
-                $this->flashMessenger()->addMessage($e->getMessage(), 'error');
+                $this->flashMessenger()->addMessage('email_failure', 'error');
             }
         }
 
--- a/module/VuFind/src/VuFind/Controller/AbstractRecord.php
+++ b/module/VuFind/src/VuFind/Controller/AbstractRecord.php
@@ -32,6 +32,7 @@ namespace VuFind\Controller;
 use VuFind\Exception\BadRequest as BadRequestException;
 use VuFind\Exception\Forbidden as ForbiddenException;
 use VuFind\Exception\Mail as MailException;
+use VuFind\Exception\SMS as SMSException;
 use VuFind\RecordDriver\AbstractBase as AbstractRecordDriver;
 use VuFindSearch\ParamBag;
 
@@ -614,7 +615,7 @@ class AbstractRecord extends AbstractBase
                 $sms->text($view->provider, $view->to, null, $body);
                 $this->flashMessenger()->addMessage('sms_success', 'success');
                 return $this->redirectToRecord();
-            } catch (MailException $e) {
+            } catch (SMSException $e) {
                 $this->flashMessenger()->addMessage($e->getDisplayMessage(), 'error');
             }
         }
--- a/module/VuFind/src/VuFind/Controller/AbstractRecord.php
+++ b/module/VuFind/src/VuFind/Controller/AbstractRecord.php
@@ -549,7 +549,10 @@ class AbstractRecord extends AbstractBase
                 $this->flashMessenger()->addMessage('email_success', 'success');
                 return $this->redirectToRecord();
             } catch (MailException $e) {
-                $this->flashMessenger()->addMessage('email_failure', 'error');
+                $this->flashMessenger()->addMessage(
+                    $e->getCode() ? $e->getMessage() : 'email_failure',
+                    'error'
+                );
             }
         }
 
@@ -615,7 +618,10 @@ class AbstractRecord extends AbstractBase
                 $this->flashMessenger()->addMessage('sms_success', 'success');
                 return $this->redirectToRecord();
             } catch (MailException $e) {
-                $this->flashMessenger()->addMessage($e->getMessage(), 'error');
+                $this->flashMessenger()->addMessage(
+                    $e->getCode() ? $e->getMessage() : 'email_failure',
+                    'error'
+                );
             }
         }
 
--- a/module/VuFind/src/VuFind/Controller/AbstractRecord.php
+++ b/module/VuFind/src/VuFind/Controller/AbstractRecord.php
@@ -549,10 +549,7 @@ class AbstractRecord extends AbstractBase
                 $this->flashMessenger()->addMessage('email_success', 'success');
                 return $this->redirectToRecord();
             } catch (MailException $e) {
-                $this->flashMessenger()->addMessage(
-                    $e->getCode() ? $e->getMessage() : 'email_failure',
-                    'error'
-                );
+                $this->flashMessenger()->addMessage($e->getDisplayMessage(), 'error');
             }
         }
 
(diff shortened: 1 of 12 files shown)


module/VuFind/src/VuFind/Controller/AbstractRecord.php
```diff
@@ -549,7 +549,10 @@ public function emailAction()
                 $this->flashMessenger()->addMessage('email_success', 'success');
                 return $this->redirectToRecord();
             } catch (MailException $e) {
-                $this->flashMessenger()->addMessage($e->getMessage(), 'error');
+                $this->flashMessenger()->addMessage(
+                    $e->getCode() ? $e->getMessage() : 'email_failure',
```
demiankatz: Since this logic is repeated a number of times, do you think it would be better to put it inside a custom method in `VuFind\Exception\Mail`? That would also open up the possibility of implementing different behavior when we're in development mode (i.e. only suppress the technical message when we're _not_ in dev mode). One approach would be to override the `getMessage` method, but the results of doing that might be less predictable than creating a new method like `getSafeUserMessage` or `getDisplayMessage`. Since the code is always specifically catching a `MailException`, it should be safe to rely on the existence of such a method.

demiankatz: I'd also suggest that maybe it's better to explicitly say `$e->getCode() !== ERROR_DEFAULT` rather than relying on the non-obvious fact that `ERROR_DEFAULT` has a falsy value. And maybe `ERROR_DEFAULT` should be `ERROR_UNKNOWN` or `ERROR_TECHNICAL` or something that better reflects its use here.

demiankatz (changes requested):
Thanks, @LuomaJuha -- this is a good idea; just a couple of suggestions for refining it!

module/VuFind/src/VuFind/Controller/AbstractRecord.php
LuomaJuha: Good points, I'll alter the checks.

demiankatz (changes requested):
Thanks, @LuomaJuha! This looks good to me now. One last question: should you do a search through all the code for instances of `VuFind\Exception\Mail` that need additional work? A couple of things I noticed in a quick search of my own:

1.) The SMSMailer also uses the same exception. Should we create a code for the "unknown carrier" scenario? Are there other SMS-related situations we need to worry about?

2.) Some of the unit tests expect Mail exceptions with specific messages. Maybe we should add code-related expectations as well now that codes are set, to provide test coverage of the new behavior.

LuomaJuha:
@demiankatz I adjusted the tests to check for a code number also and added few tests to test that the returned value of an unknown exception will be email_failure.

module/VuFind/src/VuFind/SMS/Clickatell.php
```diff
@@ -79,14 +79,14 @@ public function text($provider, $to, $from, $message)
         try {
             $result = $this->client->setMethod('GET')->setUri($url)->send();
         } catch (\Exception $e) {
-            throw new MailException($e->getMessage());
+            throw new MailException($e->getMessage(), MailException::ERROR_UNKNOWN);
         }
         $response = $result->isSuccess() ? trim($result->getBody()) : '';
         if (empty($response)) {
-            throw new MailException('Problem sending text.');
+            throw new MailException('Problem sending text.', MailException::ERROR_UNKNOWN);
```
demiankatz: Isn't this change going to cause the error message to say "Problem sending email" instead of "Problem sending text"? That might cause confusion.

module/VuFind/src/VuFind/SMS/Clickatell.php
```diff
@@ -171,9 +171,10 @@ protected function formatMessage($message)
     {
         // Clickatell expects UCS-2 encoding:
         if (!function_exists('iconv')) {
-            // @codeCoverageIgnoreStart
-            throw new MailException('Clickatell requires iconv PHP extension.');
-            // @codeCoverageIgnoreEnd
+            throw new MailException(
+                'Clickatell requires iconv PHP extension.',
+                MailException::ERROR_EXTENSION_MISSING
```
demiankatz: This seems like a more technical error, and not something meaningful to the end user -- maybe this should just stick with the UNKNOWN code.

module/VuFind/tests/unit-tests/src/VuFindTest/SMS/ClickatellTest.php
```diff
@@ -76,6 +76,41 @@ public function testCarriers()
...
     }
 
+    /**
+     * Test unknown exception message error
+     *
+     * @return void
+     */
+    public function testUnknownException()
+    {
+        $client = $this->getMockClient($this->getMockClient(), []);
```
demiankatz: I think something is wrong here -- not sure why `getMockClient` is getting passed in to `getMockClient`.

demiankatz (changes requested):
Thanks, @LuomaJuha -- I spotted a couple of small issues in the latest round of changes. (I think you fixed one of them while I was writing my comment about it, so ignore anything that's out of date. :-) ).

The other general question is that I'm perhaps second-guessing my suggestion to use getDisplayMessage instead of overriding getMessage. It certainly has the advantage of avoiding unanticipated side effects, but it also has the disadvantage of requiring us to remember to use the right method call if we want to display a "safe" message. I'm not sure which is more important. Maybe @EreMaijala has an opinion if we need another one.

module/VuFind/src/VuFind/SMS/Clickatell.php
LuomaJuha: Oh, good catch. Going to have to write a case for it too.

module/VuFind/src/VuFind/SMS/Clickatell.php
LuomaJuha: Check

LuomaJuha:
@demiankatz Hmm, yea. It could be ok to just overwrite the function as the error message will be displayed in dev mode / in log. I think there is not too many changes in the logic of the function.

LuomaJuha:
@demiankatz oh, seems like the getMessage function is final.

demiankatz:
> @demiankatz oh, seems like the getMessage function is final.

Well, I guess that makes up our mind for us! :-)

module/VuFind/src/VuFind/Exception/Mail.php
```diff
@@ -40,4 +40,66 @@
...
+     *
+     * @return string
+     */
+    public function getDisplayMessage(): string
+    {
+        // If application env is development, we can display too technical messages
+        if ('development' === APPLICATION_ENV || $this->getCode() !== self::ERROR_UNKNOWN) {
+            return $this->getMessage();
+        }
+        return 'email_failure';
```
demiankatz: It still troubles me a bit that when sending SMS, we'll display an email failure message by default. I think perhaps we need a different exception class for SMS. What if we change the default display message so that it's a property -- `i.e. protected $defaultDisplayMessage = 'email_failure';` and then we can create a new `VuFind\Exception\SMS` as a subclass of `VuFind\Exception\Mail` which overrides the defaultDisplayMessage to `'Problem sending text.'`. Then the SMS code can throw more specific exceptions, and everything else can keep working as it is.

Am I overthinking this? Maybe that's too much work for a minor point. But let me know if you think it's worth doing!

demiankatz (changes requested):
Thanks for the progress -- one last thought (I promise!).

module/VuFind/src/VuFind/Exception/Mail.php
LuomaJuha: Well, I think it would be ok and it should not be too much of a hazzle. I can add that also.

module/VuFind/src/VuFind/Controller/AbstractRecord.php
```diff
@@ -614,8 +615,8 @@ public function smsAction()
                 $sms->text($view->provider, $view->to, null, $body);
                 $this->flashMessenger()->addMessage('sms_success', 'success');
                 return $this->redirectToRecord();
-            } catch (MailException $e) {
-                $this->flashMessenger()->addMessage($e->getMessage(), 'error');
+            } catch (SMSException $e) {
```
demiankatz: I think it's okay to leave this as a MailException, since that's more general than SMSException and will catch both types -- this way, if either type is thrown, the correct behavior will occur. I think it's possible that sending an SMS will lead to a MailException being thrown, if something goes wrong in lower-level mail code.

module/VuFind/src/VuFind/SMS/Clickatell.php
```diff
@@ -79,14 +79,14 @@ public function text($provider, $to, $from, $message)
         try {
             $result = $this->client->setMethod('GET')->setUri($url)->send();
         } catch (\Exception $e) {
-            throw new MailException($e->getMessage());
+            throw new SMSException($e->getMessage(), SMSException::ERROR_UNKNOWN);
         }
         $response = $result->isSuccess() ? trim($result->getBody()) : '';
         if (empty($response)) {
-            throw new MailException('Problem sending text.');
+            throw new SMSException('Problem sending text.', SMSException::ERROR_RESPONSE_UNKNOWN);
```
demiankatz: Now that the SMSException provides an appropriate user message, I think we can change this back to ERROR_UNKNOWN and eliminate the ERROR_RESPONSE_UNKNOWN type. There is no translation defined for 'Problem sending text.' so that should not be used for end-user display. (When I first brought this up, I assumed that it was translated and didn't realize that sms_error was the correct string to use for end users).

demiankatz (changes requested):
Thanks, @LuomaJuha, this is great. I just spotted a couple more problems (most of which already existed before these changes, but which were brought into clarity by my latest review). I'm pretty sure after one more round of changes, this will be ready to merge. Thanks again for your patience with the process! :-)

module/VuFind/src/VuFind/Controller/AbstractRecord.php
LuomaJuha: Sure thing!

demiankatz (approved):
Looks good to me -- thanks, @LuomaJuha!

PR #2907, opened 2023-05-25
Upgrade php-cs-fixer to v3.17.0.

demiankatz:
I had to disable the `statement_indentation` rule for templates to avoid lots of problems, but at least some of these other changes do result in cleaner code.

TODO
- [x] Run full test suite.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Virtua.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Virtua.php
@@ -692,12 +692,11 @@ class Virtua extends AbstractBase implements \VuFindHttp\HttpServiceAwareInterfa
                 $return_string .= $d . " " . $data['data'][$i] . " ";
                 $i++;
             }
-
+       } else {
             // Chrono
             // Important note: strtotime() expects
             // 01/02/2000 = 2nd Jan 2000
             // 01-02-2000 = 1st Feb 2000 <= Use hyphens
-        } else {
             $pattern = implode("", $data['pattern']);
             switch (strtolower(trim($pattern))) {
                 // Error case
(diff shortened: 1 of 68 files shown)


demiankatz:
All tests are passing, though I plan to do another review of my changes next week with a fresher brain, just to be on the safe side. :-)

module/VuFind/src/VuFind/ILS/Driver/Virtua.php
```diff
@@ -693,10 +693,10 @@ protected function renderPartSubPattern($data)
                 $i++;
             }
 
-        // Chrono
-        // Important note: strtotime() expects
-        // 01/02/2000 = 2nd Jan 2000
-        // 01-02-2000 = 1st Feb 2000 <= Use hyphens
+            // Chrono
```
EreMaijala, line 696: I think it would be cleaner to move these inside the next `else` block.

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -125,7 +125,7 @@ public function getAvailableDigitalFormats()
                     $formatType = $format->formatType;
                     $formats[$formatType] = $formatNames[$formatType];
                 }
-            // If we aren't locked in, we can show all formats
+                // If we aren't locked in, we can show all formats
```
EreMaijala: I think it would be cleaner to rephrase this so that it can be moved into the next `else` block, like `// Not locked in, show all formats`.

EreMaijala (changes requested):
I agree that the results are good. I had just a couple of minor issues with indented comments.

demiankatz:
Thanks, @EreMaijala, I've made the requested adjustments! I wish they'd make up their mind about comment indentation, because I feel this has switched back and forth a few times, but your suggestions will make this more predictable in future even if it changes again. :-)

demiankatz:
Hmm, that's weird. I edited both files through the web-based editor here since the changes were trivial... but that somehow messed up the indentation of the else clauses. Fixed now. Not sure exactly what I did to introduce that indentation problem!

EreMaijala (approved):

EreMaijala:
@demiankatz Odd, but now it seems to be all good.

PR #2908, opened 2023-05-29
Display custom hidden filters as a temporary search tab.

EreMaijala:
Previously the default tab was displayed as active when current hidden filters did not match any of the tabs, which was quite misleading and confusing. This change makes it possible to display a temporary custom tab in this case, or optionally not display any tab as active by emptying the translation.


EreMaijala:
I can't recall the exact details, but I believe I originally did this because we didn't have the sid parameter and had to "guess" the tab in some cases. In any case I've stumbled on this myself at least when updating SearchTabsFilters in config.ini and trying to check the results without starting from the home screen. This can also occur if filters are updated and a saved search no longer matches any of the tabs, as well as for hand-crafted search URL's that contain hidden filters.

demiankatz (approved):
Makes sense to me -- thanks, @EreMaijala!

PR #2910, opened 2023-05-30
Improve scope of lessToSass; delete unused files.

demiankatz:
I noticed while reviewing #2909 that the lessToSass action was not descending into subdirectories of less/components, and also that we had committed some backup and test files that don't need to be in the repo. This PR corrects both problems.


EreMaijala (approved):

demiankatz:
Thanks, @EreMaijala!

PR #2911, opened 2023-05-30
Proxy laminas-recaptcha instead of extending it.

demiankatz:
laminas-recaptcha 2.5.0 makes Laminas\ReCaptcha\ReCaptcha final, so we can no longer extend it. This PR accomplishes the same result with proxying.

TODO:
- [x] Test!
- [x] Discuss: does this need to wait for release 10.0 or can/should we merge earlier?


demiankatz:
I've done hands-on testing of this PR and can confirm that ReCaptcha v2 works identically using the old code and the revisions here. I've simplified the logic in the proxied version of the code, since some of the changes seem to be attempting to apply changes that are now part of the core Laminas code. The only thing we really need to do at this point is disable script tags for proper integration with @mtrojan-ub's pluggable Captcha system.

I haven't tested ReCaptcha v3, but if that requires more work, that's a separate project -- my goal here is simply to retain existing functionality while upgrading the Laminas library, and I believe that I have accomplished that successfully.

The remaining question is whether we defer this change until v10 because it is backward-breaking, or if we include it in v9.1 because the likelihood of a conflict with local customizations is very small, and it enables us to upgrade dependencies sooner. Any opinions?

mtrojan-ub (approved):
Thanks, @demiankatz!

I woulnd't mind adding it to the 9.1, since it seems to be only a minor backwards compatility break that should not cause too much trouble,  and i also think that the likelihood of customizations is pretty small here.

demiankatz:
Thanks, @mtrojan-ub. I'm inclined to agree that the risks are low -- I think this is more likely to cause problems if we defer it (because Composer will keep telling us to upgrade laminas-recaptcha, which will then break the build) than if we merge it now (which I can address with a changelog note, which I will do momentarily).

PR #2912, opened 2023-05-30
Clean up legacy OCI8 logic.

demiankatz:
I'm not sure if anyone is in a position to actually test this anymore now that we've migrated away from Voyager, but this PR cleans out some old PHP7-specific logic now that we're raising our minimum version to PHP8. It also adds a new "provides" setting to composer.json so that the oci8 PHP extension is not mandatory.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
     },
     "provide": {
         "ext-memcached": "*",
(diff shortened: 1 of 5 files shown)


EreMaijala (approved):
This all looks good to me even if it cannot be tested.

PR #2915, opened 2023-05-31
SierraRest: Add more configuration options for hold availability

EreMaijala:
New settings have been added to control holdability of items and bibs. Some of the code has been refactored to be reusable and requests to be cacheable.

More specfically, item retrieval has been extracted from getItemStatusesForBiblio, and its results are used in checkRequestIsValid/checkTitleHoldRules as well to avoid requesting the same information multiple times. The items are only cached for 10 seconds by default to avoid displaying stale information. It would be nice to have a shortcut for returning title holdability directly with getHolding, but this is the best I could manage without touching the upper level logic.

The newly added explodeSetting method simplifies initialization of colon separated settings, and I intend to follow up with extended use of it.

### DIFF
--- a/config/vufind/SierraRest.ini
+++ b/config/vufind/SierraRest.ini
@@ -122,6 +122,10 @@ item_hold_excluded_item_codes = "e"
 ; holds.
 ;item_hold_excluded_item_types = "215:251"
 
+; A colon separated list of bib levels that enable item level holds. If not defined,
+; title_hold_bib_levels is used instead.
+;item_hold_bib_levels = "a:b:m:d"
+
 ; A colon separated list of bib levels that enable title level holds
 title_hold_bib_levels = "a:b:m:d"
 
--- a/config/vufind/SierraRest.ini
+++ b/config/vufind/SierraRest.ini
@@ -82,11 +82,6 @@ password_field = "pin"
 ; settings. Additional notes about some of these settings are available in the wiki:
 ; https://vufind.org/wiki/configuration:ils:holds
 [Holds]
-; If a colon separated list of item statuses is provided here, only matching items
-; will show hold links.  Skip this setting to allow all statuses.
-; Note that VuFind-style statuses are used here.
-;valid_hold_statuses = 'On Shelf:Charged'
-
 ; HMACKeys - A list of hold form element names that will be analyzed for consistency
 ; during hold form processing. Most users should not need to change this setting.
 HMACKeys = item_id:holdtype:level
@@ -113,16 +108,32 @@ defaultPickUpLocation = ""
 ; By default item holds are enabled. Uncomment this setting to disable item holds.
 ;enableItemHolds = false
 
-; This setting specifies which item codes disable item level holds
-item_hold_excluded_item_codes = e
+; If a colon separated list of item statuses is provided here, only matching items
+; will show hold links.  Skip this setting to allow all statuses.
+; Note that VuFind-style statuses are used here.
+;valid_hold_statuses = 'On Shelf:Charged'
 
-; This setting specifies which bib levels allow title level holds
-title_hold_bib_levels = a:b:m:d
+; A colon separated list of item codes (ICODE2 in Sierra) that disable item level
+; holds.
+item_hold_excluded_item_codes = "e"
 
-; By default a title hold can be placed even when there are no items. Uncomment this
-; to prevent holds if no items exist. If request groups are enabled, item existence
-; is checked only in the selected request group.
-;checkItemsExist = true
+; A colon separated list of item types (I TYPE in Sierra) that disable item level
+; holds.
+;item_hold_excluded_item_types = "215:251"
+
+; A colon separated list of bib levels that enable title level holds
+title_hold_bib_levels = "a:b:m:d"
+
+; If a colon separated list of rules is provided here, title level hold is only
+; displayed for records matching any of the rules.
+; Valid rules are:
+; item                    The title must have at least one item
+; holdable_item           The title must have at least one item that matches the
+;                         item holdability rules (regardless of whether item holds
+;                         are enabled)
+; order                   The title must have at least one active order (order
+;                         record with "ON ORDER" status)
+;title_hold_rules = "holdable_item:order"
 
 ; Fields that can be updated for holds (unless the hold is already in transit or
 ; available for pickup). Supported values are "frozen" and "pickUpLocation".
--- a/config/vufind/SierraRest.ini
+++ b/config/vufind/SierraRest.ini
@@ -164,7 +164,7 @@ title_hold_bib_levels = "a:b:m:d"
 ;allowCancelingAvailableRequests = true
 
 ; This section allows modification of the default mapping from item status codes to
-; VuFind item statuses. Defaults are
+; VuFind item statuses. Defaults are listed below.
 [ItemStatusMappings]
 ;! = "On Holdshelf"
 ;t = "In Transit"
--- a/config/vufind/SierraRest.ini
+++ b/config/vufind/SierraRest.ini
@@ -110,7 +110,8 @@ defaultPickUpLocation = ""
 
 ; If a colon separated list of item statuses is provided here, only matching items
 ; will show hold links.  Skip this setting to allow all statuses.
-; Note that VuFind-style statuses are used here.
+; Note that mapped status strings are used here (see the [ItemStatusMappings] for
+; more information).
 ;valid_hold_statuses = 'On Shelf:Charged'
 
 ; A colon separated list of item codes (ICODE2 in Sierra) that disable item level
@@ -127,12 +128,12 @@ title_hold_bib_levels = "a:b:m:d"
 ; If a colon separated list of rules is provided here, title level hold is only
 ; displayed for records matching any of the rules.
 ; Valid rules are:
-; item                    The title must have at least one item
-; holdable_item           The title must have at least one item that matches the
-;                         item holdability rules (regardless of whether item holds
-;                         are enabled)
-; order                   The title must have at least one active order (order
-;                         record with "ON ORDER" status)
+;   item           The title must have at least one item
+;   holdable_item  The title must have at least one item that matches the
+;                  item holdability rules (regardless of whether item holds are
+;                  enabled)
+;   order          The title must have at least one active order (order record with
+;                  "ON ORDER" status)
 ;title_hold_rules = "holdable_item:order"
 
 ; Fields that can be updated for holds (unless the hold is already in transit or
@@ -159,7 +160,7 @@ title_hold_bib_levels = "a:b:m:d"
 ;allowCancelingAvailableRequests = true
 
 ; This section allows modification of the default mapping from item status codes to
-; VuFind item statuses
+; VuFind item statuses. Defaults are
 [ItemStatusMappings]
 ;! = "On Holdshelf"
 ;t = "In Transit"
@@ -173,9 +174,10 @@ title_hold_bib_levels = "a:b:m:d"
 ;s = "On Search"
 ;d = "In Process"
 ;- = "On Shelf"
-; This is a special case -- not an official Sierra code, but a VuFind-specific way
-; of defining the "checked out" message text:
+; These are special cases -- not official Sierra codes, but a VuFind-specific way
+; of defining the "checked out" and "ordered" message texts:
 ;Charged = "Charged"
+;Ordered = "Ordered"
 
 ; This section allows mapping of patron block codes to VuFind translation strings
 [PatronBlockMappings]
--- a/module/VuFind/src/VuFind/ILS/Driver/AbstractBase.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/AbstractBase.php
@@ -81,4 +81,28 @@ abstract class AbstractBase implements DriverInterface
     ): void {
         throw new ILSException($msg ?? $exception->getMessage(), 0, $exception);
     }
+
+    /**
+     * Explode a delimited setting to an array
+     *
+     * @param string $value     Setting value
+     * @param bool   $trim      Whether to trim the values
+     * @param string $separator Separator
+     *
+     * @return array
+     */
+    protected function explodeSetting(
+        string $value,
+        $trim = false,
+        string $separator = ':'
+    ): array {
+        if ('' === $value) {
+            return [];
+        }
+        $result = explode($separator, $value);
+        if ($trim) {
+            $result = array_map('trim', $result);
+        }
+        return $result;
+    }
 }
--- a/module/VuFind/src/VuFind/ILS/Driver/AbstractBase.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/AbstractBase.php
@@ -81,28 +81,4 @@ abstract class AbstractBase implements DriverInterface
     ): void {
         throw new ILSException($msg ?? $exception->getMessage(), 0, $exception);
     }
-
-    /**
-     * Explode a delimited setting to an array
-     *
-     * @param string $value     Setting value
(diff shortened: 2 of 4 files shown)


module/VuFind/src/VuFind/ILS/Driver/AbstractBase.php
```diff
@@ -81,4 +81,28 @@ protected function throwAsIlsException(
...
+    /**
+     * Explode a delimited setting to an array
+     *
+     * @param string $value     Setting value
+     * @param bool   $trim      Whether to trim the values
+     * @param string $separator Separator
+     *
+     * @return array
+     */
+    protected function explodeSetting(
```
demiankatz: Would it make sense to move this to `\VuFind\Config\ExplodeSettingTrait` so it can also be used outside of ILS drivers? I suspect we're doing this kind of exploding all over the place.

config/vufind/SierraRest.ini
```diff
@@ -113,16 +108,32 @@ defaultPickUpLocation = ""
...
-title_hold_bib_levels = a:b:m:d
+; If a colon separated list of item statuses is provided here, only matching items
+; will show hold links.  Skip this setting to allow all statuses.
+; Note that VuFind-style statuses are used here.
+;valid_hold_statuses = 'On Shelf:Charged'
 
-; By default a title hold can be placed even when there are no items. Uncomment this
-; to prevent holds if no items exist. If request groups are enabled, item existence
-; is checked only in the selected request group.
-;checkItemsExist = true
```
demiankatz, line 125: I notice that you've completely removed this setting -- was this something that was never fully implemented? I don't see any code related to the setting in SierraRest.php, but there is a property with a corresponding name that doesn't seem to be used for anything. Should that be removed too?

module/VuFind/src/VuFind/ILS/Driver/AbstractBase.php
```diff
@@ -81,4 +81,28 @@ protected function throwAsIlsException(
...
+        string $value,
+        $trim = false,
```
demiankatz: Is there a compelling reason NOT to trim by default? I can't think of a scenario where we would want to keep random whitespace, and I suspect that the only reason the code doesn't already trim everywhere is laziness/inconsistency.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -290,35 +321,23 @@ public function init()
...
-            = $this->config['Holds']['defaultPickUpLocation'] ?? '';
+            = $holdCfg['allowCancelingAvailableRequests'] ?? false;
+        $this->defaultPickUpLocation = $holdCfg['defaultPickUpLocation'] ?? '';
         if ($this->defaultPickUpLocation === 'user-selected') {
             $this->defaultPickUpLocation = false;
         }
-
-        $this->checkFreezability
-            = !empty($this->config['Holds']['checkFreezability']);
+        $this->checkFreezability = $holdCfg['checkFreezability'] ?? false;
```
demiankatz: Do we want to cast this to a boolean? The previous logic guaranteed a boolean value, but this approach does not.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -2732,49 +2712,149 @@ function ($matches) {
         return $msg;
     }
 
+    /**
+     * Get a cached record and check that it has the requested fields
+     *
+     * @param string $cacheId Cache entry ID
+     * @param array  $fields  Fields
```
demiankatz: Maybe adding a word to this description would be helpful -- e.g. Requested fields or Required fields?

demiankatz (changes requested):
Thanks, @EreMaijala. Since I no longer have access to a Sierra system, I can't test this, but here are a few comments based on a quick review.

module/VuFind/src/VuFind/ILS/Driver/AbstractBase.php
EreMaijala: I considered this and decided to err on the side of caution because particularly with codes from an ILS, a single space could be a valid one. We could of course try to do something smart to allow an intentional space while trimming everything else, but I'm not sure if that's worth the complexity.

Another point of view is that this code will potentially be executed millions of times, while you only have to get the config right once. :)

demiankatz: All good points. I won't argue against them. :-)

module/VuFind/src/VuFind/ILS/Driver/AbstractBase.php
EreMaijala: Good idea, done.

config/vufind/SierraRest.ini
EreMaijala: It was never implemented at all, but I forgot to remove the property. Now done.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Cast added.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: I revised this and other descriptions as well.

config/vufind/SierraRest.ini
```diff
@@ -113,16 +108,32 @@ defaultPickUpLocation = ""
...
 ;enableItemHolds = false
 
-; This setting specifies which item codes disable item level holds
-item_hold_excluded_item_codes = e
-
-; This setting specifies which bib levels allow title level holds
-title_hold_bib_levels = a:b:m:d
+; If a colon separated list of item statuses is provided here, only matching items
+; will show hold links.  Skip this setting to allow all statuses.
+; Note that VuFind-style statuses are used here.
```
demiankatz: Should we direct users to a place where they can find a list of these -- even if just a reference to a particular method? I'm not sure it will be immediately obvious what "VuFind-style statuses" means otherwise.

config/vufind/SierraRest.ini
```diff
@@ -113,16 +108,32 @@ defaultPickUpLocation = ""
...
+; holds.
+;item_hold_excluded_item_types = "215:251"
+
+; A colon separated list of bib levels that enable title level holds
+title_hold_bib_levels = "a:b:m:d"
+
+; If a colon separated list of rules is provided here, title level hold is only
+; displayed for records matching any of the rules.
+; Valid rules are:
+; item                    The title must have at least one item
```
demiankatz: I think indenting the values (item, etc.) a couple of spaces might make this list a little more readable so that they stand out better from the "Valid rules are:" header.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -58,8 +58,42 @@ class SierraRest extends AbstractBase implements
     use \VuFind\I18n\Translator\TranslatorAwareTrait;
     use \VuFind\I18n\HasSorterTrait;
     use \VuFind\Service\Feature\RetryTrait;
+    use \VuFind\Config\ExplodeSettingTrait;
 
-    public const HOLDINGS_LINE_NUMBER = 40;
+    /**
+     * Fixed field number for location in holdings records
+     *
+     * @var int
```
demiankatz: The type says int, but the value is a string. Intentional? (Same for all the constants below).

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -241,6 +282,49 @@ class SierraRest extends AbstractBase implements
...
+     */
+    protected $bibCacheTTL = 300;
+
+    /**
+     * Life time in seconds for cached items of a bibliographic record
+     *
+     * It is recommended to keep this fairly short to ensure that any recent changes
+     * (such as placing a hold) are reflected correctly in holdings.
```
demiankatz, line 303: Wouldn't it be possible to fully place a hold in less than 10 seconds and see confusing results? Is it possible that this value should be even shorter, if we're just concerned about caching within a single request?

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -2120,91 +2200,59 @@ protected function getItemStatusesForBib($id, $checkHoldings)
...
+            if ($notes) {
+                $entry['item_notes'] = $notes;
+            }
 
-                $locationCode = $item['location']['code'] ?? '';
-                if (!empty($holdingsData[$locationCode])) {
-                    $entry += $this->getHoldingsData($holdingsData[$locationCode]);
-                    $holdingsData[$locationCode]['_hasItems'] = true;
-                }
+            if ($this->itemHoldsEnabled && $this->isHoldable($item, $bib)) {
```
demiankatz, line 2251: Would it make more sense to check itemHoldsEnabled inside isHoldable and just return false when it's disabled?

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -2568,44 +2616,35 @@ protected function getItemStatus($item)
      * Determine whether an item is holdable
      *
      * @param array $item Item from Sierra
+     * @param array $bib  Bib record from Sierra
      *
      * @return bool
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
```
demiankatz: Why is `$bib` unused? I assume this is intended as a hook for local customizations, but I'm wondering if perhaps there's more logic somewhere that should be incorporated into this method.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -2732,49 +2771,141 @@ function ($matches) {
...
     /**
      * Fetch fields for a bib record from Sierra
      *
      * Note: This method can return cached data
      *
-     * @param int    $id     Bib record id
-     * @param string $fields Fields to request
+     * @param string $id     Bib record id
+     * @param ?array $fields Fields to request or null for defaults
      * @param array  $patron Patron information, if available
```
demiankatz: This comment needs to be aligned with the `?array` type below.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -2732,49 +2771,141 @@ function ($matches) {
...
     }
 
+    /**
+     * Get all items for a bib record
+     *
+     * Note: This method can return cached data
+     *
+     * @param string $id     Bib record id
+     * @param ?array $fields Fields to request or null for defaults
+     * @param array  $patron Patron information, if available
```
demiankatz: Another inconsistent doc comment type.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3086,4 +3217,41 @@ protected function getItemsWithBibsForTransactions(
...
+        }
+
+        if (
+            in_array('order', $this->titleHoldRules)
+            && !empty($bib['orders'])
+        ) {
+            return true;
+        }
+        if (
+            in_array('item', $this->titleHoldRules)
```
demiankatz: Maybe this would be more readable if you assigned the in_array results to variables -- then you wouldn't have to search for 'item' twice, and you could incorporate the existence of the holdable_item rule into the if statement for isHoldable, which is not currently necessary but might make the code more explicit (and prevent possible bugs if more rules are added in future).

demiankatz (changes requested):
Thanks for the improvements, @EreMaijala. Since I can't test this myself, I've done a very careful code read and left a lot of nitpicky comments. I hope they're more helpful than annoying. ;-)

The other obvious question is whether any of this is worth adding tests for. We don't have much of a SierraRestTest built up yet, though, so I understand if now is not the right time for that task.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: My bad. They are all numeric but come in as strings.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Yes, that's possible. I reduced it to two seconds. In the long run I'd like to make it possible for getHolding to return title holdability right away to shortcut the need to go through checkRequestIsValid, but that'll need a bit more work.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Yes, but then I'd need an override param for the check to work in checkRequestIsValid/checkTitleHoldRules even when item holds as disabled.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: It's for any local needs because the now-extinct `itemHoldAllowed` method also had it. There was a bad title hold specific check for bib level in `itemHoldAllowed`. But now I realized that removing it made it impossible to control item hold in this respect, so I reintroduced the check but added a possibility to configure it separately. This makes the new functionality back-compatible configuration-wise.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Fixed.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Fixed.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Fixed.

config/vufind/SierraRest.ini
EreMaijala: Added.

config/vufind/SierraRest.ini
EreMaijala: Now it should be better.

EreMaijala:
@demiankatz Thanks, I appreciate all the comments! Tests would be nice, but I'll have to get back to that at some later point.

config/vufind/SierraRest.ini
```diff
@@ -148,7 +164,7 @@ title_hold_bib_levels = a:b:m:d
 ;allowCancelingAvailableRequests = true
 
 ; This section allows modification of the default mapping from item status codes to
-; VuFind item statuses
+; VuFind item statuses. Defaults are
```
demiankatz: This seems to trail off. Is something missing?

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
demiankatz: Thanks! But now I think you need to remove this SuppressWarnings annotation (looks like it's still there as of this writing).

demiankatz (changes requested):
Thanks, @EreMaijala -- nearly there, but a couple minor things I noticed in the latest revisions.

config/vufind/SierraRest.ini
EreMaijala: Oops. That's when I realized that the defaults below were not quite complete. :) Now fixed.

EreMaijala:
@demiankatz I'll get it right, eventually! Should be fixed now.

demiankatz (approved):
Thanks, @EreMaijala, it all looks good to me now!

PR #2916, opened 2023-06-01
Fix styling of badges in user lists.

demiankatz:
This PR fixes a bug identified identified by @sturkel89 during discussion of #2914 where badges stretched out unnecessarily.

Before:

image

After:

image

This applies the facet styles used to prevent badge stretching for consistency (though it does NOT apply the facet style that highlights the selected item in orange, since it seems better to keep the list highlighting consistent with the account menu highlighting above).

Note that this results in the icons and badges being vertically centered, which is inconsistent with what happens with long labels on the account menus above (which are top-aligned). I'm not sure how to easily fix this, or if changing it is even desirable. Any thoughts/opinions?


EreMaijala:
@demiankatz I'd prefer top-aligned icons and badges. Would simply adding `align-items: baseline;` to `.myresearch-menu a` work?

demiankatz:
Thanks, @EreMaijala, that has done the trick:

image

Looks to me like the icons are now a pixel or two higher than before, though. Not sure how much we should be worried about that.

demiankatz:
Also, here's how it now looks in Bootstrap3:

image

And Sandal:

image

EreMaijala:
@demiankatz I think bootstrap3 and sandal are fine, but bootprint3 is a bit off, and `align-items: center;` would work better with the icons since they're a bit too small for the line height.

ckaz (approved):
Neat! Seems you found the exact spot!

demiankatz:
> @demiankatz I think bootstrap3 and sandal are fine, but bootprint3 is a bit off, and `align-items: center;` would work better with the icons since they're a bit too small for the line height.

I'm not sure I understand what you're suggesting, @EreMaijala. Are you saying we should use `align-items: center` instead of `align-items: baseline` for bootprint3, or that we should somehow apply `align-items: center` to some more specific bit of CSS?

crhallberg:
> Looks to me like the icons are now a pixel or two higher than before, though.

I think @EreMaijala is suggesting a different alignment for the icons. I think `align-items: center` would look weird with wrapped lines in this particular case. The icons could probably be nudged with a top margin.

crhallberg (approved):
Adjusting the icons (if it's only the icons) might be best (for purity alone) in a separate PR, but it is a very minor fix.

demiankatz:
Thanks, @crhallberg -- I'll wait and see what @EreMaijala has to say before merging this. Would you be willing to create a follow-up icon-tweaking PR after that, if we all agree to go forward with this and adjust separately?

EreMaijala (approved):
@crhallberg @demiankatz Works for me!

demiankatz:
Thanks, everyone! I've merged this. @crhallberg, let me know if you need any help with the final pixel tweaking here (and please be sure to target it against the release-9.0 branch).

PR #2918, opened 2023-06-01
Revert "SierraRest: Add more configuration options for hold availability"

EreMaijala:
Reverts vufind-org/vufind#2915

I made a mistake and subsequently messed things up further than I'd like, so I think it's better to back out while I work on a better solution.


demiankatz (approved):
If at first you don't succeed, revert, revert again... ;-)

PR #2919, opened 2023-06-01
Clean up coverage tmp files after coverage reports are done.

EreMaijala:


demiankatz (approved):
Thanks, @EreMaijala -- we'll see if this makes Jenkins happier. :-)

PR #2921, opened 2023-06-02
SierraRest: Add more configuration options for hold availability (v2)

EreMaijala:
New settings have been added to control holdability of items and bibs. Some of the code has been refactored to be reusable and requests to be cacheable.

Compared to the previous attempt (#2915), this version has the following changes:

- The item holdability check really is back-compatible wrt the bib level check while now allowing configuration separately from title holds.
- Title hold rule "item" has its own item inclusion/exclusion rules so that it can be configured without getting tangled with item holdability rules. The "holdable_item" rule has been dropped.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2923, opened 2023-06-02
Alma: make item status mapping to uncertain work with getStatus call.

EreMaijala:
Previously item status mapping to uncertain based on location type only worked with getHolding.

This also fixes the getHolding test to return proper electronic holdings. The previous mocked response had wrong MARC fields.

To make this work the getItemLocationType had to be moved out from the getItemStatusFromLocationTypeMap method because getStatus only had MARC fields to manage with. This makes for the bulk of the changes.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Alma.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Alma.php
@@ -264,7 +264,13 @@ class Alma extends AbstractBase implements
     {
         // Check location type to status mappings first since they override
         // everything else:
-        [$available, $status] = $this->getItemStatusFromLocationTypeMap($item);
+        $available = null;
+        $status = null;
+        if ($this->locationTypeToItemStatus) {
+            [$available, $status] = $this->getItemStatusFromLocationTypeMap(
+                $this->getItemLocationType($item)
+            );
+        }
 
         // Normal checks for status if no mapping found above:
         if (null === $status) {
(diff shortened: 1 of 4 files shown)


module/VuFind/tests/fixtures/alma/holdings-without-mappings.json
```diff
@@ -145,18 +145,20 @@
         {
             "id": "1111",
             "source": "Solr",
-            "callnumber": "HB Ab",
+            "callnumber": "",
```
demiankatz, line 148: Is there a strong reason to remove the call number from the test fixture? Seems like adding the subfield back to minimize diffs and exercise more code might be worthwhile... but if this was done for a reason, I obviously don't object to the change. :-)

module/VuFind/src/VuFind/ILS/Driver/Alma.php
```diff
@@ -1815,9 +1810,26 @@ protected function getStatusesForInventoryTypes($ids, $types)
...
+                        [$available, $statusText] = $this->getItemStatusFromLocationTypeMap(
+                            $this->getLocationType($library, $locationCode)
+                        );
+                    }
+
+                    if (null === $available) {
+                        $availStr = strtolower($marc->getSubfield($field, 'e'));
+                        $available = 'available' === $availStr;
+                        // No status message available, so set it based on availability:
+                        $statusText = $available ? 'Item in place' : 'Item not in place';
```
demiankatz, line 1827: Is "Item in place" Alma terminology? If not, any particular reason this language was chosen?

demiankatz (commented):
Thanks, @EreMaijala -- a couple minor questions before I approve this...

module/VuFind/tests/fixtures/alma/holdings-without-mappings.json
EreMaijala: It's not relevant for electronic availability. The test returned physical item availability previously because the fixture contained wrong field (AVA instead of AVE).

module/VuFind/src/VuFind/ILS/Driver/Alma.php
EreMaijala: It is, and those are what Alma uses in the item statuses.

demiankatz (approved):
Thanks for the clarifications, @EreMaijala. Merging now! :-)

PR #2924, opened 2023-06-02
Improve SCSS-related CI behavior.

demiankatz:
This adds a check to avoid situations where a LESS change breaks the SCSS build. The build is going to fail here until #2922 is merged, so we should work on resolving that PR first.

There are also some minor changes here to bring the lessToSass check up to date.


ckaz (approved):
Looks good and reasonable to me.

demiankatz:
Hmm, the build is failing here with:

```
> grunt scss

Running "scss:sass" (scss) task
Warning: Missing binding /home/runner/work/vufind/vufind/node_modules/node-sass/vendor/linux-x64-93/binding.node
Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 16.x

Found bindings for the following environments:
  - OS X 64-bit with Node.js 16.x

This usually happens because your environment has changed since running `npm install`.
Run `npm rebuild node-sass` to download the binding for your current environment. Use --force to continue.

Aborted due to warnings.
[phingcall] /home/runner/work/vufind/vufind/build.xml:203:75: Task exited with code 3
[phingcall] /home/runner/work/vufind/vufind/build.xml:122:39: Execution of the target buildfile failed. Aborting.
```

I'm not sure why. I've tried adding a new build step to run `npm rebuild node-sass` as they suggest, but I'm not sure why this would be necessary. I wonder if it's a cache-related problem, but don't really understand why it would be.

demiankatz:
Explicitly rebuilding node-sass corrects the problem. I strongly suspect the issue here is that we're using `--ignore-scripts` on our `npm install` command (which is a security precaution) and that is likely preventing the node-sass module from fully building. I think doing an explicit rebuild here seems like an acceptable solution. I'll add a comment explaining this, then will merge.

PR #2926, opened 2023-06-06
Refresh account status when a request is placed.

EreMaijala:
This ensures account status loaded with AJAX is updated properly. This also fixes an issue where non-lightboxed request redirects to record page with an anchor that prevents any tab from becoming active when initial tab is loaded with AJAX.

A session helper is used to pass the required flag in session to the refreshed page. Lightbox has been extended to support executing scripts flagged for it.


EreMaijala:
A note about the implementation: it would be possible to return a refresh response for lightbox that would trigger a reload. That would have the benefit of refreshing the holdings information too, but would change the workflow from the current one, so I opted to do it in a way that doesn't cause visible changes apart from the information getting properly refreshed.

demiankatz (approved):
Thanks, @EreMaijala! I've added a unit test for the new view helper and my hands-on testing confirms that this fixes the problem with statuses not updating correctly in the drop-down menu.

PR #2927, opened 2023-06-06
Add config setting for custom Solr highlighting params.

demiankatz:
Solr 9.1 has changed some default highlighting behavior, and this new feature makes it easier to override/customize advanced highlighting parameters.

I've also added a test for the InjectHighlightingListener. It's not comprehensive, but it does cover the changed code (and a little more as well).

### DIFF
--- a/module/VuFind/src/VuFind/Search/Solr/InjectHighlightingListener.php
+++ b/module/VuFind/src/VuFind/Search/Solr/InjectHighlightingListener.php
@@ -67,17 +67,26 @@ class InjectHighlightingListener
     protected $fieldList;
 
     /**
+     * Extra Solr highlighting parameters.
+     *
+     * @var array
+     */
+    protected $extraHighlightingParameters;
+
+    /**
      * Constructor.
      *
      * @param BackendInterface $backend   Backend
      * @param string           $fieldList Field(s) to highlight (hl.fl param)
+     * @param array            $extras    Extra Solr highlighting parameters
      *
      * @return void
      */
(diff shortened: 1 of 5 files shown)


module/VuFind/src/VuFind/Search/Solr/InjectHighlightingListener.php
```diff
@@ -118,11 +127,14 @@ public function onSearchPre(EventInterface $event)
             if ($params = $command->getSearchParameters()) {
                 // Set highlighting parameters unless explicitly disabled:
                 $hl = $params->get('hl');
-                if (!isset($hl[0]) || $hl[0] != 'false') {
+                if (($hl[0] ?? 'true') != 'false') {
                     $this->active = true;
                     $params->set('hl', 'true');
                     $params->set('hl.simple.pre', '{{{{START_HILITE}}}}');
                     $params->set('hl.simple.post', '{{{{END_HILITE}}}}');
+                    foreach ($this->extraHighlightingParameters as $key => $val) {
```
EreMaijala, line 134: Should the extra params be set first so that even if someone sets e.g. `hl.simple.pre` in them, it doesn't break things?

EreMaijala (changes requested):

module/VuFind/src/VuFind/Search/Solr/InjectHighlightingListener.php
demiankatz: Good idea -- adjusted accordingly!

EreMaijala (approved):

PR #2814, opened 2023-03-31
Replace jumpMenu controls with menu-button components

demiankatz:
This PR supersedes #2492.

TODO
- [x] Convert card selector in librarycards/selectcard.phtml
- [x] Convert sort control in myresearch/controls/sort.phtml
- [x] Convert search scheduler in myresearch/schedulesearch.phtml
- [x] Convert limit control in search/controls/limit.phtml
- [x] Convert sort control in search/controls/sort.phtml
- [x] Convert search scheduler in search/history-table.phtml
- [ ] Make sure behavior remains consistent with (and passes tests from) #3031.
- [ ] Adjust styles for appropriate look and feel
- [x] Eliminate obsolete jumpMenu Javascript once all related code is refactored
- [ ] Update changelog to indicate removal of jumpMenu behavior
- [ ] Ensure appropriate keyboard, focus and accessibility/screen reader behavior in menu-button component.
- [ ] Run full test suite before merging

PR #2929, opened 2023-06-06
Add an option to update search results without loading the full page.

EreMaijala:
This turns on, by default, loading of results with JS when paging or changing sort order or result limit. The aim was to do this with as few template changes as possible and retain back-compatibility. It's also possible to turn off with a setting, but I believe defaulting to on due to accessibility (see below) and performance would make sense (though please let me know if this is a bad idea!).

This is related to the accessibility issue discussed in #2492, and I've spent some time trying to ensure that the controls work without issues here. @pasitiis has provided some feedback to ensure that e.g. the aria-live attribute is used properly.

Goals:
- Should still look and feel like VuFind normally does
- Page URL must reflect current state (e.g. reloading the page must lead to same results)
- Navigating history in browser must work (e.g. browser's back and forward buttons)

TODO:
- [X] Testing functionality (all search backends, history handling, error handling, page address handling etc.)
- [x] Testing with different browsers (now tested with Firefox, Chrome and Safari on MacOS)
- [x] Documentation?
- [x] openurl.js is a bit troublesome
- [x] Additional Mink tests?
- [x] Include compiled css files
- [x] Wait for PR #3046 to be merged first
- [x] Fix analytics tracking
- [x] Resolve #3292 first


module/VuFind/config/module.config.php
```diff
@@ -432,6 +432,7 @@
             'VuFind\Record\FallbackLoader\PluginManager' => 'VuFind\ServiceManager\AbstractPluginManagerFactory',
             'VuFind\Record\Loader' => 'VuFind\Record\LoaderFactory',
             'VuFind\Record\Router' => 'VuFind\Service\ServiceWithConfigIniFactory',
+            'VuFind\Record\VersionsHelper' => 'VuFind\Record\VersionsHelperFactory',
```
demiankatz: Would VersionsHelper refactoring be a good chunk of this to merge as a separate PR to simplify this one? At a glance it looks like that would shave off at least four files without interfering with other changes.

themes/bootstrap3/templates/search/results-list.phtml
```diff
@@ -0,0 +1,18 @@
...
+  $searchClass = $searchClassCss = $this->params->getSearchClassId();
+  if (!ctype_upper($searchClassCss)) {
+    $searchClassCss = preg_replace('/([a-zA-Z])(?=[A-Z])/', '$1-', $searchClassCss);
+  }
+  $this->resultsClass = 'search-results-' . strtolower($searchClassCss);
+  // Add any additional classes and attributes:
+  if ($this->results->getOptions()->loadResultsWithJsEnabled()) {
+    $this->resultsClass .= ' js-record-list';
+    if ($this->resultsAttrs) {
+      $this->resultsAttrs->add('data-backend', $searchClass);
```
demiankatz, line 16: Do we need to guard against the possibility that somebody sets resultsAttrs to a string? Without explicit documentation, that seems like an easy mistake/assumption to make.

themes/bootstrap3/templates/search/results-scripts.phtml
```diff
@@ -11,3 +11,14 @@ if (($this->listViewOption ?? 'full') !== 'full') {
...
     $this->headScript()->appendFile('embedded_record.js');
 }
+if ($this->jsResults ?? false) {
+    $this->headScript()->appendFile('search.js');
+    // Ensure OpenURL support is available even if none of the results on the initial
+    // page has one:
+    $this->headScript()->appendFile('openurl.js');
+}
+// Load only if list view parameter is NOT full:
+if (($this->listViewOption ?? 'full') !== 'full') {
```
demiankatz: This appears to be duplicate logic; maybe a merge error somewhere along the way?

themes/bootstrap3/templates/search/results.phtml
```diff
@@ -104,10 +103,16 @@
...
+      $topPaginatorStyle = $this->results->getOptions()->getTopPaginatorStyle();
+      if ($topPaginatorStyle && 'full' === $topPaginatorStyle):
+    ?>
+      <?=$this->paginationControl($this->results->getPaginator(), 'Sliding', 'search/pagination.phtml', ['results' => $this->results, 'options' => $this->paginationOptions ?? []])?>
+    <?php endif; ?>
     <form id="search-cart-form" method="post" name="bulkActionForm" action="<?=$this->url('cart-searchresultsbulk')?>" data-lightbox data-lightbox-onsubmit="bulkFormHandler">
       <?=$this->context($this)->renderInContext('search/bulk-action-buttons.phtml', ['idPrefix' => ''])?>
     </form>
-    <?=$this->render('search/list-' . $this->params->getView() . '.phtml')?>
+    <?=$this->render('search/results-list.phtml')?>
```
demiankatz: Refactoring to introduce this template might be another chunk best merged separately. I think having a single commit to highlight how things are moved around may help make conflict resolution easier if people have customized this code.

themes/bootstrap3/templates/search/pagination_simple.phtml
```diff
@@ -0,0 +1,27 @@
...
+          'params' => $this->results->getParams(),
+          'recordTotal' => $this->results->getResultTotal(),
+          'lookfor' => '',
+        ];
+        echo $this->render('search/controls/showing.phtml', $params);
+      ?>
+    </li>
+    <?php if (isset($this->next)): ?>
+      <li role="none">
+        <a class="page-next" href="<?=$this->results->getUrlQuery()->setPage($this->next)?>"><span class="sr-only"><?=$this->transEsc('page_next')?></span><?=$this->icon('page-next') ?></a>
```
demiankatz: page-prev is split up and indented while page-next is all one line. Would probably make sense to format these consistently with each other.

module/VuFind/src/VuFind/Search/Base/Params.php
```diff
@@ -1751,11 +1751,13 @@ public function getLimitList()
     {
         // Loop through all the current limits
         $valid = $this->getOptions()->getLimitOptions();
+        $defaultLimit = $this->getOptions()->getDefaultLimit();
         $list = [];
         foreach ($valid as $limit) {
             $list[$limit] = [
                 'desc' => $limit,
                 'selected' => ($limit == $this->getLimit()),
+                'default' => $limit == $defaultLimit,
```
demiankatz, line 1778: Would the changes to clear options to use default values make sense as a separate small PR as well? This seems like a reasonable stand-alone improvement to reduce URL length, though it does have the disadvantage of making saved URLs less "sticky" if defaults are changed in future.

themes/bootstrap3/templates/search/controls/limit.phtml
```diff
@@ -1,11 +1,11 @@
 <?php $limitList = $this->params->getLimitList(); ?>
 <?php if (count($limitList) > 1): ?>
-  <form class="form-inline search-result-limit" action="<?=$this->currentPath() . $this->results->getUrlQuery()->setLimit(null)?>" method="get">
-    <?=$this->results->getUrlQuery()->asHiddenFields(['sort' => '/.*/']);?>
+  <form class="form-inline search-result-limit" action="<?=$this->currentPath()?>" method="get">
+    <?=$this->results->getUrlQuery()->asHiddenFields(['limit' => '/.*/']);?>
```
demiankatz: Is this a bug fix that should be backported, or am I misunderstanding what I'm seeing?

themes/bootstrap3/js/common.js
```diff
@@ -655,7 +655,10 @@ function unwrapJQuery(node) {
 function setupJumpMenus(_container) {
   var container = _container || $('body');
   container.find('select.jumpMenu').on("change", function jumpMenu() {
-    $(this).parent('form').trigger("submit");
+    // Check if jumpMenu is still enabled (search.js may have disabled it):
```
demiankatz, line 719: Is there new code using jumpMenu? If so, maybe we should investigate alternatives, since #2814 aims to do away with jumpMenu entirely. (I realize that PR still needs a lot more work, but in theory it's scheduled to be completed before 10.0 is released).

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -306,6 +306,20 @@ abstract class Options implements TranslatorAwareInterface
...
+     * @var bool
+     */
+    protected $loadResultsWithJs = true;
+
+    /**
+     * Top pagination control style (none, simple or full)
+     *
+     * @var string
+     */
+    protected $topPaginatorStyle = 'simple';
```
demiankatz: This gets initialized in the constructor, so it's probably less confusing to avoid defaulting it here.

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -306,6 +306,20 @@ abstract class Options implements TranslatorAwareInterface
      */
     protected $firstlastNavigation = false;
 
+    /**
+     * Is loading of results with JavaScript enabled?
+     *
+     * @var bool
+     */
+    protected $loadResultsWithJs = true;
```
demiankatz: Initialized in constructor, doesn't need initialization here.

themes/bootstrap3/templates/search/pagination.phtml
```diff
@@ -1,14 +1,14 @@
 <?php if ($this->pageCount): ?>
   <nav aria-label="<?=$this->transEscAttr('pagination_label')?>">
-    <ul class="pagination">
+    <ul class="pagination<?=$this->results->getOptions()->loadResultsWithJsEnabled() ? ' js-pagination' : ''?>">
       <?php if (isset($this->previous)): ?>
         <?php if (!isset($this->options['disableFirst']) || !$this->options['disableFirst']): ?>
           <li role="none">
-            <a href="<?=$this->currentPath() . $this->results->getUrlQuery()->setPage(1)?>" aria-label="<?=$this->transEscAttr('page_first')?>">[1]</a>
+            <a href="<?=$this->results->getUrlQuery()->setPage(1)?>" aria-label="<?=$this->transEscAttr('page_first')?>">[1]</a>
```
demiankatz, line 16: Is it possible there was a reason for using currentPath() here? Perhaps it was always unnecessary, but I have a gut feeling that maybe there's some edge case where it matters.

demiankatz (changes requested):
Thanks, @EreMaijala -- I've finally found time to give this a fairly close look. I haven't done a deep dive into the most important parts of the code -- search.js and GetSearchResults.php -- as I should probably look at those two with a fresh brain. For now, I just went through everything else, looking for bugs, potential problems, and opportunities for spinning off parts of the work into separate PRs so that we can make this one a little less intimidating. I hope this is helpful!

themes/bootstrap3/js/common.js
EreMaijala: This is actually a workaround for jumpMenus, so if we can get rid of them, this can go. The purpose of this change is to be able to disable jumpMenu in case we reload the results asynchronously and don't want to jump. Since the js results init runs after this bind has been done, removing the jumpMenu class afterwards allows the action to be disabled. 

We may of course need another way to substitute the code that reloads the results if jumpMenu is replaced with something else that triggers a reload.

themes/bootstrap3/templates/search/pagination.phtml
EreMaijala: I can't think of any. We used to have the full path in quite a few places, but in Finna we got rid of them years ago without issues. Doesn't mean there can't be reason, I just can't imagine why it would be needed. The only case would be if we used the `base` tag, but fortunately we don't.

themes/bootstrap3/templates/search/results-list.phtml
EreMaijala: We could check the type, but I'm not sure what would be a useful course of action if it's a string.

themes/bootstrap3/templates/search/results-scripts.phtml
EreMaijala: Fixed.

themes/bootstrap3/templates/search/pagination_simple.phtml
EreMaijala: Fixed.

themes/bootstrap3/templates/search/controls/limit.phtml
EreMaijala: Bug fix indeed, now in #3031.

module/VuFind/src/VuFind/Search/Base/Options.php
EreMaijala: Fixed.

module/VuFind/src/VuFind/Search/Base/Options.php
EreMaijala: Fixed.

themes/bootstrap3/templates/search/results.phtml
EreMaijala: Now in #3032.

module/VuFind/config/module.config.php
EreMaijala: Yes, now in #3033.

EreMaijala:
I've now extracted all suggested pieces and the top paginator support to separate pull requests.

themes/bootstrap3/templates/search/pagination.phtml
demiankatz: Thanks, @EreMaijala, it's probably okay then. And interesting about the `base` tag -- I somehow had never encountered that before!

EreMaijala: @demiankatz That's for the better. I think the best use for the tag is to sow confusion. :)

themes/bootstrap3/templates/search/results-list.phtml
demiankatz: At least three options:

1.) Is there a way to initialize the htmlAttributes helper from a string of attributes? If so, we could promote strings to objects.

2.) We could concatenate the attribute to the string if it's a string.

3.) We could throw an exception, which is probably better than a low-level PHP error.

Since we treat resultsAttrs like a string in the list-*.phtml templates, it's not immediately obvious that we're using the htmlAttributes helper, depending on how you trace the code for the property, so I think it's probably worth doing _something_, though not necessarily anything that requires tons of time/effort.

EreMaijala: Alas, it can't be built from a string. I added a type check now since that keeps the actual code clean.

demiankatz: Maybe a mechanism for populating htmlAttributes from a string would be a useful future enhancement to that helper, but for now this does the job!

EreMaijala: Yes, but since the idea of the helper is to avoid hand-crafting attribute strings, it kind of makes sense to not support population from a string. :)

demiankatz: Fair point. :-)

module/VuFind/src/VuFind/Search/Base/Params.php
EreMaijala: This only affects the display in JS mode. The data-default attribute is used to set the correct item active when moving around in the history and the current url doesn't contain the default.

EreMaijala: Additionally, we're not really consistent here: default limit is included in the query string when making a new search, but default sort is not.

demiankatz: Possibly something we should straighten out at some point in the future, but it sounds like no action is needed at this point in time. :-)

sturkel89:
My tests show two weird behaviors.

**SELECT TWO OR MORE LIMITS -> RESULTS SORTING IS DISABLED**
In the test branch, when I set **two limits**, I find that none of the sorting options (relevance, author, title) work. They can be either two limits from the same grouping, or one each from two different groupings. 

With only one limit set, sorting works fine. 

(In dev, you can have any number of limits set, and sorting still works fine.)

@demiankatz suggests that we may need to add a new automated test for this scenario.

**STRANGE BEHAVIOR REGARDING PAGE LENGTH SELECTIONS**
I uncommented limit_options in searches.ini and tested the behavior of selecting different numbers of search results showing on a page. I found that under some circumstances, on the test branch, “Results” respects your choice, and under other circumstances it reset to the previously selected option. More details:

- When I start out with the default page length, I can do limits, searches, filters, and the page length stays the same. 

- When I change the page length to a new number, the page length changes to match my selection. BUT when I select any filter, the page length (and display of number) reverts to my original selection. Each time I try to override and select a new page length, the page length reverts to the original default when I apply any filter.

- Starting all over again (switch branch to dev and then back to test)... when I select a new page length, then click an item title to view an item record, then back out (back arrow or click Search in the breadcrumbs), my new selection sticks! Behavior is normal. I can apply filters and clear them, and the selection I made before viewing the item record is respected. 

     It seems that the act of viewing a page record sets the system to behave normally (e.g., like dev).

Please let me know if you need more specific illustrations of these behaviors.

demiankatz (changes requested):
Just putting this into "request changes" state in response to @sturkel89's bug reports. I'll do a closer review of the actual code once these are sorted out.

EreMaijala:
@sturkel89 thanks for testing. The page length selection issue should be fixed now, but I'm unable to replicate the problem. Are those two just normally selected facet values? And what exactly does the 'none of the sorting options work' mean? Can you still select them but don't get the correct results or is it something else?

demiankatz:
> @sturkel89 thanks for testing. The page length selection issue should be fixed now, but I'm unable to replicate the problem. Are those two just normally selected facet values? And what exactly does the 'none of the sorting options work' mean? Can you still select them but don't get the correct results or is it something else?

I'm confident that the problem Susan describes has to do with applying more than one facet filter, but I'm not sure exactly what broken behavior she was encountering. She should be online in a few more hours to clarify. :-)

sturkel89:
You can replicate my experience in the test branch by following these steps:

- Run an empty search to retrieve the entire contents of the database.
- Apply any two filters, e.g. under **Library**, filter to **geo.mrc**, and under **Format**, filter to **Book Chapter**.
- There are 177 results, sorted by default by Relevance. 
- Try each of the sort options in any sequence (Author, Title, Date Descending, Date Ascending, etc.).
- You will see that nothing changes in the results display. The order stays the same and the results do not reload at all.

Then, checkout the dev branch and go through the same process.
- There is a clear response to each of those sort option selections: the screen flickers and the entries reload, even if the sort doesn't change (e.g., because all the items have the same call number, so there's no resorting needed). 
- There is definitely a change in order between Title and Author -- with Title, we go from Test Publication 20001 to Test Publication 20020 on the first screen, and with Author the items are arranged by author and the pub numbers are in mixed-up order starting with 20127.

sturkel89:
Also, I just tried to test the limit_options / page length issue, and now I get an error when I try to change from the default length of 20 to any other length:

image

sturkel89 (changes requested):
@EreMaijala, please see the two comments I just added. (I keep forgetting to enter them here, under "Review changes"!) Thanks.

EreMaijala:
@sturkel89 Sorry, there was an unrelated JavaScript error in the code base. It's fixed now, so you shouldn't encounter the error again. Also, thanks for the steps to reproduce the issue. I tried them with the latest version but couldn't see any issues.

sturkel89:
I've just taken another look at the test branch and compared it with the dev branch.

**Good news:** 
The situation I described [above ](https://github.com/vufind-org/vufind/pull/2929#issuecomment-1816684231)re: limit options / page length has been fixed! I can now successfully view results at any of the Results per page options - 10, 20, 40, etc. up through 100 records per results page. 

As long as I have up to one limit set, I can sort these results by any of the sorting options at any of these page lengths. 

**Bad news:** 
The situation I described [above ](https://github.com/vufind-org/vufind/pull/2929#issuecomment-1816677437)re: Sort by not working at all once two limits are set is still happening. I compared test and dev at different page lengths, and as above once there are two limits set, there is no visible response when I change sort orders in the test branch.

EreMaijala:
@sturkel89 Got it, things work great with Firefox but not with Chrome. Will look into it now.

EreMaijala:
@sturkel89 Now things should finally with multiple filters as well.

sturkel89 (changes requested):
This branch also has [the issue I noted in PR 3046](https://github.com/vufind-org/vufind/pull/3046#pullrequestreview-1756191313) with the hierarchical facets on the VuFind front page not working. 

If you click any of those links, you are not brought to the results page. The other facet options on the front page work correctly.

sturkel89:
> @sturkel89 Now things should finally with multiple filters as well.

Yes - I have tested sorting with various numbers of filters and various page lengths, and as long as I eliminate the Overdrive records from the results set the sorting and number of records on the page all function properly - no errors, and the sorting actually happens.

sturkel89 (changes requested):
A new (?) difference between test and dev I've just noticed is that when I have a long results list and change either the sort or the results per page, **the screen automatically scrolls down** so that the results themselves are at the very top of the screen. The search box and option to X out some facets is above the fold and I'd need to scroll back up. 

In this screenshot of the test branch, I had just clicked to show 80 results per page instead of whatever I had before.

image


On dev, I can change the sort or the results per page, and my view of the screen stays the same - no automatic scroll. I prefer this! In this screenshot of the dev branch, I had just changed the results per page to 60.

image

I am pretty sure that this is a new change you've introduced?? It may be an aesthetic preference, but I'll still list this as a "request change" because I like it better the other way.  ;-)

EreMaijala:
> This branch also has [the issue I noted in PR 3046](https://github.com/vufind-org/vufind/pull/3046#pullrequestreview-1756191313) with the hierarchical facets on the VuFind front page not working.
> 
> If you click any of those links, you are not brought to the results page. The other facet options on the front page work correctly.

I'll wait for PR 3046 to be merged before checking this again. It should help alleviate the issue here.

EreMaijala:
> A new (?) difference between test and dev I've just noticed is that when I have a long results list and change either the sort or the results per page, **the screen automatically scrolls down** so that the results themselves are at the very top of the screen. The search box and option to X out some facets is above the fold and I'd need to scroll back up.

I updated the logic so that if the page is already on top, it won't scroll down. It will still scroll up when updating results if the window is scrolled down enough for the result tools (sorting, page size etc.) to not be visible. In practice this happens when using the bottom pagination control.

demiankatz:
@EreMaijala, I've just merged #3046, so this is ready for conflict resolution and re-testing when you have the time!

EreMaijala:
@demiankatz @sturkel89 Latest dev now merged here, so this should be good for another test round!

sturkel89 (approved):
> > A new (?) difference between test and dev I've just noticed is that when I have a long results list and change either the sort or the results per page, **the screen automatically scrolls down** so that the results themselves are at the very top of the screen. The search box and option to X out some facets is above the fold and I'd need to scroll back up.
> 
> I updated the logic so that if the page is already on top, it won't scroll down. It will still scroll up when updating results if the window is scrolled down enough for the result tools (sorting, page size etc.) to not be visible. In practice this happens when using the bottom pagination control.

I tested this branch in all three themes, and compared it with dev.

I see that all of the changes @EreMaijala describes are working as intended. Looks good to me! 

In dev, the only difference is that use of the bottom pagination control pops you up to the very top of the page, but I think the behavior in the test branch is better under that circumstance.

demiankatz:
Thanks, @sturkel89, that's great news!

@EreMaijala, are there any functional areas/edge cases you'd like to see get more thorough hands-on testing, or are we ready for the code review (and, if all is well, merge)?

sturkel89:
NB: I don't think I turned on hierarchical facets when I did the testing I just reported - would that have made any difference?

EreMaijala:
@sturkel89 @demiankatz Now that hierarchical facets are very similar to normal facets, they shouldn't make a difference, but minor differences can cause major issues, so anything's possible. I've tested with hierarchical facets, but I'll extend the Mink tests to verify that changed settings (like sort) are kept properly when facets are selected or removed. Other than that, I believe this is ready for final reviews.

EreMaijala:
Mink tests updated.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
```diff
@@ -0,0 +1,424 @@
+<?php
+
+/**
+ * "Get Search Results" AJAX handler
+ *
+ * PHP version 7
```
demiankatz: We should change all `PHP version 7` comments to `PHP version 8` at this point.

config/vufind/searches.ini
```diff
@@ -157,10 +157,18 @@ always_display_reset_filters = false
 ; commented out, fallback ID lookups will be disabled.
 ;fallback_id_field = previous_id_str_mv
 
+; This controls whether results are loaded with JavaScript when paging or changing
+; settings. Loading results this way improves performance and accessibility, and is
+; enabled by default. Loading will automatically fall back to non-JS mode if
+; JavaScript is not supported by the browser.
+load_results_with_js = true
```
demiankatz, line 164: Since this setting is implemented in the base class, should we copy this setting to other backend-specific .ini files (Summon, EDS, etc., etc.)?

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/BasicSearchTest.php
```diff
@@ -53,61 +55,105 @@ public function testOutOfBoundsPage()
...
+        );
+
         $session = $this->getMinkSession();
         $session->visit($this->getVuFindUrl() . '/Search/Results');
         $page = $session->getPage();
 
-        // No paginator unless configured:
-        $this->unFindCss($page, '.search-header .pagination');
-        $this->unFindCss($page, '.search-header .pagination-simple');
+        // Should never have full pagination:
```
demiankatz: Should this be "never have full top pagination"?

demiankatz (changes requested):
@EreMaijala, looks like some conflicts have been introduced recently. I also had a couple of minor observations. I haven't reviewed the two very long files (GetSearchResults.php and search.js) in depth yet -- I'll try to find time for that next week once these smaller matters have been resolved.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
EreMaijala: Oops, fixed.

config/vufind/searches.ini
EreMaijala: Done.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/BasicSearchTest.php
EreMaijala: Yes, fixed.

EreMaijala:
@demiankatz All resolved once again. :)

themes/bootstrap3/js/search.js
```diff
@@ -0,0 +1,371 @@
...
+   * Initialize result controls that are not refreshed via AJAX.
+   *
+   * Note that view type links are updated in updateResultControls, but using them
+   * will cause a reload since page contents may change.
+   */
+  function initResultControls() {
+    document.querySelectorAll(resultsControlFormSelector).forEach((form) => {
+      if (!form.dataset.ajaxPagination) {
+        form.dataset.ajaxPagination = true;
+        form.querySelectorAll('.jumpMenu').forEach(jumpMenu => {
```
demiankatz, line 55: I wonder how this should interact with the changes in #2814, where we fully eliminate the jump menu controls. Does this potentially offer a different solution? Does this replace the need to replace jump menu controls with menu-button components in some situations? Would it be simpler to just add submit buttons for accessibility in the non-JS situation, hide those submit buttons in AJAX mode, and get rid of the whole jumpMenu concept entirely?

Maybe that's a change beyond the scope of this current PR, but I wanted to capture the idea so we can figure out our next steps -- i.e. do we need to dramatically revise the work in progress on #2814? It has already grown quite stale, so I won't be too sad if we need to throw it away and approach it from a different angle.

demiankatz (commented):
@EreMaijala, still chipping away at this a bit at a time -- I took a deeper look at the JS code today, and it raised one possible point of discussion (which I'm happy to defer to a separate PR if that's easiest, but wanted to bring up in case it makes sense to handle sooner).

Also, a general question: what are the implications of this change for usage tracking -- e.g. Matomo? Is that something we need to worry about in some way?

themes/bootstrap3/js/search.js
EreMaijala: This can adapt to whatever #2814 brings, but with this the need to get rid of jump menus due to accessibility issues vanishes. Since this can keep the focus in the control that was updated, there's no sudden and unexpected jump that causes accessibility issues. #2492 and #2814 were actually the ones that got me started working on this one (along the other benefits for sure, but still).

EreMaijala:
@demiankatz Analytics tracking is a valid point. I'll need to add triggering of tracking.

themes/bootstrap3/js/search.js
demiankatz: I guess my question is: if this new logic supersedes the setupJumpMenus function, can we just get rid of that function entirely? We then presumably need to display submit buttons in cases where the AJAX search results are disabled. This should already work for the case where JS is fully disabled, but I'm wondering if additional work might be needed for the case where AJAX search results are disabled via configuration.

If we get to a place where we display submit buttons when needed and have accessible behavior in AJAX mode, then we might not need #2814 at all (or we might be able to start a new PR that takes some smaller parts without requiring so much additional work to be done).

And then that brings me back to the question: is it worth doing any of that here, or should we wait and do it as follow-up?

EreMaijala: Right, I missed that part of the idea. Yes, that'd make sense. I'd prefer to do it in a follow-up as this is already big enough, but if you think it's better to incorporate it here to keep all the major changes in one set, I can mange that as well.

EreMaijala:
@demiankatz Analytics support added. While at it I added proper support for loading scripts returned by the AJAX request. It unearthed again the problem with openurl.js that gets added for every single result that has OpenURL (in dev as well). I think we should do something about it, but I'm not sure what. It's small enough that it would be tempting to include it in theme config... But this can be left out from this PR, it's an old issue.

themes/bootstrap3/js/search.js
demiankatz: I don't mind waiting!

themes/bootstrap3/templates/search/results-scripts.phtml
```diff
@@ -11,3 +11,9 @@ if (($this->listViewOption ?? 'full') !== 'full') {
     $this->headScript()->appendFile('record.js');
     $this->headScript()->appendFile('embedded_record.js');
 }
+if ($this->jsResults ?? false) {
+    $this->headScript()->appendFile('search.js');
+    // Ensure OpenURL support is available even if none of the results on the initial
+    // page has one:
+    $this->headScript()->appendFile('openurl.js');
```
demiankatz: Would a possible compromise on the openurl.js situation be to move this outside of the conditional block and then remove it from being included in other places? Or do we have to worry about OpenURL embedding outside of search results? If that is the case, then I'd support moving it into the theme config for global activation both to simplify things and to increase the benefit of using the asset pipeline.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
```diff
@@ -0,0 +1,455 @@
...
+     * @param Params $params Request params
+     *
+     * @return ?Results
+     */
+    protected function getSearchResults(Params $params): ?Results
+    {
+        parse_str($params->fromQuery('querystring', ''), $searchParams);
+        $backend = $params->fromQuery('source', DEFAULT_SEARCH_BACKEND);
+        $searchType = $params->fromQuery('searchType', '');
+        if ('versions' === $searchType) {
```
demiankatz: I wonder if it would be beneficial to wrap this in a method like `applyTypeSpecialCases($searchType, & $searchParams): bool` to better encapsulate it (and perhaps encourage more pluggable refactoring in the future, if more special cases arise). The method could return true if search should proceed and false if the caller needs to return null.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
```diff
@@ -0,0 +1,455 @@
...
+                ->getSearchStringFromWorkKeys((array)$driverAndKeys['keys']);
+            $searchParams['type'] = $this->versionsHelper->getWorkKeysSearchType();
+        }
+
+        $results = $this->resultsManager->get($backend);
+        $paramsObj = $results->getParams();
+        $paramsObj->getOptions()->spellcheckEnabled(false);
+        $paramsObj->initFromRequest(new Parameters($searchParams));
+
+        if ('versions' !== $searchType) {
```
demiankatz: Maybe use a class property, like `if (!in_array($searchType, $this->searchTypesWithoutHistory)) {` to make this more generic as well?

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
```diff
@@ -0,0 +1,455 @@
...
+     * Render page elements
+     *
+     * @param Params  $params  Request params
+     * @param Results $results Search results
+     *
+     * @return array
+     */
+    protected function getElements(Params $params, Results $results): array
+    {
+        if ($params->fromQuery('searchType', '') === 'versions') {
```
demiankatz: `$this->customizeUrlHelperForSearchType($results, $params)` for abstraction?

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
```diff
@@ -0,0 +1,455 @@
...
+     *
+     * @param Params  $params  Request params
+     * @param Results $results Search results
+     *
+     * @return ?string
+     */
+    protected function renderResults(Params $params, Results $results): ?string
+    {
+        [$baseAction] = explode('-', $results->getOptions()->getSearchAction());
+        $templatePath = "$baseAction/results-list.phtml";
```
demiankatz, line 275: Is it a problem that this is hard-coded to list view? What happens in grid view? Obviously I don't object to starting with the most commonly-used case, but I just want to be sure other views aren't broken, and I'm curious whether there are future plans (or should at least be a JIRA ticket) to support AJAX grids.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
```diff
@@ -0,0 +1,455 @@
...
+        $options = $results->getOptions();
+        $cart = $this->renderer->plugin('cart');
+        $showBulkOptions = $options->supportsCart()
+            && ($this->config['Site']['showBulkOptions'] ?? false);
+        // Checkboxes if appropriate:
+        $showCartControls = $options->supportsCart()
+            && $cart()->isActive()
+            && ($showBulkOptions || !$cart()->isActiveInSearch());
+        // Enable bulk options if appropriate:
+        $showCheckboxes = $showCartControls || $showBulkOptions;
```
demiankatz, line 289: This feels like complex and potentially confusing logic that could cause problems if it gets out of sync. Is there any chance we could encapsulate it in a single place for sharing, or is that just not practical?

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
```diff
@@ -0,0 +1,455 @@
...
+     * @return ?string
+     */
+    protected function renderPagination(
+        Params $params,
+        Results $results,
+        string $template = 'search/pagination.phtml',
+        string $ulClass = ''
+    ): ?string {
+        $backend = $params->fromQuery('source', DEFAULT_SEARCH_BACKEND);
+        $paginationOptions = 'EDS' === $backend
```
demiankatz: Can we perhaps move this logic into the Options class so we don't need an EDS-specific special case here?

demiankatz (changes requested):
Thanks, @EreMaijala, I think we're nearly done -- I finally got through reviewing the last (and longest) PHP file, and just had a few suggestions, mainly around reducing/isolating special case logic and ensuring maintainability.

themes/bootstrap3/templates/search/results-scripts.phtml
EreMaijala: We need to support OpenURL embedding outside of search results, like in favorite lists as well. I'll move it to theme config.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
EreMaijala: results-list.phtml is the main template for the results regardless of the list type. Grid is already supported. :) Visual is not AJAX-enabled, but it doesn't need paging anyway.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
EreMaijala: Yes, and we can eliminate carrying these around everywhere. Now in #3292.

demiankatz: Guess we should work on getting that finished up before continuing here!

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
EreMaijala: I eliminated the need for this by passing the history flag to the handler. If someone would mangle the URL and save such search to history, it would still work but display as a normal search, so nothing tragic would happen.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
EreMaijala: Eliminated as above.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
EreMaijala: No longer needed.

module/VuFind/src/VuFind/AjaxHandler/GetSearchResults.php
EreMaijala: Options already has the `supportsFirstLastNavigation()` method, so I just switched to that. :)

demiankatz:
@sturkel89, now that @EreMaijala has significantly simplified the code here, do you mind doing one more round of testing to be sure nothing has gotten broken? In particular, I'd check that sort/limit/facet/pagination options all work together appropriately. It's also worth trying grid view (which can be turned on in searches.ini) just to be sure nothing weird happens there.

sturkel89 (approved):
I reviewed the test branch by trying various limits, sorts, page lengths, paging forward and back, etc. 

In normal (list) view, everything looked fine and performed perfectly. Each time I sorted in a different way, the page rewrote appropriately. When I changed page length, I was always brought back to item/page one. My choice of page length seemed to stick regardless of what I did. 

So, no complaints.

The only thing I noticed was that when I had a very long page (e.g. 80 or 100 items), there was a little delay as I tried to pull the little box at the edge of the browser window down to scroll quickly to the end of the page. The system had to rewrite information for each item, and that caused a little lag. This is probably unavoidable in the case of these long pages. (Right?)

-----

At Demian's suggestion, I also tested grid view. I had never seen it before, and don't think it's a very good choice for this type of data. I think the record components that display by default should be more consistent regardless of item type. However, this critique is not specific to the test branch, as I see the same thing on dev.

Note that the list view displays location information for every item, as well as lozenges for format for every item, and for availability when it's a print item.

image

In grid view, location is not displayed for any item, nor are format lozenges. So, the availability lozenges stand out in their presence for print items and in their absence for online items. 

image

The simplest solution would probably be to just show just the thumbnail and title, and probably the format, for each item. Locations and hyperlinks take up a lot of space and make the grid look messier. If the user wants to know more about a specific title, they can click and view the item record. 

(I know that this is not an issue that will be solved in the context of this PR. Just thought I would record my thoughts here. Feel free to disregard!)

EreMaijala:
@sturkel89 Thanks for the feedback! I believe the scrolling with Chrome is the same also with the dev branch. The loading of item statuses affects the scroll position, and Chrome can also lose the connection between the mouse movement and scroll handle.

I agree with your assessment on the grid view. I think it's best suited for visual material like images. But it could be specific to a case, and could be easily customized according to an institution's needs.

demiankatz (approved):
This is all looking good to me now. Thanks, @EreMaijala, for making it happen, and @sturkel89 for all the helpful testing!

Regarding the grid view, there's definitely a good reason that's not enabled by default. As @EreMaijala says, it's easy to customize, so my view is that it's a demonstration of a possible way to render results differently, but it's likely to require local customization to meet whatever use case justifies the grid layout... so I don't think it's worth investing a lot of time into trying to perfect it (unless we suddenly find a lot of people using it and all making similar adjustments).

demiankatz:
@EreMaijala, I have merged this and updated the changelog in the wiki.

Do you think other documentation is necessary? If so, let me know if you need help setting it up.

Also, the other open "next step" discussed in this PR is to clean up/eliminate the jumpMenu code so that we can finish (or supersede) #2814. Are you in a position to work on this, or would you like me to put it on my list?

EreMaijala:
@demiankatz Thanks! I don't think we actually need any further documentation since all the refactoring in other issues simplified this significantly. I'll see what I can come up with regarding the jumpMenu code.

PR #2931, opened 2023-06-07
Fix account notification check to default to enabled.

EreMaijala:
This matches the check in module/VuFind/src/VuFind/Auth/Manager.php. This change doesn't require constructor or factory changes, but I intend to post a follow-up for dev branch to make IlsRecords check the setting from Auth Manager.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2932, opened 2023-06-07
Make sure array-style search query params are arrays.

EreMaijala:
I noticed this issue when Outlook butchered a search URL leaving just `?hiddenFilters` as the query params. I tried to check any other search parameters as well.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2934, opened 2023-06-07
Switch to Splide Carousel Library (see #4557)

crhallberg:
Continuation of #2702.

- [x] npm install
- [x] Channels implementation
- [x] use for SimilarItems carousel
- [x] ~~remove jQuery from carousels setup while we're refactoring~~ done in `common.js`, the rest can be saved for de-jQuery project.
- [ ] Check whether changes fix [VUFIND-1691](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1691)


demiankatz:
Thanks, @crhallberg! I'll test this soon when time permits. In the meantime, is it possible to install splide via npm as we've done with some of our other dependencies, or is that not feasible?

demiankatz:
@crhallberg, I never looked closely enough at #2702 to realize that this was about the Channels view and not the [SimilarItems carousel](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/RecordTab/similaritemscarousel.phtml). Is that using native Bootstrap or something else? Should we consider converting that to Splide as well for consistency?

crhallberg:
Yes! I intend to replace all of the carousels, I just wanted to get the Channels one down first.

package.json
```diff
@@ -14,38 +14,35 @@
...
+    "@splidejs/splide": "^4.1.4",
     "autocomplete.js": "vufind-org/autocomplete.js#v2.1.5",
     "grunt": "^1.0.4",
     "grunt-contrib-less": "^2.0.0",
     "grunt-less-to-sass": "duvillierA/grunt-less-to-sass#0.0.11",
     "grunt-sass": "^3.1.0",
     "jit-grunt": "^0.10.0",
-    "node-sass": "^7.0.1"
+    "node-sass": "^7.0.1",
+    "vanilla-cookieconsent": "3.0.0-rc.14"
```
demiankatz: Why did this move up from devDependencies to dependencies? I think the idea is that dependencies that get installed by Phing are treated as devDependencies, because we only need them when developing (since the files are committed to Git for the benefit of "regular users"). For this reason, we should move splide down there as well.

demiankatz (changes requested):
Thanks for getting the ball rolling, @crhallberg. This doesn't seem to be working yet for me, though. When I go to the channels page, I get this error:

```
Uncaught Error: [splide] A track/list element is missing.
    bn assert.ts:12
    p Elements.ts:181
    t Splide.ts:126
    w forOwn.ts:20
    w forOwn.ts:19
    mount Splide.ts:123
    setupChannelSlider channels.js:96
    jQuery 2
    init channels.js:191
    init common.js:113
    commonDocReady common.js:654
    jQuery 13
assert.ts:12:10
```

And I end up with a partially-rendered screen like this:

image

Any idea what's going on?

Also, see below for an unrelated question.

package.json
crhallberg: dependencies are for libraries used on the front-end, devDependencies are for build tools. It's me being a bit strict and it's not the best time to do this, and there's clearly a lot more here to cleanup.

demiankatz: I think it's a bit murky because of the way VuFind uses Javascript. Technically, it could be argued that EVERYTHING is a build tool. You don't need to run `npm install` to run the VuFind application, because everything is already there in Git. You ONLY need Javascript tools if you're doing development of some sort. And by the standard you propose, some things currently in `dependencies` are dev tools (grunt, node-sass, etc.) and some of the things in devDependencies are not (jstree). By any definition I can think of, some of what we're doing looks wrong. I think we should talk this through and define a standard that we can all agree on -- probably with input from @EreMaijala when he's back from holiday -- but for now I'd suggest just adding splide wherever you think it belongs, but avoiding moving other stuff around until we can address the moves more holistically.

EreMaijala: I'd draw the line between non-development needs and development needs this way (and I totally understand this being arbitrary and non-scientific):
- Non-development tools are needed to customize a VuFind installation including branding, colors etc. but not for actual development work.
- Development dependencies are needed to update js libraries etc. that could also require code changes to accommodate upstream changes.

As far as I can see, with this distinction our current package.json in dev makes sense for most parts, but autocomplete.js is in wrong block since it's only needed for updating of the library.

demiankatz: Thanks, @EreMaijala, I think that rule of thumb seems workable. If @crhallberg agrees, maybe he can move the autocomplete down instead of the cookie-consent up, and then we'll be back to consistency. :-)

config/constants.config.php
```diff
@@ -12,7 +12,7 @@
 defined('APPLICATION_ENV')
     || define(
         'APPLICATION_ENV',
-        (getenv('VUFIND_ENV') ? getenv('VUFIND_ENV') : 'production')
+        (getenv('VUFIND_ENV') ? getenv('VUFIND_ENV') : 'development')
```
demiankatz: Accidental commit that needs to be reverted. I recommend setting dev mode by editing /etc/apache2/conf-enabled/vufindtest.conf to uncomment the development setting, then restarting Apache. It's two steps instead of one, but it avoids this accidental commit issue, which seems to happen a lot.

package.json
```diff
@@ -33,6 +33,7 @@
     "watch:lessToSass": "grunt watch:lessToSass"
   },
   "dependencies": {
+    "@splidejs/splide": "^4.1.4",
```
demiankatz: Based on conversation elsewhere with @EreMaijala, I think this should get moved down to the devDependencies section, since it is required for building VuFind but not for running VuFind.

demiankatz (changes requested):
I was trying to find our conversation with @EreMaijala where I thought we made a decision about dependencies vs. devDependencies in package.json, so that I could reference it in a different review. I didn't find it, but I did notice a couple of problems here during the search. I haven't done a thorough review of this yet, but I'm leaving the review before I forget about these things.

package.json
crhallberg: This is a client-side library, so it's actually not built but imported. We copy it over to the client after install.

package.json
crhallberg: I'd argue that autocomplete is in the right spot and cookieconsent needs to be moved up because they are needed to run VuFind successfully. Anything moved by copynodemodules should be in non-development tools.

EreMaijala: But I'd say that copynodemodules is a development tool (upgrading dependencies is not a simple task typically done for customizing VuFind), and we don't need these otherwise.

demiankatz: The problem with reaching consensus here is that there are solid arguments for slicing this just about any way. You could say all of it is dev (because you don't need Javascript at all to run the system) or that almost none of it is dev (because most of the tools are being used to support VuFind, not to do Javascript-related development... perhaps eslint/jshint are the only things that are specifically for Javascript development). From a practical perspective, it doesn't much matter, since when you run `npm i` you're going to get all the dependencies by default, and I don't see a reason for anyone to bother excluding the dev dependencies. Ultimately, I don't really care which way we go, and I don't think it matters. But I would like us to pick one way or the other so we can stop having this conversation, and it can be clear where new things should go. :-)

demiankatz: Maybe the best thing for the present would be to go with @EreMaijala's approach (since that's what I've been telling people to do, and that's how the majority of the dependencies are currently organized) and we can revisit this as part of the JS tooling overhaul, which is on the roadmap for 10.0.

module/VuFind/src/VuFind/RecordTab/AbstractBase.php
```diff
@@ -110,6 +110,16 @@ public function supportsAjax()
         return true;
     }
 
+    /**
+     * What CSS and JS dependencies does this tab have?
+     *
+     * @return array
+     */
+    public function getDependencies()
```
demiankatz: This approach kind of blurs the lines between front end and back end code, and also makes it impossible to use different carousel implementations across different themes (which seems an unlikely requirement, but something that would be nice to be able to do in theory).

Does the back-end code really need to know about the front-end dependencies? Might a better approach be to simply load this content in the SimilarItemsCarousel.phtml file instead?

Also, on a related note, I feel like this is going to cause problems for AJAX-loaded tabs unless we are careful about our approach. I'm not seeing anything obviously broken in my tests, but it would probably be a good idea to test that closely when time permits.

package.json
demiankatz: We're having this conversation in multiple PRs. Let's try to remember to talk this out on the next Community Call so we can put it to rest one way or another. :-)

demiankatz (changes requested):
Thanks for the progress here, @crhallberg! Things seem to be working reasonably well for me. See below for a couple of comments. Do you think this is at a point where we should ask @sturkel89 to start thoroughly testing it, or do you still have more things you want to finish up first?

sturkel89:
I'll wait to hear from you two re: when it's time for me to test this branch! :)

module/VuFind/src/VuFind/RecordTab/AbstractBase.php
crhallberg: You're right. This was probably an over-reaction to including a relatively larger library. I'll weed this out.

themes/bootstrap3/js/common.js
```diff
@@ -668,6 +672,33 @@ function setupAutocomplete() {
   });
 }
 
+function setupCarousels(scope = document) {
```
demiankatz: Would it make sense to make this a module under the VuFind namespace for better scoping? This would enable automatic initialization as part of the main VuFind.init() call and reduce the number of exports/globals you need to declare.

themes/bootstrap3/templates/record/view.phtml
```diff
@@ -7,6 +7,11 @@
   // Set up standard record scripts:
   $this->headScript()->appendFile('record.js');
   $this->headScript()->appendFile('check_save_statuses.js');
+
+  // Add carousel library for Similar Items
+  $this->headScript()->appendFile('vendor/splide.min.js');
```
demiankatz: If we want to avoid putting this in such a global place, you could move these to RecordTab/similaritemscarousel.phtml and adjust the VuFind\RecordTab\SimilarItemsCarousel class to disallow AJAX loading. Or maybe there's a way to dynamically load the content as part of the template even if it is fetched by AJAX (I'm pretty sure you can get away with that for Javascript, but I'm less sure about the CSS).

demiankatz (changes requested):
Thanks, @crhallberg -- a couple more thoughts!

EreMaijala (approved):
I'll approve this now. The package.json refactoring will solve the remaining issue.

package.json
demiankatz: I believe that #3290 has introduced a much better solution to the package.json problem. @crhallberg, can you update this PR to move the splide-related installation logic to the places introduced there?

build.xml
```diff
@@ -372,6 +372,29 @@
     <exec command="php ${srcdir}/composer.phar install ${composer_extra_params}" passthru="true" checkreturn="true" />
   </target>
 
+  <!-- Copy files from node modules to themes -->
+  <target name="copynodemodules">
```
demiankatz: This Phing action should no longer exist after the latest merge with dev; it has been replaced by a different mechanism. I think it just needs to be deleted here. Any new copy operations should be added to https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/tools/copyDependencies.mjs

package.json
```diff
@@ -33,6 +33,8 @@
...
```
demiankatz: These files now belong in https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/package.json and should be removed from this top-level file.

demiankatz (changes requested):
Thanks for the progress here, @crhallberg -- looks like there may have been some issues in the latest dev merge, though, because it seems like the dependency installation stuff hasn't moved to the correct place yet. See notes below. Once everything is set up, you should be able to `composer run update-npm-dependencies` to put the files into place.

build.xml
crhallberg: That explains it! I was confused by this diff.

demiankatz:
@sturkel89, do you mind comparing the channel behavior and the "similar items" tab on the record page between this PR and the dev branch? The carousel library we're using has been completely replaced, so there may be some significant changes -- the question is whether anything is broken or needs improvement. Thanks!

demiankatz (changes requested):
I just ran the test suite on this, and a few things are failing:

```
There was 1 error:

1) VuFindTest\Mink\LinkResolverTest::testLinkOnRecordPageWithLinkInHoldingsAndAjaxTabLoading
Behat\Mink\Exception\DriverException: ReferenceError: $ is not defined
    at <anonymous>:2:5

/usr/local/vufind/vendor/dmore/chrome-mink-driver/src/ChromeDriver.php:1131
/usr/local/vufind/vendor/behat/mink/src/Session.php:343
/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:743
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/LinkResolverTest.php:106
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/LinkResolverTest.php:240
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:104

--

There were 5 failures:

1) VuFindTest\Mink\RecordActionsTest::testAddComment
Element not found: .modal.in index 0
Failed asserting that null is of type "object".

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:408
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordActionsTest.php:121
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:104

2) VuFindTest\Mink\RecordTest::testRecordTabsOnNormalId
--- Expected
+++ Actual
@@ @@
-'http://localhost/vufind_test/Record/testsample1#details'
+'http://localhost/vufind_test/Record/testsample1'

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:628
/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:646
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordTest.php:65
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordTest.php:108
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:104

3) VuFindTest\Mink\RecordTest::testRecordTabsOnSpacedId
--- Expected
+++ Actual
@@ @@
-'http://localhost/vufind_test/Record/dot.dash-underscore__3.space%20suffix#details'
+'http://localhost/vufind_test/Record/dot.dash-underscore__3.space%20suffix'

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:628
/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:646
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordTest.php:65
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordTest.php:119
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:104

4) VuFindTest\Mink\RecordTest::testRecordTabsOnPlusId
--- Expected
+++ Actual
@@ @@
-'http://localhost/vufind_test/Record/theplus+andtheminus-#details'
+'http://localhost/vufind_test/Record/theplus+andtheminus-'

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:628
/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:646
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordTest.php:65
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordTest.php:134
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:104

5) VuFindTest\Mink\RecordTest::testLoadInitialTabWithAjax
--- Expected
+++ Actual
@@ @@
-'http://localhost/vufind_test/Record/testsample1#details'
+'http://localhost/vufind_test/Record/testsample1'

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:628
/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:646
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordTest.php:65
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordTest.php:151
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:104

ERRORS!
Tests: 2641, Assertions: 14462, Errors: 1, Failures: 5, Skipped: 8.
```

I haven't spent a lot of time trying to figure out exactly what is wrong, but I noticed that one failure was related to adding comments, and when I tried to manually add a comment, I noticed this error in my browser console:

```
Uncaught TypeError: VuFind.carousels is undefined
    registerTabEvents http://localhost/vufind_test/themes/bootstrap3/js/record.js?_=1705092015:189
    ajaxLoadTabDone http://localhost/vufind_test/themes/bootstrap3/js/record.js?_=1705092015:231
    jQuery 6
    ajaxLoadTabReal http://localhost/vufind_test/themes/bootstrap3/js/record.js?_=1705092015:220
    recordTabsClick http://localhost/vufind_test/themes/bootstrap3/js/record.js?_=1705092015:387
    jQuery 8
    recordDocReady http://localhost/vufind_test/themes/bootstrap3/js/record.js?_=1705092015:352
    jQuery 13
record.js:189:5
```

Hopefully that's some clue as to what's going on!

crhallberg:
When I moved the loading out of theme.config.php I didn't patch it in where Splide is being loaded. Fixed!

demiankatz (changes requested):
@crhallberg, thanks for the fix. Unfortunately, after pulling your latest changes, merging dev, and resolving conflicts, I'm still getting exactly the same set of test failures as before.

demiankatz:
@sturkel89, this is ready for you to take another look when time permits! The automated tests are passing, so now the question is whether things work as expected for a human user, and whether the aesthetics are good. Please take a look at the channels functionality and the "similar items" carousel tab on record views. Let me know if you need any guidance on setting up a good test environment or finding the features. (It's possible you might at least want to turn on dynamic covers so that things look better while you are testing -- and I can't remember if you've done that before. Happy to help as needed!)

sturkel89 (changes requested):
I've reviewed this branch in all three themes and compared it with dev. Here's my review, divided into one part about the Similar Items tab in record view and a second part about the Channels page. 

Overall, I would prefer we switch to test, but only after a little tweaking of the side arrows in the Bootprint3 theme.

### Similar Items tab at bottom of item record

**Navigation controls - prefer test**

- The arrows are much easier to see (exception: bootprint3, where the left-hand arrow overlaps in a less pleasing way with the line with the green background). 
- The dots across the bottom indicating which page you're on have a nice contrast in tone between white and gray. I also like how in the dev version there's a larger dot indicating your current location. It would be nice to have both the color indicator and the size indicator, but we can probably only have one or the other.

**Design - prefer test**

- I’m not a fan of the shading at the edges of this area on dev.
- The solid background on test looks more modern than the shading on dev. However, I think an even paler gray would be better. (And maybe it would be fine to have no shading and just have the contents of this slider merge with the background white?)

**Height - no preference, but a question**

- It seems to me that the height of this area is too tall given the short height of the covers, but my own screenshots show that some book titles are very long and need the entire area. Is the height responsive to the length of the longest title in the set?

**Movement - prefer dev, but test can be set to match**

- I like the self-advancing action of the items in this area on dev. Chris says that this can be turned on in test. I would like to see it turned on.

I have screenshots of everything; here's a sample:

**Sandal - Test**
image

**Sandal - Dev**
image

Bootprint3 - Test
image

Bootprint3 - Dev
image

### Channels view

**Arrows - no preference**
- On a channels page, I think the side arrows are prominent in both test and dev; they are styled differently, but neither is appreciably better than the other.

**Navigation/feedback via dots at bottom of channel - prefer test**
- The main advantage of test is that it provides dots across the bottom of the channel display so the user can tell how many screens of related items are included in the channel.

**Layout - dev is better in Bootprint3**
- The main advantage of dev is that the placement and size of the arrows works better on Bootprint3; the arrows fit in the white area between the content and the green background. In test in Bootprint2, the arrows are positioned awkwardly and overlap the green.

Sandal - Test
image

Sandal - Dev
image

Bootprint3 - Test
image

Bootprint3 - Dev
image

crhallberg:
Style-wise, do we think it would be better to always have the title overlay showing, vs hiding it until hover/focus? I feel like that's a lot of information not being shown.

demiankatz:
@crhallberg, are you proposing that we show the overlay on all items, or that we only show the overlay on one item at a time, but that we don't hide it when focus is lost? I think showing overlays on everything all the time might be excessive, but maybe ensuring that we always have one overlay visible would be a helpful cue to indicate that more information is available. Though maybe that would also make it harder to tell where focus lies. (As you can see, I don't have strong opinions about this, and perhaps I'm a little confused :-) ).

crhallberg:
I just find the amount of information available without interaction to be lacking a bit. I wondering if we can do better.

demiankatz:
@crhallberg, I'm certainly open to revisions. I think the related items carousel is a demo you put together at a long-past VuFind Summit that never got much further development, so there's certainly room for improvement. However, I'd suggest that we hold off on changing things until after we finish this PR -- let's get the existing functionality migrated to the new library first, and then we can improve from there.

crhallberg:
image

I've added a little information to keep it from being so sparse, especially if the thumbnail is missing or unclear.

crhallberg:
@demiankatz when you get the chance, I'd appreciate a run on the full test suite!

demiankatz:
@crhallberg, all tests are passing. I think the question now is whether there's still outstanding work from @sturkel89's February 8 comment and/or whether it's time for her to re-test anything.

crhallberg:
I believe I've addressed all of the style feedback. Happy that all the tests are passing! This one is ready for a final checks before merging.

demiankatz:
Thanks, @crhallberg -- I've requested a re-review by @sturkel89 and will take one last look once she's given the thumbs-up.

sturkel89:
I took another look at Splide in all three themes.

### Similar items

Splide in Similar items looks good in all three themes. I think the Similar Items feature is ready to go.

--------------
### Channels

**General observation:**
I notice the difference mousing over items in Channels does NOT cause the large highlight focus that we see in Similar Items. Instead, there’s a pop-out title on hover, and a much larger menu on click. I know that this isn’t new, but it’s noticeable to me that these two uses of a similar feature are implemented so differently. Is there a reason? Would one benefit from having the functionality of the other?

### Bootstrap3:
I think Channels are working well in this theme. The arrows look good at most widths; they are fully outside the carousel, unless it gets too narrow in which case they fully overlap the carousel items.

image


### Bootprint3:
I think Channels could use some tweaking in this theme. The arrows overlap the border between the white and the green at most widths. I like the size of the arrows, but I’m not crazy about the overlap. Would it be possible to make the carousel a bit narrower, with five items instead of six, so that there’s a wider margin to hold the arrows? (In Similar Items, the carousel is only four items wide, so there’s plenty of room for arrows.)

image

### Sandal:
Every time I try to switch to Sandal, my browser hangs and I can’t click on anything. There is something wrong with Sandal in test.

sturkel89 (changes requested):
Whoops - commented instead of reviewing. Please see my comment; I'm asking for tweaks to bootprint3 and I'm unable to load channels in sandal.

demiankatz:
Just to second @sturkel89's observation: channels are totally broken in Sandal for me too, so it's not just her environment causing this.

Also, while experimenting with this, I ran into another popover bug that seems to impact all versions of the code all the way back to the release-9.1 branch, so it seems that the fix in #3607 was not comprehensive. Here's how to reproduce:

1.) In the standard test environment, go to the Channels page.

2.) Scroll down until you find the same record in multiple channels -- the "Author, Primary" and "Author, Secondary" channels are a good spot for this.

3.) Click on a duplicated record (e.g. "Fake record 1 with multiple relators"). The popover appears.

4.) Now click on the same record in a different channel. The open popover closes, but no new popover appears. Multiple clicks will not cause anything to happen.

5.) Now click on the original copy of the record again. You'll get an incomplete popover like we saw in the #3607 bug. From this point forward, clicking on either record will cause crazy behavior -- flashes of content, incomplete popovers, etc., etc.

Fixing this problem is probably a higher priority than finishing Splide. Please let me know if you'd like me to write a test case to demonstrate it.

demiankatz:
Update: see #3649 for fix to popover bug discussed above. Once we get that merged through to here, we should resume investigating the other known problems specific to this branch.

crhallberg:
@demiankatz @sturkel89 I was not able to see the Sandal not working error. I did update the mobile styles however.

demiankatz (changes requested):
@crhallberg,

I took another look at this. I'm still seeing the sandal problem (at least in Firefox, my usual browser for testing): channels load slowly, if at all, in Sandal, even though they seem normal in bootprint3 and bootstrap3. Here's a screen shot of my latest attempt at Sandal -- note the "page is slowing down your browser" warning -- something is causing Javascript to do heavy processing, but only in one theme!

image

I also notice that `npm run build:css` is throwing errors about missing include files in the bootstrap5 theme -- I suspect that significant work will be needed to reconcile the bootstrap3 and bootstrap5 versions of the conversion. @EreMaijala might have thoughts on how complex this is likely to be.

In any case, all of this adds up to suggest to me that we're not going to get this done in time for release 10.0 -- but we can discuss on today's Community Call!

EreMaijala:
Oh no, @pasitiis noticed that there haven't been new releases for Splide since 2022, and the last commit is also more than a year ago. Seems like the development may have stalled. Apparently https://swiperjs.com/ might be an alternative, but can't vouch for it or any other.

demiankatz:
Argh, that's quite concerning. Once again, it seems like abstraction needs to be the order of the day, since we may need to hot-swap carousels from time to time. @crhallberg, do you think there's work here that might be able to be generalized? Should we, for example, open a separate PR to introduce the vf-carousel classes independently of the Splide upgrade? I think given the size of the change here and the proximity of the release, we probably should delay going forward with what we have here until we can get more information. Do you disagree?

crhallberg:
@demiankatz I don't disagree. The `vf-carousel` changes were necessary to separate the carousels from Bootstrap. This should definitely be abstracted into a component like menu-button.

@EreMaijala Swiper looks like a good replacement!

demiankatz:
I'm pushing all theme-related PRs that were not finished in time for 10.0 to the 11.0 milestone; it will be easier to finish these if we only have to operate on one theme at a time. I think we should establish a dev-11.0 branch fairly early in the 10.1 development process so that this work can move forward as soon as possible.

pasitiis:
We have been using Splide in Finna for a few years, but it’s not perfect from an accessibility perspective for our needs. Although Splide follows the ARIA carousel pattern, using it with a screen reader is somewhat awkward for browsing items. It would be better to forgo the carousel pattern altogether, structure it as a list, and create a carousel-like appearance, either with or without pagination.

I could be interested in helping, regardless of which component the work continues with.

demiankatz:
I like the idea of a styled list, if doing that correctly isn't terribly burdensome. Replacing a third-party library with standard HTML worked well for hierarchy trees, so maybe carousels could be handled similarly. I defer to @crhallberg on this decision, though, as he probably has a better sense than I do of exactly how complex this might end up becoming. :-)

crhallberg:
I do agree that a styled list would be much more accessible for screen readers and would work well on mobile where horizontal scrolling is increasingly common. However, we would need a solution for desktop. Splide not being the solution is fine, this PR has some good value of making it easier to replace the library with another. cc @demiankatz @pasitiis

crhallberg:
I am hesitant to return to rolling our own carousel, but we can piggyback some work at least. W3m has an [example accessible carousel](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-1-prev-next/). We could start with that code and add some modern touches such as [CSS scroll snapping](https://medium.com/web-dev-survey-from-kyoto/vanilla-js-carousel-that-is-accessible-swipeable-infinite-scrolling-and-autoplaying-5de5f281ef13#6082).

demiankatz:
Closing as superseded by #4557.

PR #2935, opened 2023-06-08
Exclude FOLIO reserves that have been suppressed from discovery

meganschanz:
This change is to exclude course reserves from FOLIO that have instance records marked as suppressed from discovery. We added this to our custom module a while back since it was resulting in a count mis-match (telling the user there were some higher number of course reserves but when they clicked on the reserve result, there were fewer actual reserves because we did not ingest those instance records due to them being suppressed).

The change was added a few years back in the FOLIO API: https://issues.folio.org/browse/MODCR-65 and is documented in their API documentation: https://s3.amazonaws.com/foliodocs/api/mod-courses/r/courses.html#coursereserves_reserves_get.

We thought it would be useful to suggest this change to VuFind, since as a discovery layer, you would expect it to filter suppressed records.

### DIFF
--- a/config/vufind/Folio.ini
+++ b/config/vufind/Folio.ini
@@ -166,6 +166,11 @@ in_transit[] = "Open - Awaiting delivery"
 ; only the name will be displayed:
 displayCourseCodes = false
 
+; If set to true, finding course reserves will include FOLIO instance records
+; even if they are suppressed from discovery; if false, the suppressed records
+; are excluded.
+includeSuppressed = false
+
 [Availability]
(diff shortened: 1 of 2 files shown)


demiankatz (changes requested):
Thanks, @meganschanz -- this is a good idea and makes sense to me as a default behavior, but maybe it's worth adding a setting somewhere in Folio.ini to allow suppressed items to be included in case somebody wants to do that for some reason. What do you think?

meganschanz:
Yeah that makes sense, just in case someone wanted those records for some reason.

demiankatz (approved):
Thanks, @meganschanz!

PR #1338, opened 2019-03-11
Add support for fetching summary holdings from Sierra when using the …

EreMaijala:
…REST API.

Includes a fix for ordering of results.

Note that support for fetching the required information for the holdings records was only recently added to the v5 API, so I used a pseudo-version 5.1 to indicate support for it.

PR #2936, opened 2023-06-09
Sierra: Fix retrieval of holdings.

EreMaijala:
I was caught off guard by the fact that PHP likes array indexes to be ints.

The exclusion of deleted and suppressed records was commented out since the beginning (see #1338), and I can't think of a reason other than testing something and then forgetting to uncomment them.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2937, opened 2023-06-09
[release-9.0] Sierra: Fix holdings line number comparison (again).

EreMaijala:
Since the const type is different from dev, this needs different comparison.

This will conflict with dev branch and should be resolved so that dev keeps its current version.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2938, opened 2023-06-09
[release 9.0] Sierra: Add missing mandatory fields to getHolding response.

EreMaijala:
When there was a holdings record without items for the location, the result was missing fields number and reserve.


demiankatz (approved):
Thanks, @EreMaijala!

PR #2939, opened 2023-06-09
Drop type attribute from Matomo load script.

EreMaijala:
This avoids an HTML validation notice `The “type” attribute is unnecessary for JavaScript resources.`


demiankatz (approved):
Thanks, @EreMaijala!

PR #2940, opened 2023-06-12
Fix page param out of range at search results

xmorave2:
To reproduce:
1. Enable ability to change number of result per page
2. Make e search
3. Go to last page
4. Change the results per page to higher option
5. Without the fix, you should see empty results. With this patch, you'll be redirected to last available page

Or just change the page param in url manually ;)

I am not really sure if this is the right place where to make a fix, so suggestion are more than welcome ;)


demiankatz:
Thanks, @xmorave2, this seems to me like a reasonable place for the logic. I'd also be interested to hear what @EreMaijala thinks, though, in case I'm overlooking anything.

module/VuFind/src/VuFind/Controller/AbstractSearch.php
```diff
@@ -373,7 +373,16 @@ protected function getSearchResultsView($setupCallback = null)
...
-        $view->params = $results->getParams();
+        $view->params = $params = $results->getParams();
+
+        // For page parameter being out of results list, we want to redirect to correct page
+        $page = $params->getPage();
+        $lastPage = ceil($results->getResultTotal() / $params->getLimit());
+        if ($page > $lastPage) {
+            $queryParams = $request;
+            $queryParams['page'] = $lastPage;
+            $this->redirect()->toRoute('search-results', [], [ 'query' => $queryParams ]);
```
EreMaijala: Should this be returned to avoid further processing?

xmorave2: Definitely, thanks @EreMaijala, I fixed it.

EreMaijala (approved):
Looks good to me now!

demiankatz:
I just tried running the full test suite and am getting failures, but it may be due to a configuration error on my end. I'm out of time for today but will try to give this another look tomorrow!

EreMaijala (dismissed):
Right, there's an issue when there are no results. `$lastPage` becomes 0 and causes and endless redirect.

demiankatz (changes requested):
Thanks for figuring this out, @EreMaijala. Can you fix this edge case when time permits, @xmorave2? Maybe it would also be useful to add a Mink test to capture the "access past end of list" behavior to catch future regressions.

demiankatz (approved):
Thanks, @xmorave2 -- all tests are passing now, and I added a small test to cover this edge case.

xmorave2:
Thanks @demiankatz

PR #2941, opened 2023-06-15
[VUFIND-1617] Prevent single sign-on (SSO) from breaking out of lightbox

ThoWagen:
Solution for ticket [VUFIND-1617](https://openlibraryfoundation.atlassian.net/jira/software/c/projects/VUFIND/issues/VUFIND-1617)

The idea behind the solution is the following:
When storing an url that is a lightbox in the followup plugin, the base url of the page also is stored. When returning from SSO, it is checked if there is such an extra base url and instead of redirecting to the lightbox url directly, the base page is loaded first. Then the page creates a lightbox and the returning from SSO is performed inside of it.

TODO
- [x] Note BC breaks when merging (replacement of VuFind\Controller\AbstractBase::getFollowupUrl with getAndClearFollowupUrl).
- [x] Make sure full test suite passes before merging.


module/VuFind/src/VuFind/Controller/MyResearchController.php
```diff
@@ -155,7 +155,7 @@ public function homeAction()
                     if (
                         $this->params()->fromPost('processLogin')
                         && $this->inLightbox()
-                        && empty($this->getFollowupUrl())
+                        && empty($this->hasFollowupUrl())
```
demiankatz: I don't think you need an empty check here since `hasFollowupUrl` returns a boolean.

module/VuFind/src/VuFind/Controller/AbstractBase.php
```diff
@@ -736,16 +735,45 @@ protected function normalizeUrlForComparison($url)
...
      * @return string
      */
-    protected function getFollowupUrl()
-    {
-        return $this->followup()->retrieve('url', '');
+    protected function getAndClearFollowupUrl($checkRedirect = false)
+    {
+        if ($url = $this->followup()->retrieve('url', '')) {
+            $lightboxParent = $this->followup()->retrieveAndClear('lightboxParent');
+            $this->clearFollowupUrl();
```
demiankatz: Is there a strong reason to clear this separately rather than using retrieveAndClear above? If so, is it accurate to call the method getAndClearFollowupUrl if it's only clearing conditionally?

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -107,6 +107,12 @@
     ?>
     <?=$this->cookieConsent()->render()?>
     <?=$this->headScript() ?>
+    <?php if(isset($lightboxParent)): ?>
```
demiankatz: Minor style -- we use `if (` instead of `if(`, and I don't think you need the opening whitespace inside the `<script>` tags.

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -107,6 +107,12 @@
...
+      <script> window.lightboxParent = "<?= $lightboxParent ?>"</script>
```
demiankatz: Should we apply some kind of escaping to $lightboxParent and $lightboxChild? Should we add semi-colons to terminate these assignments?

themes/bootstrap3/js/lightbox.js
```diff
@@ -168,6 +168,14 @@ VuFind.register('lightbox', function Lightbox() {
       }
       obj.url += parts.length < 2 ? '' : '#' + parts[1];
     }
+    if (window.lightboxParent){
+      let parts = obj.url.split('#');
+      obj.url = parts[0].indexOf('?') < 0
+        ? parts[0] + '?'
+        : parts[0] + '&';
+      obj.url += '&lightboxParent=' + window.lightboxParent;
```
demiankatz: Should we `encodeURIComponent` the lightboxParent?

demiankatz (changes requested):
Thanks, @ThoWagen, this looks promising, though I haven't actually tried any hands-on testing yet. A few comments below based on an initial read-through.

I'm going to set this to the 10.0 milestone, since it contains some breaking changes and may be a bigger shift than we would want to make in a minor release. If getting it finished faster is important, though, there may be ways to achieve this in a more back-compatible way, and I'm certainly open to discussing that.

Thanks again for taking the time to work on this!

demiankatz:
I'm also interested in hearing what @crhallberg and @EreMaijala think of this when they have a moment!

module/VuFind/src/VuFind/Controller/MyResearchController.php
ThoWagen: Fixed

module/VuFind/src/VuFind/Controller/AbstractBase.php
ThoWagen: You are right. Now retrieveAndClear is used and clearFollowupUrl is removed.

themes/bootstrap3/templates/layout/layout.phtml
ThoWagen: fixed

themes/bootstrap3/templates/layout/layout.phtml
ThoWagen: urlencoding $lightboxParent and $lightboxChild now. 
I don't think the missing semi-colons are causing a problem here but I added them.

themes/bootstrap3/js/lightbox.js
ThoWagen: window.lightboxParent is now already urlencoded

EreMaijala:
Sorry, it'll take at least until Monday for me review this.

demiankatz:
No rush, @EreMaijala -- I'm also not going to be back in the office until next week, but I'll look forward to discussing it when time permits. Thanks for the progress, @ThoWagen!

themes/bootstrap3/templates/layout/layout.phtml
demiankatz: Agreed, the semi-colons don't make a technical difference, but for the sake of style I like to include them. :-)

themes/bootstrap3/js/lightbox.js
```diff
@@ -168,6 +168,14 @@ VuFind.register('lightbox', function Lightbox() {
       }
       obj.url += parts.length < 2 ? '' : '#' + parts[1];
     }
+    if (window.lightboxParent){
```
EreMaijala: The prevalent style has a space after the closing bracket, I'd like to follow it.

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -107,6 +107,12 @@
     ?>
     <?=$this->cookieConsent()->render()?>
     <?=$this->headScript() ?>
+    <?php if (isset($lightboxParent)): ?>
+      <script>window.lightboxParent = "<?= urlencode($lightboxParent) ?>";</script>
```
EreMaijala: Unless there's a really good reason to store these in window, I'd follow the existing practice of using the VuFind object for state (see `$appendScripts` right above). Add the scripts to `$appendScripts`, otherwise any CSP will block them since they won't get a nonce.

I'd also prefer for the `lightboxParent` and `lightboxChild` to not be urlencoded when they're not part of a URL. Otherwise the way they're used later on may be confusing, i.e. it may look like they're missing proper encoding when added to a URL. Plus you wouldn't need the `decodeURIComponent` and handling of spaces as pluses. You could use e.g. `json_encode((string)$lightboxParent)` to ensure they're safe in the JS code.

module/VuFind/src/VuFind/Controller/AbstractBase.php
```diff
@@ -736,16 +735,44 @@ protected function normalizeUrlForComparison($url)
...
+            $lightboxParent = $this->followup()->retrieveAndClear('lightboxParent');
+            // If a user clicks on the "Your Account" link, we want to be sure
+            // they get to their account rather than being redirected to an old
+            // followup URL. We'll use a redirect=0 GET flag to indicate this:
+            if (!$checkRedirect || $this->params()->fromQuery('redirect', true)) {
+                if (isset($lightboxParent) && $this->getAuthManager()->getSessionInitiator($this->getServerUrl())) {
+                    $this->followup()->store([], $url);
+                    $url = new \Laminas\Uri\Uri($lightboxParent);
+                    $params = $url->getQueryAsArray();
+                    $params['lightboxChild'] = urlencode($this->getServerUrl());
```
EreMaijala: Why is this double-encoded? The `urldecode` calls would also be unnecessary without the `urlencode` here.

module/VuFind/src/VuFind/Controller/AbstractBase.php
```diff
@@ -736,16 +735,44 @@ protected function normalizeUrlForComparison($url)
...
-        return $this->followup()->retrieve('url', '');
+    protected function getAndClearFollowupUrl($checkRedirect = false)
+    {
+        if ($url = $this->followup()->retrieveAndClear('url', '')) {
+            $lightboxParent = $this->followup()->retrieveAndClear('lightboxParent');
+            // If a user clicks on the "Your Account" link, we want to be sure
+            // they get to their account rather than being redirected to an old
+            // followup URL. We'll use a redirect=0 GET flag to indicate this:
+            if (!$checkRedirect || $this->params()->fromQuery('redirect', true)) {
+                if (isset($lightboxParent) && $this->getAuthManager()->getSessionInitiator($this->getServerUrl())) {
```
EreMaijala: Use `null !== $lightboxParent` instead of isset.

module/VuFind/src/VuFind/Controller/AbstractBase.php
```diff
@@ -736,16 +735,44 @@ protected function normalizeUrlForComparison($url)
...
+     *
      * @return string
      */
-    protected function getFollowupUrl()
-    {
-        return $this->followup()->retrieve('url', '');
+    protected function getAndClearFollowupUrl($checkRedirect = false)
+    {
+        if ($url = $this->followup()->retrieveAndClear('url', '')) {
+            $lightboxParent = $this->followup()->retrieveAndClear('lightboxParent');
```
EreMaijala, line 772: Move this further down inside the `if (!$checkRedirect...)` block since it's not needed outside of it.

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -107,6 +107,12 @@
     ?>
     <?=$this->cookieConsent()->render()?>
     <?=$this->headScript() ?>
+    <?php if (isset($lightboxParent)): ?>
```
EreMaijala: I'd use `!empty()` instead of `isset()` so that e.g. an empty string is ignored.

EreMaijala (changes requested):
I didn't get to actually test this yet, but I've left some comments about my observations here.

themes/bootstrap3/templates/layout/layout.phtml
demiankatz: I'll second the suggestion that it's probably more clear to encode at the time we need encoding, instead of in advance. Otherwise, it can become confusing, since it requires looking in multiple parts of the code to figure out whether or not encoding has already been applied. (And if there were a strong reason to encode in advance, I would reflect the encoding in the variable name, like lightboxParentUrlEncoded, to make this really clear... but obviously not encoding seems simpler than weird variable names in most situations).

themes/bootstrap3/js/lightbox.js
ThoWagen: fixed

themes/bootstrap3/templates/layout/layout.phtml
ThoWagen: fixed

module/VuFind/src/VuFind/Controller/AbstractBase.php
ThoWagen: fixed

module/VuFind/src/VuFind/Controller/AbstractBase.php
ThoWagen: You are right. I was not aware of the url already being encoded, when its added this way. Removed it and the urldecode.

themes/bootstrap3/templates/layout/layout.phtml
ThoWagen: Thank you for the hint. I moved both urls to the VuFind object. Also using json_encode now instead of urlencode.

module/VuFind/src/VuFind/Controller/AbstractBase.php
ThoWagen: This command also clears the lightboxParent. If it is moved inside of the if block it might cause problems later on.

themes/bootstrap3/js/common.js
```diff
@@ -7,6 +7,10 @@ window.console = window.console || { log: function polyfillLog() {} };
 var VuFind = (function VuFind() {
   var defaultSearchBackend = null;
   var path = null;
+  var lightbox = {
```
ThoWagen: I just realized that this is most likely not the place where this should be added. I will change this.

ThoWagen: Done

module/VuFind/src/VuFind/Controller/AbstractBase.php
```diff
@@ -736,16 +735,44 @@ protected function normalizeUrlForComparison($url)
...
+            // If a user clicks on the "Your Account" link, we want to be sure
+            // they get to their account rather than being redirected to an old
+            // followup URL. We'll use a redirect=0 GET flag to indicate this:
+            if (!$checkRedirect || $this->params()->fromQuery('redirect', true)) {
+                if (null !== $lightboxParent && $this->getAuthManager()->getSessionInitiator($this->getServerUrl())) {
+                    $this->followup()->store([], $url);
+                    $url = new \Laminas\Uri\Uri($lightboxParent);
```
demiankatz: It might be a good idea to use a different variable name than $url here (maybe $parentUrl) since we previously used $url for something different in the same function, and that makes things slightly confusing.

module/VuFind/src/VuFind/Controller/AbstractBase.php
```diff
@@ -346,16 +350,11 @@ public function forceLogin($msg = null, $extras = [], $forward = true)
...
 
-        // We don't want to return to the lightbox
-        $serverUrl = $this->getServerUrl();
-        $serverUrl = str_replace(
-            ['?layout=lightbox', '&layout=lightbox'],
-            ['?', '&'],
-            $serverUrl
-        );
+        // store parent url of lightboxes
+        $extras['lightboxParent'] = urldecode($this->getRequest()->getQuery('lightboxParent'));
```
demiankatz: Why does this need to be URL decoded? Where is it getting encoded? Maybe a comment would be in order to help clarify what is going on here.

demiankatz (changes requested):
Thanks, @ThoWagen. I took another quick look at this and left a couple more small comments. I still haven't done hands-on testing yet, but can do so after these things are addressed if you'd like me to! (I expect that @EreMaijala won't be back to re-review until August, so we can either wait for him, or I can put a little more time into this in the meantime).

module/VuFind/src/VuFind/Controller/AbstractBase.php
ThoWagen: This way I wanted to avoid an additional return statement, but renamed it, since it is certainly less confusing this way.

module/VuFind/src/VuFind/Controller/AbstractBase.php
ThoWagen: It does not need to be decoded. I removed it.

ThoWagen:
@demiankatz I will be on vacation myself for the next two weeks. If you find the time to test this, I wouldn't mind. But this is nothing urgent.

demiankatz:
Thanks, @ThoWagen! I will be out of office for a few days as well, but I should be able to do some hands-on testing of this code before you return from your own vacation. I hope you have a good time!

demiankatz:
@ThoWagen, I haven't had a chance to try this yet, but today I built #2987 so that we potentially have a mechanism that can be used to create automated tests for SSO behavior, which will make it easier to try out things like this PR and to prevent future regressions. I'll spend some more time on this once that work has been reviewed and merged.

demiankatz:
Another update: now that #2987 is merged, I'm in a good position to test this. I started by running the full automated test suite today, which revealed a failure caused by a pre-existing bug that was revealed by code changes in this PR. I've opened #2989 to address that problem, and will resume testing here once that has been reviewed and merged.

demiankatz:
My bug fix is now merged and the full test suite is passing. Some quick hands-on testing also looks very promising. I think at this point next step is to just get some more input (still interested in hearing from @crhallberg about the Javascript changes, for example) and perhaps write some new Mink tests using the SimulatedSSO authentication to exercise the code.

ThoWagen:
I added two a Mink tests. One just tests the SimulatedSSO login and logout and the other tests the SSO login inside a lightbox by trying to save a record to the list. 

Is this sufficient or does it need more coverage?

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SsoTest.php
```diff
@@ -0,0 +1,117 @@
...
+        $this->changeConfigs($this->getConfigIniOverrides());
+        $session = $this->getMinkSession();
+        $session->visit($this->getVuFindUrl());
+        $page = $session->getPage();
+
+        // Login
+        $this->clickCss($page, '#loginOptions a');
+
+        // Log out
+        $this->clickCss($page, '.logoutOptions a.logout');
```
demiankatz: Should add an assertion after this to be sure it worked -- e.g make sure login link is back again.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SsoTest.php
```diff
@@ -0,0 +1,117 @@
...
+        $this->changeConfigs($this->getConfigIniOverrides());
+        $session = $this->getMinkSession();
+        $session->visit($this->getVuFindUrl() . '/Record/testsample1');
+        $page = $session->getPage();
+
+        // Try to save record to list
+        $this->clickCss($page, '.record-nav .save-record');
+
+        // Login in lightbox
+        $this->clickCss($page, '.modal-body .btn.btn-link');
```
demiankatz, line 100: Maybe assert something about  this link before clicking it, like confirm it says "log in" (or whatever it actually says).

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SsoTest.php
```diff
@@ -0,0 +1,117 @@
...
+        $this->closeLightbox($page);
+
+        // Check that we are still on the record page
+        $this->assertEquals(
+            'Journal of rational emotive therapy : the journal of the Institute for Rational-Emotive Therapy.',
+            $this->findCss($page, '.record .media-body h1')->getText()
+        );
+
+        // Log out
+        $this->clickCss($page, '.logoutOptions a.logout');
```
demiankatz, line 130: Not sure if logging out is strictly necessary, though it doesn't hurt. Might be worth an assertion afterward, as above. Maybe you could use a protected support method like `logOutAndAssertSuccess` to share the logic.

demiankatz (changes requested):
Thanks, @ThoWagen, this test is a great start! We could obviously potentially do a whole lot more, in terms of testing every possible place where SSO could lead to different login behavior. But I don't expect that as part of this PR -- I think you've come up with a good "minimum viable test" to get us started. I just left a few suggestions of places where I thought it might be worth adding assertions to make the tests a little more robust.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SsoTest.php
ThoWagen: Done

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SsoTest.php
ThoWagen: Done

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SsoTest.php
ThoWagen: I implemented the logOutAndAssertSuccess method. I also don't think that logging out is necessary here, but it should not hurt either.

demiankatz: Thanks for these latest changes -- looks good! You might want to consider passing $page to the logOutAndAssertSuccess method since you already have access to it in the calling context, and it will simplify the method a little bit... but no pressure to do so if you prefer it as-is.

Beyond that, I'm still waiting for feedback from others, but from my perspective, this is looking very good.

ThoWagen: I thought about this but decided against it. I prefer this because calling the method is simpler this way :)

demiankatz:
Thanks again, @ThoWagen -- this looks complete from my perspective, but I'll wait on an approval until we get some more reviews (which will hopefully happen next month when people are less busy and/or back from holiday). :-)

ThoWagen:
I made two changes:
1. In getAndClearFollowupUrl the followup url is used as the lightboxChild directly. Otherwise, there might be problems when calling the current url twice. And I don't see a reason why this should not work. Not sure why I didn't implement it this way in the first place.
2. The lightboxParent url is urlencoded again, since it is used as a query parameter for the lightbox requests.

EreMaijala:
>     2. The lightboxParent url is urlencoded again, since it is used as a query parameter for the lightbox requests.

I find this a bit confusing. I think it would be easier to handle if lightbox.parent was not encoded, but encoding be applied as necessary. In practice something like: remove [line 104 from layout.phtml](https://github.com/vufind-org/vufind/pull/2941/files#diff-9a9c1fe114b3ed709c797c908ebdde9d9f67f85bb3db14db3ead4aed5304e57fR104) and add `urlencode` to [lightbox.js line 179](https://github.com/vufind-org/vufind/pull/2941/files#diff-571ab64362ef6738924a96a721d69e7460aefdd873ea2a86f21e552e428c94e3R179). I'd also consider using the [URL object](https://developer.mozilla.org/en-US/docs/Web/API/URL_API) to manipulate the URL to avoid all the manual work.

ThoWagen:
I did the encoding in php since it is only used as a part of an url in the JS code. But, I guess you are right. I moved it into the JS to reduce confusion in the future.

When I started with these changes I first used the URL object. But unfortunately, it caused some problem. If I recall correctly, the url is not always a complete url. Sometimes it's only the path and the query parameters. And the URL object can not be used in that case.

EreMaijala:
@ThoWagen Ok, right, then URL won't do it. You could still use [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams), but it's maybe overkill for a simple case like this.

demiankatz:
Thanks, @EreMaijala and @ThoWagen -- I agree that URLSearchParams might be useful here, though I also don't object to keeping things in their current form for the moment.

ThoWagen:
I created an extra function for adding query parameters. It uses URLSearchParams.

ThoWagen:
It seems like the failing test has nothing to do with the changes in this PR. It probably has something to do with #3035.

demiankatz:
@ThoWagen, if you run `vendor/bin/phing php-cs-fixer` and commit the resulting changes, that will fix everything.

ThoWagen:
@demiankatz Thanks! That did the trick.

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -99,6 +99,21 @@
           VuFind.setCspNonce('{$cspNonce}');
           VuFind.setCurrentSearchId($searchId);
           JS;
+
+      if (!empty($lightboxParent)) {
+        $lightboxParent = json_encode((string)$lightboxParent);
+        $appendScripts[] = <<<JS
+                   VuFind.lightbox.parent = {$lightboxParent};
```
demiankatz: Since these new scripts are just one-liners, maybe it would be cleaner to do it with a single quoted string instead of HEREDOC.

module/VuFind/src/VuFind/Controller/AbstractBase.php
```diff
@@ -737,16 +736,44 @@ protected function normalizeUrlForComparison($url)
...
-        return $this->followup()->retrieve('url', '');
+    protected function getAndClearFollowupUrl($checkRedirect = false)
+    {
+        if ($url = $this->followup()->retrieveAndClear('url', '')) {
+            $lightboxParent = $this->followup()->retrieveAndClear('lightboxParent');
+            // If a user clicks on the "Your Account" link, we want to be sure
+            // they get to their account rather than being redirected to an old
+            // followup URL. We'll use a redirect=0 GET flag to indicate this:
+            if (!$checkRedirect || $this->params()->fromQuery('redirect', true)) {
+                if (null !== $lightboxParent && $this->getAuthManager()->getSessionInitiator($this->getServerUrl())) {
```
demiankatz: Does the authentication method actually matter here? Or should we instead be checking whether the request is already in a lightbox or not? In practice, this scenario currently only arises from SSO, but that may be something of a coincidence, and the thing that really impacts the desired behavior is whether we're already in a lightbox or need to restore a lightbox from prior to a redirect. Sorry if I'm off base here, but I think there might be an opportunity to generalize the code further.

themes/bootstrap3/js/lightbox.js
```diff
@@ -46,6 +49,18 @@ VuFind.register('lightbox', function Lightbox() {
     return document.dispatchEvent(event);
   }
 
+  function _addQueryParameters(url, params) {
+    let fragmentSplit = url.split('#');
+    let paramsSplit = fragmentSplit[0].split('?');
+    let searchParams = new URLSearchParams(paramsSplit.length > 1 ? paramsSplit[1] : "");
+    for ( const [key, value] of Object.entries(params)) {
```
demiankatz: Nitpick: unnecessary space before `const`.

demiankatz (changes requested):
It's been in my queue for a while now to give this another review, and I've finally gotten that far. Thanks for your patience, @ThoWagen! :-)

themes/bootstrap3/templates/layout/layout.phtml
ThoWagen: done

themes/bootstrap3/js/lightbox.js
ThoWagen: done

module/VuFind/src/VuFind/Controller/AbstractBase.php
ThoWagen: You are absolutely right. Thanks for the hint!  I changed this.

demiankatz: Great, thanks -- I'll give this one more thorough test and review tomorrow to see if I can think of or find any other issues, but I have a feeling we may be done. More soon!

ThoWagen:
Two small fixes:

1. Now setting _currentUrl in loadConfiguredLightbox() in lightbox.js. Otherwise, submitting forms that don't provide an action attribute won't work.

2. When storing followup URLs, the queryParameter 'lightboxParent' is removed. Not removing it was not really a bug but led to accumulating multiple lightboxParent parameters in successive requests.

ThoWagen:
Another two small fixes:
1. Removing lightboxChild from lightboxParent when loading the child. This is not a bug but it is more consistent that the parent does not accumulate multiple children.
2. Using `set` instead of `append` in `_addQueryParameters` in lightbox.js so that existing params get overwritten instead of duplicated.

demiankatz:
I see that @EreMaijala had requested changes, but I believe work has stabilized here, so I'm requesting another review. I also plan to take a look at this soon.

themes/bootstrap3/js/lightbox.js
```diff
@@ -467,6 +480,28 @@ VuFind.register('lightbox', function Lightbox() {
       });
     });
   }
+
+  function loadConfiguredLightbox(){
```
EreMaijala: Missing space before opening bracket.

themes/bootstrap3/js/lightbox.js
```diff
@@ -467,6 +480,28 @@ VuFind.register('lightbox', function Lightbox() {
...
+    if (VuFind.lightbox.child) {
+      // remove lightbox reference
+      let lightboxChild = VuFind.lightbox.child;
+      VuFind.lightbox.child = null;
+      let url = new URL(window.location.href);
+      let params = url.searchParams;
+      params.delete('lightboxChild');
+      url.search = params.toString();
```
EreMaijala: This can be simplified a bit (and use `const` since the object reference won't change):
```
      const url = new URL(window.location.href);
      url.searchParams.delete('lightboxChild');
```

EreMaijala (changes requested):
Just a couple more minor details. :)

EreMaijala:
Now I can't really test this properly, but I managed to set up Shibboleth so that I could try. However, at least an unsuccessful login breaks out of the lightbox. Is this a part that's intentionally not covered here?

demiankatz:
> Now I can't really test this properly, but I managed to set up Shibboleth so that I could try. However, at least an unsuccessful login breaks out of the lightbox. Is this a part that's intentionally not covered here?

Thanks for taking another look at this. Note that you can set authentication method to `SimulatedSSO` to perform single sign-on without a real system -- it always logs you in to a fixed account name. Perhaps we should make SimulatedSSO more configurable so we can simulate failures as well. If @ThoWagen can't easily reproduce the situation you're describing, I can try to find some time to improve SimulatedSSO to make it easier to test!

EreMaijala:
@demiankatz Right, of course! I tried SimulatedSSO and initially couldn't make it work. Then I realized, that it doesn't work with ChoiceAuth. Maybe it's because it's not including auth_method in the backlink?

demiankatz:
> @demiankatz Right, of course! I tried SimulatedSSO and initially couldn't make it work. Then I realized, that it doesn't work with ChoiceAuth. Maybe it's because it's not including auth_method in the backlink?

Very likely! I threw this thing together in a few minutes to test this PR, so it's not particularly well-developed. Feel free to open a PR to add the missing parameter, or let me know and I can put it on my to-do list when time permits.

EreMaijala:
And armed with this knowledge I tried Shibboleth without ChoiceAuth, but with a failed login it just ends up in MyResearch/Home?auth_method=Shibboleth with the following message:
kuva

EreMaijala:
#3201 adds the auth_method parameter to SimulatedSSO. It allows it to work with ChoiceAuth, but for some reason it breaks out of lightbox whereas using SimulatedSSO alone works as expected.

EreMaijala:
Actually, it doesn't break out lightbox, but at least when trying to email records from cart, the email form is displayed as the main page instead of the page where cart was opened.

demiankatz:
> And armed with this knowledge I tried Shibboleth without ChoiceAuth, but with a failed login it just ends up in MyResearch/Home?auth_method=Shibboleth with the following message: kuva

How do you define "failed login"? I thought the general pattern with Shibboleth was that VuFind redirects to Shibboleth, which sends the user to an external login screen. Then they either get stuck at that screen because they can't log in, or they get redirected back to VuFind with successful credentials. What is the scenario where they get sent back to VuFind without valid credentials? Does this just indicate a problem with Shibboleth configuration, or is there a "normal" path that leads to this scenario?

EreMaijala:
> > And armed with this knowledge I tried Shibboleth without ChoiceAuth, but with a failed login it just ends up in MyResearch/Home?auth_method=Shibboleth with the following message: kuva
> 
> How do you define "failed login"? I thought the general pattern with Shibboleth was that VuFind redirects to Shibboleth, which sends the user to an external login screen. Then they either get stuck at that screen because they can't log in, or they get redirected back to VuFind with successful credentials. What is the scenario where they get sent back to VuFind without valid credentials? Does this just indicate a problem with Shibboleth configuration, or is there a "normal" path that leads to this scenario?

Certainly not a typical situation, but I suspect this would happen e.g. if the user doesn't have the required attributes (e.g. user_attribute_1/user_attribute_value_1 in config.ini) even if the login was successful.

demiankatz:
> Certainly not a typical situation, but I suspect this would happen e.g. if the user doesn't have the required attributes (e.g. user_attribute_1/user_attribute_value_1 in config.ini) even if the login was successful.

Makes sense. I suspect this has not been anticipated by the existing code and probably needs to be addressed somehow. But hopefully that problem doesn't need to hold up the rest of the work here. Should we open a JIRA ticket to track this?

EreMaijala:
I agree, doesn't need to be addressed here.

ThoWagen:
@EreMaijala @demiankatz I'm not sure if I understand your discussion completely. Is there still a problem with this PR or has that what you have discussed nothing to do with this PR directly?

"Actually, it doesn't break out lightbox, but at least when trying to email records from cart, the email form is displayed as the main page instead of the page where cart was opened."

This sound to me like it return to the stored followup but does not restore it inside of a lightbox. So there is still something not working. But maybe I misunderstand the scenario.

I can have a look into it next week.

demiankatz:
@ThoWagen, I've just opened #3208 to make it possible for the SimulatedSSO authentication module to be configured to throw exceptions instead of logging in. This makes it possible to easily reproduce the scenario that @EreMaijala describes. I am seeing the same problem: when login fails, the lightbox is bypassed.

themes/bootstrap3/js/lightbox.js
ThoWagen: fixed

themes/bootstrap3/js/lightbox.js
ThoWagen: done

ThoWagen:
I tested this with SimmulatedSSO and ChoiceAuth but it works as expected for me. 

I also tested with the failing SimulatedSSO and can confirm that you will get to this page:

image

But this is expected since we are not returning to the followup yet.

 If I do this and then change the username back to "fakeuser1" in the configs and try logging in again the lightbox tries to open the record page again. The followupUrl is overwritten if the login failed (MyResearchController line 176). This can be fixed by clearing the lightboxParent in `setFollowupUrlToReferer`. But is this the expected behavior? Shouldn't we also return to the original followup if one login fails?

ThoWagen:
Clearing the lightboxParent in `setFollowupUrlToReferer` now since this should be done anyway.

module/VuFind/src/VuFind/Controller/AbstractBase.php
```diff
@@ -726,6 +726,9 @@ protected function setFollowupUrlToReferer(bool $allowCurrentUrl = true)
             return;
         }
 
+        // Clear previously stored lightboxParent.
+        $this->followup()->retrieveAndClear('lightboxParent');
```
demiankatz: You could just use `->clear` here since you're not using the retrieved value.

demiankatz (changes requested):
Thanks, @ThoWagen! Just one small suggested related to your latest code.

You make a good point about the reason that failed logins do not open in the lightbox. That makes sense! I suppose it might feel slightly more natural if the error message appeared in the lightbox instead of on a full page, but I imagine that making that happen would require significant work, and the benefit is minimal (especially because this type of failure scenario should be extremely rare in a correctly configured instance). I don't think this is a blocker unless @EreMaijala does.

module/VuFind/src/VuFind/Controller/AbstractBase.php
ThoWagen: fixed

ThoWagen:
By adding 

```
if(
    $this->followup()->retrieve('lightboxParent')
    && $url = $this->getAndClearFollowupUrl(true)
) {
    return $this->redirect()->toUrl($url);
}
```
inside of the 
```
// Not logged in?  Force user to log in:
        if (!$this->getAuthManager()->isLoggedIn()) {
```
block we could return to the action inside the lightbox and get the login screen again. 

But the error flashmessage will shown inside the parent view and not the lightbox. I think this could be fixed by adding a query parameter like "block_stored_messages=true" to the lightboxParent and using this to determine if stored flashmessages should be displayed. What do you think about this? @demiankatz @EreMaijala

demiankatz:
> But the error flashmessage will shown inside the parent view and not the lightbox. I think this could be fixed by adding a query parameter like "block_stored_messages=true" to the lightboxParent and using this to determine if stored flashmessages should be displayed. What do you think about this? @demiankatz @EreMaijala

Do we need a separate parameter at all? If we have pending flash messages and plan to open a lightbox, don't we always want to display the flashmessages in the lightbox? Can we leverage our knowledge of the lightbox state to make the right decision? (I'm open to adding a parameter if it is necessary -- I'm just checking first whether there's a simpler way).

ThoWagen:
@demiankatz You are right. Inside of the Flashmessages ViewHelper we can check if lightboxChild is set. I added it.

module/VuFind/src/VuFind/Controller/AbstractBase.php
```diff
@@ -740,16 +747,44 @@ protected function normalizeUrlForComparison($url)
...
+     * @param bool $checkRedirect Whether the query should be checked for param 'redirect'
+     *
      * @return string
      */
-    protected function getFollowupUrl()
-    {
-        return $this->followup()->retrieve('url', '');
+    protected function getAndClearFollowupUrl($checkRedirect = false)
+    {
+        if ($url = $this->followup()->retrieveAndClear('url', '')) {
```
demiankatz: Is there a reason to default to `''` here instead of omitting the second parameter to use the default `null`? I would expect both to have the same falsy behavior for the purposes of this if statement. Doesn't make a huge difference, but having the second parameter made me question why it was there, whereas relying on the default imposes less thought on the potential reader of the code. :-)

demiankatz (changes requested):
Thanks, @ThoWagen -- all tests are passing, and the failed login behavior is improved. I think this may be ready to merge, but I've asked @EreMaijala to take one more look at it, and I have one fairly trivial comment below that you can act on or ignore as you see fit.

EreMaijala (approved):
Looks great to me!

ThoWagen:
While testing this myself once more I encountered a problem.

It is possible to create race conditions in the session. This is not a problem of this PR and can be reproduced on the dev branch. 

If you enable Ajax requests like `loadInitialTabWithAjax` or `show_result_counts`, go to a record and click the `Email this` button really fast it can happen that you don't return to the email action after logging in but load the favorites. The reason for this is that if the Ajax request and the email action are performed simultaneously and the Ajax request finishes after the email action the session will be cleared again and the followup URL is deleted. This can be reproduced consistently if you set a debugger breakpoint at the start of the `writeClose` method in the `SessionManager` class of Laminas. Then you can control which request writes the session first.

@demiankatz @EreMaijala Can you reproduce this?

demiankatz:
@ThoWagen, I'm out of the office today, so I won't have time to try to reproduce this before tomorrow. However, most of our AJAX actions are supposed to disable the session in order to prevent this type of problem, but I notice that \VuFind\Controller\AbstractRecord::ajaxtabAction() does not appear to be doing that. I'm curious if calling [disableWrite](https://github.com/vufind-org/vufind/blob/33978d5dec6db9781893144ae6a5e9dc60d45e1c/module/VuFind/src/VuFind/Session/Settings.php#L72) in the AJAX tab method would solve your problem. Of course, that would also prevent tabs from writing anything to the session... but at least off the top of my head, I can't think of a reason why any of them would need to.

module/VuFind/src/VuFind/Controller/AbstractBase.php
demiankatz: On a separate topic from the race condition discussed below, please let me know your thoughts on this item -- once it's resolved, I think we can merge this and move the rest of the discussion to a JIRA ticket or new PR.

demiankatz:
On reflection, there might be some tab handlers that need to write to the session; I'm thinking in particular of the one that embeds search results in collections, since that might need to store data for the next/previous controls on the record view. Not sure if that's a use case that can actually come up... In any case, it's rare, so maybe we can make session support "opt-in" to reduce chances of trouble. Regardless, I think testing as I suggested above to see if that solves the problem is a good place to start, and if it helps, we can refine it as needed.

module/VuFind/src/VuFind/Controller/AbstractBase.php
ThoWagen: No, there is no reason for this. I removed it.

ThoWagen:
@demiankatz Thank you! Your solution indeed fixes the problem. I created a PR #3241 for this.

demiankatz (approved):
Looking good and all tests are passing. Thanks for the excellent work on this, @ThoWagen!

PR #2943, opened 2023-06-19
Fix return types in Aleph driver

xmorave2:


module/VuFind/src/VuFind/ILS/Driver/Aleph.php
```diff
@@ -1056,7 +1056,7 @@ public function getMyHolds($user)
      * @param array $holdDetails A single hold array from getMyHolds
      * @param array $patron      Patron information from patronLogin
      *
-     * @return string Data for use in a form field
+     * @return string|null Data for use in a form field
```
demiankatz: The [driver spec](https://vufind.org/wiki/development:plugins:ils_drivers#getcancelholddetails) suggests that this should be an empty string instead of null. Would it be better to align with the spec instead of changing the type? Or is the spec inaccurate and in need of updating?

demiankatz (changes requested):
Thanks, @xmorave2 -- just one comment (see below).

module/VuFind/src/VuFind/ILS/Driver/Aleph.php
xmorave2: You are right, it should return empty string instead of null. Fixed

demiankatz (approved):
Thanks, @xmorave2! Since this is a fix, I've backported it to the release-9.0 branch to include it in 9.0.2.

PR #2944, opened 2023-06-19
Add EBSCO Publication Finder data source

maccabeelevine:
Per [EBSCO Connect](https://connect.ebsco.com/s/article/Publication-Finder-API-Reference-Guide-About-Publication-Finder): 

> "The Publication Finder API provides customers the ability to search for publications designated as part of their collection in the Holdings Management module in EBSCOadmin. For more information about Full Text Finder and Publication Finder, please visit [Full Text Finder](https://connect.ebsco.com/s/topic/0TO1H000000HHfWWAW?language=en_US)."

It is closely related to the EDS API, but distinct:
- The API is documented [alongside the EDS API](https://connect.ebsco.com/s/article/EBSCO-Discovery-Service-API-Reference-Guide).
- It shares the same authentication and session endpoints.
- The specific guides to each functional component (search, retrieving a specific item, etc.) are all distinct from the EDS API.
- In practice the queries for the search and retrieve methods are a strict subset of the EDS equivalents.
- The response data is formatted similarly wherever the same type of data is returned, but of course there are differences between publication-level data (EPF) and article-level data (EDS).

TODO:
- [ ] When merging, add changelog note about altered createBackendFilterParameters method signature, added session URL setting, and removal of edsApiHost property.


maccabeelevine:
@demiankatz What I've committed so far is a **very** minimal implementation; the search results page and record page work for some basic definition of 'work', and you can adjust (in the EPF.ini) the default_view level of detail.

I'd like to have this initial PR be a sort of MVP version of the integration, and go from there in future PRs as needed.  There is a ton of functionality in the existing EDS integration, and I don't know how much of it will be applicable or useful for EPF.  Let me know if that's reasonable?  Beyond what works so far, I think an MVP would include:
- Facets, to some extent.  I definitely want to be able to filter journals-only (by default, it includes books and other content -- not clear why).
- Cover images
- Full text links, to some extent.

Of course I would also include docstrings, tests, linting fixes etc. once scope is clear.

I don't necessarily think an MVP would need to include this stuff (that I see in the EBSCO UI for Lehigh's implementation):
- Page Options via UI (i.e. view level of detail, results per page)
- Sorting options
- Browse by first letter of title

There's also a whole mechanism in the EDS API for retrieving institution-configured options from the API itself.  I think that's a nice-to-have?

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Backend.php
```diff
@@ -295,30 +295,47 @@ public function retrieve($id, ParamBag $params = null)
                 $this->profile = $overrideProfile;
             }
             $sessionToken = $this->getSessionToken();
-            $parts = explode(',', $id, 2);
-            if (!isset($parts[1])) {
-                throw new BackendException(
-                    'Retrieval id is not in the correct format.'
+
+            $backendType = (null !== $params) ? $params->get('backendType')[0] : null;
```
maccabeelevine: @demiankatz I do not like this $backendType, but did it anyway because it's tentatively the only significant change to the EDS Backend, and if that stays the same after all the MVP is done, it may be a better option than a second backend.  But I'm not convinced yet.  The two APIs share so much in common, despite being defined separately; it's very confusing.

module/VuFind/src/VuFind/Controller/EPFController.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace VuFind\Controller;
```
demiankatz, line 31: TODO: add required header comments.

config/vufind/EPF.ini
```diff
@@ -0,0 +1,22 @@
...
+; This section represents the EBSCO EDS API Account credentials.
+; If using IP Authentication, then the user_id and password should remain blank
+; and ip_auth should be set to true.
+[EBSCO_Account]
+; IP authentication for the API
+ip_auth = false
+user_name = [USERNAME]
+password  = [PASSWORD]
+profile   = [PROFILE]
+organization_id = "VuFind 2.x from MyUniversity"
```
demiankatz: Any idea why there's a "2.x" in there? If that's copied over from the EDS configuration, we should probably just delete the number there in a separate PR to prevent future confusion.

config/vufind/EPF.ini
```diff
@@ -0,0 +1,22 @@
...
+
+; This is the URL of the EPF API endpoint:
+api_url = "https://eds-api.ebscohost.com/edsapi/publication"
+
+; This is the URL of the EBSCO authorization endpoint:
+auth_url = "https://eds-api.ebscohost.com/authservice/rest"
+
+session_url = "https://eds-api.ebscohost.com/edsapi/rest"
+
+; This section represents the EBSCO EDS API Account credentials.
```
demiankatz: Should this say EPF rather than EDS? And/or do we need a note indicating that the credentials are essentially the same between the two services?

module/VuFind/src/VuFind/Controller/EPFrecordController.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace VuFind\Controller;
```
demiankatz, line 31: TODO: add header comments

module/VuFind/src/VuFind/Controller/EdsrecordController.php
```diff
@@ -125,4 +125,12 @@ protected function resultScrollerActive()
         return isset($config->Record->next_prev_navigation)
             && $config->Record->next_prev_navigation;
     }
+
+    protected function loadRecord(ParamBag $params = null, bool $force = false)
```
demiankatz: TODO: add comment

module/VuFind/src/VuFind/RecordDriver/EPF.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace VuFind\RecordDriver;
```
demiankatz, line 31: TODO: add comments

module/VuFind/src/VuFind/Search/EPF/Options.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace VuFind\Search\EPF;
```
demiankatz, line 33: TODO: add comments

module/VuFind/src/VuFind/Search/EPF/Params.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace VuFind\Search\EPF;
```
demiankatz, line 34: TODO: add comments

module/VuFind/src/VuFind/Search/EPF/Params.php
```diff
@@ -0,0 +1,33 @@
...
+        return $backendParams;
+    }
+
+    public function getView()
+    {
+        $viewArr = explode('|', $this->view ?? '');
+        return $viewArr[0];
+    }
+
+    public function getEpfView() {
```
demiankatz: Do you need this wrapper method? Seems to only be used internally, so maybe it's better to just go through getOptions() as needed.

module/VuFind/src/VuFind/Search/EPF/Results.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace VuFind\Search\EPF;
```
demiankatz, line 34: TODO: add comments

module/VuFind/src/VuFind/Search/Factory/EPFBackendFactory.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace VuFind\Search\Factory;
```
demiankatz, line 31: TODO: add comments

module/VuFind/src/VuFind/Search/Factory/EdsBackendFactory.php
```diff
@@ -69,6 +69,11 @@ class EdsBackendFactory extends AbstractBackendFactory
      */
     protected $accountData;
 
+    protected function getServiceName()
```
demiankatz, line 85: TODO: add comment

The comment should probably include the fact that this name is used for both configuration and record driver retrieval... unless we're better off having separate properties for configName/driverName, which might be worth considering (though I wouldn't insist on it).

module/VuFind/src/VuFind/Search/Factory/EdsBackendFactory.php
```diff
@@ -164,6 +159,21 @@ protected function createConnector()
         return $connector;
     }
 
+    protected function createConnectorOptions()
```
demiankatz: TODO: add comment

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Backend.php
demiankatz: Rather than passing this in through the ParamBag, would it make more sense to pass it in through the constructor, and then address it only at the factory level? I assume it's going to be constant for each backend, and we'll have a different instance of the class for each backend, so I don't think there's a reason to pass the value dynamically. And, of course, if it eventually turns out to make more sense to have separate subclasses for each service, you still only have to change the factory code to make the adjustment, instead of potentially having to eliminate some extra ParamBag-based communication.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
```diff
@@ -196,7 +200,7 @@ public function createSession(
      *
      * @return array    The requested record
      */
-    public function retrieve(
+    public function retrieveEdsItem(
```
demiankatz, line 244: There are tests that rely on the existence of a retrieve method. We either need to fix the tests or modify our approach to differentiated retrieval. I think for backward compatibility reasons, it might be preferable to keep retrieve in some form, even if only as a deprecated wrapper.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
```diff
@@ -216,6 +220,20 @@ public function retrieve(
         return $this->call($url, $headers, $qs);
     }
 
+    public function retrieveEpfItem(
```
demiankatz, line 274: TODO: add comment (if retaining).

themes/bootstrap3/templates/epf/home.phtml
```diff
@@ -0,0 +1,2 @@
+<?php
```
demiankatz: This template could be slightly simplified to:

`<?=$this->render('search/home.phtml');`

themes/bootstrap3/templates/epf/search.phtml
```diff
@@ -0,0 +1,2 @@
+<?php
+  echo $this->render('search/results.phtml');
```
demiankatz, line 3: Can be simplified as above.

demiankatz (changes requested):
Thanks, @maccabeelevine, this is a great start! I've left some suggestions (mostly reminders about missing comments, which I'm sure you're aware of -- but hopefully they're a helpful checklist rather than annoying nags).

I haven't done any hands-on testing of this yet, but I'm happy to do so whenever you're ready for somebody else to try it. I just wanted to do the basic code review/discussion first!

Regarding pagination options and sorting, I agree that these are probably not essential for an MVP, but I also suspect that they're probably pretty easy to implement (though maybe I'm mistaken, since the EDS options are a bit weird, and maybe EPF has the same complications).

In any case, I'm definitely open to the strategy of merging an MVP first and then adding additional functionality incrementally. That should make the code review process easier to manage, among other advantages.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Backend.php
maccabeelevine: Setting the backendType via the backend factory instead.  Could be done in the constructor but I used a mutator instead as that seemed to be the pattern in that file.

config/vufind/EPF.ini
maccabeelevine: So I missed editing that comment by mistake, thanks for catching, but actually / stupidly it's correct as-is.  The Publication Finder API page makes clear that auth actually uses the EDS API.  That's why the `session_url` correctly has the "edsapi" path.

>The Pub Finder API requires that all calls have both session and authentication tokens which can be obtained by making calls the [EDS API Authentication](https://connect.ebsco.com/s/article/EBSCO-Discovery-Service-API-IP-Authentication) and [Create Session](https://connect.ebsco.com/s/article/EBSCO-Discovery-Service-API-Sessions-CreateSession) methods. They are to be specified in the HTTP header in the same way they are for the methods in the EDS API.

I did update the comment to hopefully make that clearer, although I'm not sure I was successful.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
maccabeelevine: Ok, I added it back as a pass-through and with a deprecation warning.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Backend.php
```diff
@@ -137,6 +137,8 @@ class Backend extends AbstractBackend
      */
     protected $isGuest;
 
+    protected $backendType = null;
```
demiankatz, line 145: TODO: add comment to property and associated setter method.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Backend.php
demiankatz: Works for me -- thanks!

config/vufind/EPF.ini
demiankatz: I think this is pretty clear, though I notice the comment mentions a user_id setting that doesn't exist (it's user_name). This error is copied and pasted from the EDS.ini file, which should probably be updated in a separate PR (perhaps along with the weird "VuFind 2.x" example that I mentioned elsewhere).

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
demiankatz: Sounds good. If we're considering retrieve to be deprecated, we should also update the test that calls it (though that's obviously not the most urgent task at hand here :-) ).

demiankatz:
@maccabeelevine, I'm seeing a test failure:

```
1) VuFindTest\Backend\EDS\BackendTest::testRetrieve
Undefined variable $backendType

/home/runner/work/vufind/vufind/module/VuFindSearch/src/VuFindSearch/Backend/EDS/Backend.php:337
/home/runner/work/vufind/vufind/module/VuFindSearch/tests/unit-tests/src/VuFindTest/Backend/EDS/BackendTest.php:116
```

Seems likely to be related to recent refactoring. :-)

maccabeelevine:
Sorry about ignoring the documentation comments and test failures...I'm trying to get a bit more functionality in place first, as right now it's still pretty heavy refactoring mode and I keep removing things I add.

demiankatz:
> Sorry about ignoring the documentation comments and test failures...I'm trying to get a bit more functionality in place first, as right now it's still pretty heavy refactoring mode and I keep removing things I add.

No problem! I just highlighted that latest test failure because it looks like it might be reveal an issue in the new code.

config/vufind/EPF.ini
maccabeelevine: Yeah no idea -- this was copied from EDS.ini

module/VuFind/src/VuFind/Search/EPF/Params.php
maccabeelevine: I prefer to leave it.  EDS/Params.php has a slightly more complicated version of this method, which I simplified because the MVP is skipping the more advanced options/params logic from EDS.  But it's reasonable to think that may come later.  So I prefer to keep it parallel.

module/VuFind/src/VuFind/Controller/EPFController.php
maccabeelevine: Added

module/VuFind/src/VuFind/Controller/EPFrecordController.php
maccabeelevine: Added

module/VuFind/src/VuFind/Controller/EdsrecordController.php
maccabeelevine: Added

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
maccabeelevine: I updated the unit test as well, since otherwise it was failing (since now I have to mock retrieveEdsItem, not retrieve).

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
maccabeelevine: Added

themes/bootstrap3/templates/epf/home.phtml
maccabeelevine: Done.

themes/bootstrap3/templates/epf/search.phtml
maccabeelevine: This is now a two-line file, so I think the <?= simplification is no longer useful.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Backend.php
maccabeelevine: Added

maccabeelevine:
Ok I've finished everything described above for the MVP, except writing new unit tests, which I'd rather do after we're good with functionality (sorry TDD).  Please do test as needed.

>Regarding pagination options and sorting, I agree that these are probably not essential for an MVP, but I also suspect that they're probably pretty easy to implement (though maybe I'm mistaken, since the EDS options are a bit weird, and maybe EPF has the same complications).

Yeah there are two main complications.  a) the tight coupling between EDS and EPF, that may not be the same for sorting (not sure yet).  And b) all of the options should eventually be changed to pull from the EPF GetInfo() API call, as EDS does.  So it seemed premature to try to get those options right.

Although I did want to raise the broader related issue.  All of the API calls return responses that are configured by the EBSCO backend UI, i.e. which facets to allow, among (I think) many other things.  The GetInfo() call is meant to retrieve those backend settings in order to configure the UI, along with the EDS.ini settings (there's some merging that goes on).  So for example in EPF.ini > [Facets], I added the three facets I can see in actual API responses, but there may be others that Lehigh has disabled.  I have to figure out how to see all the possible options without changing our production setup.  (I did reach out to EBSCO for a contact on this and some related questions.)  But I figured for the MVP it's ok to ignore all that and just use what is set locally in EPF.ini.

themes/bootstrap3/templates/epf/search.phtml
demiankatz: Agreed.

module/VuFind/src/VuFind/Controller/EPFController.php
```diff
@@ -0,0 +1,67 @@
...
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Site
+ */
+
+namespace VuFind\Controller;
+
+use Laminas\ServiceManager\ServiceLocatorInterface;
+
+/**
+ * EDS Controller
```
demiankatz: Copy and paste error -- should be EPF.

module/VuFind/src/VuFind/Controller/EPFrecordController.php
```diff
@@ -0,0 +1,57 @@
...
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Site
+ */
+
+namespace VuFind\Controller;
+
+use Laminas\ServiceManager\ServiceLocatorInterface;
+
+/**
+ * EDS Record Controller
```
demiankatz: Copy and paste error.

config/vufind/EPF.ini
maccabeelevine: Yup I can address those items in a PR following this one.

demiankatz: If it would help, I can do some investigation and open a PR to address the existing EDS issues, and then we can just adjust this one to match.

module/VuFind/src/VuFind/Search/EDS/AbstractEDSParams.php
```diff
@@ -0,0 +1,93 @@
...
+class AbstractEDSParams extends \VuFind\Search\Base\Params
+{
+    /**
+     * Set up filters based on VuFind settings.
+     *
+     * @param ParamBag $params Parameter collection to update
+     *
+     * @return void
+     */
+    public function createBackendFilterParameters(ParamBag $params)
```
demiankatz, line 58: I notice that you've removed the second Options parameter here... which makes sense, since it appears to be unused, and the Options object could easily be obtained through $this->getOptions() from inside the method if needed anyway. However, this is a minor BC break, so there's a question of whether it's worth retaining the parameter in case somebody relies on it in a local subclass, or if we should just drop it and put a note in the changelog. I think the latter is probably fine, but I'm wracking my brain to figure out why this was here in the first place, in case there's a valid use case I'm somehow forgetting about.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
```diff
@@ -49,7 +49,9 @@ abstract class Base
     protected $debug = false;
 
     /**
-     * EDSAPI host
+     * EDSAPI host.  Although the default is the same as $sessionHost,
```
demiankatz: This comment says the two values are identical, but the values below are not identical. What is intended/expected?

demiankatz (changes requested):
Thanks, @maccabeelevine -- just a few more questions and comments based on the latest changes.

I'll try to find time for some hands-on testing tomorrow if I can (though if preparation for Monday's 6.0.2 release takes longer than planned, it might have to wait until next week).

module/VuFind/src/VuFind/Controller/EPFrecordController.php
maccabeelevine: Fixed

module/VuFind/src/VuFind/Controller/EPFController.php
maccabeelevine: Fixed

module/VuFind/src/VuFind/Search/EDS/AbstractEDSParams.php
maccabeelevine: > If it would help, I can do some investigation and open a PR to address the existing EDS issues, and then we can just adjust this one to match.

That would certainly help.  I'm constantly trying to make sense of the EDS API as part of this, so it helps if EDS.ini is as close to correct as possible.

> I notice that you've removed the second Options parameter here... which makes sense, since it appears to be unused, and the Options object could easily be obtained through $this->getOptions() from inside the method if needed anyway. However, this is a minor BC break, so there's a question of whether it's worth retaining the parameter in case somebody relies on it in a local subclass, or if we should just drop it and put a note in the changelog. I think the latter is probably fine, but I'm wracking my brain to figure out why this was here in the first place, in case there's a valid use case I'm somehow forgetting about.

Yeah, I removed the parameter because when I call it from the new EPF/Params.php, I didn't have any $options to pass in, and then I noticed that the method ignored the parameter in any case.  I wrote up the code to keep the parameter but set the default to null, can commit that if you like, if you prefer keeping it for BC.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
maccabeelevine: Sorry, I think the defaults were identical, but my comment made everything even more confusing, as did the $edsApiHost variable which was now used for both EDS and EPF endpoints.  I renamed it to $apiHost and removed the default value from Backend altogether, since there is no longer a reasonable default.  I realize that's a BC break if they don't have it set in the .ini files (which would be odd since it's there by default, but fine).  So I added the defaults to the two BackendFactory classes instead.

module/VuFind/src/VuFind/Search/EDS/AbstractEDSParams.php
demiankatz: Here's a PR to update EDS.ini: https://github.com/vufind-org/vufind/pull/2950 -- I double-checked the code to be sure that these changes are accurate/safe. I'm honestly not sure what organization_id is used for (in our local configuration, the value is set to a blank string), but it does get passed along to EDS in the code, and the value in EDS.ini is obviously a fake example, so changing it seems entirely harmless.

Regarding the second parameter to createBackendFilterParameters, I'll just add a TODO note to indicate the possible minor BC break. I can't think of a good reason to keep the parameter since the options object is accessible by other means anyway. Maybe there's some bizarre edge case where you would want to pass a _different_ Options object instead of the one already inside the Params object, but I can't think of a good reason for that -- and there would be other ways to achieve the same effect anyway.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
demiankatz: I can add a changelog note about this as well, just to be on the safe side.

config/vufind/EPF.ini
demiankatz: As noted elsewhere, it should be safe to adjust this to match #2950 now.

module/VuFind/src/VuFind/Search/EDS/AbstractEDSParams.php
maccabeelevine: > Here's a PR to update EDS.ini: https://github.com/vufind-org/vufind/pull/2950

I just made the same updates to EPF.ini.

> Regarding the second parameter to createBackendFilterParameters, I'll just add a TODO note to indicate the possible minor BC break.

Ok that make sense.  I agree it's better than leaving that param in.

config/vufind/EPF.ini
maccabeelevine: Done

config/vufind/EPF.ini
```diff
@@ -0,0 +1,106 @@
+; For the EBSCO Publication Finder data source
+[General]
+
```
demiankatz: Unnecessary whitespace?

config/vufind/EPF.ini
```diff
@@ -0,0 +1,106 @@
...
+default_noresults_recommend[] = SwitchType
+default_noresults_recommend[] = RemoveFilters
+
+; This is the URL of the EPF API endpoint:
+api_url = "https://eds-api.ebscohost.com/edsapi/publication"
+
+; This is the URL of the EBSCO authorization endpoint:
+auth_url = "https://eds-api.ebscohost.com/authservice/rest"
+
+session_url = "https://eds-api.ebscohost.com/edsapi/rest"
```
demiankatz, line 27: Missing comment?

module/VuFind/src/VuFind/Controller/EPFController.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+/**
+ * EPF Controller
+ *
+ * PHP version 8
+ *
+ * Copyright (C) Villanova University 2010.
```
demiankatz: You might want to check dates in your new files and use 2023 where appropriate.

module/VuFind/src/VuFind/Search/Factory/EPFBackendFactory.php
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Create the options array for the EDS connector.
+     *
+     * @return array
+     */
+    protected function createConnectorOptions()
+    {
+        $options = parent::createConnectorOptions();
+        $options['api_url'] = $this->edsConfig->General->api_url
+            ?? 'https://eds-api.ebscohost.com/edsapi/publication';
```
demiankatz: Maybe you could use a $defaultApiUrl property in the factory to allow this customization, instead of having to override api_url here.

module/VuFind/src/VuFind/Search/Factory/EPFBackendFactory.php
```diff
@@ -0,0 +1,73 @@
...
+
+        if (isset($this->edsConfig->General->session_url)) {
+            $options['session_url'] = $this->edsConfig->General->session_url;
```
demiankatz: Any reason not to add this option to the base factory and EDS.ini as well? If we're making it configurable, I don't see harm in having the same configuration work in both .ini files. Combined with the above suggestion, that might eliminate the need for this whole custom method.

demiankatz (changes requested):
Thanks again, @maccabeelevine. I just tested this using my organization's credentials, and it seems to be working well. Let me know if you want me to test any specific features/details. I've given the code one last review and suggested a few more minor changes. I think once that's all done, I'd be comfortable to merge this as our MVP, and you can build on it in followup PRs as time permits.

maccabeelevine:
> Thanks` again, @maccabeelevine. I just tested this using my organization's credentials, and it seems to be working well. Let me know if you want me to test any specific features/details. I've given the code one last review and suggested a few more minor changes. I think once that's all done, I'd be comfortable to merge this as our MVP, and you can build on it in followup PRs as time permits.

I'd suggest trying the default_view parameter in EDF.ini (to 'title' or 'detailed') and check the changes -- I didn't customize any of that, it's just picking up logic from the EDS driver (or somewhere).  I think the facet parameters in EDF.ini also work, but let me know if you see anything odd?

config/vufind/EPF.ini
maccabeelevine: Fixed

config/vufind/EPF.ini
maccabeelevine: Fixed.

module/VuFind/src/VuFind/Controller/EPFController.php
maccabeelevine: Crap, I did that in a few places but forgot (many) others.  Fixed.

demiankatz:
> I'd suggest trying the default_view parameter in EDF.ini (to 'title' or 'detailed') and check the changes -- I didn't customize any of that, it's just picking up logic from the EDS driver (or somewhere). I think the facet parameters in EDF.ini also work, but let me know if you see anything odd?

@maccabeelevine, I did see differences when changing default_view between brief, title and detailed, so that seems good.

I tried the facet_limit setting in the [Facet_Settings] section of EPF.ini and it did not appear to have any effect (setting it to a low number did not reduce the size of facet lists). Everything else facet-related seems fine. The HideFacetValue/ShowFacetValue settings _did_ work for me, as did changing parameters in [Results_Settings] and moving things between [Facets] and [FacetsTop] and reordering them there.

module/VuFind/src/VuFind/Search/Factory/EPFBackendFactory.php
maccabeelevine: Yes, wish I had thought of this!  Done.

module/VuFind/src/VuFind/Search/Factory/EPFBackendFactory.php
maccabeelevine: Yes, done.  Originally I had only added sessionURL on the EPF side; on the EDS side I reused the API URL value for session since it was identical.  That worked but I'd rather just be specific and cleaner about it.  So I also removed some logic I had added in createSession() that was falling back to the apiHost if sessionHost was not present -- not needed since we now have a default.

maccabeelevine:
> I tried the facet_limit setting in the [Facet_Settings] section of EPF.ini and it did not appear to have any effect (setting it to a low number did not reduce the size of facet lists).

Yes I see what you mean.  The same thing happens when I change facet_limit in EDS.ini; no effect. So not sure what to do there.

module/VuFind/src/VuFind/Search/Factory/EPFBackendFactory.php
maccabeelevine: Well, to clarify, I left the createConnectorOptions() method in EdsBackendFactory, but no more need to override. in the EPF version.  I think that's ok because I can see possibly needing to override it in the future, and it breaks out a somewhat complex part of the connector setup.  But can move it back into createConnector() if you prefer.

demiankatz: I have no objection with leaving this refactoring in place -- as you say, it has potential benefits even if we don't need to leverage it for EPF.

demiankatz:
> Yes I see what you mean. The same thing happens when I change facet_limit in EDS.ini; no effect. So not sure what to do there.

Hmm, if the problem is also impacting EDS, I wonder if either something has changed, or this never worked to begin with. I'm out of time for this week, but maybe it's worth opening a JIRA ticket as a reminder to investigate further when time permits. If there is no solution, we should remove the .ini setting to prevent confusion. But maybe it's a bug that can be fixed by changing some aspect of the API call.

maccabeelevine:
> If there is no solution, we should remove the .ini setting to prevent confusion. But maybe it's a bug that can be fixed by changing some aspect of the API call.

For now I just removed it from EPF.ini.  I will (separately) look into fixing it on the EDS side before adding it back here, since EDS is integrated with server-side config (GetInfo()) and I don't want to do too much configuration optimization on the EPF side unti the same is true here.

module/VuFind/src/VuFind/Search/EDS/Options.php
```diff
@@ -7,6 +7,7 @@
  *
  * Copyright (C) EBSCO Industries 2013
  * Copyright (C) The National Library of Finland 2022
+ * Copyright (C) Villanova University 2023
```
demiankatz: Your only change to this file is adding a copyright line. I think we can probably revert that.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
```diff
@@ -176,13 +186,13 @@ public function createSession(
             . "$profile, guest: $isGuest, authToken: $authToken "
         );
         $qs = ['profile' => $profile, 'guest' => $isGuest];
-        $url = $this->edsApiHost . '/createsession';
+        $url = $this->sessionHost . '/createsession';
```
demiankatz, line 189: Would it be slightly more equivalent to the old functionality if we defaulted sessionHost to null and then used `($this->sessionHost ?? $this->edsApiHost)` here? Otherwise, if somebody has customized their API URL in the configuration but doesn't add the new session URL parameter, they'll revert to the default URL.

demiankatz (changes requested):
I think this is just about ready to merge -- just two (hopefully) final thoughts.

Also, thanks for looking into the facet problem on the EDS side. Let me know if you need any help from me on that one!

module/VuFind/src/VuFind/Search/EDS/Options.php
maccabeelevine: Oops, fixed.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
maccabeelevine: I'm not sure.  I guess it depends why they are customizing the URL in the first place, which is hard to fathom, but maybe different domains for international users?  But if that's the case, I think this *should* be a breaking change, needing to add the new sessionHost property.  

Defaulting sessionHost to apiHost would work for EDS queries but not for EPF queries.  Granted that would be backwards compatible, and EDS is literally 100% of the uses of this file until this PR is merged, but after that it creates a relationship that's wrong for half the code that uses the file, and obscures the problem.

This goes back to an odd EPF API design decision to literally use the EDS path for creating a session.  If they had just created an alias for an EPF url to do that, there would be no need for the new property.  But they did, meaning it's something that needs to be handled separately by API users.  So I think VF should as well.

demiankatz: That is a puzzling decision. In any case, I see your point -- I'll just add this to the changelog notes instead of insisting upon a change. :-)

demiankatz (approved):
I believe this is ready to merge now. Any last-minute changes/requests before I do so, @maccabeelevine?

maccabeelevine:
@demiankatz I think it is good to go

PR #2945, opened 2023-06-20
Avoid division by zero when result limit is 0.

EreMaijala:
This is a small follow-up to #2940 to avoid a potential division by zero. While I could only trigger the error by adding 0 to `limit_options`, I was a bit worried that there could be a situation such as locally customized code that could trigger it.


xmorave2 (approved):
Thanks @EreMaijala

demiankatz (approved):
Thanks, @EreMaijala, makes sense to me!

PR #2946, opened 2023-06-20
Replace strtolower with mb_strtolower to avoid encoding issues.

EreMaijala:
Oddly, I couldn't make the test fail, so the failure only occurs under Apache. It could be related to the locale setup somehow, but I couldn't quite find out how. Regardless, the actual change makes a difference and avoids preg_match failing at Query.php line 163 due to invalid UTF-8.


demiankatz (approved):
Looks good to me, @EreMaijala. Thanks for the fix!

PR #2947, opened 2023-06-20
Inheritance of local dirs for configuration files

ThoWagen:
For each local dir a localSystem.ini can be included. In the INI a parent dir can be configured. This dir will be added to the localConfigDirStack in the PathResolver. The defaultConfigSubdir can also be changed.

This allows inheritance of configurations that can be very useful when working with multiple instances of VuFind that have common configurations. 

In the future the localSystem.ini could be extended so that e.g. all configs automatically add the parent dirs configs without adding the [Parent_Config] section to the specific INI file. This would provide a simple way to have development configs overwriting the production configs.


local/localSystem.ini
```diff
@@ -0,0 +1,6 @@
+[Parent_Dir]
```
demiankatz: It would probably be a good idea to add comments to all of these settings.

I also wonder if we should name the file differently if we don't anticipate using it for purposes other than configuration management -- e.g. something like ConfigLocations.ini.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
```diff
@@ -60,6 +62,25 @@ class PathResolverFactory implements FactoryInterface
...
+    /**
+     * Constructor
+     */
+    public function __construct()
+    {
+        // Use ASCII 0 as a nest separator; otherwise some of the unusual key names
+        // we have (i.e. in WorldCat.ini search options) will get parsed in
+        // unexpected ways.
+        $this->iniReader = new IniReader();
+        $this->iniReader->setNestSeparator(chr(0));
```
demiankatz: I see that this is copied and pasted from VuFind\Config\PluginFactory. I wonder if there's a way we can centralize this logic so there's a single factory for the IniReader instances used to load VuFind configurations. It's entirely possible that's more work than it's worth, but I wanted to at least raise the idea.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
```diff
@@ -82,14 +103,35 @@ public function __invoke(
...
-            && strlen(trim(LOCAL_OVERRIDE_DIR)) > 0
-                ? [
-                    [
-                        'directory' => LOCAL_OVERRIDE_DIR,
-                        'defaultConfigSubdir' => $this->defaultLocalConfigSubdir,
-                    ],
-                ] : [];
+        $localDirs = [];
+        $currentDir = defined('LOCAL_OVERRIDE_DIR')
+            && strlen(trim(LOCAL_OVERRIDE_DIR)) > 0 ?
```
demiankatz: In our style, we prefer operators on the start of lines rather than the ends, so I'd recommend moving the `?` down to the start of the next line.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
```diff
@@ -82,14 +103,35 @@ public function __invoke(
...
+            LOCAL_OVERRIDE_DIR : '';
+        while (!empty($currentDir)) {
+            $systemConfigFile = $currentDir . '/localSystem.ini';
+            $systemConfig = new Config(
+                file_exists($systemConfigFile) ?
```
demiankatz: I'd also move down this `?` and `:`.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
```diff
@@ -82,14 +103,35 @@ public function __invoke(
...
+                    $this->iniReader->fromFile($systemConfigFile) :
+                    []
+            );
+            array_unshift(
+                $localDirs,
+                [
+                    'directory' => $currentDir,
+                    'defaultConfigSubdir' =>
+                        $systemConfig['Local_Dir']['config_subdir'] ??
```
demiankatz: ...and I'd move down this `??`.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
```diff
@@ -82,14 +103,35 @@ public function __invoke(
...
+                        $this->defaultLocalConfigSubdir,
+                ]
+            );
+            $parentDir = '';
+            if (!empty($systemConfig['Parent_Dir']['absolute_path'])) {
+                $parentDir = $systemConfig['Parent_Dir']['absolute_path'];
+            }
+            if (empty($parentDir) && !empty($systemConfig['Parent_Dir']['relative_path'])) {
+                $parentDir = $currentDir . '/' . $systemConfig['Parent_Dir']['relative_path'];
```
demiankatz: Should we throw some kind of an exception if both absolute_path and relative_path are set? Or should we adjust the configuration to make such a conflict impossible -- e.g. by having a path setting and a path_type setting that specifies 'relative' or 'absolute'?

demiankatz (changes requested):
Thanks, @ThoWagen! This seems like a pretty straightforward and flexible solution to the problem. I still think an argument could be made for using module.config.php, but I don't think that's obviously a better solution than this one. I'd be interested in hearing some more opinions, though!

In the meantime, however, I've just left a bit of general feedback.

local/localSystem.ini
ThoWagen: Renamed into DirLocations.ini because it configures a parent local directory and could be used for more than just configs. 

Also added comments.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
ThoWagen: I created a trait for this.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
ThoWagen: fixed

module/VuFind/src/VuFind/Config/PathResolverFactory.php
ThoWagen: fixed

module/VuFind/src/VuFind/Config/PathResolverFactory.php
ThoWagen: fixed

module/VuFind/src/VuFind/Config/PathResolverFactory.php
ThoWagen: I used the Parent_Config logic as orientation, but I think you are right. It should not be possible contradicting configs.

I changed it into "path" and a boolean "is_relative_path" to define whether the path is relative or absolute.

ThoWagen:
@demiankatz I'm not sure about using module.config.php for this. As far as I can tell this would require to create own modules or overwrite VuFinds module.config.php. And one would have to create new modules for each new local directory or it would be way less flexible.

But maybe I don't understand what you want to do. Could you elaborate on how you would do this with the module.config.php?

module/VuFind/src/VuFind/Config/PathResolverFactory.php
demiankatz: Ahh, of course, great idea.

module/VuFind/src/VuFind/Config/IniReaderTrait.php
```diff
@@ -0,0 +1,68 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  Config
+ * @author   Thomas Wagener <wagener@hebis.uni-frankfurt.de>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Site
+ */
+
+namespace VuFind\Config;
```
demiankatz: We tend to put traits in a Feature sub-namespace -- e..g. VuFind\Config\Feature in this case. I see that we recently added an ExplodeSettingsTrait directly in the VuFind\Config namespace, though. If you prefer to leave this here, that's fine... but if you think separating the traits for cleaner directory organization would be worthwhile, feel free to establish the Feature subdirectory, and I'll open a separate PR to move the other trait in with it.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
```diff
@@ -82,14 +85,37 @@ public function __invoke(
...
+                    'defaultConfigSubdir' =>
+                        $systemConfig['Local_Dir']['config_subdir']
+                        ?? $this->defaultLocalConfigSubdir,
+                ]
+            );
+
+            $parentDir = '';
+            if (!empty($systemConfig['Parent_Dir']['path'])) {
+                $parentDir = $systemConfig['Parent_Dir']['path'];
+                if ($systemConfig['Parent_Dir']['is_relative_path']) {
```
demiankatz: This needs a default value to avoid an unset configuration triggering a notice. I assume `?? false` is what you intend. Maybe this whole block could be simplified to:

```
            // If there's a parent, set it as the current directory for the next loop iteration:
            if (!empty($systemConfig['Parent_Dir']['path'])) {
                $isRelative = $systemConfig['Parent_Dir']['is_relative_path'] ?? false;
                $parentDir = $systemConfig['Parent_Dir']['path'];
                $currentDir = $isRelative ? $currentDir . '/' . $parentDir : $parentDir;
            } else {
                $currentDir = '';
            }
```

module/VuFind/src/VuFind/Config/PathResolverFactory.php
```diff
@@ -82,14 +85,37 @@ public function __invoke(
...
+            $systemConfigFile = $currentDir . '/DirLocations.ini';
+            $systemConfig = new Config(
+                file_exists($systemConfigFile)
+                    ? $this->getIniReader()->fromFile($systemConfigFile)
+                    : []
+            );
+            array_unshift(
+                $localDirs,
+                [
+                    'directory' => $currentDir,
```
demiankatz, line 119: Should we check if $currentDir exists and is a directory before adding it to the stack? Maybe we should throw an exception if configuration points at something that doesn't exist or can't be used as a directory. Also, maybe it's worth throwing a `realpath` in somewhere to make the paths more readable if we do have messages about them.

demiankatz (changes requested):
@ThoWagen, this keeps getting better! Just a few more questions and suggestions.

Regarding module.config.php, I am increasingly convinced that your approach here is better than anything we could achieve there, but just for the sake of completing the conversation: my original thinking was that most custom VuFind instances have at least one custom module, and because of the way Laminas module configuration merging works, if we turned the directory stack into an array in module.config.php, then it would be possible for any module to add one or more paths to the search stack. You definitely wouldn't need one Laminas module per path -- a single module could specify the whole stack if desired. This has the advantage of allowing very flexible configuration without requiring the parsing of additional configuration files. However, it has what I increasingly believe is a much bigger disadvantage: it's difficult to control the order effectively, since the module merging is always going to merge arrays together in the same way, so it's hard to add new paths to the "middle" of the stack between the home directory and the local directory. We would probably have to use some kind of nested array format with priority values and then sort the end result, and that's less elegant, less understandable, and less efficient than what you're doing here with .ini files.

I'm still open to other input and ideas, but I think if we can reach some consensus on this approach, it could reasonably be included in release 9.1, since it doesn't break backward compatibility in any way. I've put the 9.1 milestone on this, and I'll make sure we discuss it at the next Community Call to solicit some additional reviews.

I'd be interested to hear from @EreMaijala about this as well, but he's on holiday until August, so if you're eager to get this done sooner rather than later, we may have to proceed without him. :-)

module/VuFind/src/VuFind/Config/IniReaderTrait.php
ThoWagen: I moved the trait into a new Feature subdir. I can also move the ExplodeSettingsTrait there unless you think it's better to do this in an other PR.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
ThoWagen: Yes, i think your suggestion look much cleaner. I have taken this.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
ThoWagen: Using realpath to make the paths more readable. It can also be used directly to check if the paths exists. Now, an \Exception is thrown if a path does not exist. If you know a more specific exception type, I would change this. But right now, I could not think of a fitting one.

Throwing an Exception at this point results in an white screen and it's not directly obvious where the error comes from. Maybe a warning will be sufficient, since a non existing path will not directly result in errors?

ThoWagen:
@demiankatz Thank you for the explanation. I understand your idea better now. But I'd prefer the current solution when I think about our setup. 

To avoid parsing multiple INI files each request, maybe one could cache the localConfigDirStack?

I can't participate in the next Community Call but I think you can discuss this without me.

And when it comes to the question when the PRs should be included, there is no time pressure from our side. We will update our version at release 10 at the earliest.

module/VuFind/src/VuFind/Config/IniReaderTrait.php
demiankatz: Thank you! I think it would be better to move ExplodeSettingsTrait in a separate PR just to avoid creating any confusion here. (Especially since that's a change that can be merged right away without need for much discussion, whereas this is likely going to remain open for a little while longer). If you have time to do it, please feel free. If you don't, let me know and I'll do it when time permits.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
demiankatz: Yes, I think a warning would be appropriate in this situation if the alternative is a difficult-to-troubleshoot "white screen of death."

demiankatz:
Thanks, @ThoWagen! I've replied to a couple of your comments above.

I'll let you know the outcome of the Community Call conversation, but I strongly suspect we'll all reach similar conclusions about this.

I agree that it may be possible to cache the localConfigDirStack, though it's probably best to wait and see if there's a performance need before bothering to spend time on increasing code complexity. I think odds are good that in most situations, it won't make a significant difference.

module/VuFind/src/VuFind/Config/IniReaderTrait.php
ThoWagen: I created a PR for this.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
ThoWagen: I changed the exception to a warning.

demiankatz:
Thanks again, @ThoWagen, this is looking good to me. I'll try to find some time for hands-on testing next week, and then I'll let you know the out come of Community Call discussion early next month.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
```diff
@@ -82,14 +86,46 @@ public function __invoke(
...
+                    'defaultConfigSubdir' =>
+                        $systemConfig['Local_Dir']['config_subdir']
+                        ?? $this->defaultLocalConfigSubdir,
+                ]
+            );
+
+            // If there's a parent, set it as the current directory for the next loop iteration:
+            if (!empty($systemConfig['Parent_Dir']['path'])) {
```
demiankatz, line 127: It might be a good idea to confirm that the parent directory is not already in the directory stack -- otherwise, it's possible to create an infinite loop with circular references that prevents the site from loading.

local/DirLocations.ini
```diff
@@ -0,0 +1,12 @@
+; A parent local directory can be specified. When resolving paths to configuration files
+; this dir will also be searched if the file does not exist in the current dir.
+[Parent_Dir]
+; Path to the parent directory
+;path =
+; Whether the path is relative or absolute (defaults to false)
```
demiankatz: I went ahead and added a note to the comment here to indicate the default value, since it's useful to know for reference.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
```diff
@@ -82,14 +86,46 @@ public function __invoke(
...
+                    : []
+            );
+
+            // adding directory to the stack
+            array_unshift(
+                $localDirs,
+                [
+                    'directory' => $currentDir,
+                    'defaultConfigSubdir' =>
+                        $systemConfig['Local_Dir']['config_subdir']
```
demiankatz, line 121: The config file has a [LocalConfig] section, but the code is looking for [Local_Dir]. These should be reconciled.

demiankatz (changes requested):
@ThoWagen, I did some hands-on testing today and noticed a couple of problems. I fixed one (a simple documentation omission) but would be interested in your thoughts on the others.

demiankatz:
One other thing to contemplate, @ThoWagen: we're currently set up so that .gitignore prevents local configurations from being committed, but this setting does not apply to the newly-added config file. Should we perhaps consider renaming `local/DirLocations.ini` to `local/DirLocations.ini.dist` and adding `local/DirLocations.ini` to the .gitignore, so that an example config is provided, but if it's locally applied, it doesn't get committed? That might be more consistent with the rest of our approach, but I'm open to other opinions.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
ThoWagen: Yes, good call! Implemented it. A warning is thrown if this happens.

local/DirLocations.ini
ThoWagen: Thank you. This definitely makes sense.

module/VuFind/src/VuFind/Config/PathResolverFactory.php
ThoWagen: I must have missed that when I changed the section's name. Thanks for pointing it out! Fixed it.

ThoWagen:
> One other thing to contemplate, @ThoWagen: we're currently set up so that .gitignore prevents local configurations from being committed, but this setting does not apply to the newly-added config file. Should we perhaps consider renaming `local/DirLocations.ini` to `local/DirLocations.ini.dist` and adding `local/DirLocations.ini` to the .gitignore, so that an example config is provided, but if it's locally applied, it doesn't get committed? That might be more consistent with the rest of our approach, but I'm open to other opinions.

That makes sense. I have implemented the suggestion.

demiankatz (approved):
Thanks again, @ThoWagen! I tested this again and it seems to work as expected now. I'll leave this PR open until after next week's Community Call in case any further changes are suggested in discussion there.

demiankatz:
It has been a week since our Community Call discussion, and no one has objected to this addition or made alternate suggestions, so I'm going ahead and merging it. Thanks again, @ThoWagen!

PR #2948, opened 2023-06-20
Update SCSS Bootstrap to match Less version.

demiankatz:
While working on #2922, I noticed that the Bootstrap version in the SCSS directories was slightly out of sync with the version in the Less directories. This PR brings things back into sync.


ckaz (approved):
Looks good and shouldn't affect the looks.

demiankatz:
Thanks, @ckaz!

PR #2949, opened 2023-06-20
autocomplete.js accessibility patch.

crhallberg:
Addresses [VUFIND-1616](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1616).

Using the notes from [this excellent example](https://codepen.io/ademcifcioglu/pen/xdOyXv), accessibility has been improved.

With the release of [Angolia's robust and popular autocomplete library](https://www.algolia.com/doc/ui-libraries/autocomplete/introduction/what-is-autocomplete/), this library may soon be archived.

TODO
- [x] Merge and release https://github.com/vufind-org/autocomplete.js/pull/14, and increment version number in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "watch:lessToSass": "grunt watch:lessToSass"
   },
   "dependencies": {
-    "autocomplete.js": "vufind-org/autocomplete.js#v2.1.6",
+    "autocomplete.js": "vufind-org/autocomplete.js#v2.1.7",
     "grunt": "^1.0.4",
     "grunt-contrib-less": "^2.0.0",
     "grunt-less-to-sass": "duvillierA/grunt-less-to-sass#0.0.11",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "watch:lessToSass": "grunt watch:lessToSass"
   },
   "dependencies": {
-    "autocomplete.js": "vufind-org/autocomplete.js#v2.1.5",
+    "autocomplete.js": "vufind-org/autocomplete.js#v2.1.6",
     "grunt": "^1.0.4",
     "grunt-contrib-less": "^2.0.0",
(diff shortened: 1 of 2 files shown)


demiankatz:
Thanks, @crhallberg, I'm running the test suite now and have re-based this onto the release-9.0 branch so it can be considered for inclusion in 9.0.2 next week. I'll report back once I have results (probably after lunch).

demiankatz (approved):
All tests are passing, and things seem to be working as expected in my manual tests, so from a visual perspective, I see no problems.

I don't have a screen reader setup with which to test the other behavior, however. I talked to @dltj, and he may be able to help with that testing later today. I'll hold off on merging until I hear his report, and if he's not able to get to this, I'll see if I can get a screen reader set up in my own environment tomorrow before the release packaging process begins.

demiankatz:
@crhallberg, this was not working for @dltj yesterday, so today I set up NVDA on my system and confirmed that it still wasn't quite right. I've adjusted the code here and also opened https://github.com/vufind-org/autocomplete.js/pull/14 with a fix that gets NVDA working as expected. I imagine it's probably not a best-practice solution, but it's all I could figure out that triggered the desired behavior. (Probably the only part that's strictly necessary is using aria-owns to point at the active element instead of the overall list... which is also unfortunately probably the most questionable design decision here... the other changes are designed to fix some bugs in the original listbox design, which was missing some key attributes and was assigning incorrect IDs; I figured it was worth getting those details right, even though it didn't seem to lead to appropriate NVDA behavior).

demiankatz (changes requested):
I'm changing my review from "approve" to "request changes" to be sure we don't accidentally merge this before a new version of autocomplete.js has been released. I'll approve and merge once the new release is minted (assuming nobody can find a better solution to this problem in the meantime).

demiankatz (approved):
I believe this is all good now. Thanks, @crhallberg!

PR #2950, opened 2023-06-22
Update outdated/incorrect EDS.ini language.

demiankatz:
This PR updates an inaccurate comment to use the correct config setting, and removes a version number from an example value to prevent potential confusion.


maccabeelevine:
Looks good to me.

crhallberg (approved):

PR #2951, opened 2023-06-22
Moving ExplodeSettingTrait to Feature Subdir

ThoWagen:
Traits are usually put in a Feature sub-namespace. Hence, the ExplodeSettingsTrait is moved there in this PR as discussed in PR https://github.com/vufind-org/vufind/pull/2947

### DIFF
--- a/module/VuFind/src/VuFind/Config/Feature/ExplodeSettingTrait.php
+++ b/module/VuFind/src/VuFind/Config/Feature/ExplodeSettingTrait.php
@@ -27,7 +27,7 @@
  * @link     https://vufind.org Main Site
  */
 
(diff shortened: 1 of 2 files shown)


ThoWagen:
I'm not sure why the test failed. It does not look like it has something to do with the changes of the PR.

demiankatz (approved):
Thanks, @ThoWagen, this looks good to me. I suspect that the CI failure is being caused by a GitHub Actions glitch and not any of the code changes, as you say. I've re-triggered the jobs to see if that sorts things out. If not, I'll keep trying. Once the tests pass, I'll merge this.

PR #2952, opened 2023-06-23
Adding optional virtual keyboard to searchbox

ThoWagen:
There are libraries or services that have records in many different languages or even specialize in a specific language. For those, it it nice to be able to search in the original language. However, the languages might use completely different character sets. Therefore, this PR includes the option to configure virtual keyboards for the searchbox.

 The library https://github.com/hodgef/simple-keyboard is used and all of the layouts https://github.com/simple-keyboard/simple-keyboard-layouts are supported. The layouts can be activated by uncommenting them in the searchbox.ini. The keyboards can then be selected via a dropdown menu in the searchbox. If no layout is activated in the configurations the dropdown menu is not shown. Hence, this is completely optional.


demiankatz:
Thanks, @ThoWagen, this looks good to me! Before merging, though, I'm hoping that @crhallberg can take a quick look at the styles and JS code to see if he has any suggestions, and that @sturkel can take a little time testing the functionality in all of the themes to see if she can find any usability or display issues. Once I hear back from them, I'll approve and merge unless there are problems revealed.

sturkel89:
### Overview

I checked the branch using all three themes. I found that almost everything displays and functions perfectly in Bootstrap3, but there are display issues in the other two themes. See below for screenshots.

Functionality is generally good in every theme:

- When you click the keyboard icon, you get the dropdown that allows you to select a language. 
- When you click in the search box when the keyboard icon is visible, you USUALLY get see the virtual keyboard pop out, and you can enter your search on either the virtual keyboard or your physical keyboard. 
- Each "key" on the keyboard that I clicked resulted in the same character displaying in the search box.
- If there were matching items in the catalog the search worked as expected. 
----------------------------------------------
### Bootstrap3
The keyboard icon fits very seamlessly into the search box in this theme.
image

The virtual keyboard opens nicely across the screen when you click in the search box, with good spacing between it and the search box.
image

The keyboard icon remains visible when you get to a Results screen or a No Results screen.
image

When you go to do a new search from a Results or No Results screen, the keyboard pops up and and displays as expected.
image

When you visit an item record, the keyboard icon remains clearly visible in the search box.
image

_HOWEVER: when you click in the search box, from within an item record, the virtual keyboard does NOT pop up._

-------------------------------------------------------------------------------
### Sandal

The icon for the virtual keyboard sort of overlaps the search box, and is gray instead of white. 

image

From the search screen, you click in the search box and the virtual keyboard pops out. It’s positioned a couple of pixels too high; could it be moved lower so you can still see the bottoms of the search box, field selector dropdown, and Find button with a hair of space below them?

I also noted that it would be nice if the enter key were labeled with the word Enter, instead of only displaying an arrow. See circled area. I don't know if there are settings within the virtual keyboard that we can control.

image

From the Results screen, or the No Results screen, the keyboard icon disappears.

image

When you mouse over it, the icon appears, but then when you click in the search box to type a new search, the keys appear to be blank.  (I think they’re actually white on white in this theme and under this circumstance.)

image

However, you can click the blank keys and they will populate the search box.

When you have results and go to an item record, the keyboard icon is invisible (as above). It appears only on direct mouseover.

image

However, when you click in the search box from an item record, the keyboard does not pop out (as in Bootstrap3).

------------------------------------------------------------------------
### Bootprint3

Bootprint3 has some of the same display issues as Sandal, but in other ways it operates correctly.

The keyboard icon overlaps the border of the search box. Can it be moved up slightly so that it fits inside the search box?

image
When you click in the search box, the keyboard appears appropriately. It’s positioned a few pixels too high; could it be moved lower so you can still see the bottoms of the search box, field selector dropdown, and Find button with a hair of space between them and the keyboard?

image

When you get to the search results screen, the keyboard icon is visible.

image

When you click in the search box from the search results screen, the virtual keyboard pops out and you can use the backspace key to clear your search or you can use the physical keyboard.

image

When you open an item record, the keyboard icon is visible but the keyboard does NOT appear when you click in the search box.

image

demiankatz (changes requested):
Thanks, @sturkel89, for the detailed testing. @ThoWagen, when you return from vacation, can you look into the style issues highlighted in the comments above? I imagine that @crhallberg might be able to assist if you have specific questions about the variations between the themes.

ThoWagen:
@sturkel89 Thank you for your input! 
Sorry, I didn't have look at the other themes so I missed this. I did not work with them before but I will fix this.

Sometimes the keyboard was not loaded because if the searchbox is rendered inside of layout.phtml the headScripts are already included and the JS for the keyboards is not loaded. I moved the logic to the start of layout.phtml and it should work now.

I replaced the names of the keys with arrows because unfortunately they can not be translatable without a lot of more work. I thought the arrow keys would be an international standard. But I might be wrong here. However, if you still prefer "Enter" I can change this.

themes/bootstrap3/templates/layout/layout.phtml
```diff
@@ -1,5 +1,14 @@
 <?=$this->doctype()?>
-
+<?php // Set up the search box -- there are three possible cases:
```
demiankatz: I would suggest reformatting this as:

```
<?php
  // Set up the search box -- there are three possible cases:
  // 1. No search box was set; we should default to the normal box
  // 2. It was set to false; we should display nothing
  // 3. It is set to a custom string; we should display the provided version
  // Set up default search box if no data was provided from the template;
  // this covers case 1.  Cases 2 and 3 are then covered by logic below.
  if (!isset($this->layout()->searchbox)) {
    $this->layout()->searchbox = $this->context($this)->renderInContext('search/searchbox.phtml', []);
  }
?>
```

...I think that indentation somewhat improves readability.

It might also be worth adding a new comment, to the effect of "We should make sure the search box renders before we build the HTML header, so that any Javascript dependencies can be loaded correctly."

demiankatz (changes requested):
Thanks for the progress, @ThoWagen -- a couple of minor suggestions on the latest refactoring. Beyond that, please let me know when you're ready for us to take another look at the style issues.

themes/bootstrap3/templates/layout/layout.phtml
ThoWagen: Done

ThoWagen:
@demiankatz @sturkel89 I fixed the other themes.

demiankatz:
Thanks for the progress on this, @ThoWagen! @sturkel89 is out of the office today, but I'll see if she can find some time to review this again after she returns.

sturkel89:
@ThoWagen I have reviewed the branch in all three themes, and everything seems to display and perform correctly! 

The keyboard icon looks great within the search box in every theme, and the keyboard itself fits the page well and resizes nicely as you resize the browser window.

I saw your note about the arrow symbol rather than the word Enter on the Enter key, and I understand your reasoning. I think it's okay with the arrow; keyboard users will be familiar with the function based on the location of the key and the arrow symbol.

I don't have any other critiques - this seems ready to go!

crhallberg (approved):
Looks like some conflicts need to be resolved, but you have my approval!

ThoWagen:
There have been three problems with this:
1. They keyboard selection button had a conflict with the reset button included in #2994. Now the buttons are shown next to each other. The logic was combined and moved into searchbox_controls.js.
2. The keyboard was overlapped by the autocomplete selection. So I moved the keyboard to the bottom of the screen.
3. The keyboard did not trigger the autocomplete. That also works now.

themes/sandal/less/search.less
```diff
@@ -274,7 +301,7 @@
     .filters-term {
       color: #000;
     }
-    a:hover {
+    a:hover {
```
demiankatz, line 308: I'm not sure what's going on here -- I noticed that this PR removed some whitespace that it didn't need to remove, so I re-added the whitespace... but now it's still showing up as a diff. I wonder if there was somehow some strange non-printable character here instead of a regular space, but I don't know how or why that would have happened.

package.json
```diff
@@ -39,7 +39,9 @@
     "grunt-less-to-sass": "duvillierA/grunt-less-to-sass#0.0.11",
     "grunt-sass": "^3.1.0",
     "jit-grunt": "^0.10.0",
-    "node-sass": "^7.0.1"
+    "node-sass": "^7.0.1",
+    "simple-keyboard": "^3.6.0",
```
demiankatz: I believe that the simple-keyboard and simple-keyboard-layouts dependencies should be moved under devDependencies. Since we commit the Javascript/CSS vendor files to the main project, we only need to load them from NPM in order to run the Phing `copynodemodules` task -- and that is a development activity rather than a regular day-to-day operation.  (I realize that there is some existing inconsistency here, and I believe there has been discussion about moving the autocomplete library down in a different PR... just hoping we can avoid compounding the problem here).

themes/bootstrap3/js/common.js
```diff
@@ -687,25 +687,3 @@ $(function commonDocReady() {
...
 });
-
-$(function searchFormResetHandler() {
-  const queryInput = $('#searchForm_lookfor');
-  const resetButton = $('#searchForm-reset');
-
-  let query = queryInput.val();
-  if (query !== '') {
-    resetButton.show();
-    queryInput.focus().val('').val(query);
```
demiankatz: There are some bug fixes currently being worked on in #3064 that impact this code... but it looks like you've removed or significantly rewritten it as part of the refactoring here. I'm going to wait until that PR is finished before doing final testing and merging of this one, since it seems better to fix existing bugs before adding new functionality on top of them. If you have a moment to look at #3064 and comment on it (if it hasn't been merged yet) or reconcile the fixes with this PR (if it has been completed by the time you read it), I'd appreciate the help!

demiankatz (changes requested):
I was hoping to be able to merge this today, @ThoWagen, but unfortunately things have gotten a little complicated because of the introduction of the search form reset button and related bugs. As noted below, I think we have to deal with #3064 before this one can be totally completed. I've left a few very minor comments in the meantime, and I also went ahead and merged the latest dev branch into this one to keep us up to date.

Sorry that this is dragging out so long!

themes/bootstrap3/js/common.js
crhallberg: It looks like this was moved to the [_onChange handler for the keyboard](https://github.com/vufind-org/vufind/pull/2952/files/90efef862a9b165e167aa9951ea56ac9fbc9ec00#diff-fbacfd93e0c0eb3a264f477eff397a1c48a952fc877a3f674e9aee6ebf62b103R44). It looks to me like this should work even when the keyboard is disabled, but I haven't tested it myself.

themes/bootstrap3/js/searchbox_controls.js
```diff
@@ -0,0 +1,143 @@
...
+        layoutName: currentLayoutType === "default" ? "shift" : "default"
+      });
+    }
+
+    if (button === "{enter}") {
+      document.getElementById("searchForm").submit();
+    }
+  }
+
+  function _updateKeyboardLayout(layoutName){
```
demiankatz: Minor point of style: might be nice to replace `){` with `) {` in this file for consistency with project standards. (Would also be nice to get eslint to do that for us sooner or later...)

themes/sandal/less/search.less
ThoWagen: This was an non-breaking space. My IDE showed it like this [NBSP]. I wanted to replace it with a normal space but somehow only deleted it. I think now it is correct.

themes/bootstrap3/js/common.js
ThoWagen: I am not sure what the purpose of #3064 is. Was it introduced so that "Loading..." is not shown when the reset button is used or when the page is loaded initially?

In this PR, whenever a change is made to the input by using the actual keyboard, the virtual keyboard or the reset button the _onChange function is used to update the state for all the elements. And 

>  [_onChange handler for the keyboard](https://github.com/vufind-org/vufind/pull/2952/files/90efef862a9b165e167aa9951ea56ac9fbc9ec00#diff-fbacfd93e0c0eb3a264f477eff397a1c48a952fc877a3f674e9aee6ebf62b103R44)

is used to trigger the autocomplete.

package.json
ThoWagen: I just put it in there because autocomplete.js was also there. But your explanation makes sense. Moved it to devDependencies.

themes/bootstrap3/js/searchbox_controls.js
ThoWagen: Done

themes/bootstrap3/js/common.js
demiankatz: @ThoWagen, I believe the primary purpose of #3064 is to prevent a problem (possibly introduced when the reset button was added) where after performing a search, the browser would auto-focus the search box, and then because the search box was focused, the autocomplete suggestions would drop down and immediately cover the search results. @crhallberg did some work to further ensure that autocomplete results only appear in response to user actions/input, and not automatically on page load.

I haven't had a chance to test whether this branch is subject to that problem, but it's easy enough to test: just perform a search and see whether or not the autocomplete pops open without your intervention.

I'm out of the office today (it's a holiday here) but I'll be back tomorrow and can give this (and your other PRs) a closer look then. Thanks as always for all the progress!

ThoWagen: The behavior you describe does not occur on this PR at least in my setup. But the searchbox is not auto-focused also. 

Is the auto-focus a wanted but it should just not open the autocomplete? In that case, I could add `_textInput.focus();` to the `_onChange` function and it still works.

demiankatz: I think it's generally better to avoid auto-focus for accessibility reasons, so if it's not auto-focusing at present, I don't think we need to make changes to make it do so. :-) Thanks for looking at this!

themes/bootstrap3/js/searchbox_controls.js
```diff
@@ -0,0 +1,143 @@
...
+
+    let layout = window.Cookies.get("keyboard");
+    if (layout == null) {
+      layout = "none";
+    }
+    _updateKeyboardLayout(layout);
+  }
+
+  function init(){
+    _textInput = document.querySelector(".searchForm_lookfor");
```
demiankatz: Is there a reason to use the class `.searchForm_lookfor` for the text input but the id `#searchForm-reset` for the reset button? Might it be safer to use identifiers in both cases to be sure that only one element is selected? I'm concerned that strange and unpredictable things might occur if we ever had a scenario where multiple elements matching the `.searchForm_lookfor` class were on screen at once. (And if we need to support a multi-input scenario, we would then need to revise the reset button logic).

themes/bootstrap3/templates/search/searchbox.phtml
```diff
@@ -67,12 +68,41 @@
...
+              <?=$this->icon('keyboard-o') ?>
+              <?=$this->icon('caret-down') ?>
+            </button>
+            <ul class="dropdown-menu" aria-labelledby="keyboard-selection-button">
+              <li>
+                <a class="keyboard-selection-item" href="#" data-value="none"><?= $this->transEsc('None') ?></a>
+              </li>
+              <?php foreach ($keyboardLayouts as $keyboardLayout): ?>
+                <li>
+                  <a class="keyboard-selection-item" href="#" data-value="<?= $keyboardLayout ?>"><?= $this->transEsc($keyboardLayout) ?></a>
```
demiankatz, line 124: Should we create a text domain for these keyboard layouts so we can translate them without adding strings to the primary language files? That is, create `languages/KeyboardLayout/en.ini` with mappings, and then use `$this->transEsc("KeyboardLayout::$keyboardLayout")` here.

demiankatz (changes requested):
I'm assigning this PR to the 10.0 milestone. It is very mature and should be safe to merge soon, but I still have a few questions, and since it does make some fairly impactful code changes, I think it may be safest not to merge it so late in the development of release 9.1, in case it introduces unforeseen issues.

See below for a couple of specific questions/suggestions.

Additionally, regarding the general user interface: I was initially confused that selecting a virtual keyboard didn't seem to do anything -- I expected the keyboard to pop up immediately when I chose it, rather than later, when the input area was focused. Would it make sense to focus the input when the virtual keyboard is selected? Additionally, I wonder if it would make sense to have a visual indicator that the virtual keyboard is active (such as a highlighted color on the icon) when the setting is something other than "none".

What do you think?

themes/bootstrap3/templates/search/searchbox.phtml
ThoWagen: I'm not sure about this. The layouts are names of languages and are often identical to the translation keys of the language selection except for the case of the first letter. Wouldn't it be better to use the same translations there instead of duplicating most of it?

demiankatz: We actually don't have language translations yet outside of native.ini, which serves a somewhat different purpose. Broader translations are the subject of #413, but it hasn't been resolved since no one has had time to reach consensus on language representation in the index and to build the necessary i18n files. I agree that these efforts could be related. Perhaps this is an opportunity to start something.

themes/bootstrap3/js/searchbox_controls.js
ThoWagen: Changed it to "document.getElementById" in both cases as in #3064. 

I thought about using this also for advanced searches where multiple input fields are present. This would require refactoring this anyway. But that is something for a followup PR.

ThoWagen:
The keyboard is now shown on selection and also the selection button is highlighted if a keyboard is selected, as I also think this makes using it less confusing for users.

themes/bootstrap3/templates/search/searchbox.phtml
demiankatz: (And when I say this is different from native.ini, my thought is that the language selection control helps people find their native language -- i.e. I can spot English in the list, a German- speaker will recognize Deutsche, etc. -- but I think the virtual keyboard labels should be translated to the active language. I might need a Russian keyboard even if I can't speak Russian but need to enter some special characters, but I'd have difficulty knowing which Cyrillic word means Russian.)

demiankatz: I'm open to other arguments, though; maybe my first impression doesn't reflect the real use cases!

ThoWagen: I think you are right about handling language selection translations different than in the keyboard selection.  

I shared an idea in #413 how this could be done.

ThoWagen:
I am not sure why the security check fails. The tests run with no errors locally and I don't have access to view the details.

demiankatz:
@ThoWagen, the security check is failing because it can't run while there are conflicts on the branch. I think you just need to merge dev, run `npm run build:css` and commit the resulting compiled.css artifacts to resolve conflicts.

demiankatz:
@ThoWagen, apologies for introducing more conflicts, but it looks like we need to resolve this with the changes in #3080, where I borrowed some of your ideas from here to clean up our reset button focus problems. I hope that after the merge, the code here will be slightly simplified. Please let me know if you need any help from my end or if you have any questions!

ThoWagen:
@demiankatz I resolved the conflicts. But I basically just have overwritten the changes of #3080 again. You mentioned in your comment that this code might be simplified. But I'm not sure what you mean. Did I miss something?

demiankatz:
Thanks, @ThoWagen. I thought perhaps more of the existing code from #3080 might be applicable here, in which case it would reduce the number of diffs found in this PR... but if that doesn't make sense, then I have no objection to what you have done... it will just require some more thorough testing to be sure all the functionality remains consistent. At the moment I'm focused on finishing up remaining 9.1-targeted work, but I hope to get to this fairly early in the 10.0 development process. Apologies again that things are not moving more quickly -- there's a great deal going on at the moment. :-)

demiankatz (changes requested):
@ThoWagen, now that 9.1 development has settled down, I'm back to reviewing 10.0 PRs. I see that this one needs some conflict resolution. Do you have a moment to take care of that? If so, I'll give it another look once we have a green checkmark again. :-)

ThoWagen:
I updated the branch and now it should work again.

config/vufind/searchbox.ini
```diff
@@ -56,3 +56,48 @@ type[] = External
 target[] = "http://www.google.com/search?q="
 label[] = Google
 group[] = "Other Sites"
+
+[VirtualKeyboard]
```
demiankatz, line 62: Might be worth adding a header comment here, like:

```
; Uncomment one or more of the lines below to enable virtual on-screen keyboard support
; using the specified layout(s).
```

themes/bootstrap3/templates/search/searchbox.phtml
demiankatz: Some work is moving forward now in #3200. Do you think we should focus on finishing that before wrapping this up, or should we find an interim solution here and add a TODO there to update things to handle keyboard layouts when possible?

demiankatz (changes requested):
This is looking very good to me. I just had one thought on documentation, plus we need to figure out how to resolve the translation piece (see details below).

config/vufind/searchbox.ini
ThoWagen: Added it

themes/bootstrap3/templates/search/searchbox.phtml
ThoWagen: I think we can finish this now. Other PRs might introduce conflicts again in the future. And I don't think that having translations for the layouts is that critical. Even if they are not included in 10.0 if anyone wants to use this feature they can add the translations in their local configs.

demiankatz: I've added a TODO item on #3200 to set up translation for keyboard layout names so that we don't lose track of this task.

demiankatz (approved):
I believe this is ready now -- thanks for all the work on it! I'm merging now, and will work on the translation piece as part of #3200 when time permits. I'm also open to a follow-up to add advanced search support, as discussed earlier, if and when you'd like to work on that.

PR #2953, opened 2023-06-23
EPF: Test record driver and changes to EDS Backend

maccabeelevine:
Relatively minimal new tests needed, because the EPF data source reuses much of the EDS code.


maccabeelevine:
@demiankatz I got these new tests to pass, and hopefully they are useful.  But couldn't figure out how to get around the phpstan error.  I get (I think) that when you instantiate a class via reflection you need to give some code hints to phpstan, but not clear from the link it provides how to do that.  I could rewrite the code to avoid the reflection (passing the class name as a string or something), but there must be a better way.

demiankatz (changes requested):
> @demiankatz I got these new tests to pass, and hopefully they are useful. But couldn't figure out how to get around the phpstan error. I get (I think) that when you instantiate a class via reflection you need to give some code hints to phpstan, but not clear from the link it provides how to do that. I could rewrite the code to avoid the reflection (passing the class name as a string or something), but there must be a better way.

Fortunately, I think the solution to this is much easier than you may fear -- you are already passing the class name as a string (the `::class` constant just resolves to a string containing the class name). However, your typehint on the function that phpstan is complaining about has the type set to `class` which is confusing the analysis. Just change `class` to `string` and it should pass. :-)

maccabeelevine:
Wish they were all that easy!

demiankatz (approved):
Looks good -- merging now!

PR #2958, opened 2023-06-27
wrap h2 around title and author, add span around author

elsenhans:
This PR is wrapping a `<h2>` element for the titles and authors in all lists (result-list, saved items and all other lists available in the account). Additionally the author is wrapped by a `<span>` element.

If screen reader users jump from one record heading to the next, they may search for the authors and not find them, since they logically expect this content after the heading.
By combining title and author in one `<h2>` element, screen reader users will find both information.

Since we only do have data for the Default-RecordDriver, I am not able to access the **result-lists in the other RecordDrivers**. I still can adapt these templates but I am not able to check the changes visual. 

TODO
- [x] After merging this, open a follow-up PR to refactor code to `RecordDriver/DefaultRecord/account-entry.phtml` as discussed in comments below.


demiankatz:
Thanks, @elsenhans. Unfortunately, it looks like these changes were made against an old commit, and this PR has a significant number of conflicts with the current dev branch. Please let me know if you need any help with conflict resolution!

elsenhans:
@demiankatz Can you tell me what I had done wrong?
Did I checked out the wrong source branch?
I checked out branch _release-9.0_ and forked my branch from it. Should I use the _dev_ branch in the future?
Not sure if I will be able to resolve all the conflicts. What do you think will be the best solution?
Sorry for all the trouble!

demiankatz:
@elsenhans, the release-9.0 branch is for bug fixes to release 9.0, whereas the dev branch is for development of new features for the next release. The dev branch has moved forward quite a bit as we work on release 9.1, hence the conflicts. I might be able to help with conflict resolution tomorrow if necessary; just let me know!

elsenhans:
@demiankatz  I would appreciate your help for resolving the conflicts!

demiankatz (changes requested):
I took care of merging the conflicts, @elsenhans. I think I did it right, but if you could review, I would appreciate it! Also, a couple of general questions:

1.) Should we add h2 elements to the RecordDriver/DefaultRecord/result-grid.phtml template as well? (You can see this template by turning on the grid view option in the Views section of searches.ini).

2.) Should the span around the author information have a class on it? Otherwise, what purpose does it serve?

elsenhans:
Thank you for merging the dev-branch @demiankatz !
Please excuse my mistake once again.


To your requests:

**1.** Yes, I think we should add the `<h2>`-element also for the **title** in _RecordDriver/DefaultRecord/result-grid.phtml_ although it is not for the purpose of grouping the title and author, because it seems there is no author rendered.
There are also all the _result-lists_ of RecordDrivers, where we do not have records for: _LibGuides_, _BrowZine_, _EDS_, _Pazpar2_, _SolrAuthDefault_ where at least the title is rendered. There we should probably do the same if the hierarchy of the headlines is identical to the one of _DefaultRecord/result-list_. But I am not able to check any changes on these RecordDrivers visually.

**2.** At the moment the `<span>` element is not needed for any stylings. The stylings are still working which had been defined for the `<a>`-element. The `<span>` was just set to make a clear separation between title and author. That means I can either **remove the `<span>`-element** if you think it is not necessary at all **or** I can **add a classname**. With the classname and the span it would be easier for libraries to set their specific styling. Which variation do you prefer?

demiankatz:
Thanks, @elsenhans, and no worries!

To answer your questions:

1.) Please go ahead and make the suggested changes, including to the other result-list templates. I can test the majority of these on my end.

2.) I think adding a classname to the span is probably the most flexible option, unless @crhallberg prefers to remove the span.

A related thought: is it perhaps worth considering refactoring to a record driver template for some of the repeated logic? For example, is the logic to display the title and author identical across all of the account-related templates (myresearch/*.phtml, checkedout/history.phtml, etc.)? If so, I wonder if creating a RecordDriver/DefaultRecord/account-list-header.phtml file (or something like that) might simplify maintenance/customization of this. But I didn't spend a lot of time thinking about this idea or comparing the code, so maybe it is not as uniform as I am hoping.

elsenhans:
@demiankatz I have implemented the 2 points.

Well, the title and author are not rendered in all templates together. But in general it would be great to outsource parts that appear more than once!

demiankatz:
@elsenhans, if you search all the templates for a comment that begins "If this is a non-missing Solr record", you'll find five different templates that contain very nearly identical logic for displaying basic record details. The differences between the versions seem to be due to fixes being applied inconsistently, rather than a conscious need to differentiate. I think it would make sense to refactor the block starting at `<h2>` and ending at the display of publication_year into a separate `RecordDriver/DefaultRecord/account-entry.phtml` template that could be loaded via `$this->record($resource)->renderTemplate('account-entry.phtml')`.

However, I don't think we should make more changes to this PR, in order to keep things simple. I will add a TODO item as a reminder to follow up with that refactoring after the work here is merged, unless you feel it is a bad idea.

elsenhans:
@demiankatz sounds good to me.

themes/bootstrap3/templates/RecordDriver/Pazpar2/result-list.phtml
```diff
@@ -18,32 +18,32 @@
...
-    <div>
       <?php $summAuthors = $this->driver->getPrimaryAuthorsWithHighlighting(); ?>
       <?php if (!empty($summAuthors)): ?>
-        <?=$this->transEsc('by')?>
-        <?php $authorCount = count($summAuthors); ?>
-        <?php foreach ($summAuthors as $i => $summAuthor): ?>
-          <a href="<?=$this->record($this->driver)->getLink('author', $this->highlight($summAuthor, null, true, false))?>" class="result-author"><?=$this->highlight($summAuthor)?></a><?=$i + 1 < $authorCount ? ',' : ''?>
-        <?php endforeach; ?>
+        <br>
+        <span class="authors">
```
crhallberg, line 28: Instead of a `<br>`, can we use a block element like `<div>` or block styles for authors?

themes/bootstrap3/templates/checkouts/history.phtml
```diff
@@ -107,26 +107,30 @@
...
+                    echo '<a href="' . $this->escapeHtmlAttr($this->recordLinker()->getUrl($resource)) .
+                      '" class="title">' . $title . '</a>';
+                  } elseif (isset($ilsDetails['title']) && !empty($ilsDetails['title'])) {
+                    // If the record is not available in Solr, perhaps the ILS driver sent us a title we can show...
+                    echo $this->escapeHtml($ilsDetails['title']);
+                  } else {
+                    // Last resort -- indicate that no title could be found.
+                    echo $this->transEsc('Title not available');
+                  }
+                ?><br>
```
crhallberg: Same question as above. I don't know if it's useful to flag every location.

crhallberg (changes requested):

themes/bootstrap3/templates/RecordDriver/Pazpar2/result-list.phtml
elsenhans: Thanks for the hint @crhallberg !
I removed the breaks before authors and set `display: block`.

crhallberg (approved):

demiankatz (approved):
Thanks again, @elsenhans! I've done hands-on testing of all of this. I made some minor adjustments to the styling for grid view to address some inconsistencies created by the PR; now everything visually looks the same while having better underlying semantic markup.

elsenhans:
Thanks a lot @demiankatz !

PR #2959, opened 2023-06-27
Add support for Laminas 'Dumb' CAPTCHA.

demiankatz:
This addresses the "bare minimum" suggestion on [VUFIND-1611](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1611) for a more accessible CAPTCHA option.

It would definitely be desirable to do something better than this, but this is the best we can do using only components provided in the laminas-captcha library.


mtrojan-ub:
That's pretty straightforward, although the Laminas documentation states that it should be used only for testing purposes.

https://docs.laminas.dev/laminas-captcha/adapters/

(But in the end since it's just another configurable alternative, that's fine for me.)

demiankatz:
Thanks, @mtrojan-ub -- I did note in the config.ini comment that this option is only recommended for testing. :-) If you think that's good enough, please give me an approval and I'll merge it!

mtrojan-ub (approved):

PR #2962, opened 2023-06-27
[VUFIND-1568] Make Solr port configurable during installation

demiankatz:
This PR is intended to make use of a non-standard Solr port easier to configure.

It includes some changes to the installation process, including the creation of an `env.sh` file for Unix equivalent to the existing `env.bat` used in Windows, and modifications to the .deb package installation to utilize the new file.

The command-line `install.php` script now prompts for Solr port number (and supports a `--solr-port` switch to specify it from the command line). This impacts environment variable setup and the local `import/import.properties` file. The web-based installer now reads the Solr URL out of the local `import/import.properties` when establishing baseline `config.ini` settings. This means that, as long as the installation steps are followed on the expected/appropriate order, the user is prompted for Solr port once, and this port number is applied in all the necessary/appropriate places.

TODO
- [x] Add Solr port setting to install.php script
- [x] Incorporate Solr port setting into config.ini initialization in web-based installer
- [x] Improve test coverage
- [x] Test .deb installation to be sure new process works correctly
- [x] Resolve [VUFIND-1568](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1568) when merged
- [x] Open a follow-up ticket to address inconsistencies in behavior related to existing files (right now we skip the import.properties file if it exists, but we overwrite and back up other files... it would make more sense to have a switch which controls which behavior is used for all duplicate files).


demiankatz:
I just built the .deb package and confirmed that the new workflow functions correctly. I believe this PR is ready for review. I also noticed during testing that we have an inconsistency in how we handle the installer overwriting pre-existing configuration files, so I added a TODO to follow up with a separate PR to sort that out. I didn't want to make this PR more complicated than necessary by adding more changes here.

EreMaijala (approved):
@demiankatz Awesome! This almost solved my gripe with having to run integration tests with Solr in port 8983. I pushed a commit that adds support for specifying the port in build.xml as well and verified that everything works as expected. Can you please check the last commit if it's ok for you? If so, I think this is good to go.

demiankatz:
@EreMaijala, thanks for this -- I'm glad this helps to support the build.xml use case! One question before I merge: how does the correct port number get set in config.ini? Normally that happens via the web-based installer, but in the continuous integration process, I think we might need to add an additional search-and-replace to the config.ini setup to make it work right, since we don't use the web-based installer in the CI context. Did you test this end-to-end? Maybe I'm missing something, but just wanted to make sure before merging (and I don't have enough free time to test it myself right now).

EreMaijala:
Err, oops. I thought I checked everything but the tests were actually skipped. I'll revise.

EreMaijala:
Now fixed. I also added Search2.ini for tests that use Search2 and added the env and import properties files to the cleanup so that reset_setup works properly.

demiankatz:
Thanks, @EreMaijala!

PR #2963, opened 2023-06-28
php-cs-fixer: Remove disabled braces rule configuration.

demiankatz:
The braces rule was disabled for php-cs-fixer template configuration because it was causing problems. However, in newer versions of php-cs-fixer, it is no longer an issue, and the braces rule was causing a deprecation warning. This PR simply removes the offending setting, which does not have an adverse effect on template formatting.


LuomaJuha (approved):
Seems to work :)

demiankatz:
Thanks, @LuomaJuha!

PR #2964, opened 2023-06-28
LibGuidesAZ: Add results recommendation module

maccabeelevine:
A *Results recommendation module for displaying a list of recommended databases from the LibGuides AZ data source, most likely in the sidebar.  Patterned vaguely after EDSResults, but it's simpler.


languages/en.ini
```diff
@@ -714,6 +714,8 @@ Last Name = "Last Name"
 Last Search Result = "Last Search Result"
 less = "less"
 less_ellipsis = "less…"
+libguidesaz_recommendations = "Database Results"
```
maccabeelevine, line 717: I should note that I found one other "databases" recommendation module, for Summon, where the default text is a vague 

> "You may find additional resources here:"

If that's a pattern we have to follow then fine.  But my text is more in line with the other SearchObject recommendation modules.

module/VuFind/src/VuFind/Recommend/LibGuidesAZResultsDeferred.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * Get the URL parameters needed to make the AJAX recommendation request.
+     *
+     * @return string
+     */
+    public function getUrlParams()
+    {
+        return 'mod=LibGuidesAZResults&params=' . urlencode($this->processedParams)
```
maccabeelevine, line 133: This file is a complete copy/paste from EDSResultsDeferred, except for the 'mod' param in this one line.  I think separately it should be refactored into some sort of SearchObjectDeferred.php?  Saying this partially because once this PR is through, I plan to create additional recommendation modules for LibGuides (traditional, not AZ) and EPF, which will copy/paste this twice more.

themes/bootstrap3/templates/Recommend/LibGuidesAZResults.phtml
```diff
@@ -0,0 +1,19 @@
...
+  $searchObject = $this->recommend->getResults();
+  $results = $searchObject->getResults();
+?>
+<?php if (!empty($results)): ?>
+  <h4><?=$this->transEsc('libguidesaz_recommendations')?></h4>
+  <div class="list-group">
+    <?php foreach ($results as $driver): ?>
+      <div class="list-group-item">
+        <span>
+          <a href="<?=$driver->getUniqueID()?>" class="title">
```
maccabeelevine: Is it legit to go straight to $driver for this, rather than through some view helper method?  I realize it's weird that the unique ID is the URL, but that's what the driver does, inherited from the original LibGuides driver.  It kind of makes sense as there is very little data retrieved from the data source.

demiankatz (approved):
Looks good to me, @maccabeelevine, and it works fine in my test environment. I see you have this marked as draft, but if you think you're done, I'm happy to merge whenever you're ready.

maccabeelevine:
@demiankatz Just wanted to write up my code comments before marking it ready for review.  If you ok all of those, then fine to merge.  Thanks.

languages/en.ini
demiankatz: This makes sense to me -- I like the pattern of having a module-specific translation string, since that allows targeted customization.

module/VuFind/src/VuFind/Recommend/LibGuidesAZResultsDeferred.php
demiankatz: Agreed, it probably makes sense to create an abstract base class, and then make all of these other versions subclasses that just override a property or two.

themes/bootstrap3/templates/Recommend/LibGuidesAZResults.phtml
demiankatz: If that's consistently how LibGuides behaves, I don't see a problem with it... though I think you should wrap this in `$this->escapeHtmlAttr()` to ensure that the link is properly escaped.

demiankatz (changes requested):
Thanks, @maccabeelevine -- in following up on your comments, I noticed one very small thing that would benefit from adjusting...

module/VuFind/src/VuFind/Recommend/LibGuidesAZResultsDeferred.php
maccabeelevine: Will do next.

themes/bootstrap3/templates/Recommend/LibGuidesAZResults.phtml
maccabeelevine: Done

demiankatz (approved):
Looks good -- merging now. I'll look forward to that refactoring PR whenever time permits. :-)

PR #2966, opened 2023-06-28
Refactor *Deferred recommendation modules

maccabeelevine:
Refactor some/all *Deferred recommendation modules that ultimately point to SearchObject implementations.

### DIFF
--- a/module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
+++ b/module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
@@ -129,7 +129,7 @@ abstract class AbstractSearchObjectDeferred implements RecommendInterface
      * @param \VuFind\Search\Base\Params $params   Search parameter object
      * @param \Laminas\Stdlib\Parameters $request  Parameter object representing user
      * request.
-     * @param array                      $settings Parameter array
+     * @param array                      $settings Parameter array (passed by reference)
      *
      * @return void
      */
--- a/module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
+++ b/module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
@@ -68,6 +68,13 @@ abstract class AbstractSearchObjectDeferred implements RecommendInterface
     protected $processedParams;
 
     /**
+     * Number of expected module parameters (from .ini config)
+     *
+     * @var int
+     */
+    protected $paramCount = 2;
+
+    /**
      * Store the configuration of the recommendation module.
      *
      * @return string Module name in call to AjaxHandler
@@ -107,15 +114,32 @@ abstract class AbstractSearchObjectDeferred implements RecommendInterface
         // Make sure all elements of the params array are filled in, even if just
         // with a blank string, so we can rebuild the parameters to pass through
         // AJAX later on!
-        for ($i = 0; $i < 2; $i++) {
+        for ($i = 0; $i < $this->paramCount; $i++) {
             $settings[$i] ??= '';
         }
 
+        $this->initLookFor($params, $request, $settings);
+
+        $this->processedParams = implode(':', $settings);
+    }
+
+    /**
+     * Initialize the lookFor query parameter.  Called from init().
+     *
+     * @param \VuFind\Search\Base\Params $params  Search parameter object
+     * @param \Laminas\Stdlib\Parameters $request Parameter object representing user
+     * request.
+     * @param array $settings Parameter array
+     * request.
+     *
+     * @return void
+     */
+    protected function initLookFor($params, $request, &$settings)
+    {
         // Map the user-specified query field to 'lookfor' for simplicity:
         $this->lookfor
             = $request->get(empty($settings[0]) ? 'lookfor' : $settings[0], '');
         $settings[0] = 'lookfor';
-        $this->processedParams = implode(':', $settings);
     }
 
     /**
--- a/module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
+++ b/module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
@@ -44,7 +44,7 @@ namespace VuFind\Recommend;
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org/wiki/development:plugins:recommendation_modules Wiki
  */
-abstract class SearchObjectDeferred implements RecommendInterface
+abstract class AbstractSearchObjectDeferred implements RecommendInterface
 {
     /**
      * Raw configuration parameters
--- a/module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
+++ b/module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
@@ -126,11 +126,10 @@ abstract class AbstractSearchObjectDeferred implements RecommendInterface
     /**
      * Initialize the lookFor query parameter.  Called from init().
      *
-     * @param \VuFind\Search\Base\Params $params  Search parameter object
-     * @param \Laminas\Stdlib\Parameters $request Parameter object representing user
+     * @param \VuFind\Search\Base\Params $params   Search parameter object
+     * @param \Laminas\Stdlib\Parameters $request  Parameter object representing user
      * request.
-     * @param array $settings Parameter array
-     * request.
+     * @param array                      $settings Parameter array
      *
      * @return void
      */
--- a/module/VuFind/src/VuFind/Recommend/AbstractSummonRecommendDeferred.php
+++ b/module/VuFind/src/VuFind/Recommend/AbstractSummonRecommendDeferred.php
@@ -42,13 +42,6 @@ namespace VuFind\Recommend;
 abstract class AbstractSummonRecommendDeferred extends AbstractSearchObjectDeferred
 {
     /**
-     * Name of module being loaded by AJAX -- MUST be set in constructor!
-     *
-     * @var string
-     */
-    protected $module = null;
-
-    /**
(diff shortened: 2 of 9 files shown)


maccabeelevine:
So I initially refactored just the new LibGuidesAZResultsDeferred and the class I based it off, EDSResultsDeferred.  They were identical minus a single string parameter.

Other candidates to extend the new abstract SearchObjectDeferred would be
- AbstractSummonDeferred (which itself has two subclasses)
- EuropeanaResultsDeferred
- OpenLibrarySubjectsDeferred

Those three all have a slightly more complex init() method.  The simplest thing to do would be a straight override of that method.  I could also try to override with a method that calls the abstract implementation and then adds the extra stuff, but I'm not confident in the details so I'd prefer not.  More importantly I have access to neither Summon or Europeana to test.  (In theory I can figure out how to test OpenLibrary.)

So options are:
1. Stop here
2. Add OpenLibrary only, which I can probably test.
3. Refactor those other three with simple overrides of init(), and get help testing from people who can
4. Try a more complex refactoring, with (more) help testing

module/VuFind/src/VuFind/Recommend/SearchObjectDeferred.php
```diff
@@ -0,0 +1,146 @@
...
+ * This class sets up an AJAX call to trigger a call to some SearchObject implementation.
+ *
+ * @category VuFind
+ * @package  Recommendations
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Maccabee Levine <msl321@lehigh.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:plugins:recommendation_modules Wiki
+ */
+abstract class SearchObjectDeferred implements RecommendInterface
```
demiankatz: I would normally name this AbstractSearchObjectDeferred to indicate that it's an abstract class, but I see that you're following the pattern of the existing SearchObject class, which is also abstract without having Abstract in the name. Maybe we should fix that in a separate PR -- we could rename SearchObject to AbstractSearchObject, then create a deprecated SearchObject class that extends it for backward compatibility purposes. Or maybe it's not worth the effort. Any thoughts/opinions? I like using the Abstract prefix since it groups the abstract classes together and makes it clear that you can't instantiate them... but maybe I'm overrating the value vs. effort.

demiankatz (changes requested):
Thanks for getting the ball rolling on this, @maccabeelevine. I can help test some, though not all, of the modules you reference. However, I notice that we do have [100% unit test coverage](https://vufind.org/jenkins/job/VuFind/clover/VuFind/src/VuFind/Recommend/index.html) on the deferred result classes, so presumably if refactoring breaks something, it will also break a test, so I think it's probably pretty safe to tackle all of them.

Since as you note we've already refactored some of the common functionality in the case of the Summon-related modules via [AbstractSummonRecommendDeferred](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/Recommend/AbstractSummonRecommendDeferred.php), it makes sense to me to generalize this further, with a base class like AbstractDeferredRecommend that serves as the parent for AbstractSummonRecommendDeferred and any new abstract class you end up with. What do you think?

module/VuFind/src/VuFind/Recommend/SearchObjectDeferred.php
maccabeelevine: Right I would always use Abstract in the name, but thought following the pattern was more important here.  I'd say consider separately.

demiankatz: If we're going to consider this separately, maybe we should go ahead and rename the new class to Abstract now so we don't have to make more changes later. I'm happy to open a PR to do the other renaming in parallel with this work if that would help.

maccabeelevine:
Ok I refactored the abstract Summon class (which wasn't actually marked abstract -- that's fixed) and the Europeana class to also extend the new SearchObjectDeferred abstract class.  I decided *not* to refactor OpenLibrarySubjectsDeferred since it doesn't work the way all the other deferred SearchObject classes do; it actually extends its non-deferred equivalent.

I think this is refactored now to the extend that's useful.  As I mentioned above I could go farther with common parts of init() but I think that's overkill.

module/VuFind/src/VuFind/Recommend/SearchObjectDeferred.php
maccabeelevine: Sorry missed this comment.  I'll rename the new class.

module/VuFind/src/VuFind/Recommend/AbstractSummonRecommendDeferred.php
```diff
@@ -104,11 +71,6 @@ public function setConfig($settings)
      */
     public function init($params, $request)
     {
-        // Validate module setting:
-        if (null === $this->module) {
```
demiankatz, line 108: You can also remove the property definition for $module above if it's no longer being used here.

module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
```diff
@@ -0,0 +1,146 @@
...
+     */
+    public function init($params, $request)
+    {
+        // Parse out parameters:
+        $settings = explode(':', $this->rawParams);
+
+        // Make sure all elements of the params array are filled in, even if just
+        // with a blank string, so we can rebuild the parameters to pass through
+        // AJAX later on!
+        for ($i = 0; $i < 2; $i++) {
```
demiankatz: I think it probably makes sense to factor the `$paramCount` logic from the AbstractSummonRecommendDeferred class to the base class, since it would allow you to eliminate the "magic number" 2 here. You could then just revise the property value in the Summon base class to retain existing behavior. You could also potentially set the property to 4 and use it in the Europeana module's init method, even if you don't decide to do further refactoring there.

demiankatz (changes requested):
Thanks, @maccabeelevine, this is coming along nicely -- I just had a couple of small suggestions below. I agree with your decision to leave OpenLibrarySubjectsDeferred alone. In theory we might be able to share some logic with a trait, but that's probably more trouble than it's worth, since the common methods are trivial.

module/VuFind/src/VuFind/Recommend/AbstractSummonRecommendDeferred.php
maccabeelevine: Oops.  Done.

module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
maccabeelevine: Ok, I refactored $paramCount.  And at that point it just made sense to refactor init() as well, although feedback welcome on how I did it.  

I also don't love that $paramCount is an overridden parameter and getAjaxModule is an abstract method, could use either approach for both.  I preferred the abstract method for getAjaxModule() since it forces an implementation and there is no default.  For $paramCount there is a default, but it could still be a method if that consistency is helpful.  But maybe better as is.

demiankatz: I think it's okay to use different approaches for these two parameters for the reasons you state: one has a default value, and the other is required. I think it's fine as-is!

module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
```diff
@@ -0,0 +1,169 @@
...
+        $this->processedParams = implode(':', $settings);
+    }
+
+    /**
+     * Initialize the lookFor query parameter.  Called from init().
+     *
+     * @param \VuFind\Search\Base\Params $params   Search parameter object
+     * @param \Laminas\Stdlib\Parameters $request  Parameter object representing user
+     * request.
+     * @param array                      $settings Parameter array
```
demiankatz: I think you may want to add an `&` to the doc comment here to clarify that this is passed by reference (though maybe I'm mistaken and the style checker won't like that). If you can't put the `&` in, I'd at least add "(passed by reference)" to the description to make this clear, because it's an important point!

It might also be worth putting in `use` statements at the top of the file for these classes so you don't have to use the fully qualified names in the doc comments. That might improve readability (though no pressure to change it if it feels like a hassle).

demiankatz: An alternative to consider could be having initLookFor return $settings, and then you avoid some of the possible pitfalls/confusion of pass-by-reference by making the data flow a little more explicit. Again, no pressure to change, just an option to consider.

demiankatz (changes requested):
I think we're nearly done here... just a couple of thoughts regarding the newly-introduced initLookfor support method...

module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
maccabeelevine: Good call about '&' but that's [not supported](https://docs.phpdoc.org/guide/references/phpdoc/tags/param.html#param) (and php-cs-console complains) so I just added the "passed by reference" text instead.

Yeah I considered just returning $settings, but I thought semantically that was wrong, as it's not the *point* of initLookFor(), and some implementations (such as EuropeanaResultsDeferred) don't actually touch it.  The side effect semantics seemed more correct.

I didn't add those 'use' statements, as this followed the (not great) pattern used for init() itself, as well as process().  So as usual I preferred to be consistent if maybe not what's best.

demiankatz: Perfect, thanks! No problem about leaving the use statements out for now. I imagine one of these days one of our code standard tools will just decide to do that job automatically for us. ;-)

demiankatz (approved):

PR #2967, opened 2023-06-29
Active favorites in menu

xrosecky:


demiankatz (approved):
Thanks, @xrosecky! I have backported this fix to the release-9.0 branch, so if we decide to make a 9.0.3 release it will be included there. I have also merged it forward to dev.

PR #2968, opened 2023-06-29
Browse: Use same localized number formats like in facets

mtrojan-ub:


demiankatz (approved):
Good idea -- thanks, @mtrojan-ub!

PR #2969, opened 2023-06-29
Rename abstract SearchObject class to AbstractSearchObject.

demiankatz:
This complements the work in #2966 (as discussed there).

TODO
- [x] When merging, add note to next major version deprecation cleanup ticket to remove SearchObject class.

### DIFF
--- a/module/VuFind/src/VuFind/Recommend/AbstractSearchObject.php
+++ b/module/VuFind/src/VuFind/Recommend/AbstractSearchObject.php
@@ -6,7 +6,7 @@
  *
  * PHP version 8
  *
- * Copyright (C) Villanova University 2010.
+ * Copyright (C) Villanova University 2010-2023.
  *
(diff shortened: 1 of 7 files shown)


maccabeelevine (approved):
LGTM.  My only feedback is it would be nicer in the Git history if AbstractSearchObject were a renamed SearchObject, and the new SearchObject was separate.  So you could still do a git blame on all the details.  But I don't think Git can do that if you're introducing a new SearchObject class at the same time.  Only alternative (maybe) to trick Git would be to delete the new SearchObject and add it as a separate PR?

demiankatz:
Good suggestion, @maccabeelevine -- I've manipulated the commit history in this PR to make things more clear, and I'll just rebase this onto dev instead of squashing it to give us clean results in the overall history. Thanks!

maccabeelevine:
Your gitfu is strong

PR #2973, opened 2023-06-29
EPF: Add results recommendation module

maccabeelevine:
Add a recommendation module for EBSCO Publication Finder search results.


languages/en.ini
```diff
@@ -397,6 +397,8 @@ empty_search_disallowed = "An empty query is not allowed with the search target"
 Enable Auto Config = "Enable Auto Config"
 End Page = "End Page"
 eol_ellipsis = "…"
+epf_recommendations = "Publication Results"
```
maccabeelevine, line 400: We use EPF specifically for *e-journal* title search, so we'll be re-labelling this in local/config to something like "Journal Results" or "eJournal Results".  But technically it's "publication" finder, so I figure that's what the default label should be.

demiankatz (approved):
Thanks, @maccabeelevine, this looks good and works for me, so I'm going to merge it as-is since I think it represents positive incremental progress. However, I think it is probably worth opening a follow-up PR to add some more detail to some of the EPF displays. In particular:

1.) When I do an EPF search, I see full text links accompanying each result. However, if I click through to the EPFRecord screen, I actually get less detail, and none of the URLs are accessible! This is a problem with the recommendation module since it sends the user directly to the record page instead of to search results. I suspect that we need to customize the EPF record driver and/or its core.phtml template to make these important links accessible. Let me know if you need help working out the details!

2.) It seems possible that the EPF recommendations will include identical results. For example, using Villanova's settings, if I do a search for "Chemistry" my top three recommendations are all just called "Chemistry" with no differentiating details. Would it be a good idea to add more detail to the results display (perhaps publisher name, or a date range) to make it easier to tell things apart?

maccabeelevine:
> 1.) When I do an EPF search, I see full text links accompanying each result. However, if I click through to the EPFRecord screen, I actually get less detail, and none of the URLs are accessible! This is a problem with the recommendation module since it sends the user directly to the record page instead of to search results. I suspect that we need to customize the EPF record driver and/or its core.phtml template to make these important links accessible. Let me know if you need help working out the details!

I agree and had the same thought when testing.  The recommendation module is arguably* doing the right thing, but the EPF record page should have a lot more useful information.  At the moment the record page is actually less useful than the EPF search results page listing for the same item, even on default_view=title, which is nuts.  So it's on my list and yes, I will probably need help with the details once I get into it.

* Arguably because, actually, if libraries worked in some ideal state, there would be no need for a record page at all, and the search result (or recommendation module item) would just take you directly to the publication itself, with full-text access to every issue ever, so you don't have to choose which database to use for it.  <sigh/>

> 2.) It seems possible that the EPF recommendations will include identical results. For example, using Villanova's settings, if I do a search for "Chemistry" my top three recommendations are all just called "Chemistry" with no differentiating details. Would it be a good idea to add more detail to the results display (perhaps publisher name, or a date range) to make it easier to tell things apart?

Hmm I didn't encounter that but obviously depends on the library and the search.  Yeah I think some minimal additional info would be reasonable, but would need some guidance from ERM folks on what that would be -- the two you suggest seem reasonable but I'm not sure how to weigh it.  I think we would not want to make the recommendation sidebar get too lengthy.

demiankatz:
>Arguably because, actually, if libraries worked in some ideal state, there would be no need for a record page at all, and the search result (or recommendation module item) would just take you directly to the publication itself, with full-text access to every issue ever, so you don't have to choose which database to use for it.

The problem with the "one magic link" approach comes when a publication has different ranges of holdings in different databases (or when there's bad data in EBSCO's database, which happens from time to time). I think having the ability to choose among options is often valuable. Though in theory, it might be fairly easy to link directly to an institution's link resolver from VuFind to avoid the need for intermediate screens. We couldn't offer that as a standard/default option since not everyone has a link resolver... but it would be a very easy minor customization if somebody wanted to do it.

> Hmm I didn't encounter that but obviously depends on the library and the search. Yeah I think some minimal additional info would be reasonable, but would need some guidance from ERM folks on what that would be -- the two you suggest seem reasonable but I'm not sure how to weigh it. I think we would not want to make the recommendation sidebar get too lengthy.

I suggested year/publisher based on the precedent that the existing cross-reference recommendation modules that display more than just title typically include a year and an author... and since authors aren't as relevant at the publication level, publisher seemed like the next best roughly equivalent thing. Of course, I'd welcome input from ERM folks!

PR #2974, opened 2023-07-03
Fetch only requested facets from Solr

xrosecky:
Replaces #2271.

### DIFF
--- a/module/VuFind/src/VuFind/Recommend/SideFacets.php
+++ b/module/VuFind/src/VuFind/Recommend/SideFacets.php
@@ -257,8 +257,12 @@ class SideFacets extends AbstractFacets
      */
     public function init($params, $request)
     {
+        $mainFacets = $this->mainFacets;
+        if ($request != null && ($enabledFacets = $request->get('enabledFacets', null)) !== null) {
+            $mainFacets = array_intersect_key($mainFacets, array_flip($enabledFacets));
+        }
         // Turn on side facets in the search results:
-        foreach ($this->mainFacets as $name => $desc) {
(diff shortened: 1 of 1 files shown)


xrosecky:
I'm not sure whether enabledFacets should apply also on checkbox facets?

demiankatz:
@xrosecky, I think @EreMaijala said that he has some local customizations that allow checkbox facets to include result counts, so failing to exclude checkbox fields might break that. @EreMaijala is on holiday until August, so we might have to wait until then to get more clarification from him.

Where/how is this new feature used? Do you plan to include an example of setting up the enabledFacets setting? I think Ere's requirements may depend on how the rest of the solution falls into place.

xrosecky:
It is optimalization - pass to Solr only required facets, parameter enabledFacets is already used in AJAX calls:

https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/js/facets.js#L178

demiankatz:
Ahh, thanks for the clarification, @xrosecky! We should probably wait for further input from @EreMaijala before merging this, but as long as his use case is met and the tests pass, this seems straightforward to me.

EreMaijala (approved):
This should be fine. We actually do the same but also filter checkbox facets by `enabledFacets`. I'm not sure but perhaps you could do the same here.

demiankatz:
Thanks, @EreMaijala and @xrosecky. I'm going to merge this as-is since it's forward progress. If more work on checkbox facets would further improve performance, please feel free to open a follow-up PR focused on that issue.

PR #2977, opened 2023-07-05
Add recommendation module for LibGuides Profiles

maccabeelevine:
A LibGuides Profile page contains information about a LibGuides user.  In practice it's for a librarian who created one or more subject guides.  From https://ask.springshare.com/libapps/faq/638

> Your profile page displays all of the information from your profile box, in addition to:
> - An optional personal statement
> - A list of all of the guides you own in LibGuides
> - Any LibGuides subject categories to which you're assigned as an expert

This recommendation module displays the photo and basic information about a librarian who would be relevant based on the search query.  To be specific:

- Relevancy is based on comparing the query with the names of the subject guides owned by each librarian.  That's not a great metric, but it works with no additional configuration so that's nice.  I did some testing with a local subject librarian and this worked acceptably well.  However a future enhancement could use local configuration to associate a profile to specific search terms, or to specific factors from the result set (i.e. call numbers).
- The default (boostrap3) template displays their name, title (i.e. Humanities Librarian), email address and profile photo, with the name and photo linking to the actual LibGuides Profile page.  However the template has access to the entire JSON data for the LibGuides account, so other info could also be displayed by a custom template.

I borrowed the connection logic (including OAuth token negotiation) from OverdriveConnector.  I did refactor it into a new reusable trait, but I didn't touch the original.

TODO
- [ ] Refactor OverdriveConnector to use the new OauthServiceTrait.
- [ ] Add configuration options to select an account based on call number facets in the results.


config/vufind/contentsecuritypolicy.ini
```diff
@@ -47,6 +47,8 @@ worker-src[] = "blob:"
 style-src[] = "'self'"
 style-src[] = "'unsafe-inline'"
 img-src[] = "'self'"
+; If you are using LibGuidesProfile recommendation module, uncomment the line below
+;img-src[] = libapps.s3.amazonaws.com
```
maccabeelevine, line 51: I notice most of the img-src here start with 'https:'.  A protocol is not required by the CSP definition, but I understand the intent to avoid loading from an insecure (http) source.  That said when I run a test instance of VuFind it's http so it would be annoying to not support that.  The actual URL loaded (coming from the API data) omits the protocol also (i.e. starts with '//') so it picks up https if the page already uses it.

demiankatz:
Looks like a good start, @maccabeelevine; let me know when you're ready for a full review. And don't forget to put in a comment in searches.ini to describe the new module. :-)

module/VuFind/src/VuFind/Connection/LibGuides.php
```diff
@@ -0,0 +1,192 @@
...
+     * @param bool                 $forceNewConnection Force a new connection (get a new token)
+     *
+     * @link https://ask.springshare.com/libguides/faq/873#api-auth
+     */
+    public function __construct(
+        $client,
+        $baseUrl,
+        $clientId,
+        $clientSecret,
+        $forceNewConnection = false
```
maccabeelevine: Yeah phpstan, I know I'm not using this yet.  I have to see in what cases the OverdriveConnector actually forced a new connection.

module/VuFind/src/VuFind/Connection/LibGuides.php
```diff
@@ -0,0 +1,192 @@
...
+        $this->baseUrl = $baseUrl;
+        $this->clientId = $clientId;
+        $this->clientSecret = $clientSecret;
+    }
+
+    /**
+     * Load all LibGuides accounts.
+     *
+     * @return array An array of all LibGuides accounts, or an empty array
+     * if an error occurs
```
maccabeelevine: I know returning an empty array is unusual.  It has the benefit of being useful, but if it's better to return null or throw an exception or something...?

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    /**
+     * Get terms related to the query.
+     *
+     * @return array
+     */
+    public function getResults()
+    {
+        // TODO: cache data, check it for staleness
```
maccabeelevine: I do see that VuFind has a nice object cache...planning to use it.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -0,0 +1,241 @@
...
+     */
+    protected function findBestMatch(\VuFindSearch\Query\QueryInterface $query)
+    {
+        $queryString = $query->getAllTerms();
+        if (!$queryString) {
+            return false;
+        }
+        $queryString = strtolower($queryString);
+
+        // Find the closest levenshtein match.
```
maccabeelevine, line 162: See comments in the PR description.  I recognize it's weird in a project reliant on Solr and its excellent searching and relevancy ranking to do something like this algorithm.  But it works surprisingly well, and fast enough in my own testing.  Thoughts?

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -0,0 +1,241 @@
...
+            if ($distance < $minDistance) {
+                $id = $this->subjectToId[$subject];
+                $minDistance = $distance;
+            }
+        }
+        if ($id == null) {
+            return false;
+        }
+
+        // // Find an exact match
```
maccabeelevine: So, this could be an alternative algorithm used via configuration?  But it's a really bad algorithm as it will almost never result in a match.  And substring match is almost as bad.

themes/bootstrap3/templates/Recommend/LibGuidesProfile.phtml
```diff
@@ -0,0 +1,14 @@
+<?php $account = $this->recommend->getResults(); ?>
+<?php if (!empty($account)): ?>
+    <div class="suggested-librarian">
+        <h4>Talk to a Subject Librarian</h4>
```
maccabeelevine: heading stolen from Villanova's implementation :)

maccabeelevine:
> Looks like a good start, @maccabeelevine; let me know when you're ready for a full review. And don't forget to put in a comment in searches.ini to describe the new module. :-)

Thanks for the reminder @demiankatz, fixed that.  This is ready for an initial review.  See all my comments of course, there are things left to do.  But I'd appreciate your input at this point on the general approach, and the light refactoring of OverdriveConnector, before I go further.

config/vufind/contentsecuritypolicy.ini
demiankatz: It's possible that all the https entries below were made because these sources are known to be HTTPS-only... or perhaps because those who made the entries didn't know protocol was optional. Either way, I see no problem with you doing things the way you are doing them.

config/vufind/searches.ini
```diff
@@ -288,6 +288,10 @@ WorkKeys = year
 ; FacetCloud:[ini section]:[ini name]
 ;       Same functionality as ExpandFacets, but with a more compact interface to
 ;       allow the display of more values.
+; LibGuidesProfile
+;       Display a single LibGuides Profile for a LibGuides user (likely a librarian)
+;       who owns a subject guide whose title most closely matches the search terms.
+;       Uses configuration in LibGuidesAPI.ini.
```
demiankatz: Is it worth also mentioning contentsecuritypolicy.ini?

config/vufind/LibGuidesAPI.ini
```diff
@@ -0,0 +1,20 @@
+; This file is used for the LibGuides *API*, which uses an
+; oauth connection mechanism.  This API is currently used
+; by the LibGuidesProfile recommendation module.
+; https://ask.springshare.com/libguides/faq/873
```
demiankatz, line 6: Very minor nitpick, but the way this comment is arranged, it makes me expect that this link will tell me about the recommendation module rather than the LibGuides API. Maybe just adding an `; API documentation:` comment above the link would make things more clear.

themes/bootstrap3/templates/Recommend/LibGuidesProfile.phtml
demiankatz: You're welcome to it, but you should make this a translation string for i18n purposes. I'd suggest something like `<?=$this->transEsc('suggested_librarian_heading')?>` with the actual text defined in en.ini.

themes/bootstrap3/templates/Recommend/LibGuidesProfile.phtml
```diff
@@ -0,0 +1,14 @@
+<?php $account = $this->recommend->getResults(); ?>
+<?php if (!empty($account)): ?>
+    <div class="suggested-librarian">
```
demiankatz: Not a big deal, but we typically use 2-space rather than 4-space indent in template files (except in a few places where php-cs-fixer gets fussy and forces adjustments).

themes/bootstrap3/templates/Recommend/LibGuidesProfile.phtml
```diff
@@ -0,0 +1,14 @@
...
+        <h4>Talk to a Subject Librarian</h4>
+        <a href="<?=$this->escapeHtmlAttr($account->profile->url)?>" target="_blank">
+            <img class="librarian-img" src="<?=$this->escapeHtmlAttr($account->profile->image->url)?>" />
```
demiankatz: Is there any chance of any of these fields being undefined, or is it safe to assume they're always there? If you want to keep the template code simple, maybe there should be some validation in the PHP code before objects get returned.

module/VuFind/src/VuFind/Connection/LibGuides.php
demiankatz: If the intention of this is to pass it to authenticateWithClientCredentials later on, you could still store it in a property and pass it along down below to make phpstan happy, and decide later whether/how to use it. Or maybe you don't need it and could just put a TODO comment here to investigate later as needed.

module/VuFind/src/VuFind/Connection/LibGuides.php
demiankatz: I guess it depends on whether or not you want to be able to differentiate between "something went wrong" and "there are no accounts" from the calling code. It might make sense to signal an error condition with a different value to allow calling code to be smarter. Since `null` and `[]` are both "falsy" values, code that doesn't care about the subtleties could do a simple `if (!$result)` check and get the same behavior either way, but then more precise checks (`is_array` or `=== null`) could be done if there was a reason to go into finer detail.

module/VuFind/src/VuFind/Connection/LibGuides.php
```diff
@@ -0,0 +1,192 @@
...
+
+        $headers = [];
+        if (
+            isset($tokenData->token_type)
+            && isset($tokenData->access_token)
+        ) {
+            $headers[] = "Authorization: {$tokenData->token_type} "
+                . $tokenData->access_token;
+        }
+        $headers[] = "User-Agent: VuFind";
```
demiankatz: I wonder if this should be configurable -- you could at least make it a class property so it's easy to override.

module/VuFind/src/VuFind/Connection/LibGuides.php
```diff
@@ -0,0 +1,192 @@
...
+
+        $this->client->setHeaders($headers);
+        $this->client->setMethod("GET");
+        $this->client->setUri(
+            $this->baseUrl . "/accounts?expand=profile,subjects"
+        );
+        try {
+            // throw new Exception('testException');
```
demiankatz: Test code that can be removed.

module/VuFind/src/VuFind/Connection/LibGuides.php
```diff
@@ -0,0 +1,192 @@
...
+            return [];
+        }
+        $body = $response->getBody();
+        $returnVal = json_decode($body);
+        $this->debug(
+            "Return from LibGuides API Call: " . print_r($returnVal, true)
+        );
+        if ($returnVal != null) {
+            if (isset($returnVal->errorCode)) {
+                // In some cases, this should be returned perhaps...
```
demiankatz, line 176: Your declared return type is an array, but the result of json_decode is an object. This seems to be a further argument against returning an empty array to indicate a problem.

module/VuFind/src/VuFind/Connection/LibGuides.php
```diff
@@ -0,0 +1,192 @@
...
+                $this->error("LibGuides Error: " . $returnVal->errorCode);
+                return $returnVal;
+            } else {
+                return $returnVal;
```
demiankatz: If you're returning the result regardless of what happens, you don't really need this else clause -- you could slightly simplify the code here.

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
```diff
@@ -0,0 +1,132 @@
...
+ * @package  Connection
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Brent Palmer <brent-palmer@icpl.org>
+ * @author   Maccabee Levine <msl321@lehigh.edu>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org
+ */
+trait OauthServiceTrait
+{
+    protected $tokenData = null;
```
demiankatz, line 56: We should put a comment on this.

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
```diff
@@ -0,0 +1,132 @@
...
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org
+ */
+
+namespace VuFind\Connection;
+
+/**
+ * Helper trait for OAuth 2.0 connections.
+ *
+ * Closely adapted from VuFind\DigitalContent\OverdriveConnector.
```
demiankatz, line 39: Maybe we should make a note somewhere (like a TODO checkbox at the top of this PR, or a JIRA ticket) to start a separate PR to use this trait in the OverdriveConnector after this work is merged. We'd just need to see if @bpalme (or another Overdrive user) is interested in helping to test it.

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
```diff
@@ -0,0 +1,132 @@
...
+     *
+     * @link https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/
+     */
+    public function authenticateWithClientCredentials(
+        $oauthUrl,
+        $clientId,
+        $clientSecret,
+        $forceNewConnection = false
+    ) {
+        $this->debug("connecting to API");
```
demiankatz: It looks like this trait has a dependency on the LoggerAwareTrait. We should probably at very least document that in the comments, or perhaps take stronger action like throwing an exception if the debug method is undefined.

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Authentiate via the OAuth Client Credentials grant type.
+     *
+     * @param string $oauthUrl           URL of thee OAuth service
+     * @param string $clientId           client_id for a client_credentials grant
+     * @param string $clientSecret       client_secret for a client_credentials grant
+     * @param bool   $forceNewConnection Force a new connection (get a new token)
+     *
+     * @return string token for the session or false
```
demiankatz: This looks like it's returning an object, not a string.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -0,0 +1,241 @@
...
+
+    /**
+     * Constructor
+     *
+     * @param array      $config Config object representing LibGuidesAPI.ini
+     * @param HttpClient $client VuFind HTTP client
+     */
+    public function __construct($config, HttpClient $client)
+    {
+        $this->libGuides = new LibGuides(
```
demiankatz: It's probably better to build the LibGuides object in the factory and pass it in, rather than building it here. In fact, the LibGuides object should probably be configured as a service like the WorldCat service, etc., so that it can be customized/overridden.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -0,0 +1,241 @@
...
+        // TODO: cache data, check it for staleness
+        $this->refreshData();
+
+        $query = $this->results->getParams()->getQuery();
+        $account = $this->findBestMatch($query);
+        return $account;
+    }
+
+    /**
+     * Find the VuFind account whose profile best matches the
```
demiankatz: Do you mean LibGuides account?

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * @param \VuFindSearch\Query\QueryInterface $query Current search query
+     *
+     * @return array LibGuides account
+     */
+    protected function findBestMatch(\VuFindSearch\Query\QueryInterface $query)
+    {
+        $queryString = $query->getAllTerms();
+        if (!$queryString) {
+            return false;
```
demiankatz, line 158: Might it make sense to have a "default librarian" for very generalized searches?

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
demiankatz: I don't see anything wrong with using levenshtein, though there are a lot of obvious situations where this isn't going to work so well -- e.g. long, complex or very specific queries will cause problems. That's why, in Villanova's case, we opted to map call number facets to subjects, so we could look at facets and make a recommendation based on the contents of the result set rather than the user input. You can always predict the range of values found in certain facet fields, but you can never predict the weird things that users will type in. :-)

As you say, this might serve as a baseline default, but I think it's going to be worth the effort to add some more configurable options.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
demiankatz: I wonder if one approach you could take would be to have a set of support methods for different algorithms, and a way to configure which ones are used and in what order. Then the module could loop through them and return the first truthy value it gets back. This would make it easy to subclass and plug in your own algorithms, and to expand the logic in the future.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -0,0 +1,241 @@
...
+        $this->idToAccount = $idToAccount;
+        $this->subjectToId = $subjectToId;
+    }
+
+    /**
+     * Return the list of facets configured to be collapsed
+     *
+     * @return array
+     */
+    public function isCollapsed()
```
demiankatz: I don't think this is used or needed -- maybe accidentally copy-pasted from somewhere else?

demiankatz (changes requested):
Thanks again, @maccabeelevine -- I can't test this, but I've given it a pretty thorough readthrough and left some comments/suggestions.

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
maccabeelevine: Agreed.  For now I added it to this PR since it seems odd to do the Jira referencing this PR before it is merged.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: The thing about levenshtein is it often works well even if they do type in something really long & complex, like "new approaches in neurobiology" would pull up a librarian with a "biology" guide over a humanities librarian, etc.    I agree through that using the result set's call number facets is a good idea if someone is willing to maintain the local configuration as Villanova did, and if the primary search will return call numbers (so for an ILS definitely, although not EDS or such) -- I added that as a TODO on the PR.

The tougher problem (IMHO) with levneshtein as you reference is that it's impossible by itself to know if you have no good matches at all -- the result distance just tells you better and worse, with no useful threshold for "good enough".  So no concept of "very generalized searches" that's detectable, or trolling searches like "show me the profile of a horrible librarian".  I discussed this with one subject librarian and she thought it was still acceptable -- assuming good intent by the patron, if they get the wrong librarian and send them a query, the librarian will just forward it on to the appropriate person; better to still show the patron that a librarian can help them.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: See comment below re "very generalized searches"

demiankatz: Well, by "very generalized searches" I really mean "blank searches that bring back everything." :-)

config/vufind/LibGuidesAPI.ini
maccabeelevine: Yup, fixed.

config/vufind/searches.ini
maccabeelevine: Yup.  Done.

module/VuFind/src/VuFind/Connection/LibGuides.php
maccabeelevine: Makes sense, I added a class property.

module/VuFind/src/VuFind/Connection/LibGuides.php
maccabeelevine: This was copied from Overdrive, but yeah.  Done.

module/VuFind/src/VuFind/Connection/LibGuides.php
maccabeelevine: Yup another odd one copied from Overdrive, but fixed.

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
maccabeelevine: Done

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
maccabeelevine: Fixed I think

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Oops, fixed.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Thanks, yes a copy/paste error.

themes/bootstrap3/templates/Recommend/LibGuidesProfile.phtml
maccabeelevine: Yup, done.

themes/bootstrap3/templates/Recommend/LibGuidesProfile.phtml
maccabeelevine: Yeah I am not sure, there is no formal documentation of the API return types.  I made the design decision (which, feel free to push back on -- I don't think it's necessarily VuFind standard practice) to push the return object directly to the template, so it was easier to customize at the theme level alone.  So I added empty() checks in the template.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
demiankatz: I agree that levenshtein can help with long searches that contain useful keywords... but the problem is searches that contain no directly matching words at all. Figuring out that a search for "Albert Einstein" is probably related to science or history is beyond its capabilities, whereas the call number approach might help there.

I also agree that the problem of trolling searches is unavoidable but not a huge problem if you assume good faith (our librarians at Villanova enjoy seeing who comes up for various searches -- e.g. "Who's the Christmas librarian? Who's the cat/dog librarian?"). And as you say, the key here is highlighting that there are humans available to help end users. Getting the most appropriate librarian most of the time is desirable, but if you miss now and then, the librarians can still work it out on their side -- they're just happy to get the user referrals.

themes/bootstrap3/templates/Recommend/LibGuidesProfile.phtml
demiankatz: I think it's fine to push the object directly to the template, as long as the template is tolerant of surprises -- so thanks for adding those extra checks.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Ah sorry, I didn't look closely enough at the context :)  I can see this both ways.  On the one hand the levenshtein is already dumb enough to return a bad match rather than no match in difficult cases.  On the other hand there really is no good match for an empty search, and I feel it would make it look like the recommendation is meaningless or random.  It might be useful to recommend something generic like "call the help desk", but our template will already have plenty of generic links for that.  So I think I prefer to skip the recommendation here.

themes/bootstrap3/templates/Recommend/LibGuidesProfile.phtml
```diff
@@ -0,0 +1,20 @@
+<?php $account = $this->recommend->getResults(); ?>
+<?php if (!empty($account) && !empty($account->profile->url) && !empty($account->first_name)): ?>
+  <div class="libguides-profile">
+    <h4><?=$this->transEsc('libguides_profile_suggestion_heading')?></h4>
+    <a href="<?=$this->escapeHtmlAttr($account->profile->url)?>" target="_blank">
+      <?php if (!empty($account->profile->image)): ?>
+        <img class="libguides-profile-img" src="<?=$this->escapeHtmlAttr($account->profile->image->url)?>" />
+      <?php endif; ?>
+      <div class="libguides-profile-name"><?=$this->escapeHtml($account->first_name) . ' ' . $this->escapeHtml($account->last_name)?></div>
```
demiankatz: I'd suggest making this `$account->last_name ?? ''` just in case that value is missing somehow.

demiankatz (changes requested):
One more very small thing...

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Ok thanks for the Albert Einstein example, now I get it.  I think we should leave the current algorithm in there as a zero-config baseline, and I will look at the call number enhancement as a separate PR.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: I like that idea in theory, but each algorithm except the last then has to be fail-able, which levenshtein isn't.  I do think levenshtein would be a good final-case algorithm if other, future, better ones fail -- such as the call-number based enhancement.

In any case I don't think this particular exact-match algorithm has any benefit over levenshtein *except* being fail-able, which it does 99% of the time.  So inclined to delete it.

module/VuFind/src/VuFind/Connection/LibGuides.php
demiankatz: Maybe it's worth a small PR to apply these trivial fixes to Overdrive so we keep things in alignment. I can do it if you like; just let me know!

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
demiankatz: That's fine, but it's certainly a possible future enhancement to have a special generic display if there's no specific recommendation found. Of course, this could easily be achieved with a custom template and the current code by adding an else clause, so maybe that's enough.

themes/bootstrap3/templates/Recommend/LibGuidesProfile.phtml
maccabeelevine: Yes good catch, fixed.

module/VuFind/src/VuFind/Connection/LibGuides.php
maccabeelevine: So it turns out OverdriveConnector never passed this boolean as true either!  That will teach me to copy/past :)  Removed altogether.

module/VuFind/src/VuFind/Connection/LibGuides.php
maccabeelevine: Agreed, changed to null.

module/VuFind/src/VuFind/Connection/LibGuides.php
maccabeelevine: Yup, changed to return null so I can use null coalescing when calling it.

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
maccabeelevine: I am new to traits and didn't realize some of the limitations when I started; it would be cleaner to just have this trait use, the LoggerAwareTrait, but traits don't do that.  It seems ugly for a trait to force another dependency on classes that include it?  Maybe I should just remove the debug() calls (copied from Overdrive) and change the errors to throw an exception?

demiankatz: Yeah, traits are very useful but they can get ugly when you run up against some of these limitations. It does seem that the debug calls might be useful in some situations. One approach could be to add a local wrapper method, like:

```
protected function oauthServiceTraitLog($level, $msg)
{
    if (method_exists($this, 'log')) {
        $this->log(...func_get_args());
    }
}
```

and then refactor the debug/error calls to be `$this->oauthServiceTraitLog('debug', ...)`, etc.

Not particularly elegant, but more bulletproof.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: I changed it to be created via factory (and related refactoring of config, client, and other objects being passed around).  I think it is now "configured as a service" but not 100%.

demiankatz: Yes, that's exactly what I had in mind. Thanks!

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
maccabeelevine: Ok I took that approach, although I can't stomach passing 'debug' as a string each time so I added helper methods.  That's even more boilerplate that makes me not love traits, but it works.

module/VuFind/src/VuFind/Connection/OauthServiceInterface.php
```diff
@@ -32,6 +32,8 @@
 /**
  * Interface for classes using OauthServiceTrait.
  *
+ * Classes which use this trait should also use LoggerAwareTrait.
```
demiankatz: Should we mention the interface here rather than the trait?

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
```diff
@@ -122,14 +124,53 @@ public function authenticateWithClientCredentials(
...
     }
+
+    /**
+     * Log a debug message, if $this->log exists.
+     *
+     * @param string $msg Log message
+     *
+     * @return void
+     */
+    protected function debugSafe($msg)
```
demiankatz: The problem with using the name debugSafe (as opposed to something that incorporates the name of this trait) is that if we use this technique in multiple traits, then they'll start to conflict with one another.

I'd suggest one of two possibilities:

1.) Rename these to incorporate the name of the trait (ugly, but collision-free)

2.) Instead of adding protected methods, since we only use these in one place, we could create anonymous functions inside authenticateWithClientCredentials -- e.g.:

```
$logSafe = function($level, $msg) {
    if (method_exists($this, 'log')) {
        $this->log(...func_get_args());
    }
};

$debugSafe = function ($msg) {
    $logSafe('err', $msg);
};

// ...etc
```

...and then you just change $this->debug to $debugSafe...

demiankatz (changes requested):
Thanks for the progress! A couple more related suggestions.

module/VuFind/src/VuFind/Connection/OauthServiceTrait.php
maccabeelevine: Ok thanks for explaning the collision, you're right of course this is just another problem with this trait not just being able to use the LoggingAwareTrait directly.  I like your anonymous function approach but I don't want to define the methods with within authenticateWithClientCredentials, since eventually some class will probably need to use one of the other three OAuth grant types and they we'd have to define those functions yet again.  So I just renamed them to avoid the collision.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: @demiankatz Question about caching.  I would like to take advantage of the ttl functionality [built into StorageInterface](https://docs.laminas.dev/laminas-cache/v3/storage/adapter/#basic-configuration-options), but if I use the existing object cache as I think is [recommended](https://vufind.org/wiki/development:architecture:caching), I'd be setting the ttl for everything else using it.  Is that correct?  If so I'd lean to just creating a separate cache, but maybe there's a reason I shouldn't?

demiankatz: If you use \VuFind\Cache\CacheTrait, it supports setting different ttl's for different items within a single cache. That's what we do in much of the other code. It might be more elegant to utilize the native built-in ttl, but this mechanism is a little more flexible.

maccabeelevine: Caching implemented.

module/VuFind/src/VuFind/Connection/OauthServiceInterface.php
maccabeelevine: Agree, done.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: I've deleted it.

maccabeelevine:
@demiankatz This is maybe actually ready now.

config/vufind/LibGuidesAPI.ini
```diff
@@ -20,3 +20,8 @@
 
 ; Base URL of the LibGuides API
 api_base_url = https://lgapi-us.libapps.com/1.2
+
+; Calls to GET (...)/accounts endpoint
+[GetAccounts]
+; Duration (seconds) to cache repsonse data.  Default is 600.
```
demiankatz: Typo: repsonse.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -177,28 +175,48 @@ protected function findBestMatch(\VuFindSearch\Query\QueryInterface $query)
...
         return $account;
     }
 
+    /**
+     * Load or retrieve from the cache the list of LibGuides accounts
+     * from the LibGuides API.
+     *
+     * @return array An array containing the idToAccount and subjectToId maps
+     */
+    protected function loadDataIfNeeded()
```
demiankatz: Nitpick, but I don't really like the method name `loadDataIfNeeded` -- that describes an internal implementation detail that the caller doesn't necessarily need to worry about. Maybe I'd call this one getLibGuidesData, since that's the one the user definitely wants to call, and then rename loadData to populateLibGuidesCache or something like that, to make it clear this is the primary entry point, and that is support for this.

module/VuFind/src/VuFind/Connection/OauthServiceInterface.php
```diff
@@ -32,7 +32,7 @@
 /**
  * Interface for classes using OauthServiceTrait.
  *
- * Classes which use this trait should also use LoggerAwareTrait.
+ * Classes which use this trait should also implement LoggerAwareInterface.
```
demiankatz: "this trait" --> "this interface"?

demiankatz (changes requested):
Definitely getting close to done, though a few (mostly nitpicky) comments.

Beyond that, the only other thing to consider is whether test coverage would be feasible. On the one hand, I suspect it would be a bit of a headache to collect all the necessary fixtures and put things together, so I'm sympathetic if you don't think it's worth the effort. On the other hand, this is something I can't personally test, so having some automated test coverage would give me greater confidence that it won't get broken by future refactoring.

module/VuFind/src/VuFind/Connection/OauthServiceInterface.php
maccabeelevine: Oops fixed

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Thank you, I hated those two names and tried various options but couldn't come up with something reasonable.  I actually had "maybeLoadData()" at one point.  Your suggestions are much better.

config/vufind/LibGuidesAPI.ini
maccabeelevine: Yup fixed.

maccabeelevine:
> Beyond that, the only other thing to consider is whether test coverage would be feasible. On the one hand, I suspect it would be a bit of a headache to collect all the necessary fixtures and put things together, so I'm sympathetic if you don't think it's worth the effort. On the other hand, this is something I can't personally test, so having some automated test coverage would give me greater confidence that it won't get broken by future refactoring.

Yes I agree it's a good idea.  Have to see which parts are effectively unit testable.

demiankatz (approved):
Thanks, @maccabeelevine, I think this is in good shape. I won't merge it yet in case you'd like to add tests here first, but if you prefer to merge this and do tests in a separate PR, that's fine too.

You might find it helpful to look at the json_log_file setting in the AbstractAPI ILS driver code (https://github.com/vufind-org/vufind/blob/c510883aa8bcdfa3639ce068d7faa8591066cfbc/module/VuFind/src/VuFind/ILS/Driver/AbstractAPI.php#L178) for a possible way to capture fixtures... and the FolioTest offers an example of how these can be used.

maccabeelevine:
Thanks.  I'm writing tests now and I can see they will be some work so I will do them separately.

PR #2979, opened 2023-07-06
Fix autocomplete/searchbox bugs; add autocomplete formatting configs

demiankatz:
The primary goal of this PR is to correct a problem where certain unusual titles can result in invalid search queries when autocomplete is used. It solves the problem by adding configurable query formatting, configured by default to convert everything into a phrase search with escaped double quotes, with an exception configured for tag search (which works differently). A few other small bugs discovered during the course of testing are also fixed here, including missing placeholder when search tabs are not active, missing support for alphabrowse autocomplete configuration, and missing autocomplete on course reserves search.

TODO
- [x] Discuss: is there a better way to configure which methods are phrase-escaped and which are not?
- [x] Discuss/test: what are the implications of this on non-Solr search backends?
- [x] Add integration test for tag search
- [x] Add integration test for combined search functionality (alphabrowse autocomplete, autocomplete query formatting, etc.)
- [x] Add autocomplete functionality on the alphabrowse search box


crhallberg:
@demiankatz, does this pass the test with the weird titles and ids?

demiankatz:
> @demiankatz, does this pass the test with the weird titles and ids?

The IDs don't matter in this case -- the autocomplete is only looking at the title text, and the search is also based on the title, so the underlying IDs have no impact one way or another. None of the existing "weird_ids.mrc" test records have particularly weird titles, except for one which includes angle brackets. Testing with that one revealed a separate pre-existing problem which I have addressed in #3003.

demiankatz:
I've made autocomplete query formatting behavior configurable through commit bfcd57650321521279a6ca99c3d656462b92c7b8 -- there are currently only two options ("phrase" and "none"), but this gives us a hook to add more if we need them in future.

In the process of adding this, I significantly reformatted the code that builds the search input to make it easier to manage (using the `htmlAttributes` view helper).

Now I'll do some testing of all the different contexts for autocomplete to see if any of the default configurations need to be changed.

demiankatz:
Commit 5fd531d1ec4f0d0462274ab4a3265e4478959acd simplifies my code and fixes a problem I discovered while testing the refactoring: placeholders didn't work when search tabs were disabled.

demiankatz:
Further testing revealed more pre-existing problems: the autocomplete setup logic was only designed to assign functionality to the first search box on the page, but the course reserves search page needs separate autocomplete on two different boxes. Commit 4b046eba7b8ab331d448afb044d1e843f05bf1b9 fixes this problem (rewriting the reserves search box along similar lines to the standard search box as part of the solution).

demiankatz:
All tests are now passing. I think this is potentially ready to merge. @crhallberg, do you mind taking another look at this?

demiankatz:
Update: a bit more testing revealed some further oversights here related to combined search boxes and alphabrowse search options. More code has been added to handle all these cases; I plan to add additional Mink tests to cover this potentially brittle code before merging the PR, but I may not have time for this work until later in the month. I still welcome reviews/feedback in the meantime.

damien-git:
[Related PR](https://github.com/vufind-org/vufind/pull/3024) with a fix to the normalizer.

crhallberg:
Do we want to fix the bug where the Autocomplete is opening on load in this PR?

demiankatz:
@crhallberg, if the fix can be made independently of these changes, I think a separate PR would keep things simpler... but if the changes here will conflict with changes you plan to make to dev, then I'm open to putting the additional changes here.

demiankatz:
I've resolved merge conflicts here, and all tests are passing. I've also opened [VUFIND-1638](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1638) to track one known issue I discovered during development of this PR. I will solve that in a follow-up PR after this one is merged.

config/vufind/Summon.ini
```diff
@@ -354,6 +354,13 @@ default_handler = None
 ; Auto-submit autocomplete on click or enter
 auto_submit = true
 
+; Query formatting rules for autocomplete, indexed by search handler name;
+; formatting_rule[*] defines a rule that is used by default if a more specific type
+; is not provided. Currently supported rules:
+; - phrase: Put the value in double quotes, and backslash-escape quites in the phrase
```
crhallberg: typo: quites

themes/bootstrap3/js/common.js
```diff
@@ -489,75 +489,112 @@ function setupOffcanvas() {
...
+        success: function autocompleteJSON(json) {
+          const highlighted = json.data.suggestions.map(
+            (item) => ({
+              text: item.replaceAll("&", "&amp;")
+                .replaceAll("<", "&lt;")
+                .replaceAll(">", "&gt;")
+                .replaceAll(query, `<b>${query}</b>`),
+              value: formattingRule === 'phrase'
+                ? '"' + item.replaceAll('"', '\\"') + '"'
+                : item,
```
crhallberg, line 578: in the future, it might be better to make each formatting change a function so that we can format more than just the value. Pass the whole item and return it with any and all changes. How to make this easy to customize will be the challenge.

crhallberg (approved):
Besides the typo I found, this looks like a robust way to fix intermittent search issues

config/vufind/Summon.ini
demiankatz: Good catch! This was actually in three different files -- I've fixed all instances.

themes/bootstrap3/js/common.js
demiankatz: Makes sense -- certainly open to future enhancements. Maybe more ideas will come to you as we de-jQuery and refactor this during 10.0 development.

demiankatz:
Thanks for the review, @crhallberg -- I'll leave this open in case @EreMaijala wants to take a look at it; otherwise, I'll merge it early in the day tomorrow so I can get to work on fixing VUFIND-1638.

EreMaijala (approved):
Looks good to me! I'd be happier to not see new jQuery code, but it's okay since it was just extending existing code.

demiankatz:
I agree, new jQuery code is not ideal... but I was fixing bugs in the context where they existed. We can presumably de-jQuery-ize everything when we reconcile this with #3064.

PR #2981, opened 2023-07-07
Handle OR operator for SortFacetList

RLangeUni:
Recently we noticed, that facet-links on a home page under [HomePage_Settings] defined as orFacets for ContentBlocks don't work. In SortFacetList the addFacet method is called without considering the operator. This should be improved with this PR.


demiankatz:
Thanks, @RLangeUni -- this looks like it might be a candidate for backporting as a bug fix for release 9.0.3. I'll take a closer look at it next week when I'm back in the office!

demiankatz (approved):
This looks reasonable to me. I'm backporting it to the release-9.0 branch.

PR #2982, opened 2023-07-07
Add translations to improve accessibility for account/list checkboxes

RLangeUni:
(new translations are only provided for de and en)

* change text for all cart entries
* use select_all_on_page in checkedout.phtml, list.phtml, bulk-action-buttons
* select_item_storage_retrieval_request_cancel in storageretrievalrequests
* use texts in myresearch aria-labels
* use select_item_[context] in checkbox.phtml

This is the "fourth PR" simple string / helper texts, derived from https://github.com/vufind-org/vufind/pull/2874.

TODO:
- [ ] After merging, add note to 9.1 language update ticket to deal with select_page cleanup.


languages/en.ini
```diff
@@ -1199,9 +1199,14 @@ seconds_abbrev = "s"
...
 see_all_ellipsis = "see all…"
+select_item_checked_out_renew = "Select item for renewing"
+select_item_hold_cancel = "Select item for canceling holds or recall"
+select_item_ill_request_cancel = "Select item for canceling interlibrary loan requests"
+select_item_storage_retrieval_request_cancel = "Select item for canceling storage retrieval requests"
 Select this record = "Select this record"
 Select your carrier = "Select your carrier"
-select_page = "Select Page"
+select_page = "Select all entries of the page"
+select_page_cart = "Select all entries of the clipboard"
```
demiankatz: I would change these to "on the page" (rather than "of the page") and "in the book bag" (rather than "of the clipboard") for consistency/clearer English.

themes/bootstrap3/templates/holds/list.phtml
```diff
@@ -28,7 +28,7 @@
           <div class="checkbox">
             <label>
               <input type="checkbox" name="selectAll" class="checkbox-select-all">
-              <?=$this->transEsc('select_page')?>
+              <?=$this->transEsc('select_all_on_page')?>
```
demiankatz, line 31: There is no select_all_on_page key in the language file. Should this remain as select_page, or should select_page be copied or renamed in the language file?

demiankatz (changes requested):
Thanks for getting this started, @RLangeUni! I noticed a couple of things that may need further attention.

demiankatz:
It also looks like the language files need to be re-ordered to meet project standards -- just run `cd $VUFIND_HOME ; php public/index.php language normalize languages` and commit the results to get the build to pass.

elsenhans:
Hi @demiankatz , let me know if there are further changes desired for this PR, I will be responsible for further editing.

demiankatz:
Thanks, @elsenhans! I will give this another review as soon as time permits (hopefully in the next day or two) and will let you know what I find.

themes/bootstrap3/templates/record/checkbox.phtml
```diff
@@ -1,5 +1,6 @@
+<?php $label = isset($this->context) ? 'select_item_' . $this->context : 'select_item'; ?>
```
demiankatz: Development of this file seems incomplete. 

1.) This references a `select_item` key which is undefined in the language files.

2.) I'm not sure where `$this->context` is coming from or if it is ever going to be set. I wonder if a new parameter needs to be added to the Record view helper's getCheckbox method to specify the context.

3.) I also wonder whether we need to have a fallback in case `$label` is set to a value that's not defined in the language files -- this could be done by specifying a default, like `$this->transEscAttr($label, [], $this->translate('select_item'))` which will use the translation for select_item if the specified $label value is not found.

demiankatz (changes requested):
I had more time to look at this today, @elsenhans. I made a couple of small changes:

1.) I adjusted the language of one of the English strings to make it more clear.

2.) I adjusted both language files so that they add a new select_all_on_page string instead of changing the existing select_page translation. I think this will give us more flexibility and will also ensure that we obtain the translations we need for other languages.

I also still have some concerns that the more verbose checkbox controls may impact the layout of the screen in a negative way, but I haven't actually started this up and looked at it yet -- I wanted to get other review changes made before doing hands-on testing.

See below for a few specific comments about one file that seems to be incompletely developed. I think once we get that sorted out, I'll be ready to try this and see if that leads to further suggestions.

Thanks again for moving this forward!

themes/bootstrap3/templates/record/checkbox.phtml
elsenhans: Hello @demiankatz ! 
Well, `$this->context` is already available in `function getCheckbox()`:
https://github.com/vufind-org/vufind/blob/f29c765ae1b982cb614ab4f413a225415f899732/module/VuFind/src/VuFind/View/Helper/Root/Record.php#L427
Depending on whether you are in the _result list_ or the _favorites list_ either the token `select_item_search_result` or `select_item_favorite` is generated. But you are right, both tokens are missing in the translation files. I will add them for de and en:
```
select_item_search_result = "Select item for mailing, exporting, printing, saving as favorite or adding to book bag"
select_item_favorite = "Select item for mailing, removing from favorites, exporting, printing or adding to book bag"
```

```
select_item_search_result = "Titel zum Mailen, Exportieren, Drucken, Speichern oder Merken auswählen"
select_item_favorite = "Titel zum Mailen, Löschen, Exportieren, Drucken oder Merken auswählen"
```

demiankatz: @elsenhans, the $context variable in the helper method that you linked to is an array of values to add to the template's context, but it does not itself make a value named context accessible to the template. Based on the existing code, it sets $this->id, $this->number, $this->prefix and possibly $this->formAttr, but the array does not contain a 'context' key, so there is no guarantee that $this->context will be set when this method is called, unless it already exists in the view object prior to the call. I'm also not sure where the search_result or favorite values you reference would be coming from. Doing a quick search of the template code, I don't see a context variable being set in a situation where the checkbox template will be included; it only seems to be used for a few narrow things like QR code display. I also suspect that using 'context' as a variable name may not be a good idea (even though there's already a precedent for doing it), because we have a view helper named context -- this may cause some confusion or ambiguity.

Of course, it's possible I'm still missing something somewhere, but this still seems incomplete to me. And if we continue down this road, we also still need to add the generic select_item translation and make sure it gets used as a default if no translation exists for a more specific context-based token, or else we may end up rendering something non-human-readable in certain situations.

elsenhans: OK, just talked to @RLangeUni , he still has to do the PR for the helper method. But it should be done in a separate PR.
So, either we wait with this PR for that one or we remove the label in the checkbox here and open another one (or should the label part of your PR @RLangeUni ?).
Or we set it in the template:
`($this->context ? $this->context . '-' : '') `
https://github.com/vufind-org/vufind/pull/2874/commits/84ab0775e2b98a820ec83e41c24ffc82e9548a92

Sorry, @demiankatz did not follow all the discussions you had already in the origin PR. Do you have any preferences?

demiankatz: @elsenhans, why don't we just remove this file (and any associated translation strings) from this PR for now? We can move those details into the other one when it is ready. That should allow us to finish up the rest of the work here in the meantime and hopefully get some more forward progress merged.

elsenhans: I removed the label and the translations and created a ticket as reminder.
For now we also could use for the aria-label just one token with the translation _"Select item for mailing, exporting, printing, saving as favorite or adding to book bag"_
Just the delete in the list of favorites would be missing, but the most important information would be already available.

demiankatz: @elsenhans, I like the idea of a single generic label for now, but I think we should use much less specific text. Every option on the list is configurable and may not be available in every VuFind instance, so listing every possible action is going to be confusing and misleading in many situations. I think it would be better to use a general text like "Select item for further action" and then people can customize this locally to match their actual situation if they wish to. Additionally, if we make the text fully generic, then it will also be usable as a default fallback for other types of checkboxes and will never be wrong. I understand that there are advantages to being very specific, but in the case of the core project, I just don't think there's a reasonable way to do it without a lot of complex logic, and that seems excessive. I'd rather be vague than explicitly wrong.

elsenhans: Thanks @demiankatz ! That's a good point with the individual configuration. 
I add your concerns to ticket for the following PR, so they are not lost and the best possible solution can be worked out.

themes/bootstrap3/templates/holds/list.phtml
```diff
@@ -70,7 +70,7 @@
             <?php endif; ?>
             <div class="checkbox">
               <label>
-                <input type="checkbox" name="selectedIDS[]" value="<?=$this->escapeHtmlAttr($ilsDetails['cancel_details'] ?? $ilsDetails['updateDetails']) ?>" id="checkbox_<?=$safeId?>" class="checkbox-select-item">
+                <input type="checkbox" name="selectedIDS[]" value="<?=$this->escapeHtmlAttr($ilsDetails['cancel_details'] ?? $ilsDetails['updateDetails']) ?>" id="checkbox_<?=$safeId?>" class="checkbox-select-item" aria-label="<?=$this->transEscAttr('select_item_hold_cancel')?>">
```
demiankatz, line 73: Is it intentional that the `select_page` in this file has not been changed to `select_all_on_page`? I'm guessing it's more likely that this was missed by accident.

demiankatz: The `select_page` token is also still present in the `checkouts/history.phtml` template, which does not appear to have received any attention yet -- possibly because it was added to the project AFTER this accessibility work. Would you like to take a look at that template and adjust in as part of this PR?

demiankatz (changes requested):
Thanks for the ongoing progress, @elsenhans. I did another more thorough review and noticed a couple of places where the old `select_page` translation has not been replaced yet -- see comments below.

I also took time to capture some screen shots of how this impacts various parts of the visual interface. I'm sharing them here so we can get some opinions from others about whether or not this causes problems.

BOOK BAG

Before:

image

After:

image

SEARCH RESULTS WITH BULK OPTIONS

Before:

image

After:

image

FAVORITES

Before:

image

After:

image

CHECKED OUT ITEMS

Before:

image

After:

image

...for the most part, I think this looks fine with the longer text, but the button wrapping in the book bag lightbox is perhaps not ideal. Might it make sense to set things up so that the checkbox and button list wrap to separate lines if they are too large to fit together, rather than having the buttons partially wrapped? I'd be interested in opinions from @crhallberg, @sturkel89 and others...

themes/bootstrap3/templates/holds/list.phtml
elsenhans: Thanks @demiankatz for the hints!
I will check the changes in this PR with the original requirements.

elsenhans: @demiankatz  I changed both tokens to `select_all_on_page`. 
All the others seem to be correct.

demiankatz: Thanks! I think we also need an aria-label added to checkboxes in the new checkouts/history.phtml, probably with a new translation string like `select_item_purge = "Select item for purging"`.

elsenhans: Right. Added that aria-label. Thank you!

demiankatz:
Thanks, @elsenhans -- I'm now just awaiting feedback from others in order to determine if any further action is needed before we merge this. I'll be sure to discuss it on next month's Community Call if we don't reach a resolution sooner than that. From my personal perspective, the only thing that might need more attention is the layout/wrapping of the cart controls. Do you have any opinion about that?

elsenhans:
I think the division into two lines is good.
For a uniform representation, I would suggest to take over the division into two lines at all occurring places.
image

crhallberg (approved):

crhallberg:
I like the two lines approach, although I think the checkbox would be better below the buttons, closer to the items. Another option is to widen the lightbox, since it looks like this issue will only occur in the lightbox.

crhallberg (approved):
No concerns beyond the book bag wrapping issue.

demiankatz:
> I like the two lines approach, although I think the checkbox would be better below the buttons, closer to the items. Another option is to widen the lightbox, since it looks like this issue will only occur in the lightbox.

Is there responsive markup we could use for all of these controls so that they display side by side when space permits and the checkbox drops below when necessary? Seems like something that should be possible.

Is the lightbox currently a fixed width? It seems like it would probably make sense to give it more room, though we'd have to be sure there were no unexpected side effects. That's probably a task for a separate PR, though!

elsenhans:
Widen the lightbox seems to be a good solution for me too. We are going to do that in a separate PR right?

demiankatz:
I agree, @elsenhans, if we want to widen the lightbox, we should probably do that in a separate PR. I think the only outstanding issue here is whether there's markup @crhallberg can recommend to help with the flow of the checkbox and button elements.

elsenhans:
@demiankatz @crhallberg Can we please name the open work and finish it either here or a new PR? There are other PRs waiting for this one to be completed.

One task for a new PR is to widen the lightbox.
What else needs to be done?

demiankatz:
@elsenhans, I agree that a separate PR to enlarge the lightbox is a good idea. I think the main other task is to add responsive markup to make the layout of checkboxes and buttons tidier -- that's the change I'm waiting for before approving and merging this. I did ask @crhallberg yesterday to please weigh in on this -- but I think he's still catching up on things following WOLFcon last week. :-)

crhallberg:
My mistake! I thought the screenshots were already marked up.

themes/bootstrap3/templates/cart/cart.phtml
```diff
@@ -14,7 +14,7 @@
       <div class="checkbox pull-left flip">
```
crhallberg: Removing "pull-left flip" from here would probably cause a wrap. But in case it doesn't...

themes/bootstrap3/templates/checkouts/history.phtml
```diff
@@ -47,7 +47,7 @@
           <?php if ($purgeSelectedAllowed): ?>
             <label>
               <input type="checkbox" name="selectAll" class="checkbox-select-all"/>
-              <?=$this->transEsc('select_page')?>
+              <?=$this->transEsc('select_all_on_page')?>
             </label>
```
crhallberg, line 51: It looks like the .checkbox on line 46 can be closed here instead of line... actually it may not be closed...

themes/bootstrap3/templates/myresearch/bulk-action-buttons.phtml
```diff
@@ -6,7 +6,7 @@
 <nav class="bulkActionButtons">
```
crhallberg, line 6: .bulkActionButtons may have float or flex stylings that need to be broken for a wrap.

crhallberg (dismissed):
Added comments on where to cause two-line wraps for the better, longer labels.

themes/bootstrap3/templates/cart/cart.phtml
demiankatz: Removing pull-left flip doesn't seem to make any difference. How would you feel about actually adding an extra `<div class="pull-left flip">` around all of the buttons below the checkbox? This allows the checkbox and buttons to coexist when there is room, but it wraps all the buttons together when there is not enough room. The only disadvantage is it wraps the checkbox ABOVE the buttons, and I think we may have discussed a preference for wrapping the checkbox BELOW the buttons.

demiankatz: Another option to consider: the existing search/bulk-action-buttons.phtml template wraps as-is, because it's coded as a checkbox div followed by a button ul. Would it make sense to re-code the cart buttons as a ul for consistency? Is there any reason not to do that? Again, this has the same issue of not wrapping the checkbox below the buttons, but maybe it makes sense to just retain current behavior and make the template code more consistent across the board.

themes/bootstrap3/templates/myresearch/bulk-action-buttons.phtml
demiankatz: This already wraps as expected -- and this template is coded consistently with `search/bulk-action-buttons.phtml`. So the real question is, why is the cart different, as noted above?

themes/bootstrap3/templates/checkouts/history.phtml
demiankatz: The checkbox div is closed, but it is wrapped around not just the checkbox but also all of the buttons. This is consistent with the markup style in holds/list.phtml, myresearch/checkedout.phtml, myresearch/illrequests.phtml and myresearch/storageretrievalrequests.phtml as well. As a result, buttons do not wrap as a group, which is inconsistent with the desired behavior we're working on elsewhere. However, because these templates have comparatively few buttons, it is less likely to cause problems in these contexts. It is probably worth making our checkbox+button markup consistent everywhere, but I think fixing this immediately is less critical.

demiankatz:
Thanks for the input, @crhallberg -- see my replies to your comments above. I think we have at least three approaches here:

1.) If it is really important to us to get the checkbox to wrap BELOW the buttons, instead of above the buttons, I think we'll have to do a lot of work on all of the templates referenced above. I'm not sure if that's worth the effort, though it would be nice to have.

2.) If we want to make all of our markup consistent, perhaps we should consider adopting the div + ul approach for all checkbox + button combinations. (Though if a different approach is more accessible, I'm certainly open to other options...). This is a moderate amount of work, especially if it causes any complications for the various myresearch templates that are currently implemented without a lot of structured markup around the buttons.

3.) If we want a bare minimum solution to fix the most serious visual problems and get this PR merged, maybe adjusting the cart template is all that's absolutely necessary (since I really think the long checkbox text in the cart looks bad without wrapping, but the other templates either already wrap or have enough space that wrapping is unlikely). We could then do follow-up work on option 1 or 2 as a follow-up in a separate PR, if we feel it would be beneficial.

What do you think, @crhallberg and @elsenhans? Any preferences or further ideas?

crhallberg:
I'm going to get to implementing this tomorrow.

crhallberg:
@RLangeUni, could you give me permission to push to this branch? That may be simpler than creating a separate pull request.

demiankatz:
To save @RLangeUni some time, I've gone ahead and pushed @crhallberg's latest changes to this branch. Thanks, @crhallberg!

demiankatz (approved):
This looks good to me now! There's one small inconsistency at present: the "select all" checkbox wraps ABOVE the buttons in the cart lightbox, and BELOW the buttons everywhere else. However, this will be fixed soon in a follow-up PR to make the lightbox window larger, as previously discussed, so I don't mind having it momentarily inconsistent in dev for a couple of days while we wait for that to be finalized.

PR #2983, opened 2023-07-07
FOLIO: Remove empty holdings statements

damien-git:
FOLIO can have an empty object (no key/value) in the holdingsStatements/holdingsStatementsForSupplements/holdingsStatementsForIndexes arrays. It probably shouldn't, but the result in VuFind is empty entries for things like Supplements in the holdings. We can filter out empty strings in the Folio driver to remove them.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/Folio.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/Folio.php
@@ -608,18 +608,18 @@ class Folio extends AbstractAPI implements
             array_map([$this, 'formatNote'], $holding->notes ?? [])
         );
         $hasHoldingNotes = !empty(implode($holdingNotes));
-        $holdingsStatements = array_map(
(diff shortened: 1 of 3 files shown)


demiankatz:
Thanks, @damien-git -- seems like a reasonable fix. Probably a good idea to add a test fixture to demonstrate this issue to prevent future regressions. Would you be able to do that, or would you like my help with it?

demiankatz (approved):
Thanks again, @damien-git -- I had a few moments free, so I went ahead and added a test for this functionality. I made a couple of small adjustments to help ensure consistent results in the test.

I was going to investigate whether this fix could be backported for the 9.0.3 release, but since this builds upon changes that were subsequently made in the dev branch for release 9.1, I don't think it's worth the effort to backport and have slated this for release 9.1.

damien-git:
OK, thanks.

PR #2984, opened 2023-07-07
Test LibGuides connector and LibGuidesProfile

maccabeelevine:
Test LibGuides connector and the LibGuidesProfile recommendation module, both added in #2977.

### DIFF
--- a/module/VuFind/tests/fixtures/libguides/api/accounts
+++ b/module/VuFind/tests/fixtures/libguides/api/accounts
@@ -1,0 +1,147 @@
+HTTP/1.1 200 OK
+Server: nginx
+Date: Wed, 05 Jul 2023 19:50:05 GMT
+Content-Type: application/json
+
+[
+    {
+        "id": 1234,
+        "email": "melvil@dewey.edu",
+        "uuid": "19193969-097a-4f4f-b17b-c2b402af24a7",
+        "login_fail_cnt": 1,
+        "first_name": "Melvil",
+        "last_name": "Dewey",
+        "title": "Humanities Librarian",
+        "nickname": "Mel",
+        "signature": "",
+        "created_by": 7890,
+        "created": "1814-03-21 15:57:20",
+        "updated": "1821-08-23 19:07:14",
+        "profile": {
+            "id": 4321,
+            "account_id": 1234,
+            "customer_id": 999,
+            "public_id": "c11cd6c1-9d3b-44c9-901a-526965b85dd7",
+            "box": {
+                "title": "Cataloging Librarian"
+            },
+            "pronouns": "he / him / his",
+            "image": {
+                "show": 1,
+                "file": "Melvil_Dewey.jpg",
+                "url": "//en.wikipedia.org/wiki/Melvil_Dewey#/media/File:Melvil_Dewey.jpg"
+            },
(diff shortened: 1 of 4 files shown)


module/VuFind/tests/fixtures/libguides/api/accounts
```diff
@@ -0,0 +1,147 @@
...
+Date: Wed, 05 Jul 2023 19:50:05 GMT
+Content-Type: application/json
+
+[
+    {
+        "id": 1234,
+        "email": "melvil@dewey.edu",
+        "uuid": "19193969-097a-4f4f-b17b-c2b402af24a7",
+        "login_fail_cnt": 1,
+        "first_name": "Melvil",
```
maccabeelevine, line 12: I had perhaps too much fun with this fixture data, but I learned some things!

module/VuFind/tests/unit-tests/src/VuFindTest/Connection/LibGuidesTest.php
```diff
@@ -0,0 +1,104 @@
...
+class LibGuidesTest extends \PHPUnit\Framework\TestCase
+{
+    use \VuFindTest\Feature\FixtureTrait;
+
+    /**
+     * Test loading accounts.
+     *
+     * @return void
+     */
+    public function testGetAccounts()
```
maccabeelevine, line 57: I am honestly not sure if this test class has a point, since all I'm doing is parsing the fixture.  I guess parsing the fixture is what it's testing.

module/VuFind/tests/unit-tests/src/VuFindTest/Recommend/LibGuidesProfileTest.php
```diff
@@ -0,0 +1,185 @@
...
+
+        // Mock config and caching logic in LibGuidesProfile.
+        // Caching is from a trait, which is not the point of this test suite,
+        // and the config is only used in LibGuidesProfile for caching.
+        $this->config = $this->getMockBuilder(Config::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $this->cacheAdapter = $this->getMockBuilder(CacheAdapter::class)->getMock();
+
+        // For the target class LibGuidesProfile, only mock the caching methods
```
maccabeelevine, line 107: It's probably unusual to mock (some of) the very class we're testing, but I think the reasoning is sound?

maccabeelevine:
FYI I will be on vacation most of next week, so will probably get to revisions the week after.

module/VuFind/tests/unit-tests/src/VuFindTest/Recommend/LibGuidesProfileTest.php
demiankatz: It's actually not uncommon to do this sort of thing, especially for classes that interact with the file system or database. In this particular case, there is a "Memory" cache adapter that stores data in-memory and is good for testing, so you _could_ use that instead of mocking the cache methods... but this is fine as is. I'll merge this in its current form, and if you feel like changing it in the future, we can always do a follow-up PR.

demiankatz (approved):
Thanks, @maccabeelevine! Have a great vacation!

PR #2986, opened 2023-07-10
Shibboleth: clean up comments with use statement.

demiankatz:
I happened to notice that the Shibboleth auth module contained an unused use statement, but using the use statement allowed simplification of many comments. This PR applies that simplification for readability. It should have no impact on the functionality of the code.


xmorave2 (approved):

demiankatz:
Thanks, @xmorave2!

PR #2987, opened 2023-07-10
Add simulated single sign-on for testing purposes.

demiankatz:
This pull request creates a new authentication module and controller which allow VuFind to simulate a single sign-on process for testing purposes. When enabled, clicking on the "institutional login" link will automatically log the user in with the username fakeuser1. This could potentially be extended to make the username/user attributes configurable, but I kept things simple for now.

This functionality could be used to build automated tests that exercise VuFind's single sign-on workflows. It could also be useful for developers who want to try these workflows without access to a real SSO system.

I hope to make use of this to test #2941 in my local environment.


module/VuFind/src/VuFind/Auth/SimulatedSSO.php
```diff
@@ -0,0 +1,97 @@
...
+     * Session initiator URL callback
+     *
+     * @var callable
+     */
+    protected $getSessionInitiatorCallback;
+
+    /**
+     * Constructor
+     *
+     * @param string $url Session initiator URL callback
```
xmorave2: this should be callable, not string

xmorave2: The type of `$url` parameter should be  `callable`, not  `string`.

module/VuFind/src/VuFind/Auth/SimulatedSSO.php
```diff
@@ -0,0 +1,97 @@
...
+     */
+    public function __construct($url)
+    {
+        $this->getSessionInitiatorCallback = $url;
+    }
+
+    /**
+     * Attempt to authenticate the current user.  Throws exception if login fails.
```
xmorave2: There is double space in comment.

module/VuFind/src/VuFind/Auth/SimulatedSSO.php
```diff
@@ -0,0 +1,97 @@
...
+        $user->email = 'fake@example.com';
+
+        // Save and return the user object:
+        $user->save();
+        return $user;
+    }
+
+    /**
+     * Get the URL to establish a session (needed when the internal VuFind login
+     * form is inadequate).  Returns false when no session initiator is needed.
```
xmorave2: There is double space in comment.

xmorave2 (changes requested):
Working OK. Code is looking good, I just found only few minor issues.

demiankatz:
Thanks, @xmorave2! I've corrected the issues you highlighted, though those double-space-in-comment issues are probably present in some or all of the other authentication modules, since I just copied and pasted the existing comments from Shibboleth when I built this. I'll open up a follow-up PR to fix that.

xmorave2 (approved):

PR #2989, opened 2023-07-11
Add missing container dependency to ChannelsController.

demiankatz:
#2941 revealed a bug in the ChannelsController: even though it extends the AbstractBase controller, it does not inject the top-level service manager that is required/expected by the parent class. This can lead to unexpected failures and problems. This PR adds the missing dependency. This could be viewed as a minor BC break, but I think the fact that it fixes potential bugs overrides that concern and makes it worth including in release 9.0.3. I'm open to discussing this, however!

TODO
- [ ] Add note to changelog

### DIFF
--- a/module/VuFind/src/VuFind/Controller/ChannelsController.php
+++ b/module/VuFind/src/VuFind/Controller/ChannelsController.php
@@ -29,6 +29,7 @@
 
 namespace VuFind\Controller;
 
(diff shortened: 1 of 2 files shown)


LuomaJuha (approved):
Seems good! I think the bc in this pr is something which should be fixed and so the bc is desired.

PR #2994, opened 2023-07-17
Search form reset button

xmorave2:
There are still some issues to discuss I think (like: should this be configured to show?). But I would like to have some feedback.


demiankatz:
@xmorave2, your styling seems to be having a side effect on the language control (at least in the bootprint3 theme; I haven't looked at others yet).

Before:

image

After:

image

themes/bootstrap3/templates/search/searchbox.phtml
```diff
@@ -70,7 +70,10 @@
...
+    <div class="searchForm-query">
+      <input id="searchForm_lookfor" class="searchForm_lookfor form-control search-query<?php if ($this->searchbox()->autocompleteEnabled($this->searchClassId)):?> autocomplete searcher:<?=$this->escapeHtmlAttr($this->searchClassId) ?><?=$this->searchbox()->autocompleteAutoSubmit($this->searchClassId) ? ' ac-auto-submit' : '' ?><?php endif ?>" type="search" name="lookfor" value="<?=$this->escapeHtmlAttr($this->lookfor)?>"<?php if ($placeholder): ?> placeholder="<?=$this->transEscAttr($placeholder) ?>"<?php endif ?> aria-label="<?=$this->transEscAttr("search_terms")?>">
+      <button id="searchForm-reset" class="searchForm-reset form-control" type="reset" tabindex="-1" aria-label="<?=$this->transEsc('searchform-reset-button')?>">&times;</button>
```
demiankatz: Looks like the searchform-reset-button needs to be added to the language files.

I also wonder if perhaps it would make sense to use the icon helper here instead of hard-coding a `&times;` entity.

themes/bootstrap3/templates/search/searchbox.phtml
```diff
@@ -163,3 +166,29 @@
   </form>
 <?php endif; ?>
 
+<?php
+$js = <<<JS
+        $(document).ready(function searchFormResetHandler() {
```
demiankatz: Would this make more sense in common.js rather than coded into the template? If you use the .js file you can better take advantage of caching.

demiankatz (changes requested):
Thanks, @xmorave2! I'd be interested in input from others (like @crhallberg), but from my perspective, this seems like a pretty clear and unintrusive addition that probably does not need to be made configurable as long as it passes accessibility review. In addition to the minor problem noted above, I've left a few other comments below.

xmorave2:
Thank @demiankatz, I (hopefully) addressed all issuses.

themes/bootstrap3/js/common.js
```diff
@@ -696,3 +696,25 @@ $(function commonDocReady() {
...
+      resetButton.hide();
+    } else {
+      resetButton.show();
+    }
+  });
+  resetButton.on('click', function onClick() {
+    queryInput.attr('value', '').focus();
+    resetButton.hide();
+  });
+});
```
demiankatz: We try to have an end-of-file line break in all of our files for style purposes. Probably worth adding one back here.

themes/bootstrap3/templates/search/searchbox.phtml
```diff
@@ -70,7 +70,10 @@
...
+    <div class="searchForm-query">
+      <input id="searchForm_lookfor" class="searchForm_lookfor form-control search-query<?php if ($this->searchbox()->autocompleteEnabled($this->searchClassId)):?> autocomplete searcher:<?=$this->escapeHtmlAttr($this->searchClassId) ?><?=$this->searchbox()->autocompleteAutoSubmit($this->searchClassId) ? ' ac-auto-submit' : '' ?><?php endif ?>" type="search" name="lookfor" value="<?=$this->escapeHtmlAttr($this->lookfor)?>"<?php if ($placeholder): ?> placeholder="<?=$this->transEscAttr($placeholder) ?>"<?php endif ?> aria-label="<?=$this->transEscAttr("search_terms")?>">
+      <button id="searchForm-reset" class="searchForm-reset" type="reset" tabindex="-1" aria-label="<?=$this->transEsc('searchform_reset_button')?>"><?=$this->icon('ui-remove');?></button>
```
demiankatz: It might be worth considering creating a new icon name like `search-reset` and then making it an alias of ui-remove by default, so people can customize it independently if they wish to. See [this example](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/theme.config.php#L250) of an icon alias if you don't already know how to set them up.

demiankatz (changes requested):
This is looking better, but I'm seeing a new style issue (at least in Firefox -- I didn't try other browsers). In all three provided themes, the search box and type drop down are no longer aligned with each other. Screen shots are from bootprint3, but the other two are similar.

Before (dev branch):

image

After (this branch):

image

See also a few very minor suggestions below.

I've put a 9.1 milestone on this, since I'm pretty confident we can get it polished before that release, and this will encourage discussion on the next Community Call. I'd also like to get a review from someone more UI-oriented than myself before merging it, but I won't request that until we finish fixing the minor issues from my own reviews, to avoid any redundant suggestions.

themes/bootstrap3/templates/search/searchbox.phtml
xmorave2: Good point, I've added the alias icon.

themes/bootstrap3/js/common.js
xmorave2: Line break added

xmorave2:
@demiankatz: Another round done ;) I incorporated your suggestions and fixed the visual glitch.

demiankatz (changes requested):
Thanks, @xmorave2 -- this is looking much better to me. Just one remaining problem that I can see: the clear icon is nearly invisible in the sandal theme:

image

We may need some sandal-specific styling to adjust for this problem.

That detail aside, though, I think this is ready for input from others, so I'm going to request a couple more reviews -- I'm interested in @ckaz's thoughts on accessibility and @crhallberg's comments on the specific styles and Javascript approach.

xmorave2:
I fixed the color in sandal theme

demiankatz:
Thanks, @xmorave2! Looks good to me now. I also merged dev into this branch to resolve minor conflicts caused by other recent style changes. This has my personal approval, but I'll wait to mark it as approved until we receive feedback from others.

crhallberg:
I'm not seeing any evidence of this interfering with [the clear button that some browsers add to type=search inputs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#differences_between_search_and_text_types), which is good, but that's my only concern.

demiankatz:
Thanks, @crhallberg, that's a good point! I'm glad that we're not seeing any problems related to this behavior, but do we understand why? I don't see an obvious reason why the browser wouldn't try to add its own clear button to this input, so why aren't we seeing double in some cases? I notice that Chrome isn't adding a clear button on the existing VuFind search box (e.g. at [the public demo](https://vufind.org/demo/)). Why not? It might be good to see if we can force the pseudo-element to appear so we can be sure that there is no edge case where it's going to cause problems.

I also found [this StackOverflow thread](https://stackoverflow.com/questions/20804016/editing-input-type-search-pseudo-element-button-x) which talks about styling the Webkit pseudo-element. If nothing else, maybe we should explicitly hide that in our styles if we think there's any chance we might encounter it. Or maybe we should try to leverage the native function where available, though I personally prefer not to get into supporting features that only work in certain browsers. @xmorave2, what do you think about all this?

xmorave2:
It is not interfering with standard reset button in webkit/chrome based browsers, because normalize from bootstrap is hiding it. See https://github.com/twbs/bootstrap/blob/v3.4.1/less/normalize.less#L367

I would be happy to use the standard reset button provided by browser, but Firefox is still not implementing it.

demiankatz (approved):
Thanks, @xmorave2! Since we have a clear explanation for why the browser-provided button is hidden, I have no concerns about conflicts arising from the new functionality, and I think this is the best option for the present since not all browsers offer a standard implementation. I will merge this now!

ckaz:
Sorry, I've been away on holiday leave. I have no issues with this PR -- looks perfect to me.

PR #2995, opened 2023-07-17
Optimized batch_import_marc

damien-git:
Currently when imported files are small there is a large overhead when the JVM is loaded and SolrMarc prepares to process files. But SolrMarc supports taking multiple files as input.
This PR changes batch-import-marc.sh and import-marc.sh so that several files can be sent at once (set to 10 by default). When files only contain a few records (which happens for instance with incremental harvests), performance is much better.

TODO
- [x] Add equivalent Windows .bat support, if possible
- [x] Update changelog and relevant documentation when merging


demiankatz:
Thanks, @damien-git, that's a great idea; I did not realize that SolrMarc supported this behavior. I'm adding a TODO item to port these changes to the .bat versions of the files if possible, and I'll look into doing that when time permits.

harvest/batch-import-marc.sh
```diff
@@ -109,20 +116,33 @@ then
...
 else
   function log {
-    local FILE=$1
-    cat -u - > $BASEPATH/log/`basename $FILE`.log
+    local FILES=$@
+    local LOGFILE
+    if [ $MAX_BATCH_COUNT -eq "1" ]
+    then
+      LOGFILE=$BASEPATH/log/`basename $FILES`.log
+      > $LOGFILE
```
demiankatz, line 131: Did the `cat -u - ` get dropped here by accident?

harvest/batch-import-marc.sh
```diff
@@ -109,20 +116,33 @@ then
...
+    else
+      TIMESTAMP=$( date +%Y%m%d%H%M%S )
+      LOGFILE=$BASEPATH/log/$TIMESTAMP.log
```
demiankatz: Do you think it might be worth including the first filename as part of the log, perhaps in addition to the timestamp, to help with finding things? I certainly don't insist on this, but I think it might be easier to narrow down the appropriate log file if the first filename from the batch is included than if it's purely timestamp based (though I guess you can `grep` it easily enough in either case if you know how things are structured).

harvest/batch-import-marc.sh
```diff
@@ -109,20 +116,33 @@ then
   }
 else
   function log {
-    local FILE=$1
-    cat -u - > $BASEPATH/log/`basename $FILE`.log
+    local FILES=$@
+    local LOGFILE
+    if [ $MAX_BATCH_COUNT -eq "1" ]
```
demiankatz: Is it worth adding validation to ensure that $MAX_BATCH_COUNT is a positive integer? Can passing in other types of values cause weird behaviors in the current version of the script?

demiankatz (changes requested):
Thanks again, @damien-git. I haven't actually tried this out yet, but I had time to give it a closer look today. I found one possible bug and had two suggestions that might or might not be helpful -- let me know what you think about them, but there's no pressure to implement if you don't think they're worth the effort.

Once you've had a chance to look at this, I'll see about working on the Windows port of the new functionality. I suspect that adding the -x functionality to batch-import-marc.bat may be more effort than it's worth due to the limitations of batch files (in which case I can just add an error message if somebody tries to pass -x, and we can spend more time if a Windows user actually cares), but I can probably at least add the multiple file support to marc-import.bat.

harvest/batch-import-marc.sh
damien-git: No, cat -u - happens after the if. "> $LOGFILE" is just resetting it in case the log file exists.

harvest/batch-import-marc.sh
damien-git: I was thinking of using grep, with -l or -m to avoid scanning a lot of large log files. But I would be OK with using the first filename to begin the log file name.

harvest/batch-import-marc.sh
damien-git: It would be an improvement, that variable should really be a positive integer.

harvest/batch-import-marc.sh
demiankatz: Oh, interesting, I've never seen that syntax before! Thanks for the clarification.

harvest/batch-import-marc.sh
demiankatz: Sounds good. Is this something you plan to work on, or do you need me to look into it when time permits?

damien-git: I can do it.

import-marc.sh
```diff
@@ -158,7 +161,7 @@ fi
 # Execute Importer
 #####################################################
 
-RUN_CMD="$JAVA $INDEX_OPTIONS -Duser.timezone=UTC -Dlog4j.configuration=file://$LOG4J_CONFIG $EXTRA_SOLRMARC_SETTINGS -jar $JAR_FILE $PROPERTIES_FILE -solrj $SOLRJ_DIR -lib_local "$VUFIND_HOME/import/lib_local\;$VUFIND_HOME/solr/vendor/modules/analysis-extras/lib" $MARC_PATH/$MARC_FILE"
+RUN_CMD="$JAVA $INDEX_OPTIONS -Duser.timezone=UTC -Dlog4j.configuration=file://$LOG4J_CONFIG $EXTRA_SOLRMARC_SETTINGS -jar $JAR_FILE $PROPERTIES_FILE -solrj $SOLRJ_DIR -lib_local "$VUFIND_HOME/import/lib_local\;$VUFIND_HOME/solr/vendor/modules/analysis-extras/lib" $NORMALIZED_PATHS"
 echo "Now Importing $1 ..."
```
demiankatz: Should we replace `$1` with `$NORMALIZED_PATHS` or `$*` here?

harvest/batch-import-marc.sh
demiankatz: Thanks! Meanwhile, I've pushed up the import-marc.bat changes to match import-marc.sh, and I left a comment about something I noticed during that process. :-)

import-marc.sh
damien-git: Yes.

harvest/batch-import-marc.sh
```diff
@@ -70,6 +77,13 @@ then
   exit 1
 fi
 
+# Check MAX_BATCH_COUNT is a positive integer
+if ! [[ "$MAX_BATCH_COUNT" =~ ^[1-9][0-9]*$ ]]
```
demiankatz, line 81: @damien-git, I adjusted this regex since it was still possible to say `-x 0`, which led to an xargs error instead of a more meaningful message. Obviously this will prevent someone from doing something like `-x 0110` which might technically be legal but doesn't really make sense, so I'm not too worried about it.

damien-git: OK

demiankatz:
@damien-git, I've done a little more work on the Windows equivalent functionality -- for now, if you pass a -x flag to the batch-import-marc.bat file, it fails with a message that the feature is not supported in Windows. I suspect that this _could_ be implemented -- probably by using a counter and stringing together the filenames inside the main loop -- but I'm not sure that it's worth the effort to bother doing the work, since I don't think we have many Windows users. Unless anyone objects, I'll just leave it at this until somebody complains.

Is there any other work you want to do on this, or is it now ready to merge from your perspective?

demiankatz:
I've opened [VUFIND-1626](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1626) to track the missing Windows functionality, so people can comment/vote there if they want it. For good measure, I've also linked to the ticket from the Windows error message, so if anyone encounters the situation, they'll know how to report their needs.

damien-git:
> Is there any other work you want to do on this, or is it now ready to merge from your perspective?

I think it's ready to merge.

demiankatz (approved):
Thanks again, @damien-git, this is a great improvement. Merging now!

PR #2996, opened 2023-07-18
fix: jstree elements overflow text properly with ellipses.

crhallberg:
Fixes [VUFIND-1621](https://openlibraryfoundation.atlassian.net/jira/software/c/projects/VUFIND/issues/VUFIND-1621).


crhallberg:
Will fix the CI tomorrow.

demiankatz:
Thanks, @crhallberg! I've rebased this onto the release-9.0 branch so we can merge it into release 9.0.3 if deemed appropriate. I also ran lessToSass to fix the CI build. I'm going to test it now.

One question, though: why are the same styles defined in both bootstrap-custom.css AND the less/sass files? Does it really need to be in both places, and if so, why?

demiankatz (changes requested):
I just tried this out, and it's an improvement but not quite perfect yet. If you turn on exclude facets, long text still pushes the "exclude" control off the right side of the facet list. Here's what it looks like in bootstrap3:

image

In sandal, you can see part of the X, but not the whole thing.

In that screenshot, you can see that I have `orFacets = *` and `exclude = *` turned on... but even if you turn off orFacets, the exclude behavior is still broken.

Can you take a look at this detail?

demiankatz:
Also, for comparison purposes, I just spun up an instance of version 8.1 and confirmed that the ellipses appear and the exclude control is correctly aligned there (though it seems to have some other visual issues with borders):

image

crhallberg:
> Why are the same styles defined in both bootstrap-custom.css AND the less/sass files?

It's so people not using LESS or SASS can still get the styles. There may be a way to import this into the LESS and dedupe this. I'd also be okay with rooting out bootstrap-custom.css altogether.

demiankatz (approved):
Looks much better to me. Thanks, @crhallberg!

PR #2997, opened 2023-07-19
Remove bootstrap-custom.css

crhallberg:
This file was supposed to provide developers who do not use LESS and SASS with the changes VuFind makes to base Bootstrap. However, this doesn't seem to be a well-utilized or well-documented use case and the file has fallen well out of step with the LESS and SASS changes.


demiankatz (approved):
Thanks, @crhallberg! I've actually rebased this onto dev since it's not a bug fix, and I don't think it's necessary (but might be confusing) to make this change in a minor release... I'll add a note about it to the 9.1 changelog.

PR #2999, opened 2023-07-21
Get unique html element id from record view helper

RLangeUni:
According to https://github.com/vufind-org/vufind/pull/2874#issuecomment-1625347871, this PR is the foundation for using unique ids to reference to from other tags with "aria-describedby" or similar.

$idPrefix could be the modal or the context ('list-entry', 'holds', 'illrequests', 'storage-retrieval-requests', 'search-result'). Perhaps a second parameter in getUniqueHtmlElementId could be added to differentiate between context and prefix?

TODO:
- [ ] After merging, open follow-up PR to apply new helper methods in appropriate places; see discussion on #2874.


module/VuFind/src/VuFind/View/Helper/Root/Record.php
```diff
@@ -707,4 +706,33 @@ protected function deduplicateLinks($links)
...
+    }
+
+    /**
+     * Get the source identifier + unique id of the record
+     *
+     * @return string
+     */
+    public function getUniqueIdWithSourcePrefix()
+    {
+        if ($this->driver) {
```
demiankatz, line 889: This function declares that it returns a string, but it can potentially complete without returning a value at all. Maybe it should throw an exception if `$this->driver` is undefined.

module/VuFind/src/VuFind/View/Helper/Root/Record.php
```diff
@@ -415,8 +415,7 @@ public function getSearchResult($view)
      */
     public function getCheckbox($idPrefix = '', $formAttr = false, $number = null)
     {
-        $id = $this->driver->getSourceIdentifier() . '|'
-            . $this->driver->getUniqueId();
+        $id = $this->getUniqueHtmlElementId($idPrefix);
```
demiankatz: I don't think this is working correctly, either in the existing dev code or in this PR. The `$id` value is the record identifier intended to be used as the VALUE of the checkbox. The `$idPrefix` value is a prefix for the HTML ID of the checkbox, but currently, the record/checkbox.phtml template does not set any HTML IDs at all or make use of the `'prefix'` value in the context! We definitely do NOT want to prefix the `$id` value, since that is a reference to a specific record in VuFind, and prefixing it will break functionality related to it. We probably DO want to add an HTML element ID, but we'll need to adjust the template to accept it in that case. I've gone all the way back through history, and this has NEVER been set up correctly. The $idPrefix variable has been present in the helper and unused in the template since the bootstrap3/bootprint3 themes were introduced in 2014. Maybe the solution is something like this:

```
$context = compact('number') + [
    'id' => $this->getUniqueIdWithSourcePrefix(),
    'checkboxElementId' => $this->getUniqueHtmlElementId($idPrefix),
];
```

demiankatz (changes requested):
Thanks for getting this started, @RLangeUni -- I think this has turned up some issues that require a little more work. See comments below!

demiankatz:
@elsenhans and @RLangeUni, there hasn't been further activity here for more than a month -- just wanted to be sure you saw my earlier review. Please let me know if you need any more action/information from my end!

elsenhans:
Hi, @demiankatz , thanks for the reminder!
As far as I know, @RLangeUni  had waited for the merge of https://github.com/vufind-org/vufind/pull/2982

demiankatz:
Thanks, @elsenhans -- sorry that #2982 took so long to resolve, but now that it's finally done, hopefully we can make more progress here.

demiankatz:
I'm moving this to the 10.0 milestone, as we're getting close to the 9.1 code freeze, and I don't think this is going to get done in time. Of course, happy to keep working on it regardless whenever you're ready!

demiankatz:
@RLangeUni / @elsenhans, now that 10.0 development is beginning, I'm just checking in to see where we are on this PR. Please let me know if you need any action on my end to get this moving again.

elsenhans:
@RLangeUni , @demiankatz 
If someone can tell me what is still missing or needs to be adjusted here, I would continue working on the PR. However, I am a frontend developer and it seems that the problems regarding the ID have to be solved mainly with PHP. Is that correct?

demiankatz:
@elsenhans, see my review comments above for the specific issues I highlighted on my previous review. As you say, all of this is touching PHP code, but it is PHP code that is used to set values for building the front-end templates, so it's definitely front-end-adjacent. Perhaps a conversation with @RLangeUni would be in order... or, if you'd like me to make some of my proposed changes so that you can test them, that's also an option (though I probably won't get to it quickly due to my long backlog of reviews at the moment).

elsenhans:
@RLangeUni has informed me about the current status of the PR. The ID is already provided and the access to the ID in the templates should be implemented in a separate PR. 
The comments and the error are still being worked on in this PR. For the changes of templates we would create a new PR.

demiankatz:
Thanks, @elsenhans -- then I'll just wait for the work to be completed on my earlier review comments. If I can do anything to help in the meantime, please let me know!

demiankatz:
@elsenhans / @RLangeUni, with the 10.0 release beginning to draw near, do you expect to do more work on this soon, or should we reschedule for a future release? I also notice that there is a conflict in the test class -- please let me know if you need my help to resolve that.

jpkanter:
As @RLangeUni is on vacation this week i quickly tried to take over the open points but my lack of knowledge with PHPUnit made it impossible to replace the correct withConsecutive Call. Is next week early enough?

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/RecordTest.php
```diff
@@ -365,6 +365,21 @@ public function testGetLink(
...
     }
 
+    /**
+     * Helper function for testGetCheckbox to simulate withConsecutive deprecated in PhpUnit 9.6
+     * stolen from https://stackoverflow.com/a/75536028
+     *
+     * @param ...$args
+     * @return callable
+     */
+    public function consecutiveCalls(...$args): callable
```
demiankatz: You might find it helpful to use [WithConsecutiveTrait.php](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFindTest/Feature/WithConsecutiveTrait.php) here -- that's the general-purpose solution I used to replace the deprecated withConsecutive method in other tests.

demiankatz (changes requested):
@jpkanter, thanks for the progress here. I haven't fully reviewed your changes yet since I'm not sure if you're completely done yet, but I noticed that your test changes may be able to be simplified. See below...

demiankatz:
@jpkanter, this can definitely wait a week or two if necessary. But if you need help adapting the test to use WithConsecutiveTrait as I suggested, let me know and I don't mind helping. :-)

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/RecordTest.php
jpkanter: Thats an interesting headsup, i will look into it tomorrow. I am indeed not done, i quickly threw some code together as the necessary changes seemed rather simple and i had no running test-environment on my local machine. As i cannot continue to spam not working tests i should probably get that going again.

module/VuFind/src/VuFind/View/Helper/Root/Record.php
```diff
@@ -419,10 +419,10 @@ public function getSearchResult($view)
      */
     public function getCheckbox($idPrefix = '', $formAttr = false, $number = null)
     {
-        $id = $this->driver->getSourceIdentifier() . '|'
-            . $this->driver->getUniqueId();
-        $context
-            = ['id' => $id, 'number' => $number, 'prefix' => $idPrefix];
+        $context = compact('number') + [
```
demiankatz, line 563: It looks to me like this refactoring is dropping the `prefix` element from the array. While I realize in the long-term we may no longer need that value, I think in the short-term, we probably still need it (unless there are corresponding template changes that are meant to be in this PR but are somehow missing).

module/VuFind/src/VuFind/View/Helper/Root/Record.php
```diff
@@ -419,10 +419,10 @@ public function getSearchResult($view)
...
+            'id' => $this->getUniqueIdWithSourcePrefix(),
+            'checkboxElementId' => $this->getUniqueHtmlElementId($idPrefix)
+            ];
```
demiankatz: Minor style point: I'd de-indent the closing `];` here for consistency with code style elsewhere.

demiankatz (changes requested):
A couple of other small details...

module/VuFind/src/VuFind/View/Helper/Root/Record.php
demiankatz: Interestingly, it does appear that prefix is unused by the current [checkbox.phtml](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/record/checkbox.phtml), so maybe it is okay to remove it. But it might also be worth updating the template to actually include the ID value since we're going to the trouble of creating it. The only question is whether there's any risk of that causing duplicate element IDs under any circumstances.

robertlange81:
Sorry for the delay. I'll try to fix this today.

themes/bootstrap3/templates/record/checkbox.phtml
```diff
@@ -1,5 +1,5 @@
 <label class="record-checkbox hidden-print">
-  <input class="checkbox-select-item" type="checkbox" name="ids[]" value="<?=$this->escapeHtmlAttr($this->id) ?>"<?php if (isset($this->formAttr)): ?> form="<?=$this->formAttr ?>"<?php endif; ?> aria-label="<?=$this->transEscAttr('select_item')?>">
+  <input <?php if (isset($this->checkboxElementId)): ?>id="<?=$this->checkboxElementId ?>"<?php endif; ?> class="checkbox-select-item" type="checkbox" name="ids[]" value="<?=$this->escapeHtmlAttr($this->id) ?>"<?php if (isset($this->formAttr)): ?> form="<?=$this->formAttr ?>"<?php endif; ?> aria-label="<?=$this->transEscAttr('select_item')?>">
```
demiankatz: I think we need to add `$this->escapeHtmlAttr()` around `$this->checkboxElementId` to ensure safe escaping of the value.

demiankatz (changes requested):
Thanks for the progress on this, @RLangeUni! Looks like you might need to run `composer fix` to clean up some minor style things. Also, please let me know if you need my help with the test issues discussed in earlier comments.

themes/bootstrap3/templates/record/checkbox.phtml
robertlange81: I wasn't sure whether it was escaped before, but seems to be needed.

robertlange81:
Probably it is useful to add tests for getUniqueIdWithSourcePrefix and getUniqueHtmlElementId. And should we change templates in a separate PR only?

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/RecordTest.php
```diff
@@ -372,28 +372,42 @@ public function testGetLink(
...
-            $context
+        ];
+
+        $this->expectConsecutiveCalls(
+            $context,
+            'renderInContext',
+            $expectedCalls,
+            ['success', 'success']
         );
-        // We run the test twice to ensure that checkbox incrementing works properly:
```
demiankatz, line 413: I think this comment is still true; you might not want to delete it.

module/VuFind/src/VuFind/View/Helper/Root/Record.php
```diff
@@ -711,4 +712,34 @@ protected function deduplicateLinks($links)
...
+    /**
+     * Get the source identifier + unique id of the record
+     *
+     * @return string
+     */
+    public function getUniqueIdWithSourcePrefix()
+    {
+        if ($this->driver) {
+            return "{$this->driver->tryMethod('getSourceIdentifier')}"
+                . "|{$this->driver->tryMethod('getUniqueId')}";
```
demiankatz: Both `getSourceIdentifier` and `getUniqueId` are required by the `AbstractBase` record driver, so I don't think it's necessary to use `tryMethod` here -- they should be guaranteed to exist in any legal record driver.

demiankatz (changes requested):
Thanks, @RLangeUni -- I think this could be safely merged now, but a couple of suggestions that might polish up the changes.

Also, as you say, it never hurts to be comprehensive in test coverage, so if you want to add some new test methods to cover the new support methods, that would be great too!

Please let me know if I can help with anything, and also when you're ready for me to do another (hopefully final) review.

robertlange81:
bah, my sniffer works in finc but not on dev ...

robertlange81:
I 've added an additional test for getUniqueHtmlElementId, that should cover getUniqueIdWithSourcePrefix too. If it is not appropriate I could revert it.

demiankatz:
Thanks, @robertlange81 -- this all looks good to me now. I'm going to talk with @EreMaijala about some potential edge cases in combined search that could lead to duplicate IDs on the page before I merge this -- we might be able to address that issue using whatever mechanism we come up with to solve #3557. But in the meantime, I don't think any further work is required on your part. I'll keep you posted on what we come up with! :-)

demiankatz (changes requested):
@RLangeUni, I spoke with @EreMaijala, and he agrees that there is a remaining problem that we need to solve: it is possible to have multiple columns of results coming from the same search backend in combined search, but with different applied filters. In some configurations, it would be possible for the exact same record to appear in multiple columns on a single page. As written here, this would cause duplicate HTML IDs, which is illegal.

Would it make sense to add something like a uuid suffix onto the generated IDs to ensure uniqueness? Obviously, the problem with that is that multiple calls to the getUniqueHtmlElementId() method would return different values. Depending on the expected/intended usage, this could be good or bad. Basically it all comes down to whether we expect to call getUniqueHtmlElementId once, assign the value to variable, and use that value in multiple places... or if we expect to be able to call it multiple times and always get the same value back for the same record.

Do you have any thoughts/preferences on this? Since this is foundational work, I just want to be sure we build it in a way that's sufficiently robust and flexible to meet all future needs.

robertlange81:
This sounds like a tough nut to crack. At finc, we have so far only used one result list. If a page generates two lists, both with the same backend (Solr), then the IDs would be identical, unless you specify a context/prefix. In our example with the checkbox in list-entry, you would need to call the checkbox with getCheckbox('search-results-1'). This would be error-prone and involve a lot of logic spread over several templates.

Another option is to try using UUIDs, hash codes, or timestamps to generate a unique ID for each element, but then the reference from other templates will become the problem. Here, it's a case of the chicken and the egg.

I think it depends on what we want to use the unique ID for. If it's 'just' about referring to it with aria-* attributes or adding it to a watchlist, maybe just one record with a unique ID would suffice. That is, generate the checkboxElementId only for the first element => createUniqueHtmlElementId, include it in a static list of IDs, and retrieve it with getUniqueHtmlElementId. That means the same record in list2 would have no id at all. Then we would need to test what happens with JavaScript etc. when a record is saved as a favorite; whether then both records in both lists are marked. But this problem should have already been solved without the unique ID. An exception should be records in modals - here we should use a prefix 'modal' or similar, but I have in mind, this approach was not very popular?

demiankatz:
@robertlange81, my first thought was this: what if we generate a unique record set ID (e.g. a uuid) in AbstractBackend::injectSourceIdentifier, and then push that into a new method in RecordCollectionInterface/AbstractRecordCollection (e.g. setResultSetIdentifier) that in turn pushes the value into a matching method in RecordInterface/RecordTrait? This way, every record driver will have a stable unique ID, and if the same record is retrieved in multiple contexts, they won't conflict, even if the code that retrieves them runs in isolation (e.g. via an AJAX request). It just requires that we use the same record driver object every time we want to reference an ID related to that specific object.

But on thinking deeper, that seems like a lot of work with little benefit. It would allow us to create more human-readable identifiers, but they would be very long, and does it matter if humans can read them? Maybe, as you say, all we really need is a getRecordUuid method on the record driver that generates and persists an ID unique to that particular instance of the record. As above, as long as we use the same record driver object every time we reference the instance we are working with, we can store a UUID in the object and be confident that things will match up, but separate instances of the same record will have different IDs. This would allow us to match up element IDs across templates without problems (as long as the record driver object can be shared). As you say, we would need to prefix the IDs for various contexts, but I don't think that should be a big problem.

Do you think this is an approach worth pursuing?

robertlange81:
To stay on top of the matter, I will try to implement a solution this week. Although I'm not entirely sure yet how I will reproduce the scenario with the two result lists.

demiankatz:
Thanks, @robertlange81. The easiest way to reproduce the scenario with multiple result lists is to set up combined.ini with multiple identical columns. You can do this by using a colon and an identifier after the section name -- e.g.

```
[Solr]
label = Catalog
;sublabel = "library books, videos, CDs, microform, local resources"
more_link = "More catalog results"
;advanced_link_bottom = "Advanced Search"
;limit = 10

[Solr:duplicate]
label = "Catalog (duplicate copy)"
;sublabel = "library books, videos, CDs, microform, local resources"
more_link = "More catalog results"
;advanced_link_bottom = "Advanced Search"
;limit = 10
```

If you go to /Combined under your VuFind URL and do a search, this should show you the same results twice in different columns.

Please let me know if you need more detail/assistance to set this scenario up.

demiankatz:
@robertlange81, just checking in to see if you've had a chance to make any progress here. We should probably freeze code for the 10.0 release in the next week (or two, at the extreme). Please let me know if I can do anything to help!

demiankatz:
Looks like this isn't going to get done in time for 10.0; since it's a relatively small change, I have moved the milestone to 10.1.

demiankatz:
Looks like this isn't going to get done in time for 10.0; since it's a relatively small change, I have moved the milestone to 10.1.

robertlange81:
> Looks like this isn't going to get done in time for 10.0; since it's a relatively small change, I have moved the milestone to 10.1.

"Unfortunately, I didn't make it in time. Hopefully for 10.1."

robertlange81:
I implemented a unique result set identifier for records, ensuring that each record receives a UUID-like identifier to avoid conflicts. PHPStn forced me to add the $recordCollection property in AbstractBackend. generateUuid() could be more handsome and moved to some Utils class? But I didn't found an existing appropriate. The template changes should not be merged. They are more for my own comparism. For example in result-list should be later changed like `<a id="<?=$this->escapeHtmlAttr($this->record($this->driver)->getUniqueHtmlElementId())?>" ...`  - so it seems that there are no duplicate ids in the combined search.

module/VuFindSearch/src/VuFindSearch/Backend/AbstractBackend.php
```diff
@@ -116,6 +126,39 @@ abstract public function getRecordCollectionFactory();
...
+    /**
+     * Sets the result set identifier for the record collection.
+     *
+     * @param string $uuid A valid UUID associated with the data set.
+     *
+     * @return void
+     */
+    public function setResultSetIdentifier(string $uuid)
+    {
+        $this->recordCollection->setResultSetIdentifier($uuid);
```
demiankatz: Is this method actually used? I think the property you were forced to add was related to the call in this method, but I suspect this method is unnecessary.

demiankatz (changes requested):
Thanks, @RLangeUni! I think we're nearly done here, but I think the phpstan issue you reported may have been related to some orphaned and unnecessary code; see below for details.

robertlange81:
In the AbstractBackend class, I already had a slight suspicion that I was too eager. It seems "setResultSetIdentifier" is sufficient for the records/collections. Is the generation of the UUID alright? With Doctrine, one could also use its helper methods (Doctrine\ORM\Id\UuidGenerator).

robertlange81:
"injectSourceIdentifier" is also called from InjectSpellingListener.php - should we check within AbstractBackend::injectSourceIdentifier() whether `$response->getRecords()` is empty or zero before setting the result set identifier unnecessarily?

module/VuFindSearch/src/VuFindSearch/Response/RecordTrait.php
```diff
@@ -47,6 +47,16 @@ trait RecordTrait
...
 
+    /**
+     * The unique identifier for the result set.
+     *
+     * This property stores a UUID or similar identifier that uniquely identifies
+     * the result set. It is typically set by calling the `setResultSetIdentifier` method.
+     *
+     * @var string|null
+     */
+    protected $resultSetIdentifier;
```
demiankatz: I'd suggest explicitly initializing this to null.

module/VuFind/src/VuFind/View/Helper/Root/Record.php
```diff
@@ -857,4 +858,36 @@ protected function deduplicateLinks($links)
...
+     *
+     * @return string
+     */
+    public function getUniqueHtmlElementId($idPrefix = '')
+    {
+        return preg_replace(
+            "/\s+/",
+            '_',
+            ($idPrefix ? $idPrefix . '-' : '')
+            . $this->driver->getResultSetIdentifier() . '-'
```
demiankatz: We should add handling here to ensure this value is non-null, since concatenating null into a string may cause a notice. It would probably be good to add a case to the unit test to cover the "null record set identifier" case so we can ensure it behaves as expected.

module/VuFindSearch/src/VuFindSearch/Backend/AbstractBackend.php
```diff
@@ -116,6 +119,27 @@ abstract public function getRecordCollectionFactory();
     protected function injectSourceIdentifier(RecordCollectionInterface $response)
     {
         $response->setSourceIdentifiers($this->identifier);
+        $response->setResultSetIdentifier($this->generateUuid());
```
demiankatz: I think this approach is fine for now, but maybe it's worth putting a TODO comment here to switch to using the Doctrine UUID generator once it becomes available to us in the future. Then we could potentially address that as part of (or sometime after the merge of) #2233.

demiankatz:
> "injectSourceIdentifier" is also called from InjectSpellingListener.php - should we check within AbstractBackend::injectSourceIdentifier() whether `$response->getRecords()` is empty or zero before setting the result set identifier unnecessarily?

I suppose it wouldn't hurt to count the records before generating the UUID to potentially save a few cycles, though I don't think it's a huge issue.

demiankatz (changes requested):
See below for a few more minor suggestions!

demiankatz:
I'm also requesting a review from @maccabeelevine since #3552 is going to depend on the work here, so his opinion/input may be valuable!

robertlange81:
Adopted the last suggestions. Thanks for your support!

demiankatz (approved):
Thanks, @RLangeUni, this all looks good to me now. I appreciate your persistence in seeing this through to the end. :-)

I'm not going to merge it quite yet, just in case anyone else has last-minute feedback. I'll resolve it next week if there is no conversation before then.

maccabeelevine (approved):
Works very nicely!

demiankatz:
Thanks, @RLangeUni and @maccabeelevine. @EreMaijala pointed out that it might be better to generate real UUIDs with the ramsey/uuid library to prevent confusion. I'm inclined to agree but don't want to further delay this PR, so I'll merge this as-is and open a separate UUID PR shortly.

PR #3000, opened 2023-07-21
FOLIO: Fix results pagination when limit value is exceeded

meganschanz:
We stumbled on this bug with our course reserves import, that when more than 1,000 records (the `limit`) are returned from the FOLIO API, you will not get all of the records returned. According to the [FOLIO documentation](https://github.com/folio-org/raml-module-builder#estimated-totalrecords), the `totalRecords` field is only an estimate of the number of results to expect when the there are more than 1,000 records in the results. This isn't likely to affect many of the API calls since the vast majority of them don't return more than 1,000 results. But we sure noticed it when we were missing ~2,300 course reserves due to the "estimate"!

This pull request updates the logic to no longer rely on that field and instead keep looping until there are no more pages of results (the last page will have less than the `limit` value of records). In the event that the last page happens to have exactly the number of results as the limit value, the subsequent call will simply return 0 results and the loop will exit (because the count of 0 records is no longer equal to the limit). I'm not thrilled that particular edge case results in an extra call to FOLIO, but without being able to rely on other data points in the previous response to know that we have all of the results I'm not seeing how we can avoid it -- but I'm open to suggestions if others can think of any. You could theoretically make a call with `limit` set to `0` to get an accurate `totalRecord` count, per the documentation, but that seems like a lot of overhead to add an extra call just to handle an edge case.

I also added `limit` as an optional parameter (maintaining its existing default value) to the function purely to make testing of the function easier; without having to actually make fixtures with 1,000 records in them.


module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -966,19 +966,15 @@ protected function getPagedResults($responseKey, $interface, $query = [])
...
                 yield $item ?? '';
             }
-            // Continue until the count reaches the total records
-            // found, if count does not increase, something has gone
-            // wrong. Stop so we don't loop forever.
-        } while ($count < $total && $previousCount != $count);
+            $offset += $limit;
+
+            // Continue until the results are not the limit value (i.e. the last page of results)
+        } while ($count == $limit);
```
demiankatz: I wonder if as an extra safety measure, it might still be worth keeping a running total and looking at $json->totalRecords -- we could potentially have a while condition like `while ($count == $limit || $totalCount < $json->totalRecords)` so that we're less likely to abort early if something unexpected occurs (i.e. what if some glitch in FOLIO causes it to return one less record on the first page and then we fail to iterate deeper into the set?).

Obviously I have no specific reason to believe that such a problem would occur, but it seems conceivable, and maybe it's safer to try to have this both ways at once. What do you think?

demiankatz (changes requested):
Thanks, @meganschanz! Just one thought (which you can feel free to disagree with if you think I'm being overly paranoid... :-) ).

module/VuFind/src/VuFind/ILS/Driver/Folio.php
meganschanz: That is a good idea. We've seen that scenario occur with FOLIO's OAI harvest calls, so it's not out of the realm of possibility that other API calls could have that happen at some point. I made a commit with that suggestion in it.

maccabeelevine:
The revised algorithm seems reasonable to me.  FWIW, the "total records is actually an estimate" is not an uncommon pattern, especially since the total result set can change between API calls; OAI-PMH [does the same](https://www.openarchives.org/OAI/openarchivesprotocol.html#FlowControl).  That said, I have seen some particularly wrong estimates coming from FOLIO on some API calls.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
demiankatz: Thanks for the revisions, @meganschanz. I should have done this sooner but I was a bit pressed for time yesterday -- now that I've looked more closely at the documentation you shared, I see that it makes a specific algorithmic recommendation itself:

"Note that clients should continue on the next page when totalRecords = offset + limit because there may be more records."

If we take this recommendation, we could use `while ($offset <= $totalEstimate)` and eliminate some of the manual counting from the code, which might be more efficient. What do you think? (EDIT: Note that I'm using `$offset` instead of `$offset + $limit` since we add the limit right before the bottom of the loop).

meganschanz: That sounds good. I just double checked in our FOLIO instance how the `totalRecords` field would behave when thousands of records were returned to make sure that the value wasn't always the low ~1000 value like it was on the `offset=0` pass, because I couldn't recall, and it does increase like we would need it to so I'll go ahead and work on a commit for this change.

| limit | offset | totalRecords |
|-------|--------|---------------|
| 1000  | 0      | 1000          |
| 1000  | 1000   | 2000          |
| 1000  | 2000   | 3000          |
| 1000  | 3000   | 3042          |

demiankatz (approved):
Thanks again, @meganschanz! I've tested this with my own course reserves and I'm getting identical results with this branch vs. dev, and all tests are passing. This is not only more accurate now, but the code also seems a little bit simpler. I'll merge it now!

PR #3004, opened 2023-07-26
Batch import: sort files and follow symlinks

damien-git:
VuFind 9 started using find for batch import (see [commit](https://github.com/vufind-org/vufind/commit/86e4671b5ef10405faca92c9dee4ee08d002a288)). 2 unintended consequences were that links were not followed anymore, and files were no longer sorted by name (and hence by harvest date). A possible consequence of this is that an older version of a record might get imported after a new one, replacing it. This can happen for instance if files are moved from processed/ to reimport all records.
This PR solves these 2 issues by using a find option to follow symlinks and a sort afterwards to sort files by name (which include a timestamp).

### DIFF
--- a/harvest/batch-import-marc-auth.sh
+++ b/harvest/batch-import-marc-auth.sh
@@ -113,7 +113,7 @@ else
 fi
 
 # Process all the files in the target directory:
-find $BASEPATH -maxdepth 1 \( -iname "*.xml" -o -iname "*.mrc" -o -iname "*.marc" \) -type f -print0 | \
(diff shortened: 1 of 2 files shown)


demiankatz (approved):
Thanks for catching these problems, @damien-git!

damien-git:
@demiankatz Thanks. Are you thinking of including this in 9.0.4 or 9.1 ?

demiankatz:
@damien-git, probably 9.1, since I'm expecting to have that out in October, and I don't know that we need another minor release between now and then unless we get a compelling collection of fixes. Of course, if a 9.0.4 does turn out to be needed, this would be a good fix to cherry-pick into it (along with the recently-merged #3003).

PR #3005, opened 2023-07-26
LibGuidesProfile: Add call number matching strategy

maccabeelevine:
Expand the LibGuidesProfile recommendation module to use different strategies.

Beyond the original 'subject' matching strategy, this adds a 'call_number' strategy
based on Villanova's custom code.
https://gist.github.com/demiankatz/4600bdfb9af9882ad491f74c406a8a8a

It looks at the call number facet to the search results, finds the most prominent
call number, and maps that to a LibGuides profile.  This mapping happens via
configuration in LibGuidesAPI.ini.  Future enhancements might add other sources 
for that mapping.

### DIFF
--- a/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
+++ b/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
@@ -110,7 +110,7 @@ class LibGuidesProfile implements
 
         $profile = $config->Profile;
         if ($profile) {
-            $this->strategies = $profile->get('strategies', []);
+            $this->strategies = (array) $profile->get('strategies', []);
             $this->callNumberToAlias = $profile->call_numbers ? $profile->call_numbers->toArray() : [];
             $this->aliasToAccountId = $profile->profile_aliases ? $profile->profile_aliases->toArray() : [];
             $this->callNumberLength = $profile->get('call_number_length', 3);
--- a/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
+++ b/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
@@ -171,19 +171,12 @@ class LibGuidesProfile implements
     {
         // Consider strategies in the order listed in the config file.
         foreach ($this->strategies as $strategy) {
-            if ("call_number" == $strategy) {
-                // Try the best match for call numbers in the results facets.
-                $account = $this->findBestMatchByCallNumbers($this->results);
-                if ($account) {
-                    return $account;
-                }
-            } elseif ("subject" == $strategy) {
-                // Find best match for the query terms alone.
-                $query = $this->results->getParams()->getQuery();
-                $account = $this->findBestMatchByQuery($query);
-                if ($account) {
-                    return $account;
-                }
+            $method = 'findBestMatchBy' . $strategy;
+            if (
+                method_exists($this, $method) &&
+                $account = $this->$method($this->results)
+            ) {
+                return $account;
             }
         }
         return false;
@@ -199,7 +192,7 @@ class LibGuidesProfile implements
      *
      * @return array LibGuides account
      */
-    protected function findBestMatchByCallNumbers($results)
+    protected function findBestMatchByCallNumber($results)
     {
         // Skip if no call number mapping was provided by config.
         if (empty($this->callNumberToAlias)) {
@@ -250,19 +243,20 @@ class LibGuidesProfile implements
     }
 
     /**
-     * Find the LibGuides account whose profile best matches the
-     * given query.
+     * Find the LibGuides account whose subject expertise in their
+     * profile best matches the given query.
      *
-     * @param \VuFindSearch\Query\QueryInterface $query Current search query
+     * @param \VuFind\Search\Base\Results $results Search results object
      *
      * @return array LibGuides account
      */
-    protected function findBestMatchByQuery(\VuFindSearch\Query\QueryInterface $query)
+    protected function findBestMatchBySubject($results)
     {
         $data = $this->getLibGuidesData();
         $subjectToId = $data['subjectToId'];
         $idToAccount = $data['idToAccount'];
 
+        $query = $results->getParams()->getQuery();
         $queryString = $query->getAllTerms();
         if (!$queryString) {
             return false;
--- a/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
+++ b/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
@@ -179,6 +179,10 @@ class LibGuidesProfile implements
      */
     public function getResults()
     {
+        if (empty($this->strategies)) {
+            throw new \Exception("LibGuidesAPI.ini must define at least one strategy if LibGuidesProfile is used.");
+        }
+
         // Consider strategies in the order listed in the config file.
         foreach ($this->strategies as $strategy) {
             $method = 'findBestMatchBy' . $strategy;
--- a/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
+++ b/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
@@ -108,8 +108,7 @@ class LibGuidesProfile implements
         // Cache the data related to profiles for up to 10 minutes:
         $this->cacheLifetime = intval($config->GetAccounts->cache_lifetime ?? 600);
 
-        $profile = $config->Profile;
-        if ($profile) {
+        if ($profile = $config->Profile) {
             $this->strategies = $profile->get('strategies', []);
             $this->callNumberToAlias = $profile->call_numbers ? $profile->call_numbers->toArray() : [];
             $this->aliasToAccountId = $profile->profile_aliases ? $profile->profile_aliases->toArray() : [];
--- a/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
+++ b/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
@@ -110,7 +110,7 @@ class LibGuidesProfile implements
 
         $profile = $config->Profile;
         if ($profile) {
-            $this->strategies = (array) $profile->get('strategies', []);
+            $this->strategies = $profile->get('strategies', []);
             $this->callNumberToAlias = $profile->call_numbers ? $profile->call_numbers->toArray() : [];
             $this->aliasToAccountId = $profile->profile_aliases ? $profile->profile_aliases->toArray() : [];
             $this->callNumberLength = $profile->get('call_number_length', 3);
--- a/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
+++ b/module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
@@ -116,7 +116,9 @@ class LibGuidesProfile implements
         $this->cacheLifetime = intval($config->GetAccounts->cache_lifetime ?? 600);
 
         if ($profile = $config->Profile) {
-            $this->strategies = $profile->get('strategies', []);
+            $strategies = $profile->get('strategies', []);
+            $this->strategies = is_string($strategies) ? [$strategies] : $strategies;
+
             $this->callNumberToAlias = $profile->call_numbers ? $profile->call_numbers->toArray() : [];
             $this->aliasToAccountId = $profile->profile_aliases ? $profile->profile_aliases->toArray() : [];
             $this->callNumberField = $profile->get('call_number_field', 'callnumber-first');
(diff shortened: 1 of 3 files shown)


maccabeelevine:
@demiankatz Tests not written yet, but an initial look would help.  Thanks!

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -79,6 +107,14 @@ public function __construct(
 
         // Cache the data related to profiles for up to 10 minutes:
         $this->cacheLifetime = intval($config->GetAccounts->cache_lifetime ?? 600);
+
+        $profile = $config->Profile;
+        if ($profile) {
+            $this->strategies = $profile->get('strategies', []);
```
demiankatz: You might want to add `(array)` in front of this to typecast it in case somebody configures a single string instead of an array of strings.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -62,6 +62,34 @@ class LibGuidesProfile implements
      */
     protected $libGuides;
 
+    /**
+     * List of strategies enabled to find a matching LibGuides profile
+     *
+     * @var int
+     */
+    protected $strategies = [];
```
demiankatz, line 70: Should this default to `['subject']` so that it's guaranteed to do something even if no additional configuration is provided?

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -79,6 +107,14 @@ public function __construct(
 
         // Cache the data related to profiles for up to 10 minutes:
         $this->cacheLifetime = intval($config->GetAccounts->cache_lifetime ?? 600);
+
+        $profile = $config->Profile;
+        if ($profile) {
```
demiankatz: If you want to shave off a line, you could make this `if ($profile = $config->Profile) {`.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -134,8 +170,83 @@ public function process($results)
...
+        foreach ($this->strategies as $strategy) {
+            if ("call_number" == $strategy) {
+                // Try the best match for call numbers in the results facets.
+                $account = $this->findBestMatchByCallNumbers($this->results);
+                if ($account) {
+                    return $account;
+                }
+            } elseif ("subject" == $strategy) {
+                // Find best match for the query terms alone.
+                $query = $this->results->getParams()->getQuery();
```
demiankatz: Rather than extracting the query here, maybe it's better to pass $this->results to findBestMatchByQuery, parallel to how findBestMatchByCallNumber works. That would allow you to simplify this code significantly -- instead of having a separate set of logic for each strategy, you could just use logic to identify the strategy method name, and then call that method dynamically -- e.g.

```
$strategyMap = ['call_number' => 'findBestMatchByCallNumbers', 'subject' => 'findBestMatchByQuery'];
foreach ($this->strategies as $strategy) {
    $method = $strategyMap[$strategy];
    if ($account = $this->$method($this->results)) {
        return $account;
    }
}
```

You might even be able to use some kind of convention to eliminate the need for `$strategyMap`, though then you'd want to add safety mechanisms to prevent calls to undefined methods or triggering of inappropriate methods via configuration.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -134,8 +170,83 @@ public function process($results)
...
+    protected function findBestMatchByCallNumbers($results)
+    {
+        // Skip if no call number mapping was provided by config.
+        if (empty($this->callNumberToAlias)) {
+            return false;
+        }
+
+        // Get the Call Number facet list.
+        $filter = [
+            'callnumber-first' => 'Call Number',
```
demiankatz: You should probably make the call number field name configurable in the .ini file so you could specify dewey_hundreds instead of callnumber-first when desired. Or maybe we need a way to try both maps in case mixed call numbers are available and one set fails (though that's possibly overkill).

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
```diff
@@ -62,6 +62,34 @@ class LibGuidesProfile implements
...
+     * @var int
+     */
+    protected $strategies = [];
+
+    /**
+     * Map of call number pattern to config alias
+     *
+     * @var array
+     */
+    protected $callNumberToAlias;
```
demiankatz, line 77: Rather than mapping call numbers to aliases, would it be better to map call numbers to accounts, with aliases as an optional feature? I.e. for each value, look it up in the alias table to map it to an account number, but if it's not in the alias table, assume it is _already_ an account number. This would save people from having to create aliases for accounts that are only used once or twice. (I realize that the aliases are probably going to be more useful/readable in the majority of cases, but it seems unnecessary to make them mandatory).

demiankatz (changes requested):
Thanks, @maccabeelevine -- here are a few thoughts and suggestions!

I also wonder if there's some way to move some of this logic into a trait to make it reusable across modules, if we want to add similar functionality for multiple platforms (or so that my local custom code can utilize more upstream logic). Not a top priority, but maybe something to contemplate/consider.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Yeah I considered that.  But the real benefit to aliases (which you couldn't know since you don't have LibGuides) is that the account IDs are purely numeric, and 4-6 digits(!), making them completely inscrutable.  So even if every single call number prefix maps to a different account, I don't realistically see anyone wanting their configuration file to be unreadable enough to skip the aliases.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Good idea, I made the field configurable.  I did not want to try both maps since there are actually more than two that may be needed (Lehigh will likely use dewey-ones or -tens) and why add complexity.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: That's fantastic.  I did eliminate $strategyMap and I added a method_exists($this) check, is that safe enough?

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Nice.  Done.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: I agree that would be a "quicker to get it working" solution but I'd rather force the deployer to choose.  Because 'subject' may work out-of-the-box but it's pretty poor, and I'd hate for someone to miss the fact in config that it can work better.  Also 'subject' always returns a result, so it should likely be disabled if the deployer prefers the other strategy, and having it on by default might make that less obvious.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Hmm, the problem is $config->get doesn't return an array normally, it's some object.  The default value [] works if there is no config present, but I'd prefer it as the object otherwise.

maccabeelevine:
> I also wonder if there's some way to move some of this logic into a trait to make it reusable across modules, if we want to add similar functionality for multiple platforms (or so that my local custom code can utilize more upstream logic). Not a top priority, but maybe something to contemplate/consider.

I agree with that in theory but it's hard for me to see which parts to refactor without that second use case.  Comparing with your local custom code, there is the most overlap in this loop

`foreach ($facets[$this->callNumberField]['list'] ?? [] as $current) {`

but even that is semantically different because I'm matching to user profiles and you are matching to subjects.  I am open to refactoring ideas?

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
demiankatz: Good point. The behavior of the Config object is annoying just about as often as it's helpful. :-) I think this still leaves us with the problem that `$this->strategies` may end up being set to a string if the configuration isn't quite right. Maybe something like:

```
$strategies = $profile->get('strategies', []);
$this->strategies = is_string($strategies) ? [$strategies] : $strategies;
```

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
demiankatz: I guess another approach might be to add a 'missing' strategy that throws an exception complaining that no strategies are configured, and use that as the default...

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
demiankatz: Probably true. I guess I could argue for the possibility of using the numbers with explanatory comments, but that's probably pushing it. No objections to leaving it as-is. :-)

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
demiankatz: I think it's pretty safe, though If you're feeling paranoid, you might also use a regex to strip weird characters off the strategy string to avoid shenanigans (not that I can think of any specific bad thing somebody could do) -- i.e. a `preg_replace('/[^\w]*/, '', $strategy)` type of thing.

demiankatz (changes requested):
Thanks, @maccabeelevine! A few more comments... I also agree that if there's not a lot of overlap between my example code and what you've done here, refactoring to a trait is probably premature; we can revisit that when we have a more tangible second use case.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: I'm not sure I'm wrapping my head around this, but if a method actually exists matching whatever the strategy string is, and it's somehow malicious, I'm not sure how restricting the method names to alphanumeric will necessarily help.  Couldn't the attacker just name their method something in the restricted character set?  I do get that we might want to exclude loading methods from other classes/paths somehow, but that seems to bring its own set of limitations.  I defer to your expertise here, you had the idea of dynamically determining the strategy method so I'm guessing similar approaches have been used elsewhere?

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: I added an empty($this->strategies) check in getResults(), is that ok?  It seemed cleaner than defining a fake exception-throwing strategy.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Yup that works.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Yeah if someone actually complains about this I am happy to make aliases skippable, but I don't see that happening.  I also think they may be useful for other logic in LibGuidesAPI.ini in the future.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
demiankatz: I'm just thinking about trying to prevent oddball scenarios, like a hypothetical situation where arbitrary code execution could be achieved by injecting null characters or something like that. I'm not worried about somebody running an inappropriate VuFind function so much as taking advantage of some flaw in the language itself. I don't think that's very likely either, but I try to think about input sanitization as policy just in case of crazy things happening.

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
demiankatz: Yeah, I agree that's more straightforward in this instance!

module/VuFind/src/VuFind/Recommend/LibGuidesProfile.php
maccabeelevine: Yeah fair point.  I added that sanitizing.

demiankatz (approved):
Looks good to me, @maccabeelevine. Thanks! I'm approving this but not merging it yet, just in case there's anything else you want to do on your end... but if you're done, feel free to merge at your convenience.

PR #3011, opened 2023-07-27
Upgrade Swagger UI and add associated Mink test.

demiankatz:
This PR upgrades to the next major version of Swagger UI and adds a Mink test to exercise the API code and Swagger UI. The test is very minimal at this point and could be expanded, but it's at least a foundation to build upon, and it's a quick way to ensure that Swagger UI is not completely broken after an upgrade.

Note that the test runs very slowly right now because the test suite relies on jQuery to check for availability of elements, and without jQuery, we have to wait for timeouts to be exceeded. I have opened #3012 to address that, which should correct the speed problems seen here.


xmorave2 (approved):
As main new feature is support for OpenAPI 3.1, we should investigate if we can take advantage of this new OpenAPI specification.

xmorave2:
https://spec.openapis.org/oas/v3.1.0
https://lornajane.net/posts/2020/whats-new-in-openapi-3-1
https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

demiankatz:
Thanks, @xmorave2! Looks like there may be some things worth taking advantage of -- we should probably get input from @EreMaijala on this when he returns in the next week or two.

demiankatz:
I've opened https://openlibraryfoundation.atlassian.net/browse/VUFIND-1629 to keep track of this.

PR #3012, opened 2023-07-27
Reduce test suite's reliance on jQuery.

demiankatz:
Our Mink test suite relies on jQuery for a variety of checks, which is a problem if we want to test a page that does not include jQuery. This PR uses browser-native checks in more places to reduce (but not yet fully eliminate) our reliance on jQuery. Merging this will greatly improve the performance of the tests introduced in #3011.

NOTE: I initially tried to simplify waitForPageLoad as well, but it relies on special jQuery selectors that are not supported by the native browser functionality and needs to remain as-is for now.

TODO
- [x] Run full test suite.


demiankatz:
All tests are passing with this revision.

xmorave2 (approved):

PR #3015, opened 2023-07-31
Adding the option to show the keys of translations

ThoWagen:
Sometimes it is helpful to show the keys of translations on the website directly in development systems. This can help non-technical staff members to customize the translations.

So in this PR a option for this is added. The translatable text has the format "\*\<key\>\*" to indicate that it is translatable.


demiankatz:
Thanks, @ThoWagen! This is a good idea, though I may not have time to review it in depth before I get back from vacation in a couple of weeks (though I will try if time permits). I'd welcome input from others in the meantime!

module/VuFind/src/VuFind/I18n/Translator/Loader/ExtendedIni.php
```diff
@@ -117,6 +117,11 @@ public function addToPathStack($pathStack)
      */
     public function load($locale, $filename)
     {
+
```
demiankatz: Unnecessary whitespace at top of function (I'm not sure why php-cs-fixer isn't complaining about this).

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
```diff
@@ -101,6 +101,10 @@ public function getTranslatorLocale($default = 'en')
      */
     public function translate($target, $tokens = [], $default = null)
     {
+        if ($this->translator !== null && $this->translator->getLocale() == "keys") {
```
demiankatz: You could use `$this->getTranslatorLocale()` here.

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
```diff
@@ -101,6 +101,10 @@ public function getTranslatorLocale($default = 'en')
...
+            return "*$target*";
```
demiankatz: Would it make sense to embed the $tokens in the return value if the array is non-empty? Otherwise, key parts of many strings will not be displayed at all.

module/VuFind/tests/integration-tests/src/VuFindTest/View/Helper/Root/ResultFeedTest.php
```diff
@@ -102,7 +102,8 @@ protected function getMockTranslator()
                 . '</strong> results of <strong>%%total%%</strong>',
         ];
         $mock = $this->getMockBuilder(\Laminas\I18n\Translator\TranslatorInterface::class)
-            ->getMock();
+            ->addMethods(['getLocale'])
```
demiankatz: Rather than implementing this change in multiple places, it looks like there's an opportunity to simply leverage the \VuFindTest\Feature\TranslatorTrait in this test and reduce some existing redundancy.

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/VersionsTest.php
```diff
@@ -62,10 +62,13 @@ public function testGetDescription(): void
         $obj = new Versions($config, $som);
         $obj->setRecordDriver($recordDriver);
         $translator = $this->getMockBuilder(\Laminas\I18n\Translator\TranslatorInterface::class)
-            ->getMock();
+            ->addMethods(['getLocale'])
```
demiankatz: Maybe another place to use the TranslatorTrait? Or is the expectation set below too specific for that to work? (I'm not sure if it really matters that we're expecting the 'default' namespace here... but that's the only major difference I see from using the trait).

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
```diff
@@ -76,7 +76,10 @@ public function testCursorMark(): void
      */
     public function testFacetTranslations(): void
     {
-        $mockTranslator = $this->createMock(TranslatorInterface::class);
+        $mockTranslator
+            = $this->getMockBuilder(TranslatorInterface::class)
+            ->addMethods(['getLocale'])
+            ->getMockForAbstractClass();
```
demiankatz: ...and another opportunity to use the trait. This test is specifying `exactly(2)` instead of `any` calls to the translate method, but we could address that by adding another parameter to the `getMockTranslator` method -- i.e.

```
protected function getMockTranslator($translations, $expectation = null)
{
...
    $translator->expects($expectation ?? $this->any())->method('translate')
```

demiankatz (changes requested):
Thanks again, @ThoWagen -- I did find a little time to look more closely at this, and I've left a few thoughts (many of them pretty repetitive. :-) ).

module/VuFind/src/VuFind/I18n/Translator/Loader/ExtendedIni.php
ThoWagen: Fixed

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
ThoWagen: Done

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
ThoWagen: Good call. Now, the tokens are also displayed.

module/VuFind/tests/integration-tests/src/VuFindTest/View/Helper/Root/ResultFeedTest.php
ThoWagen: Done

module/VuFind/tests/unit-tests/src/VuFindTest/RecordTab/VersionsTest.php
ThoWagen: Using the trait here works. I changed it.

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
ThoWagen: Using the trait here as well. I didn't see why using "exactly(2)" instead of "any" is important for this test. So, I ignored this for now. But, if you think this is required I will change it.

demiankatz: It's a slightly more specific assertion, but I agree that it's probbbly not a hugely important distinction here.

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
```diff
@@ -101,6 +101,12 @@ public function getTranslatorLocale($default = 'en')
      */
     public function translate($target, $tokens = [], $default = null)
     {
+        if ($this->getTranslatorLocale() == "keys") {
+            return "*$target"
+                . (!empty($tokens) ? " | ['" . implode("', '", array_keys($tokens)) . "']" : "")
```
demiankatz: This looks like it's only going to show the array keys, but seeing the values is useful too. Maybe it's worth something a little more complicated, like:

```
$callback = function ($key, $val) {
    return "$key = $val";
};
$tokenDetails = empty($tokens)
    ? ''
    : '| [' . implode(', ', array_map($callback, array_keys($tokens), array_values($tokens))) . ']';
return "*$target$tokenDetails*";
```

demiankatz (changes requested):
Thanks again, @ThoWagen -- one last thought before I do hands-on testing. :-)

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
ThoWagen: It might blow up some translations but I think it is more useful than harmful. It's implemented.

demiankatz: Thanks! I'll try this out in more detail when I return from vacation next week, but it's looking good to me.

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
```diff
@@ -101,6 +101,16 @@ public function getTranslatorLocale($default = 'en')
      */
     public function translate($target, $tokens = [], $default = null)
     {
+        if ($this->getTranslatorLocale() == "keys") {
+            $keyValueToString = function ($key, $val) {
+                return "$key = $val";
+            };
+            $tokenDetails = empty($tokens)
+                ? ''
+                : '| [' . implode(', ', array_map($keyValueToString, array_keys($tokens), array_values($tokens))) . ']';
```
demiankatz: I think this might look better if there were a space before the pipe -- i.e. `' | ['` instead of `'| ['`.

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
```diff
@@ -101,6 +101,16 @@ public function getTranslatorLocale($default = 'en')
...
+            return "*$target$tokenDetails*";
```
demiankatz: At this point in the code, $target might be an array (in situations where a text domain is represented as an array instead of using a `::` delimiter -- you can see this, for example, on the facet list on the search home screen in a default configuration). This causes the string to be displayed as `*Array*`. I think you might want to move all this logic below the extractTextDomain call so that you can guarantee that you have a string and not some other thing. You might also want to display the text domain when it's a value other than default.

demiankatz (changes requested):
Thanks, @ThoWagen -- I did a little bit of hands-on testing and found a couple of small remaining issues. See below. I also merged this with the latest dev branch to be sure I was testing the latest code, so you'll want to pull down my changes before making more of your own.

demiankatz:
One other thought: maybe this would be even more clear if the "language code" for this was "debug" rather than "keys".

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
ThoWagen: You are right. I changed this.

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
ThoWagen: I didn't think about this. Fixed it.

ThoWagen:
I renamed it to "debug". I think you are right. It is not really obvious what "keys" means on first glance.

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
```diff
@@ -104,6 +104,19 @@ public function translate($target, $tokens = [], $default = null)
         // Figure out the text domain for the string:
         [$domain, $str] = $this->extractTextDomain($target);
 
+        if ($this->getTranslatorLocale() == "debug") {
+            $targetString = $domain !== "default" ? "$domain::$str" : "$str";
```
demiankatz: Very minor nitpick, but you don't need double quotes around `"$str"` -- `$str` will do the same thing more efficiently. :-)

demiankatz: ...and if you want to be _really_ picky, it might also be slightly more efficient to use single quotes instead of double quotes for simple strings like `"debug"` that don't require interpolation. I _try_ to make a habit of this, though I don't claim that the code is currently 100% consistent. It would be nice to find a php-cs-fixer rule to do this for us someday! ;-)

demiankatz (changes requested):
Thanks, @ThoWagen -- some fairly trivial comments. I'm out of time for today but hope to get this merged tomorrow. If you're busy and would like me to make the suggested changes myself, I am happy to help when I have time to do more hands-on testing tomorrow... just passing them along to you in case you're able to get there first.

config/vufind/config.ini
```diff
@@ -1728,12 +1728,17 @@ skip_numeric = true
 ; directory.  The value of each setting is the on-screen name of the language,
 ; and will itself be subject to translation through the language files!
 ;
+; Enable keys to see the keys of the translation instead of a translation.
```
demiankatz: Also noticed that we need to change this comment to match the new `debug` name.

ThoWagen: done

module/VuFind/src/VuFind/I18n/Translator/TranslatorAwareTrait.php
ThoWagen: Cleaned this up. Although its debatable if single quotes are really more efficient its definitely nice to have this consistent. I will keep this in mind in the future.

demiankatz: Thanks, @ThoWagen. And thanks to #3035, this will soon be automated for us!

demiankatz (approved):
I've merged the latest dev branch into this and made a very small whitespace adjustment to config.ini; I think this is ready to merge now. Thanks again, @ThoWagen!

PR #3016, opened 2023-08-01
Remove deprecated tools (phpcpd/phploc).

demiankatz:
As discussed on the August, 2023 Community Call, two of our continuous integration tools (phpcpd and phploc) are no longer being maintained. There was no objection to discontinuing their use, as their benefit is limited and newer tools (like static code analysis) can meet many of the same needs.

If anyone has concerns about this removal, please comment here before August 15th. Otherwise, I will aim to finish this work up when I return from vacation.

TODO
- [x] Remove processing of output from Jenkins configuration before merging


demiankatz:
No one has objected to the removal, so I have gone ahead and turned off the phploc and phpcpd configurations for all of our Jenkins projects. It is now safe to merge this (and make equivalent changes across the subprojects) as soon as someone approves this PR.

demiankatz:
I also had a TODO note here to synchronize tool versions between the core project and the subprojects... but I have moved that note to [VUFIND-1609](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1609) since that seems like the more appropriate spot for the reminder.

maccabeelevine (approved):
Seems straightforward?

demiankatz:
Should be! Thanks, @maccabeelevine!

PR #3017, opened 2023-08-02
[VUFIND-1606] Update CookieConsent library to version 3.0.0-rc.16.

EreMaijala:
HTML validation is now enabled for the Mink tests as the validation issues have been fixed.

This version changes the order of buttons so that "Accept All" is the first one, so I had to adjust the tests, and decided to add a set of button click methods to make the tests more understandable while at it.


demiankatz (approved):
Thanks, @EreMaijala, looks good to me!

PR #3019, opened 2023-08-03
Fix API schema Authors component documentation

xmorave2:
Documentation says there are "main" author field, but it is "primary"

Also the authors field of record is defined as "array" but it is object of type "Author"


EreMaijala (approved):

EreMaijala:
@xmorave2 Thanks for catching that one!

PR #3021, opened 2023-08-03
[dev-10.0] Expose setupQRCodeLinks method.

EreMaijala:
setupQRCodeLinks used to be a global function, and calling it directly could still be needed in customized installations.


EreMaijala:
What I mean is that we need it in Finna. :)

demiankatz (approved):

PR #3023, opened 2023-08-03
Fix batch-import-marc.sh error if there no files to process

natecollins:
The use of xargs in multi-file processing to batch-import-marc.sh will always output at least one line by default, even if that line is empty. The result is that when no files are available to process (XML or MARC) then the script will attempt to call import-marc.sh with an empty value, resulting in an error.

Adding the -r flag to xargs avoides this error, preventing further commands from running if only blank data was received.


demiankatz (approved):
Thanks, @natecollins!

PR #3024, opened 2023-08-04
[VUFIND-1630] Alphabrowse: new normalizer for titles based on SolrMarc titleSortLower

damien-git:
This makes sure the same normalizer is used for indexing and browsing. It uses a new normalizer based on SolrMarc `titleSortLower`. The same field as before is used for indexing (`title_sort`).

TODO
- [x] Merge [related PR](https://github.com/vufind-org/vufind-browse-handler/pull/43) in vufind-browse-handler first.
- [x] Implement XML equivalent title_sort behavior
- [x] Add equivalent to SolrMarc cleanData in XML formatting logic
- [x] Make sure Windows .bat file works the same as Linux .sh file.
- [x] Add Mink test (if time permits)
- [x] Add changelog note (reindex needed for XML records)


index-alphabetic-browse.sh
```diff
@@ -31,11 +31,17 @@ then
   SOLR_HOME="$VUFIND_HOME/solr/vufind"
 fi
 
+# This would be /opt/bitnami/solr on a solr docker container
+if [ -z "$BITNAMI_SOLR_HOME" ]
```
demiankatz: Is this an intentional change? It seems unrelated to the rest of the PR. I'm open to supporting more Solr scenarios, but I think it might be confusing to foreground the Bitnami use case by using that as the primary environment variable name. Maybe the thing to do would be to create a new SOLR_JAR_PATH environment variable, and set that to BITNAMI_SOLR_HOME when that variable is set and the other value otherwise.

demiankatz (changes requested):
Thanks for getting this started! In addition to the issue highlighted below, two other questions/comments:

1.) Why is solrmarc_core.jar included in this PR? Did you mean to replace the browse handler jars instead?

2.) We'll need to implement equivalent XML processing to avoid breaking browse headings related to XML input. Right now, all of our XSLTs have lines like this:

https://github.com/vufind-org/vufind/blob/dev/import/xsl/dspace.xsl#L140

I think we need to add a PHP version of titleSortLower to [\VuFind\XSLT\Import\VuFind](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/src/VuFind/XSLT/Import/VuFind.php) and modify the XSLT examples to wrap this around the existing stripArticles call.

If you don't want to spend time on this because it's not relevant to your use cases, we can add a TODO checkbox and I'll try to find time to help with it when I'm back from vacation!

index-alphabetic-browse.sh
todolson: Agreed that it seems confusing to key this to Bitnami explicitly. And for non-Bitnami sites, the `CLASSPATH` below could have unexpected side effects.

index-alphabetic-browse.sh
```diff
@@ -31,11 +31,17 @@ then
...
+then
+  BITNAMI_SOLR_HOME="${SOLR_HOME}/../vendor"
+fi
+
 set -e
 set -x
 
 cd "`dirname $0`/import"
-CLASSPATH="browse-indexing.jar:${VUFIND_HOME}/import/lib/*:${SOLR_HOME}/jars/*:${SOLR_HOME}/../vendor/modules/analysis-extras/lib/*:${SOLR_HOME}/../vendor/server/solr-webapp/webapp/WEB-INF/lib/*"
+CLASSPATH="browse-indexing.jar:${VUFIND_HOME}/import/lib/*:${SOLR_HOME}/jars/*:${BITNAMI_SOLR_HOME}/modules/analysis-extras/lib/*:${BITNAMI_SOLR_HOME}/server/solr-webapp/webapp/WEB-INF/lib/*:${BITNAMI_SOLR_HOME}/server/lib/ext/log4j*.jar"
```
todolson: Related to the above comment, I'm a little concerned about having the `BITNAME_SOLR_HOME` wired into `CLASSPATH` so explicitly, especially if something like log4j needs to be loaded from a special directory. The idea for a `SOLR_JAR_PATH` might be more flexible.

todolson (commented):
I'm not yet able to test the main issue of the new normalizer, but see in-file comments about BITNAMI_SOLR_HOME.

damien-git:
It's my turn to be on vacation, and I won't be able to address all the questions for a few days. I am fine with naming this variable differently, SOLR_JAR_PATH sounds good.
The last entry for the CLASSPATH should probably be ${BITNAMI_SOLR_HOME}/server/lib/ext/* instead (because log4j*.jar might not work).
The change to use a new env variable was intentional, because we use Solr separately from Vufind (in a different container) and we currently have to recreate a special environment with a vendor link to make things work without changing the scripts. It would be nice to be able to use the indexing script without the Vufind structure dependency.

demiankatz:
@damien-git, would it make sense to set up SOLR_JAR_PATH by first checking for the existence of $VUFIND_HOME/solr/vendor, and then failing over to $BITNAMI_SOLR_HOME as a possible next resort?

I don't think using a more generic search path is a huge problem, though if I'm remembering correctly, we may have used the log4j*.jar pattern because using the entire directory was slowing things down by searching too many unneeded jars.

In any case, I hope you have a great vacation, and we can resume this conversation when we're both back in the office!

index-alphabetic-browse.sh
damien-git: > 1.) Why is solrmarc_core.jar included in this PR? Did you mean to replace the browse handler jars instead?

It needs to be in the classpath for both indexing and browsing. I did not change the browse handler jar because I assumed you do that before every release (I can include it if you wish).

> 2.) We'll need to implement equivalent XML processing to avoid breaking browse headings related to XML input. Right now, all of our XSLTs have lines like this:

I see, yes, we would need a local implementation for non-MARC input. This goes beyond what I intended to do in the scope of resolving this issue. I could possibly work on it later, but not right away.

damien-git:
> would it make sense to set up SOLR_JAR_PATH by first checking for the existence of $VUFIND_HOME/solr/vendor, and then failing over to $BITNAMI_SOLR_HOME as a possible next resort?

I think it makes sense to use `SOLR_JAR_PATH` if it is defined, and use `$VUFIND_HOME/solr/vendor` otherwise, which is what the script does now. Because `SOLR_JAR_PATH` is not used for anything else, so if it is defined we can assume users want it to be used for this script.

> I don't think using a more generic search path is a huge problem, though if I'm remembering correctly, we may have used the log4j*.jar pattern because using the entire directory was slowing things down by searching too many unneeded jars.

I think `log4j*.jar` would not work, as Java only understands `*` for names in the classpath. It might slow things down a bit, yes. I guess we could improve the `CLASSPATH` creation by looking for `log4j*.jar` in bash and assemble the related `CLASSPATH` part. _Edit: I implemented that._

index-alphabetic-browse.sh
demiankatz: 1.) SolrMarc is already in the import directory in VuFind -- I think we should avoid adding a duplicate copy if at all possible. Let me know if you need help figuring out how to make it accessible in the appropriate context.

2.) Once the chaos of August calms down, I should be able to help with the non-MARC parts. Let me know if you get there first, but otherwise I'll help as soon as I can. :-)

demiankatz: Also, regarding updating the browse handlers, if this PR depends on new browser handler jars, it probably makes sense to include the update here to enable testing... though if the PR in the vufind-browse-handler gets merged first, we could also do the update separately in advance of merging this PR.

damien-git: I think `solrmarc_core.jar` needs to be in `solr/vufind/jars` in the release. How could I copy the file from `import` ? Would `build.xml` be a good place for that ?

demiankatz: You should be able to add it to the classes included in solrconfig.xml without making an extra copy. Not at a computer right now, but I can dig up the details tomorrow if I'm being too vague right now. :-)

damien-git: I see, there is already that in `solrconfig.xml`:
`<lib dir="../../../import/lib" regex="marc4j.*\.jar" />`
So I could add .../import for solrmarc. The dependency of the solr config to vufind is not ideal for us because we run them separately, but this change won't affect that dependency much. I will also have to add `solrmarc_core.jar` to the classpath in `index-alphabetic-browse.sh`.

index-alphabetic-browse.sh
```diff
@@ -31,11 +31,18 @@ then
...
+  SOLR_JAR_PATH="${SOLR_HOME}/../vendor"
+fi
+
 set -e
 set -x
 
 cd "`dirname $0`/import"
-CLASSPATH="browse-indexing.jar:${VUFIND_HOME}/import/lib/*:${SOLR_HOME}/jars/*:${SOLR_HOME}/../vendor/modules/analysis-extras/lib/*:${SOLR_HOME}/../vendor/server/solr-webapp/webapp/WEB-INF/lib/*"
+LOG4J_CLASSPATH=$(find "${SOLR_JAR_PATH}/server/lib/ext" -name 'log4j*.jar' -printf '%p:' | sed 's/:$//')
+CLASSPATH="browse-indexing.jar:$(echo solrmarc_core*.jar):${VUFIND_HOME}/import/lib/*:${SOLR_HOME}/jars/*:${SOLR_JAR_PATH}/modules/analysis-extras/lib/*:${SOLR_JAR_PATH}/server/solr-webapp/webapp/WEB-INF/lib/*:${LOG4J_CLASSPATH}"
```
demiankatz: Is `$(echo solrmarc_core*.jar)` going to cause problems if there are multiple files matching the wildcard pattern in the directory? Obviously, this SHOULDN'T ever happen, but sometimes if people upgrade without following all the recommended best practices, they can end up with multiple copies of SolrMarc, so it's not impossible. (Maybe we don't care, because when this happens, it will also cause other problems and needs to be fixed... but just wondering if we should consider making it more robust).

demiankatz (changes requested):
Thanks, @damien-git, this is looking good to me.

I've just merged https://github.com/vufind-org/vufind-browse-handler/pull/43, so this PR can be the place where we pull the updated browse handler code into the dev branch. I notice that you've only updated the browse handler jar here, though; should the browse indexer also be updated?

Also, one other question below...

index-alphabetic-browse.sh
damien-git: Sorry I forgot about browse-indexing.jar, I will add it.
About the potential multiple solrmarc jars, what would happen if there were several is that the JVM would pick the first one listed (so probably the oldest). It could be a problem, but I feel like this is an issue with upgrades, and if we want to handle that anywhere it should be with the upgrade process.

demiankatz: My concern is that echo with a wildcard will create a space delimited list instead of a colon delimited list when multiple matches are found,  which might break the whole class path. Could we pipe it to something to just grab the first value to prevent this? Or is that concern unfounded?

damien-git: Ah, yes, it will not create a colon delimited list, and it will break... Which might be better than a silent version mismatch, but ideally we should check there is only one matching jar and fail with exit 1 otherwise.

demiankatz: I guess we could direct the output of the echo to an environment variable, and use wc to count the elements and fail with an error message if it's not equal to 1.

damien-git: Yes, I was going to do that but I got distracted by pressing bugs. I'll see you at WOLFCON.

demiankatz: Thanks! I probably won't have a chance to review and merge this until after the conference, but I'm glad to hear you'll be there -- I look forward to meeting you in person!

index-alphabetic-browse.sh
```diff
@@ -31,11 +31,24 @@ then
...
+then
+  SOLR_JAR_PATH="${SOLR_HOME}/../vendor"
+fi
+
 set -e
 set -x
 
 cd "`dirname $0`/import"
-CLASSPATH="browse-indexing.jar:${VUFIND_HOME}/import/lib/*:${SOLR_HOME}/jars/*:${SOLR_HOME}/../vendor/modules/analysis-extras/lib/*:${SOLR_HOME}/../vendor/server/solr-webapp/webapp/WEB-INF/lib/*"
+LOG4J_CLASSPATH=$(find "${SOLR_JAR_PATH}/server/lib/ext" -name 'log4j*.jar' -printf '%p:' | sed 's/:$//')
```
demiankatz: Why are we explicitly adding Log4j to the classpath? It doesn't seem that it was needed in the previous code. Is this dependency added as a result of introducing SolrMarc to the class path? If so, I'm a bit concerned, because SolrMarc currently uses Log4j 1.x and Solr uses Log4j 2.x. Also just wondering if there's an opportunity to simplify changes.

demiankatz (commented):
Sorry for the long delay getting back to this -- I'm _still_ catching up on work that piled up during WOLFcon! It looks like most of the issues under discussion have been resolved, but I had one more question about the existing code when reviewing the state of things. Once we settle this conversation, I'll work on updating the Windows .bat file to match. In the meantime, if time permits, I'll get started on updating the XML indexing logic to match.

demiankatz:
@damien-git, I've just implemented XML title_sort normalization. While I obviously wouldn't guarantee that this is 100% equivalent to what SolrMarc is doing, it's at least a great deal closer and much less likely to result in browse handler inconsistencies. Feedback on my implementation is welcome!

index-alphabetic-browse.sh
damien-git: Yes, it's a dependency. It crashes without Log4j on the classpath. I don't think we could simplify changes much. We could optimize further by checking which of the Log4j jars is really needed (I did not test that).
I did not realize the difference in versions, but since the classpath did not include Log4j before, it is safe to say that for this execution it will only be used by SolrMarc.

damien-git:
Great, we had put that into a future sprint, but I will look at your implementation instead.

damien-git:
@demiankatz OK, I took a look. I am not sure how close you are trying to be with the SolrMarc implementation, but I think we could improve 2 things:
- I am not seeing an equivalent of [cleanData(String origStr)](https://github.com/solrmarc/solrmarc/blob/be27d3bb5d601d5aa1786e01c17e153c1e86e03a/src/org/solrmarc/tools/DataUtil.java#L142) in your implementation.
- stripAccents: turning all the string into ASCII would remove a lot of characters, it would be better to replace only the ones we can recognize as having accents and leave the other characters untouched.
For implementation ideas, see for instance https://stackoverflow.com/questions/1017599/how-do-i-remove-accents-from-characters-in-a-php-string

demiankatz:
@damien-git, thanks for taking another look at this!

A few comments:

1.) I failed to find the cleanData method when I was trying to make sense of the SolrMarc code. I couldn't figure out what eCleanVal was doing, and it seems you've found the link that I missed. I'll close this gap in the PHP version as soon as time permits!

2.) The stripAccents function is not turning a string entirely into ASCII -- the Latin-ASCII transformer only selects Latin accented characters and converts those to their ASCII equivalents while leaving everything else alone. One of my UNIT tests applies the transform to Japanese text to confirm that it is unaffected. There are probably still ways we could do better, but I think for common Latin accent scenarios, the transformation should be equivalent.

3.) I'm still a bit confused about the Log4j situation. In recent VuFind releases, SolrMarc's Log4j has been replaced with Reload4j (which maintains the 1.x API but backports security fixes). This is a temporary measure until SolrMarc upgrades to Log4j 2.x. However, all the necessary .jar files for this are found in import/lib -- so I'm not sure why extra work should be needed for Log4j jars if the import/lib folder is on the classpath. I'm quite short on time this week, but if you need me to do any experimentation around this, let me know and I will try as soon as time permits.

damien-git:
OK. Regarding 3) I was not aware that reload4j was implementing the Log4j API. We run our code on a container separate from VuFind, so we copy only the jars we need into a temporary "import/lib" structure. Currently we are only copying marc4j from there. It is entirely possible that it will work with reload4j, and that we can simplify the classpath and remove Log4j. I will try that as soon as I can.

demiankatz:
Thanks, @damien-git -- fingers crossed that that helps. :-)

damien-git:
Yes, it worked. I will update the PR accordingly.

demiankatz:
Thanks for the updates, @damien-git -- I should be able to work on the Windows port of the import script tomorrow.

In the meantime, I've added a port of the cleanData logic in commit [beff9a5](https://github.com/vufind-org/vufind/pull/3024/commits/beff9a58697318ff0964214cf0fe5f59680d7f7f). I did a fairly literal port of the SolrMarc logic, even though in the context of titleSortLower, many of the subtleties get wiped out anyway. I haven't taken the time to carefully test the individual support methods, but I think the overall effect should be reasonable in most cases. I certainly welcome the addition of further test cases if anyone has time and thinks it's worthwhile.

damien-git:
Great, thanks !

demiankatz:
@damien-git, I have finished porting the index-alphabetic-browse logic to the Windows .bat file, so I think this is now functionally complete. I might try to add another Mink test case if time permits, if there's something in the sample data we can use to reproduce the problem (and thus prevent future regressions). I've asked @EreMaijala for a review as well, in case he has any concerns about how this might interact with his [RecordManager](https://github.com/NatLibFi/RecordManager) ecosystem.

index-alphabetic-browse.bat
```diff
@@ -43,8 +43,29 @@ goto javaset
 set JAVA="%JAVA_HOME%\bin\java"
 :javaset
 
+rem This would be /opt/bitnami/solr on a solr docker container
```
EreMaijala: I think it would be better to say this in a more generic way like "This can point to an external Solr in e.g. a Docker container" (also avoids documenting a current practice of a single vendor).

index-alphabetic-browse.sh
```diff
@@ -31,11 +31,23 @@ then
   SOLR_HOME="$VUFIND_HOME/solr/vufind"
 fi
 
+# This would be /opt/bitnami/solr on a solr docker container
```
EreMaijala: Same as in the .bat file.

EreMaijala (changes requested):
I think the Solr jar finding bit would warrant a separate commit, but other than that this looks ok to me. Just a small comment about a comment. :)

demiankatz:
Thanks, @EreMaijala! I've opened #3088 to extract the SOLR_JAR_PATH logic into a separate commit and revise the comment to be more generic as you suggested. Once that's merged, I'll resolve the conflicts here.

EreMaijala (approved):
Looks good to me!

demiankatz (approved):
I think this is ready to go now, though I'll wait for @EreMaijala's double-check before merging.

As promised, I added a few more tests to the Mink suite for AlphaBrowse. We could certainly still do a more thorough job of testing, especially if we added some test records to specifically highlight specific behaviors. Most of the tests I added pass both here and on the dev branch, so I'm not necessarily proving a whole lot. However, the "accent stripping" case works here and fails on dev, so we do have at least some confirmation that this version improves on the dev version.

PR #3025, opened 2023-08-05
Add new Overdrive features

bpalme:
OK here are the changes to satisfy the two issues referenced in the title and bring the Overdrive integration back into working order. One things to note... I went a different direction on the core template. Previously we had some of the holdings information in the core which was somewhat clumsy to make it work by setting some variable with HTML comment and then calling the parent core tab. I elected to make it more like other content and moved all of the holding info to a holdings tab. I think that will make more sense and make the integration more uniform. If you are OK with that, then it might make sense to remove the call in the main template to show the contents of those variables. I hope I didn't lose all the formatting fixes that were done in the intervening time. I tried to fold those in, but I doubt I got all of them.

TODO
- [x] Resolve [VUFIND-1499](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1499) and [VUFIND-1375](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1375) when merging.


demiankatz:
@bpalme, since development work on release 9.1 will be wrapping up soon, I'm moving this to the 10.0 milestone. If you get the work done very soon, I'm still open to merging it into the earlier release, but since I imagine it may take a while longer to get all the details right, I think 10.0 is probably the most realistic version target. :-)

bpalme:
Ok I think  that makes sense.

demiankatz:
@bpalme, just a quick update to say that I've marked this PR "on hold" until you are ready, since there is nothing to review until you are able to share your latest work. No rush, but let me know when you'd like me to remove that label.

demiankatz:
@bpalme, thanks for making some progress here! I just took some time to resolve all the conflicts and get this properly based against the dev branch. I hope I didn't break anything in the process! Could you please review this when you get a chance and correct anything that I may have messed up? Once you're happy that this is complete and functional, I can give it a closer review from my end.

demiankatz:
@bpalme, I notice that there is some kind of parsing error in `themes/bootstrap3/templates/RecordTab/overdriveholdings.phtml` -- any idea what that's about?

bpalme:
Thank you so much. I'll try to make some progress on it soon.

Brent

config/vufind/Overdrive.ini
```diff
@@ -36,6 +36,10 @@ noAccessString = ""
 ; more overhead with it on. Default is false.
 consortiumSupport = false
 
+;AJAX Status is supported, but you can turn it off it's slowing the search results
```
demiankatz: I think this comment is missing an "if" between "off" and "it's." We also by convention include spaces after the `;` in comments.

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -250,8 +251,9 @@ public function getPreviewLinks()
      */
     public function supportsAjaxStatus()
     {
-        // TODO: add this as an Overdrive configuration to turn it off
-        return true;
+        $this->debug('ajax status:' . $this->config->enableAjaxStatus);
+        $this->debug('all configs:' . print_r($this->config, true));
```
demiankatz: Do we need all these debug outputs here?

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -331,24 +333,34 @@ public function getOverdriveAvailability()
...
+        $this->debug('isCheckedOut', [], true);
         $overdriveID = $this->getOverdriveID();
         $result = $this->connector->getCheckouts(true);
         if ($result->status) {
             $checkedout = false;
             $checkouts = $result->data;
+            $result->data = [];
             foreach ($checkouts as $checkout) {
-                if (strtolower($checkout->reserveId) == $overdriveID) {
+                $this->debug("comparing: {$checkout->reserveId} to $overdriveID");
```
demiankatz: This debug message seems potentially outdated, since you're doing something more nuanced now.

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -563,4 +575,41 @@ public function getRealTimeTitleHold()
...
+     *   <ul>queryString: Query params to append after building route (optional)</ul>
+     * </li>
+     *
+     * @return array
+     */
+    public function getURLs()
+    {
+        $permlink =  $this->getPermanentLink();
+        //todo translate
+        $desc = 'Overdrive Resource Page';
```
demiankatz: Record drivers have access to the translator, so this TODO is not hard at all -- just make sure the string is in the language file, and change this to `$desc = $this->translate(...)`. Additionally, if you're hard-coding $desc, there's no need to do the `?: $permlink` in the next line, since $desc will never be false.

module/VuFind/src/VuFind/RecordTab/OverdriveHoldingsFactory.php
```diff
@@ -0,0 +1,62 @@
...
+     * @return object
+     *
+     * @throws ServiceNotFoundException if unable to resolve the service.
+     * @throws ServiceNotCreatedException if an exception is raised when
+     * creating a service.
+     * @throws ContainerException if any other error occurs
+     */
+    public function __invoke(\Interop\Container\ContainerInterface $container, $requestedName, array $options = null)
+    {
+        return new \VuFind\RecordTab\OverdriveHoldings();
```
demiankatz: If you're just creating the object with no extra parameters, there is no need to build a whole factory. You can just configure the plugin manager to use `\Laminas\ServiceManager\Factory\InvokableFactory` to build the object. And speaking of that, I don't see any changes to register this new class in the PluginManager. I wonder if I accidentally lost those changes when resolving conflicts. Please let me know if you need help getting them back!

module/VuFind/src/VuFind/DigitalContent/FakeOverdriveConnector.php
```diff
@@ -220,31 +298,50 @@ public function returnResource($resourceID)
...
-    public function getDownloadLink($overDriveId, $format, $errorURL)
+    public function getPermanentLinks($overDriveIds = [])
     {
-        return new \stdClass();
+        return [];
     }
 
     /**
-     * Lock In Overdrive Resource for a particular format
+     * Returns
```
demiankatz: This description doesn't seem right for some reason.

module/VuFind/src/VuFind/DigitalContent/OverdriveConnector.php
```diff
@@ -873,12 +1057,42 @@ public function getMetadata($overDriveIds = [])
...
+    /**
+     * For  array of titles passed in this will return the same array
+     * with metadata attached to the records with the property name of 'metadata'
+     *
+     * @param array $overDriveTitles Assoc array of objects with OD IDs as keys (generally what
+     *                               you get from getCheckouts and getHolds)
+     *
+     * @return array inital array with results of metadata attached as "metadata" property
+     *
+     * @todo if more tan 25 passed in, make multiple calls
```
demiankatz: Typo "tan" for "than"

module/VuFind/src/VuFind/DigitalContent/FakeOverdriveConnector.php
```diff
@@ -253,12 +350,30 @@ public function lockinResource($overDriveId, $format)
      * @param array $overDriveIds Set of Overdrive IDs
      *
      * @return array results of metadata fetch
+     *
+     * @todo if more tan 25 passed in, make multiple calls
```
demiankatz: Do we need this todo in the fake version?

module/VuFind/src/VuFind/DigitalContent/FakeOverdriveConnector.php
```diff
@@ -253,12 +350,30 @@ public function lockinResource($overDriveId, $format)
...
+    /**
+     * For  array of titles passed in this will return the same array
+     * with metadata attached to the records with the property name of 'metadata'
+     *
+     * @param array $overDriveTitles Assoc array of objects with OD IDs as keys (generally what
+     *                               you get from getCheckouts and getHolds)
+     *
+     * @return array inital array with results of metadata attached as "metadata" property
+     *
+     * @todo if more tan 25 passed in, make multiple calls
```
demiankatz: Do we need this todo in the fake version?

module/VuFind/src/VuFind/DigitalContent/FakeOverdriveConnector.php
```diff
@@ -253,12 +350,30 @@ public function lockinResource($overDriveId, $format)
...
     }
 
+    /**
+     * For  array of titles passed in this will return the same array
+     * with metadata attached to the records with the property name of 'metadata'
+     *
+     * @param array $overDriveTitles Assoc array of objects with OD IDs as keys (generally what
+     *                               you get from getCheckouts and getHolds)
+     *
+     * @return array inital array with results of metadata attached as "metadata" property
```
demiankatz: Typo: inital

module/VuFind/src/VuFind/DigitalContent/OverdriveConnector.php
```diff
@@ -873,12 +1057,42 @@ public function getMetadata($overDriveIds = [])
...
     }
 
+    /**
+     * For  array of titles passed in this will return the same array
+     * with metadata attached to the records with the property name of 'metadata'
+     *
+     * @param array $overDriveTitles Assoc array of objects with OD IDs as keys (generally what
+     *                               you get from getCheckouts and getHolds)
+     *
+     * @return array inital array with results of metadata attached as "metadata" property
```
demiankatz: Typo: inital

module/VuFind/src/VuFind/DigitalContent/OverdriveConnector.php
```diff
@@ -843,9 +953,82 @@ public function getFormatNames()
...
+                $result->status = true;
+                $result->data = (object)[];
+                $result->data = $response;
+            }
+
+            if ($checkouts) {
+                $checkoutResult = $this->getCheckouts();
+                $checkoutData = $checkoutResult->data;
+                foreach ($result->data->products as $key => $issue) {
+                    if (isset($checkoutData[strtolower($issue->id)])) {
```
demiankatz: This if/else could be simplified to:

```
$issue->checkedout = isset($checkoutData[strtolower($issue->id)]);
```

demiankatz (changes requested):
Thanks for the progress here, @bpalme! There's a lot of code to review, so I haven't looked thoroughly at everything yet, but I thought I'd get the ball rolling with a few comments based on an initial quick scan of what has changed. See below!

Please let me know if/when you'd like help with hands-on testing as well! Obviously my ability to test is limited since I do not have access to OverDrive, but with the fake connector, I can at least do some local testing. I just don't want to start that before you think the code is ready to run, so please keep me informed of the status. :-)

demiankatz:
@bpalme, I also resolved the latest round of conflicts while I was looking at this anyway. (Nothing complicated -- just had to recompile the CSS).

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
bpalme: No. Should I generally go through all the comments and trim it down to the most useful?

bpalme:
> Thanks for the progress here, @bpalme! There's a lot of code to review, so I haven't looked thoroughly at everything yet, but I thought I'd get the ball rolling with a few comments based on an initial quick scan of what has changed. See below!
> 
> Please let me know if/when you'd like help with hands-on testing as well! Obviously my ability to test is limited since I do not have access to OverDrive, but with the fake connector, I can at least do some local testing. I just don't want to start that before you think the code is ready to run, so please keep me informed of the status. 

I didn't create the FakeConnector so I didn't pay much attention to it but I'll clean it up now. I've been meaning to ask if I should try to flush it out more and change the return data to be more realistic.

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
demiankatz: It wouldn't hurt. As a rule of thumb: if it's going to help somebody who's having trouble with Overdrive, feel free to leave it. If it's something you needed during development but no longer expect to require, get rid of it. :-)

demiankatz:
> I didn't create the FakeConnector so I didn't pay much attention to it but I'll clean it up now. I've been meaning to ask if I should try to flush it out more and change the return data to be more realistic.

I created the FakeConnector because I needed to test and refactor some templates, and I needed a way to be able to see them without OverDrive. If you can easily make it more complete and realistic, that would certainly be helpful, especially if you want my assistance with deeper testing. However, I don't want to make extra work for you if you don't think this is a good use of your time, and I can always add fake data on my end based on guesswork if I need to test something -- the results just might not be as good as if you do it with your more informed perspective. :-)

module/VuFind/src/VuFind/RecordTab/OverdriveHoldingsFactory.php
bpalme: Hmm. I don't think I ever put it in the PluginManager but it seems to be working without it listed. But I added it and removed the unnecessary factory and that works too.

demiankatz: Thanks! The plugin managers have some default behavior for building objects when configuration is missing, so that's probably why this worked before... but I prefer to be explicit about configuration when possible, so thanks for adding the missing details!

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
bpalme: Remove quite a few debug statements

bpalme:
> If you can easily make it more complete and realistic, that would certainly be helpful, especially if you want my assistance with deeper testing. However, I don't want to make extra work for you if you don't think this is a good use of your time, and I can always add fake data on my end based on guesswork if I need to test something -- the results just might not be as good as if you do it with your more informed perspective. :-)

I was planning on capturing some of the actual returned data for a few real titles and putting that into the test but even serialized it was quite large so I didn't think you would want that cluttering up the code. So I focussed more just getting it done and tested manually.

bpalme:
For the changes in PR #3244, I've incorporated those here (One template was removed).

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/status-full.phtml
```diff
@@ -17,26 +17,38 @@ foreach ($ids as $id) {
...
+            $current['full_status'] .= '<span class="text-danger">';
+            $current['full_status'] .= $this->icon('status-unavailable', 'icon-link__icon');
+            $current['full_status'] .= '</span> ';
+        }
+        if ($avail->availabilityType == 'AlwaysAvailable') {
+            $current['full_status'] .= $this->transEsc('od_unlimited');
+        } else {
+            $current['full_status'] .= $avail->copiesAvailable . ' / ' . $avail->copiesOwned;
+        }
+        $current['full_status'] .= ' ' . $this->transEsc('od_copies_available');
```
demiankatz: Rather than concatenating strings here, it would probably be better to change the od_copies_available string to something like `%%copies%% copies available` and then change this code to something like:

```
$copies = $avail->availabilityType == 'AlwaysAvailable'
    ? $this->transEsc('od_unlimited');
    : ($avail->copiesAvailable . ' / ' . $avail->copiesOwned);
$current['full_status'] .= ' ' . $this->transEsc('od_copies_available', ['%%copies%%' => $copies]);
```

...and then languages that put things in different orders can adjust things appropriately in translation.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/status-full.phtml
```diff
@@ -17,26 +17,38 @@ foreach ($ids as $id) {
...
     $current['status'] = 'OK';
     $current['id'] = $id;
-    $avail = $this->result->data[$id] ?? new \stdClass();
+    $avail = $this->result->data[$id] ?? (object)['code' => ''];
+    $current['full_status'] = '';
 
     if ($noshow) {
-        $current['full_status'] = '';
-    } elseif ($isError || $avail->code == 'od_code_resource_not_found') {
+        //patron does not have access to overdrive. don't show avail
```
demiankatz: Minor style: please put a space after `//` in comments.

themes/bootstrap3/templates/RecordTab/overdriveholdings.phtml
```diff
@@ -0,0 +1,182 @@
...
+        <?php endif; ?>
+
+        <?php if (isset($avail->data->availabilityType) && $avail->data->availabilityType == 'AlwaysAvailable'): ?>
+          <?= $this->transEsc('od_unlimited') ?>
+        <?php else: ?>
+          <?= $avail->data->copiesAvailable . '/' . $avail->data->copiesOwned ?>
+        <?php endif; ?>
+        <?= $this->transEsc('od_copies_available') ?>
+        <?php if ($avail->data->numberOfHolds > 0): ?>
+          <?= '; ' . $avail->data->numberOfHolds . ' ' . ($avail->data->numberOfHolds > 1 ? $this->transEsc('od_people') : $this->transEsc('od_person')) . ' ' . $this->transEsc('od_waiting') ?>
```
demiankatz: This looks like an opportunity to use our recently-added support for ICU message strings. Instead of separate od_people/od_person strings, I'd suggest changing this to:

```
<?= '; ' . $this->transEsc('od_waiting', ['holds' => $avail->data->numberOfHolds], null, true) ?>
```

And then you can make the od_waiting translation string:

```
od_waiting = "{holds, plural, =1 {1 person} other {# people}} waiting"
```

This gives more fine-grained control over pluralization support in languages with more complex requirements.

languages/en.ini
```diff
@@ -928,63 +928,92 @@ number_decimal_point = "."
...
+od_people = "People"
+od_person = "Person"
+od_resource_page = "OverDrive Resource Page"
+od_return_confirm = "Are you sure you want to return this title?"
 od_return_failure = "This title could not be returned."
 od_return_success = "This title has been returned."
-od_video-streaming = "streaming video file"
+od_susp_after = "Deliver after %%days%% days"
+od_susp_asap = "Deliver as soon as possible"
+od_susp_deliver = "Deliver"
```
demiankatz: The `od_susp_deliver` string does not appear to be used. Is this intentional?

module/VuFind/src/VuFind/RecordTab/OverdriveHoldings.php
```diff
@@ -0,0 +1,54 @@
...
+ * Summary of OverdriveHoldings
+ *
+ * @category VuFind
+ * @package  RecordTabs
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Brent Palmer <brent-palmer@icpl.org>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development:plugins:record_tabs Wiki
+ */
+class OverdriveHoldings extends AbstractBase implements TabInterface
```
demiankatz: You don't need the `implements TabInterface` here since that is inherited from AbstractBase automatically.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/previewlink.phtml
```diff
@@ -0,0 +1,5 @@
+<?php
+$previews = $this->driver->getPreviewLinks();
+if ($previews->url): ?>
+    <a href="<?=$previews->url?>>" target="_blank" class="btn btn-info btn-xs"><?=$this->transEsc('Preview')?> </a>
```
demiankatz: Should there be a `$this->escapeHtmlAttr()` around `$previews->url`? Is the blank space really wanted before the ` </a>`?

demiankatz (changes requested):
Thanks for the latest progress, @bpalme! I haven't had a chance to exhaustively review everything yet, but I spent some time today doing a few things:

1.) I merged the latest dev branch and resolved a few minor conflicts.
2.) I updated all of the language files to use the correct "OverDrive" capitalization instead of "Overdrive."
3.) I fixed a couple of small typos.
4.) I left a few comments (see below), mostly focused on internationalization.

I may do another review and provide further feedback on other files later if time permits. In the meantime, please let me know if you have any questions about, or need any help with, this current round of feedback.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/status-full.phtml
```diff
@@ -17,26 +17,36 @@ foreach ($ids as $id) {
...
+            $current['full_status'] .= '<span class="text-success">';
+            $current['full_status'] .= $this->icon('status-available', 'icon-link__icon');
+            $current['full_status'] .= '</span> ';
+        } else {
+            $current['full_status'] .= '<span class="text-danger">';
+            $current['full_status'] .= $this->icon('status-unavailable', 'icon-link__icon');
+            $current['full_status'] .= '</span> ';
+        }
+        $copies = $avail->availabilityType == 'AlwaysAvailable'
+            ? $this->transEsc('od_unlimited');
```
demiankatz: The semi-colon at the end of this line should be removed; it's causing a syntax error.

demiankatz (changes requested):
Thanks for the ongoing progress, @bpalme -- I just happened to notice there was a continuous integration failure here, and it appears to be caused by the minor issue below.

themes/bootstrap3/templates/RecordTab/overdriveholdings.phtml
```diff
@@ -0,0 +1,182 @@
...
+          <span class="text-success">
+            <?= $this->icon('status-available', 'icon-link__icon') ?>
+          </span>
+        <?php else: ?>
+          <span class="text-danger">
+            <?= $this->icon('status-unavailable', 'icon-link__icon') ?>
+          </span>
+        <?php endif; ?>
+
+        <?php if (isset($avail->data->availabilityType) && $avail->data->availabilityType == 'AlwaysAvailable'): ?>
```
demiankatz: Looks like this code needs to be revised to match the edits in status-full.phtml related to od_copies_available. Also, in en.ini, it looks like od_copies_available needs to have the `%%copies%%` token added to the front of it.

module/VuFind/src/VuFind/Controller/OverdriveController.php
```diff
@@ -99,11 +99,38 @@ public function mycontentAction()
                 $checkoutsUnavailable = true;
             } else {
                 foreach ($checkoutResults->data as $checkout) {
+                    $mycheckout = [];
```
demiankatz: You could shave off a line of code by replacing these two lines with:

`$mycheckout = compact('checkout');`

That will initialize $mycheckout to ['checkout' => $checkout] for you.

module/VuFind/src/VuFind/Controller/OverdriveController.php
```diff
@@ -258,41 +333,114 @@ public function holdAction()
...
+                    $result->code = 'od_hold_susp_indef';
+                } else {
+                    $result->code = 'od_hold_redelivery';
+                    $result->codeParams = ['%%days%%' => $result->data->holdSuspension->numberOfDays];
+                }
+            } else {
+                $result->code = 'od_hold_update_failure';
+            }
+        } elseif ($action == 'editSuspendedHold') {
+            $actionTitleCode = 'od_susp_hold_edit';
```
demiankatz: This method is very long; it might be easier to read, troubleshoot and extend if it could be broken down into smaller parts. I certainly don't insist that you do this right now, but it might be helpful. For example, might it be possible to break down this long if/else statement into something like this?

```
$actions = [
    'editSuspendedHold' => ['titleCode' => 'od_susp_hold_edit', 'resultsMethod' => 'getEditSuspendedCodeResult'],
    // other action definitions go here...
];
if (isset($actions[$action])) {
    $actionTitleCode = $actions[$action]['titleCode'];
    $result = ($this->{$actions[$action]['resultsMethod']})();
} else {
    // action not defined logic goes here
}
```

And then you end up with a whole lot of small action processing methods instead of one big one.

Maybe there are reasons it's not as simple as this, but just an idea. If you like the idea but don't have time to deal with this right now, I'm happy to get this merged as-is and can help with refactoring in the future, as long as you can help with testing. :-)

module/VuFind/src/VuFind/DigitalContent/FakeOverdriveConnector.php
demiankatz: These todos are still here, though my comments about them are resolved. Should they be removed?

module/VuFind/src/VuFind/DigitalContent/OverdriveConnector.php
```diff
@@ -263,7 +262,7 @@ public function getAvailability($overDriveId)
             $availabilityUrl .= "$overDriveId/availability";
             $res = $this->callUrl($availabilityUrl);
 
-            if ($res->errorCode == 'NotFound') {
+            if (isset($res->errorCode) and $res->errorCode == 'NotFound') {
```
demiankatz: You can simplify some conditions in this class by using null coalescing instead of isset checks. For example, this could become:

`if (($res->errorCode ?? '') == 'NotFound') {`

module/VuFind/src/VuFind/DigitalContent/OverdriveConnector.php
```diff
@@ -1037,34 +1242,50 @@ public function getHolds($refresh = true)
...
-                            && $hold->holdListPosition == 1
-                        ) {
-                            $result->data[$key]->holdReadyForCheckout = true;
-                            // format the expires date.
-                            $holdExpires = new \DateTime($hold->holdExpires);
-                            $result->data[$key]->holdExpires
-                                = $holdExpires->format(
+
+                    if (isset($response->holds)) {
+                        $result->data = [];
```
demiankatz: You're assigning values to `$result->data` in two consecutive lines... since the empty array assigned in the first line is immediately overwritten, you don't need the first line at all.

module/VuFind/src/VuFind/DigitalContent/OverdriveConnector.php
```diff
@@ -1037,34 +1242,50 @@ public function getHolds($refresh = true)
...
+                        $result->data = $response->holds;
+                        // Check for holds ready for chechout
+                        foreach ($response->holds as $key => $hold) {
+                            // check for hold suspension
+                            if (isset($hold->holdSupension)) {
```
demiankatz: This if/else could be simplified to:

`$result->data[$key]->holdSupension = $hold->holdSupension ?? false;`

module/VuFind/src/VuFind/DigitalContent/OverdriveConnector.php
```diff
@@ -1450,12 +1673,13 @@ protected function connectToPatronAPI(
     /**
      * Get an HTTP client
      *
-     * @param string $url URL for client to use
+     * @param string $url            URL for client to use
+     * @param bool   $allowRedirects Whether to allow the client to follow redirects
      *
-     * @return \Laminas\Http\Client
+     * @return \Zend\Http\Client
```
demiankatz: This Zend should be changed back to Laminas; probably reintroduced by accident.

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -250,8 +250,7 @@ public function getPreviewLinks()
      */
     public function supportsAjaxStatus()
     {
-        // TODO: add this as an Overdrive configuration to turn it off
-        return true;
+        return $this->config->enableAjaxStatus;
```
demiankatz: Should we have a default value here in case enableAjaxStatus is unset? e.g. `return $this->config->enableAjaxStatus ?? true;`?

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
```diff
@@ -1,59 +1,81 @@
 <div class="odaction">
+  <?php if($action == 'getTitle'): ?>
+    <h2><?=$this->transEsc($actionTitleCode); ?>
+        <span class="odbrand pull-left">
+          <img class="media-left" src="https://developerportaldev.blob.core.windows.net/media/Default/images/newLogos/OverDrive_Logo_28x28_rgb.jpg">
```
demiankatz: Would it be better to include a local copy of this file rather than linking to it externally, or does OverDrive require this approach? The external links may get blocked by content security policies, plus it makes us dependent on external hosting that might change over time.

themes/bootstrap3/templates/admin/menu.phtml
```diff
@@ -5,7 +5,5 @@
   <a href="<?=$this->url('admin/maintenance')?>" class="list-group-item<?=strtolower($this->layout()->templateName) == 'maintenance' ? ' active' : ''?>"><?=$this->transEsc('System Maintenance')?></a>
   <a href="<?=$this->url('admin/tags')?>" class="list-group-item<?=strtolower($this->layout()->templateName) == 'tags' ? ' active' : ''?>"><?=$this->transEsc('Tag Maintenance')?></a>
   <a href="<?=$this->url('admin/feedback')?>" class="list-group-item<?=strtolower($this->layout()->templateName) == 'feedback' ? ' active' : ''?>"><?=$this->transEsc('Feedback Management')?></a>
-  <?php if ($this->overdrive()->showOverdriveAdminLink()):?>
-    <a href="<?=$this->url('admin/overdrive')?>" class="list-group-item<?=strtolower($this->layout()->templateName) == 'overdrive' ? ' active' : ''?>"><?=$this->transEsc('Overdrive API')?></a>
-  <?php endif;?>
+  <a href="<?=$this->url('admin/overdrive')?>" class="list-group-item<?=strtolower($this->layout()->templateName) == 'overdrive' ? ' active' : ''?>"><?=$this->transEsc('Overdrive API')?></a>
```
demiankatz: I don't think the string Overdrive API exists in en.ini -- and shouldn't it be OverDrive API in any case?

themes/bootstrap3/templates/myresearch/odmycontent.phtml
```diff
@@ -55,6 +68,8 @@
             </a>
           </h4>
 
+          <div class="od_edition"><?=$edition?></div>
```
demiankatz: Does $edition need to be escaped here, or do we expect it to contain HTML?

themes/bootstrap3/templates/myresearch/odmycontent.phtml
```diff
@@ -64,23 +79,23 @@
...
-          <div><?=$this->transEsc('od_expires_on', ['%%due_date%%' => $resource['checkout']->expires])?></div>
+          <div> <?=$this->transEsc('od_expires_on', ['%%due_date%%' => $resource['checkout']->expires])?></div>
           <div class="pull-right">
             <?php if ($resource['checkout']->isReturnable): ?>
               <a
-                class="btn btn-primary icon-link returnTitle"
+                class="btn btn-warning icon-link returnTitle"
                 data-lightbox title="<?=$this->transEsc('od_but_return')?>"
-                href="<?=$this->url('overdrive-hold') . "?od_id=$od_id&rec_id=$rec_id&action=returnTitleConfirm"; ?>"
+                href="<?=$this->url('overdrive-hold') . "?od_id=$od_id&rec_id=$rec_id&edition=$edition&action=returnTitleConfirm"; ?>"
```
demiankatz: Does $edition need to be urlencoded here?

demiankatz (changes requested):
Thanks for the ongoing progress, @bpalme -- I took another pass through the files and found a few more things to comment on. I hope this is helpful! Please let me know when you're ready for me to do more review and testing.

module/VuFind/src/VuFind/Controller/OverdriveController.php
bpalme: I like the idea. I've been meaning to but just haven't gotten up the gumption yet. I'll start working on it.

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -361,13 +374,15 @@ public function isCheckedOut()
...
     public function isHeld()
     {
-        $overDriveId = $this->getOverdriveID();
-        $result = $this->connector->getHolds(true);
-        if ($result->status) {
-            $holds = $result->data;
-            foreach ($holds as $hold) {
-                if (strtolower($hold->reserveId) == $overDriveId) {
-                    return $hold;
+        if ($this->isLoggedIn()) {
```
demiankatz, line 377: In #3459, I eliminated the isLoggedIn method that returns a user object, since it's a bit confusing to have a method that sounds like a Boolean but returns a non-Boolean value. While working on that, I noticed that Overdrive had an isLoggedIn method of its own, but that method is not used at all in the dev branch. I see that you _are_ using it here. I'd suggest either renaming the method to something like `getCatalogUser` if you want it to return an object, or changing it to return a Boolean value (since that seems to be the only way you're using it anyway).

bpalme: I changed the return type. But let me know if there is a better way to do it. I couldn't find any other drivers that checks for a user so maybe there is a more fundamental problem here.

bpalme:
It occurred to me that perhaps it was a faux-pas to resolve the conversations. Is it understood that you are the one to do that? At any rate, I left the others unresolved, but I think I'm all caught up on changes so far.

themes/bootstrap3/templates/RecordTab/overdriveholdings.phtml
demiankatz: It looks like this comment may have gotten missed...

demiankatz:
Thanks, @bpalme -- it looks like one comment related to i18n may have been missed. I've put a follow-up comment on that thread. I'll review the rest of the changes soon, as time permits!

Regarding resolving comment threads, I'm happy with whatever workflow works best for you. If you leave a comment and you want me to see/acknowledge it, it's probably best to leave the thread open -- then I can read what you said, double-check the changes, and resolve the thread. But if the comment is something really simple, like "fix this typo," then you can fix it and resolve the comment and nothing more need be said. :-)

In any case, thanks for all the work on this -- it's looking like release 10.0 will be coming in May or June, so hopefully we're on track to get this all wrapped up by then.

themes/bootstrap3/templates/RecordTab/overdriveholdings.phtml
bpalme: Either I misunderstood what you meant or the system just didn't pick up the change because the line number is little different:  https://github.com/bpalme/vufind/blob/VUFIND-1499%2BVUFIND-1375/themes/bootstrap3/templates/RecordTab/overdriveholdings.phtml#L48

demiankatz: Strange -- when I looked at this yesterday, `od_copies_available` was still missing the `%%copies%%` token in en.ini. But I see it's there now! Sometimes GitHub lags behind, and changes don't show up right away. Perhaps that is what happened. In any case, thanks for all the progress, and I'll review this again as soon as I can (probably early next week, but maybe sooner if I'm lucky ;-) ).

module/VuFind/src/VuFind/Controller/OverdriveController.php
```diff
@@ -183,17 +218,23 @@ public function holdAction()
         if (!is_array($patron = $this->catalogLogin())) {
             return $patron;
         }
-        $this->debug('patron: ' . print_r($patron, true));
+        //$this->debug('patron: ' . print_r($patron, true));
```
demiankatz: Can this be fully removed, or might you need it back?

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -272,11 +271,11 @@ public function getOverdriveAccess()
      *
      * Returns whether the current user is logged in
      *
-     * @return object|bool User if logged in, false if not.
+     * @return bool User if logged in, false if not.
```
demiankatz: This comment is no longer accurate with the changed return type -- I'd just drop the comment entirely since the method is now self-explanatory, and reduce the line to simply `@return bool`

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
demiankatz: I think this is fine for now at least!

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -326,29 +324,44 @@ public function getOverdriveAvailability()
...
-                    $result->status = true;
-                    $result->data = $checkout;
+        $result = $this->connector->getResultObject();
+        if ($this->isLoggedIn()) {
+            $overdriveID = $this->getOverdriveID();
+            $result = $this->connector->getCheckouts(true);
+            if ($result->status) {
+                $checkedout = false;
+                $checkouts = $result->data;
+                //In case of a magazine issue, we have to get all the checkouts to see if the
```
demiankatz: Minor style nitpick: we generally include a space after `//` comments.

demiankatz (changes requested):
@bpalme, I had a little bit of time to look at this again, so I resolved all of the open issues that are taken care of, and left a few more very minor suggestions. I'll try to find time to give the remaining files a more thorough review soon!

Also note that I merged the latest dev branch into this one and corrected a couple of very minor issues here to save you some time.

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -549,4 +562,39 @@ public function getRealTimeTitleHold()
...
+    }
+
+    /**
+     * Get Permanent Link to the resource on your institution's OverDrive site
+     *
+     * @return string the permanent link to the resource
+     */
+    public function getPermanentLink()
+    {
+        $libraryURL = $this->config->libraryURL;
```
demiankatz: Where does this libraryURL setting come from? I don't see it in Overdrive.ini. Am I looking in the wrong place, or do we need to add an example? Also, do we need to handle the possibility that the setting is missing? Should we throw an exception in that case, or just skip providing a URL?

demiankatz (changes requested):
Thanks, @bpalme -- I'm still chipping away at reviewing this. It looks like all open comments from prior reviews are now addressed (there were one or two small things that seem to have gotten missed, but I just went ahead and fixed them because they were minor and didn't need to waste more of your time). See below for one new question. I'll continue reviewing as time permits. My goal is to get this merged within the next 2-3 weeks so we can get all of the language file updates in place in time for the 10.0 release... but if that timeline doesn't work for you, we can always push this forward to a future release again.

themes/bootstrap3/templates/admin/menu.phtml
```diff
@@ -5,7 +5,5 @@
   <a href="<?=$this->url('admin/maintenance')?>" class="list-group-item<?=strtolower($this->layout()->templateName) == 'maintenance' ? ' active' : ''?>"><?=$this->transEsc('System Maintenance')?></a>
   <a href="<?=$this->url('admin/tags')?>" class="list-group-item<?=strtolower($this->layout()->templateName) == 'tags' ? ' active' : ''?>"><?=$this->transEsc('Tag Maintenance')?></a>
   <a href="<?=$this->url('admin/feedback')?>" class="list-group-item<?=strtolower($this->layout()->templateName) == 'feedback' ? ' active' : ''?>"><?=$this->transEsc('Feedback Management')?></a>
-  <?php if ($this->overdrive()->showOverdriveAdminLink()):?>
-    <a href="<?=$this->url('admin/overdrive')?>" class="list-group-item<?=strtolower($this->layout()->templateName) == 'overdrive' ? ' active' : ''?>"><?=$this->transEsc('Overdrive API')?></a>
-  <?php endif;?>
+  <a href="<?=$this->url('admin/overdrive')?>" class="list-group-item<?=strtolower($this->layout()->templateName) == 'overdrive' ? ' active' : ''?>"><?=$this->transEsc('od_admin_menu')?></a>
```
demiankatz: Is there a reason why we made the Overdrive menu item unconditional here? Might we still want the ability to disable it when it's not relevant?

themes/bootstrap3/templates/myresearch/odmycontent.phtml
```diff
@@ -55,6 +68,8 @@
             </a>
           </h4>
 
+          <div class="od_edition"><?=$this->escapeHtml($edition)?></div>
```
demiankatz: Do we want to display this div even if $edition is an empty string? Would it make sense to wrap in `<?php if ($edition): ?> ... <?php endif; ?>`?

themes/bootstrap3/templates/myresearch/odmycontent.phtml
```diff
@@ -64,23 +79,24 @@
...
-          <div><?=$this->transEsc('od_expires_on', ['%%due_date%%' => $resource['checkout']->expires])?></div>
+          <div> <?=$this->transEsc('od_expires_on', ['%%due_date%%' => $resource['checkout']->expires])?></div>
           <div class="pull-right">
             <?php if ($resource['checkout']->isReturnable): ?>
               <a
-                class="btn btn-primary icon-link returnTitle"
+                class="btn btn-warning icon-link returnTitle"
                 data-lightbox title="<?=$this->transEsc('od_but_return')?>"
-                href="<?=$this->url('overdrive-hold') . "?od_id=$od_id&rec_id=$rec_id&action=returnTitleConfirm"; ?>"
+                href="<?=$this->url('overdrive-hold') . "?od_id=$od_id&rec_id=$rec_id&edition=" .
```
demiankatz: Are the various IDs here guaranteed numeric, or should we be urlencoding all of them?

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
bpalme: Strange. yes it was missing before and I am sure I've added it at least once. I guess it somehow didn't get committed. I think we do need the URL since it's library specific. I'll re-add it to config.

module/VuFind/src/VuFind/Controller/OverdriveController.php
```diff
@@ -121,16 +146,27 @@ public function mycontentAction()
...
-                        = $this->serviceLocator->get(\VuFind\Record\Loader::class)
-                        ->load(strtolower($hold->reserveId));
-                    $holds[] = $myhold;
+                    try {
+                        $myhold['record']
+                            = $this->serviceLocator->get(\VuFind\Record\Loader::class)
+                            ->load(strtolower($hold->reserveId));
+                        $holds[] = $myhold;
+                    } catch (\VuFind\Exception\RecordMissing $e) {
+                        //hold is missing from Solr
```
demiankatz: Nitpick: I'd add a space after `//` here and below on line 162.

module/VuFind/src/VuFind/Controller/OverdriveController.php
```diff
@@ -210,90 +249,36 @@ public function holdAction()
...
+            'suspendHold' => ['titleCode' => 'od_susp_hold', 'resMeth' => 'getSuspendHoldRes'],
+            'editSuspendedHold' => ['titleCode' => 'od_susp_hold_edit', 'resMeth' => 'getEditSuspendedRes'],
+            'cancelHold' => ['titleCode' => 'od_cancel_hold', 'resMeth' => 'getCancelHoldRes'],
+            'returnTitle' => ['titleCode' => 'od_early_return', 'resMeth' => 'getReturnTitleRes'],
+            'getTitle' => ['titleCode' => 'od_get_title', 'resMeth' => 'getDownloadTitleRes'],
+        ];
+
+        if (isset($actions[$action])) {
+            $actionTitleCode = $actions[$action]['titleCode'];
+            $result = $actions[$action]['resMeth'] ? $this->{$actions[$action]['resMeth']}() : false;
```
demiankatz, line 281: Does it make any difference that in your old code, $result defaulted to `null` but now it gets set to `false` instead for situations where no method is used? I think it's likely the two are functionally equivalent for the purposes of your code, but maybe `null` is a more appropriate value for this situation.

module/VuFind/src/VuFind/Controller/OverdriveController.php
```diff
@@ -308,11 +293,290 @@ public function holdAction()
...
+    {
+        $od_id = $this->params()->fromQuery('od_id');
+        $rec_id = $this->params()->fromQuery('rec_id');
+        // Load the Record Driver. Should be a SolrOverdrive driver.
+        $driver = $this->serviceLocator->get(\VuFind\Record\Loader::class)->load(
+            $rec_id
+        );
+        $formats = $driver->getDigitalFormats();
+        // Looks like this is a magazine...
+        if (current($formats)->id == 'magazine-overdrive') {
```
demiankatz, line 353: Is this a safe enough check? Is there always going to be exactly one digital format? If not, would it be better to create a support method to search the whole array?

module/VuFind/src/VuFind/Controller/OverdriveController.php
```diff
@@ -308,11 +293,290 @@ public function holdAction()
...
+    public function getReturnTitleConfirmResult()
+    {
+        $result = $this->connector->getResultObject();
+        $rec_id = $this->params()->fromQuery('rec_id');
+        // Load the SolrOverdrive driver.
+        $driver = $this->serviceLocator->get(\VuFind\Record\Loader::class)->load(
+            $rec_id
+        );
+        $formats = $driver->getDigitalFormats();
+        $result->data = (current($formats)->id == 'magazine-overdrive') ?: false;
```
demiankatz, line 480: Here's another instance of the magazine check. Maybe it's worth creating a support method like `isRecordDriverMagazine` to do this work since it's repeated, whether or not the logic itself needs to be adjusted.

themes/bootstrap3/templates/RecordTab/overdriveholdings.phtml
```diff
@@ -0,0 +1,182 @@
...
+  $checkedOutResult = $this->driver->isCheckedOut();
+  $checkedOut = $checkedOutResult->data;
+  if (!$checkedOut):
+    $hold = $this->driver->isHeld();
+  endif; ?>
+
+  <div class="internetDetails">
+    <div class="odbrand pull-left media-left">
+      <img
+        src=""<?=$this->imageLink('OverDrive_Logo_28x28_rgb.jpg')?>"">
```
demiankatz: There seem to be extra double-quotes here.

themes/bootstrap3/templates/RecordTab/overdriveholdings.phtml
```diff
@@ -0,0 +1,182 @@
...
+          <?= $avail->data->copiesAvailable . '/' . $avail->data->copiesOwned ?>
+        <?php endif; ?>
+        <?= $this->transEsc('od_copies_available', ['%%copies%%' => $copies]) ?>
+        <?php if ($avail->data->numberOfHolds > 0): ?>
+          <?= '; ' . $this->transEsc('od_waiting', ['holds' => $avail->data->numberOfHolds], null, true) ?>
+        <?php endif; ?>
+      </p>
+
+      <?php if (!$checkedOut || $checkedOutResult->isMagazine): ?>
+        <?php if (isset($avail->data->copiesAvailable) && $avail->data->copiesAvailable > 0): ?>
```
demiankatz: You could simplify some of the if checks in this file by removing issets through null coalescing -- e.g.:

` <?php if (isset($avail->data->copiesAvailable) && $avail->data->copiesAvailable > 0): ?>`

could become:

` <?php if (($avail->data->copiesAvailable ?? 0) > 0): ?>`

No pressure to do this if it's too much work or you prefer not to, but making the statements more concise might be helpful.

themes/bootstrap3/templates/RecordTab/overdriveholdings.phtml
```diff
@@ -0,0 +1,182 @@
...
+          <p>
+            <a class="btn btn-primary checkout" data-lightbox title="<?= $this->transEsc('request_place_text') ?>"
+              href="<?= "$hold_url?od_id=$od_id&rec_id=$rec_id&action=checkoutConfirm" ?>">
```
demiankatz: I think I've asked this question elsewhere, but this is another place where I wonder if we need to urlencode some of these IDs just in case they contain unexpected characters. But maybe it's not necessary if they're absolutely guaranteed to be integers.

demiankatz (changes requested):
Thanks again, @bpalme -- I managed to look through a couple more files; see a few more suggestions below. Your refactoring of the controller is a big improvement over the earlier version, in any case!

bpalme:
> Thanks, @bpalme -- I'm still chipping away at reviewing this. It looks like all open comments from prior reviews are now addressed (there were one or two small things that seem to have gotten missed, but I just went ahead and fixed them because they were minor and didn't need to waste more of your time). See below for one new question. I'll continue reviewing as time permits. My goal is to get this merged within the next 2-3 weeks so we can get all of the language file updates in place in time for the 10.0 release... but if that timeline doesn't work for you, we can always push this forward to a future release again.

I would very much like to get this in the next release. I should have the current batch of revisions done by this Saturday.

demiankatz:
Excellent, thanks, @bpalme. I have just two long files left to review, so I'll see if I can get to those this week as well just in case there are more comments for me to share. Hopefully nothing major in any case!

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
```diff
@@ -1,59 +1,81 @@
...
+                <label id="select-issue-label"><?=$this->translate('Issue') ?></label>
+                <select class="form-control" name="edition" id="select-edition" data-open="false">
+
+                <?php foreach ($result->data->products as $key => $issue):
+                    $checkedout = $issue->checkedout ? ' - ' . $this->transEsc('od_mag_issue_ischeckedout') : ''; ?>
+                    <option value="<?=$issue->id?>"><?=$issue->edition . $checkedout?> </option>
+                <?php endforeach;?>
+                </select>
             </div>
+            <?php  endif;?>
```
demiankatz: Minor style thing -- might be worth a string replace from `<?php  ` (2 spaces) to `<?php ` (1 space) to clean up some unnecessary whitespace in a couple of spots here. Also worth adding a space before the `?>` in this close tag.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
```diff
@@ -1,59 +1,81 @@
 <div class="odaction">
+  <?php if($action == 'getTitle'): ?>
+    <h2><?=$this->transEsc($actionTitleCode); ?>
+        <span class="odbrand pull-left">
+          <img class="media-left" src="<?=$this->imageLink('OverDrive_Logo_28x28_rgb.jpg')?>">
+        </span>
+    </h2>
+    <iframe class="od_download" src="<?=$result->data->downloadRedirect?>"></iframe>
```
demiankatz: Might be worth `$this->escapeHtmlAttr()` around the src value here.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
```diff
@@ -1,59 +1,81 @@
...
+
+                <input name="doAction" value="doCheckout" type="hidden">
+                <input class="btn btn-success" name="doCheckout" value="<?=$this->transEsc('od_but_checkout_s'); ?>" type="submit">
             </form>
+
+        <?php elseif ($result->code == 'OD_CODE_NOMAGISSUES'): ?>
+            <div class="alert alert-warning">
+                <div class="moreInfo">
+                    <?=$this->transEsc('od_docheckout_failure') ?>
+                    <?=$result->msg?>
```
demiankatz: This might need `$this->escapeHtml` if it's not already in HTML.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
```diff
@@ -116,144 +134,140 @@
...
-                <div class="form-group pull-right">
-                    <input name="doAction" value="cancelHold" type="hidden">
-                    <input class="btn btn-primary" name="cancelHold" value="<?=$this->transEscAttr('od_but_cancel_hold'); ?>" type="submit">
-                </div>
-            </form>
+    <?php elseif ($action == 'editHoldEmailConfirm'): ?>
+        <form class="" method="POST" name="editHoldEmail" data-lightbox-onclose="VuFind.refreshPage">
+            <div class="form-group">
+                <label class="control-label"><?=$this->transEsc('Email Address'); ?></label>
+                <input id="email" name="email" value="<?=$result->data->hold->emailAddress?>" size="" class="form-control" type="text">
```
demiankatz: `$this->escapeHtmlAttr` on the email address?

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
```diff
@@ -116,144 +134,140 @@
...
+    <?php elseif ($action == 'editHoldConfirm'): ?>
+        <form class="" method="POST" name="editHold" data-lightbox-onclose="VuFind.refreshPage">
 
-                <input name="doAction" value="returnTitle" type="hidden">
-                <input class="btn btn-primary" name="returnTitle" value="<?=$this->transEscAttr('od_but_return'); ?>" type="submit">
+            <div class="form-group">
+                <label id="select-suspension-label"><?=$this->transEsc('od_susp_hold_confirm'); ?></label>
+                <select class="form-control" name="suspendValue" id="select-suspension" data-open="false">
+                    <option value="0" selected=""><?=$this->transEsc('od_susp_asap'); ?></option>
+                    <option value="7"><?=$this->transEsc('od_susp_after', ['%%days%%' => '7']); ?></option>
```
demiankatz: Would it make sense to refactor this to:

```
<?php $durations = [7, 14, 21, 28, 60, 90, 180]; ?>
<?php foreach ($durations as $duration): ?>
  <option value="<?=$duration?>"><?=$this->transEsc('od_susp_after', ['%%days%%' => $duration]); ?></option>
<?php endforeach; ?>
```

?

That would also open the door to making this a config setting if there were any value in doing that.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
```diff
@@ -116,144 +134,140 @@
...
-                <div class="odformats">
-                    <h3><?=$this->transEsc('od_dl_formats'); ?></h3>
-                    <?php foreach ($formats as $key => $format): ?>
-                        <a data-lightbox-ignore class="btn btn-primary" target="_blank"
-                           href="<?=$this->url('overdrive-hold') . "?od_id=$od_id&rec_id=$rec_id&action=getTitle&getTitleFormat=$key"?>">
-                            <?=$this->translate($format) ?></a>
-                    <?php endforeach;?>
-                </div>
+                    <option value="7"><?=$this->transEsc('od_susp_after', ['%%days%%' => '7']); ?></option>
+                    <option value="14"><?=$this->transEsc('od_susp_after', ['%%days%%' => '14']); ?></option>
```
demiankatz: Ditto here.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
```diff
@@ -116,144 +134,140 @@
...
-            <div class="moreInfo"><?=$result->msg?></div>
-          </div>
-        <?php  endif;?>
-        <div class="pull-right">
-          <button type="button" class="btn btn-primary" data-dismiss="modal" aria-hidden="true"><?=$this->transEsc('close'); ?></button>
-        </div>
+    <div class="returnTitleDialog">
+
+        <?php  if ($result->data): ?>
+          <h4><?=$edition?></h4>
```
demiankatz: Does this need to be escaped?

module/VuFind/src/VuFind/DigitalContent/OverdriveConnector.php
```diff
@@ -326,7 +330,7 @@ public function getAvailabilityBulk($overDriveIds = [])
             if (!$res) {
                 $result->code = 'od_code_connection_failed';
             } else {
-                if ($res->errorCode == 'NotFound' || $res->totalItems == 0) {
+                if ((isset($res->errorCode) and $res->errorCode == 'NotFound') or $res->totalItems == 0) {
```
demiankatz: Could this be:

```
if (($res->errorCode ??  null) == 'NotFound' || $res->totalItems == 0) {
```

demiankatz (changes requested):
Okay, I think I've finished looking at everything -- here are a few more questions and nitpicks, but hopefully nothing that will take up too much of your time!

themes/bootstrap3/templates/admin/menu.phtml
bpalme: Well the permission wasn't working for some reason and I noticed none of the others had something like that so I changed it to be like them. If Overdrive is not configured, it just displays empty values which seems appropriate to me. But let me know if you want me to put it back and make it work again.

themes/bootstrap3/templates/myresearch/odmycontent.phtml
bpalme: As far as I know they are always alphnum but I'll just encode them to be on the safe side.

module/VuFind/src/VuFind/Controller/OverdriveController.php
bpalme: I think it is. Ebooks and audiobooks can have multiple formats but magazines just come in one format

themes/bootstrap3/templates/admin/menu.phtml
bpalme: This is probably why it wasn't working: [VUFIND-1398](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1398)

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
bpalme: I think I caught all of these. Also removed all space after short echo tags based on other templates I looked at.

themes/bootstrap3/templates/admin/menu.phtml
demiankatz: I took a couple of minutes to look into this and pushed up commit 3a3bd264fe51d6b65e069a808b433e53e7837a89 to reinstate the configuration. I think it's worth having, since people may have varying levels of comfort with providing web-based access to data about their settings, etc. Of course, feel free to adjust if you don't like anything I've done here.

module/VuFind/src/VuFind/RecordDriver/SolrOverdrive.php
```diff
@@ -549,4 +562,42 @@ public function getRealTimeTitleHold()
...
+    {
+        return $this->getIsMarc()
+            ? parent::getURLs()
+            : $this->getPermanentLink();
+    }
+
+    /**
+     * Get Permanent Link to the resource on your institution's OverDrive site
+     *
+     * @return string the permanent link to the resource
```
demiankatz: Looks like the return type here is actually `array` rather than `string`.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
```diff
@@ -1,59 +1,81 @@
...
-        <div>
-        <?=$this->transEsc($actionTitleCode); ?>
-        </div>
+    <h2><?=$this->transEsc($actionTitleCode)?>
+        <span class="odbrand pull-left"
+          <img class="media-left" src="<?=$this->imageLink('OverDrive_Logo_28x28_rgb.jpg')?>">
+        </span>
     </h2>
-
+         <div class="pull-right  img-col">
```
demiankatz: There's an extra space between the two classes here.

themes/bootstrap3/templates/RecordDriver/SolrOverdrive/hold.phtml
```diff
@@ -111,149 +129,138 @@
...
+    <?php elseif ($action == 'editHoldConfirm'): ?>
+        <form class="" method="POST" name="editHold" data-lightbox-onclose="VuFind.refreshPage">
 
-                <input name="doAction" value="returnTitle" type="hidden">
-                <input class="btn btn-primary" name="returnTitle" value="<?=$this->transEscAttr('od_but_return'); ?>" type="submit">
+            <div class="form-group">
+                <label id="select-suspension-label"><?=$this->transEsc('od_susp_hold_confirm')?></label>
+                <select class="form-control" name="suspendValue" id="select-suspension" data-open="false">
+                    <option value="0" selected=""><?=$this->transEsc('od_susp_asap')?></option>
+                    <?php $durations = $this->config()->get('Overdrive')->Overdrive->suspendDurations; ?>
```
demiankatz: Is it worth having a default fallback in case the config is unset, e.g.

```
<?php $durations = $this->config()->get('Overdrive')->Overdrive->suspendDurations ?? [7, 14, 21, 28, 60, 90, 180]; ?>
```

That way outdated configurations won't break the functionality. If you do it here, you'll also need to do it in the other place that displays the menu.

demiankatz (changes requested):
Thanks again, @bpalme -- just a couple more very trivial suggestions. I also pushed up a few more minor adjustments while I was reviewing the latest changes. If you clean up these last couple of things, and as long as you don't object to what I've done with the admin code, we may be ready to merge this very soon.

demiankatz (approved):
Thanks, @bpalme, for all the hard work on this (and for bearing with all of my feedback). As far as I can tell, this seems to be ready to merge now. Are there any last-minute things you want to do before I hit the merge button?

bpalme:
No. I say go for it. I'll try to do another round of testing after you merge and before we do the final release.

bpalme:
I have some related edits for the ovedrive wiki page. When should I add those and what is the process?

demiankatz:
@bpalme, feel free to make those edits at your convenience -- just note which things are 10.0-specific, if that matters. And if you have trouble accessing the wiki, let me know.

PR #3027, opened 2023-08-07
Add support for persistent login

rajaro:
Adds support for persistent login with a login token cookie and a database entry. Allows the user to save their login and automatically log in on their next visit. The user can also view and delete their saved logins. Configurable support for different login methods.

I'm opening this as a draft, because this is still missing at least tests. Any feedback would be appreciated.

TODO
- [x] Add and test PostgreSQL support
- [x] Note backward compatibility breaks in changelog when merging (e.g. constructor signature change on \VuFind\Auth\Manager).
- [x] Resolve [VUFIND-619](https://openlibraryfoundation.atlassian.net/browse/VUFIND-619) when merging.


config/vufind/config.ini
```diff
@@ -596,6 +596,15 @@ username_hint = username_only_letters_numbers_and_basic_punctuation
 ; Default is true.
 ;delete_ratings_with_user = false
 
+; "Remember me" functionality for listed login methods. If the user chooses to save the login,
+; a login token cookie and a database entry will be created. Users browser information will be
```
demiankatz: Should add an apostrophe to "Users".

module/VuFind/sql/mysql.sql
```diff
@@ -409,3 +409,24 @@ CREATE TABLE `access_token` (
...
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE utf8mb4_unicode_ci;
 /*!40101 SET character_set_client = @saved_cs_client */;
+
+--
+-- Table structure for table `login_token`
+--
+
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8mb4 */;
+CREATE TABLE `login_token` (
```
demiankatz, line 419: Once things are finalized, we should add equivalent code to the PostgreSQL configuration. I've added a TODO checkbox to the top of the PR to remind us of this.

module/VuFind/src/VuFind/Auth/LoginToken.php
```diff
@@ -0,0 +1,280 @@
...
+        $token = bin2hex(random_bytes(32));
+        $series = $series ? $series : bin2hex(random_bytes(32));
+        $browser = '';
+        $platform = '';
+        try {
+            $userInfo = get_browser(null, true) ?? [];
+            $browser = $userInfo['browser'];
+            $platform = $userInfo['platform'];
+        } catch (\Exception $e) {
+            // Problem with browscap.ini, continue without
```
demiankatz: Do we want to tolerate this situation? Would it be better to have a config setting to "opt in" to ignoring browscap.ini problems, so people don't accidentally turn on the feature in a less secure configuration than they expect?

module/VuFind/src/VuFind/Auth/LoginToken.php
```diff
@@ -0,0 +1,280 @@
...
+     *
+     * @return void
+     */
+    public function sendLoginTokenWarningEmail(\VuFind\Db\Row\User $user)
+    {
+        if (!empty($user->email)) {
+            $this->mailer->send(
+                $user->email,
+                $this->config->Site->email,
+                $this->translate('login_warning_email_subject'),
```
demiankatz: It might be a good idea to make the email subject line configurable via config.ini (but still translatable) so it's easier to change without modifying language files. It might also be worthwhile to make the email body a template so that can also be more easily customized. It might also be a good idea to include the site title in here somewhere, since otherwise with the default settings, the source of the email may not be immediately obvious to the end user if it isn't customized.

module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -267,6 +277,42 @@ public function supportsConnectingLibraryCard($authMethod = null)
...
+     * Is persistent login supported?
+     *
+     * @param User $user optional; check user auth method in database
+     *
+     * @return bool
+     */
+    public function supportsPersistentLogin($user = null)
+    {
+        if (!empty($this->config->Authentication->persistent_login)) {
+            if ($user) {
```
demiankatz: Under what circumstances is the active auth method in the manager different from the auth method in the user object? I'm sure there's a reason for this, but it's not immediately obvious to me why `$this->getAuthMethod()` alone is not enough. Maybe it's worth adding a comment to explain and/or refactoring this to a separate method like `getLastAuthMethodForUser()`.

module/VuFind/src/VuFind/Controller/MyResearchController.php
```diff
@@ -1815,7 +1815,7 @@ protected function sendVerificationEmail($user, $change = false)
                         $this->sendChangeNotificationEmail($user, $to);
                     }
                 } catch (MailException $e) {
-                    $this->flashMessenger()->addMessage($e->getDisplayMessage(), 'error');
+                    $this->flashMessenger()->addMessage($e->getMessage(), 'error');
```
demiankatz: What is the reason for this change? I think the getDisplayMessage was introduced to avoid information disclosure risks.

themes/bootstrap3/templates/myresearch/profile.phtml
```diff
@@ -84,6 +84,42 @@
     <?php endif; ?>
   </div>
 
+  <?php $user = $this->auth()->isLoggedIn(); ?>
```
demiankatz, line 87: Do you think it might be worth promoting this to a separate menu entry / controller action?

demiankatz (changes requested):
Thanks, @rajaro, this looks like a great start. I'm on vacation this week so didn't take the time to give it a very deep look, but I left a few comments based on a quick perusal. I can review more deeply, and do some hands-on testing, after I return next week. I also linked this to the VUFIND-619 JIRA ticket, which was opened a long time ago for discussion of this feature. It will be nice to finally offer an implementation. Thanks again for taking the time to work on this!

module/VuFind/src/VuFind/Controller/MyResearchController.php
rajaro: This was unintended. Reverted the change.

themes/bootstrap3/templates/myresearch/profile.phtml
rajaro: I don't have strong feelings either way. I think it is small enough to justify its place in the profile menu and it makes sense to me to have saved logins in the same menu as "Change password" and other account actions.
Then again if it was a separate menu entry maybe it would be easier for the user to find their saved logins. And I guess saved logins are a bit different than basic account information. I'm fine with either way.

module/VuFind/src/VuFind/Auth/Manager.php
rajaro: The problem happens when logging in with a token, basically this:
Log in and save the login -> delete session -> go to /MyResearch/Profile
The logging in works, but saved logins are not displayed because auth method is not updated. Looking at it now, I think we should probably make sure to update auth method when logging in with a token. And if that is the case, where should the info about the auth method come from? From database, or should the login token cookie carry auth method info?

module/VuFind/src/VuFind/Auth/LoginToken.php
rajaro: Will do.

module/VuFind/src/VuFind/Auth/LoginToken.php
rajaro: Not quite sure how to handle this. I'm not sure if there is actually a realistic scenario where this feature would be on without browser detection. We could do the opt in, but should we simply require a working browscap.ini? If we expect someone to use this without browser detection, should there be changes in the UI as well when no platform or browser is detected?

config/vufind/config.ini
rajaro: Fixed. Also removed the last sentence "Login methods are written in lowercase." because it is not needed.

themes/bootstrap3/templates/myresearch/profile.phtml
demiankatz: I don't have strong feelings -- it just seems like we keep adding more and more things to the profile page and it's potentially getting quite long. But maybe all we need to do is introduce tabs or some equivalent navigation mechanism there if it gets too long.

module/VuFind/src/VuFind/Auth/LoginToken.php
demiankatz: I think requiring a working browscap.ini is reasonable. It makes sense to implement this as securely as possible, and that seems like an important layer to include. If somebody wants it without that feature, we can address the need when we have an understood use case for it.

module/VuFind/src/VuFind/Auth/Manager.php
demiankatz: I think it probably makes sense to include that information in the database -- if it's carried by the cookie, isn't it easier for the user to manipulate, or do I misunderstand what you mean?

EreMaijala:
Might be worth noting that this is based on [Improved Persistent Login Cookie Best Practice.md](https://gist.github.com/oleg-andreyev/9dcef18ca3687e12a071648c1abff782), right?

config/vufind/config.ini
```diff
@@ -605,6 +605,9 @@ username_hint = username_only_letters_numbers_and_basic_punctuation
 ; Persistent login token lifetime in days
 ;persistent_login_lifetime = 60
 
+; Subject for the email sent after a suspicious login has been detected.
```
demiankatz, line 613: Probably worth adding a comment that the default will be the translation key login_warning_email_subject, and that any custom value used here instead will also be run through the translator.

module/VuFind/src/VuFind/Auth/LoginToken.php
```diff
@@ -229,11 +239,15 @@ public function deleteActiveToken()
     public function sendLoginTokenWarningEmail(\VuFind\Db\Row\User $user)
     {
         if (!empty($user->email)) {
+            $message = $this->viewRenderer->render('Email/login-warning.phtml', ['title' => $this->config->Site->title]);
```
demiankatz: I don't see the Email/login-warning.phtml file. Is it possible it didn't get committed with the rest of the change?

demiankatz (changes requested):
@rajaro, apologies for taking so long to get back to this -- it has been a very busy month!

I've taken a look at the latest changes and have a couple more comments; I think at least one of the issues below needs to be fixed before I can do hands-on testing of this.

Also, I notice that CI is failing due to a very minor style issue:

```
FILE: .../work/vufind/vufind/module/VuFind/src/VuFind/Auth/LoginToken.php
----------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
----------------------------------------------------------------------
 243 | WARNING | Line exceeds 120 characters; contains 121 characters
----------------------------------------------------------------------
```

Thanks again for the progress on this, and please let me know if I can do more to help!

demiankatz:
I'm also setting a 10.0 milestone for this one. I'm open to merging it sooner if it gets done in time for 9.1, but since that version is soon going to have a freeze on new language strings, I think 10.0 is probably a more realistic target.

module/VuFind/src/VuFind/Auth/Manager.php
rajaro: @demiankatz  Now the auth method is updated when the user logs in with a cookie. Auth method is checked from the database. I also changed `makeAuth()` function to make the comparison between user auth method and legal auth methods case insensitive. This is because auth methods are stored lowercase in the database.

module/VuFind/src/VuFind/Auth/LoginToken.php
rajaro: That was indeed the case

demiankatz:
Thanks for the progress on this, @rajaro -- I haven't forgotten about you, but I've been very busy finishing up release 9.1. I look forward to completing further review on this once the rest of the 9.1 work is settled.

languages/en.ini
```diff
@@ -1019,6 +1025,7 @@ pick_up_location = "Pickup Location"
 Place a Hold = "Place a Hold"
 Place of birth = "Place of birth"
 Place of death = "Place of death"
+Platform = "Platform"
```
demiankatz: It might make sense to give this translation a more specific key to clarify the context -- e.g. saved_login_platform instead of just "Platform". That will help the translators understand how it is used, and also allow more focused customization if we use the word "Platform" in another context elsewhere.

module/VuFind/sql/pgsql.sql
```diff
@@ -380,6 +380,24 @@ revoked boolean NOT NULL DEFAULT '0',
 PRIMARY KEY (id, type)
 );
 
+--
+-- Table structure for table `login_token`
+--
+
+DROP TABLE IF EXISTS "login_token";
+
+CREATE TABLE login_token (
```
demiankatz, line 389: We will also need a migration script under https://github.com/vufind-org/vufind/tree/dev/module/VuFind/sql/migrations/pgsql (you can create a new 10.0 subdirectory) to create this new table. If you have not been able to test PostgreSQL support and want to be sure it works before copying this to a migration, let me know and I can do some testing for you. Otherwise, I'll wait to do my testing until you have made the migration, since then I can test that it works at the same time as everything else.

languages/en.ini
```diff
@@ -159,6 +159,7 @@ browse_lcc = "Call Number (LC)"
 browse_publishDate = "Year of Publication"
 browse_title = "Title"
 browse_topic = "Topic"
+Browser = "Browser"
```
demiankatz: Related to my comment about Platform -- does "Browser" need its own separate translation, or should we create a single heading translation to capture "Platform / Browser"? (I don't have a strong opinion one way or the other -- just raising the question. It might be more efficient to have just one translation, but more flexible to keep things separate.) If we do keep this, it's another string that might benefit from a more descriptive key.

themes/bootstrap3/templates/myresearch/profile.phtml
```diff
@@ -84,6 +84,42 @@
...
+    <?php $tokens = $user->getLoginTokens($user->id); ?>
+    <?php if (count($tokens) === 0): ?>
+      <div class="my-profile-col"><?=$this->transEsc('No saved logins')?></div>
+    <?php else: ?>
+      <table class="login-tokens-table table table-striped">
+        <thead>
+          <tr>
+            <th scope="col" class="login_token_browser"> <?=$this->transEsc('Platform')?> / <?=$this->transEsc('Browser')?></th>
+            <th scope="col" class="login_token_last_login"><?=$this->transEsc('Last Login')?>:</th>
+            <th scope="col" class="login_token_delete"><?=$this->transEsc('Delete')?></th>
```
demiankatz: Having a "Delete" heading on a column containing only "Forget Login" buttons seems a little redundant. Would a more general heading like "Actions" make more sense?

Might there also be a benefit to a "Forget All" button to quickly clear everything out?

For accessibility purposes, should we put labels on the buttons to help differentiate them from one another (e.g. by including more details in addition to the "Forget Login" text)?

module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -267,6 +278,40 @@ public function supportsConnectingLibraryCard($authMethod = null)
...
+     * Is persistent login supported?
+     *
+     * @param User $user optional; check user auth method in database
+     *
+     * @return bool
+     */
+    public function supportsPersistentLogin($user = null)
+    {
+        if (!empty($this->config->Authentication->persistent_login)) {
+            if ($this->getAuth() instanceof ChoiceAuth) {
```
demiankatz: This could be simplified with a ternary:

```
$method = $this->getAuth() instanceof ChoiceAuth
    ? $this->getAuth()->getSelectedAuthOption() : $this->getAuthMethod();
```

module/VuFind/tests/unit-tests/src/VuFindTest/Auth/LoginTokenTest.php
```diff
@@ -0,0 +1,249 @@
...
+            $config,
+            $userTable,
+            $tokenTable,
+            $cookieManager,
+            $sessionManager,
+            $mailer,
+            $viewRenderer
+        );
+    }
+}
```
demiankatz: This file is missing a newline at the end -- one of a few minor style issues that you can probably fix automatically with `vendor/bin/phing phpcbf`.

module/VuFind/tests/unit-tests/src/VuFindTest/Auth/LoginTokenTest.php
```diff
@@ -0,0 +1,249 @@
...
+        $userTable->expects($this->once())->method('getById')
+            ->with($this->equalTo(0))
+            ->will($this->returnValue($this->getMockUser()));
+        $tokenTable = $this->getMockLoginTokenTable();
+        $tokenTable->expects($this->once())->method('matchToken')
+            ->will($this->returnValue($mockToken));
+        $loginToken = $this->getLoginToken($cookieManager, $tokenTable, $userTable);
+
+        // Expect exception due to browscap.ini requirements
+        $this->expectException(\VuFind\Exception\Auth::class);
```
demiankatz: Should the name of this test method indicate that we're expecting a failure/exception? The name suggests that this is testing a normal successful run, but that doesn't appear to be the case.

module/VuFind/src/VuFind/Db/Table/LoginToken.php
```diff
@@ -0,0 +1,193 @@
...
+     *
+     * @return mixed
+     * @throws LoginTokenException
+     */
+    public function matchToken(array $token)
+    {
+        $row = $this->select(
+            [
+                'user_id' => $token['user_id'],
+                'series' => $token['series'],
```
demiankatz: The table does not seem to enforce a unique key on user_id/series, but we're only retrieving one row here. Is there any possibility that multiple matching rows might exist, and we could somehow miss the best one? Should we iterate on results to be safe and/or impose more restrictions on the table definition?

(I admit that I'm asking these questions without closely reviewing the algorithm or where these data elements come from... so apologies if they're silly or irrelevant, but it may not hurt to try to address them if any weird edge cases are even a slim possibility).

module/VuFind/src/VuFind/Db/Table/LoginToken.php
```diff
@@ -0,0 +1,193 @@
...
+     * Check if a login token matches one in database.
+     *
+     * @param array $token array containing user id, token and series
+     *
+     * @return mixed
+     * @throws LoginTokenException
+     */
+    public function matchToken(array $token)
+    {
+        $row = $this->select(
```
demiankatz: Could you use getBySeries here to reduce redundancy?

demiankatz: (Obviously, that might not make sense if you stop using current based on my other comments here... but if this code stays the same, it might make sense to use the support method for brevity)).

demiankatz (changes requested):
@rajaro, thanks again for your patience as I've worked through the 9.1 release process. Now that that is done, I should be able to devote considerably more time to getting this project finished. See below for my latest round of feedback, based on a closer review of the code (but no hands-on testing yet).

languages/en.ini
rajaro: Having them combined to a single translation makes sense. So something like "saved_login_platform_and_browser"?

themes/bootstrap3/templates/myresearch/profile.phtml
rajaro: Sure, "Actions" does sound better. I'll also add a "Forget All" button.

For accessibility, I'll add the platform/browser and login time info to the label

languages/en.ini
demiankatz: Yes, that works!

module/VuFind/src/VuFind/Db/Table/LoginToken.php
rajaro: True, I think there is a possibility that there could be multiple matching rows for a given series/user_id. I'll fix that

demiankatz:
@sturkel89, when time permits, could you do a bit of hands-on testing of this new feature? If you uncomment the `persistent_login` setting in config.ini, it should enable a "remember me" option when you log in. If you go to the profile section of your account, you can manage persistent logins from there.

A couple of notes of things to test in particular:

1.) It would be a good idea to try logging in from multiple browsers so you have several sessions to experiment with. Then you can test, for example, that deleting a session on the profile screen in one browser causes a different browser to prompt for a new login on your next access.

2.) We should test this functionality with both MySQL and PostgreSQL since it adds new database tables. To test with PostgreSQL instead of the default MySQL, just add `-Ddbtype=pgsql` to the command line when you run both the startup and shutdown Phing commands.

Please let me know if you need more detail on anything!

module/VuFind/sql/mysql.sql
```diff
@@ -409,3 +409,25 @@ CREATE TABLE `access_token` (
...
+  `user_id` int NOT NULL,
+  `token` varchar(255) NOT NULL,
+  `series` varchar(255) NOT NULL,
+  `last_login` datetime NOT NULL,
+  `browser` varchar(255) NULL,
+  `platform` varchar(255) NULL,
+  `expires` int NOT NULL,
+  `last_session_id` varchar(255) NULL,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `user_id_series` (`user_id`, `series`)
```
demiankatz: I think there's a missing comma at the end of this line -- it's preventing the database from being created correctly in my test instance. I don't have write access to this branch, so I can't fix it myself. Do you mind taking a look when you have a moment?

demiankatz (changes requested):
See below -- there's a problem that's preventing us from moving forward with testing this branch.

module/VuFind/sql/mysql.sql
rajaro: comma added

demiankatz: Thanks! I'll take another look at this after @sturkel89 has had a chance to do some testing.

module/VuFind/sql/mysql.sql
```diff
@@ -409,3 +409,25 @@ CREATE TABLE `access_token` (
...
+  `token` varchar(255) NOT NULL,
+  `series` varchar(255) NOT NULL,
+  `last_login` datetime NOT NULL,
+  `browser` varchar(255) NULL,
+  `platform` varchar(255) NULL,
+  `expires` int NOT NULL,
+  `last_session_id` varchar(255) NULL,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `user_id_series` (`user_id`, `series`),
+  KEY `user_id` (`user_id`),
```
demiankatz: This still isn't working for me. I think the comma after the KEY line needs to be removed -- adding the missing comma in the line above wasn't the whole solution.

demiankatz (changes requested):
Looks like one more fix is needed to the SQL before we can test this.

module/VuFind/sql/mysql.sql
rajaro: Oops. Now it should work.

sturkel89 (changes requested):
I was able to checkout the branch and attempt to test it, but I was unable to get persistent login to work. I will explain the process I followed:

I uncommented this line in config.ini:  `persistent_login = "database,multiils"`

I opened the branch in my Virtual Machine and clicked Login. I could see the regular login lightbox with the additional "Remember my login for 0 days" checkbox. 

I clicked Create New Account and was able to fill out the form. The act of filling out the form logged me into my account, but since there was no checkbox for staying logged in on the Create New Account screen I logged out.

Then, when I clicked Login again, I tried to login with the checkbox checked and got an error. The error on the screen was:
image

I tried to log in **without** the checkbox checked, and was successful.

I turned on development mode in order to troubleshoot and tried again to login with the checkbox checked. Instead of getting a full error report, I got the lightbox with this message:
image

@demiankatz noted that "It's possible the errors are getting hidden by the login system due to the way it handles exceptions." It would be helpful to allow us to see these errors in order to help diagnose problems like the one I am having.

I had the same problem getting persistent login to work whether or not I uncommented the persistent_login_lifetime line in config.ini. 

I would be happy to test again once this is fixed and it is possible to check the box and log in to the system. @rajaro, please let me know if there is a step I am missing in my attempt to enable this feature. 

**Another comment:** 
The default time period for persistent login should be a positive number of days, e.g. 14 or 60, not zero, or else there should be an .ini file comment saying that the admin needs to uncomment the line in order to set a real default.

rajaro:
@sturkel89 Thanks for the feedback. I changed the default time period to 14.
I believe the error is likely caused by a problem with browscap.ini. Make sure to have a browscap.ini file and enable it in php.ini like this:
`[browscap]`
`browscap = /path/to/browscap.ini`

demiankatz:
Thanks, @rajaro! It's probably worth investigating why turning on development mode doesn't make this error message visible to the end user. If we can find a way to do that, it will likely reduce end user confusion if people miss/forget about the comment about browscap.ini (which I somehow managed to do while working with @sturkel89, even though I should have known better!). I'm a bit short on time this week, but if you need help with the specifics, I can dig into it a little later. Feel free to just add a TODO checkbox at the top of the PR if you want me to look into this.

Also, have you looked at the [browscap-php](https://github.com/browscap/browscap-php) library? I wonder if using this might allow us to do things a little more seamlessly, and simplify the setup work for users.

rajaro:
@demiankatz browsecap-php library looks good and I think it will be a smoother way to handle things. I will look into it. I'll also check if I can make the error messages better

sturkel89:
I have installed browscap.ini, restarted apache, and given php.ini the information about the browscap.ini file's location. 

I am still getting an internal server error when I try to create a new account when logged in to the test branch.  

I will look at this later with @demiankatz , but for now I remain unable to test this branch.

themes/bootstrap3/templates/Auth/AbstractBase/login.phtml
```diff
@@ -13,6 +13,9 @@
       <?php if ($account->supportsRecovery()): ?>
         <a class="btn btn-link" href="<?=$this->url('myresearch-recover') ?>?auth_method=<?=$account->getAuthMethod()?>"><?=$this->transEsc('Forgot Password')?></a>
       <?php endif; ?>
+      <?php if ($account->supportsPersistentLogin()) : ?>
+        <div><input type="checkbox" name="remember_me"> <?=$this->transEsc('remember_me', ['%%days%%' => $account->getPersistentLoginLifetime()])?></div>
```
demiankatz: This control should be surrounded by a `<label>` tag so that clicking on the text toggles the checkbox.

module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -704,11 +755,43 @@ public function login($request)
         // Update user object
         $this->updateUser($user);
 
+        if ($request->getPost()->get('remember_me') && $this->supportsPersistentLogin()) {
+            try {
+                $this->loginToken->createToken($user, '', $this->sessionManager->getId());
+            } catch (\Exception $e) {
+                throw new AuthException('authentication_error_technical', 0, $e);
```
demiankatz, line 769: It would be helpful to add `$this->logError((string)$e);` here so there is easier visibility of the underlying error when logging is configured.

demiankatz (changes requested):
@rajaro, I was able to get this working in my test environment -- a couple of small suggestions based on quick testing.

@sturkel89, when you are ready, I should be able to help you get this set up the rest of the way so that you can do more comprehensive tests.

sturkel89 (dismissed):
I tested this branch with both MySQL and PostgreSQL. Here are my results.

**MySQL**
I was able to create a login and then log in with multiple browsers and see all of those saved logins listed in my profile. I could close a window and then open it again and still be logged in. I could "Forget login" and then have that window be logged out of VuFind. Everything seemed to work properly with MySQL. 

The only thing that seemed inaccurate was that the platform was always detected as Win10 -- but my computer is running Win11.

**PostgreSQL**

I was able to create a login and then log in with a couple of browsers. I found that I would try to log in and the first time would get this error: "We cannot log you in at this time. Please try again later." -- and then I could re-enter the password and be logged in. However, after a couple of times, I would only get an error message and not be logged in. HOWEVER, when I checked my Saved Logins in the first logged-in window, I could see that **each of the failed login attempts was still listed as if it were active**:

image

I also tried closing a window that had successfully logged in and then reopened it, but it was no longer logged in.

**_I also note that the logins are listed out of order; can you impose a chronological ordering on the list?_**

We checked the error logs and found the following repeated many times. I hope it will help with your debugging.

> 2023-12-18T15:02:10-05:00 ERR (3): VuFind\Auth\Manager: VuFind\Exception\Auth: Failed to save token in /usr/local/vufind/module/VuFind/src/VuFind/Auth/LoginToken.php:191
> Stack trace:
> #0 /usr/local/vufind/module/VuFind/src/VuFind/Auth/Manager.php(760): VuFind\Auth\LoginToken->createToken()
> #1 /usr/local/vufind/module/VuFind/src/VuFind/Controller/MyResearchController.php(152): VuFind\Auth\Manager->login()
> #2 /usr/local/vufind/vendor/laminas/laminas-mvc/src/Controller/AbstractActionController.php(72): VuFind\Controller\MyResearchController->homeAction()
> #3 /usr/local/vufind/module/VuFind/src/VuFind/Controller/Feature/CatchIlsExceptionsTrait.php(75): Laminas\Mvc\Controller\AbstractActionController->onDispatch()
> #4 /usr/local/vufind/vendor/laminas/laminas-eventmanager/src/EventManager.php(320): VuFind\Controller\MyResearchController->onDispatch()
> #5 /usr/local/vufind/vendor/laminas/laminas-eventmanager/src/EventManager.php(178): Laminas\EventManager\EventManager->triggerListeners()
> #6 /usr/local/vufind/vendor/laminas/laminas-mvc/src/Controller/AbstractController.php(105): Laminas\EventManager\EventManager->triggerEventUntil()
> #7 /usr/local/vufind/vendor/laminas/laminas-mvc/src/DispatchListener.php(117): Laminas\Mvc\Controller\AbstractController->dispatch()
> #8 /usr/local/vufind/vendor/laminas/laminas-eventmanager/src/EventManager.php(320): Laminas\Mvc\DispatchListener->onDispatch()
> #9 /usr/local/vufind/vendor/laminas/laminas-eventmanager/src/EventManager.php(178): Laminas\EventManager\EventManager->triggerListeners()
> #10 /usr/local/vufind/vendor/laminas/laminas-mvc/src/Application.php(319): Laminas\EventManager\EventManager->triggerEventUntil()
> #11 /usr/local/vufind/public/index.php(23): Laminas\Mvc\Application->run()
> #12 {main}
>

rajaro:
Saved logins are now sorted by last_login.

@demiankatz I got my postgre setup working and managed to test and repeat the bug but the bug is quite confusing. The error that causes failed login is "While processing primary key data, a known key id was not found in the data array".
Yet a new row is created and saved in the database with correct values which is why the logins are displayed on profile page. 
By simply ignoring the exception thrown when saving the token, everything seemed to work fine. 

Any insight into this would be helpful. I'm going to be on vacation for two weeks, after which I will return to this.

demiankatz:
> Saved logins are now sorted by last_login.
> 
> @demiankatz I got my postgre setup working and managed to test and repeat the bug but the bug is quite confusing. The error that causes failed login is "While processing primary key data, a known key id was not found in the data array". Yet a new row is created and saved in the database with correct values which is why the logins are displayed on profile page. By simply ignoring the exception thrown when saving the token, everything seemed to work fine.
> 
> Any insight into this would be helpful. I'm going to be on vacation for two weeks, after which I will return to this.

Thanks for looking at this. I suspect the problem may be that the database abstration layer expects a sequence mapping to know how to generate primary keys in PostgreSQL. You may need to add a line to this configuration array:

https://github.com/vufind-org/vufind/blob/dev/module/VuFind/config/module.config.php#L606

I hope you have a pleasant holiday break, and we can discuss further in the new year! :-)

demiankatz (changes requested):
Thanks for the update, @rajaro; I'm hoping that @sturkel89 can test this again sometime this week. In the meantime, I notice the build is failing because of the need to run `vendor/bin/phing php-cs-fixer` to correct a missing comma. I tried to do it for you to save time, but I don't have permission to push to your branch.

sturkel89:
I have tested this branch using postgresql, and it seems to me that it is working properly! 

I can log in on multiple browsers without experiencing errors, and I see the list of active logins under my Profile, sorted in a logical order. 

If I click "forget login" for one of the other sessions, I can see that I get logged out of that session.

Each of the sessions has its own search history, but saved items lists and saved searches persist across the different logins. That makes sense, as they are associated with the account.

image

My only criticisms at this point are the following:
- My platform is identified as Windows 10, when in fact it is Windows 11
- The login name seems to be case sensitive. I agree that passwords should be case sensitive, but it would be more flexible and easier for the user if the username were allowed to be case insensitive.

module/VuFind/sql/pgsql.sql
demiankatz: I think we've gotten PostgreSQL all sorted out, so it's probably time to add the migration. Do you want to do this, or would you like my help? I'm willing to do it as needed.

demiankatz (changes requested):
Thanks, @rajaro and @sturkel89. I think we're nearing the end of the process and will be able to merge this soon.

I'm not sure why Susan's Windows version is getting misidentified, but I suspect that is more likely to be related to something in browscap.ini than something in the code here; fixing it is probably beyond the scope of the work here (unless @rajaro feels differently).

The case-sensitive username issue in PostgreSQL is also unrelated to the changes here and seems to be a pre-existing problem that nobody ever noticed before! I've opened PR #3316 to correct it; that work can proceed independently of this work.

demiankatz:
@rajaro, also note that I went ahead and resolved the latest merge conflicts for you here, since it was easy to do through the web interface (just a minor language file issue).

I notice that this is still marked as a draft; feel free to take it out of draft mode when you feel ready!

rajaro:
@demiankatz I also believe the problem with Windows versions is related to browscap.ini rather than the code here. Feel free to add the migration. You should now have permissions to edit this branch.

demiankatz:
Thanks, @rajaro! I've created and tested a migration script; it's now pushed to this branch.

Two questions before I dive into what I hope will be the final round of testing here:

1.) I notice that the MySQL table definition includes a separate index on the user_id field, but the PostgreSQL table definition does not. Is this difference intentional?

2.) We talked last year about the [browscap-php](https://github.com/browscap/browscap-php) library. Is that something you still intend to investigate? Should I hold off on further review/merging until you have looked at that, or would it be better to try to get this merged as-is and then treat that as a follow-up improvement project?

rajaro:
@demiankatz 
1.) It was not intentional, and now I added a create index statement to postgre definition. 
2.) I think it would be better to try to get this merged as-is and treat that as a follow-up

demiankatz:
Thanks, @rajaro! I have updated the migration to match the revisions to the table definition. I'll continue testing/review (and hopefully get this merged, barring any new problems discovered) as soon as time permits -- probably early next week at the latest.

module/VuFind/src/VuFind/Auth/LoginToken.php
```diff
@@ -0,0 +1,314 @@
...
+/**
+ * Class LoginToken
+ *
+ * @category VuFind
+ * @package  VuFind\Auth
+ * @author   Jaro Ravila <jaro.ravila@helsinki.fi>
+ * @license  https://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Page
+ */
+class LoginToken implements \VuFind\I18n\Translator\TranslatorAwareInterface
```
demiankatz: Maybe this class should be renamed to LoginTokenManager, since that's what it is; leaving off the "Manager" potentially makes it sound like a class that represents a single login token.

module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -138,6 +145,7 @@ class Manager implements
      *
      * @param Config         $config         VuFind configuration
      * @param UserTable      $userTable      User table gateway
+     * @param LoginToken     $loginToken     Login Token
```
demiankatz: Is there a strong reason to add the new argument here in the list? Might it be better to add it to the end of the list for greater backward compatibility?

module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -271,6 +282,37 @@ public function supportsConnectingLibraryCard($authMethod = null)
             && $this->getAuth($authMethod)->supportsConnectingLibraryCard();
     }
 
+    /**
+     * Is persistent login supported?
+     *
+     * @param User $user optional; check user auth method in database
```
demiankatz: This parameter is never used. Is it really needed?

module/VuFind/src/VuFind/Auth/Manager.php
```diff
@@ -512,6 +555,14 @@ public function isLoggedIn()
                 $results = $this->userTable->createRow();
                 $results->exchangeArray($this->session->userDetails);
                 $this->currentUser = $results;
+            } elseif ($this->cookieManager->get('loginToken')) {
```
demiankatz, line 556: Do we ever want to attempt this if persistent login is disabled or inapplicable? Would it make sense to do a check first? For example, what if tokens exist out in the wild, but the administrator decides to turn off the feature. Shouldn't that stop the tokens from being used?

module/VuFind/src/VuFind/Db/Table/LoginToken.php
```diff
@@ -0,0 +1,192 @@
...
+     * Get token by series
+     *
+     * @param string $series Series identifier
+     * @param int    $userId User identifier
+     *
+     * @return LoginToken
+     */
+    public function getBySeries(string $series, int $userId)
+    {
+        return $this->select(['user_id' => $userId, 'series' => $series])->current();
```
demiankatz, line 177: Does it matter that this result set is not sorted? Is there a reason to sort by last_login DESC when using user_id, but not when using series?

module/VuFind/src/VuFind/Db/Table/LoginToken.php
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * Save a token
+     *
+     * @param string $userId    User identifier
+     * @param string $token     Login token
+     * @param string $series    Series the token belongs to
+     * @param string $browser   User browser
+     * @param string $platform  User platform
+     * @param int    $expires   Token lifetime in days
```
demiankatz: This comment says expires is token lifetime in days, but lower down, it's compared against the current Unix timestamp to expire a row. Is something inconsistent somewhere, or am I just not reading carefully enough?

themes/bootstrap3/templates/myresearch/profile.phtml
```diff
@@ -84,6 +84,46 @@
...
+  <?php $user = $this->auth()->isLoggedIn(); ?>
+  <?php if ($this->auth()->getManager()->supportsPersistentLogin($user)): ?>
+    <div class="myresearch-profile-header">
+      <h3><?=$this->transEsc('Saved logins')?></h3>
+    </div>
+    <?php $tokens = $user->getLoginTokens($user->id); ?>
+    <?php if (count($tokens) === 0): ?>
+      <div class="my-profile-col"><?=$this->transEsc('No saved logins')?></div>
+    <?php else: ?>
+      <table class="login-tokens-table table table-striped">
```
demiankatz: Should this table be given responsive styling similar to what was recently added in #3308?

demiankatz (changes requested):
I found a little time to do a code read today, so see below for a few suggestions and questions. I apologize if any of my observations are off-base; please feel free to ignore anything that is unhelpful or irrelevant.

Once you've had a chance to look through this and take any necessary actions, I'll do one more round of testing and then hopefully get things merged.

module/VuFind/src/VuFind/Db/Table/LoginToken.php
rajaro: Comments were wrong. It is supposed to be a timestamp. Comments fixed.

themes/bootstrap3/templates/myresearch/profile.phtml
rajaro: Table is now responsive

module/VuFind/src/VuFind/Db/Table/LoginToken.php
rajaro: When getting tokens for display, getByUserId is used, so that is why it is sorted. It should not matter that getBySeries is not sorted

module/VuFind/src/VuFind/Auth/Manager.php
rajaro: Surely tokens should stop working when the feature is turned off. I added a check for this, but a couple rows below to make sure it works properly with ChoiceAuth. So if possible, we check the users auth method and then check whether persistent login is supported.

module/VuFind/src/VuFind/Auth/Manager.php
rajaro: Not needed, removed.

module/VuFind/src/VuFind/Auth/Manager.php
rajaro: Moved to the end of the list

module/VuFind/src/VuFind/Auth/LoginToken.php
rajaro: Sure, LoginTokenManager is more descriptive. Renamed

demiankatz (approved):
Thanks for the patience and hard work on this, @rajaro! I think it's ready to go now, so I'm going to merge it.

As discussed earlier, it's probably worth a follow-up action to revise the code to use the browscap library to simplify that process. It would also be a good idea to add some Mink tests to cover the new behavior. Please let me know if I can help with either or both!

PR #3029, opened 2023-08-14
Add check:scss task and use it in QA to avoid writing out the compile…

EreMaijala:
…d files.

Also moves the `outputStyle` setting to the task-specific settings and removes the `style`option that's not used by the node-sass plugin.


demiankatz (approved):
Looks good! Thanks, @EreMaijala.

PR #3030, opened 2023-08-14
Fix email-based catalog login from record holdings tab.

EreMaijala:
Without route parameters the route doesn't include the record id or current tab. Without the catalogLogin parameter record controller wouldn't process the login request. The issue arises when a user has already logged in to VuFind but doesn't have a library card attached. If they try to add it using the link on the record holdings tab, it would result in a broken link to be sent via email.

This changes the signature of EmailAuthenticator::sendAuthenticationLink, but I believe it's worth it.


demiankatz (approved):
I agree, this seems worth the change -- I'll add a changelog note.

PR #3031, opened 2023-08-14
Fix result limit and sort templates.

EreMaijala:
This has been spun off from #2929. Makes the limit template similar to sort template.


EreMaijala:

demiankatz (changes requested):
I agree that changing the limit shouldn't reset the sort -- that seems like a possible copy and paste bug. But looking into this raised another issue: should changing both sort and limit reset the page number? Right now, if you use these controls on page 2, you remain on page 2, even though the meaning of "page 2" has been totally redefined by the changed parameter.

Maybe that's an issue better solved separately -- I think the fixes here make things better even without fixing that problem -- but I thought I'd leave this open until we could discuss. Let me know if you think it's worth tackling that here or starting a separate PR.

demiankatz:
I should also note that this PR is going to conflict with #2814 when we get that far, but I agree that solving this for the current dev branch in the meantime still makes sense. We should just make sure that any changed behavior/fixes get correctly merged into that other PR after this work is completed.

EreMaijala:
@demiankatz I decided to add filtering of the page parameter in both templates, now here to avoid having to revisit this later. :)

demiankatz (changes requested):
As discussed in Slack, this looks good to me, but I'll hold off on merging until you've had time to add tests.

EreMaijala:
Tests now added.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SearchLimitTest.php
```diff
@@ -67,7 +69,7 @@ protected function setUpLimitedSearch(
         $config = ['default_limit' => $default, 'limit_options' => $options];
         $this->changeConfigs(['searches' => ['General' => $config]]);
         $session = $this->getMinkSession();
-        $session->visit($this->getVuFindUrl() . '/Search/Results?limit=' . $limitParam);
+        $session->visit($this->getVuFindUrl() . '/Search/Results?lookfor=%22test publication%22&limit=' . $limitParam);
```
demiankatz: Should we perhaps add the appropriate building filter here to limit to a single MARC file, just to ensure that future records are less likely to break the test?

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SearchSortTest.php
```diff
@@ -0,0 +1,185 @@
...
+                    ],
+                    'Sorting' => [
+                        'title' => 'Title',
+                        'title desc' => 'Title Reversed',
+                    ],
+                ],
+            ]
+        );
+        $session = $this->getMinkSession();
+        $session->visit($this->getVuFindUrl() . "/Search/Results?lookfor=%22test publication%22&sort=$sortParam");
```
demiankatz: Same here...

demiankatz (changes requested):
Thanks, @EreMaijala -- one small suggestion to hopefully make this more future-proof. I was hoping to make the change myself and test it today, but I never quite got there before running out of time!

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SearchLimitTest.php
EreMaijala: Makes sense. I replaced the search phrase with a building filter.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/SearchSortTest.php
EreMaijala: Replaced here as well.

demiankatz (approved):
Thanks, @EreMaijala!

PR #3032, opened 2023-08-14
Introduce results-list template between results and list templates.

EreMaijala:
This paves way for extended use (i.e. the possibility of rendering search results without the surrounding controls).

Extracted from #2929.

### DIFF
--- a/themes/bootstrap3/templates/search/results-list.phtml
+++ b/themes/bootstrap3/templates/search/results-list.phtml
@@ -1,0 +1,9 @@
+<?php
+  // Create a css class for results from the search class:
+  $searchClass = $searchClassCss = $this->params->getSearchClassId();
+  if (!ctype_upper($searchClassCss)) {
+    $searchClassCss = preg_replace('/([a-zA-Z])(?=[A-Z])/', '$1-', $searchClassCss);
+  }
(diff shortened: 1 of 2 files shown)


EreMaijala:

themes/bootstrap3/templates/search/results-list.phtml
```diff
@@ -0,0 +1,9 @@
+<?php
+  // Create a css class for results from the search class:
+  $searchClass = $searchClassCss = $this->params->getSearchClassId();
+  if (!ctype_upper($searchClassCss)) {
```
demiankatz, line 5: It took me a minute to figure out why this was here. While we're touching the code anyway, maybe it's worth adding a comment like:

`// Convert camelCase to hyphenated-text, unless it's all-uppercase, like EDS`

demiankatz (changes requested):
One tiny suggestion prior to merging...

themes/bootstrap3/templates/search/results-list.phtml
EreMaijala: Good idea, comment added.

demiankatz (approved):
Thanks, @EreMaijala!

PR #3033, opened 2023-08-14
Move record version key handling to the new Versions helper.

EreMaijala:
This makes the functionality reusable and overridable.

Extracted from #2929.


EreMaijala:

module/VuFind/src/VuFind/Controller/Feature/RecordVersionsSearchTrait.php
```diff
@@ -94,8 +90,8 @@ public function versionsAction()
         // won't in RSS mode):
         if (isset($view->results)) {
             $view->results->getUrlQuery()
-                ->setDefaultParameter('id', $id)
-                ->setDefaultParameter('keys', $keys)
+                ->setDefaultParameter('id', $this->params()->fromQuery('id'))
+                ->setDefaultParameter('keys', $this->params()->fromQuery('keys'))
```
demiankatz: Is there a reason not to use `$driverAndKeys['keys']` here to avoid re-processing the user input?

It's also tempting to use `$record->getUniqueId()` in the line above, but I suspect that won't work since `$record` might be null (though I didn't trace the code super-thoroughly, so maybe the null case can/should avoid this logic). Should we include the original ID in the $driverAndKeys array, or is there another better way to handle this?

module/VuFind/src/VuFind/Record/VersionsHelper.php
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @return string
+     */
+    public function getSearchStringFromWorkKeys(array $keys): string
+    {
+        $mapFunc = function ($val) {
+            return '"' . addcslashes($val, '"') . '"';
+        };
+
+        return implode(' OR ', array_map($mapFunc, (array)$keys));
```
demiankatz: Do you need to cast `$keys` to an array here, since the parameter is already typed above?

demiankatz (changes requested):
Thanks, @EreMaijala. A couple of small comments. Also, is it worth adding a unit test for the new class? If this is already fully covered by integration tests, perhaps that's not a high priority, though generally I like to add tests for new code when practical.

module/VuFind/src/VuFind/Controller/Feature/RecordVersionsSearchTrait.php
EreMaijala: I changed the helper method to include the id in the result and use it.

module/VuFind/src/VuFind/Record/VersionsHelper.php
EreMaijala: No, fixed.

demiankatz (approved):
Looks good to me. Thanks, @EreMaijala!

(I know I said I was out of time, but then I couldn't stop...)

PR #3034, opened 2023-08-14
Add support for configurable paginator on top of search results.

EreMaijala:
Extracted from #2929.


EreMaijala:

config/vufind/searches.ini
```diff
@@ -153,6 +153,13 @@ retain_filters_by_default = true
 ; commented out, fallback ID lookups will be disabled.
 ;fallback_id_field = previous_id_str_mv
 
+; This setting can be used to configure pagination control on top of results.
+; Possible options are:
+; empty string or false  No top pagination (default)
+; simple                 Simple next/prev button pagination
+; full                   Full pagination alike to the one at the bottom of results
+;top_paginator = simple
```
demiankatz, line 161: Do we also need to add this setting to all the other search-related .ini files? Has it been tested to work in the context of other backends?

Is it worth considering a bottom_paginator setting, so both paginators can be configured independently? Maybe that adds too much complexity (since we would probably want to validate that the user doesn't turn them both off and render the result set impossible to navigate), but it seems conceivable that somebody might want to have ONLY a top paginator, and there's currently no way to configure that.

themes/bootstrap3/templates/search/controls/showing.phtml
```diff
@@ -10,21 +10,25 @@
...
-  $showingResults = $this->translate(
-      isset($this->skipTotalCount) ? 'showing_results_for_html' : 'showing_results_of_for_html',
-      $transParams
-  );
-endif; ?>
 
+  // Use simple format if search-heading is set, search type is not basic or we have a simple top paginator enabled:
+  $topPaginator = $this->results ? $this->results->getOptions()->getTopPaginatorStyle() : '';
+  if ($this->slot('search-heading')->isset() || $this->params->getSearchType() !== 'basic' || 'simple' === $topPaginator) {
+    $showingResults = !empty($this->skipTotalCount)
```
demiankatz, line 12: Isn't `$showingResults` _already_ set to this value at this point in the code due to lines 9-12 above? Since this if..else guarantees that the variable will be set to something, can we just remove those earlier lines?

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/BasicSearchTest.php
```diff
@@ -58,4 +58,97 @@ public function testOutOfBoundsPage()
...
+        $session->visit($this->getVuFindUrl() . '/Search/Results');
+
+        // No prev page on first page:
+        $this->unFindCss($page, '.search-header .pagination-simple .page-prev');
+
+        $secondPage = $this->findCss($page, '.search-header .pagination-simple .page-next');
+        $secondPage->click();
+        $this->waitForPageLoad($page);
+        // Prev page now present:
+        $this->findCss($page, '.search-header .pagination-simple .page-prev');
```
demiankatz: Is it worth clicking on the prev button and asserting that it returns you to the previous page, so that both buttons are tested?

demiankatz (changes requested):
Thanks, @EreMaijala -- a few comments/questions.

config/vufind/searches.ini
EreMaijala: I've now tested with Blender, EDS and Primo, and can't see any reason why it wouldn't work with others as well, so I went ahead and added it, and tried to follow the existing convention in each files regarding comments.

If you really want to, you can at least hide the bottom navigation with css. But I wouldn't want to encourage that, since I can't see that being a good idea in normal cases.

themes/bootstrap3/templates/search/controls/showing.phtml
EreMaijala: Oops, yes, though the condition wasn't quite the same! Now cleaned up.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/BasicSearchTest.php
EreMaijala: Good idea, now added.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/BasicSearchTest.php
```diff
@@ -58,4 +58,126 @@ public function testOutOfBoundsPage()
...
+     * @return void
+     */
+    public function testFullTopPagination(): void
+    {
+        $session = $this->getMinkSession();
+        $session->visit($this->getVuFindUrl() . '/Search/Results');
+        $page = $session->getPage();
+
+        // No paginator unless configured:
+        $this->unFindCss($page, '.search-header .pagination');
```
demiankatz: Why are you checking for `.search-header .pagination` here but using `.pagination-top` when you actually interact with the page? It seems like it might be worth making the selectors more consistent (or at least adding additional find/unfind checks) to make it more clear that the test is proving what it says that it's proving.

demiankatz (changes requested):
Thanks for the progress, and I agree that adding more configuration to hide the bottom paginator is probably not necessary at this point (and could be added in future if there's a compelling need). Just one small question that arose when I looked more closely at the Mink tests.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/BasicSearchTest.php
EreMaijala: Ah, that check was added before I added the pagination-top class. Now simplified.

demiankatz (approved):
Thanks, @EreMaijala -- looks good to me now, and all Mink tests are still passing in my environment.

PR #3035, opened 2023-08-16
Single quotes rule for php-cs-fixer

xmorave2:
There is rule for PHP CS Fixer for changing double quotes to single ones for simple string.  Do we wan't this? And what about strings_containing_single_quote_chars configuration? I don't think it is more readable with backslashes for escaping, so i did'nt incorporate this configuration, but I am always open to discussion.

https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/string_notation/single_quote.rst


demiankatz:
Thanks, @xmorave2 -- guess I should have looked at the php-cs-fixer documentation when I wished this existed, so I could have learned that it does in fact exist! I don't think I'll have time to fully review all of these changes today, but what I've looked at so far looks very promising to me. I agree that leaving the default `strings_containing_single_quote_chars` behavior makes sense -- I'm fine with allowing double quotes around single quotes to avoid escaping for better readability.

demiankatz (approved):
Thanks again, @xmorave2 -- this all looks good to me!

PR #3036, opened 2023-08-17
Add more files to clean up in reset_setup task.

EreMaijala:
Removing any .bak files ensures that the tests don't restore obsolete backups. .key files include the OAuth2 server test keys.


demiankatz (approved):
Thanks, @EreMaijala!

PR #3040, opened 2023-08-21
fix: missed icon link in advanced search

crhallberg:
Found a link with an icon that escaped fixing before.

### DIFF
--- a/themes/bootstrap3/templates/search/advanced/layout.phtml
+++ b/themes/bootstrap3/templates/search/advanced/layout.phtml
@@ -140,9 +140,9 @@
                   <option value="NOT"<?php if (isset($setSearchGroups[$group]) && 'NOT' == $setSearchGroups[$group]):?> selected<?php endif; ?>><?=$this->transEsc('search_NOT')?></option>
                 </select>
               </div>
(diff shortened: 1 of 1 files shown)


demiankatz (approved):
Thanks, @crhallberg!

demiankatz:
Before I merge this, just to be on the safe side: @sturkel89, could you run the full test suite on this, and also double-check that it looks right in all three themes. (I'd be very surprised if it looks wrong anywhere, but I want to be absolutely sure it doesn't break any tests through the reorganized HTML). The way to see it is to go to the advanced search screen and add a search group. The issue affects the "remove search group" link, which has ugly underlining prior to the fix added here.

sturkel89:
I ran the full test suite on this branch. No errors or failed tests!

Then I loaded the branch and tried a complex advanced search. "Remove Search Group" looks great in all three themes!

I switched back to dev and could see the extended underlining in the Remove Search Group link in all three themes. Subtle difference -- but we've made an improvement! :)

demiankatz (approved):
Thanks, @sturkel89 and @crhallberg!

PR #3046, opened 2023-08-23
Refactor hierarchical facets

EreMaijala:
Refactor hierarchical facets to use hierarchical lists with details/summary tags based on [an article describing the mechanism](https://iamkate.com/code/tree-views)

This moves rendering normal phtml templates and removes a lot of the code that subsequently is no longer required.

PR #3041, opened 2023-08-22
Replace jsTree with <details> and CSS

crhallberg:
Based on [this article](https://iamkate.com/code/tree-views) found by @EreMaijala. Thank you, Ere!

- [x] Hierarchical facets
- [ ] Record hierarchy views
- [ ] Lightbox hierarchy views
- [ ] Abstract some JS (maybe)

PR #3047, opened 2023-08-23
Change hierarchical facets from jsTree to a plain HTML tree.

EreMaijala:
This is the minimal version of #3046. This could be more appropriate for the dev branch.

The js changes are based on @crhallberg's work in #3041. I just extended it to mimic the output of single-facet.phtml and changed the classes to align with #3046. If I made any new mistakes, they're totally mine. :)


EreMaijala:
Unfortunately this doesn't seem to work with Safari, so I'll need to do a bit more work.

EreMaijala:
Ok, now done with the updated mechanism from #3046.

themes/bootstrap3/js/facets.js
```diff
@@ -1,87 +1,143 @@
...
+      }
+      linkEl.className = classes + " " + linkEl.className;
+      facetEl = linkEl;
+    }
+
+    // Create toggle button
+    const toggleButton = el("button", "facet-tree__toggle-open");
+    toggleButton.setAttribute('aria-expanded', facet.hasAppliedChildren ? 'true' : 'false');
+    toggleButton.setAttribute('data-toggle-aria-expanded', '');
+    toggleButton.setAttribute('aria-label', facet.displayText);
```
crhallberg, line 85: Do you know of any better or clearer ways to build elements in JS than this? A few that I'm considering are `<template>` elements like we use in the advanced search or [using DOMParser](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString) to write HTML more directly. This might be a good place to introduce a library but I worry about creating multiple places to keep HTML in sync.

themes/bootstrap3/js/facets.js
```diff
@@ -1,87 +1,143 @@
...
+      toggleButton.setAttribute('aria-controls', childUlId);
+      toggleButton.innerHTML = VuFind.icon("facet-opened", "facet-tree__opened") + VuFind.icon("facet-closed", "facet-tree__closed");
+
+      const childrenEl = buildFacetNodes(facetName, facet.children, currentPath, allowExclude, excludeTitle, showCounts, counter, locale);
+      childrenEl.id = childUlId;
+
+      liEl.append(toggleButton, itemContainerEl, childrenEl);
+    } else {
+      toggleButton.innerHTML = VuFind.icon("facet-noncollapsible", "facet-tree__noncollapsible");
+      toggleButton.setAttribute('disabled', '');
```
crhallberg, line 105: Any concerns about mixing single and double quotes? I'm not going to make a fuss about it, but I'm happy to standardize.

themes/bootstrap3/js/facets.js
```diff
@@ -1,87 +1,143 @@
...
-
-    var url = currentPath + facet.href;
-    var excludeUrl = currentPath + facet.exclude;
-    var item = document.createElement('span');
-    item.className = 'text';
-    if (facet.isApplied) {
-      item.className += ' applied';
+  // Helper function to create elements
+  // #todo: abstract to VuFind.el?
+  function el(tagName, className = null) {
```
crhallberg, line 6: I was considering abstracting this to `VuFind.el` and modeling it [closer to this el utility](https://thescottyjam.github.io/snap.js/#!/utils/el). What do you think?

themes/bootstrap3/js/facets.js
EreMaijala: I tried to keep the changes here minimal, because #3046 will completely revamp this and we won't need this code anymore. That's why I wouldn't want to introduce a new library or templates here. Also need to keep in mind that this could be called a thousand times so any performance considerations are important as well.

themes/bootstrap3/js/facets.js
EreMaijala: Ah, yes. I changed most to single ones since they were prevalent in the file, but we could discuss this more generally. I find single quotes easier to read, and because I write a lot of PHP code as well, they come naturally. But either way is fine for me. Anyway, I wouldn't do any more than

themes/bootstrap3/js/facets.js
EreMaijala: That's a nice one! But I'd do that separately from here and apply it in all code we have (maybe apart from this piece that's just a stopgap).

EreMaijala:
One more thing I'm a bit undecided with: should the icons be 'expand' and 'collapse' instead of 'closed' and 'opened'? So named after action or state? It's now state-based to align with other facet icons, but for buttons the action-based naming might make more sense.

themes/bootstrap3/js/facets.js
demiankatz: That last sentence seems to just trail off... ;-)

demiankatz:
> One more thing I'm a bit undecided with: should the icons be 'expand' and 'collapse' instead of 'closed' and 'opened'? So named after action or state? It's now state-based to align with other facet icons, but for buttons the action-based naming might make more sense.

I don't have a really strong opinion and could be persuaded either way, but I think I do maybe slightly favor the expand/collapse approach rather than the closed/opened approach, since it seems a little less potentially ambiguous.

themes/bootstrap3/js/facets.js
EreMaijala: Oops. :) ... I wouldn't do any more than absolutely required here to get things working, so we want to standardize on single or double quotes, I'd do that in another task. And preferably with a test that any future changes follow the choice.

EreMaijala:
> > One more thing I'm a bit undecided with: should the icons be 'expand' and 'collapse' instead of 'closed' and 'opened'? So named after action or state? It's now state-based to align with other facet icons, but for buttons the action-based naming might make more sense.
> 
> I don't have a really strong opinion and could be persuaded either way, but I think I do maybe slightly favor the expand/collapse approach rather than the closed/opened approach, since it seems a little less potentially ambiguous.

Even a slight preference is enough to tip the scale. :) Now changed.

themes/bootstrap3/js/facets.js
demiankatz: I guess the real question is how to split up the work between releases 9 and 10. We don't want to do major refactoring before 10, but maybe we could introduce the helper and start to use it in new code, and then open another PR to do bigger refactoring targeted at release 10.

There's also the question of how to group things together -- do we want to create a util or helper module to contain this type of thing? Or maybe an html module? There might be some other existing global functions that could be relocated. (I'm not sure about that, but a review of places like common.js might reveal some candidates).

themes/bootstrap3/js/common.js
```diff
@@ -163,7 +174,7 @@ var VuFind = (function VuFind() {
           var sliceEnd = sliceStart;
           var value = _attrs[attr];
           var regex = new RegExp(` ${attr}=(['"])([^\\1]+?)\\1`);
-          var existing = html.match(regex);
+          var existing = mod.match(regex);
```
demiankatz, line 175: You changed `html` to `mod` here, but `html` is still used to define sliceStart in line 173 above. Is that a potential problem? Maybe this would all be more clear if we defined `html` AFTER this function, so the scoping is more confined (unless there's a reason I'm missing why we would want to do it this way).

themes/bootstrap3/js/facets.js
EreMaijala: Bearing in mind that this new code would be removed in the refactoring in release 10 by #3046, it would leave us with an unused utility if #3046 goes through. That's why I'm trying to keep things contained here. If it was introduced now in other code, I'd be happy to take advantage of it here, but otherwise I'd keep it separate.

demiankatz (changes requested):
One small question...

themes/bootstrap3/js/facets.js
demiankatz: Agreed, it wouldn't make sense to create an orphaned utility just for this. So unless we want to create a refactor targeted for 10 and then backport something from that to here, it's probably easiest to simply wait for the moment (and maybe remove the TODO comment, if it's also just going to get deleted before any action is taken).

EreMaijala: I already took care of the TODO comment. :)

themes/bootstrap3/js/common.js
EreMaijala: Oops, another one. Now fixed, and html moved lower. Would probably be better to use `let` as well instead of `var`, but that's another change for another time.

demiankatz (changes requested):
@EreMaijala, have you tested this in multiple themes? It looks okay in bootstrap3 but the icons wrap strangely in both bootprint3 and sandal for me:

image

EreMaijala:
Apparently I missed that. I'll check.

EreMaijala:
@demiankatz It was a problem with overflowing the space by 0.1px. I changed the button to fill the space it gets so it should be better and more resistent to trouble now.

demiankatz:
Thanks, @EreMaijala! I talked to @sturkel89 about this PR, so she will do some more testing across themes and configuration combinations in the near future to see if any other anomalies can be identified before we merge this. I'll await her report when time permits. :-)

sturkel89:
I am still working through all of the combinations of features and themes, but I wanted to give an interim report.

There seems to be an issue in @EreMaijala's branch with tree display **when a facet has a lengthy label**. 

I used Chrome's DevTools (F12) to replace the facet labels with multi-word long labels, and in some cases the text wrapped to multiple lines while in other cases it pushed the counts out of view and off the screen. 

Examples, in Bootstrap3:

[Hierarchy is turned on, OR facet is off, Exclude facet is off]
image

[Hierarchy is turned on, Exclude facet is turned on]
image

[Hierarchy is turned on, OR facet is turned on]
image

Compare these with the dev branch. In every circumstance, the long label is truncated with ellipses, and the numbers stay visible. (I think I prefer the word wrapped version in the test branch, but I don't think we can allow for NO wrap and the numbers disappearing from view.)

image
image
image

I will provide a more detailed report next week, but wanted to show this one issue so if you have time you can start looking at it.

demiankatz (changes requested):
Thanks, @sturkel89! Definitely looks like some work is needed in association with orFacets. Can you take a look, @EreMaijala?

EreMaijala:
@sturkel89 @demiankatz Oops! I put some emphasis on testing this as it had been an issue before, but one of the tweaks after that broke everything and I failed to test with the latest changes. Should be fixed now.

By the way, I've also used DOM replacement in the browser to test long strings, but doing it over and over again for different settings is a lot of work, so enabling translation and adding a couple of long ones makes it easier:

local/config/vufind/facets.ini:
```
[Results]
hierarchical_facet_str_mv = hierarchy

[SpecialFacets]
hierarchical[] = hierarchical_facet_str_mv

[Advanced_Settings]
translated_facets[] = hierarchical_facet_str_mv
```

local/languages/en.ini:
```
0/level1a/ = "Level 1a has a long description"
1/level1a/level2a/ = "Level2loooooooooooooong"
2/level1a/level2a/level3a/ = "Level3 again is so long"
```

demiankatz:
Thanks, @EreMaijala! Great suggestion regarding the translated facets, too -- I think I've actually done that in the past, but it didn't occur to me when I was giving @sturkel89 advice on testing, and it's a real time-saver!

Villanova is closed for a holiday today, but we'll get back to this later in the week. Thanks for the fix in the meantime!

sturkel89:
I've finished with this round of reviews of the test branch. I see that the issue I identified earlier has been fixed: with the OR facet turned on, long facet text is now truncated with ellipses and with good spacing between the text and the count indicator in all three themes. The number column is no longer pushed beyond the edge of the Hierarchy box. Thanks, @EreMaijala, for the tip about enabling translation and adding those long labels to en.ini!

See below for my evaluation. For the most part, everything looks good and functions properly. However, I have four issues I wanted to point out. I'll put each in its own section.

***Elongated count dots in Sandal***
With the OR facet turned on, there is one weird issue that differs from dev: When the browser window is narrowed, but not narrow enough to move the filter area to the top of the screen (as in mobile), something weird happens to the count circles:

- In the test branch, they're elongated, are bumped down below their referent, and span the entire following row in all areas. 
- In dev, the shape is slightly elongaged in the hierarchy facets, but they stay in place. They only bump down and span the next row in the non-hierarchy facets.

**This is only an issue in Sandal. The dots stay circular and in their proper locations in all other themes.**

image

***Spacing issue with long facet names in conditions other than having the OR facet enabled***
As noted above, long facet names now display properly with the OR facet turned on, in both test and dev.

However, under the other conditions, in the test branch there is **not enough padding between the facet text and the count numbers**. (This is when neither OR nor EXCLUDE are enabled, or when only EXCLUDE is enabled.) This is particularly a problem with the very long facet name that doesn't contain spaces, but at narrow window widths it's a problem with all lengthy facet names. They run right up against the count numbers.

The dev branch uses ellipses which abut the count numbers, but they're light so they don't seem to be crowding the numbers. To be consistent with the update you made to the display of the OR condition, you'd need to also truncate and use ellipses under these other two conditions.

image





***Location of "Reset Filters" button***
I notice an issue that when several long facets are selected, the “Reset Filters” text/button seems to move to the bottom of its area rather than being aligned with the top of the first row of facets. I compared this under test and dev, across themes, and it was the same under all circumstances.

I checked an older installation (the Falvey instance) and see that the Reset Filters text/button remains top aligned. So, this may be a change in VuFind 9? Or it might have to do with having the hierarchy tree turned on? Not sure. In any case, the button looks strange floating in that spot below the facets it refers to; I think it fits much better aligned across the top of that area next to the top row of selected filters.

image


***Aesthetic comments on vertical spacing and arrows***
In general, I prefer the tighter vertical spacing between elements in the tree that we have in dev. This spacing is the same in all three branches; this example is Bootstrap3. 

I also prefer the larger arrows in the dev branch, with the open arrows being hollow and the closed arrows being shaded. (This is all just personal aesthetic preference!)

image

demiankatz:
Thanks for all this analysis, @sturkel89! My thoughts, for what they're worth:

1.) Elongated count dots in Sandal: I think it's probably a good thing that the changes here make the problem occur consistently between hierarchical and regular facets -- that probably means that it can be fixed by a single change to improve everything. Since that's a pre-existing problem, it might make sense to address that as a follow-up after merging this.

2.) Spacing issue with long facet names in conditions other than having the OR facet enabled: I agree that this looks like a problem we might want to address here to make everything consistent. Hopefully it's not too difficult!

3.) Location of "Reset Filters" button: this seems worth fixing but isn't directly related to the changes here, so it should probably be addressed in a separate PR.

4.) Aesthetic comments on vertical spacing and arrows: While I also like the tighter spacing personally, I suspect it may be more accessible to include more spacing. I wouldn't mind larger arrow icons if they're readily available, but I suspect that @EreMaijala is using icons that are consistent with other parts of the VuFind interface, and the larger ones may be peculiar to jsTree. So for the sake of consistency, it may be preferable to keep the smaller arrows (or maybe reconfigure things to make all such arrows larger... though again, I would probably address that separately to avoid making too many unrelated changes here).

EreMaijala:
Thanks @sturkel89 for testing this thoroughly!

1.) This seems intentional (see https://github.com/vufind-org/vufind/blob/dev/themes/sandal/less/sandal.less#L263) because in sandal there's very little space in the narrow version. The result is certainly not perfect, but anyway.

2.) Should be fixed now. I had a mistake in the JS that creates the nodes, and a css rule that targeted only the icon label but should have been `.facet-value`.

3.) Yeah, unrelated but not nice.

4.) It's all for consistency, but I reduced the margins for child items a bit in an attempt to find a middle ground. What do you think? Luckily it's now much easier to customize than before. The icons used to be defined in CSS but now use the icon helper, which makes it easier to customize them, but I understand the result is perhaps not as nice. it's more consistent with the other icons, though! I can follow up with a PR that adds a new CSS icon set and restores the previous icons.

EreMaijala:
I also reduced the left padding of the facet text with bootprint3 since there was a fairly large gap between the collapse/expand icon and the text.

sturkel89:
Thanks for your responses, @EreMaijala and @demiankatz! I loaded updated versions of the test branch and the dev branch and checked on your updates. Here’s what I see.

**1. Elongated count dots in Sandal** 
At this point, Ere’s branch now behaves the same as dev. Only the non-hierarchy count dots bump down and span across the whole column. It sounds like this design in Sandal is actually deliberate (although it's still a little weird to my eyes...) I guess there’s nothing left to fix!

**2. Spacing issue with long facet names in conditions other than having the OR facet enabled**
I checked in all conditions and across themes. Thanks - looks like this has been completely fixed!

**3. Location of "Reset Filters" button**
I agree that this can be dealt with in a separate PR.

**4. Aesthetic comments on vertical spacing and arrows**
The new spacing looks better than the old! However, I’m not really sure that it accomplishes anything to have that extra padding between level 1 and level 2, especially since all spacing after that seems to be consistent (and looks good). 

image

I would advocate for consistent vertical spacing regardless of level within the tree; the tree is what gives the information that this is a subset of that. Thoughts?

Thanks in advance for potentially bringing back the previous icons! And thanks for the quick turnaround after my comments! :)

demiankatz (changes requested):
Thanks once again, @EreMaijala and @sturkel89! I think we're nearly there. I agree with @sturkel89 that the inconsistent vertical spacing seems undesirable. Is this easily fixed? If so, I think we should fix it. If it's very difficult, I could be persuaded to live with it as-is and merge what we have here. :-)

EreMaijala:
Thanks again for testing @sturkel89! The problem with the padding is that the inverse background for a selected item looks a bit odd if the bottom padding of the top level is reduced or the top padding of the first child is reduced. There's nothing between them but their own padding. Personally I'd condense the whole facet list to have a bit less padding, but if it's only done for hierarchical facets, they don't fit in with the non-hierarchical ones. To illustrate the selection:
image

I'm open to further suggestions, but I think that reducing padding for all facet items is a bit out of scope here.

EreMaijala:
Worth a note: this issue isn't as visible in dev, because the highlighting isn't the same as with non-hierarchical facets. And this is an inconsistency I'd like to avoid.

EreMaijala:
Actually, scratch that. I can counter the padding with some negative margin and it just might go unnoticed. Fix coming up.

EreMaijala:
@sturkel89 @demiankatz Should be better now. :) I don't really like to use negative margins, but here it seems warranted.

sturkel89:
@EreMaijala @demiankatz First off, I don't think this issue is THAT important! In my testing, it's only under one condition that a user would notice the slight differences in the heights of the facet rows: when there are multiple shaded rows with an unshaded row between them. I don't think this is worth devoting a lot of time and work to, unless it is a pretty quick fix. In any case, here are my observations:

- Unshaded, I think all the row heights look great - no complaint. 

- With most configurations of facet selections, the shading looks reasonable. (Question: why do we no longer get to know how many items we've selected once we select them? It seems a little strange to me that the counts disappear under the shading -- but of course we see the count at the top of the results list.) Here are examples of acceptable shading:

image

- However, when you have that unshaded row in the middle, things look a bit off. Level 1 is clearly taller than the other levels, and in Bootstrap2 and Bootprint3, the lower shading of selected Level 1 touches the count circle:

image

- I prefer the way it's done in **dev**, where the shading leaves some blank space above Level 1. All shading seems to be even. Can we do that?

image

Thanks for looking this over!

EreMaijala:
@sturkel89 It could be done, but then it would not be consistent with non-hierarchical facets where the shading fills the whole box. Not sure which way is better.

EreMaijala:
@sturkel89 Here's an attempt at a good enough compromise. I made the padding 0.1rem higher to give a very small space between the badge and the shading.

sturkel89:
@EreMaijala I think that's an excellent compromise!! I'll take it.

image

demiankatz (approved):
Thanks for all the work on this, @EreMaijala and @sturkel89. I think due diligence has been satisfied -- merging now!

EreMaijala:
@sturkel89 #3078 posted about the icons. :)

PR #3048, opened 2023-08-24
Add Lokalise i18n string import command.

demiankatz:
This PR adds a new command line utility that, given paths to a directory containing exported Lokalise language files, and a directory containing VuFind's existing language files, will import new strings into the project. It also normalizes some quirks of Lokalise in both file naming and contents -- specifically, it matches up Lokalise language codes to VuFind language codes, and it removes single quotes surrounding language strings.

TODO
- [x] Merge #3044 before merging this
- [x] Manually test with real Lokalise files
- [x] Figure out how to address remaining quote issues (see #3089)
- [x] Figure out how to address "empty space" issue (see #3106 for work in progress)
- [x] Add test coverage


demiankatz:
@lahmann, I haven't tested this with real Lokalise data yet, but it worked as expected with some fake data that I created. If it's not quite perfect for real data, I suspect only minor tweaks will be needed to get us to where we need to be. Let me know if you have any time to try this out, or if you'd like me to do more testing when time permits! And thanks for all of your help with this project!

lahmann:
Thanks a lot for your work! I tested this pull request with a full language fileset downloaded from our lokalise.com project.
Changes I made:
* added German translations for `loan_history_selected_purged` and `view_in_opac`
* added test-translations for DDC23 `999` in Basque and Chinese
* altered translations for `Account` in Basque and Chinese

The diff after importing the lokalise files shows some minor issues:

1. Trailing and leading whitespaces are being removed - this might result in minor formatting issues of combined strings but could cause more major issues as e.g. for translation tokens such as `number_thousands_separator`.
Some examples:
```
-Export to = "Exportieren nach "
+Export to = "Exportieren nach"

-number_thousands_separator = " "
+number_thousands_separator = ""
```
2. If for usage of placeholders embedded double quotes are at the beginning or end of the translation string the first/last double quote gets removed. As this in some cases will cause problems it also cleans up the language files from multiple double quotes. I suggest to clean up the language files from the irrelevant double quotes and after that fixing the code to skip needed double quotes.
Some examples:
```
-nearby_items = "Werke neben "%%title%%""
+nearby_items = "Werke neben "%%title%%"

-hold_required_by_date_invalid = ""Bitte geben Sie ein 'erforderlich bis'-Datum an"
+hold_required_by_date_invalid = "Bitte geben Sie ein 'erforderlich bis'-Datum an"

-highlight_snippet_html = ""...%%snippet%%...""
+highlight_snippet_html = "...%%snippet%%..."

-tree_search_limit_reached_html = "Roedd gormod o ganlyniadau i'w dangos yn y goeden. Ond yn dangos y <b>%%limit%%</b> eitem gyntaf. I chwilio'r cyfan, cliciwch <a id=fullSearchLink" href="%%url%%" target="_blank">yma.</a>""
+tree_search_limit_reached_html = "Roedd gormod o ganlyniadau i'w dangos yn y goeden. Ond yn dangos y <b>%%limit%%</b> eitem gyntaf. I chwilio'r cyfan, cliciwch <a id=fullSearchLink" href="%%url%%" target="_blank">yma.</a>"

-storage_retrieval_request_error_blocked = "Je hebt onvoldoende rechten om een aanvraag op dit object te plaatsen"""
+storage_retrieval_request_error_blocked = "Je hebt onvoldoende rechten om een aanvraag op dit object te plaatsen""
```

Apart from these issues the import works well and also works fine with creating new files and renaming filenames correctly (as tested with DDC23 files being created for Basque and Chinese). Thanks again!

EreMaijala:
1. I think that "Export to" should be replaced with a translation containing a placeholder for the service name. But the thousand separator is a problem. Perhaps we need a way to escape the space? Or use a non-breaking space (which would probably be the appropriate thing anyway), if it doesn't get trimmed.

demiankatz:
1. I agree, "Export to" should be replaced with a string containing a placeholder. I'm not sure how this got missed in past rounds of language file review designed to prevent this kind of problem... but obviously it did! I can open a follow-up PR next week, unless anyone else has time to do it sooner.

2. I don't think all of the double quotes can be removed -- some of them look like errors, but others are actually needed. For example, the snippet string is designed to enclose snippets in double quotes -- so with VuFind's current language file parsing logic, we need double-double quotes because the outermost quotes get stripped off by the parser, and the remaining ones get displayed. Obviously, the parser is absurdly simplistic, and I've never liked having unescaped double quotes inside other double quotes. Maybe we need to add more sophisticated support for quoting, and a new rule in the normalizer that puts single quotes around strings containing double quotes. I suspect that might play better with Lokalise. What do you think?

demiankatz:
Update: I've opened #3056 to eliminate duplicate double quotes where possible, and I've opened #3057 to use a token in the "export to" translation. This solves many but not all of the identified issues.

@lahmann, do you think it's worth spending time changing the normalization rules so that they put single quotes around strings that contain double quotes? Do you think that will help with the Lokalise import? If so, would you like me to try adding it to #3044, or do you think we should merge that PR as-is and then open a new follow-up PR to experiment with further changes?

I also wonder whether single quotes might help with the `number_thousands_separator` case... or if perhaps a better solution would be to use a non-breaking space (or some other special Unicode space character) to differentiate that from trimmable whitespace. Any thoughts/preferences?

demiankatz:
Another update: #3044 has been approved, so I went ahead and merged it. If we need to add new normalization rules, we can open a separate PR, which is probably better for clarity of the commit history anyway. @lahmann, I'm still interested in your thoughts on the above questions/ideas when time permits.

demiankatz:
I've just opened #3089 with a proposed update to address the remaining quote issues. I think that just leaves the "blank space" issue as the final thing to answer. That might be able to be addressed through another normalizer change that turns empty spaces into non-breaking spaces, or something similar.

demiankatz:
Now that #3106 is merged, this seems to be working entirely correctly -- adding translations on Lokalise, exporting them, and loading them with this tool has the desired/intended effect.

EreMaijala (approved):

PR #3053, opened 2023-08-25
SierraRest: Check holdings also for item statuses by default.

EreMaijala:
This fixes a discrepancy between status information on the search results page and record holdings tab. The issue is particularly noticeable for records that only have summary holdings for storage items.


demiankatz (approved):
Makes sense to me. Thanks, @EreMaijala!

PR #3054, opened 2023-08-28
Check for saved lists element before trying to access it.

EreMaijala:
This eliminates the JS error "savedListsEl is null" when save status check runs on a page that contains records that don't include the element (e.g. favorites).

The error can be reproduced easily by including the check_save_statuses.js script in theme.config.php (which would be necessary to combine all scripts to a single one with the asset pipeline).


demiankatz (approved):
Makes sense -- thanks!

PR #3055, opened 2023-08-28
Fix hierarchy tree search limit default value handling.

EreMaijala:
The default -1 was passed as 0 to getFullFieldFacets which caused the search to complete but to return no results.


demiankatz (approved):
Thanks, @EreMaijala!

PR #3057, opened 2023-08-28
Improve 'Export to' translation string to include token.

demiankatz:
This reduces potentially problematic translation string concatenation, making export menu output easier to customize on a per-language basis. This problem was identified during testing of #3048.

TODO
- [x] Add changelog note about revised translation string when merging.


EreMaijala (changes requested):
There were several cases where the trailing space was missing from the translations, but it didn't cause problems because the template has a space between the translation and the target name. I believe all translations need to be checked and a space added as needed.

demiankatz:
@EreMaijala, I think what you describe is a separate problem that is worth investigating, but which does not necessarily need to delay the merging of this PR. The problem being fixed here is that correct display specifically depended on a blank space being incorporated into the translation string. The problem you describe is that translation strings are incomplete, and spacing is provided by templates. We MUST fix the first problem in order to be compatible with Lokalise; investigating the second problem may reveal some more strings that would be more appropriately presented with tokens, but changing those strings is not critical in order to begin using Lokalise since the platform will not modify them in any way that will negatively impact/change existing behavior.

My suggestion is that we should go ahead and merge this work as-is in order to allow the Lokalise project to move forward, and we should have some further discussion about the other strings so that perhaps we can more comprehensively clean them up as a 10.0 project. If you have time to open a JIRA ticket with some examples that you think need attention, we could continue the conversation over there.

Of course, please feel free to disagree if I'm misunderstanding the situation!

EreMaijala:
@demiankatz Perhaps I wasn't expressing myself clearly. I mean that some of the newly added translation strings are obviously missing a space, e.g. https://github.com/vufind-org/vufind/pull/3057/files#diff-adc13dc35e2770ba83f916b2c07cfbcd2065b28c7657396d4e28fd9472c0849dR239

For the old "Export to" translation it didn't matter, because the templates constructed the results as `transEsc('Export to') + <space> + <target name>`, but now that the placeholder is in the translation, we must check each one and make sure there's a space before `%%target%%`.

EreMaijala:
@demiankatz So the blank space was in fact *not* required to be there in the old translation string.

demiankatz:
Apologies for the misunderstanding, @EreMaijala! I've normalized the spacing in https://github.com/vufind-org/vufind/pull/3057/commits/81e679663ae607977b5831a2a7b38c34dc23390d

EreMaijala (approved):
Thanks, looks good to me now!

PR #3059, opened 2023-08-29
Generate useful OpenURL for eBook

maccabeelevine:
'eBook' is a recognized out-of-the-box VuFind format defined in import/translation_maps/format_map.properties.  getOpenUrlFormat() currently doesn't recognize it and falls back to 'UnknownFormat'.  I would argue that treating it as a book is more useful.


maccabeelevine:
My use case here is specifically for a COinS URL to be picked up by Zotero.  No idea what (bad) side effects this would have for OpenUrl generation in general.  So it's a starting point for discussion hopefully.

EreMaijala (approved):
This is what we do locally as well, so approved. :)

demiankatz (approved):
I was already inclined to approve this, but if @EreMaijala is already doing it in production successfully, I can think of no possible objection. Thanks, @maccabeelevine!

PR #3060, opened 2023-08-30
Always add quotes around names in email headers.

EreMaijala:
This avoids issues when the name contains periods etc.

Since laminas-mail is in security issue maintenance, we need to override the relevant method locally.


EreMaijala:
I have verified that this fixes a problem where DKIM signature check failed before the change apparently because the intermediate SMTP server fixed the header before passing it on.

EreMaijala:
Also worth noting: [RFC 5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4) notes that unquoted form should be used unless the name contains characters that require quotes, but I've added quotes for all cases to keep the change simple. I also expect we will want to migrate to another mail component in v10 or so.

demiankatz:
Thanks, @EreMaijala -- a few questions/comments:

1.) It looks like in the old code, the double quotes were run through the MIME encoding logic, and in the new code they are not. Not sure if this makes a difference, but maybe worth double-checking.

2.) Do we want to move this logic into a trait so that we can apply the same fix to other address lists, not just From?

3.) I've opened [VUFIND-1633](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1633) to track the need to find a better mail library in future.

EreMaijala:
@demiankatz 

1.) I think the old code was a bit off. It added double quotes in a couple of special cases before MIME encoding, but if the name contained non-ASCII characters as well, it would have encoded the quotes as part of the name. The quotes need to surround whatever the end result is, not the other way round. Otherwise they would be displayed as part of the name by the mail client.

2.) We use name only with From address, but since Mailer supports them in all fields, you're right. Now added.

3.) Thanks!

demiankatz (approved):
This seems reasonable to me. Thanks for taking the time to generalize the fix!

PR #3061, opened 2023-08-30
Add place of publication to book OpenUrl

maccabeelevine:
Because it's useful when sent via COinS as a Zotero citation (according to my librarian colleagues).  

That said I'm not clear on how OpenUrl operates, whether all the fields have to match for the link resolver to surface a result.  I wouldn't want a book to fail to resolve due to a difference in place of publication.

### DIFF
--- a/module/VuFind/tests/unit-tests/src/VuFindTest/RecordDriver/DefaultRecordTest.php
+++ b/module/VuFind/tests/unit-tests/src/VuFindTest/RecordDriver/DefaultRecordTest.php
@@ -292,8 +292,9 @@ class DefaultRecordTest extends \PHPUnit\Framework\TestCase
             . 'te=1992&rft_val_fmt=info%3Aofi%2Ffmt%3Akev%3Amtx%3Abook&rft.genre=book&rft.btitl'
             . 'e=La+congiura+dei+Principi+Napoletani+1701+%3A+%28prima+e+seconda+stesura%29+%2F'
             . '&rft.series=Vico%2C+Giambattista%2C+1668-1744.+Works.+1982+%3B&rft.au=Vico%2C+Gi'
-            . 'ambattista%2C+1668-1744.&rft.pub=Centro+di+Studi+Vichiani%2C&rft.edition=Fiction'
-            . 'al+edition.&rft.isbn=8820737493&rft_id=info%3Adoi%2Fxxx&rft_id=pmid%3Ayyy';
+            . 'ambattista%2C+1668-1744.&rft.pub=Centro+di+Studi+Vichiani%2C&rft.place=Morano+%3A'
+            . '&rft.edition=Fictional+edition.&rft.isbn=8820737493&rft_id=info%3Adoi%2Fxxx'
+            . '&rft_id=pmid%3Ayyy';
 
         // Parameters returned by getBookOpenUrlParams with rft_id added
         $openUrlParams = [
@@ -309,6 +310,7 @@ class DefaultRecordTest extends \PHPUnit\Framework\TestCase
             'rft.series' => 'Vico, Giambattista, 1668-1744. Works. 1982 ;',
             'rft.au' => 'Vico, Giambattista, 1668-1744.',
             'rft.pub' => 'Centro di Studi Vichiani,',
+            'rft.place' => 'Morano :',
             'rft.edition' => 'Fictional edition.',
             'rft.isbn' => '8820737493',
             'rft_id' => [
--- a/module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
+++ b/module/VuFind/src/VuFind/RecordDriver/DefaultRecord.php
@@ -825,6 +825,10 @@ class DefaultRecord extends AbstractBase
         if (count($publishers) > 0) {
             $params['rft.pub'] = $publishers[0];
         }
+        $placesOfPublication = $this->getPlacesOfPublication();
+        if (count($placesOfPublication) > 0) {
+            $params['rft.place'] = $placesOfPublication[0];
+        }
         $params['rft.edition'] = $this->getEdition();
         $params['rft.isbn'] = (string)$this->getCleanISBN();
         return $params;


module/VuFind/tests/unit-tests/src/VuFindTest/RecordDriver/DefaultRecordTest.php
```diff
@@ -309,6 +310,7 @@ public function testGetOpenUrl()
             'rft.series' => 'Vico, Giambattista, 1668-1744. Works. 1982 ;',
             'rft.au' => 'Vico, Giambattista, 1668-1744.',
             'rft.pub' => 'Centro di Studi Vichiani,',
+            'rft.place' => 'Morano :',
```
maccabeelevine, line 313: Don't ask me why the actual Solr data (and presumably the MARC source) has a colon after the place of publication, but it does, and that's consistent with my own institution's data.  I guess it's no different than the publisher ending with a comma.  But I hate that stuff.  I mean I know those punctuation will show up in the citation, but it looks silly as data.

demiankatz: Blame the MARC21 standard, which puts trailing punctuation in a lot of fields, undermining the "machine-readable" part of MARC. Not my favorite thing! Maybe it's worth adding a regex to trim trailing punctuation and whitespace when adding the value to the OpenURL. Something like `preg_replace('/[:,\s]*/', '', $placesOfPublication[0])` would probably do the trick for 99% of cases (at least for records cataloged by American standards).

maccabeelevine: Yeah it would make the OpenURL look better but probably break Zotero, if Zotero is assuming that the data already has the MARC21 punctuation and doesn't have to add its own.  I mean MLA, Chicago, APA probably don't all use the same formatting and it probably already has to remove the commas & colons for some of them, but Zotero is mature so I hope it knows what it's doing.  And it must expect this stupid trailing punctuation since as you say it's standard.

maccabeelevine: Ok well I overcame my laziness and actually tested Zotero (with the latest Chicago style), and it has pretty good logic for adding the trailing punctuation if it's missing but not duplicating it if it's present.  That said I think any effort to clean up the MARC-cruft from the OpenUrl export should be across all the affected fields, not a one-off here.

demiankatz: Well, it's standard in MARC, but I don't think it's standard in OpenURL. Is Zotero stripping it if you try it as-is? I don't think stripping the punctuation would hurt Zotero, since there's certainly no guarantee/expectation for such punctuation to be present in the OpenURL.

demiankatz: But fair point about consistency. :-)

maccabeelevine: I'd also say if we did the stripping (in a separate PR) it should be done in SolrMarc or some trait, not DefaultRecord, because who knows if ending whitespace or punctuation might actually be valid for some records.  I hate this stuff :)

demiankatz: It would have to be done in a trait, because some of this data is pulled from the raw MARC record stored in Solr, not from the separate fields extracted by SolrMarc (because the Solr schema doesn't have the required level of granularity).

demiankatz:
@maccabeelevine, it might be helpful to try some real-world experiments to see if the place of publication field in the OpenURL has any practical effect on successful linking. My suspicion is that it should not cause any harm, but it would be nice to see whether or not we can find a way to break link resolution with the place field.

For starters, I did a very simple test with EBSCO Full Text Finder -- I grabbed a random OpenURL from a COinS field on a random book record in my catalog, and I passed it to FTF. Since this was an OpenURL for a physical book, my only option was to request via ILL. I tried this, and my ILLiad form correctly pre-filled (though the publisher field was blank even though there was a publisher name in the OpenURL). I then added the place field and tried again. Exact same result. So it seems that FTF is just ignoring all the publisher information in the scenario I tested, which might be due to some configuration issue in Villanova's instance. Bottom line: this is entirely anecdotal and not very conclusive, but in this particular situation, it doesn't seem to make any difference at all.

It would be better to find an eBook openURL that actually resolves to a full text platform, and do similar experimentation there. I'm out of time for testing this today, but is that something you might be able to dig up?

EreMaijala:
Just a couple of notes:

1.  OpenURL shouldn't include MARC punctuation, and no system should rely on that.
2. I'd think of building an OpenURL as "try to put everything we have into it because we don't know which fields the resolver will actually need". The more metadata there is, the more likely the resolution works (and extra metadata shouldn't hurt, unless it's an invalid ISBN or something like that), and even if not all the data is needed, it could be used for display purposes, just for collecting the metadata or passed on by the resolver.
3. Automatically cleaning up some of the punctuation is kinda hard, particularly with periods. RecordManager has [a method](https://github.com/NatLibFi/RecordManager/blob/dev/src/RecordManager/Base/Utils/MetadataUtils.php#L565) that does a reasonable job, but it requires a list of abbreviations to avoid removing period after an abbr.

demiankatz:
Thanks, @EreMaijala, that all matches my gut feelings, but it's helpful to have somebody say it with more confidence than I felt competent to. ;-)

So I guess the question is whether to merge this as-is and introduce punctuation stripping later, or to use this as the first case to introduce a punctuation stripping mechanism. I think in the case of place names, stripping punctuation other than periods should be safe, and I can't think of a scenario where an unwanted period is likely to be introduced.

However, if we want to introduce more sophisticated functionality from RecordManager, I wonder if it's worth considering starting a new vufind-org/vufind-metadata-utils library that could pool some of the current utilities from the XSLT namespace with some of the (generally much richer/more robust) functionality from RecordManager. No pressure if that doesn't sound like a good idea to you, but I'm certainly open to investigating such a thing if you think it would be valuable to extract and share the logic.

maccabeelevine:
> It would be better to find an eBook openURL that actually resolves to a full text platform, and do similar experimentation there. I'm out of time for testing this today, but is that something you might be able to dig up?

I'm not managing to figure out how in our institution's FTF setup to get an OpenUrl to an eBook.  Happy to hear @EreMaijala confirm that more metadata is better and link resolvers are smart enough to use what they need.

> Automatically cleaning up some of the punctuation is kinda hard

Yeah I still favor not using this PR as a lead case for stripping punctuation -- I know I raised the issue, but more to explain why I was leaving it in.  Last week at WOLFcon we ate at the hotel restaurant "Truth Be Told," with trailing comma.  There are so many edge cases and exception cases.  I think what RecordManager is doing is probably smart, but applying it to this 16-year-old OpenUrl functionality is probably a larger conversation.

EreMaijala:
@demiankatz No objection to adding a new module with metadata utils, but I agree with @maccabeelevine that it should be done separately. And there will always the edge case where things go wrong. Luckily, the punctuation should make no difference with OpenURL resolution itself.

demiankatz (approved):
Thanks, @EreMaijala -- I think that means it makes the most sense to merge this now as-is, and we can follow up with metadata cleanup as time permits. @maccabeelevine, is this something you are interested in looking into? If so, let us know how we can help. If not, maybe we should open a JIRA ticket and target release 10.0 so we don't forget to double back to this when time permits.

PR #3062, opened 2023-08-30
Import native functions to improve opcache performance

LuomaJuha:
This pr imports native functions so opcache can turn them into opcode and improves loading times.

Here is some conversation about the topic: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/3048 and the rule added: https://cs.symfony.com/doc/rules/function_notation/native_function_invocation.html


LuomaJuha:
Seems like small side-effects happened. Trying to adjust them at some point.

demiankatz:
Thanks, @LuomaJuha. Is it possible that php-cs-fixer has a setting to insert use statements for functions instead of the leading backslashes? I think that might improve readability and reduce the number of long line issues created by the change.

LuomaJuha:
@demiankatz     'global_namespace_import' => [
        'import_constants' => null,
        'import_functions' => true,
        'import_classes' => null,
    ], seems to do what I am looking for but for some reason it is not working. Gonna have to try and see why,

demiankatz:
Thanks, @LuomaJuha -- that does look like it would be a better solution if you can get it to work. If you continue to have trouble with it, please let me know and I'll be happy to experiment on my end when time permits.

LuomaJuha:
@demiankatz Hmm, seems like I have to backslash the functions before the global namespace import understands it, will try and push the new changes soon

LuomaJuha:
Well, it almost worked like a charm but the use statements looks very awful. And non-class php files dont add use statements, but I think that is okay?

tests/vufind.php-cs-fixer.php
```diff
@@ -19,12 +19,18 @@
...
+        'import_functions' => true,
+        'import_classes' => null,
+    ],
     'is_null' => true,
     'linebreak_after_opening_tag' => true,
     'lowercase_cast' => true,
     'magic_constant_casing' => true,
     'modernize_strpos' => true,
     'native_function_casing' => true,
+    'native_function_invocation' => true,
```
EreMaijala: Would it make sense to do this only for the optimizable functions? See https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/v2.15.1/src/RuleSet.php#L194 and https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/4448.

LuomaJuha: Good idea, ill add and test that also

demiankatz:
@LuomaJuha, it makes sense that the template files would not add use statements, since you only need a use statement to eliminate ambiguity when you declare a namespace, and none of the template files declare namespaces. Thus, I suspect they don't need any changes at all -- the function calls in those files MUST be the PHP-land versions, since there is no way to declare a user-land override without a namespace. (Though please correct me if I'm overlooking something!)

LuomaJuha:
@demiankatz holy smokes, I managed to get it working as intended. I also updated the php-cs-fixer to higher version in this pr.

tests/vufind.php-cs-fixer.php
```diff
@@ -19,12 +19,21 @@
     'concat_space' => ['spacing' => 'one'],
     'ereg_to_preg' => true,
     'get_class_to_class_keyword' => true,
+    'global_namespace_import' => [
+        'import_constants' => null,
```
demiankatz: Since null is the default value for import_constants, would it be better to omit it here? Or do you think it might be worth changing to true in a subsequent PR?

tests/vufind.php-cs-fixer.php
```diff
@@ -19,12 +19,21 @@
...
+        'import_functions' => true,
+        'import_classes' => null,
```
demiankatz, line 24: I see that the default value for import_classes is true. I assume you're setting it to null here to avoid lots of unrelated changes getting mixed together -- which I think is a good idea! But should we consider setting this to true in a follow-up PR?

tests/vufind.php-cs-fixer.php
```diff
@@ -45,7 +54,10 @@
     'no_useless_return' => true,
     'no_whitespace_in_blank_line' => true,
     'non_printable_character' => true,
-    'ordered_imports' => true,
+    'ordered_imports' => [
```
demiankatz, line 56: You changed ordered_imports here, but not in the template configuration below. I realize that this is probably because the setting is not particularly relevant to templates... but maybe it's better to make the configurations as consistent as possible so this doesn't confuse us in future if we compare the two files. Or is there a reason they need to be different?

demiankatz (changes requested):
Thanks, @LuomaJuha, this is great! I have a few small questions about the configuration settings and possible next steps, but as far as I can tell, all of the automatically generated changes here are correct and appropriate, and the full test suite is still passing. I'll be happy to approve and merge this once we finish making any minor adjustments to the configuration files  (which will likely have no immediate impact on any of the other files).

tests/vufind.php-cs-fixer.php
LuomaJuha: I think it would be a great followup

tests/vufind.php-cs-fixer.php
LuomaJuha: sure.

tests/vufind.php-cs-fixer.php
demiankatz: Great, please feel free to try it out after I merge this, and if you like the results, you can open a follow-up PR!

demiankatz (approved):
Looks good to me, and all tests still pass. Thanks again, @LuomaJuha!

PR #3064, opened 2023-08-31
Autocomplete: reorganize modules and update to v2.1.8

crhallberg:
Separated from #2979.

TODO
- [x] Merge/incorporate https://github.com/vufind-org/autocomplete.js/pull/15 before completing this
- [x] Will update vendor when all tests pass and errors are resolved.
- [x] Update changelog when merging (note refactored Javascript).


themes/bootstrap3/js/common.js
```diff
@@ -706,6 +707,7 @@ $(function searchFormResetHandler() {
   });
   resetButton.on('click', function onClick() {
     queryInput.attr('value', '').focus();
+    queryInput[0].ac.hide(); // not sure why AC is using previous value
```
EreMaijala: That's because autocomplete listens to the input event, and clearing the contents programmatically doesn't trigger the event. Something like this should help:

`queryInput.attr('value', '').trigger('focus')[0].dispatchEvent(new Event('input'));`

(`focus()` is deprecated, that's why `trigger('focus')` instead). This will still display the "Loading..." message, and I'm not sure if that's correct, but it's the same with the initial keystrokes so not in scope here.

EreMaijala (dismissed):

themes/bootstrap3/js/common.js
crhallberg: This definitely answers my question, but I think the hide method is better since it doesn't flash Loading or anything else.

EreMaijala: I'm fine with that, but perhaps autocomplete could also listen for the `change` event to reset when a field is emptied? I think it would be better to not leave the old state dangling regardless of whether it shows.

demiankatz:
@crhallberg, after looking at @ThoWagen's work on #2952, where he significantly refactored the search box reset button to incorporate it with his virtual keyboard support, I begin to wonder why we need to focus the search box in order to make the reset button work. I'm pretty sure it's that added focus call that introduced the problem that @EreMaijala first observed. It's entirely possible that the autocomplete improvements made here are worthwhile and helpful, but I wonder if we should also consider introducing a simplified version of searchbox_controls.js from #2952 into this PR (or a new PR) in order to begin taking advantage of that refactoring, which seems to keep things working while eliminating the extra focus call. What does everyone think? I'd also be interested to hear @xmorave2's opinion, since he might have more insight into whether the focus is needed for something I'm overlooking.

themes/bootstrap3/js/common.js
crhallberg: True. I’ll add a “change” listener and see if it fixes this issue without doubling up on what “input” is already doing.

crhallberg:
@demiankatz I agree that would be easier, but my understanding was that the focusing is not to activate the clear button, but as an autofocus functionality. I’ll give it a second look though, you may be right.

demiankatz:
@crhallberg, I don't think autofocus was intended -- I believe we had previously discussed the negative accessibility implications of autofocus and decided against it, but I just let the change through without realizing the consequences.

crhallberg:
I think you were right about the autofocus. I've addressed it in the latest commit.

themes/bootstrap3/js/common.js
```diff
@@ -365,21 +365,6 @@ function escapeHtmlAttr(str) {
   });
 }
 
-function extractClassParams(selector) {
```
demiankatz, line 433: extractClassParams is also used by openurl.js, so it should be moved back and added to the "exported" list here (that must have been an oversight before). Also, since you seem to have changed the signature, we need to be sure that nothing gets broken.

themes/bootstrap3/js/searchbox_controls.js
```diff
@@ -0,0 +1,132 @@
...
+    if (classes[i].indexOf(':') > 0) {
+      const pair = classes[i].split(':');
+      params[pair[0]] = pair[1];
+    }
+  }
+
+  return params;
+}
+
+function setupAutocomplete() {
```
demiankatz: Should this function be moved inside the "searchbox_controls" module since it is only used there?

themes/bootstrap3/js/searchbox_controls.js
```diff
@@ -0,0 +1,132 @@
...
+      const value = typeof event.detail === "string"
+        ? event.detail
+        : event.detail.value;
+      searchbox.setAttribute("value", value);
+      document.getElementById("searchForm").submit();
+    });
+  }
+}
+
+function setupSearchResetButton() {
```
demiankatz: Should this function be moved inside the "searchbox_controls" module since it is only used there?

demiankatz (changes requested):
Thanks, @crhallberg. There are a few issues noted below, but overall I think this is a step in the right direction! However, this also seems like more refactoring and changes than is strictly necessary to fix the problem at hand. I'm going to open a separate PR to just extract the bare minimum changes that I think are necessary to solve our issues in 9.1, and then we can target the rest of this to 10.0. Any objections?

demiankatz:
Also, I'm encountering a test failure:

```
There was 1 failure:

1) VuFindTest\Mink\AutocompleteTest::testBasicAutocomplete
--- Expected
+++ Actual
@@ @@
-'http://localhost/vufind_test/Search/Results?lookfor=Fake+DOI+test+1&type=AllFields'
+'http://localhost/vufind_test/Search/Results?lookfor=fake+doi+test&type=AllFields'

/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/AutocompleteTest.php:90
/usr/local/vufind/module/VuFind/src/VuFindTest/Feature/AutoRetryTrait.php:104

FAILURES!
Tests: 1, Assertions: 55, Failures: 1.
```

This test is passing on dev, so something has somehow gotten broken here.

demiankatz:
See #3080 for my simplified minimal fix. I have moved this PR's milestone to 10.0 -- we should revisit this after #3080 is merged and conflicts are resolved.

demiankatz:
Okay, now that #3080 is merged, we need to resolve conflicts here when time permits. For the moment, finishing up the 9.1-related pull requests is a higher priority, but hopefully we can clean this up early in the 10.0 development process.

demiankatz:
@crhallberg, can you take another look at this when time permits? Is it worth trying to resolve conflicts here, or have things moved so much that we're better off closing and taking a fresh approach? I honestly can't remember what we were trying to accomplish here (apart from fixing the bug cited in the description, which is already fixed now).

crhallberg:
I merged in dev but that caused some of the de-jQuery work to be reverted. We can address that after merge.

demiankatz:
Thanks, @crhallberg. It looks like there's some unaddressed feedback from a previous review, if you have a chance to look at that too. Also, I'm pretty sure this PR needs to be renamed, because the autocomplete opening problem was solved elsewhere. Is this applying more generalized fixes/improvements now?

themes/bootstrap3/js/common.js
demiankatz: Just a reminder that this still needs to be done.

demiankatz (changes requested):
Just a reminder of an open review comment from before... plus it looks like the build is failing due to some automated JS test failures. Is it possible that some code got duplicated in a merge? (Just a guess based on a very cursory glance).

themes/bootstrap3/js/vendor/autocomplete.js
```diff
@@ -73,6 +73,7 @@ function Autocomplete(_settings) {
     list.classList.remove("open");
     _currentIndex = -1;
     lastInput = false;
+    lastCB = false;
```
demiankatz, line 80: The lastCB variable is not declared until line 172 below. Is it a problem that you access it here? Does the declaration need to be moved up?

themes/bootstrap3/js/searchbox_controls.js
```diff
@@ -157,6 +297,8 @@ VuFind.register('searchbox_controls', function SearchboxControls() {
       });
     }
 
+    // Setup keyboard
+
```
demiankatz: I'd remove the blank line here so the comment is touching the related code.

themes/bootstrap3/js/searchbox_controls.js
```diff
@@ -138,7 +139,146 @@ VuFind.register('searchbox_controls', function SearchboxControls() {
...
+      queryInput.focus();
+      queryInput.ac.hide();
+    });
+  }
+
+  function init() {
+    setupAutocomplete();
+    setupSearchResetButton();
+
+    // Setup reset button
```
demiankatz: I'm a little confused by the refactoring around the reset button. This comment seems redundant, and I'm not sure where the setupSearchResetButton() logic is coming from. The reset button logic is already incorporated into the `_handleInputChange` method, so these other listeners seem redundant.

demiankatz (changes requested):
Thanks for the progress, @crhallberg! Tests are passing, but I think there's still a bit more cleanup needed here.

themes/bootstrap3/js/searchbox_controls.js
crhallberg: It looks like some redundancy was added when the virtual keyboard was merged in. I think I've cleaned it all up.

themes/bootstrap3/js/searchbox_controls.js
```diff
@@ -138,25 +133,145 @@ VuFind.register('searchbox_controls', function SearchboxControls() {
...
+          input.value = value;
+          input.form.submit();
+        });
+      }
+    });
+  }
+
+  function setupSearchResetButton() {
+    _textInput = document.getElementById("searchForm_lookfor");
+    _resetButton = document.getElementById("searchForm-reset");
```
demiankatz, line 278: We should return early if _resetButton is not found, in case the element is not on the page for any reason.

themes/bootstrap3/js/searchbox_controls.js
```diff
@@ -18,11 +19,6 @@ VuFind.register('searchbox_controls', function SearchboxControls() {
   };
 
   function _handleInputChange(input, triggerInputEvent = true) {
-    _textInput.value = input;
```
demiankatz, line 21: I think this removed code is necessary because of the virtual keyboard -- this function is not only called in response to normal input events, but also in response to user interaction with virtual keyboards. For the same reason, it may not be possible to give the reset button its own separate event handler -- it may be better to restore this code instead of setting up a separate event listener below. (In any case, note that the use of classList.toggle here is a bit more compact than the if/else in the version below).

demiankatz (changes requested):
Thanks, @crhallberg -- I think we're getting there, but if I'm remembering correctly how this works, there may still be some issues related to the virtual keyboard implementation.

themes/bootstrap3/js/searchbox_controls.js
crhallberg: I like the more compact syntax but I'm in favor of keeping the listeners separate. We should only set up a listener if that part is enabled. Keeping them separate means the autocomplete, reset button, and virtual keyboard are all independent of each other instead of doing the if-statement checks below.

crhallberg: Also, setting the value is redundant because we're passing the value into this function and the dispatch is redundant because this is triggered by the same event.

demiankatz: Have you tested this with the virtual keyboard enabled? Maybe I'm misunderstanding something, but my impression is that the virtual keyboard does not trigger normal JS events but instead directly calls the _handleInputChange method... so if you remove the reset button logic and value setting from _handleInputChange, it will break the virtual keyboards.

demiankatz (changes requested):
Looks like some recent fixes to the virtual keyboard have introduced conflicts here... and it's probably not worth trying to resolve them until after #3250 is finished as well. After that, we should make sure that all of those functional improvements are not lost in the refactoring here.

demiankatz (changes requested):
Now that #3250 has been merged, there are more conflicts to resolve. Can you sort it out when time permits, @crhallberg? Hopefully this gets the virtual keyboard logic into a stable place and we won't have to reconcile any further changes.

crhallberg:
I've resolved the conflict but I'm seeing two issues that I'm investigating:

- [x] When I move the caret to the middle of the text, the virtual keyboard correctly inserts a few characters before jumping to the end.
- [x] When starting with a search and clicking the reset button, the keyboard appears but the text does not clear.

*EDIT: I patched these errors but I'm unsure of the true cause. My top culprit is the virtual keyboard itself trying to sync the input based on the settings we're using.*

crhallberg:
This is behaving all the ways I would expect - I cannot seem to break it. @ThoWagen, I would appreciate a look from you to double-check my refactoring. @sturkel89, if you have time, I wouldn't mind you trying to break it as well. 

Thank you!

config/vufind/searchbox.ini
```diff
@@ -69,7 +69,7 @@ group[] = "Other Sites"
 ;layouts[] = "burmese"
 ;layouts[] = "chinese"
 ;layouts[] = "czech"
-;layouts[] = "english"
+layouts[] = "english"
```
ThoWagen: Is the english keyboard supposed to be enabled by default?

themes/bootstrap3/js/searchbox_controls.js
```diff
@@ -155,31 +164,150 @@ VuFind.register('searchbox_controls', function SearchboxControls() {
...
+  function setupSearchResetButton() {
+    _resetButton = document.getElementById("searchForm-reset");
 
-    if (!_textInput) {
+    if (!_resetButton) {
       return;
     }
 
-    _resetButton = document.getElementById('searchForm-reset');
+    _textInput = document.getElementById("searchForm_lookfor");
```
ThoWagen: `_textInput` should not be initialized here. It would be possible that there is no `_resetButton` but you still need `_textInput` for the virtual keyboard.

ThoWagen (commented):
Besides two small comments, I don't see any problems with this. I could not break the behavior as well. The virtual keyboard, the reset button and autocomplete are working well together. 

The idea of `_handleInputChange` was to have one method that can handle all of the events and behavior independent of what caused the change. But I'm not convinced myself that this is a better approach than this.

demiankatz (changes requested):
Thanks, @ThoWagen, I agree that these things need to be fixed.

@crhallberg, can you make the necessary adjustments and then request a review from @sturkel89 to get some final hands-on testing of everything?

config/vufind/searchbox.ini
crhallberg: This was a dev change that slipped through. Thank you!

demiankatz (changes requested):
Thanks, @crhallberg, I think this looks good now! All that's left to do is to mint the new release at the autocomplete.js repo and then update package.json here. Can you do that?

demiankatz (approved):
Thanks, @crhallberg, looking good!

PR #3065, opened 2023-08-31
Fix SolrAuth typo

maccabeelevine:
Looks to me like "S"olrAuth should be capitalized in these uses like the rest of the file?  Not tested.


EreMaijala (approved):

EreMaijala:
@maccabeelevine Oops, you're right, it should have been capitalized. Thanks for the fix!

PR #3067, opened 2023-09-01
Fix broken list tag query (and add test)

demiankatz:
While testing @skellamp's work on Doctrine migration, I discovered that the list tag functionality in the current dev branch was broken due to illegal grouping. (It probably worked in the past because MySQL used to have much more tolerant defaults... but I bet it was always broken for PostgreSQL). This PR fixes the grouping problem by removing some unused/unnecessary fields from the select clause and also adds a very basic Mink test to make sure that lists can be tagged, and tagged lists can be retrieved by the ListItems ChannelProvider.

TODO
- [x] Fix PostgreSQL compatibility

### DIFF
--- a/module/VuFind/src/VuFind/Db/Table/ResourceTags.php
+++ b/module/VuFind/src/VuFind/Db/Table/ResourceTags.php
@@ -230,7 +230,7 @@ class ResourceTags extends Gateway
             $publicOnly,
             $andTags
         ) {
-            $columns = [Select::SQL_STAR];
+            $columns = ['id' => new Expression('min(resource_tags.id)'), 'list_id'];
             if ($andTags) {
                 $columns['tag_cnt'] = new Expression(
                     'COUNT(DISTINCT(?))',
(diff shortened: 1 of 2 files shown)


demiankatz:
I just ran the test and discovered that even after my fixes, this remains broken in PostgreSQL. Here's the error message:

```
Statement could not be executed (42703 - 7 - ERROR: column "tag_cnt" does not exist LINE 1: ...er($2)) GROUP BY "resource_tags"."list_id" HAVING tag_cnt = ... ^ HINT: Perhaps you meant to reference the column "resource_tags.tag_id".)
```

I'm not particularly familiar with PostgreSQL, so I'm not sure why this is breaking. Are you not able to use a named value from the SELECT clause in the HAVING clause? Any ideas?

demiankatz:
Looks like I answered my own question. We don't actually need to select the tag count because it's only used for the HAVING clause, so I rearranged and simplified the code, and now things are working in both MySQL and PostgreSQL.

EreMaijala (approved):

PR #3068, opened 2023-09-04
Add support for advanced search with Blender.

EreMaijala:
This PR adds support for advanced search with Blender. It also adds a proper controller for Blender for the different actions and so that Blender can be used as the default module as desired. The old Search/Blended route is still supported for back-compatibility. The help pages are reduced from the pretty Solr-specific advsearch.phtml and search.phtml. The new mechanism for overriding them could be used e.g. for EDS and Primo as well.

TODO
- [x] Update changelog when merging (renaming of Search/Blended route)


config/vufind/Blender.ini
```diff
@@ -74,13 +74,27 @@ retain_filters_by_default = true
...
 Title = Title
 Author = Author
 Subject = Subject
 
+; This section defines which search options will be included on the advanced
+; search screen.
+; The name of each setting below corresponds with a Search/Fields mapping in
+; BlenderMappings.yaml.
+; Note that you will need to ensure that all enabled backends support advanced
+; searches (Solr, SolrAuth, EDS and Primo do).
```
demiankatz: Maybe we should recommend an action here if they DON'T support advanced searches. Something like "You can comment out or delete all of the options in this section to disable advanced search in situations where all backends cannot support it."

module/VuFind/src/VuFind/Controller/BlenderController.php
```diff
@@ -0,0 +1,68 @@
...
+     * Is the result scroller active?
+     *
+     * @return bool
+     */
+    protected function resultScrollerActive()
+    {
+        $config = $this->serviceLocator->get(\VuFind\Config\PluginManager::class)
+            ->get('config');
+        return isset($config->Record->next_prev_navigation)
+            && $config->Record->next_prev_navigation;
```
demiankatz: This could be `$config->Record->next_prev_navigation ?? false` -- I imagine this was copy-pasted from elsewhere, but probably doesn't hurt to optimize it.

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
```diff
@@ -455,6 +455,7 @@ protected function unFindCss(
             try {
                 $elements = $page->findAll('css', $selector);
                 if (!isset($elements[$index])) {
+                    $this->assertNull(null);
```
demiankatz, line 462: Is this to satisfy tests that would otherwise complain that they have no assertions? If so, maybe adding a comment here to explain that will prevent future confusion.

demiankatz (changes requested):
@EreMaijala, I haven't had time for a hands-on test yet, but here are a few initial comments.

config/vufind/Blender.ini
EreMaijala: Good idea, and thanks for the suggestion. Now added.

module/VuFind/src/VuFind/Controller/BlenderController.php
EreMaijala: Fixed.

module/VuFind/src/VuFindTest/Integration/MinkTestCase.php
EreMaijala: Yes, and it also makes the check count in test statistics. Comment added.

demiankatz (approved):
Looks good and works for me. I wish there was a better way to manage the help translations since so much of the text is redundant, but I don't have any brilliant ideas there.

EreMaijala:
@demiankatz I agree that there's a lot of redundant text, but I think it still makes sense from the point of customization. It's easier to update complete pages e.g. if you want to give your audience specific hints or search examples, and you can be sure that they don't affect other pages.

PR #3069, opened 2023-09-05
Search Result Explanation Feature

ThoWagen:
It is not always obvious why a record was found or how it is ranked in the result list. Solr can give you an explanation for this but it is far from easy to understand on first glance. This feature allows to view this explanation in a more readable and visual format. It can help librarians, developers to optimize the searchspecs.yaml or help them and even advanced users to better understand the search results. This feature was presented at Wolfcon 2023.

It can be activated by enabling it under [Explain] in the searches.ini and should work with all Solr versions. However, the format of Solr's explanation varies from version to version and it might be a good idea to test this on multiple systems and with different indexes.

Note: Under languages/ExplainFieldName, descriptions can be included for fields in Solr that might not be obvious. This obviously depends on the index schema. So for now, I only included three fields as an example. I am not sure if and which fields of the standard VuFind schema would need a description.


demiankatz:
Thanks for sharing this, @ThoWagen! I'm putting it in the 10.0 milestone, since I think it is unlikely that this will be fully reviewed and revised in time for release 9.1, but I want to be sure we don't lose track of it. I'm presently very busy trying to finish up 9.1-related work, but I will try to circle back and give this a first review as soon as it is practical to do so!

languages/ExplainFieldName/de.ini
```diff
@@ -0,0 +1,4 @@
+;description of fields in explain
+allfields_unstemmed = "Alle Felder"
```
demiankatz: I notice that the English strings mention copyField, but the German ones do not. Is this intentional?

languages/de.ini
```diff
@@ -1130,6 +1141,7 @@ relais_success_message = "Bestellnummer %%id%% wurde angelegt. Sie erhalten eine
 Related Author = "Ähnliche Verfasser"
 Related Items = "Ähnliche Datensätze"
 Related Subjects = "Ähnliche Schlagworte"
+Relevance = "Relevanz"
```
demiankatz, line 1149: There seem to be a few strings in the German file here which are not present in English. Is this intentional?

themes/bootstrap3/templates/record/explain.phtml
```diff
@@ -0,0 +1,13 @@
+<?php if ($this->disabled): ?>
+  <h3><?= $this->transEsc('explain_disabled', ['%%searchClassId%%' => $this->driver->getSourceIdentifier()]) ?></h3>
```
demiankatz: Very minor style suggestion: we don't usually include a space after `<?=`.

demiankatz (changes requested):
Thanks again, @ThoWagen! I haven't had time for a full review yet, but I took a few minutes today to resolve conflicts here and to begin to look over the code. See below for a few initial comments. I will add more as I have time to review more closely.

languages/de.ini
ThoWagen: For those strings the English version is the key itself. So as far as I know they don't need to be translated. E.g. for "Relevance" I would only add Relevance = "Relevance".
Please let me know if it is better to provide a translation in these cases anyway.

demiankatz: Please do include translations in these cases as well, even if the key matches the value -- we use the English language file to identify which strings need to be translated into all of the other languages, so if they are omitted from English, they will not be added to any of the other languages during our translation processes.

EreMaijala: Also consider e.g. the situation where English is a secondary language in a multilingual UI. If a translation is not found for the current language, VuFind will try the fallback languages.

languages/ExplainFieldName/de.ini
ThoWagen: I changed this. Hopefully this is better understandable now. But as mentioned in the initial description I am not sure about those field explanations. For now these are just three examples. Someone with more experience with the standard VuFind index could/should have a look into it.

languages/de.ini
ThoWagen: Makes sense. Thanks for the explanation :)
I fixed this.

themes/bootstrap3/templates/record/explain.phtml
ThoWagen: Fixed. In my opinion spaces improve the readability. But maybe I'm just more used to it.

config/vufind/searches.ini
```diff
@@ -861,3 +862,16 @@ maxLimit = 100
 ; ttl, Time to Live, i.e. cache entry life time in seconds. 300 seconds by default.
 ;options[ttl] = 300
 
+; This section provides settings for the explain feature. When enabled the result
+; list contains links to the explanation why a title was found and how the relevance
+; is calculated.
+[Explain]
+;enabled = true
```
EreMaijala, line 929: This could use a comment such as 'The explain feature is disabled by default. Uncomment the following line to enable it.'

languages/ExplainFieldName/de.ini
```diff
@@ -0,0 +1,4 @@
+;description of fields in explain
```
EreMaijala: How about naming the text domain "IndexFieldName"? I can imagine that one might want to include the descriptions in the staff record view as well, and this domain name wouldn't limit the scope to the explain feature.

module/VuFind/src/VuFind/Search/Base/Explanation.php
```diff
@@ -0,0 +1,148 @@
...
+     * @var \Laminas\Config\Config
+     */
+    protected $config;
+
+    /**
+     * Configuration file to read search settings from
+     *
+     * @var string
+     */
+    protected $searchIni = 'searches';
```
EreMaijala, line 65: How would this get set for e.g. Search2?

module/VuFind/src/VuFind/Search/Explanation/ExplanationFactory.php
```diff
@@ -0,0 +1,89 @@
+<?php
```
EreMaijala, line 1: We usually have the factory in the same directory as the class. I think that makes the relationship more obvious, and if you'd need its own factory for e.g. the Solr version it would also be in the Solr directory.

module/VuFind/src/VuFind/Search/Solr/Explanation.php
```diff
@@ -0,0 +1,531 @@
...
+     * @var array
+     */
+    protected $explanation = [];
+
+    /**
+     * Describes the rest. It has restValue and the percentage from total value.
+     *
+     * @var array
+     */
+    protected $rest = null;
```
EreMaijala, line 101: Shouldn't this be initialized as `[]`?

module/VuFind/src/VuFind/Search/Solr/Explanation.php
```diff
@@ -0,0 +1,531 @@
...
+            $this->boost = $this->parseLine(array_pop($lines));
+            if ($this->boost['value'] > 0) {
+                $this->baseScore = $this->baseScore / $this->boost['value'];
+            }
+        }
+
+        // handle coord
+        if (!empty($lines) && str_contains($this->parseLine(end($lines))['description'], 'coord')) {
+            $this->coord = $this->parseLine(end($lines));
+            $this->baseScore = $this->baseScore / $this->coord['value'];
```
EreMaijala: Is `$this->coord['value']` guaranteed to be non-zero?

module/VuFind/src/VuFind/Search/Solr/Explanation.php
```diff
@@ -0,0 +1,531 @@
...
+            $maxChild = null;
+            foreach ($this->getChildLines($lines, $curLevel) as $child) {
+                if ($this->parseLine($child)['value'] > $maxValue) {
+                    $maxValue = $this->parseLine($child)['value'];
+                    $maxChild = $child;
+                }
+            }
+        }
+
+        //summery of lower childes
```
EreMaijala: "summary of lower children"?

themes/bootstrap3/less/components/explain.less
```diff
@@ -0,0 +1,46 @@
...
+  .exact-match {
+    color: black;
+  }
+
+  .inexact-match {
+    color: #436c71;
+  }
+}
+
+
```
EreMaijala: Trim trailing lines

config/vufind/searches.ini
```diff
@@ -861,3 +862,16 @@ maxLimit = 100
...
+; The explanation is split into the main fields and a rest.
+; minPercent gives a lower bound for fields to be included. All fields with
+; a percentage lower than minPercent are summarized in the rest.
+;minPercent = 0
+; maxFields gives a upper bound for the number of main fields. Set negative for no boundary.
+;maxFields = -1
+; Number of decimal places to be displayed.
+;decimal_places = 2
```
EreMaijala: Other settings are camelCaps, so maybe `decimalPlaces`?

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain.phtml
```diff
@@ -0,0 +1,173 @@
...
+        <?=$this->localizedNumber($explainElement['percent'], $decimalPlaces)?>%
+      </td>
+      <td>
+        <span class="<?=$explainElement['exactMatch'] ? 'exact-match' : 'inexact-match'?>"><?=$explainElement['fieldValue']?>"</span>
+        <?=$this->transEsc('in')?>: <?=$explainElement['fieldName']?>
+        <?php if ($fieldNameKey !== $fieldName): ?>
+          (<?=$fieldName?>)
+        <?php endif; ?>
+        <?php if (isset($explainElement['fieldModifier'])): ?>
+          , <?=$this->transEsc('explain_modifier', ['%%modifier%%' => $this->localizedNumber($explainElement['fieldModifier'], $decimalPlaces)])?>
```
EreMaijala: This causes a space to be included before the comma. It would look better without the space or the comma.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain.phtml
```diff
@@ -0,0 +1,173 @@
...
+        $shortExplain[] = $explainElement['value'];
+        $fieldNameKey = $explainElement['fieldName'];
+        $fieldName = $this->transEsc('ExplainFieldName::' . $fieldNameKey);
+      ?>
+      <td class="percentage">
+        <?=$this->localizedNumber($explainElement['percent'], $decimalPlaces)?>%
+      </td>
+      <td>
+        <span class="<?=$explainElement['exactMatch'] ? 'exact-match' : 'inexact-match'?>"><?=$explainElement['fieldValue']?>"</span>
+        <?=$this->transEsc('in')?>: <?=$explainElement['fieldName']?>
```
EreMaijala: I'd leave the colon out and add a span with a class around the field name so that it can be styled to stand out.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain.phtml
```diff
@@ -0,0 +1,173 @@
...
+          . ' (' . $explainElement['fieldValue'] . ')';
+        $shortExplain[] = $explainElement['value'];
+        $fieldNameKey = $explainElement['fieldName'];
+        $fieldName = $this->transEsc('ExplainFieldName::' . $fieldNameKey);
+      ?>
+      <td class="percentage">
+        <?=$this->localizedNumber($explainElement['percent'], $decimalPlaces)?>%
+      </td>
+      <td>
+        <span class="<?=$explainElement['exactMatch'] ? 'exact-match' : 'inexact-match'?>"><?=$explainElement['fieldValue']?>"</span>
```
EreMaijala: There's either a double quote missing from the beginning or an extra one at the end.

module/VuFind/src/VuFind/Search/Solr/Explanation.php
```diff
@@ -0,0 +1,531 @@
...
+     * @param float  $percentage    Percentage
+     * @param float  $fieldModifier Field Modifier
+     *
+     * @return array
+     */
+    protected function parseExplainElement($value, $description, $percentage, $fieldModifier)
+    {
+        $res = [
+            'value' => $value,
+            'percent' => $percentage,
```
EreMaijala, line 518: Initialize all required elements here so that the template doesn't try to access anything non-existent if preg_match fails.

module/VuFind/src/VuFind/Search/Solr/Explanation.php
```diff
@@ -0,0 +1,531 @@
...
+        ];
+        $pattern = '/weight\((?<fieldName>[^:]+):(?<fieldValue>\"([^\W]+\s?)+[^\\W]+\"|[^\W]+)(.+?(?= in))?/';
+        if (preg_match($pattern, $description, $matches)) {
```
EreMaijala: This is not matching what I get for a phrase search, e.g. ` weight(title:"(kyläkunta kylä) kunta väylä" in 433347) [SchemaSimilarity], result of:`. This is Solr 9.3 and default searchspecs.yaml, though not the vanilla VuFind index. Can you reproduce?

I'm also seeing `Synonym(title_short` in fieldName for `" weight(Synonym(title_full:kylä title_full:kyläkunta) in 64561) [SchemaSimilarity], result of:"`

themes/bootstrap3/js/explain.js
```diff
@@ -0,0 +1,208 @@
...
+          }
+        }
+      });
+    });
+  }
+
+
+  function _setupExplainPieChart() {
+    const pieChart = document.getElementById('explain-pie-chart');
+    if (pieChart != null) {
```
EreMaijala: Return early if !pieChart

themes/bootstrap3/templates/RecordDriver/DefaultRecord/result-list-explain.phtml
```diff
@@ -0,0 +1,24 @@
+<?php
+  $link = $this->recordLinker()->getActionUrl($this->driver, 'Explain', $this->request);
+?>
+<a target="explain" href="<?=$this->escapeHtmlAttr($link)?>"><?=$this->transEsc('Explain')?></a>
```
EreMaijala: A named target can be confusing and annoying e.g. if you want to compare records. Could we use `target="_blank"` here?

module/VuFind/src/VuFind/Search/Solr/Explanation.php
```diff
@@ -0,0 +1,531 @@
...
+     *
+     * @return array
+     */
+    protected function parseExplainElement($value, $description, $percentage, $fieldModifier)
+    {
+        $res = [
+            'value' => $value,
+            'percent' => $percentage,
+        ];
+        $pattern = '/weight\((?<fieldName>[^:]+):(?<fieldValue>\"([^\W]+\s?)+[^\\W]+\"|[^\W]+)(.+?(?= in))?/';
```
EreMaijala: Should match in UTF-8 mode `/.../u`. Otherwise this messes up non-ascii characters. This should be `kylä`:
kuva

EreMaijala (changes requested):
- I'm seeing this error in JS console in search results:
```
Uncaught TypeError: columnChart is null
    _setupExplainColumnChart http://localhost/vufind2/themes/bootstrap3
/js/explain.js?_=1699432031:137
```
- I'd prefix any class used as selector in JS with `js-`. This is not consistent in the code base at the moment, but a practice I'd recommend.
- How would you feel about making the score canvas the link to explain in results list by default and adding a description to the title attribute or something? I'm afraid the simple "Explain" link could be confusing, and "Explain the score" or such would be a bit long.
- Nit: there's a bit of inconsistency in formatting of comments. I'd ensure there's a space after the '//' everywhere. Also I believe the prevalent style in the code base is to start with a capital letter.

config/vufind/searches.ini
ThoWagen: Added the comment.

languages/ExplainFieldName/de.ini
ThoWagen: Changed the text domain to "IndexFieldDescription".

module/VuFind/src/VuFind/Search/Solr/Explanation.php
ThoWagen: Currently this doesn't make a difference. And `$rest` is optional. In my opinion null is a better default value in that case. But if that's not how we do it usually I can change this.

module/VuFind/src/VuFind/Search/Solr/Explanation.php
ThoWagen: fixed

module/VuFind/src/VuFind/Search/Solr/Explanation.php
ThoWagen: fixed

themes/bootstrap3/less/components/explain.less
ThoWagen: fixed

config/vufind/searches.ini
ThoWagen: fixed

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain.phtml
ThoWagen: fixed

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain.phtml
ThoWagen: fixed

module/VuFind/src/VuFind/Search/Solr/Explanation.php
ThoWagen: fixed

themes/bootstrap3/js/explain.js
ThoWagen: done

themes/bootstrap3/templates/RecordDriver/DefaultRecord/result-list-explain.phtml
ThoWagen: fixed

module/VuFind/src/VuFind/Search/Solr/Explanation.php
ThoWagen: fixed

module/VuFind/src/VuFind/Search/Base/Explanation.php
ThoWagen: I guess we could add Explanation.php in Search/Search2 that extends /Search/Solr/Explanations and overwrites $seachIni to "Search2". Or is there a better place for the config so it can be used for all explanations that use Solr?

module/VuFind/src/VuFind/Search/Explanation/ExplanationFactory.php
ThoWagen: I did this because it's the same for Results, Options, and Params. And this way the factory could also be used for other backends.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain.phtml
ThoWagen: done. I made the name bold.

ThoWagen:
@EreMaijala I fixed the JS error, added `js-` to the class names if necessary and fixed the comments. 

I am not sure about the link in the result list. I agree that "explain" alone might be confusing. But if we only have the chart with mouseover text then it might not be clear that you can interact with it. What do you think about this @demiankatz?

module/VuFind/src/VuFind/Search/Explanation/ExplanationFactory.php
```diff
@@ -0,0 +1,89 @@
...
+     * @throws ServiceNotCreatedException if an exception is raised when
+     * creating a service.
+     * @throws ContainerException&\Throwable if any other error occurs
+     */
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        // Replace trailing "Results" with "Params" to get the params service:
```
demiankatz: This comment should say "Explanation" instead of "Results."

themes/bootstrap3/templates/record/explain.phtml
demiankatz: It would be nice to have php-cs-fixer normalize this for us one way or the other -- maybe something to investigate when time permits.

module/VuFind/src/VuFind/Search/Base/Explanation.php
demiankatz: I haven't looked at the code in much depth yet, but this seems like a reasonable approach to me.

module/VuFind/src/VuFind/Search/Solr/Explanation.php
demiankatz: If you keep it this way, you'll at least need to change the return type of getRest to `?array`.

demiankatz (changes requested):
A couple of replies below, along with a small issue I noticed while reviewing.

Regarding the UI question, what if instead of the "Explain" text, we put an icon next to the score bar, like a little question mark? Then we could give the bar alt text/labeling to say something like "Score: N" and label the icon with "Explain score" (or something similar). What do you think?

module/VuFind/src/VuFind/Search/Solr/Explanation.php
ThoWagen: For the first case it should be enough if we allow more characters in the field value.

I'm not sure about the second problem. Do you use a synonym filter? If yes how did you configure it? Then we can try to reproduce this.

ThoWagen:
> Regarding the UI question, what if instead of the "Explain" text, we put an icon next to the score bar, like a little question mark? Then we could give the bar alt text/labeling to say something like "Score: N" and label the icon with "Explain score" (or something similar). What do you think?

That's a good idea!

module/VuFind/src/VuFind/Search/Explanation/ExplanationFactory.php
ThoWagen: fixed

module/VuFind/src/VuFind/Search/Solr/Explanation.php
ThoWagen: I will do this if @EreMaijala doesn't have any objections.

module/VuFind/src/VuFind/Search/Base/Explanation.php
ThoWagen: Enabled explain for search2.

ThoWagen:
I implemented @demiankatz suggestion for the result list link and added support for Search2.

EreMaijala (changes requested):
Looks pretty good to me! Synonym parsing is not quite working yet, though. You can see the results with the default Solr index by searching for e.g. `II` (the first line is for `" weight(Synonym(title_short:2 title_short:ii title_short:two) in 3) [SchemaSimilarity], result of:"`):
kuva

ThoWagen:
@EreMaijala I implemented something for synonyms and it works with the default Solr index for me. Could you confirm that it works for you too?

If yes I will refactor and clean up the code. I also ran the Mink Tests with explain enabled and some of them failed. I don't think that should happen and will have a look into this next week.

module/VuFind/src/VuFind/Search/Solr/Explanation.php
EreMaijala: No objections as long as docs match reality. :)

module/VuFind/src/VuFind/Search/Solr/Explanation.php
EreMaijala: This was with VuFind's default index definition. See https://github.com/vufind-org/vufind/blob/dev/solr/vufind/biblio/conf/synonyms.txt.

EreMaijala (changes requested):
Synonyms look good to me now. I'll just mark this as "Requested changes" until you've done the minor fixes and cleanup.

ThoWagen:
I think all the requested changes are now implemented and I also fixed if the the result list if the necessary information (request and scores) is not provided. That's the case for e.g. searches in tags or collections.

config/vufind/searches.ini
```diff
@@ -889,3 +890,18 @@ maxLimit = 100
 ; Filters are defined as in [RawHiddenFilters] in order to handle more complex
 ; filter combinations.
 ;0 = "format:Video"
+
+; This section provides settings for the explain feature. When enabled the result
+; list contains links to the explanation why a title was found and how the relevance
+; is calculated.
+[Explain]
+; The explain feature is disabled by default. Uncomment the following line to enable it
+enabled = true
```
demiankatz: This line should be commented out by default to match the comment above.

demiankatz (changes requested):
I'll do a deeper dive into this code after @EreMaijala approves it, but in the meantime, I noticed one minor problem.

module/VuFind/src/VuFind/View/Helper/Root/ExplainElement.php
```diff
@@ -61,9 +61,14 @@ public function __invoke($explainElement, $decimalPlaces)
         $fieldValue = $explainElement['fieldValue'] ?? false;
         $shortLabel = '';
         if (is_array($fieldName) && is_array($fieldValue) && count($fieldName) == count($fieldValue)) {
-            $shortLabel = 'Synonym [' . implode(', ', array_map(function ($name, $value) {
-                return $name . '(' . $value . ')';
-            }, $fieldName, $fieldValue)) . ']';
+            $shortLabel = $view->transEsc('Synonym') . '['
```
EreMaijala: I don't think escaping here is the right thing. That should be something for the template. I'd prefer the translation to be done there as well, but in this case it's not worth the effort.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
```diff
@@ -3,53 +3,38 @@
...
-        )
-        : '';
-      ?>
-      ]<?=$modifier?>
-    <?php else: ?>
-      <?php
-      if (!empty($fieldName)) {
-        $fieldDescription = $this->transEsc('IndexFieldDescription::' . $fieldName);
-        $fieldDescription = $fieldName !== $fieldDescription ? ' (' . $fieldDescription . ')' : '';
+    <?php if (!is_array($fieldName)): ?>
```
EreMaijala: The indentation and not using curly brackets for pure-PHP blocks makes this a bit difficult to read. Also, could we just make it so that everything is always an array so that the template doesn't need to check these?

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
```diff
@@ -3,53 +3,38 @@
...
-        : '';
-      $exactMatchClass = ($explainElement['exactMatch'] ?? 'unkown') . '-match';
-      ?>
-      <?php if ($fieldValue): ?>
-        <span class="<?=$exactMatchClass?>">"<?=$fieldValue?>"</span>
-      <?php endif; ?>
-      <?=$this->transEsc('in')?> <span class="field-name"><?=$fieldName?></span><?=$fieldDescription?><?=$modifier?>
-    <?php endif; ?>
+      $exactMatchClass = ($exactMatch ?? 'unkown') . '-match';
+      echo '<span class="' . $exactMatchClass . '">"' . $value . '"</span>' . ' ' . $this->transEsc('in');
```
EreMaijala: `$value` needs to be escaped

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
```diff
@@ -3,53 +3,38 @@
...
+      echo ' <span class="field-name" title="' . $this->escapeHtmlAttr($description) . '">' . $name . '</span>';
```
EreMaijala: `$name` needs to be escaped

EreMaijala (changes requested):

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
ThoWagen: fixed

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
ThoWagen: fixed

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
ThoWagen: Refactored to always use arrays and changed the block in the template to pure-PHP with curly brackets.

module/VuFind/src/VuFind/View/Helper/Root/ExplainElement.php
ThoWagen: fixed

config/vufind/searches.ini
ThoWagen: fixed

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
```diff
@@ -0,0 +1,36 @@
...
+    if (count($fieldName) > 1) {
+      echo $this->transEsc('Synonym') . ': [';
+    }
+
+    for ($i = 0; $i < count($fieldName); $i++) {
+      $name = $fieldName[$i];
+      $value = $fieldValue[$i] ?? '';
+      $exactMatch = $explainElement['exactMatch'][$i] ?? '';
+      if (!empty($name)) {
+        $description = $this->transEsc('IndexFieldDescription::' . $name);
```
EreMaijala: I think this should be `translate` instead of `transEsc`. Otherwise you could end up comparing an escaped value with `$name` that's not escaped (at least in theory), but more importantly using `transEscAttr` below would double-escape the description.

EreMaijala (changes requested):
Looks pretty good to me, just one more little thing.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
ThoWagen: fixed

EreMaijala (approved):

sturkel89 (approved):
I was asked to view the explanation features in all three themes. I found that the display was consistent all three themes, in both the search results summary screen and in each page that detailed the explanation.

Everything worked as expected, as far as I can tell.

languages/en.ini
```diff
@@ -339,6 +339,7 @@ Description = "Description"
 Desired Username = "Desired Username"
 Detailed View = "Detailed View"
 Details = "Staff View"
+Difference to top result relevance = "Difference to top result relevance"
```
demiankatz: Is there a reason most of the other new strings have `explain_` prefixes but this one does not?

module/VuFind/src/VuFind/RecordDriver/SolrDefaultWithoutSearchServiceFactory.php
```diff
@@ -67,8 +67,15 @@ public function __invoke(
         if (!empty($options)) {
             throw new \Exception('Unexpected options sent to factory.');
         }
-        $config = $container->get(\VuFind\Config\PluginManager::class)
-            ->get('searches');
+
+        if (str_contains($requestedName, 'Search2')) {
```
demiankatz: What is the purpose of this special case? This feels like logic that is not in the most appropriate place. Could we instead create a custom Search2Factory to accomplish the same goal in a more targeted way?

demiankatz (changes requested):
Thanks, everyone, for keeping this moving forward! I haven't finished a full review of this yet myself (running short on time before the holidays), but I was able to look through about half of the files here, and I had a couple of minor questions. See below...

languages/en.ini
ThoWagen: This translation is not used anymore. So I deleted it. But for two other translations I also added the 'explain_' prefix.

module/VuFind/src/VuFind/RecordDriver/SolrDefaultWithoutSearchServiceFactory.php
ThoWagen: I did not want to duplicate the code. But you are probably right that this is not the best way to do it. I created a Search2DefaultFactory.

module/VuFind/src/VuFind/RecordDriver/Search2DefaultFactory.php
```diff
@@ -0,0 +1,77 @@
...
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options sent to factory.');
+        }
+        $config = $container->get(\VuFind\Config\PluginManager::class)
+            ->get('Search2');
```
demiankatz: Perhaps a clearer and less redundant way to achieve this effect would be to add a `protected $searchIni = 'searches';` property to SolrDefaultWithoutSearchServiceFactory and then use that in place of the hard-coded string. Then this class could extend SolrDefaultFactory and simply override `protected $searchIni = 'Search2';` without changing any other code.

module/VuFind/src/VuFind/Search/Base/Explanation.php
```diff
@@ -0,0 +1,148 @@
...
+     *
+     * @return string
+     */
+    public function getLookfor()
+    {
+        return $this->lookfor;
+    }
+
+    /**
+     * Get the record id of title the explanation is build for.
```
demiankatz: Typo: build should be built.

module/VuFind/src/VuFind/Search/Solr/Explanation.php
```diff
@@ -0,0 +1,562 @@
...
+
+    /**
+     * Describes the rest. It has restValue and the percentage from total value.
+     *
+     * @var ?array
+     */
+    protected $rest = null;
+
+    /**
+     * Contains the fields that where removed from the main explanation.
```
demiankatz: Typo: where should be were.

module/VuFind/src/VuFind/Search/Solr/Explanation.php
```diff
@@ -0,0 +1,562 @@
...
+     *
+     * @return float
+     */
+    public function getBaseScore()
+    {
+        return $this->baseScore;
+    }
+
+    /**
+     * Get value of the boost used is Solr query.
```
demiankatz: Typo: should be "used in"

module/VuFind/src/VuFind/Search/Solr/Explanation.php
```diff
@@ -0,0 +1,562 @@
...
+            $maxChild = null;
+            foreach ($this->getChildLines($lines, $curLevel) as $child) {
+                if ($this->parseLine($child)['value'] > $maxValue) {
+                    $maxValue = $this->parseLine($child)['value'];
+                    $maxChild = $child;
+                }
+            }
+        }
+
+        // summery of lower children
```
demiankatz: Typo: should be summary

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
```diff
@@ -0,0 +1,36 @@
...
+      $name = $fieldName[$i];
+      $value = $fieldValue[$i] ?? '';
+      $exactMatch = $explainElement['exactMatch'][$i] ?? '';
+      if (!empty($name)) {
+        $description = $this->translate('IndexFieldDescription::' . $name);
+        $description = $name !== $description ? $description : '';
+      } else {
+        $description = '';
+      }
+      $exactMatchClass = ($exactMatch ?? 'unkown') . '-match';
```
demiankatz: Typo: unkown (and if this is really going to be displayed, should it be translated?)

demiankatz (changes requested):
Thanks for the progress on this! I managed to finish reading through all of the files, and I found a few more minor things to fix.

One other general issue: I'm concerned that in the long term, this code may be fragile if Solr changes its output format. I realize that there's no perfect way to test for this, but I wonder if some kind of Mink test might be worth developing, to at least make sure that explanation returns some kind of values and not an error message or garbage. It's obviously impossible to write a test that will reliably get the same scores all the time, but maybe something that just confirms we're getting numbers where we expect numbers, etc., would be enough to catch the most severe of errors. What do you think?

module/VuFind/src/VuFind/RecordDriver/Search2DefaultFactory.php
ThoWagen: Yes, that is much better. I changed it.

module/VuFind/src/VuFind/Search/Base/Explanation.php
ThoWagen: fixed

module/VuFind/src/VuFind/Search/Solr/Explanation.php
ThoWagen: fixed

module/VuFind/src/VuFind/Search/Solr/Explanation.php
ThoWagen: fixed

module/VuFind/src/VuFind/Search/Solr/Explanation.php
ThoWagen: fixed

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
ThoWagen: fixed. This is a class and is not displayed directly. So it does not need to be translated.

ThoWagen:
I added a Mink test class that tests that at least some results are displayed and are not empty. I agree that it probably is not possible to test the exact values.

themes/bootstrap3/templates/RecordDriver/DefaultRecord/explain-line.phtml
demiankatz: Of course! Sorry for misreading the context. :-)

demiankatz (approved):
This looks good to me, @ThoWagen.

Before I merge, just one last question: do you think it's worth investing time in adding more descriptions to the IndexFieldDescription text domain? If you think it would be helpful, I'm certainly willing to help with this. That being said, though, I wonder if this is worth having at all -- right now, since they are only used in a title attribute, it is fairly difficult to even see these descriptions or realize that they are available. Do you anticipate other uses for these translations, or is it possible that the work of maintaining and translating them is greater than the value they provide?

ThoWagen:
We want to use the descriptions but use other fields in our index. So I can't simply copy them in this PR. 

We could simply use the descriptions from here https://vufind.org/wiki/development:architecture:solr_index_schema. Or we don't include any translations is this PR at all but still include the logic so that anyone can add their own descriptions. I don't have strong feelings about this.

I can also see that it's not obvious that the description exists. We also add a small hoverable question mark here.

demiankatz:
Thanks, @ThoWagen. I agree that it might be valuable to add a hoverable question mark icon when a description is set --  maybe we can make that last change before merging this, if you have time. Beyond that, I would suggest that we just leave the translations that are currently here as examples but don't add more yet (so as not to create significant amounts of work for translators on a feature that may be rarely used). If there is interest/demand, we can always expand them in future, using the wiki page as a starting point as you suggest.

ThoWagen:
@demiankatz I added the question mark and agree with your take on the translation.

demiankatz (approved):
Thanks, @ThoWagen, this looks good to me now!

PR #3039, opened 2023-08-21
Accessibility: Moravian Library Report

crhallberg:
Applying the recommendations from [Moravian Library Accessibility Report](https://docs.google.com/document/d/1qII8IRWOk0eP3er-91C5clhTfQq92yPkUbvAtq3xdNU/edit#heading=h.kvcl9sm0nhrb) (thank you, Josef!)

## Needs Clarification

- Advanced Search
  - From and To need labels _(couldn't replicate)_
  - Search Tips to (ul) _(not sure which tips, headings might be better)_

## To Research

- [x] Top Recommendations to (ul) #3107 
- [x] Refine Search to (ul) #3108 
- [ ] Advanced Search
  - [x] Identical "Remove" links (PR) #3109 
  - [ ] From and To need labels
  - [ ] Search Tips to (ul)
- [ ] Headings
  - [ ] Alternative Searches (h3)
  - [x] Search Tools (h2) -- handling in #3070 
- [ ] Autocomplete (looking into alternatives)
- [ ] aria-expanded in collapsed facets
- [x] Pick up location label -- handling in #3076
- [ ] My Account
  - [ ] Your Account and Favorites headings
  - [ ] Add alert after changing account
  - [ ] Make individual loan checkbox labels unique
    - [ ] Likely applies to search checkboxes as well

## To Research

- [ ] H5 in Footers _(h5 seems high)_

PR #3070, opened 2023-09-05
Fix Search Tools heading and list accessibility

maccabeelevine:
From #3039 Accessibility: Moravian Library Report

### Search Tools (h2)

Details from https://docs.google.com/document/d/1qII8IRWOk0eP3er-91C5clhTfQq92yPkUbvAtq3xdNU/edit:

> LH: Now the static text “Search Tools:” should be an h2 heading and the links behind it should also be listed in an unordered list (`<ul>`).


maccabeelevine:
I changed the `<strong>` to an `<h2>` and surrounded the various links with a `<ul>`/`<li>` structure.  I added some CSS to keep the visual display as similar as possible to what it was before, since I don't think we actually want it to *look* like a heading and a vertical list of actions; although that can be easily removed if I'm wrong.

I also took the liberty of moving the separator n-dashes from HTML into css pseudo-elements, because it seemed semantically wrong now that each item is wrapped in a list item.  Hope that's ok.

Following the wiki I made the CSS changes in search.less and compiled them with `npm run build:css`.  It looks like I also need to add the compiled .scss and .css files to the PR, but I wanted to check that first.  (Update: I committed the .scss file because the build failed without it.  Asking for confirmation on the .css files.)

demiankatz:
Thanks, @maccabeelevine! I'll give this a closer look tomorrow, but your assumptions all sound correct to me. Please do commit all the generated css and scss files when you have a chance; we include those assets so people can deploy VuFind from GitHub without requiring Node.js in their environment.

maccabeelevine:
> Please do commit all the generated css and scss files when you have a chance; we include those assets so people can deploy VuFind from GitHub without requiring Node.js in their environment.

Yes that makes sense, thanks @demiankatz .  I agree with the goal and that it's worth doing.  I hate node dependencies in particular.  Just wanted to check because it's non-standard to commit compiled files.

demiankatz:
Thanks, @maccabeelevine! This looks like a good candidate for review by @sturkel89, who can double-check that everything looks reasonable in all of our themes.

@sturkel89, this PR changes the "Search Tools" list at the bottom of the search results (which includes RSS feed, email search, save search, etc.). The goal is to make the underlying markup more accessible without drastically changing the visuals. Could you just take a look at this and compare it against dev in all three themes? I also recommend testing what happens when you turn on schedule_searches in the Account section of config.ini, since this should add an additional option to the tool list.

I strongly suspect from reviewing the code that this is going to be just fine and nothing is going to look different -- but always good to get another set of eyes on things.

maccabeelevine:
> I strongly suspect from reviewing the code that this is going to be just fine and nothing is going to look different -- but always good to get another set of eyes on things.

At @demiankatz 's optimistic statement I took another look and noticed my changes had introduced some extra vertical whitespace.  Just added some css tweaks to remove that; it's not pixel-perfect but close I think (on boostrap3).  Thanks @sturkel89 for the review!

sturkel89:
I've looked over this test branch in all three themes. 

When the browser is wide enough that all of the Search Tools can fit in a single row, **test and dev are identical**. 

When I narrow the screen a bit, I like @maccabeelevine's version better than dev! He seems to have introduced a line break after the heading "Search Tools:" so that the options are listed below in a row. 

The only one that works slightly less well is Sandal, as the full complement of options scroll to three lines rather than two with the separate header.

Overall, I think the new code is an improvement! 

Examples from each theme, with browser window deliberately narrowed:

image

demiankatz (approved):
Thanks, @sturkel89, I agree that this looks good and improves upon the original. I can live with the sandal issue, as I think this is an unavoidable consequence of the more spacious font choices in that theme.

PR #3071, opened 2023-09-05
Add package-lock.json to .gitignore

maccabeelevine:
This seems like a node.js build artifact just like node_modules.  But that was added to .gitignore 7 years ago so I'm probably missing something.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ import/solrmarc.log*
 lessphp_*.list
 module/VuFind/tests/.phpunit.result.cache
 node_modules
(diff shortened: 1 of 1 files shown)


demiankatz (approved):
I have no objection to adding this; there might be an argument for adding package-lock.json to the repository if we change our approach to Javascript dependencies in the future and want to capture more exact versions of everything, but for the moment I think it makes sense to allow these dependencies to be resolved locally based on package.json without additional restrictions (since I imagine that imposing such restrictions might create problems for users with different Node.js or npm versions).

PR #3072, opened 2023-09-06
Fix record redirect to collection to not mess up encoding of record id.

EreMaijala:
Same issue as in other places where `normalize_path` is set to false.


demiankatz (approved):
Thanks for catching this, @EreMaijala!

PR #3073, opened 2023-09-06
Add a composer script to update npm dependencies.

EreMaijala:
Since we don't currently include package-lock.npm in git, this provides a convenient command for updating npm dependencies and copying them to correct places.


demiankatz (approved):

PR #3074, opened 2023-09-06
Improve coverage of list tagging test.

demiankatz:
This improves the test created in #3067 to cover more edge cases. It is currently passing in my environment, and in the work-in-progress branch that ports tag logic to Doctrine.


EreMaijala (approved):
Nice! I also wasn't aware of named tests with a data provider, so something new learned as well. :)

demiankatz:
Great, thanks for the approval. Yes, I really like the named test feature -- extremely helpful when something fails!

PR #3075, opened 2023-09-06
Simplify resultScrollerActive() code.

demiankatz:
Since I complained about this in #3068, I figured I ought to fix it everywhere else for consistency....


EreMaijala (approved):
Thanks! That was in quite a few places.

PR #3077, opened 2023-09-07
Add grid-based stack placement for combined search

maccabeelevine:
This is a CSS Grid based layout for combined search, as an alternative to the existing flex-based modes.  This has two primary advantages which I cared about:

- combined.ini says "The order of sections in this file will control the display order of search results on screen."  This is actually true only on wide screens.  On mobile the columns are collapsed so that everything in column 1 displays above everything in column 2, etc.  Grid mode actually does display things in the intended order regardless of width.
- All columns in a given row get the same height, without any JS or additional CSS.  This is a visual preference I see in many single-search implementations, and something our institution likely wants.

Of course there are other flexibilities you can get from a grid layout as well, such as more complex layouts of the individual search modules.  I would probably leave those to local customizations, unless someone has a compelling idea for how to generalize it.  But I do think that the basics of grid layout is worth including as an option in the core, for the advantages above and so it's easier to do those local customizations.

Now that all browsers support grid layout, which was obviously not the case when 'combined search' was developed, one could argue for actually replacing the existing 'distributed' mode.  But I think it's best to leave it as-is for backwards compatibility, and because some people prefer the uneven / adaptive heights of each module.


composer.json
```diff
@@ -128,7 +128,6 @@
         "post-install-cmd": "@phing-install-dependencies",
         "post-update-cmd": "@phing-install-dependencies",
         "qa": "phing qa-console -Ddefaultconfigs=true",
-        "install-npm-dependencies": "npm install && phing copynodemodules",
-        "update-npm-dependencies": "npm update && phing copynodemodules"
+        "install-npm-dependencies": "npm install && phing copynodemodules"
```
maccabeelevine: I did not make this change (intentionally), it either came from upstream or from installing node locally...but it matches what's currently in `dev` and has been for years...

demiankatz: @EreMaijala added that new line in #3073. Is it possible you accidentally undid his change when resolving merge conflicts?

maccabeelevine: Argh thank you @demiankatz .  I didn't see it in the git history for that file because I was accidentally looking at a commit-specific history.  I will revert my accidental change.

maccabeelevine: Fixed

demiankatz (changes requested):
Looks like this has some conflicts that need to be resolved (nothing hard -- just merge and rebuild the CSS). Other than that, it looks good to me, but since I'm not nearly the HTML/CSS expert that @crhallberg is, I'd be interested in hearing whether he has any suggestions before merging this.

maccabeelevine:
> Looks like this has some conflicts that need to be resolved (nothing hard -- just merge and rebuild the CSS). 

@demiankatz Yup lots of PRs lately with CSS changes.  Fixed.

> Other than that, it looks good to me, but since I'm not nearly the HTML/CSS expert that @crhallberg is, I'd be interested in hearing whether he has any suggestions before merging this.

Yes please!

demiankatz (changes requested):
@maccabeelevine, I took a look at this from a visual perspective, and I notice that distributed mode and grid mode look nearly identical, except that distributed mode has more right margin. Is this intentional? If it's easy to change, it might make sense to align these so that they look identical except for the ordering. Screen shots:

Distributed:
image

Grid:
image

maccabeelevine:
> distributed mode and grid mode look nearly identical, except that distributed mode has more right margin

@demiankatz Not intentional so much as the result of minimal CSS and I figured it was benign.  But I agree it's better to match the look of distributed by default, since I've specified what were meant to be the only differences.  The two modes should now have identical widths at all breakpoints.

themes/bootstrap3/templates/combined/stack-grid.phtml
```diff
@@ -0,0 +1,37 @@
...
+        'domId' => $currentSearch['domId'],
+      ];
+      if (isset($currentSearch['ajax'])) {
+        $colParams['ajax'] = $currentSearch['ajax'];
+      }
+    ?>
+    <?php // This empty div prevents margin collapse, as the otherwise unneeded div.combined-column does elsewhere. ?>
+    <div>
+      <div id="<?=$this->escapeHtmlAttr($colParams['domId'])?>" class="combined-list">
+        <?php $templateSuffix = (isset($colParams['ajax']) && $colParams['ajax']) ? 'ajax' : 'list'; ?>
```
demiankatz: `isset($colParams['ajax']) && $colParams['ajax']` could be simplified to `$colParams['ajax'] ?? false` for improved readability. (php-cs-fixer is supposed to spot and correct this kind of scenario, but it doesn't always work).

themes/bootstrap3/templates/combined/stack-grid.phtml
```diff
@@ -0,0 +1,37 @@
+<div class="combined-search-container grid"
+  style="grid-template-columns: repeat(<?=$this->escapeHtmlAttr($columns)?>, 1fr)"
+>
+  <?php $columnedParams = []; ?>
+  <?php $columnIndex = 0; ?>
+  <?php foreach ($this->combinedResults as $searchClassId => $currentSearch): ?>
+    <?php
+      if (
+        (!isset($currentSearch['ajax']) || !$currentSearch['ajax'])                    // AJAX column
```
demiankatz: These checks can be simplified with null coalescing.

themes/bootstrap3/templates/combined/stack-grid.phtml
```diff
@@ -0,0 +1,37 @@
+<div class="combined-search-container grid"
+  style="grid-template-columns: repeat(<?=$this->escapeHtmlAttr($columns)?>, 1fr)"
+>
+  <?php $columnedParams = []; ?>
```
demiankatz: This variable is never used.

themes/bootstrap3/templates/combined/stack-grid.phtml
```diff
@@ -0,0 +1,37 @@
...
+  <?php $columnIndex = 0; ?>
```
demiankatz, line 4: This value is never incremented. Could that mean that `supportsCartOptions` is not being correctly applied below?

demiankatz (changes requested):
@maccabeelevine, I took a closer look at your new template and compared it against existing ones, which led me to notice some potential optimizations and possible bugs.

The potential optimizations were in code that was copied from other existing templates, so I opened https://github.com/vufind-org/vufind/pull/3118 to optimize everything.

maccabeelevine:
> The potential optimizations were in code that was copied from other existing templates,

@demiankatz Yes, exactly.  I realized I was increasing the duplication among these stack-x templates but figured any refactoring should be done in a separate PR, as you're doing now.

themes/bootstrap3/templates/combined/stack-grid.phtml
maccabeelevine: Addressed by mirroring #3118

themes/bootstrap3/templates/combined/stack-grid.phtml
maccabeelevine: Thanks, removed.

themes/bootstrap3/templates/combined/stack-grid.phtml
maccabeelevine: Yes you are probably right; I added back the incrementor to match stack-distributed.  The variable name is misleading, $columnIndex is not about columns, it's about the individual search result modules.

So possibly fixed now, but I'm reading through code and wiki and not seeing how to actually enable the cart to confirm it.  Still looking

maccabeelevine: > Still looking

Ok I found `showBulkOptions` and enabled it.  I see checkboxes next to the modules that are supposed to have them, so I think it's ok now.

demiankatz: Thanks for tracking that down. Yes, it looks good now!

demiankatz (approved):
Thanks, @maccabeelevine -- this should be ready to merge, but I'll leave it open in case @crhallberg has a chance to give it a look. If you don't hear from him by Thursday, feel free to merge.

PR #3078, opened 2023-09-08
Add an icon set for unicode symbols and use it in hierarchical facets.

EreMaijala:
As promised in #3047, here's a follow-up to introduce plain Unicode symbols as a set in the icon helper. Also switches the facet tree icons to use the same symbols as before #3047. 

TODO
- [x] Merge #3084 before this and make sure escaping here is consistent with standards there.
- [x] Add test coverage


themes/root/templates/Helpers/icons/unicode.phtml
```diff
@@ -0,0 +1 @@
+<span class="icon icon--font icon--unicode" role="img" aria-hidden="true">&#x<?=$this->icon?>;</span>
```
demiankatz: Is it worth escaping the icon, just in case somebody puts something unexpected in the configuration?

demiankatz (changes requested):
Thanks, @EreMaijala! One very minor technical suggestion, which doesn't need to delay @sturkel89 from testing this from a visual perspective.

sturkel89:
I took a look at this new branch in each theme, with and without OR and EXCLUDE turned on. Everything seems to work - no issues, and I definitely like these better than the smaller arrows. Thanks, @EreMaijala!

themes/root/templates/Helpers/icons/unicode.phtml
EreMaijala: Escaping added. Note that there's no escaping in font.phtml either..

themes/root/templates/Helpers/icons/unicode.phtml
```diff
@@ -0,0 +1 @@
+<span class="icon icon--font icon--unicode" role="img" aria-hidden="true">&#x<?=$this->transEsc($this->icon)?>;</span>
```
demiankatz: I was thinking `escapeHtml` rather than `transEsc` -- was there a reason you chose the translated version?

demiankatz (changes requested):

themes/root/templates/Helpers/icons/unicode.phtml
demiankatz: Also, all the other icon templates offer a mechanism for passing through additional classes and attributes. Did you intentionally avoid aligning with the other icon templates in those regards?

demiankatz:
I've opened #3084 to straighten out the escaping. As it happens, the icon value was being escaped in the helper code, so I was wrong to suggest adding the escaping here. However, after #3084 is merged, the escaping will be moved to the templates, since there are some situations where escaping the icon value early will break template logic. So I think we should just merge #3084 first, then finish cleaning this up. Working on #3084 also reminded me that the icon helper has a pretty thorough test class, and we should add support for this new scenario there as well.

themes/root/templates/Helpers/icons/unicode.phtml
EreMaijala: Oops and oops. Somewhere along the road I realized we don't need the `::before` trick to display the icon, and then I lost all attribute support. Should be fixed now.

themes/root/templates/Helpers/icons/unicode.phtml
```diff
@@ -0,0 +1,5 @@
+<?php
+    $this->headLink()->appendStylesheet($this->src);
+    $className = trim('icon icon--font icon--unicode ' . ($this->extra['class'] ?? ''));
+?>
+<span class="<?=$className?>"<?=$this->attrs ?> role="img" aria-hidden="true">&#x<?=$this->escapeHtml($this->icon)?>;</span>
```
demiankatz: As per #3084, we should probably escape `$className` here for consistency.

demiankatz (changes requested):
I think we're nearly there -- just one more detail needed to fully match up with where we landed in #3084.

themes/root/templates/Helpers/icons/unicode.phtml
EreMaijala: Should be all sorted now.

demiankatz (approved):
Looking good to me -- thanks, @EreMaijala!

PR #3079, opened 2023-09-08
Add 'combined' indicator to backend result-list

maccabeelevine:
In the context of a combined / multi-column search (Combined Controller) it is useful to be able to abbreviate the display of an item's search result, since
- it has a narrower space to fit
- it's ideal to keep the total height of each result equal across backends, which doesn't matter in a single-backend results screen

I couldn't find any state present in templates/RecordDriver/*/result-list.phtml that indicated whether the list was being rendered in the context of a combined search.  This addresses that, so that the result-list can use `if ($combined ?? false)` etc. logic.

If it was already possible and I missed it, that would be great.


demiankatz (changes requested):
@maccabeelevine, a couple of thoughts on this:

1.) It's possible that you might be able to use `$this->layout()->templateDir === 'combined'` to detect this scenario without making changes to core code.

2.) I can't remember all the details without reviewing the code, but I suspect that your solution, and my proposed alternative, will not work correctly if combined with AJAX-loaded columns. Not sure if that's a problem for your use case, but if we want to create a general solution, we should make sure that it works correctly for both AJAX and non-AJAX situations.

maccabeelevine:
@demiankatz Yes I definitely want this to work in both AJAX=true and =false modes.  As you suggested the `templateDir` solution does not work in AJAX mode; that variable is undefined.  

But the 'combined' flag solution I'm proposing here does work in both modes.  I think ultimately the AJAX call ends up triggering the same templates.

themes/bootstrap3/templates/combined/results-list.phtml
```diff
@@ -74,6 +74,7 @@
       'params' => $params,
       'showCheckboxes' => $this->showCartControls || $this->showBulkOptions,
       'excludeResultIds' => true, // Don't add id attributes to results to avoid duplicates
+      'combined' => true,
```
demiankatz: Maybe add a comment explaining why this is here, since it's not used in core code -- e.g. `// Flag that we're in combined mode; may be useful to customize template behavior`

demiankatz (changes requested):
Thanks, @maccabeelevine -- clearly my memory of this code was a bit too hazy, and my suggestions were not helpful. I see no harm in merging this as-is, but one small suggestion first...

themes/bootstrap3/templates/combined/results-list.phtml
maccabeelevine: Good call.  Done.

demiankatz (approved):

themes/bootstrap3/templates/combined/results-list.phtml
demiankatz: Thanks!

PR #3081, opened 2023-09-08
[VUFIND-1637] Do not override explicit searchClassId w/search memory.

demiankatz:
This is a possible fix to [VUFIND-1637](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1637), but I'm not sure if it will significantly reduce the benefits of the work introduced in #2701. We should discuss further before merging.

TODO
- [x] Run full test suite
- [x] Resolve [VUFIND-1637](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1637) when merging.


demiankatz:
@EreMaijala, I've revised this based on our Slack conversation. It does appear that we're only explicitly overriding searchClassId in the case of the Author module, so it seemed sensible to make the override more explicit.

EreMaijala (approved):
Looks good to me! Mink tests and manual tests are passing.

PR #3082, opened 2023-09-08
Remove unnecessary IE polyfill.

demiankatz:
Happened to spot this while working on other code -- pretty sure we don't need to support IE <9 anymore.


maccabeelevine (approved):
Die IE, die.

PR #3083, opened 2023-09-08
Display jump links above combined search results

maccabeelevine:
The idea here is to optionally provide a sort of table-of-contents above combined search results, each item a jump link within the page to the bento box below.

Inspiration examples from USC:
usc

and Cornell discovery layers:
cornell


themes/bootstrap3/templates/combined/results.phtml
```diff
@@ -60,6 +60,8 @@
       <?php endforeach; ?>
     </div>
   <?php endif; ?>
+  <?php // TODO check config file to enable ?>
```
maccabeelevine: TODO

themes/bootstrap3/templates/combined/jump-links.phtml
```diff
@@ -0,0 +1,10 @@
+<div class="combined-jump-links">
+  <span>Jump to Results:</span>
```
maccabeelevine: TODO: translation

maccabeelevine:
Here is a first take on the idea for boostrap3.  Some styling advice would be welcome.  It probably needs at least some surrounding box similar to the facet and search result summaries above.

Screenshot 2023-09-08 171035

maccabeelevine:
Updated screenshot, I added record counts to each jump link, similar to one of the examples.

Screenshot 2023-09-11 135210

themes/bootstrap3/templates/combined/results.phtml
```diff
@@ -60,6 +60,9 @@
       <?php endforeach; ?>
     </div>
   <?php endif; ?>
+  <?php if ($config['Layout']['jump_links'] ?? false): ?>
+    <?=$this->context($this)->renderInContext('combined/jump-links.phtml', ['combinedResults' => $this->combinedResults])?>
```
demiankatz: Do you need to use renderInContext here, or would a plain render work? (The renderInContext helper is a weird VuFind-specific thing that I'm hoping to someday eliminate... if regular render works, which I think it should in this case, it is preferable).

demiankatz (changes requested):
I'm short on time so haven't given this a very deep review yet, but I did notice one thing that might be easy to improve.

Also, in line with my comments on an earlier PR, have you tested that this works with AJAX-loaded content? I see you do have some Javascript here, but it wasn't clear to me at a glance if that would handle the AJAX situation or if it was for a different purpose.

I'd also be interested to hear @crhallberg's thoughts on the aesthetic side of things. I might have opinions of my own when I have time to try this hands-on as well, but that may be a little while, and I thought I might as well share my initial thoughts in the meantime.

maccabeelevine:
> Also, in line with my comments on an earlier PR, have you tested that this works with AJAX-loaded content? I see you do have some Javascript here, but it wasn't clear to me at a glance if that would handle the AJAX situation or if it was for a different purpose.

@demiankatz If I understand the question, yes it does work with both ajax=true and ajax=false modules defined in combined.ini (and a mix of both).

> I'd also be interested to hear @crhallberg's thoughts on the aesthetic side of things. I might have opinions of my own when I have time to try this hands-on as well, but that may be a little while, and I thought I might as well share my initial thoughts in the meantime.

Yes please @crhallberg and others.  I didn't spend any time really on aesthetics yet because I don't know how "far" to go with it.  I.e. I think the two non-VuFind examples I provided look great, but neither would be consistent with bootstrap3's general design.  Suggestions please.

themes/bootstrap3/templates/combined/results.phtml
maccabeelevine: @demiankatz Hmm.  Just tried this, a plan render() failed due to an undefined $columns param used by stack-??.phtml.  I fixed that by adding it to the $viewParams array above, but now I'm not clear if there are any other context variables that might be expected in certain conditions in any of the stack-* files.  I think it's safer to leave as is but your call.

demiankatz: That happened after you changed `<?=$this->context($this)->renderInContext('combined/jump-links.phtml', ...` to `<?=$this->render('combined/jump-links.phtml', ...`? I wouldn't expect that, since as far as I can tell, jump-links.phtml only refers to one variable and doesn't include any sub-templates. But either I'm missing something, or we're not talking about the same code change. I wasn't suggesting changing any other renderInContext calls, just the newly-added jump-links one.

maccabeelevine: Argh you are correct of course, I changed the wrong one.  I hate days I have to wake up at 5am :)

demiankatz: Ha, I know exactly how you feel -- I was also up pretty early today, and I'm dragging at this point. :-)

maccabeelevine:
I noticed yesterday this doesn't work right for sections with a filter.  Working on that this morning.

maccabeelevine:
> I noticed yesterday this doesn't work right for sections with a filter. Working on that this morning.

Fixed.

themes/bootstrap3/templates/combined/jump-links.phtml
```diff
@@ -0,0 +1,10 @@
+<div class="combined-jump-links">
+  <span><?=$this->translate('combined_jump_links_intro')?></span>
+  <ul>
+    <?php foreach ($combinedResults as $section) : ?>
+      <li class="<?= $section['domId'] ?>">
```
demiankatz: The 'domId' value needs to be run through `$this->escapeHtmlAttr` and the 'label' needs to be run through `$this->transEsc` to display things correctly.

themes/bootstrap3/templates/combined/results-list.phtml
```diff
@@ -15,6 +15,18 @@
     ? $this->url($advancedSearchAction, [], ['query' => ['edit' => $results->getSearchId()]])
     : false;
 ?>
+<?php
+    $classId = $domId ?? 'combined_' . $searchClassId;
+    $countsJs = <<<JS
+        $(document).ready(function() {
```
demiankatz: Would it be easy to do this without jQuery? Since we're trying to move away from jQuery, it might be better to avoid it here so we don't need to port code later... but if it's a lot of work to convert now, I don't insist on the change. :-)

demiankatz (changes requested):
Thanks, @maccabeelevine -- I finally had time to look more closely at this, and I spotted one significant bug and one possible optional improvement.

Additionally, I notice that the jump links have no left margin, at least in the bootprint3 theme; that might be worth adjusting too:

image

Since this adds a new language string, we're really brushing up against the translation file freeze for release 9.1. If it's important to you to get this included in that release, maybe we can merge the new language string independently of the rest of the code so that it can be included in the translation process while giving us more time to work through the remaining issues here. What do you think?

maccabeelevine:
> Since this adds a new language string, we're really brushing up against the translation file freeze for release 9.1. If it's important to you to get this included in that release, maybe we can merge the new language string independently of the rest of the code so that it can be included in the translation process while giving us more time to work through the remaining issues here. What do you think?

Good idea.  I've pulled out the translation as #3119.

maccabeelevine:
> Additionally, I notice that the jump links have no left margin, at least in the bootprint3 theme; that might be worth adjusting too

@demiankatz Yes that's true, I hadn't really tried to make it look good initially:

>I didn't spend any time really on aesthetics yet because I don't know how "far" to go with it. I.e. I think the two non-VuFind examples I provided look great, but neither would be consistent with bootstrap3's general design. Suggestions please.

But I should have at least done something clean to start.  I've given it some side margins.  I also stopped it from wrapping within a single jump link.  Current look: 

image

themes/bootstrap3/templates/combined/jump-links.phtml
maccabeelevine: Thank you for the catch.  Fixed.

themes/bootstrap3/templates/combined/results-list.phtml
maccabeelevine: Ok, I rewrote the code to avoid jQuery.  I used several design patterns that are possibly better to extract into a separate JS module, but that would introduce race conditions and other complexity, so not doing that for now.

Used as guidance:

- https://youmightnotneedjquery.com/#ready
- https://youmightnotneedjquery.com/#create_elements

There is another part of the ready() logic needed to support IE <= 8, but I'm glad to see we don't support that.

demiankatz (approved):
Looks good to me! As with #3077, I recommend leaving this open another couple of days in case @crhallberg has anything to add before leaving on vacation. :-)

crhallberg (dismissed):
There's an AJAX bug holding this PR back, where we need to add this to the viewParams on line 125 of the CombinedController.

```php
'domId' => 'combined_' . str_replace(':', '____', $sectionId),
```

demiankatz:
Thanks for catching that, @crhallberg -- I thought @maccabeelevine and I had both tested the scenario that caused this bug (an ajax-loaded filtered column), but it appears that we missed that edge. I've confirmed that merging your fix corrects the underlying problem (which was causing the count of the filtered column to appear next to the unfiltered column instead of its proper place), so I'm going to dismiss your review as completed and merge this now.

maccabeelevine:
Thanks @crhallberg for catching that and the fix!

PR #3086, opened 2023-09-12
Widen Lightbox

crhallberg:
Allow the Lightbox to fit content from its current maximum of 600px to a dynamic range between 600px and 1200px.


demiankatz:
Thanks, @crhallberg. I haven't had a chance to look at this in detail yet, but I'm curious about the form changes -- requesting a review from @EreMaijala in case he has specific concerns or suggestions related to these changes.

module/VuFind/tests/unit-tests/src/VuFindTest/Form/FormTest.php
```diff
@@ -367,22 +367,25 @@ public function testSenderFieldMergingWithSettings()
                     'name' => 'name',
                     'group' => '__sender__',
                     'label' => 'Sender Name',
-                    'settings' => ['size' => 100],
+                    'settings' => [
+                        'maxlength' => 50, // from Form.php defaults
+                        'size' => 100, // from feedbackforms/test.yaml
```
EreMaijala, line 365: I know this is just a test, but why have a 100 characters wide field if you only can input 50? Can we revise to like maxlength=100 and size=50?

EreMaijala (changes requested):
I don't think that changing size directly to maxlength is the correct action for most fields. I'm fine with having a max length, but 50 is a bit on the low side. 

Do we still want to keep the default width of 50 for text areas (does it even have any effect with the other settings controlling the field width?)?

There's a problem with missing right margin on narrow screens:

dev branch:
image

this branch:
image

module/VuFind/tests/unit-tests/src/VuFindTest/Form/FormTest.php
crhallberg: This is all a result of a misunderstanding on my end on how `size` works. Would you agree that size and maxlength can be eliminated in favor of CSS?

EreMaijala: At least when feedback is stored in the database we should enforce maximum length somehow (unlike we now do), so setting maxlength to what the database supports is one sensible option. But that's not necessarily the scope of this PR, so I'd be fine with leaving both out here.

demiankatz (commented):
I did some testing and found some potentially problematic results...

1.) In some cases, the lightbox is now smaller than it used to be (for example, turn on tab_enabled under [Feedback] in config.ini and compare the old code with this code -- the new lightbox is narrower). Is this intentional, or should we raise the min-width?

2.) For very wide lightboxes, there are problems with the close button either getting hidden off-screen or else pushing the left side of the lightbox out of the visible window. This happens before the responsive event where the close button moves inside the lightbox. I haven't tested all themes, but I'm seeing this behavior in bootprint3.

I'm interested in hearing @EreMaijala's opinion on whether the form size removal is okay, but I think we'll need to at least resolve issue 2 above before we can merge this.

demiankatz (approved):
Thanks for the latest updates, @crhallberg -- this addresses all of the problems I encountered. The full test suite is passing. I'll just leave this open until @EreMaijala has a chance to comment (he's welcome to hit merge if he wants, once he has looked at it).

EreMaijala (approved):
Looks good to me now.

The thing that I don't really like is how the lighbox actually gets narrower when switching to the the external close button around the 768px breakpoint, but that was the case before as well, so not an issue here.

PR #3087, opened 2023-09-13
Fix issues with too long sets of work keys in queries and URLs

EreMaijala:


module/VuFind/src/VuFind/Search/Base/Params.php
```diff
@@ -1912,16 +1913,31 @@ public function getOverrideQuery()
     /**
      * Return search query object.
      *
-     * @return \VuFindSearch\Query\AbstractQuery
+     * @return AbstractQuery
      */
-    public function getQuery()
+    public function getQuery(): AbstractQuery
```
demiankatz: Is there a strong reason to add the return type here? It obvious is a good idea in the long term, but I'm concerned it might be a BC break if any classes are extending this method. Maybe better to omit for now and do in a separate dev-10.0 PR.

module/VuFind/src/VuFind/Search/QueryAdapter.php
```diff
@@ -106,6 +107,11 @@ public static function display(AbstractQuery $query, $translate, $showName)
             return $query->getString();
         }
 
+        // Work keys query:
+        if ($query instanceof WorkKeysQuery) {
+            return $query->getId();
```
demiankatz: Is it an issue that this might return null? Do we need some kind of default fallback (even if it's just an empty string)?

module/VuFind/src/VuFind/View/Helper/Root/RecordLinker.php
```diff
@@ -304,7 +305,13 @@ public function getVersionsSearchUrl($driver)
         ];
 
         $urlHelper = $this->getView()->plugin('url');
-        return $urlHelper($route, [], ['query' => $urlParams]);
+        $url = $urlHelper($route, [], ['query' => $urlParams]);
+        if (strlen($url) > 2048) {
```
demiankatz, line 309: Should we use mb_strlen? (Probably doesn't make much difference one way or the other in this instance... but I'm trying to be in the habit of using mb_strlen when appropriate).

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Backend.php
```diff
@@ -428,12 +432,11 @@ public function workExpressions($id, $workKeys, ParamBag $defaultParams = null)
...
         $this->injectResponseWriter($params);
-        $query = [];
+        $terms = [];
         foreach ($workKeys as $key) {
-            $key = addcslashes($key, '+-&|!(){}[]^"~*?:\\/');
-            $query[] = "work_keys_str_mv:(\"$key\")";
+            $terms[] = addcslashes($key, ',+-&|!(){}[]^"~*?:\\/');
         }
-        $params->set('q', implode(' OR ', $query));
+        $params->set('q', '{!terms f=work_keys_str_mv}' . implode(',', $terms));
```
demiankatz: Can we guarantee that no term will ever contain a comma?

demiankatz (changes requested):
I haven't tried this hands-on yet, but it looks logical to me. I've left a handful of minor comments.

This might also be a good time to review the Mink RecordVersionsTest and make sure it's covering everything that it should, if you have a moment... want to be sure we can reliably use that as a before-and-after to be confident that the refactoring hasn't obviously changed/broken anything.

module/VuFind/src/VuFind/Search/Base/Params.php
EreMaijala: Oops, right, reverted.

module/VuFind/src/VuFind/Search/QueryAdapter.php
EreMaijala: Yes, fallback added.

module/VuFind/src/VuFind/View/Helper/Root/RecordLinker.php
EreMaijala: I don't think the URL can ever contain multibyte characters. Even international domain names should be encoded as IDNA that's ascii safe.

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Backend.php
EreMaijala: No. I thought escaping the delimiter above would be enough, but now that I checked it, that doesn't work. I'll think about an alternative.

module/VuFind/src/VuFind/View/Helper/Root/RecordLinker.php
demiankatz: Okay, in that case, I'll withdraw my suggestion. :-)

module/VuFindSearch/src/VuFindSearch/Backend/Solr/Backend.php
EreMaijala: I changed the delimiter to chr(1F) and removed the useless escaping (verified manually with Solr that all the characters can be in a term as is).

demiankatz (approved):
This is looking good to me now, and all tests are passing. Thanks for the fix, @EreMaijala!

PR #3091, opened 2023-09-14
[VUFIND-1638] Improve flexibility of autocomplete autosubmit.

demiankatz:
This PR addresses [VUFIND-1638](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1638) and adds autocomplete autosubmit support to the course reserves search.

TODO
- [x] Simplify Javascript if possible

### DIFF
--- a/themes/bootstrap3/js/common.js
+++ b/themes/bootstrap3/js/common.js
@@ -591,7 +591,12 @@ function setupAutocomplete() {
           ? event.detail
           : event.detail.value;
         input.value = value;
-        $("#searchForm").trigger("submit");
+        const $form = $searchbox.closest("form");
+        // Try to submit the form:
+        $form.trigger('submit');
+        // If form submit failed (which currently happens for course reserves),
+        // click the input button instead:
+        $form.find('input[type="submit"]').trigger('click');
       });
     }
   });
--- a/themes/bootstrap3/js/common.js
+++ b/themes/bootstrap3/js/common.js
@@ -591,12 +591,7 @@ function setupAutocomplete() {
           ? event.detail
           : event.detail.value;
         input.value = value;
-        const $form = $searchbox.closest("form");
-        // Try to submit the form:
-        $form.trigger('submit');
-        // If form submit failed (which currently happens for course reserves),
-        // click the input button instead:
-        $form.find('input[type="submit"]').trigger('click');
+        input.form.submit();
       });
     }
   });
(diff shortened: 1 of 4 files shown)


demiankatz:
As of this writing, the implementation meets the need but seems overly complicated. I can't figure out why the course reserves form doesn't submit when I trigger a submit event, but does submit when I click the submit button inside the form. Right now, the Javascript code is doing the same thing twice to cover two different situations, which is obviously not ideal. I also didn't attempt to de-jQuery-ize anything, but maybe doing that will solve some part of the problem.

If you want to set this up to try it for yourself, follow these steps:

1.) Spin up a standard test instance with `phing startup`
2.) In config.ini, change the ILS `driver` setting to `Demo` in the `[Catalog]` section and set `search_enabled` to `true` in the `[Reserves]` section
3.) At the command line, run `php $VUFIND_HOME/util/index_reserves.php` to get some sample data into the index
4.) Go to the /Search/Reserves path in your running instance to test.

themes/bootstrap3/js/common.js
```diff
@@ -591,7 +591,12 @@ function setupAutocomplete() {
           ? event.detail
           : event.detail.value;
         input.value = value;
-        $("#searchForm").trigger("submit");
+        const $form = $searchbox.closest("form");
+        // Try to submit the form:
+        $form.trigger('submit');
+        // If form submit failed (which currently happens for course reserves),
+        // click the input button instead:
+        $form.find('input[type="submit"]').trigger('click');
```
crhallberg, line 599: I think the issue here is that the form has multiple submit forms. We might be able to fix this issue by making each of the select+button pairs its own form. It's not invalid HTML but I imagine the different submits is what's holding up the JS.

EreMaijala (approved):

demiankatz:
Mystery solved: the inconsistency was caused by a button named "submit" inside the form interfering with the form's native "submit" function. Removing the name from the button solved everything.

themes/bootstrap3/js/common.js
```diff
@@ -591,7 +591,7 @@ function setupAutocomplete() {
           ? event.detail
           : event.detail.value;
         input.value = value;
-        $("#searchForm").trigger("submit");
+        input.form.submit();
```
crhallberg, line 594: Well I'm learning new things every day!

crhallberg (approved):

demiankatz:
All tests are passing -- merging now. Thanks for the troubleshooting help, everyone! :-)

PR #3093, opened 2023-09-14
Align Reset Filters Button

crhallberg:
Align button to the top and use flex to improve the look and spacing.

Fixes https://openlibraryfoundation.atlassian.net/browse/VUFIND-1636.


sturkel89:
Good news! The Clear Filters button is now behaving properly, and looks great in all three themes!! Thanks for this fix, @crhallberg!
image

-----------------------------------
The nitpicky news is that I found two formatting things that could stand to be improved related to the display of applied filters at the top of the screen:

image

This demonstrates that if you have multiple very long entries for a single filter, they scroll to another line, which is indented in an unsatisfactory way (half a space?). I think the bumped-down entry should either be flush with the left margin along with the other filter names, or indented by several spaces. (This looks the same in dev, so it's not an issue with your branch.) NB: it seems to me that once you have four filters in the same category, they're grouped into a single expandable dropdown.

-----------------------------

This leads to my second issue. The expandable dropdown looks different in the three themes, two of which would benefit from tweaking:

**Bootstrap3**: 
The button is very tall; it would look better with less top and bottom padding.
image

**Bootprint3**: 
Looks good!

image

**Sandal**: 
The button seems to be bottom-aligned with the line, and the text within the button is bottom aligned within the button. I think it would look better centered vertically and with the text centered within the button.

image

demiankatz (approved):
Thanks, @sturkel89 and @crhallberg -- merging now.

@crhallberg, can you look into fixing the other issues Susan reported in a follow-up PR?

PR #3094, opened 2023-09-15
FOLIO: Display bound-with titles

maccabeelevine:
"Bound-with" titles are a situation where multiple independent works / bibliographic titles are bound together into the same physical item.  This breaks the most common ILS convention where a single work can contain multiple holdings/items but not vice-versa.  In a bound-with, a single item physically contains multiple works.  It is sometimes even a many-to-many relationship, where a single bibliographic title is bound into multiple bound-with items, each of which also contains many other works.

In VuFind, a patron looking at a bibliographic record page, on the holdings tab, needs to be able to see if the record is bound-with other titles, and what those titles are.  A bound-with item often has no printed table of contents, and may have dozens of works bound together, sometimes with no relationship to each other and no obviously inherent order.  Yes this is truly annoying. 

The original FOLIO data model did not support this, as each single item can only be linked to one holdingsRecord, and from there to one instance.  A parallel bound-with-parts API was developed and implemented over the last few years.  Each bound-with-part links a holdingsRecord to an item.  See the orange lines in [this diagram](https://docs.google.com/drawings/d/15Od4OcnKYj6R88UcrME8HdK-MFrjUE8HM1nxiQBWsaA/edit).  Support for this data model has been built over the last year into the FOLIO Inventory app such that you can view, delete and create bound-with relationships.

This PR uses the FOLIO bound-with data model to add bound-with titles to the holdings tab.

### Future improvements
- [ ] Optimize getBoundWithRecords to bulk query for holdings records and instances, batched for URL length.

### TODO
- [x] Update ILS driver spec when merging
- [x] Update changelog (FOLIO BC breaks) when merging


maccabeelevine:
UX developed with input from Lehigh librarians, and meeting with U Chicago bound-with experts for a second perspective this afternoon.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -783,10 +819,13 @@ public function getHolding($bibId, array $patron = null, array $options = [])
...
                 $this->getPagedResults(
                     'items',
-                    '/item-storage/items',
+                    '/inventory/items-by-holdings-id',
                     $query
                 ) as $item
             ) {
+                if ($item->discoverySuppress) {
+                    continue;
+                }
```
maccabeelevine: `/inventory/items-by-holdings-id` is an optimized call to retrieve both directly linked and bound-with items at once.  They are distinguished in the result set by the isBoundWith attribute.  The downside is this API does not support passing the discoverySuppress parameter so that filter is done to the results intead.

themes/bootstrap3/templates/RecordTab/holdingsils.phtml
```diff
@@ -123,12 +123,12 @@
         try {
             echo $this->context($this)->renderInContext(
                 'RecordTab/holdingsils/' . $this->tab->getTemplate() . '.phtml',
-                ['holding' => $row]
+                ['holding' => $row, 'callNos' => $callNos]
```
maccabeelevine, line 126: There has to be a better way to do this, but two levels deep in `boundwith.phtml` I want to format this item differently based on whether there are multiple call numbers being displayed.  In `holdingsils.phtml` the `$holding` variable can provide this info, but that name is being overridden on this line by the specific row, so it's no longer accessible.

themes/bootstrap3/templates/RecordTab/holdingsils/boundwith.phtml
```diff
@@ -0,0 +1,18 @@
...
+            $callNumberCount < 2 ? 'bound_with_heading_one_callnumber' : 'bound_with_heading_multiple_callnumbers', 
+            ['%%callnumber%%' => $this->escapeHtml($callNumber)])
+        ?>
+    </h4>
+    <p><?=$this->translate('bound_with_intro', ['%%title%%' => $this->escapeHtml($this->truncate($driver->getTitle(), 50))])?></p>
+    <ol>
+        <?php foreach ($boundWithRecords as $record): ?>
+            <li class="<?=($record['bibId']==$driver->getUniqueId()) ? 'direct-item' : 'bound-with-item'?>">
+                <a href="<?=$this->escapeHtmlAttr($record['bibId'])?>">
+                    <?=$this->escapeHtml($this->truncate($record['title'], 45)) ?>
```
maccabeelevine: This truncation is only necessary because the list of titles is displayed in the second column of this items table, taking up half its width, and approx 1/3 of total width including the sidebar column.  

- I would prefer it to be full-width on the table, but couldn't think of a way to do that while preserving the `th` / `td` structure for accessibility.  
    - If I simply put this info in a separate `<tr>` following the main one for this item, it messes up the alternating color CSS.  
    - I could alternately just adjust the column withs to something like 25%/75%, but the table has fixed layout and it might look odd for it to be relative, especially if there are other item tables on the same record (i.e. for other locations) that continue to have fixed width.  
    - I can't just display the list outside of the table itself, because it's semantically relevant to the specific item -- not to the bib/instance record.  
- I also suspect individual institutions will want to change that specific number.  I could move that to a config file but don't see a good design pattern for using config (beyond the base config.ini) within templates.  Alternately I could do the truncation earlier in `Folio.ini`.  But again I'd prefer to just have a wider column.

Ideas welcome.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -737,6 +739,41 @@ function ($a, $b) use ($sortField) {
...
+                'GET',
+                '/holdings-storage/holdings/' . $boundWithHoldingId
+            );
+            $holding = json_decode($response->getBody());
+            // Get the bound-with holding's instance record.
+            $instance = $this->getInstanceById($holding->instanceId);
+            $boundWithRecords[] = [
+                'title' => $instance->title,
+                'bibId' => $this->getBibId($instance),
+            ];
```
maccabeelevine, line 778: If you think it's important, I could try to figure out how to contain each $instance in a RecordDriver, in case a template wants something else out of the instance.  Or I suppose just put the instance itself in the array.  Neither seemed necessary but I'm not the best judge.

maccabeelevine:
First example screenshot, an bib record with two items, the second being a bound-with.  In the list of contents, the bolded item is this bib record.  
image

themes/bootstrap3/templates/RecordTab/holdingsils/boundwith.phtml
crhallberg: I would be in favor of changing the table layout away from a fixed 50/50%. It might be best to give these tables a class like `holdings-table` since the fixed layout is coming from the Bootstrap framework.

crhallberg: As far as truncating goes, make sure to test it on the sandal theme for landing on a number. It has the largest font and padding. 

EDIT: it occurred to me that having the full title would actually be a good option. The list will format it clearly. If we still want to truncate, a title or making this an abbr element would give the user an option to see the full title.

crhallberg (commented):
Everything looks accessibly marked up and clearly presented.

themes/bootstrap3/templates/RecordTab/holdingsils/boundwith.phtml
maccabeelevine: @crhallberg I've submitted a separate PR #3096 to adjust the columns on all holdings tab tables to 75/25 as we discussed in the zoom on Friday.  I've changed the truncation here to a level that seems to work fine (with a bit of margin to spare) even on Sandal.  Where it wraps on mobile widths it still doesn't seem to exceed 2 lines each.

themes/bootstrap3/templates/RecordTab/holdingsils/boundwith.phtml
```diff
@@ -0,0 +1,21 @@
+<div class="bound-with-records">
+    <h4>
+        <?=$this->transEsc(
+    $callNumberCount < 2 ? 'bound_with_heading_one_callnumber' : 'bound_with_heading_multiple_callnumbers',
+    ['%%callnumber%%' => $this->escapeHtml($callNumber)]
+)
```
maccabeelevine: I really don't like the way this is (not) indented, but it's what php-cs-fixer insists on.

maccabeelevine:
> UX developed with input from Lehigh librarians, and meeting with U Chicago bound-with experts for a second perspective this afternoon.

The UX feedback from U Chicago folks and @crhallberg was generally positive.  I made a few changes based on the conversation:
- The bound_with_intro paragraph is now optional (not even an empty `<p>` will display if it's left out.
- It turns out that the order of bound-with titles is not necessarily the actual bound order (even though it is for Lehigh).  I changed the `<ol>` to a `<ul>` and changed the intro text to match.
- The group agreed on the general layout but preferred changing all holdings tab tables to a wider first column.  I did that in #3096 since it has much wider scope than bound-with and may need more input.

I think the UX is now reasonably good, so ready for full review.

themes/bootstrap3/templates/RecordTab/holdingsils/boundwith.phtml
demiankatz: Sometimes it makes bad decisions in templates. You might be able to get better results by leaving the `<?=` on its own line, then indenting the content on the next line, then closing with `?>` below that.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
demiankatz: We don't typically load record drivers inside ILS drivers, so I think it's fine to just export the IDs. If we need record drivers, we should probably fetch them "on the outside" anyway.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -645,13 +645,14 @@ protected function getHoldingDetailsForItem($holding): array
...
-     * @param array  $holdingDetails Holding details produced by
-     * getHoldingDetailsForItem()
-     * @param object $item           FOLIO item record (decoded from JSON)
-     * @param int    $number         The current item number (position within
-     * current holdings record)
-     * @param string $dueDateValue   The due date to display to the user
+     * @param string $bibId            Current bibliographic ID
+     * @param array  $holdingDetails   Holding details produced by
+     *                                 getHoldingDetailsForItem()
+     * @param object $item             FOLIO item record (decoded from JSON)
```
demiankatz, line 656: It looks to me like you're retrieving $item from a different API than the previous code used, which is necessitating some additional changes (like the way effective location ID is retrieved below). Should this comment reflect exactly where the item came from? Also, it seems like we'll need a changelog note about this, because this could be a breaking change if people have customized formatHoldingItem, and the values they rely on are no longer available or formatted differently. (I didn't dig in to see how many potential differences there are -- but it seems like a possible issue).

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -783,10 +820,13 @@ public function getHolding($bibId, array $patron = null, array $options = [])
             foreach (
                 $this->getPagedResults(
                     'items',
-                    '/item-storage/items',
+                    '/inventory/items-by-holdings-id',
                     $query
                 ) as $item
             ) {
+                if ($item->discoverySuppress ?? false) {
```
demiankatz, line 832: Is it no longer possible to pre-filter because of the API change? Is this a possible performance concern?

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -710,6 +712,7 @@ protected function formatHoldingItem(
             'location_code' => $locationCode,
             'reserve' => 'TODO',
             'addLink' => true,
+            'bound_with_records' => $boundWithRecords,
```
demiankatz, line 720: It would probably be a good idea to add bound_with_records support to the Demo driver as well, so people without FOLIO can demonstrate and experiment with this feature. I'll also add a TODO checkbox to update the driver spec when we merge this.

themes/bootstrap3/templates/RecordTab/holdingsils.phtml
demiankatz: This seems like a case of bad variable naming. It would make more sense if 'holding' were called 'item' in this context, and then you could also pass in the true holdings record as 'holding'. It's unfortunate that this wasn't noticed much earlier! That might be a knot worth untangling eventually, but I think under the circumstances, just passing down the call numbers isn't a terrible solution. Since work was already done to process and deduplicate them, it seems better to provide them to the template instead of recalculating them, even if the subtemplate did have access to full holdings.

themes/bootstrap3/templates/RecordTab/holdingsils/boundwith.phtml
demiankatz: I see that my earlier comment got a bit garbled -- that'll teach me to write something in a rush from my phone. What I was trying to suggest was something like this:

```
        <?=
            $this->transEsc(
                $callNumberCount < 2
                  ? 'bound_with_heading_one_callnumber'
                  : 'bound_with_heading_multiple_callnumbers',
                ['%%callnumber%%' => $this->escapeHtml($callNumber)]
            )
        ?>
```

I think with this basic arrangement, php-cs-fixer might tweak the indentation a little bit, but it shouldn't completely de-indent anything in ugly ways.

themes/bootstrap3/templates/RecordTab/holdingsils/standard.phtml
```diff
@@ -78,6 +78,19 @@
             </ul>
           </div>
         <?php endif; ?>
+        <?php if (!empty($holding['bound_with_records'])): ?>
```
demiankatz, line 81: Once things are finalized, we should remember to add this to holdingsils/extended.phtml as well to keep the two versions in sync with each other.

demiankatz (changes requested):
Thanks, @maccabeelevine -- this looks pretty reasonable to me. I've left some feedback and replies to your comments below.

I gather from @crhallberg that there is some desire to get this into release 9.1 if possible. We're rapidly closing in on the freeze on new translation strings, however, so I'm not sure how realistic that is. Since it looks like there are some potential BC breaks in the FOLIO driver changes, I don't want to rush this without proper consideration of the implications.

If you do want to increase the likelihood of this getting into 9.1, one approach might be to implement the Demo driver version of the functionality first, instead of FOLIO. Then we could merge the UI changes (including new language strings) along with the Demo functionality first before the freeze, and spend a little more time discussing the FOLIO implementation details.

If you think it's okay to wait until 10.0, though, there's no need to make things more complicated -- just brainstorming options.

maccabeelevine:
> If you think it's okay to wait until 10.0, though, there's no need to make things more complicated -- just brainstorming options.

@demiankatz I have no problem with this going into 10.0.  Actually I'm less concerned with the reasons you mentioned than that this effectively relies on #3096 to not look horrible, and that PR may need more time for discussion (despite @crhallberg and I agreeing) since it's a noticeable visual change.

demiankatz:
Thanks, @maccabeelevine -- I've given this a 10.0 milestone so it won't slip through the cracks, and I'll try to give #3096 some priority in the meantime.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
maccabeelevine: Yes, unfortunately the `/inventory/items-by-holdings-id` API does not support the discoverySuppress query parameter.  I can submit that as a FOLIO Jira if it's a concern.  

The post-filtering itself is not a performance concern, I think, because in practice I'm not sure how often there will be an unsuppressed instance record, with an unsuppressed holdings record, with a significant number of suppressed item records.  Even if that were the case, we immediately ignore any suppressed records after the API call.  Eliminating the filter in the call itself might actually speed things up.

That said, the switch from an item-storage API to the higher-level inventory API does have a negative performance impact.  In my very ad-hoc testing it adds an average of maybe 100ms to the call, ranging from almost zero difference to ~200ms at worst.  So then yeah, if there are 5 holdings records on the instance, that could in theory add a full second to the retrieval.  Again I am not sure if this is a practical concern, and there are probably other advantages to having the full (not storage-only) item record.  But if needed I could add a config param loadFullItem and allow the code to work both ways for the (probably large) majority of institutions that won't use bound-with.

themes/bootstrap3/templates/RecordTab/holdingsils/boundwith.phtml
maccabeelevine: Yes that worked!  Much cleaner.

themes/bootstrap3/templates/RecordTab/holdingsils/standard.phtml
maccabeelevine: Done.  I didn't realize extended.phtml was a thing!  Nice how it's full-width there so no two-column problems.

demiankatz:
@maccabeelevine, do you want to sort out the conflicts here now that #3096 is merged?

module/VuFind/src/VuFind/ILS/Driver/Folio.php
maccabeelevine: To be clear do you want me to do Demo driver work as part of this PR?  Or is that a later thing?

demiankatz: If it won't take a lot of time/effort, it might be worth doing now so we can use it during the review process to quickly test edge cases, etc. But if it's going to be a problem, it's not essential!

maccabeelevine: Ok sure I hadn't worked with the Demo driver before, didn't realize it was mostly random data.  Will build this.

maccabeelevine: Done.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -865,6 +865,21 @@ public function getHolding($id, array $patron = null, array $options = [])
             $seriesIssue = $issue % 4;
             $issue = $issue + 1;
             $status[$i]['enumchron'] = "volume $volume, issue $seriesIssue";
+            if (rand(0, 1)) {
```
maccabeelevine: I figured not every fake holding should have bound-with titles.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -865,6 +865,21 @@ public function getHolding($id, array $patron = null, array $options = [])
...
+                $status[$i]['bound_with_records'] = [];
+                $boundWithCount = 3;
+                for ($j = 0; $j < $boundWithCount; $j++) {
+                    $randomRecord = array_combine(['bibId', 'title'], $this->getRandomBibIdAndTitle());
+                    $status[$i]['bound_with_records'][] = $randomRecord;
+                }
+                $boundWithIndex = rand(0, $boundWithCount + 1);
+                array_splice($status[$i]['bound_with_records'], $boundWithIndex, 0, [
+                    [
+                        'title' => 'The Title on This Page',
```
maccabeelevine, line 878: This should in theory actually display the title from the current record page, but that wasn't currently available without a search and didn't seem worth the query.

themes/bootstrap3/templates/RecordTab/holdingsils/boundwith.phtml
```diff
@@ -0,0 +1,22 @@
+<div class="bound-with-records">
+    <h4>
```
demiankatz: This file seems to have gotten double-indented somehow (we use a 2-space indent in templates, and this has 4-space indents).

module/VuFind/src/VuFind/ILS/Driver/Demo.php
demiankatz: The 50% chance seemed a bit overwhelming, so I added a Demo.ini config setting and made it 25% by default. Now it can be turned globally on or off, or adjusted for other situations as needed.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
demiankatz: I can live with this. :-)

demiankatz (changes requested):
Thanks, @maccabeelevine, I think this is nearly done! Just one minor formatting complaint, plus I pushed up a small Demo driver improvement that I'd welcome your input on if you have opinions.

themes/bootstrap3/templates/RecordTab/holdingsils/boundwith.phtml
maccabeelevine: Fixed!

maccabeelevine:
> Thanks, @maccabeelevine, I think this is nearly done! Just one minor formatting complaint, plus I pushed up a small Demo driver improvement that I'd welcome your input on if you have opinions.

I like the percentage configuration; hadn't considered that.

demiankatz (approved):
Looks good to me! Thanks, @maccabeelevine!

demiankatz:
@maccabeelevine, I've updated the [driver spec](https://vufind.org/wiki/development:plugins:ils_drivers#getholding) (see bottom of table). Please feel free to adjust if you can improve upon this!

maccabeelevine:
> @maccabeelevine, I've updated the [driver spec](https://vufind.org/wiki/development:plugins:ils_drivers#getholding) (see bottom of table). Please feel free to adjust if you can improve upon this!

I think your description is good -- this is such a cataloging niche, anyone who doesn't know what the phrase means won't actually care. :)

PR #3095, opened 2023-09-15
Improve Author module search box handling.

demiankatz:
This is a follow-up to #3081 which solves the same problem in a more comprehensive way. It eliminates the newly-introduced overrideSearchClassId in favor of an override driven by the Search Options class. This makes the override "stickier" so that it works in more edge cases (e.g. when author results are saved to the search history as a result of next_prev_navigation being turned on).

TODO
- [x] Run full test suite


EreMaijala (changes requested):
Looks otherwise good, but I'm seeing this warning on the search-home page:

`Warning: Undefined variable $overrideSearchClassId in /Users/ermaijal/src/vufind-test/themes/bootstrap3/templates/search/searchbox.phtml on line 25`

demiankatz:
Whoops, there was a very silly typo there that was causing the override to be applied 100% of the time (which didn't break anything, but was inefficient). Should be fixed now. Can you check again?

EreMaijala (approved):

PR #3097, opened 2023-09-18
Sierra: Fix PHP error when an item doesn't have a barcode.

EreMaijala:
Apparently Sierra doesn't return the barcode field if an item doesn't have a barcode.


demiankatz (approved):
Thanks, @EreMaijala!

PR #3098, opened 2023-09-18
Add option to enable facet filters for advanced search facets

LuomaJuha:
Seems like the facet filters are not enabled in the advanced search screen, so decided to introduce an optional setting to enable them in advanced search menu. Forgot to add the description when opening the pull request.

TODO
- [ ] Note BC breaks in changelog when merging (signature changes in \VuFind\AjaxHandler\GetSideFacets)


module/VuFind/src/VuFind/Controller/AbstractSolrSearch.php
```diff
@@ -146,11 +146,9 @@ protected function processAdvancedFacets(
...
     ) {
         // Process the facets
-        $facetHelper = null;
-        if (!empty($hierarchicalFacets)) {
-            $facetHelper = $this->serviceLocator
-                ->get(\VuFind\Search\Solr\HierarchicalFacetHelper::class);
-        }
+        $facetHelper = $this->serviceLocator
+            ->get(\VuFind\Search\Solr\HierarchicalFacetHelper::class);
+        $facetConfig = $this->getConfig('facets');
```
demiankatz: I don't think we should hard-code the 'facets' configuration here. This controller can be shared by multiple Solr-based searches, and not all of them use the same configuration files. It would probably be better if we could leverage the appropriate Options object to figure out this behavior -- i.e. add a new getter there. (It should be possible to retrieve a generic Options object based on $this->searchClassId if a specific instance is not passed in via $searchObject).

demiankatz (changes requested):
@LuomaJuha, I think this PR actually brings attention to an existing problem in the code base. I suspect that the FacetFilters / ExcludeFilters settings are National Library of Finland customizations that have been partially included in the main code base by accident. I say this because as far as I can tell, they are ONLY referenced in the GetSideFacets AJAX handler, and they are not mentioned in any of our configuration files. This means that, if configured, they're only going to work when SideFacetsDeferred is turned on.

I also notice that the GetSideFacetsFactory has the facets.ini configuration hard-coded into it, which will cause problems if using SideFacetsDeferred on non-default Solr backends.

This also raises another question: if this is an incompletely-implemented partial feature in the core code, do we need a separate configuration setting to turn it on for advanced search, or should it just apply consistently across the board if enabled?

Perhaps @EreMaijala can shed some additional light on the context here....

module/VuFind/src/VuFind/Controller/AbstractSolrSearch.php
LuomaJuha: oh yea, makes sense. Hmm have to readjust and if Ere has something to say about this implementation.

EreMaijala:
@demiankatz @LuomaJuha Oops! This functionality was not meant to be there as only partially supported. I must have accidentally left it in there while porting our local changes. I'd say that it's useful functionality, though, and I'd propose completing the implementation to correct the situation. If you agree, and @LuomaJuha has the means to do that, please go ahead. Otherwise I can clean things up.

demiankatz:
Thanks, @EreMaijala! I'm certainly open to a full implementation if it's possible, though without configuration and comments, I confess that I'm not entirely sure I understand what this is for (I could likely figure it out by studying the code more closely, but I haven't put in that time yet). I'll put this on the 10.0 milestone on the assumption that we're most likely to decide to go forward with full implementation, but we're unlikely to get the work done in time for 9.1. If we decide that it's not worth doing after all, though, we could have a 9.1 PR to simply remove the partially-implemented functionality.

LuomaJuha:
@demiankatz @EreMaijala I'll tune this pr and put it as a draft for now and return later when the appropriate changes have been made :)

config/vufind/facets.ini
```diff
@@ -293,6 +293,9 @@ translated_facets[] = callnumber-first:CallNumberFirst
 ; limits must be separated using the same character set here.
 ;limitDelimiter = "::"
 
+; Optional setting to enable FacetFilters and ExcludeFilters for advanced search facets.
+;enable_filters[] = true
```
demiankatz: Why is this an array?

config/vufind/facets.ini
```diff
@@ -358,3 +361,15 @@ custom_filter_field = "vufind"
 ; combination with checkbox facets if you want to exclude values by default
 ; and provide a checkbox that allows them to be included.
 ;inverted_filters[include_books] = '-format:"Book"'
+
+; Exclude filters can be specified to exclude certain filters from showing in hierarchical
+; facets.
+;[ExcludeFilters]
```
demiankatz: Do these settings really apply only to hierarchical facets? If so, would it make sense to name the sections accordingly for greater clarity?

demiankatz (changes requested):
Thanks, @LuomaJuha! I haven't had a chance to look into this in great detail yet, but I had a couple of questions based on reviewing the configuration settings, and I thought it might be best to get those answered before I look at the code, so I can better ensure that everything is consistent and as intended. :-)

config/vufind/facets.ini
LuomaJuha: Yea, tuned them!

config/vufind/facets.ini
LuomaJuha: ops, was left by a mistake

module/VuFind/src/VuFind/Controller/AbstractSolrSearch.php
```diff
@@ -170,6 +182,18 @@ protected function processAdvancedFacets(
                 $list['list'] = $facetHelper->flattenFacetHierarchy($tmpList);
             }
 
+            if (
```
demiankatz: I see that you're now loading $facetHelper unconditionally above, instead of only using it when hierarchical facets exist. Could we move this if statement inside the other if statement that checks for hierarchical facets, and make the loading conditional again, to avoid doing unnecessary processing in situations where hierarchical facets are not used?

config/vufind/facets.ini
```diff
@@ -293,6 +293,10 @@ translated_facets[] = callnumber-first:CallNumberFirst
 ; limits must be separated using the same character set here.
 ;limitDelimiter = "::"
 
+; Optional setting to enable HierarchicalFacetFilters 
+; and HierarchicalExcludeFilters for advanced search facets.
+;enable_filters = true
```
demiankatz: Should we rename this to enable_hierarchical_filters for consistency?

module/VuFind/src/VuFind/Controller/AbstractSolrSearch.php
```diff
@@ -89,6 +89,18 @@ public function advancedAction()
         return $view;
     }
 
+    /**
+     * Get proper options file for search class
+     *
+     * @return \VuFind\Search\Base\Options
+     */
+    public function getOptionsForClass(): \VuFind\Search\Base\Options
```
demiankatz: We're doing exactly the same thing [here](https://github.com/vufind-org/vufind/blob/9e56d78621677947fa961d6a2547de17d6f4350c/module/VuFind/src/VuFind/Controller/AbstractSearch.php#L97) -- maybe it's worth moving this method up to AbstractSearch and using it there so it can be more easily shared across more controllers.

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -320,6 +320,28 @@ abstract class Options implements TranslatorAwareInterface
      */
     protected $firstlastNavigation = false;
 
+    /**
+     * Should HierarchicalFacetFilters and HierarchicalExcludeFilters
+     * apply in advanced search
+     *
+     * @var bool
+     */
+    protected $filterFacetsInAdvanced = false;
```
demiankatz: Maybe this needs "Hierarchical" added as well.

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -320,6 +320,28 @@ abstract class Options implements TranslatorAwareInterface
...
+
+    /**
+     * Excluded filters
+     *
+     * @var array
+     */
+    protected $excludeFilters = [];
```
demiankatz: Again, probably would benefit from more specific naming if we're not going to use this for non-hierarchical filters too.

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -320,6 +320,28 @@ abstract class Options implements TranslatorAwareInterface
...
+
+    /**
+     * Facet filters
+     *
+     * @var array
+     */
+    protected $facetFilters = [];
```
demiankatz: Same here.

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -1169,4 +1198,34 @@ public function limitOrderOverride($limit)
         $limitConf = $limits ? $limits->get($limit) : '';
         return array_map('trim', explode($delimiter, $limitConf ?? ''));
     }
+
+    /**
+     * Is FacetFilters and ExcludeFilters enabled in advanced search?
+     *
+     * @return bool
+     */
+    public function getFilterFacetsInAdvanced(): bool
```
demiankatz: Add Hierarchical to name?

module/VuFind/src/VuFind/Search/Base/Results.php
```diff
@@ -881,6 +881,18 @@ protected function buildFacetList(array $facetList, array $filter = null): array
                     'isApplied'
                 );
             }
+            if ($hierarchical && method_exists($this->hierarchicalFacetHelper, 'filterFacets')) {
+                $excludeFilters = $this->getOptions()->getHierarchicalExcludeFilters();
```
demiankatz: Might it make sense to make $field a parameter to the getters? It could be optional (if null, return all values), but if provided it could return just the requested field, and that could make some of the subsequent code a little more concise.

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
```diff
@@ -420,6 +420,259 @@ public function testGetFacetList(): void
...
+                            ['1/Book/EBook/', 12],
+                            ['1/Book/ColoringBook/', 12],
+                        ],
+                    ],
+                ],
+            ],
+            $this->getParams(null, $config)
+        );
+
+        $results->setHierarchicalFacetHelper(new HierarchicalFacetHelper());
```
demiankatz: I didn't study this in great depth, so maybe there's a strong reason to keep these tests in this place -- but this seems more like an integration test than a unit test, since it's testing the HierarchicalFacetHelper through the Solr Results object. Would it be better to do this testing in the [HierarchicalFacetHelper test](https://github.com/vufind-org/vufind/blob/dev/module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/HierarchicalFacetHelperTest.php) and test here with a simple mock to confirm that the Results object is passing along the values as expected?

If this is a major ordeal to refactor, please don't feel obligated to change, but if it's not too hard, it might keep our test suite a little better organized.

demiankatz (changes requested):
Thanks for the ongoing work on this, @LuomaJuha! I haven't tried hands-on testing yet, but here are a few thoughts based on the latest changing (mostly to do with naming and organization).

module/VuFind/tests/unit-tests/src/VuFindTest/Search/Solr/ResultsTest.php
LuomaJuha: sounds good! I was also a littlebit lost as to where I should put this test but im going to try and adjust.

config/vufind/facets.ini
```diff
@@ -358,3 +362,15 @@ custom_filter_field = "vufind"
 ; combination with checkbox facets if you want to exclude values by default
 ; and provide a checkbox that allows them to be included.
 ;inverted_filters[include_books] = '-format:"Book"'
+
+; Exclude filters can be specified to exclude certain filters from showing in hierarchical
+; facets.
```
demiankatz: Is it worth adding a note to these two sections to explain that these filters will always be applied to search result facets, but that enable_hierarchical_filters needs to be turned on above to apply them to advanced facets as well? Maybe that's too verbose, but it might be a useful cross-reference.

module/VuFind/src/VuFind/AjaxHandler/GetFacetData.php
```diff
@@ -130,6 +132,7 @@ public function handleRequest(Params $params)
             $this->facetHelper->sortFacetList($facetList, $sort);
             $facets = $this->facetHelper
                 ->buildFacetArray($facet, $facetList, $queryHelper, false);
+            $this->facetHelper->filterFacets($facet, $facets, $results->getOptions());
```
demiankatz: You don't seem to be doing anything with the filtered facets here. Should this be `$facets = ...`? Do we need a test to cover this situation, since the problem was apparently missed during development?

demiankatz (changes requested):
Thanks, @LuomaJuha! This is making great progress, but I think I may have spotted one last bug, plus I had a small documentation suggestion (which you're free to ignore if you think it's unhelpful).

module/VuFind/src/VuFind/AjaxHandler/GetFacetData.php
LuomaJuha: Oh, yea, that could also be good. Ill try and see if I can come up with a test for this.

config/vufind/facets.ini
LuomaJuha: Sure, I could do that. I'll adjust

demiankatz (changes requested):
@LuomaJuha, thanks for the added test and other adjustments. I think we're nearly done, but unfortunately the enable_hierarchical_filters setting doesn't seem to be working for me. I set up a standard test environment, added `hierarchical_facet_str_mv = Hierarchy` to the `[Results]` and `[Advanced]` sections, and applied this filter:

```
[HierarchicalFacetFilters]
hierarchical_facet_str_mv[] = 1/level1z/level2y
```

Now, my advanced search facets are filtered regardless of how I set enable_hierarchical_filters.

I thought perhaps this was a caching-related problem, but I don't think that is the case -- even when I clear my cache, the filtering is always applied.

Can you investigate?

LuomaJuha:
@demiankatz oh, seems like when I restructured the whole function I forgot to add the check in abstract solr search... Should be fixed now.

demiankatz (approved):
Thanks for the fix, @LuomaJuha -- looks good to me now!

PR #3099, opened 2023-09-18
Add link to advanced search in combined results

maccabeelevine:
When performing a combined search, there is the option in each results list to open that backend's results separately.  This makes sense as the user may have done the combined search as a default option (or to get an overview) and now they know which specific backend's results they want.  

Similarly they may know that they want to do an advanced search on an individual backend.  On individual backend results pages there is a link to open an advanced search on the same query.  This PR presents the same option in the context of advanced search.

The idea is based on UX (and identical language) seen at USC

image

and Cornell

image


themes/bootstrap3/templates/combined/results-list.phtml
```diff
@@ -7,6 +7,12 @@
 
   // More link should use default limit, not custom limit:
   $moreUrl = $this->url($params->getOptions()->getSearchAction()) . $results->getUrlQuery()->setPage(1)->setLimit($params->getOptions()->getDefaultLimit());
+
+  // Advanced search link
+  $advancedSearchAction = $params->getOptions()->getAdvancedSearchAction();
+  $advancedSearchUrl = $advancedSearchAction
+    ? $this->url($advancedSearchAction, [], ['query' => ['edit' => $results->getSearchId()]])
+    : false;
```
maccabeelevine: This code is adapted from searchbox.phtml.  In that file there is more complicated logic for creating the URL based on $hiddenFilters if no results are present.  I don't think that's needed here but I may be wrong.

themes/bootstrap3/templates/combined/results-list.phtml
```diff
@@ -79,11 +85,25 @@
...
+  <?php if ($currentSearch['advanced_link'] ?? false): ?>
+    <?php if (!($currentSearch['more_link'] ?? false)): ?>
+      <p>
+    <?php endif; ?>
+      <a class="icon-link" href="<?=$advancedSearchUrl?>">
+        <span class="icon-link__label"><?=$this->transEsc($currentSearch['advanced_link'])?></span>
+        <?=$this->icon('more', 'icon-link__icon') ?>
+      </a>
+   </p>
+  <?php endif; ?>
```
maccabeelevine: The `more_link` is displayed twice, with different text, at the top and bottom of the results.  I haven't looked at the history of that to see if I should do the same here; can do so if helpful.

config/vufind/combined.ini
```diff
@@ -81,6 +83,7 @@ stack_placement = distributed
 label = Catalog
 ;sublabel = "library books, videos, CDs, microform, local resources"
 more_link = "More catalog results"
+;advanced_link = "advanced search"
```
demiankatz: It would probably be best to use an example that's translated in the language files (e.g. "Advanced Search" instead of "advanced search").

themes/bootstrap3/templates/combined/results-list.phtml
demiankatz: @EreMaijala might have the strongest opinion on this point.

themes/bootstrap3/templates/combined/results-list.phtml
demiankatz: That decision comes out of the original Villanova design, the intent of which was simply to provide multiple access points to the full search results to meet different user needs. The top link is meant to be an "access more options" indicator, whereas the one at the bottom is an "access more results" indicator, but of course they both lead to the same place. This has worked out reasonably well in our own usability testing for more than a decade, but it's not the result of consultation with lots of other institutions. :-)

themes/bootstrap3/templates/combined/results-list.phtml
```diff
@@ -79,11 +85,25 @@
...
-    <p>
+    <p class="more_link">
       <a class="icon-link" href="<?=$moreUrl?>">
         <span class="icon-link__label"><?=$this->transEsc($currentSearch['more_link'])?></span>
         <?=$this->icon('more', 'icon-link__icon') ?>
       </a>
     </p>
   <?php endif; ?>
+  <?php if (($currentSearch['more_link'] ?? false) && ($currentSearch['advanced_link'] ?? false)): ?>
+    <p class="advanced_link">
```
demiankatz: Maybe I'm not reading the code correctly, as there are a lot of possible code paths, but it feels to me like there may be some scenarios that lead to unclosed p tags. If you disagree, I can take a more careful read and either find an example or reassure myself. But maybe this is a sign that it would be helpful to restructure the code to make it easier to follow the various situations.

demiankatz (changes requested):
Thanks, @maccabeelevine -- getting the ball rolling with some contextual information and suggestions. :-)

themes/bootstrap3/templates/combined/results-list.phtml
maccabeelevine: @demiankatz That's fair.  I guess my question is with adding the advanced search link to the bottom, is there a need to keep the parallel idea and add it to the top as well.  From your explanation, the intended idea of the two links is different, even if the links are the same, so no reason to treat them the same.  That said, I think an advanced search link is actually closer to the "more options" concept, although in a confusing way.  But my instinct is to keep things as-is with the the advanced link only on the bottom.

maccabeelevine:
Screenshot of the initial implementation, if it helps:

image

config/vufind/combined.ini
maccabeelevine: Well I can do that, but then the pair of links would read "More catalog results or use Advanced Search" which is a bit odd capitalization.  But I get the translation advantage.  Your call.

maccabeelevine: Never mind, I changed my mind on this.  If Advanced Search is linked alone, it needs to be capitalized.  I will change the default.  Either way implementers can adjust it easily.

themes/bootstrap3/templates/combined/results-list.phtml
maccabeelevine: Yeah, I agree it was unreadable, thanks for the nudge.  I think it's clearer now.

themes/bootstrap3/templates/combined/results-list.phtml
EreMaijala: I think that search id should be enough, but this needs to be tested with SearchTabs and SearchTabsFilters.

config/vufind/combined.ini
```diff
@@ -10,6 +10,8 @@
 ; sublabel = Text to display below the header (optional)
 ; more_link = The text of the link to access more results; set to false to
 ;        disable all "more results" links for this entry.
+; advanced_link = The text of the link to load the advanced search page for
```
EreMaijala: Why does the link text need to be configurable?

themes/bootstrap3/templates/combined/results-list.phtml
demiankatz: @maccabeelevine, maybe the most future-proof thing to do would be to name the config option in such a way that if we decide to add a second configurable link at the top in the future, it doesn't become confusing... :-)

themes/bootstrap3/templates/combined/results-list.phtml
demiankatz: Much improved -- thanks!

languages/en.ini
```diff
@@ -246,6 +246,7 @@ collection_disambiguation = "Found Multiple Matching Collections"
 collection_empty = "No items to display."
 collection_view_record = "View Record"
 Collections = "Collections"
+combined_more_advanced_separator = "or use"
```
EreMaijala: I strongly recommend using a complete translation instead of stitching the result from multiple translations. Something like this would work and be much easier to translate:
`combined_more_advanced_link_html = "or use <a href="%%url%%">Advanced Search</a>"`

See e.g. `alphabrowselink_html` for existing usage.

EreMaijala (changes requested):

languages/en.ini
demiankatz: I think the idea here is that there could be just an advanced link or a more link or both... and there's a separator in between the two links only in the case where both are displayed. I think trying to do this as a single language string could get awkward, but I also agree that stringing strings together is not ideal either. I like the fact that the string is named as a separator to make its purpose clear... but then based on the initial screen shot above, it doesn't really look like a separator since it's just part of a continuous sentence.

I realize my comments are not necessarily pointing toward any particular solution... but just trying to clarify the problems. :-)

config/vufind/combined.ini
demiankatz: The more_link is configurable, so I imagine this is for consistency.

languages/en.ini
EreMaijala: In that case we may need two different translation strings. Perhaps not so elegant, but easier to manage! :)

config/vufind/combined.ini
EreMaijala: Okay. why is the more_link configurable? :)

maccabeelevine: Yes consistency was the intent.  Plus there has to be a way to enable/disable it, so there needs to be something in the .ini anyway.

maccabeelevine: > Okay. why is the more_link configurable? :)

That one for @demiankatz  :)

EreMaijala: I can accept it for consistency. The comments could reflect the fact that they can be translation keys as well, so that anyone doesn't think that it can't be translated.

demiankatz: The more_link is configurable because each column of the combined search might need a different custom message (i.e. "Show more results for Solr", "Show more results for Summon", etc.). It's possible that the advanced search is less likely to need that kind of individualization, but I don't think the flexibility hurts.

languages/en.ini
maccabeelevine: @demiankatz is correct that I did this because each of the two links may be there on its own; the separator is only used when both are.  I could create translations that accounted for all three possibilities if you all prefer that.  This seemed (slightly) simpler.

I could also simplify the whole thing by making the advanced link a sub-option of the more-link, i.e. there is no possibility of advanced link alone.  But that seemed wrong.

I can do the two separate translation strings if folks prefer that, but that's what's effectively in combined.ini since it's also an enable/disable toggle.

config/vufind/combined.ini
maccabeelevine: > The comments could reflect the fact that they can be translation keys as well, so that anyone doesn't think that it can't be translated.

Good idea, I will add that to combined.ini.

languages/en.ini
EreMaijala: Could we do without the "or use" part? That would simplify things, and since there's the icon and line break between the two links, they're not so naturally parts of a sentence. The problem is that Starting with "or use" may cause the rest to need a different inflexion or something (typical with Finnish), which makes it difficult to formulate in a way that fits both cases.

maccabeelevine: @EreMaijala Personally I am fine with that.  For my local theme I would likely add the "or use" part via css i.e. 

`p.advanced_link:before { content: 'or use'; }`

@demiankatz Do you agree?

demiankatz: Seems like an elegant solution.

maccabeelevine: Done.

image

themes/bootstrap3/templates/combined/results-list.phtml
maccabeelevine: > name the config option in such a way that if we decide to add a second configurable link at the top in the future, it doesn't become confusing

Agreed, and done.

themes/bootstrap3/templates/combined/results-list.phtml
maccabeelevine: > I think that search id should be enough, but this needs to be tested with SearchTabs and SearchTabsFilters.

Ok I tested this.  The code works fine with SearchTabs in use or not.  It also works when hidden filters are in place for the combined sections; the advanced link contains the search ID that somehow has saved the hidden filter behind the scenes.  Honestly I thought this was broken for a while before I saw that it was working -- I haven't worked with hidden filters before.  Kudos for such a cool capability!

maccabeelevine:
I think all the feedback has been addressed.

EreMaijala (approved):

themes/bootstrap3/templates/combined/results-list.phtml
```diff
@@ -7,6 +7,12 @@
 
   // More link should use default limit, not custom limit:
   $moreUrl = $this->url($params->getOptions()->getSearchAction()) . $results->getUrlQuery()->setPage(1)->setLimit($params->getOptions()->getDefaultLimit());
+
+  // Advanced search link
+  $advancedSearchAction = $params->getOptions()->getAdvancedSearchAction();
```
demiankatz: To avoid unnecessary work, what if you change this line to:

```
$advancedSearchAction = ($currentSearch['advanced_link_bottom'] ?? false)
    ? $params->getOptions()->getAdvancedSearchAction() : false;
```

and then use `if ($advancedSearchUrl)` in the template logic below?

demiankatz (changes requested):
Thanks, @maccabeelevine -- I made a tiny edit to remove double-spaces from comments in the .ini file (we generally use just one space after a period). I also have a minor optimization suggestion. Once that conversation is resolved, I think we can get this merged!

themes/bootstrap3/templates/combined/results-list.phtml
demiankatz: On further reflection, I realize that this may not be ideal since it would make it more awkward to add advanced_link_top in the future... but I prefer to avoid generating URLs and doing options lookups if we know we're not going to use them. If you prefer to leave it as-is, I won't fight you, but maybe there's an intermediate solution that's a little more efficient. Let me know what you think! :-)

maccabeelevine: I made the optimization -- I agree it's a small improvement and we can always undo it later if there's an advanced_link_top.

themes/bootstrap3/less/components/search.less
```diff
@@ -428,6 +428,9 @@ body.rtl {
   margin-left: 1rem;
   margin-right: 1rem;
   padding: 0;
+  .combined-list span.more-advanced-separator {
```
demiankatz: In reviewing your latest class changes, I was reminded of this -- does this class no longer exist? Should we just remove the style changes, or am I overlooking something?

demiankatz (changes requested):
Sorry, @maccabeelevine, I really do want to get to the bottom of this, but I think I caught one last thing to clean up.

themes/bootstrap3/less/components/search.less
maccabeelevine: No, you are right.  Fixed!

demiankatz (approved):
Thanks, @maccabeelevine -- very simple and straightforward now! Merging... :-)

PR #3100, opened 2023-09-19
Add prepend operation for custom munge

LuomaJuha:
Noticed that there is an append operation so decided to add prepend also.

### DIFF
--- a/module/VuFind/tests/fixtures/configs/1.1/searchspecs.yaml
+++ b/module/VuFind/tests/fixtures/configs/1.1/searchspecs.yaml
@@ -120,7 +120,6 @@
 # [lowercase] - Convert string to lowercase
 # [preg_replace, pattern, replacement] - Perform a regular expression replace
 #                                        using the preg_replace() PHP function
-# [prepend, text] - Prepend text to the beginning of the user's search string
 # [uppercase] - Convert string to uppercase
 #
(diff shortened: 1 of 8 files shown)


module/VuFind/tests/fixtures/configs/1.1/searchspecs.yaml
```diff
@@ -120,6 +120,7 @@
 # [lowercase] - Convert string to lowercase
 # [preg_replace, pattern, replacement] - Perform a regular expression replace
 #                                        using the preg_replace() PHP function
+# [prepend, text] - Prepend text to the beginning of the user's search string
```
demiankatz: You don't need to add this to the legacy-version-specific test fixtures -- these are historical versions of the configuration files used to test upgrade/migration, so they should not be changed.

demiankatz (changes requested):
Thanks. @LuomaJuha, this makes sense to me -- just one minor request regarding the 1.x test fixtures.

module/VuFind/tests/fixtures/configs/1.1/searchspecs.yaml
LuomaJuha: oh, okay! will remove it from them

demiankatz (approved):
Thanks, @LuomaJuha!

PR #3101, opened 2023-09-19
FOLIO: Fix grouping holdings by holdings_id

maccabeelevine:
Per [VUFIND-1641](https://openlibraryfoundation.atlassian.net/jira/software/c/projects/VUFIND/issues/VUFIND-1641)

> The default holdings_grouping defined in config.ini (and if unspecified there, in Holds.php->getHoldingsGroupKey) is `holdings_id,location`.  The `holdings_id` key is matched against keys in $copy, which ultimately come from the individual holding/item data supplied by the ILS driver in getHolding().  In the FOLIO driver, getHolding() calls formatHoldingItem() to generate this data.  formatHoldingItem() creates a key called `holding_id` (singular "holding"), which then fails to match "holdings_id" in the grouping logic.  That would be easy to fix, but the singular "holding_id" is also used in Folio.ini for something called HMACKeys which I don't understand.  ... Of course an easy workaround is to set config.ini to "holding_id,location" instead, but no one would know to do that.

This PR fixes the typo in FOLIO.php->formatHoldingItem() and in FOLIO.ini->HMACKeys.

TODO
- [ ] Update changelog when merging


demiankatz:
Thanks, @maccabeelevine -- I tried to rebase this against the dev-10.0 branch, but that branch is currently a bit behind dev and it made things confusing, so I've changed it back for now. :-)

demiankatz (approved):
I just tested, and this does not negatively impact my ability to place holds. I think this probably qualifies as a bug fix that could be merged into release 9.1, but let me know if you'd prefer for me to wait and merge this to dev-10.0 instead once conflicts there are resolved.

maccabeelevine:
> I just tested, and this does not negatively impact my ability to place holds. I think this probably qualifies as a bug fix that could be merged into release 9.1, but let me know if you'd prefer for me to wait and merge this to dev-10.0 instead once conflicts there are resolved.

@demiankatz I am fine either way, this particular issue is irrelevant to Lehigh, I just shared it on Jira so figured I should fix it.

PR #3102, opened 2023-09-19
[VUFIND-1641] Use holdings_id instead of holding_id for Alma.

demiankatz:
I _think_ this cleans up the issue reported in [VUFIND-1641](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1641) for Alma, which would let us close that ticket. However, I can't test with Alma, so I am not 100% sure I didn't break anything. Can an Alma user test this?

TODO
- [x] Close [VUFIND-1641](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1641) when merging


module/VuFind/src/VuFind/ILS/Driver/Alma.php
```diff
@@ -464,7 +465,7 @@ public function checkRequestIsValid($id, $data, $patron)
         if ('copy' === $level) {
             // Call the request-options API for the logged-in user
             $requestOptionsPath = '/bibs/' . rawurlencode($id)
-                . '/holdings/' . rawurlencode($data['holding_id']) . '/items/'
+                . '/holdings/' . rawurlencode($data['holdings_id']) . '/items/'
```
EreMaijala: I think we need `$data['holdings_id'] ?? $data['holding_id']` here to provide back-compatibility for existing Alma ini files that have holding_id in HMACKeys.

EreMaijala (changes requested):

module/VuFind/src/VuFind/ILS/Driver/Alma.php
demiankatz: Good point. I think there are actually two places where we need to do this. I've added it in both, along with comments.

EreMaijala: True, I didn't get far enough to see that. The latter fallback has wrong variable name, though.

demiankatz: Early morning copy/paste strikes again! Fortunately, continuous integration caught this, so our systems are working. :-) Fixed now!

EreMaijala (approved):

module/VuFind/src/VuFind/ILS/Driver/Alma.php
EreMaijala: Thanks, all good now!

demiankatz: Thanks!

PR #3104, opened 2023-09-20
Trim unnecessary whitespace from translation strings.

demiankatz:
Further experimentation with Lokalise reveals that there's a fair amount of unnecessary and unwanted whitespace in our language files. This PR trims strings that need trimming while retaining intentional whitespace in a few places (mainly `number_thousands_separator`).


EreMaijala (approved):
Looks good to me!

PR #3106, opened 2023-09-20
Final translation cleanup for Lokalise.

demiankatz:
I believe this PR should make the final changes needed to begin using Lokalise with VuFind. To summarize:

- For languages that use a blank space as a thousands separator, I have replaced the normal space with a FIGURE SPACE to avoid Lokalise trimming behavior. This seems appropriate for the context.
- In hy.ini, I have changed the "by" translations to empty string instead of blank space; the intention here is to eliminate an unwanted translation, which can be done equally well with an empty string.
- In es.ini, I happened to notice an error that I had previously missed, so I fixed it while I was in the area.


EreMaijala (approved):
Looks good to me!

PR #3107, opened 2023-09-20
accessibility: top recommendations to ul.

crhallberg:
Separated from #3039.


sturkel89:
I've tested this branch and compared it with dev. 

As far as I can tell, the experience on the front end is identical. Everything looks the same between test and dev in all three themes, using a combination of facet options (exclude, or, and neither). 

Good work, and ready to merge, as far as I can tell!

demiankatz (approved):
Looks good to me, and full test suite is passing.

PR #3108, opened 2023-09-20
accessibility: refine search facets to ul.

crhallberg:
Separated from #3039.


sturkel89:
I've tested this branch and compared it with dev. 

As far as I can tell, the experience on the front end is identical. Everything looks the same between test and dev in all three themes, using a combination of facet options (exclude, or, and neither). 

Good work, and ready to merge, as far as I can tell!

demiankatz (changes requested):
I've run the full test suite and it is passing, and this mostly looks good to me. However, I've noticed one small problem: when a facet has enough values to trigger a "see all" link leading to the lightbox, that link is no longer concealed behind the "more" button -- it always shows up. In dev, this is not the case. I'm guessing there may be a missing class on a special list element somewhere that's causing this issue.

demiankatz:
I've fixed the problem I noted above through commit c60e4c3ed9e25f1f7112e71d356c8014e5d9880a. I think this is good to go now, but I'd like to get one more review before merging in case I did anything wrong in my fix.

maccabeelevine (approved):
The code LGTM and I don't see any visible changes.

demiankatz (approved):
Thanks, @maccabeelevine -- merging now.

PR #3111, opened 2023-09-21
Add some missing Northern Sami translations.

EreMaijala:
@nikopartanen Could you check that this set of translations makes sense in VuFind's context. I left out translations that mention Finna by name, but I couldn't judge if the others are worded in a general way suitable for VuFind. Thanks!


nikopartanen:
Looks good, I didn't spot anything really Finna-specific now. The ones added now are generic and good messages. 

Looking to the entire file, there are quite many double spaces inside the strings which cannot be intentional. Also `eBook = "E- girji"` should to be `eBook = "E-girji"`.

EreMaijala:
@nikopartanen Thanks! I'll fix the double spaces and eBook translation.

nikopartanen (approved):
I couldn't find anything clearly specific to only Finna, so I would approve these updates.

demiankatz (approved):
Thanks, @EreMaijala and @nikopartanen!

PR #3112, opened 2023-09-21
Composed ILS Driver

ThoWagen:
This new ILS Driver can be used to combine multiple other ILS Drivers. It allows to define multiple drivers for different driver methods and for many methods support drivers can be configured that provide additional information.

Here are some possible use cases for this:
1. Using the combination of PAIA and DAIA was implemented by letting the PAIA driver inherit from the DAIA driver. With this they can be split into two separate drivers and combined by this driver. PAIA can be used as the default main driver and DAIA can be configured for the methods getHolding, getStatus and getStatuses. 
2. This also allows to use PAIA in combination with other drivers for the record status methods.
3. One could deactivate all the methods that require a catalog login by using the NoILS-Driver as default main driver and only configure another driver for the methods that don't require a catalog login.
4. One can supplement the getHolding information with additional information, like summaries, notes etc. from an external API if they are not supported by the main driver.
5. One can get additional information about a user with the getMyProfile method from another source.

However, the driver does not support MultiILS for login. So currently one can only combine one driver that provides patronLogin with drivers that don't use patron information. An exception to this is when the other drivers can use the patron information of the driver that is used for the patronLogin. But this could be implemented in the future if it is required .

This driver was heavily inspired by the Multibackend driver. Much of the code both drivers have in common was moved into the new AbstractMultiDriver class.

To understand how the driver is supposed to work it might be a good idea to first look into the unit-tests for it.


demiankatz (commented):
As the primary maintainer of MultiBackend, I imagine that this work may be of interest to @EreMaijala, so I've asked for a review from him (though I don't think there's a huge rush to complete it).

I also don't object to the name of CombinedDriver, but I wonder if something like ComposedDriver might be slightly more descriptive, since if I understand correctly, the purpose of this driver is to create a single interface by composing functionality from multiple places. Just a thought!

I'll give this a closer look myself after the 9.1 release process is wrapped up.

config/vufind/CombinedDriver.ini
```diff
@@ -0,0 +1,36 @@
...
+[Drivers]
+;driver1 = PAIA
+;driver2 = DAIA
+;driver3 = Demo
+
+; For each driver method it can be specified which drivers should be used and how they can be combined.
+; The head needs to be [<driver method name>]. main_driver specifies the driver that is primarily used.
+; The results of the main driver can be combined with other driver results. Therefore, one needs to specify a
+; merge_key. All results of the support drivers are then matched with the keys of the main driver. E.g. for the
+; method getHolding the results of multiple drivers can get combined if the match on "item_id".
```
EreMaijala: Should this be `if *they* match`?

config/vufind/CombinedDriver.ini
```diff
@@ -0,0 +1,36 @@
...
+; but it most likely wont work with the Demo driver
+;[getStatus]
+;main_driver = "driver2"
+;
+;[getStatuses]
+;main_driver = "driver2"
+;
+;[getHolding]
+;main_driver = "driver2"
+;merge_key = "item_id"
```
EreMaijala: Would it be imaginable that a different merge key would be needed for different support drivers, i.e. should it be specific to each support driver?

module/VuFind/src/VuFind/ILS/Driver/AbstractMultiDriver.php
```diff
@@ -0,0 +1,206 @@
+<?php
+
+/**
+ * Abstract Multi Driver.
+ *
+ * PHP version 8
+ *
+ * Copyright (C) Hebis Verbundzentrale 2023.
```
EreMaijala: I think you'll need to carry over the copyrights for the existing moved code.

module/VuFind/src/VuFind/ILS/Driver/CombinedDriver.php
```diff
@@ -0,0 +1,1346 @@
...
+     * @param string $username The patron barcode
+     * @param string $password The patron password
+     *
+     * @throws ILSException
+     * @return mixed           Associative array of patron info on successful login,
+     * null on unsuccessful login.
+     */
+    public function patronLogin($username, $password)
+    {
+        return $this->defaultCall('patronLogin', func_get_args());
```
EreMaijala: Default return value should be null.

module/VuFind/src/VuFind/ILS/Driver/CombinedDriver.php
```diff
@@ -0,0 +1,1346 @@
...
+    }
+
+    /**
+     * Get Default "Hold Required By" Date (as Unix timestamp) or null if unsupported
+     *
+     * @param array $patron   Patron information returned by the patronLogin method.
+     * @param array $holdInfo Contains most of the same values passed to
+     * placeHold, minus the patron data.
+     *
+     * @return int
```
EreMaijala: In Aleph driver this is `int|null`, so we should standardize on that.

module/VuFind/src/VuFind/ILS/Driver/CombinedDriver.php
```diff
@@ -0,0 +1,1346 @@
...
+     */
+    public function getHoldDefaultRequiredDate($patron, $holdInfo)
+    {
+        return $this->defaultCall('getHoldDefaultRequiredDate', func_get_args());
```
EreMaijala: Default return value can't be an array.

module/VuFind/src/VuFind/ILS/Driver/CombinedDriver.php
```diff
@@ -0,0 +1,1346 @@
...
+     * to place Holds.
+     *
+     * @param string $id      The id of the bib record
+     * @param array  $details Item details from getHoldings return array
+     *
+     * @return string         URL to ILS's OPAC's place hold screen.
+     */
+    public function getHoldLink($id, $details)
+    {
+        return $this->defaultCall('getHoldLink', func_get_args());
```
EreMaijala: Default return value should be a string.

module/VuFind/src/VuFind/ILS/Driver/CombinedDriver.php
```diff
@@ -0,0 +1,1346 @@
...
+     * an array with result details
+     *
+     * @param array $holdDetails An array of item and patron data
+     *
+     * @return mixed An array of data on the request including
+     * whether or not it was successful and a system message (if available)
+     */
+    public function placeHold($holdDetails)
+    {
+        return $this->defaultCall('placeHold', func_get_args());
```
EreMaijala: This, like other methods that mention "a system message" in @return, should return an array such as `['success' => false, 'status' => 'ils_action_unavailable']` on failure.

EreMaijala (changes requested):
This is an interesting development that provides a nice layer of flexibility to the ILS drivers! I didn't see any major issues, so I've commented on some minor ones. 

I think that the default return values will need the most work. It might also make sense to make defaultCall throw an ILS exception by default unless instructed otherwise. This would eliminate the need to specify the default return values for cases where no suitable default is not available (or, make it more difficult to return an invalid value by accident).

ThoWagen:
@demiankatz I renamed the driver to ComposedDriver.

@EreMaijala I think you are right about throwing exceptions instead of default return values. Since calling driver methods that are not supported should either not be possible or otherwise should be an error. I removed the default return values and now always an exception is thrown.

config/vufind/CombinedDriver.ini
ThoWagen: Fixed

config/vufind/CombinedDriver.ini
ThoWagen: We don't have a use case for this but I think it makes sense to have the option. I implemented it.

module/VuFind/src/VuFind/ILS/Driver/AbstractMultiDriver.php
ThoWagen: Fixed

module/VuFind/src/VuFind/ILS/Driver/CombinedDriver.php
ThoWagen: I changed this here and also in the Demo driver.

module/VuFind/src/VuFind/ILS/Driver/CombinedDriver.php
ThoWagen: Now, an exception is thrown if the method is not supported.

module/VuFind/src/VuFind/ILS/Driver/CombinedDriver.php
ThoWagen: Now, an exception is thrown if the method is not supported.

module/VuFind/src/VuFind/ILS/Driver/CombinedDriver.php
ThoWagen: Now, an exception is thrown if the method is not supported.

module/VuFind/src/VuFind/ILS/Driver/CombinedDriver.php
ThoWagen: Now, an exception is thrown if the method is not supported.

demiankatz:
Thanks, @ThoWagen -- I've renamed this PR to match the new name of the driver itself.

config/vufind/ComposedDriver.ini
```diff
@@ -0,0 +1,36 @@
...
+
+; Definition of all the used drivers in the form <DriverName> = <DriverClass>
+; The name of config file for a driver needs te be the same <DriverName>
+[Drivers]
+;driver1 = PAIA
+;driver2 = DAIA
+;driver3 = Demo
+
+; For each driver method it can be specified which drivers should be used and how they can be composed.
+; The head needs to be [<driver method name>]. main_driver specifies the driver that is primarily used.
```
EreMaijala: I just noticed this one: I think we use `section` instead of `head`, so I'd change this for consistency.

EreMaijala (changes requested):
Looks good to me! Just one more nit to pick in a comment. :)

config/vufind/ComposedDriver.ini
ThoWagen: done

EreMaijala (approved):

config/vufind/ComposedDriver.ini
```diff
@@ -0,0 +1,36 @@
...
+; Default driver for all methods
+;main_driver = "driver1"
+
+; (Optional) the path to the driver configurations relative to VuFind config
+; dir -- omit to locate those in VuFind config dir -- it also could be an
+; absolute path
+;drivers_config_path = private
+
+; Definition of all the used drivers in the form <DriverName> = <DriverClass>
+; The name of config file for a driver needs te be the same <DriverName>
```
demiankatz: Typos -- should be "to be the same as"

config/vufind/ComposedDriver.ini
```diff
@@ -0,0 +1,36 @@
...
+; For each driver method it can be specified which drivers should be used and how they can be composed.
+; The section needs to be [<driver method name>]. main_driver specifies the driver that is primarily used.
+; The results of the main driver can be composed with other driver results. Therefore, one needs to specify
+; merge_keys for the support drivers. All results of the support drivers are then matched with the keys of the main
+; driver. E.g. for the method getHolding the results of multiple drivers can get composed if they match on "item_id".
+; The array support_keys specifies which of there keys of the given support driver should be composed with the main result.
+; The format is: support_keys["<DriverName"] = "<key1>,<key2>,...,<keyN>"
+
+; Here are some examples how a configuration could look like for a combination of DAIA and PAIA
+; but it most likely wont work with the Demo driver
```
demiankatz: "examples of how" might read a little better. "wont" should be "won't". I'm not entirely sure I understand why the Demo driver won't work, or why that is mentioned here (driver2 is not set to Demo in the example above).

module/VuFind/src/VuFind/ILS/Driver/ComposedDriver.php
```diff
@@ -0,0 +1,1351 @@
...
+     */
+    protected $mainDriver;
+
+    /**
+     * Constructor
+     *
+     * @param \VuFind\Config\PluginManager $configLoader Configuration loader
+     * @param PluginManager                $dm           ILS driver manager
+     */
+    public function __construct(
```
demiankatz: Since the configLoader and driverManager are dependencies of the base class, I think you could move this constructor there. I realize that MultiBackend has a different signature, but you could still move this constructor into the base class, and then simplify the MultiBackend constructor to call parent::__construct with the appropriate parameters.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -1530,7 +1530,7 @@ function ($loc) use ($excluded) {
      * @param array $holdInfo Contains most of the same values passed to
      * placeHold, minus the patron data.
      *
-     * @return int
+     * @return int|null
```
demiankatz, line 1533: `?int` might be more concise here (though I suspect you copied this from Aleph, so if you prefer to keep it consistent with that, I have no objection).

module/VuFind/src/VuFind/ILS/Driver/ComposedDriverFactory.php
```diff
@@ -0,0 +1,78 @@
...
+ * Factory for ComposedDriver ILS driver.
+ *
+ * @category VuFind
+ * @package  ILS_Drivers
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Thomas Wagener <wagener@hebis.uni-frankfurt.de>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+class ComposedDriverFactory implements FactoryInterface
```
demiankatz: If you decide to move the constructor, you could also rename this to AbstractMultiDriverFactory, so it would be reusable if a different subclass of the abstract class was created using the same signature.

module/VuFind/src/VuFind/ILS/Driver/MultiBackend.php
```diff
@@ -1231,7 +1171,7 @@ protected function getSourceFromParams(
      *
      * @param string $source The source name of the driver to get.
      *
-     * @return mixed  On success a driver object, otherwise null.
+     * @return mixed On success a driver object, otherwise null.
```
demiankatz, line 1173: Out of scope for this PR, but maybe something @EreMaijala should consider in a follow-up -- I see that $source can be false-y based on the code below, but the comment and type here don't seem to reflect that.

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/AbstractMultiDriverTest.php
```diff
@@ -0,0 +1,422 @@
...
+    /**
+     * Get a mock ILS authenticator
+     *
+     * @param string $userSource Source id, if the authenticator should emulate a
+     * situation where a user has logged in. Set to null for the attempt to cause an
+     * exception.
+     *
+     * @return \VuFind\Auth\ILSAuthenticator
+     */
+    protected function getMockILSAuthenticator($userSource = '')
```
demiankatz: This method is only relevant in MultiBackendTest (at least for now). Is it worth moving, or do you prefer to keep it in the abstract base class?

demiankatz (changes requested):
Thanks, @ThoWagen and @EreMaijala! I gave this another look and found a few minor things to make suggestions about.

Once these are addressed, I think it's ready to merge, but I'm curious whether you think this should be included in release 9.1 or if it's safer to delay to 10.0 due to the significant refactoring of MultiBackend? I'm thinking that a 9.1 release is probably safe, since @EreMaijala is the primary user of MultiBackend and has approved this, but I'm happy to wait if either of you would prefer.

module/VuFind/src/VuFind/ILS/Driver/Demo.php
ThoWagen: I did this since @EreMaijala suggested to standardize it here: https://github.com/vufind-org/vufind/pull/3112#discussion_r1335444249

It looks like it is done like this in a lot of places. So, I don't think this this violates any code style rules. But if you prefer `?int` I can change this.

config/vufind/ComposedDriver.ini
ThoWagen: fixed

config/vufind/ComposedDriver.ini
ThoWagen: fixed. I also explained why the Demo driver won't work.

module/VuFind/src/VuFind/ILS/Driver/ComposedDriver.php
ThoWagen: done

module/VuFind/src/VuFind/ILS/Driver/ComposedDriverFactory.php
ThoWagen: done

module/VuFind/tests/unit-tests/src/VuFindTest/ILS/Driver/AbstractMultiDriverTest.php
ThoWagen: done

ThoWagen:
It is absolutely fine for me if it is included in 10.0

module/VuFind/src/VuFind/ILS/Driver/Demo.php
demiankatz: Feel free to leave it alone for now -- since it's consistent with other things, there's no need to change it. I think in the long term I'd prefer the `?` syntax instead of the `|null` syntax since it is more concise -- but fixing that should probably become a separate project, and it's best to back it up with some kind of style checker if we make that decision. All out of scope for what you're doing here. :-)

demiankatz (approved):
Thanks, @ThoWagen -- this looks good to me. I've just made one small edit to add a missing word in a comment.

If @EreMaijala wants to include this in release 9.1, he's welcome to hit the merge button. Otherwise, I'll merge this as soon as we're done with work on release 9.1.

demiankatz:
(Setting a 10.0 milestone to be sure this doesn't get forgotten if it isn't merged sooner).

module/VuFind/src/VuFind/ILS/Driver/Demo.php
EreMaijala: I didn't mean to imply that you should use `int|null`as `?int` is indeed better, though I also value keeping the docs of same function consistent. Perhaps we could also add more interfaces and use `@inheritDoc` to avoid copy-pasting and ensure that the method definitions don't diverge.

demiankatz: The problem with inheritDoc is that php codesniffer doesn't support it properly, so we'd either need to disable comment validation or ignore lots of errors. I wish this would get addressed, but I think it's a very hard problem to solve based on the way codesniffer works.

PR #3114, opened 2023-09-22
Fix search history behavior when login is disabled.

demiankatz:
When login functionality is disabled (e.g. ILS driver is set to NoILS, and NoILS.ini has hideLogin set to true), the search history page was still displaying login-related functionality. This PR corrects that bug.

### DIFF
--- a/themes/bootstrap3/templates/search/history.phtml
+++ b/themes/bootstrap3/templates/search/history.phtml
@@ -2,17 +2,25 @@
   // Set page title.
   $this->headTitle($this->translate('Search History'));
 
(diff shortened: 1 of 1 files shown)


maccabeelevine (approved):
Code looks fine to me.  I verified the change to the breadcrumb.  I'm not seeing how to enable the search-filter-toggle in the first place so couldn't test that part.

demiankatz:
Thanks, @maccabeelevine -- search-filter-toggle is a responsive thing; it only shows up on very small screens. You can test by shrinking your browser window. (I did confirm this myself).

PR #3118, opened 2023-09-22
Use null coalescing to simplify combined search templates.

demiankatz:
This applies some simplifications/optimizations that I noticed while reviewing #3077.


maccabeelevine (approved):
Looks fine, works fine

demiankatz:
Thanks, @maccabeelevine!

PR #3119, opened 2023-09-22
Add translation for combined_jump_links_intro

maccabeelevine:
This adds a translation that will be used by #3083 once that is merged.  Separated out so we can meet the translations deadline.

The context is functionality that will do an anchor-jump down the page to a particular combined results box.

Screenshot 2023-09-11 135210


demiankatz (approved):
Thanks, @maccabeelevine!

PR #3120, opened 2023-09-26
add aria-hidden to coins titles

elsenhans:
The Coins title in the result list (and other lists) is read by the NVDA screen reader for each entry. 

So that the coins title is no longer read out I added aria-label="hidden" to all Coins titles.


demiankatz (approved):
Makes sense -- thanks!

PR #3121, opened 2023-09-26
Update MLA label to 9th edition.

demiankatz:
MLA has released a 9th edition handbook. This edition does not change any rules; all changes are simply to the organization of the handbook. Thus, no code changes are needed, but this PR updates the label we put on our MLA citations to reflect the latest version.

TODO
- [x] Confirm that Arabic translation update is correct


maccabeelevine (approved):
LGTM.  The only one I have no idea is Arabic and you said that was separately approved.

PR #3124, opened 2023-09-27
SierraRest: Trim whitespace from call number and volume.

EreMaijala:
Also cleans up the call number extraction code.

Sierra may include some extra spaces in the fields, which leads to apparent duplicate call numbers being displayed.


demiankatz (approved):
Looks reasonable to me. Thanks, @EreMaijala!

PR #3125, opened 2023-09-27
fix: give ajax-loaded combined results a domId.

crhallberg:
A fix that solves the AJAX issue with #3083.


demiankatz (approved):
Good catch, @crhallberg!

PR #3126, opened 2023-09-28
Add option to return an HTMLElement from the icon helper

LuomaJuha:
Sometimes it is easier to manipulate the icon when it is an HTMLElement like when assigining it into a value, so here is a small adjustment for the icon function.

### DIFF
--- a/themes/bootstrap3/js/common.js
+++ b/themes/bootstrap3/js/common.js
@@ -119,10 +119,6 @@ var VuFind = (function VuFind() {
   };
 
   var init = function init() {
-    // Create a template element for icon function
-    if (!_elementBase) {
-      _elementBase = document.createElement('div');
-    }
     for (var i = 0; i < _submodules.length; i++) {
       if (this[_submodules[i]].init) {
         this[_submodules[i]].init();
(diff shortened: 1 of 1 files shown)


demiankatz:
Thanks, @LuomaJuha! I'd like to get @crhallberg's opinion on this, since I believe he wrote most of the original code that's impacted. He's going to be on vacation for the next week or so, but hopefully he can look at it when he returns if he can't get to it before he leaves tomorrow. :-)

themes/bootstrap3/js/common.js
```diff
@@ -154,46 +161,58 @@ var VuFind = (function VuFind() {
...
-          mod = mod.slice(0, sliceStart) +
-              " " + attr + '="' + value + '"' +
-              mod.slice(sliceEnd);
+    function addAttrs(_element, _attrs = {}) {
+      Object.keys(_attrs).forEach(key => {
+        let newAttrs = _attrs[key].split(" ");
+        const oldAttrs = _element.getAttribute(key);
+        if (oldAttrs !== null) {
+          newAttrs = [...newAttrs, ...oldAttrs.split(" ")];
+          // Remove duplicate values
```
crhallberg: You could use a Set to remove duplicate values as you compile them.

```JS
const newAttrsSet = new Set([...newAttrs, ...oldAttrs.split(" ")]);
//...
_element.setAttribute(key, Array.from(newAttrsSet).join(" "));
```

crhallberg: Are there other attributes that require space separated strings other than class? If we're going to be working on an Element directly, I'd like Element to do as much work as possible. This actually makes me think we should be overriding the attributes instead of appending them (except for class). I know this is based on my code but I've since learned [how complicated this can get](https://github.com/johnnesky/imperative-html/blob/main/src/elements-base.ts#L3).

crhallberg (approved):
My comments are nitpicks. This looks very good and I like the direction!

themes/bootstrap3/js/common.js
LuomaJuha: Hmm, I would think that class could be the only one, only if there is some custom datasets but otherwise shouldn't be, ill try and adjust

LuomaJuha: @crhallberg made some small changes where if the key is not class then just sets it instantly.

demiankatz: Thanks, @LuomaJuha! @crhallberg, do you mind taking another look at this when you have a chance?

demiankatz:
I suspect we'll be able to merge this soon, but since things are getting locked down for 9.1, it will probably have to wait for release 10.0. I'm setting a milestone to be sure we don't lose track of it.

crhallberg (approved):
Thanks for making those changes! I was thinking about using classList for this, but I think that would lead to more confusing code. Thank you!

demiankatz:
Thanks, @LuomaJuha and @crhallberg!

PR #3128, opened 2023-09-28
LibGuides and AZ: Display descriptions

maccabeelevine:
Optionally display the description field from LibGuides (research guides) and LibGuidesAZ (databases).

image

Future:
- [ ] Consider doing the same to the recommendation modules for these data sources.  Although I lean against it.


maccabeelevine:
The test failure is occurring because the fixture doesn't match what an actual LibGuides response looks like.  I'm guessing it was made with version 1 of the search widgets "API" instead of v2.  I will need to create a new fixture (and will look for the instructions on that again -- it's binary so a pain.)

Details: The fixture is returning just a series of `<a>` links separated by `<br>`s.  The current widgets return `<li>`s around each item.

module/VuFindSearch/src/VuFindSearch/Backend/LibGuides/Connector.php
```diff
@@ -233,7 +262,7 @@ protected function prepareParams(array $params)
                 'list_format' => 1,
                 'output_format' => 1,
                 'load_type' => 2,
-                'enable_description' => 0,
+                'enable_description' => 1,
```
maccabeelevine: In theory this is a breaking change, but in practice changing this value adds more data to the search widget (API) response that was previously ignored, so there is no practical impact.  The description is not actually displayed unless the config property is set, so there is no visible effect by default.

themes/bootstrap3/templates/RecordDriver/LibGuides/result-list.phtml
```diff
@@ -10,5 +10,12 @@
         </a>
       </h2>
     </div>
+    <?php foreach ($this->driver->getSummary() as $summaryItem): ?>
+      <?php if (!empty($summaryItem)): ?>
+        <div class="resultItemLine1">
```
maccabeelevine: I have seen this `resultItemLine1` class used elsewhere, even repeated in a for loop in an [EDS template](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/templates/RecordDriver/EDS/result-list.phtml#L32).  So just following an apparent pattern.  This could just as easily be a `<p>` with no class.

module/VuFindSearch/src/VuFindSearch/Backend/LibGuides/Connector.php
```diff
@@ -187,15 +196,35 @@ protected function process($data)
 
         $items = [];
 
-        // Extract titles and URLs from response:
-        $regex = '/<a href="([^"]*)"[^>]*>([^<]*)</';
-        $count = preg_match_all($regex, $data, $matches);
+        $itemRegex = '/<li>(.*?)<\/li>/';
+        $linkRegex = '/<a href="([^"]*)"[^>]*>([^<]*)</';
+        $descriptionRegex = '/<div class="s-lg-(?:widget|guide)-list-description">(.*?)<\/div>/';
```
maccabeelevine, line 201: The class name varies (widget vs. guide) depending whether you are loading research guides or databases.  So this regex works for both.  This seemed simpler than introducing more state and conditional logic.

module/VuFindSearch/src/VuFindSearch/Backend/LibGuides/Connector.php
```diff
@@ -187,15 +196,35 @@ protected function process($data)
...
+            $linkCount = preg_match_all($linkRegex, $itemMatches[1][$i], $linkMatches);
+            if ($linkCount != 1) {
+                throw new \Exception('LibGuides result item included more than one link: ' . $itemMatches[1][$i]);
+            }
+            $item = [
+                'id' => $linkMatches[1][0],    // ID = URL
+                'title' => $linkMatches[2][0],
             ];
+
+            // Extract the descrpition.
```
demiankatz: Typo: descrpition.

module/VuFindSearch/src/VuFindSearch/Backend/LibGuides/Connector.php
```diff
@@ -187,15 +196,35 @@ protected function process($data)
...
+            if ($this->displayDescription) {
+                $descriptionCount = preg_match_all($descriptionRegex, $itemMatches[1][$i], $descriptionMatches);
+                if ($descriptionCount > 1) {
+                    throw new \Exception('LibGuides result item included more than one description: ' .
```
demiankatz: Does this need to be a fatal exception? Might it be better to log a warning and just proceed without a description?

module/VuFindSearch/src/VuFindSearch/Backend/LibGuides/Connector.php
demiankatz: No objection, though it seems it would be easy to set this based on the need for descriptions if you wanted to (e.g. to save a little bit of bandwidth).

themes/bootstrap3/templates/RecordDriver/LibGuides/result-list.phtml
demiankatz: I don't really like this class very much, and I think there was some discussion recently about possibly doing away with it... but if you want to follow the precedent, it shouldn't hurt anything.

demiankatz (changes requested):
Thanks, @maccabeelevine -- a few comments. Let me know if I can of more help with the test fixture issue.

module/VuFindSearch/src/VuFindSearch/Backend/LibGuides/Connector.php
maccabeelevine: You're right and actually it's simpler just to use the first description if there are 1+, no logging needed either.

module/VuFindSearch/src/VuFindSearch/Backend/LibGuides/Connector.php
maccabeelevine: Fixed

module/VuFindSearch/src/VuFindSearch/Backend/LibGuides/Connector.php
maccabeelevine: Agreed & done.

themes/bootstrap3/templates/RecordDriver/LibGuides/result-list.phtml
```diff
@@ -10,5 +10,12 @@
...
+          <?=$this->escapeHtml($summaryItem)?>
```
demiankatz, line 16: Based on the way the summary is getting extracted from HTML, is it possible that it may already include escaped characters or markup? We should be sure there is no possibility of double-escaping anything (or letting in dangerous unsanitized HTML, for that matter).

demiankatz (commented):
Thanks for the improvements -- see below for one more thought. Other than that, let me know when it's ready to test and I can kick the tires more closely. :-)

themes/bootstrap3/templates/RecordDriver/LibGuides/result-list.phtml
maccabeelevine: Yeah that's a good point.  But I definitely see unescaped characters in the data coming in from the widget HTML, i.e. the ampersand in `<a href="https://libraryguides.lehigh.edu/electrical_engineering" target="_blank">Electrical & Computer Engineering<i class="fa fa-external-link pad-left-sm"></i></a>`.  So I think it is correct to escape here.

maccabeelevine:
@demiankatz ready to test now, thanks.  

For future reference, this is the query used to generate the new fixture.  The old one was clearly Bryn Mawr so I repeated that, but the response format has changed thoroughly.  The data of course changed as well so tests are updated.  IMHO saving the uncompressed version of the response is easier to understand when fixing tests.

themes/bootstrap3/templates/RecordDriver/LibGuides/result-list.phtml
demiankatz: Is it possible that you should `html_entity_decode` the text extracted from the widget HTML to be on the safe side? Do you have the ability to create a test description containing HTML or at least `<` / `>` characters and see how it comes out in the widget?

maccabeelevine: I went into the LibGuides admin UI and added HTML to bold a word in a database description.  When pulled into VuFind, the HTML is rendered.

image

I think(?) that is evidence that a) the current escaping is sufficient and b) no one should use HTML in these descriptions in the first place.  But feel free to push back if you still think html_entity_decode is needed somewhere -- I'm not clear on that.

demiankatz: How does this description display in LibGuides itself? I think our goal should be to mirror that as much as possible. So if the bold word is bold in LibGuides but shows as HTML in VuFind, we need to fix that. But if it shows as HTML in both places, then I think all is well. I would be interested in how a description like `< &lt; &gt; > & &amp;` shows up in both places too -- curious about handling of HTML entities and unescaped special HTML characters.

maccabeelevine: Yeah, the description does get bolded within LibGuides itself.

image

But I assumed actually parsing the entities as entities would be a problem, since `<script>` etc., although I guess the CSP policy with nonce would alleviate some of that risk.

maccabeelevine: I'll try `html_entity_decode` and see if that improves things.

demiankatz: You could also use `strip_tags` to get rid of HTML tags (in addition to `html_entity_decode` to handle the entities) if you want to avoid the issue of displaying HTML safely. Obviously not a perfect solution, but maybe the easiest compromise.

EreMaijala: Want to borrow our [cleanHtml](https://github.com/NatLibFi/NDL-VuFind2/blob/dev/module/Finna/src/Finna/View/Helper/Root/CleanHtml.php) helper? :) It's meant for cases like these where external HTML needs to be displayed in a safe way.

maccabeelevine: I went with @demiankatz 's combo of `html_entity_decode` and `strip_tags`.  

I like @EreMaijala 's utility class and I think it would be useful in a lot of cases.  I just don't see the UX here where HTML is necessary on the description of a journal or database.  LibGuides itself doesn't provide a rich text input; the fact that you can supply HTML may be more of a hidden feature.  I can't imagine the UI horrors that would result if people started adding random style attributes to parts of the description.  And while links might be useful in some contexts, I think it would confuse the primary call-to-action of the result item which is to click into the journal (or database).

demiankatz: Thanks, @maccabeelevine! I think this is fine for now, except that I would reverse the order of the functions -- I think we want to `html_entity_decode(strip_tags(...))` instead of the reverse. Otherwise, if the HTML input is `This book talks about the HTML &lt;b&gt; tag` then the output will be incorrect. :-)

maccabeelevine: Good call and fixed.

demiankatz (approved):
Thanks, @maccabeelevine! I spotted one last tiny issue which I just took care of myself to expedite things (the type annotations in one file said "boolean" instead of the expected "bool"). I believe it's ready to merge now, as soon as CI passes.

PR #3129, opened 2023-09-29
Accessible cart update announcements

ckaz:
…kbag

When screen reader users add a medium to the bookbag, the bookbag will not properly announce such changes even though the bookbag item count increases/decreases. Following the suggestions of out accessibility testers, we suggest
marking the bookbag container as a live region using `aria-live="polite"`
In addition, we should use `aria-atomic="true"` so screen readers will not only re-read the textual changes but the entire element.
This solution was built into out catalogs by @RLangeUni and works well. We are currently handicapped as we cannot test VuFind 8 or 9 locally.

Sorry, I'll be on leave until 18. October. More on this after that date.


demiankatz (commented):
This makes sense to me @ckaz: only one small question: is there a strong reason why we should do this via Javascript instead of simply adding the attributes to the element in header.phtml (https://github.com/vufind-org/vufind/blob/3b57baa764ccbbdfa5d3bfc72d8f1b38293e0b1e/themes/bootstrap3/templates/header.phtml#L23)? I would be fine with either approach, but modifying the template might be simpler, unless you want to be sure the cart works correctly even if it is applied to a different target element (which makes sense in theory but doesn't seem likely in practice).

demiankatz:
@RLangeUni, since @ckaz is away, are you able to comment on my question above regarding the possibility of doing this in the template vs. in the Javascript?

demiankatz:
@ckaz, since you're back today, I thought I'd go ahead and revise the PR as suggested (since I really don't see a specific advantage to doing this in Javascript, and the template-based approach is simpler and more readable). If you can think of a reason to change it back, I'm happy to revert my commit, but otherwise I'm hoping we can merge this soon in its new form. I'd still like input from others as well!

ckaz (commented):
Perfect! Please go ahead

ckaz:
Dear Demian, yes, I'm just back. Sorry for the delay. I can't think of any good reason why we shouldn't use your simplified code version. From our local tickets it was impossible to find out why it was done in Javascript though I have an inkling that it was done this way to apply it to all the catalogs we manage for other institutions without having to change the header template of each of them. 
So, please go ahead! Cheers!

demiankatz (approved):
Great, this has my approval, then! I'll just leave this open a little longer in case anyone else has something to add.

crhallberg (approved):
Thank you for this!

PR #3130, opened 2023-09-29
Fix hyphen-in-path bug in ImportLokaliseCommand.

demiankatz:


ThoWagen (approved):
This fixes the problem. Thanks!

demiankatz:
Thanks for reporting the bug!

PR #3131, opened 2023-09-29
Do not jump to single search result when jumpto is disabled

ThoWagen:
If one has jump_to_single_search_result set to true and only gets one result in a column of combined search one jumps to this record. This should not happen and is fixed in this PR.

### DIFF
--- a/module/VuFind/src/VuFind/Controller/AbstractSearch.php
+++ b/module/VuFind/src/VuFind/Controller/AbstractSearch.php
@@ -472,6 +472,8 @@ class AbstractSearch extends AbstractBase
      */
     protected function processJumpToOnlyResult($results)
     {
+        // If jumpto is explicitly disabled (set to false, e.g. by combined search),
+        // we should NEVER jump to a result regardless of other factors.
         $jumpto = $this->params()->fromQuery('jumpto', true);
         if (
             $jumpto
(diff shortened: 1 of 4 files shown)


demiankatz (commented):
Thanks, @ThoWagen, this is a good catch! However, do we really need to add a blockJumpto parameter? I thought the existing code to set jumpto to false was designed to have the same effect. Could it be we just need to check for a false jumpto in `processJumpToOnlyResult` and the other changes are unnecessary?

(I'm happy to merge this as-is if there's a reason I'm overlooking for its necessity, but just wondering if we can simplify).

ThoWagen:
I was not sure if there are cases where jumpto is set to false in the query parameter but if jump_to_single_search_result=true there should still be a redirect. This would not be very intuitive but it is the current logic and I only wanted to fix this for combined searches.

I can definitely change this if there are no unwanted side effects.

demiankatz:
@ThoWagen, the jumpto parameter as currently implemented should either be a positive integer (the index of the record to jump to), or false (indicating that jumping is explicitly disabled). So I think it's fair to say that `jumpto=false` serves exactly the same function as your proposed `blockJumpTo` parameter. The only thing to be careful about in processJumpToOnlyResult is the fact that we only want to block jumping if the parameter is EXPLICITLY set to false, but NOT if it is unset. This could probably be most easily addressed by defaulting to true in processJumpToOnlyResult, e.g.

```
$jumpto = $this->params()->fromQuery('jumpto', true);
if (
    $jumpto
    && ...
```

What do you think?

ThoWagen:
That makes sense. I changed it.

demiankatz (approved):
Looks good to me, and passes both the automated test suite and my own manual checks. I've just added a comment to the code to provide some additional context. Thanks again for catching and fixing this!

PR #3132, opened 2023-09-29
Upgrade to Solr 9.3.0.

demiankatz:
This PR upgrades to the latest Solr release and removes the deprecated enableRemoteStreaming setting (which is now controlled by an environment variable instead of the config file). While our existing enableRemoteStreaming settings were inconsistent with one another, I do not believe that there was any good reason to have it enabled under any circumstances, so removing it should not cause a problem.

TODO
- [x] Rebuild browse handlers
- [x] Test everything
- [ ] Finish https://github.com/vufind-org/vufind-browse-handler/pull/44 after merging


solr/vufind/authority/conf/solrconfig.xml
```diff
@@ -32,7 +32,7 @@
        that you fully re-index after changing this setting as it can
        affect both how text is indexed and queried.
     -->
-  <luceneMatchVersion>9.1.1</luceneMatchVersion>
+  <luceneMatchVersion>9.3.0</luceneMatchVersion>
```
EreMaijala: I don't think this is correct. Solr 9.3.0 ships with Lucene 9.7.0, so we should bump this to `9.7`. Solr's examples don't include the third part, so I'd follow that convention.

EreMaijala:
@demiankatz As far as I can see this could well be included in VuFind 9.1.

demiankatz:
Thanks, @EreMaijala, let's discuss on tomorrow's Community Call!

EreMaijala (changes requested):

solr/vufind/authority/conf/solrconfig.xml
demiankatz: I've fixed this and also re-run the full test suite to confirm that nothing has broken.

EreMaijala (approved):

PR #3133, opened 2023-10-02
Sierra: Fix and improve item/bib handling with transaction history

EreMaijala:
- Allows a bib to be fetched also when the item has already been deleted
- Fixes an issue when the transaction list contained multiple items for the same bib where only one item had it's bib mapped to it.

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
@@ -2271,6 +2271,9 @@ class SierraRest extends AbstractBase implements
                     );
                 }
             }
+            $callNumber = isset($item['callNumber'])
+                ? $this->extractCallNumber($item['callNumber'])
+                : $bibCallNumber;
             $volume = isset($item['varFields']) ? $this->extractVolume($item) : '';
 
             $entry = [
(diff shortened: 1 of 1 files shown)


module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -2366,18 +2365,6 @@ protected function getItemStatusesForBib(string $id, bool $checkHoldings, ?array
...
     }
 
-    /**
-     * Extract the actual call number from item's call number field
-     *
-     * @param string $callNumber Call number field
-     *
-     * @return string
-     */
-    protected function extractCallNumber(string $callNumber): string
```
demiankatz: This seems to be reverting changes from another recent Sierra PR. Is that intentional? It doesn't seem related to the intent of this one...

demiankatz (changes requested):
One thing caught my eye right away, and I thought I'd better ask about it before looking at the rest of the code in more detail.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Oops, I thought I took care of that. That's what you get for working with so many different versions. Should be fixed now.

demiankatz (approved):
Looks good to me -- thanks, @EreMaijala!

PR #3134, opened 2023-10-02
Update dependencies for release 9.1.

demiankatz:
This PR updates all dependencies to latest versions for release 9.1, with two exceptions:

  * dev tools are kept at current versions, as we have already synchronized all subprojects; we should update these again when we begin work on 10.0.
  * laminas-diactoros cannot be upgraded without a parallel update to laminas-psr7bridge, and that update requires PHP 8.1... so that update will have to wait until we raise our PHP minimum requirement.

TODO
- [x] Run full test suite


demiankatz:
All tests are passing!

EreMaijala (approved):

PR #3138, opened 2023-10-03
Bolder text in filter dropdown.

demiankatz:
@sturkel pointed out that the drop-down control for combined filters used regular-weight font while all the surrounding buttons and filters have heavier font weights. This PR adjusts the combined filter to match for a more consistent visual experience.


sturkel89:
### **DEV BRANCH:** 

**Bootstrap3** - Combined filters button is normal, rest of filter buttons are bold

image

**Bootprint3** -Button text text is bold on all filter buttons (no change needed)

image

**Sandal** - Combined filters button is normal, rest of filter buttons are bold

image

### **TEST BRANCH:**

**Bootstrap3** - All filter buttons have same bold font. 

image

**Bootprint3** - All filter buttons have same bold font. 

image

**Sandal** - All filter buttons have same bold font.

image

### **CONCLUSION**
In each of the three themes, all filter buttons now have the same bold font. So, the problem this PR seeks to address has been solved. 

**FOR A FUTURE PR**
We still may want to consider revamping the look of the combined filter button. In all cases, the combined filter button is taller than the other filter buttons, with extra padding in all dimensions. Also, it has rounded corners, while the rest of the filter buttons have less padding and sharp corners. I think the heights should be brought into alignment so that vertical line spacing is normalized; not sure whether rounded corners should be changed or not.

One theme-specific nitpick: in Bootprint3, there seems to be more top padding than bottom padding in the **regular filter buttons**. The text should probably be vertically centered within the button, and the button might look better if it were a bit taller overall.

Thanks, @demiankatz!

xmorave2 (approved):

PR #3140, opened 2023-10-04
Sierra: Add translation prefix support for ILS messages.

EreMaijala:
Very much alike the one in Alma.

### DIFF
--- a/config/vufind/SierraRest.ini
+++ b/config/vufind/SierraRest.ini
@@ -36,6 +36,14 @@ use_prefixed_ids = false
 ; Default is 2. Set to 0 to disable retries.
 ;http_retries = 2
 
+; Translation prefix to use for messages coming from Sierra. By default a prefix is
+; not used, but a prefix may be used to distinguish the codes from any other
+; translations (or other libraries). To use a simple prefix in the default text
+; domain:
+; translationPrefix = "sierra_"
+; translationPrefix = "ILSMessages::"
+;translationPrefix = "ILSMessages::"
+
 [Authentication]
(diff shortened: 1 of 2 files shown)


config/vufind/SierraRest.ini
```diff
@@ -36,6 +36,14 @@ use_prefixed_ids = false
 ; Default is 2. Set to 0 to disable retries.
 ;http_retries = 2
 
+; Translation prefix to use for messages coming from Sierra. By default a prefix is
+; not used, but a prefix may be used to distinguish the codes from any other
+; translations (or other libraries). To use a simple prefix in the default text
+; domain:
+; translationPrefix = "sierra_"
```
demiankatz: The comments are a little confusing here; you have text explaining the first example, but no text explaining the second example... and then the "actual" commented-out setting is identical to the second example. How would you feel about eliminating the two example lines, and changing the last sentence to something like the following?

You can use a simple prefix in the default domain (e.g. "sierra_") or a custom text domain (e.g. "ILSMessages::").

demiankatz (changes requested):
Looks good to me, apart from a minor complaint about the comments.

config/vufind/SierraRest.ini
EreMaijala: It was just copied from Alma.ini, but I'll revise.

EreMaijala: Should be better now.

demiankatz (approved):
Thanks, @EreMaijala!

PR #3141, opened 2023-10-04
Alma: Revise comment of translationPrefix setting.

EreMaijala:
Follow-up from #3140. I changed the example to "ILSMessages::" because it's an actual existing domain.


demiankatz (approved):
Thanks, @EreMaijala!

PR #3142, opened 2023-10-04
Multi page selection in favorite list

ThoWagen:
Currently, only the records of one page can be selected in the favorite list.

In this PR, multi page selection is introduced. This means that one can select records from multiple pages in the favorite list or even all.

This could also be implemented for the result list in the future but it would be more complicated to store the selection for different searches and the option to select all results wouldn't make sense for thousands of results.


demiankatz:
Thanks, @ThoWagen, this looks like a useful feature. I've scheduled this PR for 10.0 to be sure it gets more attention after 9.1 is completed. My only thought in the meantime is that in some situations, the limitation of only selecting one page at a time may have benefits, since it imposes limits on functionality that may have performance or capacity caps. As you say, search results are challenging because they can be very large... but it is also theoretically possible for a favorite list to have hundreds or thousands of records as well. This is very likely to exceed the capacity of the book bag, so "select all" and "add to cart" is likely to be problematic here. And other things like exports may not scale up so well. Have you done any testing around the edges of this implementation? Is it possible that we need to set some kind of configurable upper limit on how many things can be selected and/or make this whole feature configurable so it can be disabled in situations where it will cause problems?

ThoWagen:
You have a good point. We use this in our system for more than a year now. Once we had a user with over a thousand of records in his favorite list. He wanted to print them all at once and this resulted in running out of memory for loading all of the records. It is very rare that users have that many favorites but we still should address this. I think there are multiple possible solutions for this: 

1. One could restrict the number of favorites a user can have. This is probably the laziest solution and I don't think we should do this. But I just wanted to mention the option.
2. Restricting the number of ids that are provided by VuFind/Search/Favorites/Results and informing the user if the limit is exceeded. Here informing the user might be a tricky part but we wouldn't have to change any other parts.
3. Restricting the the number of records that can be used by the different bulk actions. If the limit is exceeded don't perform the action and inform the users. The limits could be individually configured for the different actions and are also used for any other ways when the actions are used.

I think 3. would be the best solution.

Maybe we could provide both `select_all` and `select_all_on_page` and make them configurable.
What do you think?

demiankatz:
Thanks, @ThoWagen, I agree that option 3 seems like the best solution. I also like the idea of offering the option of both types of "select all" buttons. Perhaps it would make sense to tackle the bulk action limits in a separate PR to prevent the work here from getting too complicated. I could imagine that some are more complicated than others -- e.g. for "export," different methods might potentially have different limitations, so we might need to add some more complex configuration there.

ThoWagen:
Now, multi page selection and the select-all checkboxes are configurable. Supporting all the different options was trickier than expected. But I created Mink tests for them and now it should work.

demiankatz:
@ThoWagen, as discussed on today's Community Call, I think the next step before we can move this forward is to begin implementing limits on bulk actions. We should figure out which actions need limits, where to configure them, etc. Do you have any thoughts on this, or would you like me to try to find time to prototype something? (I'm willing, but it may be a couple of months before I get to it at the current rate of project activity).

ThoWagen:
Maybe we can create a BulkAction section in the config.ini and set the limits there. Then we can experiment with the different actions and find out the best values. But I don't know yet when I will be able to get into this.

demiankatz:
> Maybe we can create a BulkAction section in the config.ini and set the limits there. Then we can experiment with the different actions and find out the best values. But I don't know yet when I will be able to get into this.

Yes, that makes sense to me. Such a section might also offer the opportunity to individually disable unwanted bulk options (by setting limit to 0, for example). Since neither of us knows when we will have time for this yet, whoever gets there first can comment here when beginning work, and then we will ensure that we don't both do it at the same time. :-)

ThoWagen:
@demiankatz I will have a look into the limits now.

demiankatz:
Thanks, @ThoWagen! Sorry I didn't manage to get there myself. Also, it looks like there's a small conflict that needs to be resolved here when you have a moment.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ListItemSelectionTest.php
```diff
@@ -0,0 +1,777 @@
...
+     *
+     * @return \Behat\Mink\Element\DocumentElement
+     */
+    protected function setupTest($config)
+    {
+        $this->changeConfigs(
+            [
+                'config' =>
+                    [
+                        'Social' => [
```
demiankatz: Why not just use `'Social' => $config` here? The extra complexity doesn't seem necessary. If you're going to set the settings individually, maybe it would be better to have two arguments to the function instead of a config array.

demiankatz (changes requested):
Thanks for the progress here, @ThoWagen -- see below for one minor suggestion related to the test suite.

Beyond that, the only remaining question is whether anything should be adjusted to be more future-proof in case we decide to add this functionality to search results as well. As you noted in the description here, there are some obvious challenges to implementing that, and it makes sense to get this functionality merged before attempting to add that. I'm just wondering if we should make any changes to the configuration setting names so that if we add more options in the future, they do not become ambiguous or confusing. Would it make sense to use `multi_page_favorites_selection` and `checkbox_select_all_favorites_type` (or something like that) to further make explicit the limited scope of this functionality? I think if we added similar functionality for other bulk actions, we would need to configure it independently, and if we had the same setting names in multiple sections of config.ini, that might make certain types of errors more common. What do you think?

Once we have settled these finishing touches, I will ask @sturkel89 to do some hands-on testing and see if she can find any remaining problems with the functionality (though I'm optimistic that it is already quite solid, given the thoroughness of the test suite).

ThoWagen:
Unfortunately, there are still some bugs in this PR. I will fix them when I have the time for it. But in this state it is not ready to be reviewed further.

demiankatz:
> Unfortunately, there are still some bugs in this PR. I will fix them when I have the time for it. But in this state it is not ready to be reviewed further.

Thanks, @ThoWagen! Let me know if I can be of any help along the way.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ListItemSelectionTest.php
ThoWagen: I think I expected to have more configs here. But since that isn't the case I changed it.

ThoWagen:
The bugs are fixed after significant refactoring in list_item_selection.js. Also implemented the suggested changes of the config names. 

Now, it should not be hard to add multi page selection for searches as long as we don't also add a "select all global" button. But we would need a "clear selection" button instead. And it could also be implemented for checked out items and holds. Actually in all places where we have select checkboxes on multiple pages. But this is something for followup PRs of course. 

I think we could support some bulk actions for much larger numbers of records if we implement this with some kind of async background jobs and present the users a loading bar or something similar. But since this is not a huge priority for us I don't this I will look into this anytime soon.

config/vufind/Demo.ini
```diff
@@ -46,7 +46,7 @@ services[] = 'custom'
 ; This setting can be used to create fake checked out items for specific records.
 ; The value is a JSON document representing the status information returned by the
 ; driver.
-;transactions = '[{"id":"1234", ... "renewable": true}]';
+transactions = '[{"id":"1234", ... "renewable": true}]';
```
demiankatz: This looks like an unintentional commit.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php
```diff
@@ -646,15 +646,18 @@ public function testEditingDifferentPickupLocations()
         // Place the second hold:
         $page = $this->gotoRecordById('dollar$ign/slashcombo');
         $this->placeHoldAndGoToHoldsScreen($page, ['#pickUpLocation' => 'C']);
+        //        sleep(5);
```
demiankatz: I think these commented lines are not necessary.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ListItemSelectionTest.php
```diff
@@ -0,0 +1,767 @@
...
+ *
+ * Class must be final due to use of "new static()" by LiveDatabaseTrait.
+ *
+ * @category VuFind
+ * @package  Tests
+ * @author   Demian Katz <demian.katz@villanova.edu>
+ * @author   Thomas Wagener <wagener@hebis.uni-frankfurt.de>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org Main Page
+ * @retry    4
```
demiankatz: The AutoRetryTrait has been removed due to incompatibility with PHPUnit 10, so this retry attribute should also be removed.

themes/bootstrap3/templates/myresearch/mylist.phtml
```diff
@@ -87,8 +87,10 @@
     <div><strong><?=$this->transEsc('Tags')?>:</strong> <span class="list-tags"><?=implode(', ', array_map([$this, 'escapeHtml'], $listTags))?></span></div>
   <?php endif; ?>
   <?php if ($recordTotal > 0): ?>
-    <form class="form-inline" method="post" name="bulkActionForm" action="<?=$this->url('cart-myresearchbulk')?>" data-lightbox data-lightbox-onsubmit="bulkFormHandler">
+    <?php $multiPageSelection = $this->config()->get('config')['Social']['multi_page_favorites_selection'] ? ' multi-page-selection' : ''; ?>
```
demiankatz: This could probably use some null coalescing in case the setting is unset, to avoid notices.

demiankatz (changes requested):
Thanks for the progress, @ThoWagen -- I just noticed a few small cleanup things that are probably needed here. Once those are addressed, I'll ask @sturkel89 to do some hands-on testing to see if she can find any problems... and if not, this is probably ready to merge!

config/vufind/Demo.ini
ThoWagen: Sorry, I should have seen this myself. Reverted it.

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/HoldsTest.php
ThoWagen: Should have seen this as well. Sorry. Fixed

module/VuFind/tests/integration-tests/src/VuFindTest/Mink/ListItemSelectionTest.php
ThoWagen: Done

themes/bootstrap3/templates/myresearch/mylist.phtml
ThoWagen: Done

config/vufind/config.ini
```diff
@@ -601,7 +601,7 @@ username_hint = username_only_letters_numbers_and_basic_punctuation
 ;delete_ratings_with_user = false
 
 ; "Remember me" functionality for listed login methods. If the user chooses to save the login,
-; a login token cookie and a database entry will be created. Users' browser information will be 
+; a login token cookie and a database entry will be created. Users' browser information will be
```
ThoWagen: My IDE auto formats things like this. Do you prefer not to have this in PRs or does it not bother you? I could deactivate this kind of formatting in my IDE.

demiankatz: This is fine. I prefer removing trailing whitespace, and my IDE also fixes it. Finding things like this should be fairly rare, and is only likely to happen in config files anyway due to tools fixing it elsewhere.

demiankatz:
Thanks for the fixes, @ThoWagen! I think this should now be ready for @sturkel89's testing -- I'd encourage her to experiment with the settings added to config.ini here (multi_page_favorites_selection and checkbox_select_all_favorites_type) and make sure they behave as expected in various combinations/scenarios.

sturkel89 (commented):
I have reviewed the test branch, and as far as I can tell the branch is producing the desired behavior and is working properly. 

There are a few oddities and wish list items I noticed in both test and dev that I'll mention below, but they are probably outside the scope of this PR.

My only tiny critique that is specific to the test branch is this: this branch adds an **orphan checkbox** below the final line in the Favorites list; the box is not there in dev. When you **Select All Entries**, the orphan checkbox is checked, but nothing corresponds to it in the output. When you **Select All Entries on Page**, the orphan checkbox is not checked. This is not particularly problematic, but it would be slightly more elegant to not have the floating extra box visible to the user.

image

Otherwise, all of the option settings work properly. 
- With multi_page_favorites_selection set to **true**, selection is sticky; if you select everything on one page, then go to the other page, then go back to the first page, the selections are still checked. 
- When the option is set to **false**, selection is not sticky; you can select everything on the current page, but when you leave and then come back, the selections are lost. 
- Each of the options under checkbox_select_all_favorites_type operates as described in the variable definitions in config.ini.

----------
### Other Issues
Here are my non-specific comments on problems/issues/desired enhancements in the Favorites list.

### Sorting is broken
The favorites list can only be sorted in Title order, which is the default. You get a **red box error** when you try any other kind of sort. This seems to be true in dev as well. Example: 
http://localhost/vufind_test/MyResearch/Favorites?sort=author

image

NB: Title sort seems to be accurate, although _Al Gore_ sorts in an unexpected location, between _On the culture of the Silesian beet_ and _Premiums offered by the Dublin Society, for agriculture and planting. 1787_. Maybe there is a hidden title in the _Al Gore_ record that is used for sorting?

### Numbering
We only see numbers on the items on favorites lists when we print. Otherwise, the lists are unnumbered. I would prefer that the favorites list display (and the other areas of My Account) included numbers, as we see in search results and browse lists outside the My Account area.

Regular view:


Print view:


### Weird shading in Availability lozenge
In Demo catalog, in print view, both test and dev display weird shading in the availability lozenge:
image

demiankatz (changes requested):
Thanks, @sturkel89!

@ThoWagen, can you investigate the extraneous checkbox that @sturkel89 is seeing? That seems like the main problem that needs to be addressed before this can be merged.

Many of @sturkel89's other observations seem to indicate problems or enhancements that need to be addressed. I'll begin investigating them, and they can be fixed through separate PR's or JIRA tickets as appropriate.

I also notice that there is a conflict in config.ini that needs to be resolved here.

demiankatz:
Update on @sturkel89's other findings:

1.) The broken favorites sorting is a recently introduced bug. I am currently working on a PR to fix it and add a test to prevent regressions.

2.) The "Al Gore" title sorting issue is just caused by bad MARC coding in the test data set -- there is a skip indicator on the title field that causes the first four characters to be ignored, so it's sorting as "ore"

3.) It's probably worth opening a JIRA ticket to discuss numbering of lists. I agree that consistency would be good, but do we want to make this configurable? Do we need to align CSS styles so numbering can be disabled via styles? It would be helpful to have a place to work through these things.

4.) The odd print display is probably a minor CSS problem. I'll take a look if I have time after finishing the bug fix I'm currently working on. If I don't get to it, maybe @crhallberg can help.

demiankatz:
See #3427 for a fix to the broken favorites sorting.

ThoWagen:
@sturkel89 Thanks for the hint. The checkbox was supposed to be hidden. It is fixed now.

sturkel89 (approved):
The orphan checkbox is now hidden, and all of the multi-page selection features in Favorites are still working properly. Ready to merge!

sturkel89 (changes requested):
I spoke too soon! I notice that the label for selecting all on page is now showing a variable name instead of the appropriate public-facing label. Can you fix this before the branch is merged? THANKS!

ThoWagen:
@sturkel89 `select_all` is still being translated. Is it possible that you tested other branches in the meantime and the languages cache got replaced? That is usually the reason why translations don't work properly when switching between different branches.

sturkel89 (approved):
> @sturkel89 `select_all` is still being translated. Is it possible that you tested other branches in the meantime and the languages cache got replaced? That is usually the reason why translations don't work properly when switching between different branches.

Yes, that is definitely a possibility! I went into another branch and came back out, and also cleared some caches along the way to repair another problem. I only noticed this issue when I went back into your branch; I hadn't noticed it when I went in the first time today, so it probably wasn't a problem then.

Sorry for the mix-up!

demiankatz:
@sturkel89 , the problem you describe may be caused by an outdated cache. Please see if deleting the local/cache/languages directory fixes it.

module/VuFind/src/VuFind/Controller/Feature/ListItemSelectionTrait.php
```diff
@@ -0,0 +1,69 @@
...
+        $checkedDefault = $this->params()->fromPost('checked_default') !== null;
+        $nonDefaultIds = $this->params()->fromPost('non_default_ids');
+        $allIdsGlobal = $this->params()->fromPost('all_ids_global', '[]');
+        if ($nonDefaultIds !== null) {
+            $nonDefaultIds = json_decode($nonDefaultIds);
+            return array_values(array_filter(
+                json_decode($allIdsGlobal),
+                function ($id) use ($checkedDefault, $nonDefaultIds) {
+                    $nonDefaultId = in_array($id, $nonDefaultIds);
+                    return $checkedDefault ^ $nonDefaultId;
```
demiankatz: Would it be better to use the logical `xor` instead of the bitwise `^` here? This seems like a logical rather than bit operation.

module/VuFind/src/VuFind/Controller/Feature/ListItemSelectionTrait.php
```diff
@@ -0,0 +1,69 @@
...
+trait ListItemSelectionTrait
+{
+    /**
+     * Get selected ids
+     *
+     * @return array
+     */
+    protected function getSelectedIds()
+    {
+        $checkedDefault = $this->params()->fromPost('checked_default') !== null;
```
demiankatz, line 55: It might be helpful to add a comment here explaining exactly what these three values represent. I'm pretty sure that checked_default and all_ids_global are reasonably self-explanatory, but non_default_ids is less obvious. I had to review other code to fully understand what was happening here.

Maybe something like:

```
// Values may be stored as a default state (checked_default), a list of IDs that do not
// match the default state (non_default_ids), and a list of all IDs (all_ids_global). If these
// values are found, we need to calculate the selected list from them.
```

themes/bootstrap3/templates/myresearch/bulk-action-buttons.phtml
```diff
@@ -4,10 +4,23 @@
...
+  $checkboxSelectAllType = $socialConfig['checkbox_select_all_favorites_type'] ?? 'none';
+  ?>
+  <?php if(in_array($checkboxSelectAllType, ['on_page', 'both'])): ?>
+    <div class="bulk-checkbox">
+      <input type="checkbox" name="selectAll" class="checkbox-select-all" id="myresearchCheckAll">
+      <label for="myresearchCheckAll"><?=$this->transEsc('select_all_on_page')?></label>
+    </div>
+  <?php endif; ?>
+  <?php if($multiPageSelection && in_array($checkboxSelectAllType, ['global', 'both'])): ?>
+    <div class="bulk-checkbox">
```
demiankatz: Should we hide this one by default, then unhide it in the Javascript? Otherwise, when Javascript is disabled, the control appears but does not function.

themes/bootstrap3/js/list_item_selection.js
```diff
@@ -0,0 +1,213 @@
+/*global VuFind */
+
+VuFind.register("listItemSelection", function ListItemSelection() {
+
+  function _sessionSet(form, key, data) {
+    let formId = form.id;
+    let formStorage = JSON.parse(window.sessionStorage.getItem(formId) || "{}");
```
demiankatz, line 7: Would it make sense to call `_sessionGet` here to reduce redundant logic?

module/VuFind/src/VuFind/Controller/Feature/ListItemSelectionTrait.php
demiankatz: If we do this, it might also be helpful to add, above the final return statement:

```
// If we got this far, values were passed in a simpler format: a list of checked IDs (ids),
// a list of all IDs on the current page (idsAll), and whether or not the whole page is
// selected (selectAll):
```

themes/bootstrap3/js/list_item_selection.js
```diff
@@ -0,0 +1,213 @@
...
+  function _checkIfAllSelected(form) {
+    let nonDefaultIdsInput = form.querySelector('.non_default_ids');
+    let checkedDefaultInput = form.querySelector('.checked_default');
+
+    if (nonDefaultIdsInput !== null && checkedDefaultInput !== null) {
+      let nonDefaultIds = JSON.parse(nonDefaultIdsInput.value);
+      let checkedDefault = checkedDefaultInput.checked;
+      form.querySelectorAll('.checkbox-select-item').forEach(itemCheckbox => {
+        let id = itemCheckbox.value;
+        if (checkedDefault ^ itemCheckbox.checked) {
```
demiankatz, line 124: This is another instance of using bitwise XOR for a logical operation. I guess it will work as long as both values are Booleans (which they should be), but it feels a little strange to me. (I realize that Javascript has no logical XOR operator, so maybe it's the best we can do).

themes/bootstrap3/js/list_item_selection.js
```diff
@@ -0,0 +1,213 @@
...
+    }
+    if (data.checkedDefault !== undefined) {
+      _check(form.querySelector('.checked_default'), data.checkedDefault);
+    }
+    if (data.selectAllOnPage !== undefined) {
+      _check(getItemCheckboxes(form), data.selectAllOnPage);
+    }
+  }
+
+  function _checkIfAllSelected(form) {
```
demiankatz: I wonder if this would be clearer if renamed to `_updateSelectionState`, or something like that. Starting the method name with `_check` implies that it is checking a state rather than performing an action. An unfortunate case of homonyms.

themes/bootstrap3/js/list_item_selection.js
```diff
@@ -0,0 +1,213 @@
...
+    document.querySelectorAll('.checkbox-select-all-global').forEach((checkbox) => {
+      checkbox.addEventListener('change', () => _selectAllGlobalCheckbox(checkbox));
+    });
+    document.querySelectorAll('.checkbox-select-item').forEach((checkbox) => {
+      checkbox.addEventListener('change', () => {
+        let form = checkbox.form ? checkbox.form : checkbox.closest('form');
+        if (form == null) {
+          return;
+        }
+        _writeToForm(form);
```
demiankatz: What is the point of this _writeToForm call? If no data is provided, doesn't all the logic in _writeToForm get skipped?

themes/bootstrap3/js/list_item_selection.js
demiankatz: A related question is whether function headers might be helpful in this code due to its complexity, though I wouldn't insist on it.

demiankatz (changes requested):
Thanks for the progress here, @ThoWagen and @sturkel89. I did a bit more thorough reviewing and testing myself and had a few comments (see below).

Also, a general user interface question: might there be value in having some indicator of how many items are selected somewhere? I realize that space is at a premium, so maybe it's not worth it, but it's possible to get into a situation where many checkboxes are selected, but there is no visible indication that this is the case -- e.g. click "select all entries" and then click "select all entries on page". If there's more than one page of results, this gets you to a state where nothing is checked on the current page, but lots of values are selected. If a user doesn't entirely understand the user interface, this could lead them to unexpected results.

themes/bootstrap3/js/list_item_selection.js
ThoWagen: Since `.checked' are always boolean values in JS I think we can stick to the bitwise XOR here.

themes/bootstrap3/js/list_item_selection.js
ThoWagen: We need the complete storage object here but `_sessionGet` only returns a value. But we could put `JSON.parse(window.sessionStorage.getItem(formId) || "{}");` into an own function though if you think that is worth it.

module/VuFind/src/VuFind/Controller/Feature/ListItemSelectionTrait.php
ThoWagen: changed

module/VuFind/src/VuFind/Controller/Feature/ListItemSelectionTrait.php
ThoWagen: Good idea. I added the comments.

themes/bootstrap3/js/list_item_selection.js
ThoWagen: You are right. Changed it and added a doc comment.

themes/bootstrap3/js/list_item_selection.js
ThoWagen: This is probably a leftover from before refactoring. I removed it.

themes/bootstrap3/templates/myresearch/bulk-action-buttons.phtml
ThoWagen: Done

themes/bootstrap3/js/list_item_selection.js
demiankatz: Of course, sorry for overlooking that. I think the code is readable enough as-is that further refactoring is probably not worth the effort.

ThoWagen:
@demiankatz Concerning your question about the selection count indicator I am working on adding a "clear selection" button that also shows the current selection count.

themes/bootstrap3/templates/myresearch/bulk-action-buttons.phtml
```diff
@@ -4,10 +4,23 @@
...
 <?php $user = $this->auth()->isLoggedIn(); ?>
 <nav class="bulkActionButtons">
-  <div class="bulk-checkbox">
-    <input type="checkbox" name="selectAll" class="checkbox-select-all" id="myresearchCheckAll">
-    <label for="myresearchCheckAll"><?=$this->transEsc('select_all_on_page')?></label>
-  </div>
+  <?php
+  $socialConfig = $this->config()->get('config')['Social'];
+  $multiPageSelection = $socialConfig['multi_page_favorites_selection'] ?? false;
+  $checkboxSelectAllType = $socialConfig['checkbox_select_all_favorites_type'] ?? 'none';
```
demiankatz: Typically, the default values for settings match the default configurations in config.ini -- but here, the config.ini defaults are different from the fallbacks if those settings are commented out. It might be good to either align the defaults in both places, or else to add a note in the config.ini comments indicating what the defaults are if the settings are omitted.

demiankatz (changes requested):
Thanks for addressing all of my review comments, @ThoWagen -- I think we are nearly done.

I have one more minor comment below, plus I'm curious if you saw my question about the UI design in my previous review. Now that I look at how the checkboxes and buttons are wrapped:

image

It seems to me that it wouldn't be too hard to add a [x entries selected] indicator to the right of the checkboxes without impacting the overall screen real estate. I think this might be helpful to further highlight the effect of clicking some of the checkboxes, and to alert the user of off-screen checked boxes. If you dislike the idea, I won't insist on it, but I think it might be useful if it's not hard to implement.

demiankatz:
Thanks, @ThoWagen, regarding the clear button: I see you must have answered that part of my question while I was in the process of reviewing your other changes. Sorry for missing your earlier comment!

ThoWagen:
I added a button that clears the complete selection and is shown if multi page selection is enabled. It also has an indicator for the selection count. It should be useful especially if multi page selection is enabled but the select all global checkbox is not. 

I also removed the flex design that ensures that the "select all" checkbox is shown in the same row as the action buttons since that does not work with multiple checkboxes and the new button. If you still want that behavior in the case that multi page selection is disabled I can look into it. But it might make things more complicated so that it is not worth the effort.

If you approve the current solution I will also add a Mink test for the clear selection button.

demiankatz:
Thanks, @ThoWagen, I think that functionally speaking, this is very helpful and works the way I would expect. Just two thoughts:

1.) I wonder if it would be more flexible from an i18n perspective if the translation string had a placeholder for the count in it (e.g. `Clear Selection (%%count%%)`), instead of using a separate HTML element to hold the count. That would make it easier to customize with something like `Clear all %%count%% selected checkboxes` if somebody wanted to do something like that. I realize that supporting this would require leveraging our Javascript-side translation system, which adds a bit of overhead to everything (passing the translation string to the client through the jsTranslations view helper, and using VuFind.translate to generate the new output). Maybe it's not worth the effort, but I thought it might be worth considering at least.

2.) I think the aesthetics of the button look a little bit off. In some themes, the button font doesn't match other fonts on the page. In some places the button text is bold while other text is not. In every situation, the text seems out of alignment with the checkboxes, and the height of the clear button is different from the buttons on the line above (which isn't necessarily a bad choice but might look a little distracting). Here's what I'm seeing, at least:

bootstrap3:
image

bootprint3:
image

sandal:
image

Do you think there might be some simple improvements we could make to the styles?

themes/bootstrap3/templates/myresearch/bulk-action-buttons.phtml
ThoWagen: Changed the fallback values to match the default values in config.ini

ThoWagen:
@demiankatz 
About 1.): I think thats a good idea. I changed it.

About 2.): I revised the design. Now the text in the button should be aligned with the checkbox labels and also the font style should match. I intended to make have a lower height  for the button. I think it would be to large if it has the same size as the buttons above. But I can see that the design may not yet be optimal. However, I am not sure how to improve it. Do you have any idea? Or maybe some one else could share their opinion.

demiankatz:
Thanks, @ThoWagen, I think this is much better now. I agree that increasing the button height would likely make things worse rather than better. I'll ask @sturkel89 to give this one more look in case she has any further opinions, and if it looks good to her, I think we can merge it.

sturkel89 (approved):
I think the Clear Selection button and the selected number indicator are a great improvement to this feature! 

I reviewed the improved versions in all three branches. The font size and styling are consistent within each branch between the relevant elements, and the button height seems to be just right. Functionality is perfect, as far as I can see (including sorting within Favorites lists!).

I think we're ready to merge! Thanks for all of your good work on this, Thomas!

demiankatz (approved):
Thanks, @sturkel89 -- merging now!

PR #3143, opened 2023-10-04
LibGuides Profile: Remove whitespace if no photo

maccabeelevine:
This fixes a display bug where there was an attempt to display the profile image even if none was present.  When that happens, $account->profile->image is still defined but ...image->url is an empty string.


demiankatz (approved):
Good catch, @maccabeelevine!

PR #3144, opened 2023-10-05
Fix API specification

xmorave2:


EreMaijala (approved):
Thanks! Particularly fullRecord was an odd one.

PR #3145, opened 2023-10-05
FOLIO: account for spelling change in fulfillmentPreference property

pawelmol:
Book orders doesn't work properly starting from the Orchid version of Folio. There was a spelling error in property name 'fulfilmentPreference' in Folio. So I did appropriate change in Vufind. Should this change be also merged to any of Vufind release versions?

Link to Folio changes:
https://github.com/folio-org/mod-circulation/commit/2cba42dee2fc4dad46e9e641819aea6424dedb26


demiankatz (changes requested):
@pawelmol, thanks for catching this! However, I do not believe that this change impacts the Orchid release -- my institution is currently running Orchid (we upgraded in June), and in our instance, the current code works, and this code causes an error about a missing `fulfilmentPreference` setting. Is it possible the change is actually part of the forthcoming Poppy release?

Since Poppy is not officially released yet and many institutions are still running Orchid, I don't think we should make this as an automatic change. Perhaps the best approach would be to create a new setting in Folio.ini to control the name of the setting, so that users can adjust as appropriate to their situations... or to introduce a version number setting and adjust according to that (though this may be hard to keep track of, since FOLIO emphasizes non-numeric versioning).

Please let me know what you think, and whether you need any help with implementation of one of these suggestions.

demiankatz:
I should also note that I wish that the FOLIO community would treat these types of updates as breaking changes that require a major version bump, or else maintain backward compatibility -- it is currently very hard to keep track of this sort of thing without the help of community members. So I do appreciate the warning, @pawelmol!

demiankatz:
Also, I'm not really sure what motivated this change -- fulfilment is the correct spelling in British English, and fulfillment is the correct spelling in American English. Does FOLIO actually have a standard for this? Otherwise, I could see various people changing this setting back and forth for all eternity. :-)

demiankatz:
Also note that there are some failing tests here -- once we settle on a solution, we'll need to make sure that the test suite also accounts for it. :-)

maccabeelevine:
Is it possible to just send both versions of the spelling as separate properties?  I think many Okapi interfaces will ignore an unknown property (whichever one is unknown to that version of mod-requests), although maybe not POSTs.

Then at some future point the old spelling version could be removed.

pawelmol:
We are using Folio with Vufind for test purposes in our local environment. We are testing snapshot version of Folio, and we firstly used a snapshot from the beginning of this year. Orders in Vufind worked perfectly. Then in June we updated the Folio to June snapshot and the problem with orders began. After some investigation in source code I came up with this solution.

I have found jira ticket connected to this change (https://issues.folio.org/browse/CIRC-253) and @demiankatz you are right, it's not in Orchid release. I am not sure now if this change is/will be in any folio release, because there is no Folio release specified in the jira ticket. But on the other hand CIRC-253 blocks other tickets and there version is set to Poppy and status set to closed. So I don't know what to think. I mean, is there any chance that CIRC-253 is now in the master branch and then will be excluded before Poppy release?

pawelmol:
> Is it possible to just send both versions of the spelling as separate properties? I think many Okapi interfaces will ignore an unknown property (whichever one is unknown to that version of mod-requests), although maybe not POSTs.
> 
> Then at some future point the old spelling version could be removed.

I have tested this:
`        $requestBody = $baseParams + [
            'requestType' => $holdDetails['status'] == 'Available'
                ? 'Page' : $default_request,
            'requesterId' => $holdDetails['patron']['id'],
            'requestDate' => date('c'),
            'fulfilmentPreference' => 'Hold Shelf',
			'fulfillmentPreference' => 'Hold Shelf',
            'requestExpirationDate' => $requiredBy,
            'pickupServicePointId' => $holdDetails['pickUpLocation'],`

 and I have:

`Unrecognized field "fulfilmentPreference" (class org.folio.rest.jaxrs.model.Request), not marked as ignorable (29 known properties: "cancellationAdditionalInformation", "holdShelfExpirationDate", "searchIndex", "itemId", "instanceId", "item", "proxyUserId", "instance", "fulfillmentPreference", "id", "requestDate", "cancelledDate", "awaitingPickupRequestClosedDate", "patronComments", "requestType", "requestExpirationDate", "cancelledByUserId", "metadata", "status", "requestLevel", "requesterId", "holdingsRecordId", "tags", "cancellationReasonId", "position", "proxy", "pickupServicePointId", "deliveryAddressTypeId", "requester"])`

so in okapi version from June, this doesn't work. The old property name should be mark as ignorable in Folio code to make this work.

pawelmol:
But after this error message holding looks like ordered in Vufind :) :)

demiankatz:
Thanks for testing, @pawelmol. Given that all these tickets are closed and the commit is in the master branch, it does seem likely to me that this change will be included in some future version of FOLIO (and Poppy seems likely, but not certain). I'd strongly recommend that the FOLIO community not make this change, as the costs seem to hugely outweigh the benefits, but it seems too late to undo now, given the large number of dependent tickets on CIRC-253 that indicate equivalent changes made across other parts of the project.

I'm curious about the fact that you're getting an error response but seeing the request placed anyway. If @maccabeelevine's proposed fix actually works, it's certainly temptingly simple... but if it's going to cause errors and potentially unpredictable future behavior, I wonder if we'd be better off trying one field name and checking for an error, then falling back to the other field name if the first name fails. That would double up API requests when using older versions of FOLIO, but I don't think hold volume is usually high enough that this would be a major problem, and it might prevent different problems from cropping up in the future.

damien-git:
If we don't want to add a config, we could check the module's version with a call like `/_/proxy/modules?filter=mod-circulation`. The change starts at version 24.

maccabeelevine:
> If we don't want to add a config, we could check the module's version with a call like `/_/proxy/modules?filter=mod-circulation`. The change starts at version 24.

That's elegant.

demiankatz:
> > If we don't want to add a config, we could check the module's version with a call like `/_/proxy/modules?filter=mod-circulation`. The change starts at version 24.
> 
> That's elegant.

Yes, definitely nicer than my proposed "try it, and if it fails try something else" approach. And we could potentially cache version numbers to reduce lookup frequency.

pawelmol:
> If we don't want to add a config, we could check the module's version with a call like `/_/proxy/modules?filter=mod-circulation`. The change starts at version 24.

I like this solution too. I will change code next week I think.

demiankatz:
Thanks, @pawelmol! Let me know if you need help with anything.

demiankatz:
@pawelmol, just checking in to see if you've had a chance to look at this further. I want to be sure we get this done in time for the 9.1 release at the end of the month, so if you need help getting it done, let me know and I'll see what I can do to support you next week.

pawelmol:
Sorry I had a busy week, I am working on it right now :) @damien-git I used '/_/proxy/tenants/tenant/modules?filter=mod-circulation&latest=1' endpoint. From okapi documentation this returns modules enabled for a tenant (https://s3.amazonaws.com/foliodocs/api/okapi/p/okapi.html#proxy_tenants__tenant_id__modules_get) Is it right?

pawelmol:
@demiankatz I don't have php development environoment set up properly for vufind yet, and if we want to include this fix in 9.1 version could you you create tests for it? This would definately speed things up.

demiankatz:
Thanks, @pawelmol! I should be able to help with tests later today -- stay tuned.

damien-git:
@pawelmol You might want to parse the response to extract the version number. I would pick the id of the last entry in the returned array, and extract the version from that id (an environment can have several versions of the same module). The id can be something like "mod-circulation-23.5.7".
_Edit_: I missed that you were using the "latest" parameter. I guess you will only get one back then. But there might be numbers elsewhere in the result, and it would also be better to parse the result to check consistency.

demiankatz:
@pawelmol, as promised, I have fixed the broken tests as well as adding a new test to confirm that the code works correctly in mod-circulation versions 23 and 24 (covering both sides of the cut-over). I've also made a few fixes and optimizations to the proposed driver changes.

@damien-git, I agree that parsing the response a little more specifically is probably a good idea. I'll propose a change shortly...

demiankatz:
@damien-git, what do you think of https://github.com/vufind-org/vufind/pull/3145/commits/ca794051b417f78ecb976c41ed4f6723c284b086? This does somewhat more targeted parsing of the JSON, and it logs a debug message if it runs into something unexpected.

demiankatz:
I've also made a few more minor tweaks -- applied more specific typing to the new method, switched from private to protected to improve extensibility, and added logic to avoid caching "0" values, since these may represent error conditions that we do not want to persist.

I believe this is probably ready to merge, but I'm going to hold off on approving it until at least one or two other people give it the final thumbs up, since I don't want to approve my own work without other eyes on it.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
```diff
@@ -1409,6 +1409,40 @@ public function getMyHolds($patron)
         return $holds;
     }
 
+    /**
+     * Get latest version of a $moduleName enabled for a tenant.
+     * Result is cached.
```
maccabeelevine: What this module actually returns is not the "version" but the "major version" in semver terminology.  Which I think is what we want given we only care about breaking changes (and it works great).  But the description should be specific.  "Version" would be major.minor.patch.

maccabeelevine (commented):
I can't get an end-to-end test going (local issues) but the new getModuleVersion function works great!  And the general logic looks sound.

module/VuFind/src/VuFind/ILS/Driver/Folio.php
demiankatz: Good point! I've revised the method name and comment. This way, if we have to get to more specific versions for a future check, we can more easily refactor.

maccabeelevine (approved):
The module version function now looks good to me!

demiankatz (approved):
Thanks, @maccabeelevine, I think this is ready now, but I'll wait to see if @pawelmol or @damien-git have any final comments before I merge. Please let me know before end of week if you have any concerns!

damien-git:
> @damien-git, what do you think of [ca79405](https://github.com/vufind-org/vufind/commit/ca794051b417f78ecb976c41ed4f6723c284b086)? This does somewhat more targeted parsing of the JSON, and it logs a debug message if it runs into something unexpected.

Looks fine.

pawelmol:
Looks great! Thank you @demiankatz for help. Thank you all for good ideas how to solve this issue :)

demiankatz:
Yes, excellent collaboration, everyone -- much appreciated. Looks like this is ready to merge!

PR #3151, opened 2023-10-05
Create visualization.phtml

diegojmacedo:
creation of the visualization.phtml file for Brazilian Portuguese


demiankatz (approved):
Thanks, @diegojmacedo! There was a small typographical error which I fixed in https://github.com/vufind-org/vufind/pull/3151/commits/36175ec1b1447042d2d3eff87005149cbe89fb65, but this otherwise looks good to me. Merging now... (but please let me know if I made any mistake in my change).

PR #3153, opened 2023-10-06
Deprecate custom EDS logging

maccabeelevine:
It looks like this custom EDS logging code dates to the original EDS implementation.  All of these classes already `use LoggerAwareTrait` (either directly or via inheritance) so logging works fine without it.  Logging actually works better without it, because it inherits the formatting of the logger.  Out-of-the-box this seems to include the class name, which is already a big advantage for grep-ing through to find something.

Example before:
`2023-10-06T13:22:28-04:00 DEBUG (7): POST: https://eds-api.ebscohost.com/edsapi/rest/search?`
after:
`2023-10-06T13:29:07-04:00 DEBUG (7): VuFindSearch\Backend\EDS\Connector: POST: https://eds-api.ebscohost.com/edsapi/rest/search?`

### DIFF
--- a/module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
@@ -45,7 +45,9 @@ use function is_array;
  */
 abstract class Base implements LoggerAwareInterface
 {
-    use \VuFind\Log\LoggerAwareTrait { debug as public debugTrait; }
+    use \VuFind\Log\LoggerAwareTrait {
+        debug as public debugTrait;
+    }
 
     /**
      * A boolean value determining whether to print debug information
--- a/module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
+++ b/module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
@@ -45,18 +45,7 @@ use function is_array;
  */
 abstract class Base implements LoggerAwareInterface
 {
-    use \VuFind\Log\LoggerAwareTrait {
-        debug as public debugTrait;
-    }
-
-    /**
-     * A boolean value determining whether to print debug information
-     *
-     * @var bool
-     *
-     * @deprecated Set the logging level in the Logging section of config.ini.
-     */
-    protected $debug = false;
+    use \VuFind\Log\LoggerAwareTrait;
 
     /**
      * EDS or EPF API host.
@@ -115,7 +104,6 @@ abstract class Base implements LoggerAwareInterface
      * @param array $settings Associative array of setting to use in
      *                        conjunction with the EDS API
      *    <ul>
-     *      <li>debug - boolean to control debug mode</li>
      *      <li>orgid - Organization making calls to the EDS API </li>
      *      <li>search_http_method - HTTP method for search API calls</li>
      *    </ul>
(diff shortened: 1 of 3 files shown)


demiankatz:
Makes sense to me, @maccabeelevine -- would you like me to slate this for 10.0 (since it's technically a breaking change if anyone has local code depending on the debugPrint method)? If you're in a hurry, we could make debugPrint a wrapper around the logger where it isn't already and deprecate it so we can merge sooner without breaking anything, then remove it separately in 10.0.

maccabeelevine:
> Makes sense to me, @maccabeelevine -- would you like me to slate this for 10.0 (since it's technically a breaking change if anyone has local code depending on the debugPrint method)? If you're in a hurry, we could make debugPrint a wrapper around the logger where it isn't already and deprecate it so we can merge sooner without breaking anything, then remove it separately in 10.0.

Yes I had a feeling we'd talk about a smoother deprecation here!  I'll implement that so we can get it in 9.1 (if possible, not critical) along with one other logging improvement.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Base.php
```diff
@@ -41,12 +43,16 @@
...
+    use \VuFind\Log\LoggerAwareTrait { debug as public debugTrait; }
+
     /**
      * A boolean value determining whether to print debug information
      *
      * @var bool
+     *
+     * @deprecated Set the logging level in the Logging section of config.ini.
      */
     protected $debug = false;
```
maccabeelevine: Because of keeping the $debug boolean around (if deprecated), I can't just use LoggerAwareTrait this->debug() as is.  `debugTrait` is a ridiculous function name but we can remove it next month for 10.0.

demiankatz: Ooh, that is annoying! I'd be tempted to fudge a little bit and just chop out the $debug property now. It's a lot more likely that somebody is calling debugPrint in a local subclass than that they're changing the way that property gets assigned. Otherwise, the `@deprecated` comment in this file isn't really correct, and we have to do significant re-rewriting in 10.0. I'd be comfortable putting a changelog note about the removed property and deprecation in 9.1 unless you think that's terribly irresponsible. ;-)

demiankatz: ...and while I say "a lot more likely" I actually think it's extremely unlikely that anybody is extending any of this. :-)

maccabeelevine: Yeah ok I agree 100%.  Will cut out that property and restore some sanity to the function name.

demiankatz: I'm nearly out of time for today, so I'll probably get this merged next week, assuming you finish it up tonight. (No rush in any case).

maccabeelevine: I have a second logging PR to do anyway.  No rush.

module/VuFindSearch/src/VuFindSearch/Backend/EDS/Connector.php
```diff
@@ -136,20 +136,4 @@ protected function httpRequest(
...
     }
-
-    /**
-     * Print a message if debug is enabled.
-     *
-     * @param string $msg Message to print
-     *
-     * @return void
-     */
-    protected function debugPrint($msg)
```
maccabeelevine, line 147: This one can be removed without deprecation since the parent class has another version.

demiankatz (approved):
Thanks, @maccabeelevine! Merging now.

demiankatz:
(If you want to open a follow-up PR to remove this in 10.0, feel free... but no pressure, since I've also noted this in the JIRA ticket for removal of 9.x deprecations in 10.0).

PR #3154, opened 2023-10-06
Syndetics cover images with VuFind fallback

damien-git:
Currently VuFind's implementation for Syndetics cover images creates an image URL based on the desired image size and returns that URL. When Syndetics does not have an image cover, it will either return a generic icon (different from VuFind's) or a fake cover image created based on information it has about the ISBN or other number passed (the title it uses can be different from the one in VuFind). This implementation prevents a fallback using other cover images sources configured in VuFind.
The proposed PR adds a call to get the metadata first, and check if Syndetics actually has a cover image (in that case the metadata contains an image filename). If not, a VuFind fallback can be used.
The PR also adds a new config parameter to restore the previous behavior, because people using only Syndetics as a cover image source might like the Syndetics fallback images as opposed to Vufind's "no image available" icon (that image will still appear when there is no number associated to the record that Syndetics can use, alongside Syndetics fallback images).


demiankatz:
Thanks, @damien-git, this makes a lot of sense! Since it introduces some BC breaks (changed constructor signature, etc.), I'm scheduling it for release 10.0. I expect that it can be merged pretty quickly once work on 9.1 finishes up later this month. Let me know if there's any greater urgency for a resolution here!

module/VuFind/src/VuFind/Content/Covers/Syndetics.php
```diff
@@ -38,23 +40,33 @@
...
+     *
+     * @var bool
+     */
+    protected $useSyndeticsCoverImageFallback;
+
     /**
      * Constructor
      *
-     * @param bool $useSSL Use SSL URLs?
+     * @param \Laminas\Config\Config $config Syndetics configuration
```
demiankatz: This should probably be `?\Laminas\Config\Config` since the factory may pass in null. Might be worth putting the typehint on the actual constructor argument too since we're changing it anyway.

damien-git:
@demiankatz No particular urgency. I am planning to improve other fallbacks in the future to take advantage of this change. For instance I plan to add support for oclc numbers for google books. It will be useful for records with only an oclc number (Syndetics is often not finding a cover image for them). By the way, I noticed you have a different fallback using the title at Villanova. Is this part of VuFind, another cover image source, or a VuFind extension ?

demiankatz (changes requested):
Thanks, @damien-git. I have just one very small suggestion from a more thorough review of the code (see below).

Also, to answer your question, at Villanova, we're using the [makeDynamicCovers](https://github.com/vufind-org/vufind/blob/dev/config/vufind/config.ini#L1041) setting. Is that what you're referring to? If so, let me know if you're interested in seeing all of our font settings. If that's not what you mean, could you share a specific item link that displays the type of cover you're interested in? That will help me figure out where it's coming from.

Finally, this might be a good opportunity to add some test coverage for this class, since it currently has none. If you don't have time to do that, perhaps you could at least share a couple of example XML responses, and then I can put something together when time permits.

damien-git:
> Also, to answer your question, at Villanova, we're using the [makeDynamicCovers](https://github.com/vufind-org/vufind/blob/dev/config/vufind/config.ini#L1041) setting. Is that what you're referring to?

Yes, I think this is it. I missed that for some reason. I will have to try it. I would be interested in any additional config you have about it.

damien-git:
I added the typehint to the constructor. I will work on adding a test next week.

module/VuFind/src/VuFind/Content/Covers/Syndetics.php
```diff
@@ -71,46 +83,141 @@ public function __construct($useSSL = false)
...
+    /**
+     * Calculate the image filename based on the size, without checking if it exists in the metadata.
+     *
+     * @param string $size Size of image to load (small/medium/large)
+     *
+     * @return string|bool Image filename, or false if the size is not 'small', 'medium' or 'large'
+     */
+    protected function getImageFilenameFromSize($size)
+    {
+        switch ($size) {
```
xmorave2: Just small suggestion: As VuFind currently needs PHP 8.0 at least, you can utilize match expression instead of switch statement. I think it may be a bit more readable, and is definitely shorter ;) It could look like this:
```
return match ($size) {
    'small' => 'SC.GIF',
    'medium' => 'MC.GIF',
    'large' => 'LC.GIF',
    default => false,
};

```

demiankatz: Thanks, @xmorave2, I agree that this is a great opportunity to start taking advantage of new PHP 8 features, and cutting back on case statements would be an improvement for sure. :-)

demiankatz:
> > Also, to answer your question, at Villanova, we're using the [makeDynamicCovers](https://github.com/vufind-org/vufind/blob/dev/config/vufind/config.ini#L1041) setting. Is that what you're referring to?
> 
> Yes, I think this is it. I missed that for some reason. I will have to try it. I would be interested in any additional config you have about it.

Here are our non-default settings, though I don't imagine they're particularly interesting:

```
[DynamicCovers]
backgroundMode = villanova
titleFont = "Gotham-Bold.otf"
authorFont = "Gotham-Bold.otf"
titleFontSize = 14
authorFontSize = 10
titleFillColor = white
titleBorderColor = black
authorBorderColor = black
size[*] = 150x190
topPadding = 25
bottomPadding = 10
wrapWidth = 140
```

Note that we've built our own local custom module to render backgrounds behind the text (it just picks a color from a more limited Villanova-specific palette), and we use a University-licensed font to render the text. Some of our other settings are likely optimized to play well with these branding-related configurations.

damien-git:
Here is an example of a generated fallback image from Syndetics, for people who don't use it. The background color is random, and there can also be the author's name on the top left. The type of item at the bottom ("electronic book text" in this example) is a nice addition.
syndetics cover image fallback example

demiankatz (approved):
Thanks, @damien-git. This looks good to me. Since we're already refactoring two of the other cover loaders as part of release 9.1, I think the changes here are consistent with those updates, so I'm going to merge this now and adjust the changelog accordingly.

PR #3155, opened 2023-10-09
Render combination of standard and alternative script fields in RecordDataFormatter

ThoWagen:
Adds the option to render standard and alternative script fields together when using the RecordDataFormatter.

### DIFF
--- a/module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
+++ b/module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
@@ -107,7 +107,7 @@ class RecordDataFormatter extends AbstractHelper
      */
     protected function allowValue($value, $options)
     {
-        if (!empty($value)) {
+        if (!empty($value) || ($options['renderType'] ?? 'Simple') == 'CombineAlt') {
             return true;
         }
         $allowZero = $options['allowZero'] ?? true;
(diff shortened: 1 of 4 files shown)


module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
```diff
@@ -377,6 +377,50 @@ protected function getLink($value, $options)
...
+     *
+     * @return string
+     */
+    protected function renderCombineAlt(
+        $data,
+        array $options
+    ) {
+        // Determine the rendering method to use, and bail out if it's illegal:
+        $method = empty($options['combineAltRenderType'])
+            ? 'renderSimple' : 'render' . $options['combineAltRenderType'];
```
demiankatz, line 395: Is it worth adding a test case to cover combineAltRenderType? I don't think I see one...

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/RecordDataFormatterTest.php
```diff
@@ -369,6 +427,13 @@ public function testFormatting(string $function): void
             'c' => 'c',
             'a' => 'a',
             'b' => 'b',
+            'CombineAlt' => 'Standard Title Alternative Title',
+            'CombineAltPrioritizeAlt' => 'Alternative Title Standard Title',
+            'CombineAltDataMethod' => 'Standard Title Other Alternative Title',
+            'CombineAltNoAltFunction' => 'Book',
+            'CombineAltNoAltValue' => '0',
```
demiankatz, line 457: Not sure I understand why '0' is an appropriate value here -- is it intentional that you mock `getBuildings` and don't provide an expectation with a return value for it? Maybe I'm missing or misunderstanding something...

demiankatz (changes requested):
Thanks, @ThoWagen -- this looks good to me, but I had some test-related questions.

Also, is there a possible use case where two parallel arrays of values need to get matched up instead of being displayed as side-by-side blocks? That's obviously more complicated and maybe not worth worrying about, but just curious what happens when one or both of the methods return arrays instead of strings. Is that a possibility we need to worry about?

demiankatz:
(I see that getFullTitle is already returning an array in your test, but it contains only one value...)

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
ThoWagen: Added one for renderType "RecordDriverTemplate"

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/RecordDataFormatterTest.php
ThoWagen: The mock `getBuildings` already existed. I just added `getBuildingsAltScript` to test the behavior when the altScript method exists but does not return anything.

ThoWagen:
I added a test for a method that returns an array.

It's an interesting idea to match entries of arrays. But we don't have a use case for it and as you said it makes things more complicated. So I'd it can be added in a later PR if necessary.

module/VuFind/tests/unit-tests/src/VuFindTest/View/Helper/Root/RecordDataFormatterTest.php
demiankatz: Ahh, I see, should have read the comments a little higher up the page. Sorry! :-)

demiankatz (approved):
Thanks, @ThoWagen! I think this is safe to include in release 9.1, since it adds functionality without breaking anything. I have updated the [wiki documentation](https://vufind.org/wiki/development:architecture:record_data_formatter) (but please feel free to edit that if you can improve it), and I will merge this now.

PR #3156, opened 2023-10-09
Logging: Add ReferenceId support for username

maccabeelevine:
Log the username (if logged in) as part of each log message.  For example, to track who is responsible for excessive article downloading in violation of license agreements.


demiankatz (approved):
Works for me, and makes sense. Thanks!

PR #3157, opened 2023-10-09
VUFIND-1604: Multiple calls to parentTheme view helper

crhallberg:
Fixes a found error where multiple calls to the `parentTheme` view helper failed due to reuse of the template path stack.


demiankatz (approved):
Thanks, @crhallberg -- good catch!

PR #3158, opened 2023-10-10
Configurable similar items carousel item count

rajaro:


module/VuFind/src/VuFind/RecordTab/SimilarItemsCarousel.php
```diff
@@ -56,14 +56,25 @@ class SimilarItemsCarousel extends AbstractBase
...
      *
-     * @param \VuFindSearch\Service $search Search service
+     * @param \VuFindSearch\Service  $search Search service
+     * @param \Laminas\Config\Config $config Configuration
      */
-    public function __construct(\VuFindSearch\Service $search)
-    {
+    public function __construct(
+        \VuFindSearch\Service $search,
+        \Laminas\Config\Config $config
```
demiankatz: Might be worth defaulting this parameter to null to make it optional -- this would improve backward compatibility if anybody is using an outdated factory.

demiankatz (changes requested):
Thanks, @rajaro! This looks straightforward. If you make one minor update (see below) to address backward compatibility, I think we could safely include this in release 9.1.

module/VuFind/src/VuFind/RecordTab/SimilarItemsCarousel.php
rajaro: Done.

demiankatz (approved):

demiankatz:
Thanks, @rajaro!

PR #3159, opened 2023-10-11
Handle lookfor[] array in deferred search object recommendations.

demiankatz:
In reviewing server logs, I found a weird edge case where an external server was making a VuFind request with a lookfor[] array instead of a flat lookfor value, and this was causing errors in the deferred recommendations. This PR offers a workaround. I think that flattening the array into a space-delimited string is probably the closest we can manage to what the non-deferred version does through the search objects.

### DIFF
--- a/module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
+++ b/module/VuFind/src/VuFind/Recommend/AbstractSearchObjectDeferred.php
@@ -139,6 +139,11 @@ abstract class AbstractSearchObjectDeferred implements RecommendInterface
         $this->lookfor
             = $request->get(empty($settings[0]) ? 'lookfor' : $settings[0], '');
         $settings[0] = 'lookfor';
+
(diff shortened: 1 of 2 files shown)


maccabeelevine (approved):
The code itself looks fine.  I don't really agree with supporting an unintended type of a parameter, even if it's easy to do -- seems like a can of worms that would affect lots of parameters.  But I get that it's easy to handle in this case, and in general VuFind tries to be flexible that way.

demiankatz:
Thanks, @maccabeelevine -- in this instance, I believe VuFind 1.x used to allow this syntax, and we have some very old integrations that still use it -- which is how I encountered the bug! Other parts of the stack already handle it correctly, so I just fixed this bit for consistency.

maccabeelevine:
> Thanks, @maccabeelevine -- in this instance, I believe VuFind 1.x used to allow this syntax, and we have some very old integrations that still use it -- which is how I encountered the bug! Other parts of the stack already handle it correctly, so I just fixed this bit for consistency.

That makes sense!

PR #3160, opened 2023-10-11
Recommend databases based on result facets

maccabeelevine:
This introduces a recommendation module to recommend research databases based on EDS results.  It loops through the results list to aggregate a list of databases (`$record->getDbLabel()`).  Each listed database is a link to the individual database front-end.  

So, this is different than the facet that would narrow down the results to that database within VuFind.  Why: Instruction librarians may teach students to use a particular research database for DB-specific functionality, discipline-specific terminology, etc.  When a student doesn't know what database(s) to use, this allows VuFind to be the first stop and recommend those databases.

Unfortunately the EDS API does not provide an actual link to the database UI.  It can be inferred (using a code) if the database is hosted on EBSCOhost, but that's no guarantee.  So this module loads the institution's list of databases from LibGuidesAZ (via its API -- not the search box method used in the LibGuidesAZ recommendation module), caches it, and uses LibGuidesAZ data to associate the database with a URL.

TODO
- [x] After merging, open follow-up PR to normalize recommendation module heading level (currently they're a mix of h3/h4)


maccabeelevine:
I marked this for 10.0, given the new translation, and also the lack of time. :)

languages/en.ini
```diff
@@ -356,6 +356,8 @@ edit_list_fail = "Sorry, you are not permitted to edit this list"
 edit_list_success = "List successfully updated."
 Edition = "Edition"
 EDS Results = "EDS Results"
+eds_databases = "Search Relevant Databases"
+eds_databases_intro = "Search directly in databases for enhanced functionality."
```
maccabeelevine: This suggested language comes from an instruction librarian colleague.  But discussion welcome.  I asked her to consider how to concisely explain to patrons something that's not at all obvious:

- This is not a facet to narrow down these results, like the list of sources on the Articles tab.  This will actually open up the database.
- That said, the list of databases is still drawn from the EDS results to the left.
- Given that in theory all content of those databases is indexed by EDS and available here via the API, what (concisely) is the reason someone should exit to the database UI?

module/VuFind/src/VuFind/Recommend/EDSDatabases.php
```diff
@@ -0,0 +1,174 @@
...
+     */
+    public function getResults()
+    {
+        $records = $this->results->getResults();
+        $nameToDatabase = $this->getLibGuidesDatabases();
+        $databases = [];
+        foreach ($records as $record) {
+            $databaseInfo = $nameToDatabase[$record->getDbLabel()] ?? null;
+            if ($databaseInfo) {
+                $databases[$record->getDbLabel()] = $databaseInfo;
```
maccabeelevine: It is of course ridiculous to use the actual database name/label as the match point between EDS and LibGuides AZ, but without some custom configuration there doesn't seem to be any other option.  In practice things are not always named the same, and LibGuides AZ may not be complete.  One silver lining is that LibGuides has a second "alternate titles" field that I may use as well, to see if the EDS label is a substring of it.

module/VuFind/src/VuFind/Recommend/EDSDatabasesFactory.php
```diff
@@ -0,0 +1,77 @@
...
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options passed to factory.');
+        }
+        return new $requestedName(
+            $container->get(\VuFind\Connection\LibGuides::class),
+            $container->get(\VuFind\Config\PluginManager::class)
+                ->get('LibGuidesAPI'),
```
maccabeelevine: I need the LibGuides API config for caching duration.  It seems like I should also pass the EDS config "just in case" but it's not used for anything currently, so I didn't.

module/VuFind/src/VuFind/Connection/LibGuides.php
```diff
@@ -143,10 +186,21 @@ public function getAccounts()
...
+
+        return true;
+    }
+
+    /**
+     * Perform a GET request to the LibGuides API.
+     *
+     * @param string $url Full request url
+     *
+     * @return object|null A JSON object of the repsons data, or null if an error occurs
```
demiankatz: Typo: repsons.

module/VuFind/src/VuFind/Recommend/EDSDatabases.php
```diff
@@ -0,0 +1,174 @@
...
+     *
+     * @return array
+     */
+    public function getResults()
+    {
+        $records = $this->results->getResults();
+        $nameToDatabase = $this->getLibGuidesDatabases();
+        $databases = [];
+        foreach ($records as $record) {
+            $databaseInfo = $nameToDatabase[$record->getDbLabel()] ?? null;
```
demiankatz: Doesn't EDS support the database as a facet? If so, would it be better to use the facet data instead of iterating through records? Then you could recommend the most commonly-matched databases in the result set, instead of the potentially more arbitrary "whatever shows up on the current page."

module/VuFind/src/VuFind/Recommend/EDSDatabases.php
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Load or retrieve from the cache the list of LibGuides A-Z databases.
+     *
+     * @return array An array mapping a database name to the full object
+     * from the LibGuides /az API.
+     */
+    protected function getLibGuidesDatabases()
+    {
+        $nameToDatabase = $this->getCachedData('libGuidesAZ-nameToDatabase');
```
demiankatz: Seems like it would be pretty easy to fall back to an .ini file if LibGuides is unsupported. Maybe it's worth considering adding some generic functionality of that sort to make this more broadly useful for institutions not using LibGuides. Or maybe it's worth only using an .ini file and building separate tooling to create that .ini file using a LibGuides lookup, and you could refresh it with a cron job...

module/VuFind/src/VuFind/Recommend/EDSDatabasesFactory.php
demiankatz: You might also consider just passing the Config\PluginManager if you think you might need a variety of configs and want flexibility in the future.

demiankatz (changes requested):
A few comments and ideas...

module/VuFind/src/VuFind/Recommend/EDSDatabases.php
maccabeelevine: It does support database as a facet ("source" I think), and yes that would probably give better results.  But I believe that would require actually enabling that facet (and facets in the first place) for this recommendation module to get data from it -- which you almost definitely would not want to do, since then you would have two sidebar "lists" of databases with conflicting functionality.  (And in my case, this is for a bento solution where I don't want any facets.)  However this still may be a better approach, combined with just hiding the facts with CSS -- it would solve the underlying 'better results' problem and just waste some CPU cycles computing the other (unused, hidden) facets.

module/VuFind/src/VuFind/Recommend/EDSDatabases.php
maccabeelevine: > Seems like it would be pretty easy to fall back to an .ini file if LibGuides is unsupported. Maybe it's worth considering adding some generic functionality of that sort to make this more broadly useful for institutions not using LibGuides.

I liked this idea; implemented.  The .ini configuration either replaces or overrides LibGuides's list.  There may though be better ways to define that configuration, I don't love url[] and open to ideas.  I am thinking in the future there may be other metadata we would also want to configure about each database via EDS.ini, matching some of the other info coming from LibGuides, if it was useful in the template.

module/VuFind/src/VuFind/Recommend/EDSDatabases.php
maccabeelevine: > But I believe that would require actually enabling that facet (and facets in the first place) for this recommendation module to get data from it

On more digging I was happy to discover I was wrong on this.  I've re-implemented it to use the list of databases from the facets i.e. from the full results.

That has a side effect of listing way too many databases, so I will need to do another change to limit it to the first 'n' matches.

maccabeelevine: > That has a side effect of listing way too many databases, so I will need to do another change to limit it to the first 'n' matches.

Done

demiankatz (changes requested):
Thanks for the great progress on this... but now I'm wondering if we could make this even more generic and flexible. Is there really any reason this needs to be EDS-specific? Couldn't this now work with any search backend that returns a facet format consistent with EDS (e.g. Solr/Primo)? Obviously, as of right now, EDS may be the only system that provides a "Database" facet, but I imagine that might be a possibility elsewhere now or in future, and I can think of other possible use cases (e.g. use the high-level call number facets and map those to relevant subject-specific databases).

Broadening this to more use cases would require a few more relatively minor adjustments:

1.) Allow the facet field name to be specified somewhere (either directly in the recommendation module configuration line, or through a secondary setting somewhere).

2.) Allow the configuration file and section to be configured through the recommendation module configuration line.

3.) Rename a few things to get rid of "EDS".

4.) Add a new text domain for translating database names -- e.g. in the template, change `$this->escapeHtml($name)` to `$this->transEsc("DatabaseNames::$name")`. We wouldn't need to populate this with anything by default, but it would allow my "call number to database" idea to be implemented.

If I'm overlooking some reason why this isn't actually feasible, or if you think some parts of my idea are good and others are excessive, that's fine. I'm also open to merging this as-is. But I think if there's a reasonable opportunity to make this as reusable and flexible as possible, we might as well do that up front rather than refactoring later, if it's not too much work!

maccabeelevine:
> Thanks for the great progress on this... but now I'm wondering if we could make this even more generic and flexible. Is there really any reason this needs to be EDS-specific? 

@demiankatz I decided this made a lot of sense.

> 1.) Allow the facet field name to be specified somewhere (either directly in the recommendation module configuration line, or through a secondary setting somewhere).
> 3.) Rename a few things to get rid of "EDS".

Done.

> 2.) Allow the configuration file and section to be configured through the recommendation module configuration line.

I did this for the configuration file.  I don't see a real need to make the section name itself a configuration option vs the complication that adds to the code.

> 4.) Add a new text domain for translating database names -- e.g. in the template, change `$this->escapeHtml($name)` to `$this->transEsc("DatabaseNames::$name")`. We wouldn't need to populate this with anything by default, but it would allow my "call number to database" idea to be implemented.

I didn't do this, not that it's a bad idea, but I think it's a change we could make in the future if someone needed it that would be non-breaking.

config/vufind/EDS.ini
```diff
@@ -323,3 +323,27 @@ AU = None
 ; documentation on options for different adapters.
 ; ttl, Time to Live, i.e. cache entry life time in seconds. 300 seconds by default.
 ;options[ttl] = 300
+
+; This section supports the Databases recommendation module,
+; which displays a list of the results' databases, each linking
+; to that database's individual website.
+[Databases]
```
maccabeelevine, line 348: With the generalization beyond EDS.ini, this boilerplate could technically be in any backend .ini file.  But this still seems reasonable?

demiankatz:
Thanks! I'll give this a closer look tomorrow, and get it merged then if it's fully ready.

> > 2.) Allow the configuration file and section to be configured through the recommendation module configuration line.
> 
> I did this for the configuration file. I don't see a real need to make the section name itself a configuration option vs the complication that adds to the code.

Makes sense!

> > 4.) Add a new text domain for translating database names -- e.g. in the template, change `$this->escapeHtml($name)` to `$this->transEsc("DatabaseNames::$name")`. We wouldn't need to populate this with anything by default, but it would allow my "call number to database" idea to be implemented.
> 
> I didn't do this, not that it's a bad idea, but I think it's a change we could make in the future if someone needed it that would be non-breaking.

Agreed -- also an easy local template customization if there is a need for it.

config/vufind/EDS.ini
demiankatz: The recommendation module itself is documented in searches.ini. If we keep these settings here, we should make sure that documentation has a reference to this so it can be found. I think the searches.ini documentation also needs to be updated to include the setting to specify the config file.

maccabeelevine: Thanks, I updated the seraches.ini description to include that parameter and reference EDS.ini.

maccabeelevine:
> and get it merged then if it's fully ready.

Yes there's one more enhancement I want to do (so far) but I think it's best to implement in a separate PR.

module/VuFind/src/VuFind/Connection/LibGuides.php
```diff
@@ -122,14 +122,57 @@ public function __construct(
...
+        if (isset($result->errorCode)) {
+            return null;
+        }
+        return $result;
+    }
+
+    /**
+     * Authentiate to the LibGuides API and set authentiation headers.
+     *
+     * @return bool Indicates if authentiation succeeded.
```
demiankatz: Typo: authentiation

themes/bootstrap3/templates/Recommend/Databases.phtml
```diff
@@ -0,0 +1,21 @@
+<div class="databases">
+  <?php $databases = $this->recommend->getResults(); ?>
+  <?php if (!empty($databases)): ?>
+    <h4><?=$this->transEsc('databases')?></h4>
```
demiankatz, line 4: Any particular reason you chose an h4 here? A quick glance reveals some weird inconsistencies in our recommendation module heading levels, so I imagine you copied this from some existing template, but maybe it's not the best option -- it may be a whole separate project to straighten that out. (I notice, for example, that SummonResultsDeferred has an h3 that changes into an h4 when SummonResults is loaded... that can't be intentional).

config/vufind/searches.ini
```diff
@@ -251,6 +251,10 @@ WorkKeys = year
 ;       Display catalog search results matching the terms found in the specified
 ;       GET parameter (default = "lookfor"), limited to a specified number of
 ;       matches (default = 5).  This is designed for use with non-catalog modules.
+; Databases:[result limit]:[ini name]
```
demiankatz, line 265: Maybe also worth adding an example of this (commented out) to the `default_side_recommend` settings in EDS.ini, so the example there is more complete.

module/VuFind/src/VuFind/Recommend/Databases.php
```diff
@@ -0,0 +1,275 @@
...
+     */
+    public function setConfig($settings)
+    {
+        $settings = explode(':', $settings);
+        $this->limit
+            = (isset($settings[0]) && is_numeric($settings[0]) && $settings[0] > 0)
+            ? intval($settings[0]) : 5;
+        $databasesConfigFile = $settings[1] ?? 'EDS';
+
+        $databasesConfig = $this->configManager->get($databasesConfigFile);
```
demiankatz: I wonder if it's worth creating a shortcut variable to $databasesConfig->Databases so you can shorten all your various configuration references below. Not required, just a thought for readability.

module/VuFind/src/VuFind/Recommend/Databases.php
```diff
@@ -0,0 +1,275 @@
...
+    public function getResults()
+    {
+        if (count($this->resultFacet) < 1) {
+            throw new \Exception('At least one facet key required.');
+        }
+
+        $resultDatabasesTopFacet = array_shift($this->resultFacet);
+        $resultDatabases = $this->results->getFacetList([$resultDatabasesTopFacet => null])[$resultDatabasesTopFacet];
+        while (count($this->resultFacet) && $resultDatabases) {
+            // Throw exception if the key is not found
```
demiankatz: You have some comments about throwing exceptions, but I don't see any exceptions being thrown (unless you're referring to the native PHP errors caused by accessing undefined indexes).

demiankatz (changes requested):
Thanks again, @maccabeelevine -- I didn't end up having time to do hands-on testing today, but I did find a few more things to comment on in another eyeball review of the code. I think once these things are addressed (or you decide not to address them), it should be ready for a final hands-on review.

config/vufind/EDS.ini
maccabeelevine: Yup searches.ini documentation does include the .ini file setting and references the example in EDS.ini.

module/VuFind/src/VuFind/Connection/LibGuides.php
maccabeelevine: Somehow I managed to make this same typo three times in the same method doc!!!  Fixed.

themes/bootstrap3/templates/Recommend/Databases.phtml
maccabeelevine: Yeah I don't recall which template I copied this from, but it looks like the majority (12/17) of recommendation headers are `<h4>`.  I agree there should be a separate PR to sort this out, and to figure out what heading level they should actually be.

image

config/vufind/searches.ini
maccabeelevine: Done

module/VuFind/src/VuFind/Recommend/Databases.php
maccabeelevine: Yes agreed, I simplified those references and threw an earlier exception if the section is not found.

module/VuFind/src/VuFind/Recommend/Databases.php
maccabeelevine: Yes I was referring to the native exceptions -- I thought they were reasonable to let bubble up here; I don't think any defaults / fallbacks are helpful.  If it's better I can throw explicit exceptions.

themes/bootstrap3/templates/Recommend/Databases.phtml
demiankatz: I can see how the `h2` in SideFacets is usually the topmost heading, and then some of the other `h3`'s might be seen as subdivisions of that. But I really don't see how we get to `h4` at all. In any case, though, it is fine to keep this as-is to follow the predominant example, and we can straighten it out later. I've added a TODO as a reminder to open a follow-up after this is done. (I'm happy to do something to get the ball rolling when the time comes).

module/VuFind/src/VuFind/Recommend/Databases.php
demiankatz: My main concern is that the behavior may not be predictable depending on the error_reporting settings in php.ini. It's possible that this will throw a fatal error when you're in development mode (with zero tolerance for PHP notices) but this will silently ignore the error in production mode (when the settings are generally more error-tolerant). I don't think that's the desired behavior! But maybe I'm misremembering what counts as a notice and what doesn't, so feel free to correct me if you've already tested this and I'm wrong. :-)

themes/bootstrap3/templates/Recommend/Databases.phtml
maccabeelevine: Yeah whatever heading level it ends up after that other PR with is fine.

module/VuFind/src/VuFind/Recommend/Databases.php
maccabeelevine: Huh no I didn't think of that; have been testing in production mode and looking at the logs.  I'm honestly not sure the right approach, since in my experience VuFind seems to use defaults to avoid exceptions, and I don't think that's right here.  Not sure how best to handle it.

config/vufind/EDS.ini
```diff
@@ -323,3 +324,27 @@ AU = None
...
+
+; Define the facet hierarcy path leading to each database.
```
demiankatz: Typo: hierarcy. (This has to be one of the worst words to type correctly...)

module/VuFind/src/VuFind/Recommend/Databases.php
demiankatz: I guess the question is, what's the desired behavior and intended goal here? Is the idea that if the user configures an invalid path to facet data, it causes a fatal error? If so, is that really desirable? I'm not sure that it would actually make this any easier to troubleshoot. Might it be better if we use null coalescing to treat missing data as empty so that it will be ignored without exceptions, and perhaps add some debug logging in cases where an expected configured key is undefined?

demiankatz (changes requested):
In addition to continuing our conversation below, I also happened to spot another small typo... :-)

config/vufind/EDS.ini
maccabeelevine: Indeed.  Fixed.

module/VuFind/src/VuFind/Recommend/Databases.php
maccabeelevine: Ok that's a good way to look at it -- I think it's better to log an error than break everything.  Presuming anyone is paying attention to the log output. :)  I am not using null coalescing here though because there is no default handling of these keys not being found that should not cause at least a log message; in theory I could coalesce to null, check for null, and log the error, but using try/catch makes it clearer IMHO.

maccabeelevine:
>I guess the question is, what's the desired behavior and intended goal here? Is the idea that if the user configures an invalid path to facet data, it causes a fatal error? If so, is that really desirable? I'm not sure that it would actually make this any easier to troubleshoot. Might it be better if we use null coalescing to treat missing data as empty so that it will be ignored without exceptions, and perhaps add some debug logging in cases where an expected configured key is undefined?

GitHub seems to have hid my response to this after I changed the code enough that it can't locate where the comments apply.  In short, I agree with not breaking the app and I changed this to use logging instead.  Not using null coalescing though because I would need to check for the coalesced result anyway, I think exception handling is cleaner for this situation.

module/VuFind/src/VuFind/Recommend/DatabasesFactory.php
```diff
@@ -0,0 +1,76 @@
...
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options passed to factory.');
+        }
+        return new $requestedName(
+            $container->get(\VuFind\Config\PluginManager::class),
+            $container->get(\VuFind\Connection\LibGuides::class),
```
demiankatz: I ran into an interesting problem when testing this -- because I'm not using LibGuides, I get a fatal error when I turn on this recommendation module, because the LibGuides constructor complains that LibGuides is not configured correctly. But it would be silly to force users to configure LibGuides when they're not actually using it.

The solution might be to replace the direct LibGuides dependency in the constructor with a callable getter. Then you could set up something like this in the factory:

```
$libguidesGetter = function () use ($container) {
    return $container->get(\VuFind\Connection\LibGuides::class);
}
```

...and your recommendation module can call this function to get the LibGuides object only if it actually needs it.

demiankatz (changes requested):
The code is starting to look pretty solid to me, so I started hands-on testing... and immediately ran into a problem! :-) See details below...

module/VuFind/src/VuFind/Recommend/DatabasesFactory.php
maccabeelevine: Thanks, that's a silly miss on my part.  Fixed as you suggested.

demiankatz (approved):
Thanks, @maccabeelevine -- I pushed up one small commit to add some missing styles so the icon/link combinations in your template display correctly (otherwise, there's an underlined space between the icon and the text). Other than that, this looks good to me, so I'm going to merge it now!

PR #3161, opened 2023-10-11
Add support for OCLC numbers to Google cover images and improve medium/large images

damien-git:
Current Google cover image implementation only supports ISBN numbers, but the API supports OCLC numbers, so we can use that support. It makes a difference for records that only have an OCLC number.
Also, the zoom in the image URL is modified to get a larger image when the requested size is not small.


demiankatz (approved):
Thanks, @damien-git -- these are useful enhancements! I've adjusted the existing test class to cover most of the new functionality.

damien-git:
@demiankatz Thanks !

damien-git:
@demiankatz I am not seeing a release tag. Will this be in 9.1 with other cover image improvements ?

demiankatz:
Yes, it will, @damien-git -- tag added. (Usually I only tag things if they stay open long enough for me to need to keep track of them...)

PR #3162, opened 2023-10-17
Only update FOLIO location cache data when appropriate.

demiankatz:
I happened to notice this bug while reviewing #3145. We don't want to write data to the cache if that data already came from the cache. This is not only inefficient but also potentially prevents appropriate cache expiration.


maccabeelevine (approved):
Good catch -- with a request every 30 seconds it would have kept using the stale data forever.  Tested that the fix works.

PR #3164, opened 2023-10-17
FOLIO: do not display credentials in debug messages.

demiankatz:
This is intended to reduce the chances of accidental disclosure of sensitive information.


meganschanz:
I was excited to see this change! This was a ticket we had in our backlog for a while now to work on and submit a PR for but never got around to it unfortunately. I tested this change in our environment and it worked perfectly. It successfully suppressed the credentials from being shown on the page when debug was true and in the logs when debug logging was enabled.

crhallberg (approved):

PR #3170, opened 2023-10-20
Fix escaping of backslashes in ExtendedIniNormalizer.

EreMaijala:


EreMaijala:
Perhaps a test would be useful, I'll add one.

demiankatz:
Thanks, @EreMaijala -- I've also added comments to this whole function for future reference. I'll wait until your test is added before approving and merging, but this looks good to me.

demiankatz:
Also, should we rebase against release-9.1?

EreMaijala:
@demiankatz Either way works for me. But this could be considered a bug. I added a test, but I'm not entirely happy with how confusing the strings look in the test.

demiankatz (approved):
Looks good to me, @EreMaijala -- while the test examples are a bit painful to look at, I think some pain is unavoidable when dealing with escaping. At least it's clear that the test is round-tripping the data successfully.

demiankatz:
Manually backported to the release-9.1 branch (and merged from there to dev).

PR #3171, opened 2023-10-23
Add support for a description column in fines.

EreMaijala:
Includes driver support for Demo and SierraRest. SierraRest also removes some custom handling of types and adds optional mappings. Finnish and Swedish translations include a fix for duplicate column heading.

TODO:
- [x] Update [ILS drivers wiki page](https://vufind.org/wiki/development:plugins:ils_drivers)


module/VuFind/src/VuFind/ILS/Driver/Demo.php
```diff
@@ -1044,7 +1051,9 @@ public function getMyFines($patron)
                     'createdate' => $this->dateConverter
                         ->convertToDisplayDate('U', time()),
                     // After 20 days it becomes 'Long Overdue'
-                    'fine'     => $day_overdue > 20 ? 'Long Overdue' : 'Overdue',
+                    'fine'     => $type,
```
demiankatz, line 1054: I think the comment above has gotten orphaned and needs to move up.

config/vufind/SierraRest.ini
```diff
@@ -211,6 +211,13 @@ title_hold_excluded_item_codes = "e"
 ;a = patron_status_address_missing
 ;b = patron_status_card_blocked
 
+; This section allows modification of the default mapping from Sierra fine types to
+; VuFind fine types. By default fine types are displayed as is. Examples are listed
+; below.
+[FineTypeMappings]
+;Manual = ""
+;Overdue Renewal = "Overdue"
```
demiankatz, line 219: Would it make sense to uncomment this by default to retain the historical behavior, or is there a reason we no longer want that? (No strong feelings about the matter -- just asking).

demiankatz (changes requested):
Thanks, @EreMaijala -- one small question and one minor issue to fix.

config/vufind/SierraRest.ini
EreMaijala: I opted to remove the defaults since they came from us and the changes were requested by the Sierra library. Also hiding the "Manual" type no longer makes so much sense now that it has its own column.

EreMaijala: And for the "Overdue Renewal" I've no recollection why it was done in the first place since similar mappings are not used anywhere else, and the library specifically asked not to do that since you lose some information that might be of value.

demiankatz (changes requested):
The code looks great, and the output is also fine in most themes. However, this happens in bootprint3:

image

Might there be an easy CSS fix to reduce the odds of this happening? I'm not sure why the browser isn't deciding to make the "Fee" column a little smaller so that it doesn't have to wrap the "Description" heading.

demiankatz:
Lacking a smarter solution, we could also abbreviate the English translation to "Desc." -- but then I imagine we'll still see similar problems for other languages with headings of similar length.

demiankatz:
...or maybe we should just widen the default bootprint3 width, since it's feeling pretty cramped these days. If that's our best solution, it might be worth addressing in a separate PR and just living with this as-is for now.

EreMaijala:
@demiankatz There are these rules in bootstrap.less that have been there since dawn of the file:
```
/* ------ Table wrapping to prevent horizontal overflow ------ */
.table {
  table-layout: fixed;
  word-wrap: break-word;
}
```
I understand that particularly letting words wrap in the middle helps fit the table to narrow screens as well, though I think we should try to avoid that since the result is quite horrible. The downside of the fixed layout is that it practically makes all columns equal width for a table with no explicit widths. I added a rule now to override this for the fines table on wide screens.

EreMaijala:
Also note that this is not really a new issue. Even without the description column we have some issues displaying the table in the minimum 768px width with dates wrapping in a not very nice way.

EreMaijala:
Example of how the headings wrap in Finnish:
kuva

demiankatz (approved):
Thanks, @EreMaijala, this is much better. I'm not sure in your screen shot why the statistics on the ILL link are overflowing the page, though; I'm not seeing the same behavior on my end.

In any case, I'll merge this as-is, but we should do follow-up work to improve our tables generally -- by adding more responsive styling and/or by eliminating tables where other HTML elements might be more useful and appropriate. I spoke to @crhallberg about this a little bit; perhaps he can make some proposals in a JIRA ticket or PR in the near future.

PR #3172, opened 2023-10-23
Translate sysMessage of renewals.

EreMaijala:
sysMessage is translated almost everywhere, but translation was missing for renewal results, which caused translation prefix at least in SierraRest driver to be displayed on screen.

Also fixes an unexpected trailing period from Swedish translation of renew_fail message.


demiankatz (approved):
Thanks, @EreMaijala!

PR #3175, opened 2023-10-23
Accessibility: Use $headTitle as <h1> on results

maccabeelevine:
On boostrap3 theme there seems to be no `<h1>` element on the search results page.  But there are plenty of `<h2>`s -- each search result title, plus more in the footer or if you have facets enabled.  This is I think a clear accessibility problem.

I'm proposing that the search results title -- effectively the breadcrumb -- be the `<h1>`.  I'm writing it out separately in a `sr-only` element so it doesn't affect layout.

### DIFF
--- a/themes/bootstrap3/templates/combined/results.phtml
+++ b/themes/bootstrap3/templates/combined/results.phtml
@@ -51,6 +51,7 @@
   $this->headLink()->appendStylesheet('combined-search.css');
 ?>
 <?=$this->flashmessages()?>
+<h1 class="sr-only"><?=$this->escapeHtml($headTitle)?></h1>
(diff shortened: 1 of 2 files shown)


demiankatz:
Makes sense to me, but I'd like to get input from @crhallberg before I approve, in case there's some wrinkle I'm not thinking of.

maccabeelevine:
I added support for `combined` search results.  Everything else I checked (EDS, blender) was handled by the `search` change.

crhallberg (approved):

crhallberg:
Thank you for your contribution and your patience!

demiankatz:
Thanks, @maccabeelevine and @crhallberg!

PR #3177, opened 2023-10-24
Remove obsolete basemap examples; add more notes.

demiankatz:
This pull request removes the Stamen map tile examples, as they are going offline later this month. It also adds a note about licensing and alternative options, with a link to the [Geographic Features](https://vufind.org/wiki/configuration:geographic_features) wiki page. I have updated that page to make it more readable, and to remove a lot of version-specific notes that are very old and likely serve no further purpose. I'm hoping that @EreMaijala and/or @lmgonzales can add a section to the page covering map tiles and map servers, but I wrote my revisions here so that we should be able to merge this whether or not we make other changes there.

TODO
- [x] Add changelog note about stamen map tiles.

### DIFF
--- a/config/vufind/geofeatures.ini
+++ b/config/vufind/geofeatures.ini
@@ -8,10 +8,20 @@
 [Basemap]
 ; Configures the basemap that is used to display geographic featuers.
 ; Default basemap configuration is the osm-intl option below, and other
-; open source basemap options are provided.
+; open source basemap options are provided as examples.
+;
+; IMPORTANT: most free map services have significant limitations on
+; usage, so you should consult their terms of service. If you are going to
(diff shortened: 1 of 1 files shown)


demiankatz:
I've added a stub [Notes on Map Tiles](https://vufind.org/wiki/configuration:geographic_features#notes_on_map_tiles) section where we can add more details as time permits.

lmgonzales:
Thanks for updating the .ini file @demiankatz with the notes to the Wiki page. I'll work on updating the wiki pages today.

lmgonzales (approved):

demiankatz:
Thanks, @lmgonzales -- I see that you figured out the approval process. (It's a bit hidden away until you get used to it!)

PR #3179, opened 2023-10-25
Improve stability of testAddCommentWithCaptcha test.

EreMaijala:
I've seen intermittent retries from the `testAddCommentWithCaptcha` test. This should address the timing issue.


demiankatz (approved):
Thanks, @EreMaijala, I've been noticing this too but hadn't gotten around to investigating. I appreciate the fix!

PR #3180, opened 2023-10-25
[9.1] Use null coalescing to cover for invalid active tab.

EreMaijala:
This avoids a PHP notice if you try to use a tab that doesn't exist for the record (e.g. /Record/[id]/Foo). In practice this could happen e.g. if user has a link to a tab that's been disabled.


demiankatz (approved):
Thanks, @EreMaijala!

PR #3184, opened 2023-10-27
Fix raw type warnings

maccabeelevine:
This just resolves a bunch of compiler warnings.  All the types seemed obvious from the surrounding code, and don't generate any compiler errors when run.


demiankatz (approved):
Thanks, @maccabeelevine, this all makes sense to me, and the test suite is still passing.

PR #3185, opened 2023-10-27
Fix Java 9 deprecation warnings

maccabeelevine:
Just because one of these was polluting my solr import log file.

### DIFF
--- a/import/index_java/src/org/vufind/index/CallNumberTools.java
+++ b/import/index_java/src/org/vufind/index/CallNumberTools.java
@@ -325,7 +325,7 @@ public class CallNumberTools
                 float currentVal = Float.parseFloat(callNum.getClassification());
 
                 // Round the call number value to the specified precision:
-                Float finalVal = new Float(Math.floor(currentVal / precision) * precision);
+                Float finalVal = Double.valueOf(Math.floor(currentVal / precision) * precision).floatValue();
 
                 // Convert the rounded value back to a string (with leading zeros) and save it:
                 // TODO: Provide different conversion to remove CallNumUtils dependency
--- a/import/index_java/src/org/vufind/index/DatabaseManager.java
+++ b/import/index_java/src/org/vufind/index/DatabaseManager.java
@@ -94,7 +94,7 @@ public class DatabaseManager
                 prefix = "postgresql";
             }
 
-            Class.forName(classname).newInstance();
+            Class.forName(classname).getDeclaredConstructor().newInstance();
             String[] parts = dsn.split("://");
             if (parts.length > 1) {
                 parts = parts[1].split("@");
--- a/import/index_java/src/org/vufind/index/UpdateDateTracker.java
+++ b/import/index_java/src/org/vufind/index/UpdateDateTracker.java
@@ -149,7 +149,6 @@ public class UpdateDateTracker
         insertSql.close();
         selectSql.close();
         updateSql.close();
-        super.finalize();
     }
 
     /* Get the first indexed date (IMPORTANT: index() must be called before this method)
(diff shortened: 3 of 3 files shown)


import/index_java/src/org/vufind/index/CallNumberTools.java
```diff
@@ -325,7 +325,7 @@ public Set<String> getDeweyNumber(Record record, String fieldSpec, String precis
                 float currentVal = Float.parseFloat(callNum.getClassification());
 
                 // Round the call number value to the specified precision:
-                Float finalVal = new Float(Math.floor(currentVal / precision) * precision);
+                Float finalVal = Double.valueOf(Math.floor(currentVal / precision) * precision).floatValue();
```
maccabeelevine, line 328: Math.floor returns a double, so this appears to be an appropriate way to convert it to float.

import/index_java/src/org/vufind/index/DatabaseManager.java
```diff
@@ -94,7 +94,7 @@ private void connectToDatabase()
                 prefix = "postgresql";
             }
 
-            Class.forName(classname).newInstance();
+            Class.forName(classname).getDeclaredConstructor().newInstance();
```
maccabeelevine, line 97: Explanation: https://docs.oracle.com/javase/9/docs/api/java/lang/Class.html#newInstance--

import/index_java/src/org/vufind/index/UpdateDateTracker.java
```diff
@@ -149,7 +149,6 @@ protected void finalize() throws SQLException, Throwable
         insertSql.close();
         selectSql.close();
         updateSql.close();
-        super.finalize();
```
maccabeelevine: "The finalization mechanism is inherently problematic."
https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/Object.html#finalize--

Arguably instead of just removing the call to Object.finalize() here, the entire method should be removed and reworked to close the connections in a better way.  Fixing this Java 9 issue in some sense hides a larger problem.  But that's a larger research effort IMHO and maybe is worth a Jira.

demiankatz: The documentation you shared suggests using a Cleaner, which seems like it might be a viable replacement. Is it worth trying that here and replacing the entire finalize method, rather than just eliminating the call to super, which could be misleading? See:

https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Cleaner.html

If you think we should do that work separately, maybe we should just leave the super.finalize() call in place here until we have time to do that other work.

demiankatz (changes requested):
Thanks for the effort on this -- see some thoughts below on the database finalization.

maccabeelevine:
> If you think we should do that work separately, maybe we should just leave the super.finalize() call in place here until we have time to do that other work.

Yeah I started to look into Cleaner and it seems it will take some time to learn.  Removing that one for now, at least we'll be down to a single Java deprecation for now.

demiankatz (approved):
I've specifically tested Dewey and change tracker functionality, and everything seems fine with the revised code.

Thanks again, @maccabeelevine!

Do you plan to follow up with the Cleaner revisions in the near future, or should we open a JIRA ticket so we don't lose track of it?

maccabeelevine:
> Do you plan to follow up with the Cleaner revisions in the near future, or should we open a JIRA ticket so we don't lose track of it?

Jira please, just in case.  Won't be the next week or two at least.

demiankatz:
> Jira please, just in case. Won't be the next week or two at least.

Done: [VUFIND-1651](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1651)

PR #3188, opened 2023-11-01
Better config validation for XCNCIP2 ILS driver

xmorave2:


demiankatz (approved):
Looks good to me -- thanks, @xmorave2! Might be worthwhile in future to add a test case or two to cover this (since it wouldn't be hard to do), but I won't hold up the merge because of that.

PR #3189, opened 2023-11-02
Optional dataMethod params for RecordDataFormatter

ThoWagen:
Adding the option to add parameters when calling dataMethods in RecordDataFormatter.

I.e. this can be used to call `getAllSubjectHeadings` with `$extended` set to `true`.

TODO:
- [x] Update wiki documentation and changelog when merging


demiankatz (changes requested):
Thanks, @ThoWagen -- this makes sense to me! I think it's worth adding a test to exercise the new feature, though. Do you have time to do that? If not, I'm willing to help -- I don't think it should be too difficult to just add a line to the existing test that uses the new option and has an assertion to confirm the parameter is passed as expected.

I'm also going to add a TODO to update the wiki documentation when this is merged. I'll take care of that when the time comes!

ThoWagen:
I added a test.

demiankatz (approved):
Thanks, @ThoWagen!

PR #3190, opened 2023-11-03
Sierra: Add support for statistic group in holds and renewals.

EreMaijala:
Sierra 6.0 adds support for statgroup parameter that can be used to specify the "statistic group" to use in Sierra for specific functions (renewal, holds, payments).

### DIFF
--- a/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
+++ b/module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
@@ -288,6 +288,13 @@ class SierraRest extends AbstractBase implements
     protected $apiBase = 'v5';
 
     /**
+     * Statistic group to use e.g. when renewing loans or placing holds
+     *
+     * @var ?int
+     */
+    protected $statGroup = null;
+
+    /**
      * Whether to sort items by enumchron. Default is true.
      *
      * @var array
@@ -467,6 +474,13 @@ class SierraRest extends AbstractBase implements
             // Default to API v5 unless a lower compatibility level is needed.
             if ($this->apiVersion < 5) {
                 $this->apiBase = 'v' . floor($this->apiVersion);
+            }
+        }
+        if ($statGroup = $this->config['Catalog']['statgroup'] ?? null) {
+            if ($this->apiVersion >= 6) {
+                $this->statGroup = (int)$statGroup;
+            } else {
+                $this->logWarning("Ignoring statgroup for API Version {$this->apiVersion}");
             }
         }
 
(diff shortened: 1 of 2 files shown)


EreMaijala:
Oops, renew is not right, should be a query param.

EreMaijala:
Should be fixed now, but I'll need to verify and can't renew the test loan again today.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -1898,7 +1921,11 @@ protected function getApiUrlFromHierarchy(array $hierarchy): string
     {
         $url = $this->config['Catalog']['host'];
         foreach ($hierarchy as $value) {
-            $url .= '/' . urlencode($value);
+            if (is_array($value)) {
+                $url .= (str_contains($url, '?') ? '&' : '?') . http_build_query($value);
```
demiankatz: I'm not sure I understand the purpose of this -- does it really make sense to add array support to the hierarchy array rather than introducing another parameter for these types of things? Would it really work/make sense to pass an array with multiple arrays, or with an array before strings? If those scenarios wouldn't work, maybe this should be approached in a different way.

demiankatz (changes requested):
One question for you. In any case, I'll wait for an update on your testing before I approve anything.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: The whole thing of having query parameters in the URL is not a recommended practice. You're right though, this could be used to create bogus url's, but I couldn't come up with anything that wouldn't get confusing for GET requests without breaking back-compatibility.

EreMaijala: If you think it's better to add an "additional GET parameters" (that would get merged with $params for GET requests), I could go with that.

demiankatz: I think the additional GET parameters approach feels cleaner and would reduce the chances of invalid URL generation... but if it requires a significant amount of extra infrastructure to make it work (e.g. changing lots of method signatures and/or revising large amounts of code), I could live with this as-is since its scope of use is very limited.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -1799,6 +1824,11 @@ protected function requestCallback(
 
         // Set up the request
         $apiUrl = $this->getApiUrlFromHierarchy($hierarchy);
+        // Add additional query parameters directly to the URL becuase they cannot be
```
demiankatz: Typo: becuase

demiankatz (changes requested):
Thanks, @EreMaijala, I think this is safer and more clear! Just one tiny nitpick. Happy to approve and merge this on Monday if your testing comes out successfully.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
EreMaijala: Fixed!

EreMaijala:
After testing it appears that this did not work as expected in Sierra. I'm closing this now. We can always revisit in future if necessary.

EreMaijala:
After further inspection it seems that we will need this after all.

demiankatz (approved):
I can't test with Sierra, but it looks fine to me. Thanks, @EreMaijala!

PR #3192, opened 2023-11-03
Support filters and custom heading on recommendation modules

maccabeelevine:
This adds a hiddenFilter capability on recommendation modules that derive from AbstractSearchObject.  It also adds an optional custom heading.

I have a few specific use cases in mind:
- The primary search results are from a tab with a hidden filter -- say a media-specific Solr search.  An EDSResults sidebar could show media-specific results from EDS (or pick your other backend).
- The primary search results are blended, but filters narrow down the results from one or more backend.  A recommendation module could search the same backend but apply a different filter.
- The primary search results are combined, and include several Solr bentos, each filtered on "high use" content like books or media.  As above, a CatalogResults recommendation module called something like "Additional Sources" could be filtered on the combined NOT of all of the bentos, i.e. all the other resources no one will actually use, that we don't want in the more prominent bento real estate.

As a visual for demonstration only, here's an EDS primary results that also has a CatalogResults recommendation module that I've randomly filtered on videos.

image

TODO
- [ ] Add `:[hidden filters key]:[custom heading]` to the rest of the affected modules in searches.ini
- [ ] Add deprecation notes to changelog and JIRA about InjectSearchRunnerFactory when merging


config/vufind/searches.ini
```diff
@@ -254,10 +254,14 @@ WorkKeys = year
...
 ;
-; CatalogResults:[GET parameter]:[result limit]
+; CatalogResults:[GET parameter]:[result limit]:[hidden filters key]:[custom heading]
 ;       Display catalog search results matching the terms found in the specified
 ;       GET parameter (default = "lookfor"), limited to a specified number of
-;       matches (default = 5).  This is designed for use with non-catalog modules.
+;       matches (default = 5). This is designed for use with non-catalog modules.
+;       Hidden filters key, if defined, is a reference to hidden filters for these
+;       results, defined in the [AbstractSearchObjectHiddenFilters] section of the
+;       searches.ini file. The optional custom heading overrides the default for this
```
maccabeelevine: I know it's weird to refer to "the searches.ini" file within itself.  But as the earlier note says CatalogResults is more likely to be used in another config file, while the [AbstractSearchObjectHiddenFilters] would still be defined here.

config/vufind/searches.ini
```diff
@@ -861,3 +865,12 @@ maxLimit = 100
...
 ;options[ttl] = 300
 
+; This section provides hidden filters for AbstractSearchObject recommendation
+; module whose backend search behavior is configured in this file.
+[AbstractSearchObjectHiddenFilters]
+
+; Hidden filters are defined as in [RawHiddenFilters], i.e. each is an array in
+; order to handle more complex filters.  The key must match a [hidden filters key]
+; in a recommendation module, defined in the configuration file for the primary
+; search results.
```
maccabeelevine: I cannot manage to write this in a way that makes sense.  I know it's Friday afternoon but still.  If you understand it maybe you can suggest some better wording.  If not I will explain.

module/VuFind/src/VuFind/Recommend/InjectSearchRunnerFactory.php
```diff
@@ -70,6 +70,10 @@ public function __invoke(
         if (!empty($options)) {
             throw new \Exception('Unexpected options passed to factory.');
         }
-        return new $requestedName($container->get(SearchRunner::class));
+        $module = new $requestedName($container->get(SearchRunner::class));
+        if (method_exists($module, 'setConfigManager')) {
```
maccabeelevine: First off, there may be a better way to make the config manager available within the recommendation module.  Even if not, `method_exists` seems ugly but I'm not sure what is best.  Maybe an interface that I check with instanceof, and define the setter in a trait?  But that seems overkill and I couldn't come up with a middle ground.

themes/bootstrap3/templates/Recommend/CatalogResults.phtml
```diff
@@ -1,9 +1,10 @@
 <?php
   $searchObject = $this->recommend->getResults();
   $results = $searchObject->getResults();
+  $heading = $this->recommend->getCustomHeading() ?? 'Catalog Results';
 ?>
 <?php if (!empty($results)): ?>
-  <h4><?=$this->transEsc('Catalog Results')?></h4>
+  <h4><?=$this->transEsc($heading)?></h4>
```
maccabeelevine: The custom heading bit could be extracted to a separate PR if needed.

config/vufind/searches.ini
```diff
@@ -254,10 +254,14 @@ WorkKeys = year
 ;
 ; Available modules recommended for use in the side area:
 ;
-; CatalogResults:[GET parameter]:[result limit]
+; CatalogResults:[GET parameter]:[result limit]:[hidden filters key]:[custom heading]
```
maccabeelevine: [hidden filters key] would work in any AbstractSearchObject recommendation module.  [custom heading] would also if I update the templates.  I don't know how far to go documenting that for each of these recommendation modules.  Every AbstractSearchObject recommendation module already has a lot of duplicate explanation (which makes sense since  they're near-identical implementations).

config/vufind/searches.ini
demiankatz: I don't have a problem with referring to searches.ini here, though I guess if you wanted to avoid that you could say something more technical about using the search configuration file configured in the Options class (since that's technically more accurate if somebody has overridden that for some reason...).

config/vufind/searches.ini
demiankatz: I think it's worth applying similar language to all the other modules in this family, though I don't mind waiting until everything is finalized first since the details might change in the meantime. If you prefer to wait, we should put a TODO checkbox at the top of the PR to keep track of this item.

config/vufind/searches.ini
demiankatz: Would this be less confusing if you specified an .ini section name in the recommendation module configuration, instead of having sections inside a section? That is, what if rather than `video[]` inside `[AbstractSearchObjectHiddenFilters]`, you just did `[CatalogResultsVideoFilter]`? Then you could just document it as an example, and reference the example from the module documentation, but there's nothing hard-coded or deeply nested.

themes/bootstrap3/templates/Recommend/CatalogResults.phtml
demiankatz: I don't mind adding these two enhancements in a single PR, though maybe we should rename the PR to something broader if we do.

module/VuFind/src/VuFind/Recommend/InjectSearchRunnerFactory.php
demiankatz: Maybe the solution is just to create a different factory. The InjectSearchRunnerFactory is designed to do exactly what it says: inject the SearchRunner into the constructor. But now you want to inject the SearchRunner AND set the config manager. Maybe the easiest solution is to create an AbstractSearchObjectFactory -- then you can do all the injections you want safely since the factory is tailored for AbstractSearchObject and can assume that's how it will be used.

(I think the only reason we have InjectSearchRunnerFactory is because the various Results modules didn't inherit from a common base until you refactored them, and we probably didn't think to change the factory at that time).

In any case, if you agree with this approach, we can change all the Results modules and see if anything else is still using InjectSearchRunnerFactory. If not, we can probably deprecate it and remove it in 11.0.

demiankatz (changes requested):
Thanks, @maccabeelevine -- I haven't done a thorough review or test yet, but I started by replying to your comments and sharing a few general ideas. :-)

config/vufind/searches.ini
maccabeelevine: Yeah that's more accurate but I think even more confusing. :)  I'd rather leave it as is if you are ok with it.

config/vufind/searches.ini
maccabeelevine: I added a TODO

maccabeelevine: I might also refactor the various AbstractSearchObject modules somehow so this explanation can be in a single place under side modules.

config/vufind/searches.ini
maccabeelevine: Thank you, I made this change and I think it's a lot clearer.

module/VuFind/src/VuFind/Recommend/InjectSearchRunnerFactory.php
maccabeelevine: This makes sense; implemented.  And yes InjectSearchRunnerFactory is only used for those specific modules.

themes/bootstrap3/templates/Recommend/CatalogResults.phtml
maccabeelevine: Done

config/vufind/searches.ini
```diff
@@ -254,10 +254,13 @@ WorkKeys = year
...
 ; Available modules recommended for use in the side area:
 ;
-; CatalogResults:[GET parameter]:[result limit]
+; CatalogResults:[GET parameter]:[result limit]:[ini section]:[custom heading]
 ;       Display catalog search results matching the terms found in the specified
 ;       GET parameter (default = "lookfor"), limited to a specified number of
-;       matches (default = 5).  This is designed for use with non-catalog modules.
+;       matches (default = 5). This is designed for use with non-catalog modules.
+;       Hidden filters may be defined in the specified section of the searches.ini
+;       file. The optional custom heading overrides the default for this
```
demiankatz: I would add "See [CatalogResultsVideoFilter] below for an example of this."

module/VuFind/src/VuFind/Recommend/AbstractSearchObject.php
```diff
@@ -139,6 +165,21 @@ public function init($params, $request)
...
                 $params->getOptions()->getHandlerForLabel($typeLabel)
             );
+
+            // Set any hidden filters configured for this search
+            if (!empty($this->iniSection)) {
+                $ini = $params->getOptions()->getSearchIni();
+                $config = $this->configManager->get($ini);
+                $iniSection = $this->iniSection;
+                try {
+                    $hiddenFilters = $config->$iniSection->toArray() ?? [];
```
demiankatz: I believe that you could use `$config->{$this->iniSection}->toArray()` if you wanted to avoid the need to create a separate `$iniSection` variable. Feel free to leave as-is if you prefer it this way for readability, but thought I'd mention this approach in case you weren't already aware of it (it's something I rarely use but occasionally find handy).

demiankatz (changes requested):
Thanks, @maccabeelevine, this is shaping up nicely. A couple more minor suggestions below. Once you've responded to those, I think it should be safe to finish up the remaining documentation formatting.

I've added a TODO note to myself to document the deprecation of the old factory.

I'll do some hands-on testing of this once you've responded to this latest round of feedback.

config/vufind/searches.ini
maccabeelevine: Done.

module/VuFind/src/VuFind/Recommend/AbstractSearchObject.php
maccabeelevine: I was not aware of that syntax!  Done.

config/vufind/searches.ini
```diff
@@ -259,8 +259,8 @@ WorkKeys = year
 ;       GET parameter (default = "lookfor"), limited to a specified number of
 ;       matches (default = 5). This is designed for use with non-catalog modules.
 ;       Hidden filters may be defined in the specified section of the searches.ini
-;       file. The optional custom heading overrides the default for this
-;       recommendation module. 
+;       file; see [CatalogResultsVideoFilter] below for an example ofthis. The
```
demiankatz: Missing space in "ofthis".

demiankatz: (I can fix this myself if you prefer, but I didn't want to confuse matters by pushing a commit while you were actively working on this :-) ).

config/vufind/searches.ini
```diff
@@ -300,16 +310,22 @@ WorkKeys = year
...
 ; LibGuidesResultsDeferred:[GET parameter]:[result limit]
 ;       Same as LibGuidesResults, but loaded via AJAX.
-; LibGuidesAZResults:[GET parameter]:[result limit]
+; LibGuidesAZResults:[GET parameter]:[result limit]:[ini section]:[custom heading]
 ;       Display LibGuides A-Z Databases search results matching the terms found in
 ;       the specified GET parameter (default = "lookfor"), limited to a specified
 ;       number of matches (default = 5).
+;       Hidden filters may be defined in the specified section of the LibGuidesAZ.ini
+;       file; see [CatalogResultsVideoFilter] below for an example of this. The
+;       optional custom heading overrides the default for this recommendation module.
```
maccabeelevine: I've added the documentation to these other modules for discussion purposes.  Two obvious problems though.

1. [custom heading] doesn't work until I implement it in the templates.  I can either add it to the templates or remove it from the doc for now.
2. Filtering doesn't work for many of these backends -- definitely not for LibGuides / LibGuidesAZ, not yet implemented for EPF, and I don't know about Web.  I assume it's implemented for Summon but can't test that.  So I'll presumably remove the [ini section] from those but wanted to get your take first.

config/vufind/searches.ini
maccabeelevine: Thanks yeah I caught this as well :)

config/vufind/searches.ini
demiankatz: 1.) I think you might as well go ahead and implement the custom headings for all of the templates. I wonder if it would actually make sense to just create a $heading property and getHeading() method in AbstractSearchObject. Each AbstractSearchObject subclass could set a different default property, and then the template could just display the result of calling getHeading(). Then it's a single configurable setting instead of an override option.

2.) Maybe you should reverse the order of the heading and ini section settings so it makes more sense to drop the ini section when unsupported. Otherwise you'll have to put in an extra colon separator to override the heading, and that might cause confusion. I also wonder if we should drop the word "hidden" since we're not using hidden filters here, just regular filters. We're filtering the results fetched from the backends, but we don't need to use the hidden filter mechanism since this isn't a normal user-initiated search with faceting/filtering controls. And regarding where filtering is supported, you're correct that it should work for Summon. It should also work for Web, since that's a standard Solr-based search.

maccabeelevine: This all made sense to me.  I've implemented the default heading property as an abstract method so it's easier to check compliance.  Everything else I did as suggested.

I forgot that we removed the client-side filtering settings from EDS.ini recently since they don't work.  So I removed the [ini section] for EDS and EPF, although it's something that's worth investigating along with the larger issue.

I will need help from someone to test the Summon and Web functionality.

module/VuFind/src/VuFind/Recommend/AbstractSearchObject.php
```diff
@@ -61,6 +61,20 @@ abstract class AbstractSearchObject implements RecommendInterface
...
+     * @var string
+     */
+    protected $heading;
+
+    /**
+     * Config section with filters for this search
+     *
+     * @var string
+     */
+    protected $iniSection;
```
demiankatz: Maybe call this $filterIniSection in case we have another .ini section for some other thing in future?

demiankatz (changes requested):
One minor suggestion. Once this is adjusted, I'll help test WebResults. I think we'll have to take Summon on faith (but given the nature of the changes, I'm okay with that).

module/VuFind/src/VuFind/Recommend/AbstractSearchObject.php
maccabeelevine: Good call

demiankatz (changes requested):
The recent merge of #3209 has created a bunch of conflicts here. Do you mind resolving them when time permits?

maccabeelevine:
> The recent merge of #3209 has created a bunch of conflicts here. Do you mind resolving them when time permits?

All set.

demiankatz (approved):
This looks good to me, and I tested that WebResults works as expected. Thanks again!

PR #3193, opened 2023-11-06
Configurable fields for RecordDataFormatter

ThoWagen:
This adds the option to make fields configurable when using the RecordDataFormatter.
If one sets the option "configurable" to true for a given field in the RecordDataFormatterFactory it needs to be configured in the `RecordDataFormatter.ini`. Otherwise it will not be shown. For this one needs to create a section with the field's name. One can also overwrite or add more options.

This can be useful if one has multiple instances that differ in some of the fields. Instead of creating multiple modules for them with different RecordDataFormatterFactories one can now make the fields configurable in one module and just have different configs.

TODO
- [x] Update changelog when merging (note change to constructor signature in RecordDataFormatter).


module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
```diff
@@ -126,6 +143,14 @@ protected function allowValue($value, $options, $ignoreCombineAlt = false)
      */
     protected function render($field, $data, $options)
     {
+        if ($options['configurable'] ?? false) {
+            if ($this->config[$field] ?? false) {
```
demiankatz: A very minor point, but we usually use object rather than array syntax on Laminas\Config objects. Is there any reason not to use `$this->config->$field` instead of `$this->config[$field]`?

demiankatz (changes requested):
Thanks, @ThoWagen, this makes sense and seems potentially quite useful.

Apart from the optional nitpick below, I have one other question: would it make sense to create a mechanism that would allow us to globally toggle "configurable" on for all fields? I can imagine people wanting to be able to control everything via .ini file without having to drastically override the factory or default configuration. Of course, if we did that, we'd probably want to have a more robust example .ini file set up. What do you think?

ThoWagen:
I think we could add default options in the configs that are applied to all fields. Then all fields can also be made configurable by default.

The default options could also be used for example to set the renderType to "CombineAlt".

And maybe we could also allow to add new fields via the config file. This should be easy for simple fields. But for more complex fields like "Authors" we should stick to overwriting the RecordDataFormatterFactory.

demiankatz:
Thanks, @ThoWagen, I think those are all good suggestions. I like the idea of a global changes section, though we'll have to be careful about naming to be sure it won't conflict with an actual field name.

I also wonder: would it be better to add an explicit "disabled = true/false" option to disable fields, instead of disabling them when they are omitted from the .ini file? That feels like it might be easier to understand and control than the current mechanism. And if you want to "opt in" to a small number of fields, you could set "disabled = true" in the global section, and "disabled = false" in the fields that you want to enable.

ThoWagen:
I added the option to enable or disable fields, global options and also the possibility to add extra fields.

Now that we have the "enabled" option I am not sure if we still need the "configurable" option. You either configure the fields or not. I don't know why we would want to make some fields unconfigurable. But maybe I am missing something. What do you think @demiankatz?

demiankatz:
Thanks, @ThoWagen, I agree, I can think of no reason we need the configurable option now that everything else has been made more flexible. Please remove it, and then I'll give this a final review!

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
ThoWagen: fixed

ThoWagen:
I removed the "configurable" option.

config/vufind/RecordDataFormatter.ini
```diff
@@ -0,0 +1,45 @@
+; This file can be used to set global default options for fields, to add new fields and to configure fields
+; in the RecordDataFormatter.
+
+; This section is used to set global default option.
+[Global]
+; Enable or disable all fields by default (default is true)
+;enabled = false
+; Use change default renderMethod (default is "Simple")
```
demiankatz: This comment doesn't quite make sense. Should we just remove the "Use change" part and start with "Default" or is there some other meaning that needs to be conveyed?

config/vufind/RecordDataFormatter.ini
```diff
@@ -0,0 +1,45 @@
...
+;renderType = "CombineAlt"
+; Change the default separator for array values (default is <br />)
+;separator = "; "
+; One can also add any other option.
```
demiankatz: It might be worth while to document a link to the [wiki page](https://vufind.org/wiki/development:architecture:record_data_formatter) here for reference.

config/vufind/RecordDataFormatter.ini
```diff
@@ -0,0 +1,45 @@
...
+;core[] =
+;description[] =
+
+; Example:
+; core[] = "Extra"
+; core[] = "MySpecialField"
+; to add the two fields "Extra" and "MySpecialField" to core. See below for the configuration.
+
+; To change options on a specific field one has to add a section [Field_<field's name>].
+; E.g. to enable ISBNs and to add the separator option use:
```
demiankatz: Maybe say "override" instead of "add" here?

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
```diff
@@ -126,6 +143,10 @@ protected function allowValue($value, $options, $ignoreCombineAlt = false)
      */
     protected function render($field, $data, $options)
     {
+        if (!($options['enabled'] ?? true)) {
+            return null;
```
demiankatz, line 147: This is a pre-existing error, but return type above should be `?array`.

demiankatz (changes requested):
This looks really good, @ThoWagen! Just a few suggestions/questions on the documentation side before I do my final review (and hopefully merge).

config/vufind/RecordDataFormatter.ini
ThoWagen: fixed

config/vufind/RecordDataFormatter.ini
ThoWagen: Good idea. Included it.

config/vufind/RecordDataFormatter.ini
ThoWagen: done

module/VuFind/src/VuFind/View/Helper/Root/RecordDataFormatter.php
ThoWagen: fixed

demiankatz (changes requested):
This is working well and adds a lot of flexibility. However, it has one significant limitation, which I think we can overcome with a relatively small amount of refactoring.

The problem we currently have is that the granularity here is at the field label... but the same field label could exist in multiple places in the configuration. If you wanted to disable the Language field from the collection-oriented displays but leave it intact for the core displays, I don't think this offers a way to do it, because the context in which the field is used is not taken into account.

I think the simple fix is to move the configuration processing out of getData() and into getDefaults(). By the time we're in getData, we don't know our context anymore... but we DO know our context when fetching defaults. Then we could add some additional configuration characteristics to filter by context -- this could be something like `overrideContext[]` in the field section (defaulting to "everywhere" if not specified), or we could incorporate the context into the section heading text, or maybe you can think of some better way.

Even if you don't think it's worth implementing context filtering right now, I think moving the logic would make the addition of that feature easier in the future, so it might be worth doing now.

What do you think?

ThoWagen:
@demiankatz For now I moved the `addOptions` from `getData` to `getDefaults`.

I don't mind making the configs context sensitive in this PR but I'm not sure how to do it. Adding the context in the sections headings should work well and give allot of flexibility. But the config could get become harder to read. Heaving sections like [Collection_Record_Field_Published in] are not really nice. But maybe INIs are not really made to have a more complex structure.

I don't understand first idea though. How could the array `overrideContext[]` be used to have multiple context specific options?

We could also use YAML instead of INI.

demiankatz:
> I don't understand first idea though. How could the array `overrideContext[]` be used to have multiple context specific options?

I didn't really think this one through the first time I wrote it, though there is a way we could make it work:

```
; Only show MyField in the core view
[Field_MyField]
enabled = false
overrideContext[core] = CoreOverrides_MyField

[CoreOverrides_MyField]
enabled = true
```

In other words, the field-specific section could redirect to a different arbitrarily named section for specific contexts.

This is a little more flexible than the "section name based on convention" since you could theoretically share a custom section between multiple fields with this remapping. It also keeps the basic, common use case more readable.

As I said, I don't think we need to implement this right now without a use case, but I just wanted to be sure we had a path forward -- and I think this could be a workable path forward. Let me know if you think it's worth trying it out now, or if you prefer to merge as-is and maybe open a JIRA ticket to capture this idea as a future enhancement so we don't forget this conversation. :-)

ThoWagen:
Ok, now I understand. It was rather simple to add this so I did it. I don't think we can get a much better solution for this.

demiankatz (approved):
Thanks, @ThoWagen -- this is a really exciting improvement to the system. At last, the RecordDataFormatter really does make simple changes simple to implement without code changes. I wish we had built something like this a lot sooner, but at least we've gotten here now. I appreciate your willingness to collaborate. :-)

PR #3195, opened 2023-11-07
Fix a checkCapability call to include the patron.

EreMaijala:
Fixes compatibility with the MultiBackend driver.


demiankatz (approved):
Thanks, @EreMaijala!

PR #3196, opened 2023-11-07
Replace substr with e.g. str_starts_with or str_ends_with where appropriate.

EreMaijala:
While profiling VuFind I noticed that there was a relatively high number of `substr` calls. While many of them are necessary, there was a lot of old code using them for comparison, so here's a bit cleanup. All tests are passing for me, but this was all manual work, so some scrutiny is welcome.

TODO
- [x] When merging, note removal of \VuFindSearch\Backend\EDS\SearchRequestModel::endsWith() in changelog


module/VuFind/src/VuFind/Config/Upgrade.php
```diff
@@ -1470,16 +1470,16 @@ protected function extractComments($filename)
 
             // Is the current line a comment?  If so, add to the currentComments
             // string. Note that we treat blank lines as comments.
-            if (substr($trimmed, 0, 1) == ';' || empty($trimmed)) {
+            if (empty($trimmed) || str_starts_with($trimmed, ';')) {
```
demiankatz: Any particular reason you left this as `empty` while changing other `empty` calls to `!== ''`?

module/VuFind/src/VuFind/Config/Upgrade.php
```diff
@@ -1496,7 +1496,7 @@ protected function extractComments($filename)
                 // Is the current line a setting?  If so, add to the return value:
                 $set = trim(substr($trimmed, 0, $equals));
                 $set = trim(str_replace('[]', '', $set));
-                if (!empty($section) && !empty($set)) {
+                if ('' == $section && '' == $set) {
```
demiankatz: This seems to be reversing the logic. If this didn't cause a test to fail, maybe we'd better add another test.

module/VuFind/src/VuFind/ILS/Driver/NoILS.php
```diff
@@ -269,8 +269,9 @@ protected function getFormattedMarcDetails($recordDriver, $configSection)
             // ID prefix, strip it off!
             $idPrefix = $this->getIdPrefix();
             if (
-                isset($result[0]['id']) && strlen($idPrefix)
-                && $idPrefix === substr($result[0]['id'], 0, strlen($idPrefix))
+                isset($result[0]['id'])
+                && '' !== $idPrefix
```
demiankatz, line 273: Do we need this check? If the string is empty, I would expect str_starts_with to return false anyway.

demiankatz (changes requested):
Thanks, @EreMaijala, this makes a lot of code significantly more readable, in addition to improving efficiency. I just had a couple of questions and observations from my review....

module/VuFind/src/VuFind/Config/Upgrade.php
EreMaijala: Just missed it. Fixed!

module/VuFind/src/VuFind/Config/Upgrade.php
EreMaijala: Oops, fixed! It'd only affect inline comments, but yes, not caught by any test.

module/VuFind/src/VuFind/ILS/Driver/NoILS.php
EreMaijala: Yes, it's needed. "[All strings start with the empty string](https://www.php.net/manual/en/function.str-starts-with.php)"...

demiankatz: Ah, sorry, I misread that and thought you were checking for an empty haystack rather than an empty needle! :-)

module/VuFind/src/VuFind/Config/Upgrade.php
demiankatz: I've just pushed up a test to cover this code, and confirmed that it failed before your fix and passes after.

demiankatz (approved):
Looks good to me now -- thanks again, @EreMaijala!

PR #3197, opened 2023-11-07
Add an option to always display the "Reset Filters" button.

EreMaijala:
By default the "Reset Filters" button is only displayed when retain_filters_by_default is enabled. The new setting allows it to be displayed regardless of retain_filters_by_default.

This also consolidates the handling of `retain_filters_by_default` to the base class (hope I didn't miss any important detail on why it wasn't there).


config/vufind/EDS.ini
```diff
@@ -47,6 +47,10 @@ default_noresults_recommend[] = RemoveFilters
 ; always perform new searches unfiltered (false)?
 retain_filters_by_default = true
 
+; Whether to always display the "Reset Filters" button regardless of the
+; retain_filters_by_default setting. Default is
```
demiankatz: Default is unspecified. :-)

config/vufind/Primo.ini
```diff
@@ -52,6 +52,10 @@ bulkSize = 20
 ; always perform new searches unfiltered (false)?
 retain_filters_by_default = true
 
+; Whether to always display the "Reset Filters" button regardless of the
+; retain_filters_by_default setting. Default is
```
demiankatz: Same here.

config/vufind/Summon.ini
```diff
@@ -70,6 +70,10 @@ snippets = true
 ; always perform new searches unfiltered (false)?
 retain_filters_by_default = true
 
+; Whether to always display the "Reset Filters" button regardless of the
+; retain_filters_by_default setting. Default is
```
demiankatz: ...and here.

config/vufind/searches.ini
```diff
@@ -105,6 +105,11 @@ snippets = true
 ; always perform new searches unfiltered (false)?
 retain_filters_by_default = true
 
+; Whether to always display the "Reset Filters" button regardless of the
+; retain_filters_by_default setting. Default is
```
demiankatz: ...and here.

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -124,6 +124,13 @@ abstract class Options implements TranslatorAwareInterface
      */
     protected $retainFiltersByDefault = true;
 
+    /**
+     * Should we display a "Reset Filters" link regardless of retainFiltersByDefault?
+     *
+     * @var bool
+     */
+    protected $alwaysDisplayResetFilters = false;
```
demiankatz: Since this is initialized by the constructor, it doesn't really need a default value.

module/VuFind/src/VuFind/Search/Base/Options.php
```diff
@@ -929,6 +938,16 @@ public function getRetainFilterSetting()
...
     }
 
+    /**
+     * Should the "Reset Filters" button be displayed?
+     *
+     * @return bool
+     */
+    public function shouldDisplayResetFilters()
+    {
+        return $this->retainFiltersByDefault || $this->alwaysDisplayResetFilters;
```
demiankatz: Would it be better to call getRetainFilterSetting here instead of directly accessing the property in case something is overridden in a custom subclass? (Not likely, but the abstraction seems harmless and potentially helpful).

demiankatz (changes requested):
Thanks, @EreMaijala, this makes sense to me -- just a few small comments.

Also, do you think it's worth adding a Mink test to check the various permutations of this behavior?

module/VuFind/src/VuFind/Search/Base/Options.php
EreMaijala: Fixed.

config/vufind/EDS.ini
EreMaijala: All fixed!

module/VuFind/src/VuFind/Search/Base/Options.php
EreMaijala: Yes, changed.

EreMaijala:
@demiankatz Other things addressed, but I'll add a Mink test.

EreMaijala:
@demiankatz Test added!

demiankatz (approved):
Looks good to me -- thanks, @EreMaijala!

PR #3198, opened 2023-11-07
Translate searchbox placeholder

xmorave2:
Lost in #2979


demiankatz (approved):
Thanks, @xmorave2! I have backported this to the release-9.1 branch so we can include it in release 9.1.1.

PR #3199, opened 2023-11-07
ProxyUrl: Determine whether to proxy via web service

maccabeelevine:
The current behavior of the ProxyUrl view helper is to proxy everything or nothing.  However an institution's EZproxy server is often set up to selectively identify specific URLs that do not need to be proxied, i.e. local or open resources.  This is problematic for two reasons:
- Historically this was handled (globally) via the [RedirectUnknown](https://help.oclc.org/Library_Management/EZproxy/Configure_resources/Option_RedirectUnknown) option, now renamed [UnsafeRedirectUnknown](https://help.oclc.org/Library_Management/EZproxy/Configure_resources/Option_UnsafeRedirectUnknown).  As the link explains this is a security issue, as phishing sites use the blind redirection to take advantage of the university domain trust to redirect to unexpected sites.
- The replacement [RedirectSafe](https://help.oclc.org/Library_Management/EZproxy/Configure_resources/RedirectSafe) directive works, but creates a burden to create all of those RedirectSafe directives, which can also have conflicts with existing EZproxy config directives and cause other side effects.

This PR takes advantage of a [separate, lightweight OSS tool](https://github.com/lehigh-university-libraries/ezproxy-url-checker) from @joecorall that integrates with EZproxy, parses its config files, and exposes a web service that translates a URL (or domain) into a yes/no answer of whether it requires the proxy prefix.  This PR integrates that web service into the existing ProxyUrl view helper, to prefix only when needed and avoid both the UnsafeRedirectUnknown security issues and all the RedirectSafe configuration setup & maintenance.

TODO
- [x] Note BC break (change to ProxyUrl helper constructor signature) in changelog when merging
- [x] Document in [wiki](https://vufind.org/wiki/configuration:ezproxy) when merging


module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -66,10 +78,66 @@ public function __construct($config = null)
      */
     public function __invoke($url)
     {
-        $usePrefix = !isset($this->config->EZproxy->prefixLinks)
-            || $this->config->EZproxy->prefixLinks;
+        $useWebService = $this->config->EZproxy->prefixLinksWebServiceUrl ?? false;
+        if ($useWebService) {
+            $usePrefix = $this->checkUrl($url) ?? $this->checkConfig();
```
maccabeelevine, line 85: In theory some part of this could happen asynchronously so it did not hold up view rendering.  I think that is unnecessary because
- config.ini specifies the web service should be hosted closely so the call is trivially fast.
- responses are cached
- in many cases the URLs are part of templates which are already rendered asynchronously, such as a record tab or a combined bento box

Making this async would also add a lot of unnecessarily complexity.

config/vufind/config.ini
```diff
@@ -1449,6 +1449,18 @@ replace_other_urls = true
...
 ;prefixLinks = true
 
+; Use a web service to determine whether to prefix each link, ignoring prefixLinks
+; above.  Query the configured URL via HTTP GET and a single query parameter, 'url',
+; set to the link domain. The web service should return a body with a single number,
+; '1' to prefix and '0' if not.  The web service should be hosted closely such that
+; query time is minimal.
+; See https://github.com/lehigh-university-libraries/ezproxy-url-checker for an OSS
+; implementation of this protocol that extrats the answers from EZproxy config.
+;prefixLinksWebServiceUrl = http://localhost:8080/proxyUrl
```
joecorall:
```suggestion
;prefixLinksWebServiceUrl = http://localhost:8888/proxyUrl
```

I changed the default port for that service to use `8888` since `8080` was being used by something else. May want that default set here, too.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -66,10 +80,67 @@ public function __construct($config = null)
...
+    {
+        $prefixLinksWebServiceUrl = $this->config->EZproxy->prefixLinksWebServiceUrl;
+        $queryUrl = $prefixLinksWebServiceUrl . '?url=' . $domain;
+        try {
+            $client  = $this->httpService->createClient($queryUrl);
+            $response = $client->send();
+            $responseData = $response->getContent();
+        } catch (Exception $ex) {
+            $this->logError('Exception during EZproxy web service request: ' . $ex->getMessage());
+            return null;
```
joecorall, line 141: When the service is unavailable, should we fallback to proxying the URL? This would keep the same functionality that currently exists in the event the service is unavailable for whatever reason.

```suggestion
            return TRUE;
```

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -66,10 +80,67 @@ public function __construct($config = null)
...
+     */
+    protected function queryWebService($domain)
+    {
+        $prefixLinksWebServiceUrl = $this->config->EZproxy->prefixLinksWebServiceUrl;
+        $queryUrl = $prefixLinksWebServiceUrl . '?url=' . $domain;
+        try {
+            $client  = $this->httpService->createClient($queryUrl);
+            $response = $client->send();
+            $responseData = $response->getContent();
+        } catch (Exception $ex) {
```
joecorall, line 139: I am not sure what class `$this->httpService` is, but if it's something like Guzzle, the exceptions that may occur get thrown as things other than PHP's default Exception class. Just double checking that we're catching all exceptions here and we won't get a 5XX if the service is unavailable.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
joecorall: Ah, I see. If `NULL` is returned, `checkConfig` will use the default behavior 👍

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: Yeah I'm not sure of all the possible 'throwables' (to use the java superclass) here either.  I tested with a fake hostname/port and those exceptions were caught.  But this might be missing some other cases.

config/vufind/config.ini
```diff
@@ -1449,6 +1449,18 @@ replace_other_urls = true
 ; to false.
 ;prefixLinks = true
 
+; Use a web service to determine whether to prefix each link, ignoring prefixLinks
+; above.  Query the configured URL via HTTP GET and a single query parameter, 'url',
+; set to the link domain. The web service should return a body with a single number,
+; '1' to prefix and '0' if not.  The web service should be hosted closely such that
+; query time is minimal.
+; See https://github.com/lehigh-university-libraries/ezproxy-url-checker for an OSS
+; implementation of this protocol that extrats the answers from EZproxy config.
```
demiankatz: Typo: extrats => extracts

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -50,11 +61,14 @@ class ProxyUrl extends \Laminas\View\Helper\AbstractHelper
     /**
      * Constructor
      *
+     * @param CacheAdapter           $cache  Cache for web service repsonses
```
demiankatz: Typo: repsonses

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -66,10 +80,67 @@ public function __construct($config = null)
...
             ? $this->config->EZproxy->host . '/login?qurl=' . urlencode($url)
             : $url;
     }
+
+    /**
+     * Return the configured prefixLinks setting.
+     *
+     * @return bool The configured setting, or the default
+     */
+    private function checkConfig()
```
demiankatz: I'd make this protected -- we avoid private in VuFind so that everything is extensible.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -66,10 +80,67 @@ public function __construct($config = null)
...
+    {
+        return $this->config->EZproxy->prefixLinks ?? true;
+    }
+
+    /**
+     * Check whether the given URL requires the proxy prefix.  Cache the repsonse.
```
demiankatz: Typo: repsonse

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -66,10 +80,67 @@ public function __construct($config = null)
...
+     * @param string $url The raw URL to check
+     *
+     * @return mixed Whether the URL should be prefixed, or null if it can't be determined
+     */
+    protected function checkUrl($url)
+    {
+        $domain = parse_url($url, PHP_URL_SCHEME)
+         . '://'
+         . parse_url($url, PHP_URL_HOST);
+        $usePrefix = $this->getCachedData("proxyUrl-domainToUsePrefix-$domain");
```
demiankatz: Might be worth storing the cache key in a variable to ensure consistency, instead of concatenating things twice.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -66,10 +80,67 @@ public function __construct($config = null)
...
+     * Query the web service on whether to prefix URLs to a given domain.
+     *
+     * @param $domain The domain
+     *
+     * @return mixed Whether the URL should be prefixed, or null if it can't be determined
+     */
+    protected function queryWebService($domain)
+    {
+        $prefixLinksWebServiceUrl = $this->config->EZproxy->prefixLinksWebServiceUrl;
+        $queryUrl = $prefixLinksWebServiceUrl . '?url=' . $domain;
```
demiankatz: Would it make sense to urlencode the domain, or is the service specifically set up to expect an unencoded domain?

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
demiankatz: `$this->httpService` is a wrapper around `Laminas\Http`. I think catching generic exceptions should be okay here.

demiankatz (changes requested):
Just took a quick first look at this!

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: Fixed

config/vufind/config.ini
maccabeelevine: Fixed

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: Fixed

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: Done

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: Yeah I was thinking about that as a well, forgot to raise it.  @joecorall what's your preference?  Easy enough to url-encode here.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: Yes definitely

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
joecorall: I don't think it's necessary to urlencode. IIUC punycode allows for non-ASCII alphanumeric characters in domain names, but I just scanned all of EZProxy's supported stanzas and there are no domains that have characters outside of the character range `[^A-Za-z0-9\.\-]`. Though if ya'll feel strongly I can modify the service to require the URL passed to be percent encoded.

demiankatz: I don't have a strong preference, though I tend to encode parameters habitually to avoid subtle problems. Given the nature of the data, it might make sense to set up the service so that it can accept encoded or non-encoded data for flexibility. (That is to say, if you urldecode it and it's not already urlencoded, that shouldn't hurt anything, since there aren't going to be hostnames with % characters in them anyway).

demiankatz:
@maccabeelevine, this is looking pretty good -- I'll wait for the encoding conversation to get resolved before doing hands-on testing, but I expect this can be approved and merged pretty quickly once it's finalized. Please just let me know when you're feeling ready!

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
joecorall: Fair point. I'll ensure the service can handle encoded domains.

demiankatz: Sounds good -- and once encoded hostnames are supported, I'd encourage @maccabeelevine to encode the values here, since that saves us from the future question of "did we forget to encode something here, or are we really not supposed to?"

Of course, we could also solve that problem by adding a comment saying "URL encoding is not required by this service" if going without encoding is preferred. :-)

joecorall: @maccabeelevine - looks like the service already handled percent encoded parameters as-is. I added a test for it, too https://github.com/lehigh-university-libraries/ezproxy-url-checker/pull/4

maccabeelevine: I've implemented the encoding.  As a practical matter since I am passing along the scheme + domain, the "://" will end up encoded.

maccabeelevine:
> Please just let me know when you're feeling ready!

@demiankatz I think it's ready for you.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -50,11 +61,14 @@ class ProxyUrl extends \Laminas\View\Helper\AbstractHelper
     /**
      * Constructor
      *
+     * @param CacheAdapter           $cache  Cache for web service responses
```
demiankatz: Is there a strong reason to add the new argument first in the list? I think putting it at the end would be more backward-compatible, though if you strongly prefer it this way, it's not likely to cause much trouble.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -66,10 +80,68 @@ public function __construct($config = null)
...
+     * @param $domain The domain
+     *
+     * @return mixed Whether the URL should be prefixed, or null if it can't be determined
+     */
+    protected function queryWebService($domain)
+    {
+        $prefixLinksWebServiceUrl = $this->config->EZproxy->prefixLinksWebServiceUrl;
+        $queryUrl = $prefixLinksWebServiceUrl . '?url=' . urlencode($domain);
+        try {
+            $client = $this->httpService->createClient($queryUrl);
```
demiankatz: Since you're using the httpService, I believe that you could simplify several lines of this code into:

```
$response = $this->httpService->get($prefixLinksWebServiceUrl, ['url' => $domain]);
```

(See the [get method](https://github.com/vufind-org/vufindhttp/blob/dev/src/VuFindHttp/HttpService.php#L197) here if you want all the details of what it's doing).

demiankatz (changes requested):
I saw a couple of possible opportunities for improvement/simplification in my latest review. Let me know what you think!

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: Yes and no.  I added it first because I initially wanted it to be a required argument, and $config is optional.  Then for some unclear reason I made it optional as well, but that may have been unintentional.  Easier just to require it?

demiankatz: Yeah, if it's a required dependency, then making it required makes sense. If it's optional, then I'd move it to the end. :-)

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: Nice, updated to use get().  Wherever I looked for httpService in the code wasn't doing that.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: Sure I just wasn't sure if it was ok to make it required, since the cache is only used in practice if the web service mode is enabled.  But since the object cache already exists for many other modules, I doubt it adds much overhead to retrieve it here even if it's not used.  Changed to required.

maccabeelevine: And... now I see it was optional for the sake of some tests.  I guess I will change them both to optional and fix the order.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -66,10 +80,66 @@ public function __construct($config = null)
      */
     public function __invoke($url)
     {
-        $usePrefix = !isset($this->config->EZproxy->prefixLinks)
-            || $this->config->EZproxy->prefixLinks;
+        $useWebService = $this->config->EZproxy->prefixLinksWebServiceUrl ?? false;
+        if ($useWebService) {
+            $usePrefix = $this->checkUrl($url) ?? $this->checkConfig();
```
demiankatz, line 85: On closer examination, I'm not really sure if it makes sense to use `checkConfig()` as the fallback here. It looks to me like the main purpose of the prefixLinks setting is to allow prefixing to be globally disabled. I think if somebody configures both prefixLinks and prefixLinksWebServiceUrl at the same time, that's essentially a configuration error. I suppose it doesn't necessarily hurt anything to leave this as it is, but it just feels to me like it might not be the best option. Maybe we should just fall back to true instead of falling back to checkConfig, for example...

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
```diff
@@ -66,10 +80,66 @@ public function __construct($config = null)
...
+    {
+        $prefixLinksWebServiceUrl = $this->config->EZproxy->prefixLinksWebServiceUrl;
+        try {
+            $response = $this->httpService->get($prefixLinksWebServiceUrl, ['url' => $domain]);
+            $responseData = $response->getContent();
+        } catch (Exception $ex) {
+            $this->logError('Exception during EZproxy web service request: ' . $ex->getMessage());
+            return null;
+        }
+        return '1' === $responseData;
```
demiankatz, line 143: Do you think it might be a good idea to trim $responseData? I tested this by putting a text file in VuFind's public directory and pointing the configuration at that -- I could then edit the file to change it to 0 or 1 to test both cases. However, my editor initially added line breaks, and `1\n` is interpreted as "not 1." This caused me a little bit of confusion, and I imagine it's possible if somebody builds their own custom web service that the responses might end up having stray whitespace in them.

demiankatz (changes requested):
A couple more minor thoughts, but I think this can be merged after these are addressed.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: I disagree on this, although not strongly.  That fallback to `checkConfig()` is going to happen on any exception checking the web service, such as a connection error, which itself is likely enough from time to time (even if the services are co-hosted; things break).  So people should think about what the fallback default should be, i.e. it's better as a configured param than hard-coded, and so we may as well reuse `prefixLinks`.  Admittedly the config.ini isn't clear that it's a fallback, so I just fixed that.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
maccabeelevine: Agreed, done.

module/VuFind/src/VuFind/View/Helper/Root/ProxyUrl.php
demiankatz: Yes, that makes sense -- I was thinking about it differently, but tweaking the comments is a good solution. :-)

demiankatz (approved):
Thanks, @maccabeelevine!

PR #3200, opened 2023-11-08
Add ISO 639-3 language code translations.

EreMaijala:
The Northern Sámi translation is a bit incomplete, but at least it contains the most common (for us) languages.

TODO
- [x] Obtain mappings/translations for other supported languages
- [x] Extract additional translations from [VUFIND-583](https://openlibraryfoundation.atlassian.net/browse/VUFIND-583) and [VUFIND-753](https://openlibraryfoundation.atlassian.net/browse/VUFIND-753) if possible.
- [x] Make sure appropriate attribution is in place and there are no licensing issues with data before merging.
- [x] Implement aliasing functionality for translator as described in #413 
- [x] Finish up other functionality from #413 when merging
- [x] Close [VUFIND-583](https://openlibraryfoundation.atlassian.net/browse/VUFIND-583) and [VUFIND-753](https://openlibraryfoundation.atlassian.net/browse/VUFIND-753) when merging.
- [x] Document new util/i18n directory in wiki when merging


demiankatz:
Thanks, @EreMaijala! I've merged #2933, since I believe that work is now ready to move forward, and I have added some TODO's here. I imagine there should be some open data somewhere that we could use to generate these translations for other supported languages; it's probably worth looking into that before merging this. I'm also not sure whether we want to keep this simple and just merge the language files on their own, or if we should also implement the aliasing translation feature here to leverage these files. There's also all the other outstanding functionality from #413 to think about. For now, I just wanted to capture all that needs to be done; as we make more progress, we can discuss further how best to split up the work.

EreMaijala:
I haven't actually tried, but https://github.com/ihmpavel/all-iso-language-codes might be useful.

demiankatz:
Looks very promising -- thanks, @EreMaijala!

demiankatz:
I've tagged this as an architecture PR since it's currently the home of the full TODO list for this project. We can recategorize if we decide later to do this in smaller parts.

demiankatz:
@EreMaijala, I've just done a quick experiment with the all-iso-language-codes library you mentioned. You can now generate language files for a specific language code like this:

`node $VUFIND_HOME/jstools/create_language_translations.js [language code]`

Unfortunately, it appears that the data used by this tool is very incomplete. When I run the command with `fi` as the code, it reverts a lot of Finnish translations to non-internationalized versions. I believe this is because the language-specific source data for all-iso-language-codes is stored in ISO 639-1 format, which is less granular than ISO 639-3... and any translation request for a 639-3 code not present in 639-1 just uses the English text.

There are a few places we can potentially go from here:

1.) Since this project seems active, we could talk to its maintainer and see if there's a way we could improve it to more fully support 639-3 translation.

2.) We could generate full language files using the script I wrote, on the theory that something is better than nothing, and we can update the translations as time permits.

3.) We could write a more intelligent script that ONLY outputs the 639-1 codes, so we can take advantage of the available translations and inherit the 639-3 codes from English until better versions can be found.

Any thoughts/preferences?

demiankatz:
From a little more digging: the problem appears to be that the data used by all-iso-language-codes comes from [Intl.DisplayNames](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames), which doesn't include 639-3 support as discussed in [this thread](https://stackoverflow.com/questions/68282742/intl-displaynames-in-chrome-not-working-with-3-letter-codes).

demiankatz:
One more update: I've tried to make the script a little smarter so it filters out information that is less likely to be useful. I'm sure this is imperfect, but it might at least be a way to get some incomplete baseline files that cover the most commonly used languages.

demiankatz:
...and a final update before I let this rest: I've revised the script so that it only adds new language strings to existing files instead of overwriting them, so we can use this to augment what we have without replacing manually-created strings. I think this might be good enough to set up some baseline files, but I'd be interested in more input before I actually commit to doing the work of building all the files.

EreMaijala:
> ...and a final update before I let this rest: I've revised the script so that it only adds new language strings to existing files instead of overwriting them, so we can use this to augment what we have without replacing manually-created strings. I think this might be good enough to set up some baseline files, but I'd be interested in more input before I actually commit to doing the work of building all the files.

This would have been my suggestion if you hadn't already done it. :) I think this approach is good enough for starters, and allows people to contribute more complete versions as necessary.

demiankatz:
Thanks, @EreMaijala! I've adjusted the script so that it can generate multiple languages at once, and I've used it to create baseline files for all of the currently supported languages. I think this is probably good enough to check off the first TODO box here, though obviously more translation work will be needed to polish it up. I've also used inheritance in some places where it may not be appropriate (e.g. for Chinese Traditional vs. Chinese Simplified), but for now the theory is "something is better than nothing." :-)

So now the question is whether we merge this to get the files in place in Lokalise (in which case work could continue in #413), or whether we continue working here to actually start utilizing them before we merge (in which case we could probably close #413 as superseded by this one). Any opinions?

EreMaijala:
I don't have a strong preference, but I think I'd slightly prefer starting a new PR for the actual use of the translations.

demiankatz:
Update: I did a quick search of JIRA and found a couple of language-related JIRA tickets that we may be able to use to further enhance the results here. See [VUFIND-583](https://openlibraryfoundation.atlassian.net/browse/VUFIND-583) and [VUFIND-753](https://openlibraryfoundation.atlassian.net/browse/VUFIND-753). When time permits, I'll see if I can extract more translations from the work done there; I've added more TODO checkboxes to account for this.

demiankatz:
> I don't have a strong preference, but I think I'd slightly prefer starting a new PR for the actual use of the translations.

I'll at least leave this open until I can incorporate the newly-rediscovered translations noted above. Then maybe we can start another separate PR to begin work on the aliasing mechanism, and we can distribute the remaining TODO checkboxes between that PR and #413.

demiankatz:
Another update: I have succeeded in obtaining more translations from the files in [VUFIND-583](https://openlibraryfoundation.atlassian.net/browse/VUFIND-583), significantly expanding the French, German and Spanish language files. This also revealed that the English file was missing some of the extended codes currently found in our language map file, so I added those for completeness.

I think in theory, we could get even more complete data for most/all of the languages using [lexvo](http://www.lexvo.org/), but I don't see a simple way to extract that data without setting up a triplestore.

For the moment, I think I've done as much as I can do here. Next time I have time to make forward progress, perhaps I'll look into starting a new PR to implement aliasing functionality for translations. That might be the most logical first thing to merge, and then we can continue mapping/translation work here.

EreMaijala:
@demiankatz Perhaps we could just parse the lexvo dump with [hardf](https://packagist.org/packages/pietercolpaert/hardf)?

demiankatz:
> @demiankatz Perhaps we could just parse the lexvo dump with [hardf](https://packagist.org/packages/pietercolpaert/hardf)?

Possible! I haven't looked at the dump to see exactly what data is present or how it is structured... but if it's relatively straightforward, this might work.

demiankatz:
Also, just for the record, before I throw it away, this is the crazy script I used to extract useful values from the VUFIND-583 files (after putting them in my /tmp directory, and getting a legacy version of language_map.properties from an earlier commit, before some of the values were updated). I doubt this will ever be used/needed again, but if it's in this comment, I can find it again if I somehow need it.

```
<?php

$infile = "/tmp/de.ini";
$outfile = "/tmp/de-out.ini";
$srcData = "/usr/local/vufind/import/translation_maps/language_map.properties";

$handle = fopen($outfile, 'w');

$incoming = [];
foreach (load_file($infile) as $line) {
    $incoming[$line[0]] = $line[1];
}

$reverseMap = [];
foreach (load_file($srcData) as $line) {
    $reverseMap[$line[1]] = $line[0];
}
foreach (load_file("/tmp/language_map.properties.orig") as $line) {
    $reverseMap[$line[1]] = $line[0];
}
foreach (load_file("/usr/local/vufind/languages/ISO639-3/en.ini") as $line) {
    $reverseMap[$line[1]] = $line[0];
}

foreach ($incoming as $code => $val) {
    if (!isset($reverseMap[$code])) {
        echo "$val\n";
    } else {
        fputs($handle, "{$reverseMap[$code]} = $val\n");
    }
}

fclose($handle);

function load_file($filename) {
    $callback = function ($line) {
        $parts = explode('=', $line);
        $callback = function ($val) {
            return trim($val, "\n\r\t\v\x00 \"");
        };
        return array_map($callback, $parts);
    };
    return array_map($callback, file($filename));
}
```

demiankatz:
See #3249 for work in progress on aliasing (still an early draft as of this writing, but just wanted to get the ball rolling...)

demiankatz:
Good idea, @EreMaijala! I've managed to extract a lot more language translations using the lexvo data and hardf. I've included the script I used in the util directory for reference.

As a result of all this work, this PR is adding two new dev dependencies (one JS, one PHP) and two new command line scripts... that may be a maintenance overkill for something that we're unlikely to use frequently (though I imagine will need these whenever we add a new language to the system). Should we keep these somewhere else to keep the main project cleaner?

demiankatz:
The other question is whether there are any licensing concerns/obligations that we need to worry about. The lexvo data is released under a Creative Commons license (see [this page](http://lexvo.org/legal.html)) but it mentions some data deriving from the Ethnologue database, which has [its own terms](https://www.ethnologue.com/codes/). We should probably figure out how to acknowledge these sources and/or obtain permission for use -- just not sure where the best starting point is. It would be easy enough to include a standard comment at the top of all of the impacted language files, and we could also put acknowledgments in the wiki. It gets trickier if there's an expectation of credit in the VuFind interface itself. Any thoughts on where we should begin here?

EreMaijala:
@demiankatz I'd prefer keeping the scripts in the main project for simplicity.

I'm not an expert on licensing, but from what I see there's no requirement to display the acknowledgement in the UI, so the comments and perhaps a mention in wiki would seem to be enough, but again, I could be mistaken!

demiankatz:
Thanks, @EreMaijala -- I've added attribution statements to all of the language files (with an extra line in the English file to acknowledge Ethnologue, which is not used by the other languages). I've also added comments to the new PHP utility script and moved both of the utility scripts to the util/i18n directory to group them together and keep them separate from more general-purpose scripts. If you have a better idea of placement or handling of these, I'm open to suggestions.

demiankatz (approved):
I believe this is functionally complete, and it now incorporates all of the work from #413.

Since I made most of the recent changes, I don't want to merge on a self-approval; if @EreMaijala or @ThoWagen could take a look at the last few commits and let me know if they have any concerns about my changes, I would appreciate the double-check!

EreMaijala:
@demiankatz Looks good to me! I just can't add my approval since I opened this PR. :)

PR #3203, opened 2023-11-08
Setup recaptcha on google script load jquery to javascript update

padmasreegade:
Setup recaptcha on google script load jquery to javascript update


crhallberg (approved):
This looks good to me! Thank you!

demiankatz (approved):
Excellent work, @padmasreegade! I've confirmed that the updated code works the same as the original code in my test environment. Thank you!

PR #3204, opened 2023-11-09
Responsive table for search history.

pasitiis:
Currently, the search history table easily extends beyond the screen on mobile. The issue has been highlighted in accessibility audits. To fix the problem, I applied a relatively well-known pattern for responsive tables, which should address the issue. Later, the same method could be used to present other tables as well.


demiankatz:
Thanks, @pasitiis!

@crhallberg, could you review the changes here and share feedback if you have any?

@sturkel89, could you do some hands on testing of this and see what you think about the UI changes? You should be able to perform a few searches, then create an account and save some, and then look at the resulting search history displays on a full-size browser and a smaller window. It may be worth also turning on schedule_searches in config.ini to see how that behaves, since it adds additional controls to the search history display (see [Email Alerts](https://vufind.org/wiki/configuration:email_alerts) for more details as needed).

sturkel89:
I'm not sure if I'm doing something wrong, but I can't see a difference between dev and this test branch. I've used a wide and the most narrow browser, and I have tried a mobile emulator browser plugin. I toggle back and forth between dev and test, and when I refresh each time I don't see any shift or change.

This is what I see, on either branch. Please let me know if you have suggestions.

image

demiankatz (changes requested):
@pasitiis, it looks like the compiled CSS has not been committed to this PR. Could you run `npm run build:css` and add the resulting files here?

Thanks!

pasitiis:
Oops. I accidentally removed the compiled CSS files from PR. CSS files have now been added. Should look on mobile like this. 

Screen-Shot-2023-11-10-at-08 51 34

sturkel89:
Thanks for updating the code; I see the differences now! Please see below for my comments and suggestions.

**Test branch**

image

The functionality of the test branch is good; it seems to work as intended. However, there is one issue that needs fixing, and several possible enhancements I'd suggest trying.

***Necessary change:*** 

- I turned on schedule_searches in config.ini and see that the heading "Alert schedule" is not displayed above the dropdown box in the test branch. That label needs to appear.

***Suggested enhancements:***

- I'd like to see a stronger outline around each entire saved search entity, to provide more visual distinction between the searches on the list.

- If there are no limits imposed in a saved search, it would save space to not display the **Limits** row at all.

- The new display takes up a lot of vertical space. It would be more efficient to have each set of information for a specific search presented in a two-column arrangement, with the headings in the first column and the variable content in the second column. (I don't know if alternating colors are needed; this display is imposed by GitHub's table formatting. Also, the limits could be displayed as a list; I couldn't get them to roll over to two rows.) Example:

|     |    |
| --- | --- |
| **Time** | 11-10-2023 09:36 |
| **Search** | adult children of aging parents |
| **Limits** | **Format:** Book     **Language:** English |
| **Results**| 7 |
| **Alert schedule** | [dropdown box] |
| *Delete* | |

With these minor tweaks, I think this is a great improvement to what we have on dev!

themes/bootstrap3/templates/search/history-table.phtml
```diff
@@ -43,7 +43,7 @@
             <?php endif; ?>
           <?php endforeach; ?>
         </td>
-        <td class="history_results"><?=$this->escapeHtml($this->localizedNumber($info->getResultTotal()))?></td>
+        <td class="history_results" data-label="<?=$this->transEscAttr('history_results')?>"><?=$this->escapeHtml($this->localizedNumber($info->getResultTotal()))?></td>
```
demiankatz, line 48: I think we need a data-label on the search-schedule-header td below to resolve @sturkel89's reported problem.

Is it possible that we should also label the history_actions td further down (the container for the save/delete buttons), or is that okay without a label? (I think it seems fairly self-explanatory without the label, but just making sure there's not a reason to add something).

demiankatz (changes requested):
@pasitiis, just a couple of specific notes as a follow-up to @sturkel89's testing feedback.

themes/bootstrap3/templates/search/history-table.phtml
pasitiis: I think actions is okay without a label. Next, I'll check @sturkel89 suggested enhancements.

crhallberg:
This all looks good to me. I tried to compile the CSS but I don't have branch permissions.

demiankatz (changes requested):
Thanks, @pasitiis!

Please let us know when you are ready for the next review.

Also, as Chris points out, there are some conflicts that need to be resolved here (just a matter of running `npm run build:css` during the process of merging dev to replace the conflicted versions with new ones). If you're willing to give project contributors write access to your PR branches, we can help with this kind of routine maintenance during the review process, but as Chris points out, right now, we don't seem to have permission to push to your branch.

pasitiis:
I was busy with other work for the rest of the week, so I didn't have time to touch this. I will continue working on this next week and resolve those conflicts.

pasitiis:
@sturkel89 I finally made layout improvements based on the feedback. It was a bit tricky to align the header of the table cell correctly.  I'm still considering that it might be a good idea to display the limits even if there is nothing to show, as the same cell header is displayed in the desktop.

Tested on the following browsers: Safari (iOS), Chrome (MacOS), Safari (MacOS), Firefox (MacOS)
Screen Shot 2023-11-24 at 10 55 49

sturkel89 (changes requested):
I've reviewed the branch using a Chrome plugin that emulates a mobile interface (and can be set to display as various flavors of iPhone, Android, etc.).

The recent changes look great! I only have a few small nitpicks. Aside from these concerns, I think this branch is ready to be merged.

1. **The word Delete appears twice**, as a label and in the action link to Delete. Please remove the label. Ideally, the red Delete link with the trash can icon would be located in the left-hand column, replacing the current location of the label.

2. When the "Alert schedule" label is displayed on one line, it is nicely centered vertically. But when it scrolls to two lines in some phone interfaces, it seems to be bottom-aligned within the row. The rest of the labels seem to be top-aligned; see Limits (when there is a multi-line display of limits) and Search (when the search string flows onto two lines). **The "Alert schedule" label should be top-aligned like the rest of the labels.**

### Alert schedule on one line: looks good!
image

### Alert schedule on two lines: alignment needs correction
image
image

demiankatz:
Thanks, @sturkel89 -- I don't think the suggestion of moving the delete icon into the label column is going to work, though, as the icon-plus-label link is really a single unit as far as the interface is concerned. I think we're probably best off with either no label or a generic "Take Action" kind of label.

sturkel89:
In that case, I think that no label would be fine.

pasitiis:
> 2\. **The "Alert schedule" label should be top-aligned like the rest of the labels.**

I'll fix this.

pasitiis:
@sturkel89 I adjusted the layout of labels and removed action label. Can you check that the changes are as you intended?

sturkel89 (changes requested):
There were two issues for me to check in this review: the duplication of "Delete" in each saved search, and the alignment of the "Alert schedule" label (which had been bottom-aligned in its cell when it rolled over to two lines).

- **DELETE issue**

The action label has been removed, so only the clickable "Delete" with the trash can icon remains. This works well!!

image

- **ALERT SCHEDULE label issue**

I found that in almost every mobile phone model in the free version of my emulator, the phrase "Alert schedule" appears on one line -- it looks great that way, with everything vertically centered in its cell:

image

However, the iPhone 5 has a smaller screen, so the "Alert schedule" label rolls over to two rows. I see that the **label** is now top-aligned (or perhaps centered in the cell) -- but the **dropdown element** where the user sets the schedule is now bottom-aligned! 

image

The dropdown element needs to also be top-aligned, if possible. Once that is fixed, I think this branch will be ready to merge.

pasitiis:
> The dropdown element needs to also be top-aligned, if possible. Once that is fixed, I think this branch will be ready to merge.

The dropdown element alignment is a bit tricky with the label element. The label element can be either centered in the cell or aligned to the first baseline or last baseline. Other fields look better with first baseline alignment. The best solution would be to avoid the label wrapping.

I changed the label alignment back to the first baseline and added a bit more space for the label, but it's impossible to prevent the label from wrapping in all situations.

demiankatz:
Thanks, @pasitiis! @sturkel89, can you give this another look? Sounds like we may have to consider "good enough to be good enough" due to the limitations around styling drop-down menus... so as long as it's functional, I think we can let it go if it's not 100% perfect. :-)

sturkel89 (approved):
I have tested the latest update in all three themes, using a variety of emulated smartphones and also a narrowed Chrome desktop browser.

The new spacing and alignment settings seem to be doing the trick! I can't find a way to make "Alert schedule" scroll to two lines, and the other fields that have the potential to overflow onto multiple lines (Title and Limits) are both top-aligned in both the label column and the content column.

All looks good! I think we are ready to merge.

demiankatz (approved):
Thanks, everyone!

PR #3205, opened 2023-11-09
fetch icon jquery to javascript update

padmasreegade:


themes/bootstrap3/js/common.js
```diff
@@ -157,26 +157,24 @@ var VuFind = (function VuFind() {
...
+    if (typeof _icons[name] === "undefined") {
       console.error("JS icon missing: " + name);
       return name;
     }
     // Create a template element for icon function
     if (!_elementBase) {
       _elementBase = document.createElement('div');
     }
-    const cacheKey = `${name}||${JSON.stringify(attrs)}`;
+    const cacheKey = name + '||' + JSON.stringify(attrs);
```
demiankatz, line 169: The backtick template syntax with `${}` placeholders is a native JavaScript feature, not a jQuery thing (in spite of the presence of dollar signs); I don't think we need to change it.

themes/bootstrap3/js/common.js
```diff
@@ -157,26 +157,24 @@ var VuFind = (function VuFind() {
...
     if (_iconsCache[cacheKey]) {
       return returnElement
-        ? _iconsCache[cacheKey].cloneNode(true)
+        ? _iconsCache[cacheKey].cloneNode(true).outerHTML
```
demiankatz, line 172: I believe that returning the raw node here was intentional -- recently added by #3126. Was this causing a problem for you?

demiankatz (changes requested):
Thanks, @padmasreegade. Could you provide a little more context for this change; I'm not sure I understand the goals (see specific questions below).

PR #3206, opened 2023-11-09
Setup QR code links without jQuery

padmasreegade:


themes/bootstrap3/js/common.js
```diff
@@ -271,18 +271,20 @@ var VuFind = (function VuFind() {
...
+        if (holder.find('img').length === 0) {
+          // We need to insert the QRCode image
+          var template = holder.find('.qrCodeImgTag').html();
+          holder.html(template);
+        }
+      });
     });
   }
-
+
```
demiankatz: Looks like some unnecessary whitespace was added here. Might want to check your IDE's "trim whitespace" setting.

themes/bootstrap3/js/common.js
```diff
@@ -271,18 +271,20 @@ var VuFind = (function VuFind() {
...
-      if (holder.find('img').length === 0) {
-        // We need to insert the QRCode image
-        var template = holder.find('.qrCodeImgTag').html();
-        holder.html(template);
-      }
+    var container = _container || document.body;
+    var qrcodeLinks = container.querySelectorAll('a.qrcodeLink');
+    qrcodeLinks.forEach((link) => {
+      link.addEventListener('click', () => {
+        var holder = this.nextElementSibling;
```
demiankatz, line 278: I tested this and it is not working for me. I suspect that this line may be the issue -- I'm not sure that `this.nextElementSibling` is equivalent to `$(this).next('.qrcode')` since it is not accounting for the requested class name. However, I could be wrong!

In any case, if you would like to test this yourself so you can troubleshoot more easily, just edit your `$VUFIND_LOCAL_DIR/config/vufind/config.ini` file and turn on `showInResults` in the `[QRCode]` section. Then when you perform a search, you'll see QR Code drop-downs on the right side of every search result. If you expand the drop-downs, you'll find that you see a QR code in dev, but in this branch the image does not load.

demiankatz (changes requested):
Thanks, @padmasreegade! This isn't working for me in my test environment -- see below for details. I hope this helps you make further progress, but please let @crhallberg or me know if we can offer additional assistance!

demiankatz:
I think [this article](https://gomakethings.com/converting-the-jquery-next-method-to-vanilla-js/) may help, too!

themes/bootstrap3/js/common.js
padmasreegade: Thanks @demiankatz, I will try it out.

demiankatz (approved):
This is working correctly for me now, and all tests are passing. Thanks, @padmasreegade!

crhallberg:
Fun fact: this fixes a bug that never reared its head: I apparently used jQuery next entirely wrong!

PR #3207, opened 2023-11-10
Increase contrast in form elements.

pasitiis:
Currently, the contrast of the form element borders is too low (1.4:1), it should be at least 3:1 (WCAG AA). Also, the focus color is a bit too low. In this PR, I have increased the contrast of the elements.


demiankatz:
Thanks, @pasitiis!

@crhallberg, please share any thoughts you might have on this. @sturkel89, can you check that form element borders look okay in all of our themes with this change?

pasitiis:
A contrast issue emerged during accessibility auditing. Previously, I had thought that the contrast of the form field border didn't need to be 3:1. More details on non-text contrast can be found [here](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html).

sturkel89:
I can see the subtle differences in each theme. The main difference seems to be that the **non-selected form boxes have a stronger outline** on the test branch than they do on dev. Below are examples of both versions in each of the three themes.

Bootprint3 and Sandal also show color differences in the selected form field box outline, although I think the dev version on Sandal shows more contrast with the orange line. (I don't know what's best for accessibility, though!)

**BOOTSTRAP3**

Test:
image

Dev:
image

**BOOTPRINT3**

Test:
image

Dev:
image

**SANDAL**

Test:
image

Dev:
image

demiankatz:
Thanks, @sturkel89! It looks like perhaps @pasitiis' changes are somehow overriding some of our theme-specific color choices. I don't object to this if it's in the interest of accessibility, but I wonder if @crhallberg has any concerns on that front.

crhallberg:
I do see a concern with the increased contrast on sandal. The bottom border color changing is likely not enough contrast and it might need a halo like the other themes. This is poor design from my end, not just from the increase in contrast.

Something like this would do the trick. There may be an easier way to adjust colors compatibly. Happy to fix this myself if you give me branch permissions.

```css
/* #LESS> */
--outline-color: fadeout(@brand-secondary, 50%);
/* <#LESS */
/* #SCSS>
--outline-color: color.change($brand-secondary, $alpha: 0.5);
<#SCSS */

box-shadow: 0 0 0 4px var(--outline-color);
```

demiankatz (changes requested):
Just putting this into "request changes" mode until @pasitiis has a chance to address @crhallberg's feedback.

pasitiis:
> Happy to fix this myself if you give me branch permissions.

Go ahead and fix it. I had some issue using global less-variables with forms. Is there anything specific I need to do to grant branch permissions?

demiankatz:
@pasitiis, I tried to resolve conflicts here, since I see that @crhallberg was able to push up commits. Unfortunately, it looks like I still don't have permission. Could you do for me whatever you did for @crhallberg so that I can help keep your PRs up to date when simple conflicts are introduced?

pasitiis:
> Unfortunately, it looks like I still don't have permission.

Oops. I added the permissions now.

demiankatz:
Thanks, @pasitiis -- I was able to resolve conflicts!

sturkel89 (commented):
I've checked the test branch in all three themes. It looks like the main change is to the highlighting around the selected box in Sandal. This review will focus on Sandal.

The border is very visible now, and the color is consistent all around (instead of being orange across the bottom). 

Speaking on the basis of my personal taste, I would prefer that the thickness of the outline be consistent all the way around the box, rather than heavier across the bottom. What do @demiankatz and @crhallberg think?

### Test branch
image

### Dev branch
image

### Bootstrap3, to show consistent thickness on all sides
image

demiankatz:
Thanks, @sturkel89 -- I defer to @crhallberg on this one; it seems likely that some custom style rule was added to sandal to further emphasize highlighted form elements, and that's probably what's causing the discrepancy you're seeing. It looks like @pasitiis has changed the color but left the emphasis intact. I don't know why this was done originally, though -- and from my perspective, I don't think it makes much difference one way or another. I think there's a good chance that @crhallberg knows more of the history here and might have a stronger opinion. Maybe we can just remove some extra styles and simplify the sandal theme... or maybe it's best to leave it alone. I'm happy either way!

crhallberg:
I've addressed @sturkel89's feedback about more consistent styling and removed the bottom border styling. I also differentiated between hover and focus and made sure sandal does not to override @pasitiis' new border color.

crhallberg (commented):
Resolving my review in case it will block future merging, but I'm a contributor now, so making it a comment.

demiankatz:
Thanks, @crhallberg!

@sturkel89, do you mind taking one more look at this one?

sturkel89 (approved):
With Chris's changes, the border is solid and the same weight all the way around the selected form field:

image

The orange really stands out against the rest of the color scheme!

Of course, this nice bright border makes me wish that the highlighted field in the other themes was also so visible! Not sure if that kind of change is also on the table. In case it isn't up for discussion, I will approve the current version for merging since it completes the changes we set out to make.

demiankatz:
Thanks, @sturkel89. I will wait to hear @pasitiis's final word before merging this, in case there are any remaining concerns about color choices, or a desire to extrapolate the Sandal-specific styling to the other themes. (I have no strong opinions one way or the other).

pasitiis:
@demiankatz  Sorry for delay. To me, the change looks good, and it can be merged.

demiankatz (approved):

PR #3208, opened 2023-11-10
Make SimulatedSSO module configurable.

demiankatz:
This PR introduces a new .ini configuration which can be used to override the behavior of the SimulatedSSO authentication module. It is now possible to log in as multiple users and to simulate failures using SimulatedSSO. This should help with testing/troubleshooting of #2941.


module/VuFind/src/VuFind/Auth/SimulatedSSO.php
```diff
@@ -71,10 +91,31 @@ public function __construct($url)
...
-        $user = $this->getUserTable()->getByUsername('fakeuser1');
-        $user->firstname = 'Test';
-        $user->lastname = 'User';
-        $user->email = 'fake@example.com';
+        $username = $this->simulatedSSOConfig['General']['username'] ?? 'fakeuser1';
+        if (!$username) {
+            throw new AuthException('Simulated failure');
+        }
+        $user = $this->getUserTable()->getByUsername($username);
+        $attribs = array_merge(
```
ThoWagen: Maybe the default attributes should be allowed to be unset. Or we always have set firstname, lastname and email? I that is not necessary we can discard the default attributes in this file and only use the config.

EreMaijala (approved):
Excellent! :)

module/VuFind/src/VuFind/Auth/SimulatedSSO.php
demiankatz: Those three attributes are generally expected, but you raise the good point that it is useful to be able to test what happens when they are absent. I've refined the configuration a bit so that defaults are provided in a backward-compatible, convenient way, but it's also possible to entirely disable attributes, or to test with a different basic set of values (without having any defaults forced in). See d494014.

PR #3209, opened 2023-11-10
Change side recommendation headings to <h2>

maccabeelevine:
The first-level headers for side recommendations tend to be `h4` for no obvious reason.  A few are `h3` or `h2`.  For accessibility it appears they should all be `h2`.

- [x] Add support for Databases after #3160


themes/bootstrap3/templates/Recommend/SummonResults.phtml
```diff
@@ -1,7 +1,7 @@
 <?php $searchObject = $this->recommend->getResults(); ?>
 <?php $results = $searchObject->getResults(); ?>
 <?php if (!empty($results)): ?>
-  <h4><?=$this->transEsc('Summon Results')?></h4>
+  <h2><?=$this->transEsc('Summon Results')?></h2>
```
maccabeelevine, line 4: I can't test Summon

themes/bootstrap3/templates/Recommend/SummonResultsDeferred.phtml
```diff
@@ -4,7 +4,7 @@
     . "VuFind.loadHtml(\$('#SummonDeferredRecommend'), url);";
 ?>
 <div id="SummonDeferredRecommend">
-  <h3><?=$this->transEsc('Summon Results')?></h3>
+  <h2><?=$this->transEsc('Summon Results')?></h2>
```
maccabeelevine, line 7: I can't test Summon

themes/bootstrap3/templates/Recommend/WebResults.phtml
```diff
@@ -4,7 +4,7 @@
 ?>
 <?php if (!empty($results)): ?>
 <div class="sidegroup">
-  <h4><?=$this->transEsc('Library Web Search')?></h4>
+  <h2><?=$this->transEsc('Library Web Search')?></h2>
```
maccabeelevine, line 7: I can't test Web search

themes/bootstrap3/templates/Recommend/SummonResults.phtml
demiankatz: I'm not sure if we have any Summon users left, but I'd be willing to trust this change. ;-)

themes/bootstrap3/templates/Recommend/FavoriteFacets.phtml
```diff
@@ -2,7 +2,7 @@
 <?php $sideFacetSet = $this->recommend->getFacetSet(); ?>
 
 <?php if (isset($sideFacetSet['tags']) && !empty($sideFacetSet['tags']['list'])): ?>
-  <h4 id="acc-menu-favs-header" class="tag"><?=$this->transEsc($sideFacetSet['tags']['label'])?></h4>
+  <h2 id="acc-menu-favs-header" class="tag"><?=$this->transEsc($sideFacetSet['tags']['label'])?></h2>
```
maccabeelevine: This recommendation module is not documented in searches.ini so I'm not sure how to use / test it.

themes/bootstrap3/templates/Recommend/ExpandFacets.phtml
```diff
@@ -6,7 +6,7 @@
 <?php if ($expandFacetSet): ?>
   <div class="sidegroup">
  <?php foreach ($expandFacetSet as $title => $cluster): ?>
-    <h4><?=$this->transEsc($cluster['label']) ?></h4>
+    <h2><?=$this->transEsc($cluster['label']) ?></h2>
```
maccabeelevine, line 9: I changed this to `h2` for consistency.  I'm not sure how it's used in practice, but I imagine I would put a static heading above it and make all of these actually `h3`s.  But I guess that would be up to an implementer, and `h2` seems safer here unless we actually add a static heading.

themes/bootstrap3/templates/Recommend/SummonResultsDeferred.phtml
maccabeelevine: Resolving as equiv to SummonResults

demiankatz (commented):
Thanks, @maccabeelevine! I definitely prefer having these consistent with one another, and I have no objection to this approach. I'm interested in @crhallberg's thoughts.

However, just to play devil's advocate, one alternative I can think of to this would be to have an h2 at the top of each recommendation area, and then use h3's to further subdivide those recommendation sections. I think this is probably a less desirable solution since it would require us to figure out wording for those top headings, etc., and it would be less likely to be universally appropriate for everyone... but just brainstorming!

maccabeelevine:
> Thanks, @maccabeelevine! I definitely prefer having these consistent with one another, and I have no objection to this approach. I'm interested in @crhallberg's thoughts.

Yes definitely.  I am also not sure whether we want any CSS changes to reflect the change to `h2`.  The size right now is parallel to the column headings for combined search, which is intuitively right, and none of the default (en.ini) headings wrap to multiple lines at full desktop width.  But @crhallberg might have a different thought. 

> However, just to play devil's advocate, one alternative I can think of to this would be to have an h2 at the top of each recommendation area, and then use h3's to further subdivide those recommendation sections. I think this is probably a less desirable solution since it would require us to figure out wording for those top headings, etc., and it would be less likely to be universally appropriate for everyone... but just brainstorming!

Yeah, this is a possibility.  I think the problem is that it implies the recommendations are in some way homogenous, which may be the case but is at least as likely not to be.  So as you say it would be hard to come up with a meta heading that matches all cases, and I'm not sure one is even semantically appropriate.  As you might also separate the recommendation modules into a few groups, each with their own `h2`.  So I favor leaving it as is but interested in @crhallberg 's thoughts.

themes/bootstrap3/templates/Recommend/FavoriteFacets.phtml
demiankatz: This is used in the account area -- it displays the tag filters when looking at favorites and lists.

themes/bootstrap3/templates/Recommend/ExpandFacets.phtml
demiankatz: I think long ago this was used by default in the Author module, but now it's probably never used (but kept around just in case). :-)

demiankatz:
> Yeah, this is a possibility. I think the problem is that it implies the recommendations are in some way homogenous, which may be the case but is at least as likely not to be. So as you say it would be hard to come up with a meta heading that matches all cases, and I'm not sure one is even semantically appropriate. As you might also separate the recommendation modules into a few groups, each with their own `h2`. So I favor leaving it as is but interested in @crhallberg 's thoughts.

I guess the most flexible solution to this problem would actually be to create a recommendation module that just outputs a heading, and then you could plug this between the other modules as/where needed. It would then be nice to have a configuration setting to control the heading level used by all the other recommendation modules, to avoid having to customize every single template (i.e. to use h3 if there's an h2 above, but an h2 otherwise). A view helper might be able to encapsulate this. I'm not suggesting that we should do any of this -- just pointing out that a possible solution exists if this use case becomes important, and in the meantime, doing things as you're doing them here doesn't really get in the way of that.

I think the only real question is whether any styling size changes are needed, as you mentioned above. I have no strong opinion one way or another, so we'll see what @crhallberg thinks!

themes/bootstrap3/templates/Recommend/FavoriteFacets.phtml
maccabeelevine: Ok thanks found it.  In this case it should actually be `h3` for consistency with Your Lists and Your Account above, unless it was intentionally meant to be a subheading of Your Lists, which maybe?

demiankatz: I guess you could argue either way regarding "Your Tags" being a subheading of "Your Lists," but I think keeping them parallel with `h3` makes sense.

themes/bootstrap3/templates/Recommend/Databases.phtml
```diff
@@ -1,7 +1,7 @@
 <div class="databases">
   <?php $databases = $this->recommend->getResults(); ?>
   <?php if (!empty($databases)): ?>
-    <h4><?=$this->transEsc('databases')?></h4>
+    <h2><?=$this->transEsc('databases')?></h2>
```
maccabeelevine: So *this* heading is too long now, it wraps at full width.  Assuming @crhallberg wants to keep the font size as is, I will work with my librarian colleague for a shorter name.  I have some thoughts.

maccabeelevine: I shortened it.

demiankatz: While we're changing this, it occurs to me that it might be wise to rename the translation key as well, to something like `databases_recommend_heading` since `databases` is pretty vague. (I think I failed to catch this earlier because it started life as `eds_databases` which seemed better-differentiated, but then things evolved a bit from there).

maccabeelevine: Agreed, done.

themes/bootstrap3/templates/Recommend/ExpandFacets.phtml
crhallberg: This is also the only one I wonder about. It seems like this would be rendered from another template with a static header, but as Demian said, it's probably not used. I'll keep an eye out.

crhallberg (approved):

themes/bootstrap3/templates/Recommend/WebResults.phtml
demiankatz: Before:

image

After:

image

... looks fine to me!

themes/bootstrap3/templates/Recommend/ExpandFacets.phtml
demiankatz: Definitely not used by default; it's an option that can be enabled.

Before:
image

After:
image

...this one might benefit from a top-level heading and making the individual sections smaller, but since I doubt anyone uses it, I don't think it's worth bothering to change. We can revisit if anyone complaints.

demiankatz (approved):
This all looks good to me (see screen shots for the bits you weren't able to test). Merging now!

PR #3210, opened 2023-11-10
Fix display of checkbox filters.

demiankatz:
I noticed that checkbox filters were displaying inconsistently with regular filters. This PR adjusts the markup for checkbox filters to more closely match regular filters, for consistent display and accessibility features.

To reproduce:

1. In facets.ini, uncomment the "First Edition" example under [CheckboxFacets]
2. In search results, apply the "First Edition" checkbox facet, and any other facet.
3. Compare the appearance/behavior of the applied filters above the search results.

### DIFF
--- a/themes/bootstrap3/templates/search/filters.phtml
+++ b/themes/bootstrap3/templates/search/filters.phtml
@@ -28,25 +28,24 @@
 <?php ob_start(); ?>
   <?php foreach ($this->checkboxFilters as $filter): ?>
     <?php if ($filter['selected']): ?>
-      <span class="filter-value">
-        <?php
-          $removeLink = isset($urlQuery)
-            ? $urlQuery->removeFilter($filter['filter'])
-            : $this->searchMemory()->getEditLink(
-                $this->searchClassId,
-                'removeFilter',
-                $filter['filter']
-            );
-          $desc = $this->translate($filter['desc']);
-          $ariaLabel = $this->translate('Remove filter') . ' ' . $this->escapeHtmlAttr($desc);
-        ?>
-        <?=$desc?>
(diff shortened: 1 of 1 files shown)


sturkel89:
In most ways, checkbox filters behave the same way as other filters; the unique quality of the checkbox filter is that it always appears in the first position among the list of applied filters, regardless of the order in which the filters were applied. 

(Every other applied filter appears from left to right, top to bottom, in the order in which the user clicks to set the filters.)

The one difference I found in behavior had to do with **mouseover to remove the filter**.

In the case of all of the "regular" filters, mousing over any part of the box around the named filter causes the entire box to change to red color with the text struck through; the mouse pointer is a finger anywhere over this color-changed box. 

image

However, the checkbox filter's behavior does not match this behavior on dev.

On dev, most of the checkbox filter's real estate does not respond to mouseover. There is no color change or strikethrough regardless of where you place your pointer. The mouse pointer becomes a finger ONLY when you mouse over the X. Clicking the X eliminates the filter, with a perceptible delay.

image

On test, the checkbox filter behaves exactly like the other filters! Mouseover changes the entire "First edition" box's color to red, strikes through text, and the pointer is a finger anywhere over that box. Mouse click eliminates the filter immediately.

image

Seems to me like it's ready to merge!

sturkel89 (approved):
See my earlier comment. I approve the merging of this commit! :)

demiankatz:
Thanks, @sturkel89 -- you describe exactly what I had intended to change, so it sounds like I did it right. :-) I'll leave this open a little longer in case @crhallberg or others have any concerns about my specific implementation, but I expect we should be able to merge this soon.

crhallberg (approved):

PR #3212, opened 2023-11-13
Add log writability check to SystemStatus.

EreMaijala:
Since errors writing to log can cause fatal exceptions, it makes sense to verify that the log file is writable (or any other logging method can be initialized). Since logger opens the log file for appending regardless of whether the logged message matches the logging level, the message doesn't need to get written for this check to work.


demiankatz (approved):
Thanks, @EreMaijala!

PR #3213, opened 2023-11-13
[9.1] Add explicit int casts to cover generator layers.

EreMaijala:
Fixes some PHP 8.1 deprecation notices.


demiankatz (approved):
Thanks, @EreMaijala -- I just made one small edit to add `|false` to the getColor typehint, since that method can also return false.

PR #3214, opened 2023-11-13
Primo: Replace commas in search terms with spaces instead of pluses.

EreMaijala:
Looks like the code has been the same for the past nine years, but perhaps something has changed in Primo to make it less lenient. In any case, this changes the commas to spaces as [documented](https://developers.exlibrisgroup.com/primo/apis/webservices/xservices/search/briefsearch/).


demiankatz (approved):
Thanks, @EreMaijala!

PR #3216, opened 2023-11-13
Databases: Include search query matches

maccabeelevine:
The new Databases recommendation module is "smart" in that it looks to the result facets to find relevant databases.  But it should also work in a simpler way -- if the user actually searches for the name of a database, that database should appear in the results as well.

### DIFF
--- a/module/VuFind/src/VuFind/Recommend/Databases.php
+++ b/module/VuFind/src/VuFind/Recommend/Databases.php
@@ -105,6 +105,13 @@ class Databases implements RecommendInterface, \Laminas\Log\LoggerAwareInterface
     protected $configFileDatabases;
 
     /**
+     * Configuration of whether to use the query string as a match point
+     *
+     * @var bool
+     */
+    protected $useQuery;
+
+    /**
      * Configuration of whether to use LibGuides as a data source
      *
      * @var bool
@@ -162,6 +169,8 @@ class Databases implements RecommendInterface, \Laminas\Log\LoggerAwareInterface
         $this->resultFacet = isset($databasesConfig->resultFacet)
             ? $databasesConfig->resultFacet->toArray() : [];
         $this->resultFacetNameKey = $databasesConfig->resultFacetNameKey ?? 'value';
+
+        $this->useQuery = $databasesConfig->useQuery ?? true;
 
         $this->useLibGuides = $databasesConfig->useLibGuides ?? false;
         if ($this->useLibGuides) {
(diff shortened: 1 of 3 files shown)


module/VuFind/src/VuFind/Recommend/Databases.php
```diff
@@ -230,6 +239,21 @@ public function getResults()
         }
         $nameToDatabase = $this->getDatabases();
         $databases = [];
+
+        // Add databases from search query
+        if ($this->useQuery) {
+            $query = strtolower($this->results->getParams()->getQuery()->getString());
```
demiankatz, line 254: Should we perhaps have a minimum length requirement for this logic? Otherwise, I'm guessing that something like a search for a single letter is going to lead to very weird results.

demiankatz (changes requested):
Thanks, @maccabeelevine!

For starters, just one comment (see below).

Also, might it be worth starting a unit test for this module? As things start to get more complex, I think there could be value in having a way to capture various use cases and avoid regressions.

module/VuFind/src/VuFind/Recommend/Databases.php
maccabeelevine: I was going to include that, but I considered that a lot of database names are acronyms, i.e. EPIC or JSTOR or are likely to be searched by partial names i.e. LEXIS or ABI (Inform), so in practice I can't see setting a minimum larger than like 3, and so what's the point.  But practicality aside, I agree with the idea so can add it if you prefer.

demiankatz: I think it might be worth making a configurable setting for it, with a default of 2 or 3, so people can decide for themselves. Acronyms aside, I also think blank searches may behave in an undesirable way without some kind of check (because str_contains always returns true if you ask for containment of the empty string -- but I would think the most commonly-used databases in the index would be a more useful default result than the first arbitrary databases listed in the configuration or API response).

maccabeelevine: Good point about blank searches especially.  Implemented.

maccabeelevine:
> Also, might it be worth starting a unit test for this module? As things start to get more complex, I think there could be value in having a way to capture various use cases and avoid regressions.

Yes definitely, this has gotten past that point of complexity.

demiankatz:
> > Also, might it be worth starting a unit test for this module? As things start to get more complex, I think there could be value in having a way to capture various use cases and avoid regressions.
> 
> Yes definitely, this has gotten past that point of complexity.

Do you want to add tests here, or do you see that as a follow-up action after we get through this PR? (I'm fine either way, just asking so I can prioritize my reviews accordingly).

maccabeelevine:
> Do you want to add tests here, or do you see that as a follow-up action after we get through this PR? (I'm fine either way, just asking so I can prioritize my reviews accordingly).

Honestly doesn't matter to me, I was assuming to add them here but can work either way.

demiankatz:
If you're not in a hurry to get this merged, might as well do it here, and then if we uncover other issues to fix, we can merge it all at once.

maccabeelevine:
> If you're not in a hurry to get this merged, might as well do it here, and then if we uncover other issues to fix, we can merge it all at once.

Yup, added test coverage.

demiankatz (approved):
Looks good to me! Thanks, @maccabeelevine!

PR #3217, opened 2023-11-14
Improve comments about solrmarc.path setting.

demiankatz:
There have been some recent problems where the solution turned out to be changing the order of the SolrMarc search path. This PR adds a comment to make the correct configuration more clear.


maccabeelevine (approved):
I haven't run into this problem, but I would definitely be at least unclear about the order, and that line definitely helps.

PR #802, opened 2016-09-22
Form attribute support

crhallberg:
Add support for the [`form="id"` attribute on input elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#attr-form). This new feature is then applied to [remove the hack](https://github.com/vufind-org/vufind/pull/802/files#diff-8c13fdc465a565f98bde2c279226f8f8L69) that allowed us to use forms within forms for embedded records.

Unfortunately, IE removed support from conditional loading after IE 9. This means we need to load the [polyfill](https://github.com/vufind-org/vufind/pull/802/files#diff-5) for all browsers and skip the logic addition for those that [support it](http://caniuse.com/#feat=form-attribute). The asset pipeline makes this negligible for production performance.
- [x] IE testing
    - [x] Duplicate fields if multiple form submits occur (lightbox)
    - [x] Bottom bar not working
    - [x] Print taking a long time
- [x] Merge master and Run full test suite
- [x] Fix polyfill
- [x] Test jQuery serialize
- [x] Search for more form serialization places
- [x] Add search form parameter
- [x] idPrefix
- ~~IDs or classes (discussion at VuFind)~~ in the [wrong PR](https://github.com/vufind-org/vufind/pull/814)
- [x] recaptcha logic consolidation

PR #3219, opened 2023-11-14
Combined: Fix scope of search cart form

maccabeelevine:
In search/results.phtml the form element
`<form id="search-cart-form" ...`
wraps just a single line, the call to render search/bulk-action.buttons.phtml.

In combined/results.phtml, the same form wraps most of the whole file.  It looks like search/results.phtml used to do the same, but was refactored in #802.

This is a problem just now because after #3135 one of the things it wraps is side recommendations, which can include SideFacets.phtml, which can include range-slider.phtml, which defines its own form.  It's invalid HTML to have a form nested within another form, and forces the browser to move the entire sidebar div outside of the search-cart-form.  This in turn messes up the styling assumptions where div.mainbody and div.sidebar are both floated and assumed to be adjacent elements.

This PR changes combined/results.phtml so the form scope mirrors search/results.phtml.

### DIFF
--- a/themes/bootstrap3/templates/combined/results.phtml
+++ b/themes/bootstrap3/templates/combined/results.phtml
@@ -52,51 +52,51 @@
 ?>
 <?=$this->flashmessages()?>
 <h1 class="sr-only"><?=$this->escapeHtml($headTitle)?></h1>
+<?php $recs = $combinedResults->getRecommendations('top'); ?>
+<?php if (!empty($recs)): ?>
+  <div>
+    <?php foreach ($recs as $current): ?>
+      <?=$this->recommend($current)?>
+    <?php endforeach; ?>
+  </div>
+<?php endif; ?>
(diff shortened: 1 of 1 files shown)


maccabeelevine:
I am admittedly no expert on how the search cart form works.  I turned on showBulkOptions in config.ini, and I was still able to use one of the bulk operations (export).  Not sure what else to test.

demiankatz (approved):
Thanks, @maccabeelevine, it definitely makes sense to fix the inconsistency between the combined and regular templates. I've confirmed that the full test suite passes and that manual testing of combined cart and bulk functionality reveals no problems after this change.

Even though this is a bug fix, it does not appear to cleanly backport to the release-9.1 branch due to other changes to combined search on the dev branch. Thus, I'm going to merge it to dev for inclusion in release 10.0; I think the situation where this will cause a problem is rare enough that the fix can wait for the major release. Of course, I'm happy to revisit that decision if anyone needs me to!

PR #3220, opened 2023-11-14
Combined: Fix column widths in grid layout

maccabeelevine:
This addresses two issues with the CSS grid layout for combined search added in #3077 that occur when one column happens to have a value that can't be wrapped, and is longer than the intended width of the column.  (The intention is that each column have equal width, as with the other combined column modes.)  This happens in practice with EDS data that has a long, visible URL.

1) When one column is forced longer than expected, but the combined width of all columns is still small enough to fit in .mainbody.  The result is the columns are uneven.

image

2) When once column is forced so large that the total columns can't fit in .mainbody.  Then the sidebar overlaps the main content.  This example is fake but I've seen it happen for real. 

image


themes/bootstrap3/less/components/search.less
```diff
@@ -438,6 +438,10 @@ body.rtl {
 .combined-search-container.grid {
   margin-left: 1rem;
   margin-right: 1rem;
+
+  .combined-list {
+    overflow-x: auto;
+  }
```
maccabeelevine: I don't love the horizontal scrollbar, but it's better than overlapping content.  clip or hidden would be option if @crhallberg prefers.

themes/bootstrap3/templates/combined/stack-grid.phtml
```diff
@@ -1,5 +1,5 @@
 <div class="combined-search-container grid"
-  style="grid-template-columns: repeat(<?=$this->escapeHtmlAttr($columns)?>, 1fr)"
+  style="grid-template-columns: repeat(<?=$this->escapeHtmlAttr($columns)?>, minmax(0, 1fr))"
```
maccabeelevine, line 2: I'm really not clear why setting an explicit minimum has the effect of enforcing the expected maximum.  Maybe someone else can parse MDN better.  https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns

crhallberg: It looks like minmax is overriding the overflow consideration. It is strictly setting the width rather than let them grow. It seems to be the better option.

themes/bootstrap3/less/components/search.less
crhallberg: I also agree that the overflow scrollbar is not good. I'd prefer hidden but there's more we can do, I think. Since this happens when long words come into play, we can allow for more breaks in words:

```css
.combined-list {
    word-break: break-word;
    hyphens: auto;
}
```

We can also do [what we do locally at Villanova](https://library.villanova.edu/Find/Combined/Results?type=&lookfor=behold&limit=20&sort=relevance) and make the container on the combined screen full width.

maccabeelevine:
Ugh sorry I killed this with a manual merge from dev, I forgot we had rebased against 9.1.  Will try to fix or if not will open a new branch off 9.1, it's a tiny PR.

maccabeelevine:
Replaced with #3232

PR #3222, opened 2023-11-15
Add bootstrap5 theme

EreMaijala:
This is initial work for adding a bootstrap5 theme that uses Bootstrap 5.x. and to include enough back-compatibility support to make it possible to inherit other themes from it. Still pretty much work in progress but comments welcome.

Principles:
- Keep template changes from bootstrap3 to minimum, and only do string replaces
- Add bs3-compat.scss and bs3-compat.js as a compatibility layer

Here are the most significant template changes:

data-toggle="collapse" => data-bs-toggle="collapse"
data-toggle="dropdown" => data-bs-toggle="dropdown"
data-target => data-bs-target
data-dismiss="modal" => data-bs-dismiss="modal"

Other things worth noting:
- `visually-hidden` replaces the `sr-only` class, but the latter is still available via bs3-compat.scss
- font-awesome has been updated (`$fa-font-path` has changed too).
- popover, tooltip, collapse and modal are different, so the following files have differences between bootstrap3 and bootstrap5: 
  - account_ajax.js
  - cart.js
  - channels.js
  - facets.js
  - lightbox.js
- Where bootstrap3 uses the `in` class, bootstrap5 uses `open`
- legend does not have huge bottom margin in bootstrap5
- dark theme does not work properly at the moment
- header is a bit different and needs manual adjustment from bootstrap3
- Tabs now have the `active` class on the `a` element instead of `li` (bs3-compat.js handles some of this, but style changes may be required as well).
- `npm run build:scssonly` can be used to compile scss for scss-only themes (bootstrap5, sandal5). Also `npm run build:css` does that.
- compiling scss requires `npm install` to be run in the theme directory first

TODO:
- [x] Commit templates before merging.
- [x] Do we want to clean up LESS/SCSS conditionals from scss files or keep them for parity with bootstrap3/sandal?
- [x] Ensure that there are no conflicts with Bootstrap 5 and VuFind themes regarding accessibility (Wave and/or Axe could be useful tools)
- [x] [Document](https://vufind.org/wiki/development:architecture:user_interface:bootstrap5#template_changes) changes required to templates
- [x] [Document](https://vufind.org/wiki/development:architecture:user_interface:bootstrap5#sass_only) changes required with less to sass transition
- [x] Update channels.js with latest changes from bootstrap3
- [x] Enable check-bootstrap5 in QA before merging.
- [x] When merging, add changelog note about theme transition plan (bootstrap5 = beta in 10.0; bootstrap3 will be deprecated in 10.1 and bootstrap5/sandal5 promoted to default; bootstrap3 and its children will be removed in 11.0)

TODO (Issues):
- [x] [White on white text](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2064030558) - VuFind link at top left and theme/language dropdowns appear as white-on-white (Bootstrap5 theme only).
  - Should be fixed.
- [x] [Similar Items](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2064120165) - formatted incorrectly in sidebar and record tabs (Bootstrap5 and Sandal5).
  - Should be fixed.
- [x] [Field selection dropdown in search box](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2064410537) - out of alignment (Bootstrap5 and Sandal5).
  - Should be fixed.
- [x] [Print Record](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2064504199) - formatting is strange (Bootstrap5 and Sandal5).
  - Should be fixed.
  - I'm still seeing the issue in both Chrome and FF, with cleared caches - ST 
  - Oops, I was looking at the actual print results, fixed again. -EM
  - Big improvement! Remaining nitpick: I've posted an updated screenshot below showing that **text color** still varies in print preview (everything is monochrome in legacy themes) so the last thing to possibly fix is text color. - ST
  - Should be fixed as well now (BS5 makes color utils !important by default, had to disable that). -EM
- [x] [Format and availability lozenges](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2064735378) - font too small (Sandal5, when compared with Sandal).
  - Should be fixed. -EM
- [x] [Floating "back"](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2064758161) - extra link text at top of filter columns (Bootstrap5 and Sandal5).
  - Should be fixed (was a problem in dev branch, see #3573).
  - I'm still seeing the issue in both Chrome and FF, with cleared caches - ST 
  - Missed a file in commit, should be properly fixed now. -EM
- [x] [Main VuFind logo](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2066651170) - The main VuFind logo needs more emphasis and possibly the registered trademark symbol (Bootstrap5 needs bold/etc., both Bootstraps lack the registered trademark symbol) 
  - Underlining and a bit more emphasis added. -EM
- [x] [Print preview for multi-selected items in Favorites list](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2066697629) - Print view for selected items in Favorites lists contains graphics that don't render correctly. (Bootstrap5 and Sandal5)
  - Should be fixed now along with the other print issues. -EM
- [x] [Record tabs ghost highlighting](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2066720008) - Previously selected tab still looks selected (but with slightly different styling than the actually selected tab). (Bootstrap5 and Sandal5)
  - Should be fixed. -EM
- [x] [Contrast in form element borders](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2067092225) - Borders outside search boxes and other form elements are not contrasty enough. (Sandal5, and also Bootstrap5)
  - Focus outline for sandal5 fixed, but I can't see the other issues. -EM
  - I see a small difference in the two bootstrap versions, but I think the bootstrap5 version is slightly MORE contrasty. No change needed (unless someone who knows accessibility best practices better says differently.)
- [x] [Hover behavior in header](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2067120306) - Links in header need updates to styling in new themes (Bootstrap5 especially; also Sandal5)
  - Should be much better now. However, the hover highlight is not quite the same between bootstrap3 and bootstrap5, and I'm not sure it should be either. Now everything has the same hover effect. -EM
- [x] [Advanced search limit options](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2067141364) - Height of this element needs to be increased. (Bootstrap5 and Sandal5)
  - Fixed. -EM
  - Yes, you fixed those three... but I just noticed that in [New Items](http://localhost/vufind_test/Search/NewItem?ui=sandal5), there's a similar element for Department that needs to be expanded in the same way in the new themes.
  - Now fixed as well. -EM
- [x] [Links in Channels don't work](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2067249367) - Links in the pop-out item menus don't work when clicked. (Bootstrap5 and Sandal5)
  - Fixed. -EM
- [x] [Problematic click behavior in Your Account menus](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2070030903) - Clicking a menu item results in illegible black boxes (Bootstrap5, in contrast with Bootstrap3)
  - Fixed. -EM
- [x] [Mouseover is black in Browse and Alphabrowse](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2070314783) - Hovered-over menu items are illegible due to black highlighting; in legacy version, highlighting is gray. (Bootstrap5)
  - Fixed. I agree that the highlight color is odd, but it's equally odd in old sandal. -EM
  - I'm still seeing hovered-over menu items as having black highlighting in bootstrap5 - even on a browser I've never used before. Can you check?
  - Had to tweak nav-link colors and make the navbar ones a special case instead of the other way round. -EM
- [x] [Narrow dropdown on System Maintenance page in Admin area](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2077629439) - The dropdown box in the last row on this page is not wide enough to display the word "standard" in any of the themes.
  - Fixed. The width is now relative to the font size, and padding has been reduced to allow for more text in select fields in general. -EM 
  - On recheck, I see that the entire word "standard" is now visible in bootprint3 and in sandal5. However, the final 'd' is cut in half in the two bootstraps, and in sandal "standard" ends in the middle of the 'r.' So, a little more tweaking? -ST
  - Tweaked more for bootstrap5 and sandal5. I'm avoiding touching the old themes here. -EM
  - The word "standard" is fully visible now in bootstrap5 and sandal5! This is done. :)
- [x] [Extra X in search box while typing](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2121094458) - There's an extra X to clear your search in the searchbox. (Bootstrap5 and Sandal5)
  - Should be fixed now. -EM
  - Confirmed - fixed! - ST
- [x] [Facets filtering lightbox layout](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2121112211) - This lightbox has issues in the new themes - see below. (Bootstrap5 and Sandal5)
  - Should be fixed now. -EM
  - Confirmed - fixed! - ST
- [x] [Responsive table for search history](https://github.com/vufind-org/vufind/pull/3222#issuecomment-2121150448) - The enhancements from [PR 3204](https://github.com/vufind-org/vufind/pull/3204) are not fully implemented in the new themes. (Bootstrap5 and Sandal5)
  - Should be fixed now. -EM
  - Confirmed - fixed! - ST


crhallberg:
I wonder if we can reduce the template changes even further by re-mapping the data attributes with bs3-compat.js.

```js
const dropdownElementList = document.querySelectorAll('[data-dropdown]');
dropdownElementList.forEach((dropdownToggleEl) => {
   const target = dropdownToggleEl.getAttribute("data-target");
   if (target) {
       dropdownToggleEl.setAttribute("data-bs-target", target)
   }
});

const dropdownList = [...dropdownElementList].map(
    (dropdownToggleEl) => new bootstrap.Dropdown(dropdownToggleEl)
);
```

For the templates that aren't changed from bs3, do we want to inherit those to make it clear which templates we need to update when we make changes?

EreMaijala:
> I wonder if we can reduce the template changes even further by re-mapping the data attributes with bs3-compat.js.

I haven't actually tried it yet, but I suspect we'd have trouble with the initialization order of the scripts. We could make bs3-compat run early, but then it could miss something. We'd also have to employ mutation observers for any dynamically loaded content. But I have a couple of other ideas for runtime patching..

> For the templates that aren't changed from bs3, do we want to inherit those to make it clear which templates we need to update when we make changes?

My aim currently is to make it so that you could run a script to update bootstrap5 from bootstrap3. This would allow the theme to be completely standalone to avoid any confusion. But inheriting bootstrap3 is certainly an option, and I'm not committed to any specific approach yet. :)

EreMaijala:
Regardless of the approach, we could of course leave all template files out of git if the script works. That'd make this PR quite a bit smaller.

demiankatz:
> Regardless of the approach, we could of course leave all template files out of git if the script works. That'd make this PR quite a bit smaller.

That would definitely make review easier, though of course we'll want to commit the files in the end. (If we're able to go with a script-based approach, we should also add a CI test similar to lessToSass to ensure nobody forgets to keep the files in sync). In any case, I definitely like the fool-proof nature of the script-based approach, if we're able to make that work... and if not, @crhallberg's idea of inheriting files that are unchanged seems like the next best thing.

sturkel89:
I have run the full test suite on this PR. Here are the results:

**ERRORS!**
Tests: 2567, Assertions: 10663, Errors: 37, Failures: 8, Skipped: 64.

Please let me know if you'd like me to paste more detail here in the PR.

EreMaijala:
I've now added a `vendor/bin/phing update-bootstrap5-templates` task and removed the templates from git. Next step would be to fix pagers.

demiankatz:
@EreMaijala, I see that GitHub is complaining about some things in the Javascript. Are these new issues, or is it re-reporting things we had already investigated in bootstrap3?

crhallberg:
@demiankatz it looks like [Javascript that is in bootstrap3](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/js/embedded_record.js#L46)

EreMaijala:
With the latest updates this is now in a much better shape, though I'm sure there's still a lot I've missed so far.

demiankatz:
Thanks, @EreMaijala! I've challenged @sturkel89 to look around and see if she can find anything broken... :-)

sturkel89 (changes requested):
I added bootstrap5 as an option in my theme-switcher version of config.ini. 

The first thing I noticed in the test branch is the **fun new color** of the lozenge labels for format in bootstrap3! 

(Actually, I prefer the more sedate dark blue, but it's good to have a visual indicator of which theme we're viewing and which branch we're in.)
image

------

The second thing I noticed is that in the test branch, **the AVAILABLE lozenge seems to be missing in three out of four themes!** (On the dev branch, the AVAILABLE lozenge appears as expected in all three available themes.)

**Explanation:**
The AVAILABLE lozenge uses white text, and has background shading in bootstrap3 only. In the other three themes, it is missing its background shading. See below:

### bootstrap3:
image


### bootprint3:
image


### sandal:
image


### bootstrap5:
image

@EreMaijala, maybe this is an easy fix? Thanks!

sturkel89 (changes requested):
Here's one more weirdness I noticed in bib record view, as a bonus:

### Test branch
In the test branch, bootstrap3 displays the text about tags and the **+Add Tag** "button" as two columns within the central column of the record view. The other themes display these two elements in a vertical pile.

**bootstrap3:**

image

**bootprint3:**

image

**sandal:**

image

**bootstrap5:**

image

### Dev branch:
In the dev branch, every theme shows the split column:

**bootstrap3:**

image

**bootprint3:**

image

**sandal:** (NB: The spacing on the text in _No Tags, Be the first to tag this record!_ is wonky here. This is probably true in the test branch as well; it's only visible here because of word wrap!)

image

crhallberg:
Gave the changes here a read today. I know that `dropdown-item` isn't required by BS5 anymore, but do we want to keep it for easier customization?

EreMaijala:
> Gave the changes here a read today. I know that `dropdown-item` isn't required by BS5 anymore, but do we want to keep it for easier customization?

I think it's worth keeping, but we can revisit that later.

sturkel89:
In the bootstrap5 theme, the VuFind logo in the upper left corner of the screen in every type of view is present in white-on-white and therefore is invisible. The same is true for the theme's name in the theme selector dropdown and the current language in the language selector dropdown.

This quirk is not an issue in Sandal5, because that uses the **vufind(r) search.discover.share.** icon in the upper corner, and contrasting highlighting for the selected theme and language.

image

image

sturkel89:
**Similar Items** in the right sidebar display strangely in bootstrap5 and sandal5; the padding and separators that are present in all the other themes are missing.

**Bootstrap5**
image

**Bootstrap3**
image

**Sandal5**
image

**Sandal**
image

This looks even worse with more fleshed-out similar items:

image

-----------------------

It's even weirder in the footer. The top of the section looks like this, and then there are several screens below containing row after row of single similar items with the 'no image available' icon is centered on the page.

**Bootstrap5**
image
In **Bootstrap3**, there's a slider. I guess the slider isn't working correctly in Bootstrap5 or Sandal5.

image

sturkel89:
The **field selection dropdown** in the search box at the top of results and item pages (and to a lesser extent on the home page) is a few pixels higher than it should be, in Bootstrap5 and Sandal5. Also, the text is not centered within the box.

These things are aligned perfectly in the legacy themes in the test branch.

### Examples:

**Sandal5**
image

**Sandal**
image

**Bootstrap5**
image

**Bootstrap3**
image

sturkel89:
The **Print record** view in the new themes is not working correctly. I'm comparing themes within the new branch.

**Bootstrap3 - looks normal:**
image

**Bootstrap5 - all text wants to have a white background:**
image

I could provide examples comparing Sandal and Sandal5, but the effect is exactly the same there.

------------
### UPDATE

The shading is gone and the print preview looks much better! 

My only remaining comment is that in the legacy themes, EVERYTHING in the print preview is monochrome, while in the new themes text color can vary.  

See below - with the demo driver, there are all sorts of unusual holdings, and electronic holdings apparently present availability in green. This screenshot is from the print preview view in Sandal5. In old Sandal, all the text is black.

This probably doesn't matter, but for consistency you might change that in the code.

image

sturkel89:
The font inside the availability and format lozenges in Sandal5 is **extremely tiny**. It's noticeably smaller than the font in the lozenges in Sandal. 

(I also prefer the green color in Sandal to the gray color in Sandal5.)

**Sandal**


**Sandal5**

sturkel89:
### Floating "back"
There is a strange floating "back" text link at the top of the filters column in Bootstrap5 and Sandal5. This text link does not appear in the legacy themes.

**Bootstrap5**
image

**Bootstrap3**
image

**Sandal5**
image

**Sandal**
image

sturkel89:
### Main VuFind logo
The VuFind logo is now visible in the upper left corner of the screen in Bootstrap5. However, it's much less prominent than the one on Bootstrap3 which is larger, bold, and underlined to show it's a hyperlink. 

Also, do we want to add the **registered trademark symbol** in the two Bootstrap themes?

**Bootstrap5:**
image

**Bootstrap3:**
image

sturkel89:
### Print view for multi-selections on Favorites list

When I'm in a Favorites list and select multiple items, I can ask to print that selection. The print preview display in Bootstrap5 and Sandal5 include lozenge shapes that have contrasting text on them. This was fixed in an earlier PR so that print preview has none of those graphics. Can we make that same change to this feature in these two new themes?

I think this is related to the still-present issue with Print Record; the white background surrounding text in some fields would not show if all the shading were removed.

**Bootstrap5**
image

**Bootstrap3**
image

sturkel89:
### Record tabs ghost highlighting

In the new themes, previous tabs remain highlighted when you switch and click on another tab. There is always one recently-viewed tab that shows highlighting and the changed text color, in addition to the currently selected tab. This occurs with both Bootstrap5 and Sandal5, and not in the legacy themes.

**Bootstrap5**
image

**Bootstrap3**
image

sturkel89:
### Contrast in form element borders

The borders of search boxes and other form elements have less contrast in the new themes than in the legacy themes. 

I think the outlines of the form elements in the new themes need to be set to match those in the legacy themes on dev.

There was a PR a few months ago that led to the more contrasty versions ([PR 3207](https://github.com/vufind-org/vufind/pull/3207)) that you can reference.

**Sandal5**


**Sandal**


**Bootstrap5**


**Bootstrap3**


I think the orange outline in Sandal is very visible. It's not as clear to my eye that the search box outline on Bootstrap3 is an improvement over Bootstrap5, so I'm not sure about that one. I wanted to document the difference.

-----------
### UPDATE

Thanks for fixing the issue in Sandal5! That was the important one. 

I see a DIFFERENCE between the focus outline in bootstrap3 and bootstrap5, but to my eye the bootstrap5 version is MORE visible than the other. I don't think it needs fixing, unless someone who knows more about accessibility thinks the other version is better.

sturkel89:
### Hover behavior in header

With all options enabled and a logged-in user, there are four items in the header area: two links (Your Account and Log Out) and two dropdowns (Theme and Language). In the legacy themes, when you mouse over any of those four elements, you see identical contrasting shading in all cases.

In the two new themes, mouseover of the **links** behaves differently from mouseover of the **dropdowns**. 

In Bootstrap5, the shading for mouseover of the Your Account and Log Out links renders them difficult to read: white on very light gray. In Sandal, the items are legible, but it's clear that they're highlighted in a different color than the dropdowns.

**Bootstrap5**
image
image


**Bootstrap3**
image

**Sandal5**
image
image

**Sandal**
image

sturkel89:
### Advanced Search limit selections - height needs adjustment

On the Advanced Search screen, there are three browsable "limits" available: **Call Number**, **Language**, and **Format**. 

In the legacy themes, each of these elements is set to display the **first 10 values** on the options list.

In the new themes, each of these elements is tall enough to show **one value** plus a few pixels of another value. It's still possible to scroll through the list, but I think we want these fields to be 10 rows high.

**Bootstrap5**


**Bootstrap3**


**Sandal5**


**Sandal**


-----------------------------
### EDIT: one more field to fix!

The **New Items** search page has the same three elements as Advanced Search, plus the additional field **Department** which also needs to be expanded in Sandal5 and Bootstrap5.

image

sturkel89:
### Links in Channels don't work

The new themes include improved versions of the pop-out informational elements for items on Channels pages. The new version appears to the side of the item rather than below (which usually displays below the fold). It also looks more modern.

_The problem is that the links to **View Record** and **Explore related channels** can't be clicked_. (You _can_ click, but all that happens is that the pop-out menu disappears and you're still on the Channels page.

The new themes DO NOT display the bug ([VUFIND-1686](https://openlibraryfoundation.atlassian.net/browse/VUFIND-1686)) I've discovered across all of the legacy themes, which is that if you click on a given item on the Channels page, click away, and then click again, the second and subsequent viewings show an abbreviated menu with no content aside from the **View Record** and **Explore related channels** buttons. (In most cases, the links in the abbreviated menus are not clickable either.)

-----------------
Here's a comparison between the new and old full menus. 

### **Sandal:**




### **Sandal5:**

sturkel89:
### PROBLEMATIC CLICK BEHAVIOR IN MY ACCOUNT MENUS (bootstrap5)

### Bootstrap3
In bootstrap3, the **hover color** in the main screen dropdowns is black text on a light gray background. This differs from the appearance of the **four buttons** across the top of the screen (Your Account, Log Out, Theme, and Language), which are white text on a black button when moused over. The **active theme** is indicated by white text on a dark blue background. The **moused-over theme** or language in the dropdowns is white text on a light gray background. When you click to change theme or language, the selection retains its black on light gray appearance and then the menu disappears when the change is applied.

Screenshot of bootstrap3 appearance:
image

### Bootstrap5
The **hover color** in bootstrap5 main screen dropdowns is white text on a black background, to match the color scheme of the **four buttons** across the top of the screen (Your Account, Log Out, Theme, and Language). The **active theme** or language is white text on a dark blue background. When you click to change theme or language, the selection retains its white on black appearance and then the menu disappears when the change is applied.

In this screenshot, the active theme is bootstrap5, and the mouse just hovered over bootprint 3 so it displays in black:
image

So far, so good!

Now, let's look at the **Your Account** area in the two themes.

### Bootstrap5
In bootstrap5, the **active menu item** is white text on a dark blue background, as in the top menu. The **moused over menu item** is black text on a gray background. 

image

When you **click** on that menu item, the text becomes dark blue on a black background and is pretty much illegible.

image

### Bootstrap3
In bootstrap3, the **active menu item** is white text on a dark blue background, as in the top menu. The **moused over menu item** is black text on a gray background. When you **click** on that menu item, the button retains its black text over a gray background appearance, and then the selection changes so the newly active menu item is white text on a dark blue background.

image

I'd suggest that the click behavior in bootstrap5 should be changed to be the same as bootstrap3; let the button look the same upon click as it was upon mouseover, so that you don't have that moment of illegibility.

sturkel89:
### Mouseover is black in Browse and Alphabrowse (bootstrap5)

Hovered over text in Browse and Alphabrowse in bootstrap5 is illegible due to being highlighted in dark text on a black box.

image

It looks fine on bootstrap3:
image

Can you change the highlight to gray instead of black on bootstrap5? 

----------------
**Side note:** 
Hover text is fine in Sandal and sandal5, although I'm not a fan of the very bright highlight color on the active/selected items in each menu! I don't think that color appears anywhere else in VuFind themes. Feel free to ignore this editorial comment!

image

sturkel89:
I looked at the Admin area in the four relevant themes. Overall, everything looks fine. There are some styling differences between the legacy themes and the new themes, but nothing radical. (I think the new themes look a bit better, with less boldface.) 

The only issue is a dropdown box in the System Maintenance window that is never wide enough to display the entire word "standard" -- it's in the bottom row. Is this something you'd fix here, so it will look better in the future? Seems way too minor to make a Jira ticket...

### Bootstrap3
image

### Bootstrap5
image

### Sandal
image

### Sandal5
image

sturkel89 (approved):
All of my issues to date have been addressed - thanks, @EreMaijala ! 

I think we're in shape to merge this branch, and then we'll continue to move forward.

config/vufind/config.ini
```diff
@@ -43,6 +43,8 @@ titleSeparator = "::"
 ;                (named after the earlier, now-deprecated blueprint theme)
 ;   sandal     = bootstrap3 theme with a "flat" styling applied (a newer look
 ;                than bootprint3).
+;   bootstrap5 = HTML5 theme using Bootstrap 5 with minimal styling (beta).
+;   sandal5    = Like sandal, but based on bootstrap5
```
demiankatz: Should we mark this beta as well?

demiankatz (changes requested):
@EreMaijala, one nitpick below, and one larger question before we merge this:

What is the best way to keep things in sync going forward, once we merge this? Is the idea that we'll run the automatic template conversion process every time a bootstrap3 template changes? If so, should we add something to the GitHub Actions setup to test for uncommitted bootstrap5 changes so we don't forget (e.g. similar to our current lessToSass check)?

EreMaijala:
@sturkel89 Thanks! I'll merge the latest changes from dev at least once more so that we're fully up to date.

config/vufind/config.ini
EreMaijala: Yes, I'll make sure to do that.

EreMaijala:
@demiankatz Something like that, or the other way round (update bootstrap3 from bootstrap5). Regardless of that, I think a check for inconsistencies would be very useful. It could also be applied to most js and scss files. At least until the themes start to diverge on purpose.

demiankatz:
> @demiankatz Something like that, or the other way round (update bootstrap3 from bootstrap5). Regardless of that, I think a check for inconsistencies would be very useful. It could also be applied to most js and scss files. At least until the themes start to diverge on purpose.

Great! I think at least at first, it makes sense to test from bootstrap3 --> bootstrap5, since existing work in progress is going to have bootstrap3 template changes, and it will probably be easier to make a transition if we start in that direction. At some point, we may want to decide to flip the direction once we've merged this work into all open PRs and gotten things in sync.

Let me know if you'd like any help on setting up a mechanism!

EreMaijala:
@demiankatz There's now phing task check-bootstrap5 that checks templates, js and scss files (apart from those known to have differences):

templates:
- header.phtml
- RecordTab/similaritemscarousel.phtml

scss:
- bootstrap.scss
- components/similar-carousel.scss
- vendor/**

js:
- account_ajax.js
- cart.js
- channels.js
- facets.js
- lightbox.js
- vendor/bootstrap-accessibility.min.js
- vendor/bootstrap.min.js

demiankatz (changes requested):
When I run the full test suite with -Dtest_theme=sandal5, I get a few errors/failures:

```
There were 2 errors:

1) VuFindTest\Mink\RecordVersionsTest::testVersions
Exception: Failed to get text after 6 attempts.

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:765
/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:684
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordVersionsTest.php:68
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordVersionsTest.php:91

2) VuFindTest\Mink\RecordVersionsTest::testVersionsInSearch2
Exception: Failed to get text after 6 attempts.

/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:765
/usr/local/vufind/module/VuFind/src/VuFindTest/Integration/MinkTestCase.php:684
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordVersionsTest.php:68
/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/RecordVersionsTest.php:101

--

There was 1 failure:

1) VuFindTest\Mink\CollectionsTest::testContextLinks
--- Expected
+++ Actual
@@ @@
-'http://localhost/vufind_test/Collection/subcollection1/HierarchyTree#tabnav'
+'http://localhost/vufind_test/Collection/subcollection1/HierarchyTree'

/usr/local/vufind/module/VuFind/tests/integration-tests/src/VuFindTest/Mink/CollectionsTest.php:166

ERRORS!
Tests: 2816, Assertions: 17962, Errors: 2, Failures: 1.
```

Are you seeing the same thing? Is this expected?

sturkel89:
I did some additional testing of this branch today and found a couple of new (?) issues with the bootstrap5 and sandal5 themes that don't appear in the legacy themes.

### Extra X in search box while typing

I discovered this while testing the Virtual Keyboard but then realized it's an issue even without that feature enabled. In the two new themes, an extra X appears while entering text and seeing search suggestions. There's only one X visible under the same circumstances in the three legacy themes.

**_GOOD - LEGACY THEMES_**
**Sandal**
image

**Bootstrap3**
image

**_NOT SO GOOD - NEW THEMES_**

**Sandal5**
image

image

**Bootstrap5**
image

**Hebrew example shows different locations of the two Xes**
image

sturkel89:
### Facets filtering layout

The layout of the facets lightbox looks different in the new themes vs. the legacy themes. I am not sure that this is a big problem, but it looks strange in the new themes and I think something may be wrong with the styling -- especially since the lightbox behaves better after you resize the browser.

NB: facet sorting, filtering, and clearing the search box all behave normally in every theme.

Examples come from Sandal vs. Sandal5, but the same is seen in Bootstrap3 vs. Bootstrap5.

**GOOD BEHAVIOR - LEGACY THEMES**
In sandal, the Sort and Filter boxes are aligned in a single row, and the contents of the long table are scrollable within a lightbox that is fully visible on the screen. 

When you resize the browser window in the legacy themes, the behavior of the lightbox stays the same – it just gets taller or shorter, as appropriate.

image

**QUESTIONABLE BEHAVIOR - NEW THEMES**
In sandal5, the Sort options are on one line and the Filter box is on a second line. The contents of the table are in VERY long lightbox that doesn't fit on the page at first.

image

When you resize the browser window, either making it longer or shorter, the scroll bar appears on the right and the CLOSE button pops up from the bottom.

image

This post-resizing appearance/behavior is more in line with what I'd expect. I would prefer that the Sort and Filter options appear on one line at the top of the lightbox in fullscreen.

sturkel89:
### Responsive table in search history - not fully implemented in new themes

In reference to [PR 3204](https://github.com/vufind-org/vufind/pull/3204), each item in the search history should be displayed with information in different shades in alternating rows in mobile view. There should be space between the each "recent search", and each item should have a bit of shading to give it a slightly 3D/raised appearance

Search history looks fine in the legacy themes, but in the new themes some of this styling is missing. These screenshots show Bootstrap*, but the same is found in the two versions of Sandal.

**Bootstrap3**

image

**Bootstrap5**

image

EreMaijala:
@sturkel89 Thanks for reporting the issues. They should be fixed now.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/Search.php
```diff
@@ -38,7 +38,7 @@
  * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
  * @link     https://vufind.org/wiki/development Wiki
  */
-class Search extends \VuFind\View\Helper\AbstractSearch
+class Search extends \VuFind\View\Helper\Bootstrap5\Search
```
demiankatz, line 41: Did you forget to remove the code from this class? It looks to me like the contents are identical to the Bootstrap5 version, but maybe I'm missing something.

demiankatz (changes requested):
I think I may have spotted one minor oversight. Other than that, we may be ready, though I'll give @crhallberg a little longer to comment as well.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/Search.php
EreMaijala: And I thought I even double-checked what I did... but yes, I forgot. Now fixed.

build.xml
```diff
@@ -958,4 +960,154 @@ ${git_status}
       </filterchain>
     </reflexive>
   </target>
+
+  <!-- Update bootstrap3 templates from bootstrap5 -->
+  <!-- N.B. This destroys any existing bootstrap5 templates (except explicitly excluded ones) -->
```
crhallberg: "N.B. This destroys any existing _bootstrap3_ templates"?

themes/bootstrap5/js/bs3-compat.js
```diff
@@ -0,0 +1,150 @@
+/*global VuFind*/
+VuFind.register('bs3-compat', function cookie() {
```
crhallberg: Could we consider a more descriptive name for this function? Other ones have object-like names like ItemStatuses.

It's a shame (my shame specifically) that you needed to register a module to make sure the timing of this correct. I should add an `init` and a `ready` event to common.js.

Also, not a priority, but some of the below might be able to be moved into build.xml. Eventually.

themes/bootstrap5/js/facets.js
```diff
@@ -0,0 +1,301 @@
...
+              );
+            }
+          } else if (typeof facetData.html !== 'undefined') {
+            $facetContainer.html(VuFind.updateCspNonce(facetData.html));
+            activateFacetBlocking($facetContainer);
+          }
+          $facetContainer.find('.facet-load-indicator').remove();
+        });
+        VuFind.lightbox.bind($('.sidebar'));
+        VuFind.emit('VuFind.sidefacets.loaded');
```
crhallberg, line 177: A favor for me: this shouldn't have `VuFind.` at the front.

demiankatz (changes requested):
There's no new feedback from me, but I see @crhallberg has caught a few more small things, so I'm going to put this into "request changes" mode until you have a chance to address his feedback.

themes/bootstrap5/templates/header.phtml
```diff
@@ -0,0 +1,117 @@
...
+                  <span class="full<?=!$cart->isFull() ? ' hidden' : '' ?>">(<?=$this->transEsc('bookbag_full') ?>)</span>
+                </span>
+              </a>
+            </li>
+          <?php endif; ?>
+          <?php if (is_object($account) && $account->loginEnabled()): // hide login/logout if unavailable ?>
+            <?php if ($account->getIdentity()): ?>
+              <li class="logoutOptions<?php if ($account->dropdownEnabled()): ?> with-dropdown<?php endif ?>">
+                <a href="<?=$this->url('myresearch-home', [], ['query' => ['redirect' => 0]])?>" class="icon-link">
+                  <span id="account-icon" class="icon-link__icon" role="link"><?=$this->icon('my-account') ?></span>
```
crhallberg, line 36: I'm unsure about the addition of `role="link"` here. Since we're inside a link already, is it superfluous?

EreMaijala: Sort of, yes, but Bootstrap's tooltip component adds, for better or worse, aria-label attribute to the span, and that's not allowed without a proper role.

themes/bootstrap5/js/facets.js
EreMaijala: Sorry, I'm not following. This file is a direct copy from themes/bootstrap3/js/facets.js apart from the session storage handling.

build.xml
EreMaijala: Fixed!

themes/bootstrap5/js/bs3-compat.js
EreMaijala: Oops, that was a silly one. I changed the module and function name to a more descriptive one.

EreMaijala: And yes, some of the changes could perhaps be moved to build.xml (or just done manually), but the intention here is to allow also bootstrap3-based themes with their own templates to work properly with minimal changes. In the long run I expect this to be changed to opt-in, or get deprecated and finally removed.

themes/bootstrap5/js/facets.js
crhallberg: No worries, it's a mistake of mine from the listen-emit pull request.

themes/bootstrap5/templates/header.phtml
crhallberg: I see! Thanks for explaining!

themes/bootstrap5/js/facets.js
EreMaijala: Ok. :) 

This reminds me that there's a slight problem with events in the current facet lightbox code: the modal resize relies on the fact that Bootstrap 3 triggers the show.bs.modal event on each show command even when the modal is already visible, while Bootstrap 5 doesn't. I'll post a fix for that separately after this PR has been merged.

demiankatz (approved):
Looks like everything we expect to address here has been addressed, and there are a couple of follow-up actions we can take after merging this. Thanks, @crhallberg and @EreMaijala!

PR #3225, opened 2023-11-16
Fix searchbox_controls init function if textInput or resetButton are not defined

LuomaJuha:
Added a check that both elements are actually present. Seems like advanced search results does not have those.

### DIFF
--- a/themes/bootstrap3/js/searchbox_controls.js
+++ b/themes/bootstrap3/js/searchbox_controls.js
@@ -140,12 +140,13 @@ VuFind.register('searchbox_controls', function SearchboxControls() {
 
   function init(){
     _textInput = document.getElementById('searchForm_lookfor');
-    _resetButton = document.getElementById('searchForm-reset');
 
     if (!_textInput) {
       return;
     }
-  
+
+    _resetButton = document.getElementById('searchForm-reset');
+
     _textInput.addEventListener("input", function resetOnInput(event) {
       _handleInputChange(event.target.value, false);
     });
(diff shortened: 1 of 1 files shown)


themes/bootstrap3/js/searchbox_controls.js
```diff
@@ -144,6 +144,10 @@ VuFind.register('searchbox_controls', function SearchboxControls() {
     _textInput = document.getElementById('searchForm_lookfor');
     _resetButton = document.getElementById('searchForm-reset');
 
+    if (!_textInput || !_resetButton) {
```
demiankatz: I don't think a missing reset button should cause the whole initialization to fail. Maybe we should only return if _textInput is unset, and wrap the _resetButton.addEventListener call in an `if (_resetButton) {` check. What do you think?

demiankatz (changes requested):
Thanks, @LuomaJuha. One suggestion below; I'm also interested in hearing @ThoWagen's opinion if he has time.

themes/bootstrap3/js/searchbox_controls.js
ThoWagen: @demiankatz I think you are right. Currently the _resetButton should always exist if _textInput exists so _testInput alone should be enough. But if that changes in the future it might be a good idea to check for _resetButton as well.

Besides of adding an event listener in the init method _resetButton is also used in _handleInputChange. So the check should be also made there.

LuomaJuha: I think it could be better. I'll adjust

ThoWagen (commented):
Except my comment on @demiankatz suggestion I don't have anything to add.

ThoWagen (approved):
This looks good to me now.

demiankatz (approved):
Looks good to me too -- thanks, @LuomaJuha and @ThoWagen!

PR #3226, opened 2023-11-16
Adding the option to make html elements sticky

ThoWagen:
Sometimes it's tedious if you search for something, scroll down the page and if you want to change the search term you have to scroll up again.

This PR adds the option to make html elements sticky so that they always stay on the top of the page.

To make a element sticky one has to add the class `sticky-element`. There is a ViewHelper "StickyElement" that can be used
to make the stickiness of elements configurable. As default the navbar banner the navbar searchbox and the search box on the home page are made configurable. They can be enabled by uncommenting the`stickyElements[] =` lines in the config.ini. You can also make other elements sticky but maybe that requires to change the styling depending on what you want to do.

To define the order of the elements use the attribute `data-sticky-pos`. The elements that have this attribute are sorted by their integer value. All other elements come behind that.

To hide elements if the parent element is sticky one can add the class `sticky-hidden`. This is not used in the default themes. But for an example see: https://ubffm.hds.hebis.de/

TODO
- [x] Wait for #3064 to be merged and integrated before merging this (since this includes some required autocomplete fixes).


demiankatz (commented):
Thanks, @ThoWagen! This seems like a useful feature. I'm going to defer to @crhallberg for a first review since this is mostly UI-oriented, and he might have stronger opinions than me.

One general question, though: it looks like this is mostly pure Javascript, but there are a few bits of jQuery here and there. Would it be worth trying to eliminate the new jQuery code entirely? We're currently starting a project to get rid of jQuery, so I'm trying to avoid introducing too much new jQuery code. (It's not a deal-breaker if it's too hard to change right now, though -- we can always refactor later; this project will take a while).

ThoWagen:
I agree that no new jQuery should be introduced when the plan is to get rid of it anyway. I removed it.

themes/bootstrap3/js/vendor/autocomplete.js
```diff
@@ -276,6 +276,16 @@ function Autocomplete(_settings) {
           },
           false
         );
+        window.addEventListener(
```
demiankatz: autocomplete.js is a separate project (note that this is in a vendor directory), so changes to this code should be made through a PR to https://github.com/vufind-org/autocomplete.js. It looks to me like this might be a general improvement not directly tied to sticky elements in any case.

themes/bootstrap3/js/common.js
```diff
@@ -549,6 +549,131 @@ function bulkFormHandler(event, data) {
...
+      stickyElement.parentNode.insertBefore(gapFiller, stickyElement);
+
+      setPlaceholderStyle(stickyElement);
+      stickyElement.parentNode.style.backgroundColor = getInheritedBackgroundColor(stickyElement.parentNode);
+      stickyElement.previousSibling.style.backgroundColor = getInheritedBackgroundColor(stickyElement);
+    }
+  );
+  handleStickyElements();
+
+  window.addEventListener("resize", () => {
```
demiankatz: Should we skip setting up these event listeners when no sticky elements are found, to avoid doing unnecessary work?

themes/bootstrap3/js/common.js
```diff
@@ -549,6 +549,131 @@ function bulkFormHandler(event, data) {
 }
 
 // Ready functions
+function setupStickyElements() {
```
demiankatz: Rather than having a setupStickyElements function in the global namespace, it would probably be better to register this into the VuFind namespace as a module. I don't think much would need to change about the code, just a matter of changing the top-level to `VuFind.register('stickyElements', ...` and moving the initialization logic into an `init` method of the registered module. It should then bootstrap automatically as part of VuFind initialization without requiring a separate explicit call. See [check_item_statuses.js](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/js/check_item_statuses.js) for an example of this pattern. (This might also open the door to moving this into a sticky.js file that only gets included when sticky elements are turned on... though I'll leave that to your discretion).

module/VuFind/src/VuFind/View/Helper/Bootstrap3/StickyElementFactory.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+/**
+ * StickyElement helper factory.
+ *
+ * PHP version 8
+ *
+ * Copyright (C) Villanova University 2018.
```
demiankatz: Might want to use the current year here.

themes/bootstrap3/templates/header.phtml
```diff
@@ -1,5 +1,5 @@
 <?php $account = $this->auth()->getManager(); ?>
-<div class="banner container navbar">
+<div class="banner container navbar <?= $this->stickyElement('navbar-banner') ?>" data-sticky-pos="1">
```
demiankatz: It seems slightly less than ideal that we use a view helper to display a sticky class, but we always include the data element. A few possibilities:

1.) You could capture the sticky class to a variable, and then only output the data-sticky-pos if that value is non-empty.

2.) You could change stickyElement to just return a boolean, wrap it around the data-sticky-pos attribute, and associate sticky styles with the presence of the data element instead of with a specific class. (Though I think this may have negative performance implications).

3.) You could refactor the impacted elements to use the [HtmlAttributes view helper](https://docs.laminas.dev/laminas-view/v2/helpers/html-attributes/) instead of raw HTML for rendering attributes, and then you could change the stickyElement helper so that it injects multiple attributes into the HtmlAttributes result instead of directly into markup.

demiankatz (changes requested):
Thanks, @ThoWagen -- a few issues and suggestions below.

themes/bootstrap3/js/vendor/autocomplete.js
ThoWagen: Sorry, I missed that this is in vendor. I created a PR in the project https://github.com/vufind-org/autocomplete.js/pull/15. Should I remove this here or can it stay since it will be included anyway?

themes/bootstrap3/js/common.js
ThoWagen: Returning early now if no elements are found.

themes/bootstrap3/js/common.js
ThoWagen: I was not sure if it was worth to register it in `VuFind` since there is only the `init` method. But I moved it there.

Do you most of the time prefer registering new features to `VuFind`?

module/VuFind/src/VuFind/View/Helper/Bootstrap3/StickyElementFactory.php
ThoWagen: Fixed

themes/bootstrap3/templates/header.phtml
ThoWagen: I prefer that the ViewHelper returns the class. That way it is easier to change the class if necessary. But now the ViewHelper also returns the data attribute.

themes/bootstrap3/js/common.js
ThoWagen: The js file is only included if any element is configured in the config.ini. We have to keep in mind that it would be possible to create elements with the sticky class without using the config. In that case the file would have to be included in the theme config.

themes/bootstrap3/js/vendor/autocomplete.js
demiankatz: It's fine to leave it for now; I've added a TODO checkbox to the top of this PR to be sure we don't merge this one before the autocomplete updates are done.

themes/bootstrap3/js/common.js
demiankatz: Thanks, @ThoWagen. Yes, using the VuFind namespace is the preferred method for new code -- all of the global functions are legacy that we haven't found time to update/convert yet.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/StickyElement.php
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Returns the sticky-element class if the element is configured and
+     * optionally adds the position attribute.
+     *
+     * @param string $elementName Name of the element
+     * @param int    $pos         Sticky position of the element
+     *
+     * @return array
+     */
+    public function getElementAttributes($elementName, $pos = null)
```
demiankatz: Might be worth putting an `int` typehint on this to ensure that other values can't get passed in (which then means you can safely skip escaping the value when building the attribute below).

themes/bootstrap3/templates/header.phtml
demiankatz: Thanks, this makes sense to me! Just one other thing to think about: is it worth putting in any effort to avoid having extra whitespace in the markup when the setting is turned off? I can think of a few ways to do it, just not sure if it's worth bothering. (I tend to be a perfectionist about these kinds of things, but I also realize that it doesn't really matter much :-) ).

demiankatz (changes requested):
Thanks for the ongoing progress -- a couple more small comments below.

themes/bootstrap3/templates/header.phtml
ThoWagen: If there are no functional implications in having the extra whitespaces I would prefer not to change this. I also think it would be nice but if that means that we have to make the code more complicated or harder to read it's not worth it in my opinion.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/StickyElement.php
ThoWagen: Makes sense. I added it.

themes/bootstrap3/templates/header.phtml
demiankatz: There are no functional implications. I think the best way to solve this, if we wanted to solve it, would be to use the htmlAttributes view helper instead of assembling the code by hand... but whether that's better than the current approach is a matter of opinion -- depends on how much abstraction you prefer to see. The htmlAttributes approach will generate slightly tidier HTML output and is less prone to human error, but its extra layer of abstraction might negatively impact code readability. I don't insist on any changes, but that's probably the best one to consider, if changes are to be made at all.

crhallberg: I wonder if a good way to split the difference is to encode the data attribute into a class and use [`extractClassParams`](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/js/common.js#L432-L446) to get that data. For example, the helper would return `sticky sticky-pos:1`.

crhallberg (commented):
Thank you for the contribution and sorry for the novel! I've been doing a bunch of research on ways to clean up some of the hard-coded styles and complexities. Overall, I think the effect is working very well, but I have a few ideas on how the approach can be improved.

1. Due to the if-else nature of handleStickyElements, style is being calculated many times, often with the same results. Adjusting the conditions will result in less calculations. Something like:

```js
if (window.scrollY + currentOffset >= stickyContainer.offsetTop - Number(stickyElementStyle.marginTop)) {
  if (isSticky) {
    return;
  }
  // calculations
} else {
  if (!isSticky) {
    return;
  }
  // calculations
}
```

2. Additionally, I wonder how difficult it would be to cache the styles.

3. I'm unsure if the gap-filler is necessary. It seems to be 1px tall and the containers are set to offset-1. There may be a case with background color I'm not considering.

4. I've been doing tests using [position: sticky](https://caniuse.com/css-sticky) instead of position: fixed and it seems to cover many of the complexities of this effect. You can see below an experiment I made where the headers on the left are fixed and the ones on the right are sticky with the same top. They maintain their width and height, ignore their margins, stick to the top conditionally, and they also leave a gap where they were (no placeholder needed). Might be worth a look to simplify the margin-padding-placeholder bits.

image

5. One final idealistic idea: I wonder how chaotic it would be to set selectors in the config file instead of hard-coded names.

demiankatz:
Just to update the status of the autocomplete integration: https://github.com/vufind-org/autocomplete.js/pull/15 has been merged, so now we need to work through #3064 to get the vendor file updated. (We could make a release with just the latest changed code, but since there are other open questions here, I think it's safe to work through #3064 before resuming this... if priorities change, we can adjust course).

demiankatz:
@ThoWagen, #3064 has finally been merged, so I think work can continue here at your convenience. Please let me know if I can help with anything in the meantime!

demiankatz:
@ThoWagen, for some reason, the autocomplete.js updates were breaking our build, so I have reverted them. Thus, you may want to hold off on this a little longer. Working now to pinpoint exactly which commits were responsible for the break.

demiankatz:
@ThoWagen, fixing autocomplete was trickier than anticipated, but #3413 should finally take care of everything. Please go ahead and resume work here at your convenience.

ThoWagen:
@crhallberg Thank you for your input!
1. We always have to calculate at least some of the styles since they can depend on other elements and we also have to calculate all the styles on resizing the window. But I refactored it a bit so that not all the styles are calculated on scrolling.
2. I'm not completely sure what you mean by that. But the styles can always change a little bit depending on the screen size and how many elements are sticky.
3. When the positions are calculated the values are floating point numbers and it can happen that there a one pixel gap between two sticky elements depending on the browser and screen size. If we don't include the gap filler the content behind the sticky elements can be shown in the gap.
4. When I first started implementing this this was also my approach. But I never got it to work. Some problems are that the position is only sticky with respect to its parent element. And the position is also sticky vertically. I'm not saying it is not possible with `position: sticky` but it is probably not easier than this approach.
5. Sorry, here I'm not sure what you mean. Could you tell me please what exactly you refer to with "selectors in the config file" and "hard-coded names"

themes/bootstrap3/templates/header.phtml
ThoWagen: I did some refactoring using the htmlAttributes view helper now

crhallberg:
So I'm here to clarify and update my feedback. This is good news! You'll soon read that 1, 2, 4, and 5 are no longer in play. Just 3 would warrant further exploration at this time. Thank you for your patience!

1-2. I misunderstood how this was working, please disregard the too many calcs and cache styles feedback.

3. Good attention to detail. I'm not strongly against this approach. Would it be simpler to use rounding to account for this? I'm not strongly against this approach.

4-5. Starting with 5, I think it would make this tool more customizable to allow more elements to be made sticky. Right now, there are only three options, and I think that's a good place to focus for v10. In the future, the ability to add any id or selector to the configs, or to add a data-attribute (like lightbox does) might be more useful. This is probably way more complicated and would be made even more complicated if I required the use of `position: sticky` (so item 4 from before is also out)

ThoWagen:
@crhallberg Thank you for clarifying your points!

3. I checked again if we can simplify this. If we round the positions instead of filling up the gap the sticky elements might jump for one pixel when switching between the sticky and non-sticky state. That is not the most pleasant user experience. And we also can not replace the gap filler element by adding padding-top to the sticky-container since it might have a different background than the sticky element. We could add some padding to the sticky element itself when it is sticky. However, this it is has the base style for all the other additional elements and it might not be safe to mess with its padding. So I would prefer to stick to the current solution.

5)  Now I understand what you mean. In custom themes you always have the option to add the "sticky-element" class to any element yourself. But I think being able to add any selector to the configuration is a very good idea. We could get rid of the `StickyElement` view helper that I am not really happy with in the first place. What do you think about that @demiankatz ?

Besides those points I realized that having the whole header sticky is not good on mobile phones. So I'm thinking about making the stickiness dependent on the screen size.

demiankatz:
@ThoWagen, if you think this can be handled through configuration of selectors instead of the current model, I think that's both clearer and more flexible than the current approach, so I'd be in favor of it. I'm not exactly sure how that would work, but it seems you have something in mind, so I'll trust that there is a reasonable solution!

Making stickiness dependent on screen size also makes sense -- something else to think about when designing the configuration, to ensure the right balance between clarity and flexibility.

I'm hoping to freeze code at the end of next week at the very latest (and sooner would be better). Do you want to try to press forward and include this in release 10.0, or would it be better to move this milestone to 10.1? (I don't mind trying for either approach, but just wanted to raise awareness of the timeline, since 10.0 is now very close).

ThoWagen:
@demiankatz I don't mind moving the milestone to 10.1.

@crhallberg I think I found a way to get rid of the gap filler a way after all. But it needs some more testing before I push the new version.

demiankatz:
Thanks, @ThoWagen -- I'm moving the milestone so we won't feel too rushed to finish this in an unreasonable hurry. :-)

ThoWagen:
The replacement for the gap filler was actually straight forward. Replacing `offsetHeight` with `getBoundingClientRect().height` did the trick since the later return the exact float instead of the approximation. So thank you for pointing me in that direction @crhallberg and sorry that I initially rejected your advice.

Additionally, I added the possibility to configure the scope in which the effect is being active, moved the configuration to theme.config.php and added some smaller fixes and improvements. 

This should be ready for another review.

demiankatz:
Thanks, @ThoWagen, this looks much improved to me. I'll wait to hear what @crhallberg has to say before I approve it. It may also be worth getting @sturkel89's testing assistance to confirm consistent behavior across all themes and multiple browsers, since there's some complex Javascript/CSS magic going on here. I can talk to her about the details once Chris has given the thumbs-up.

crhallberg (approved):
This looks excellent! Thanks for wading through my unclear feedback!

sturkel89:
I've tested the listed sticky elements (and the hidden sticky elements) in the base themes and in bootstrap/sandal5, on desktop and mobile. I really like this feature, and it works well! 

My only critique is that in **bootstrap5** only, there is insufficient padding above the search box and below the Find button, so the element looks a little cramped or skimpy. This is especially noticeable on mobile:

### Bootstrap5:
image


### Other themes:
image
image
image


### Desktop:
(The thin horizontal line at the top is the VERY top of my browser, before the personal bookmarks bar.)
image

### Sandal5:
image

demiankatz (approved):
Thanks, everyone! I'm going to approve this now but will wait to hear from @ThoWagen before merging it in case he wants to take any further action before it is merged.

I suspect that the Bootstrap5 issues may just be part of the nature of the Bootstrap5 theme (which is designed to impose as little customization as possible and leverage Bootstrap-provided defaults). @EreMaijala might have more to say about that, but he's out of office until August, so we'll have to wait a while for his input. :-)

So I guess the remaining questions are:

1.) Merge now or later?

2.) Are the aesthetic bootstrap5 issues worth trying to fix? If so, now or later?

demiankatz:
I just noticed that there was a small conflict in js-configs.phtml, so I have resolved it. @sturkel89 has offered to double-check my changes to be sure nothing is broken... so unless she requests changes in the next few minutes. this should still be fine to merge. :-)

sturkel89:
Everything still works! Search tabs are visible when scrolling in full screen, and disappear on scroll on mobile. This works properly in all five themes.

I think we're ready to merge!

ThoWagen:
Thanks for all the feedback! If @EreMaijala thinks we should address the insufficient padding in bootstrap5 I will happily create a followup PR. So I think we can merge this now.

demiankatz:
Thanks again, everyone -- merging now!

EreMaijala:
@ThoWagen @demiankatz I believe bootstrap3 has a small margin as a side-effect of the button padding, which isn't quite the same with bootstrap5. It would make sense to define a margin properly.

demiankatz:
Thanks, @EreMaijala -- @ThoWagen, do you mind opening a PR as proposed? Please let me know if I can help with anything!

ThoWagen:
I created the PR #3865

PR #3227, opened 2023-11-16
Check exit codes for commands in the browse_build function

meganschanz:
We had noticed that it was possibly for the `sort` command in the `browse_build` function in the `index-alphabetic-browse.sh` script to fail, but since it was not checking the exit code of that `sort` command, it was improperly returning an exit code of zero. Instead of just adding the check for that command I thought it wouldn't hurt to have the exit code check on all the commands in that `browse_build` function.

This is particularly useful for us because we have a wrapper script around `index-alphabetic-browse.sh` that will alert us when there are failures (i.e. the script exits with a non-zero exit code).

### DIFF
--- a/index-alphabetic-browse.bat
+++ b/index-alphabetic-browse.bat
@@ -112,20 +112,24 @@ if "!%3!"=="!1!" goto skipauth
 set args="%bib_index%" "%field%" "%auth_index%" "%browse%.tmp"
 :skipauth
 
-rem Extract lines from Solr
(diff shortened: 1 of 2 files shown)


demiankatz (approved):
Thanks, @meganschanz, this is a useful improvement. I also appreciate the addition of comments!

I don't have the time or inclination to fully port this to the Windows version of the script, but I did at least align the comments between the two versions -- see 3d47b63c9a15ada3c3517eb1c9180f013b796252.

PR #3228, opened 2023-11-17
Fix hierarchical facet expand/collapse buttons with mobile Safari.

EreMaijala:
Mobile Safari has extra padding for buttons which needs to be disabled. It also displays the “Black Right-Pointing Triangle” as a play button (see [example](https://symbl.cc/en/25B6/)), so I changed it to a white one.


demiankatz:
The code looks fine to me, but @sturkel89, do you mind checking this out in all the themes and making sure it's still okay? I think we expect the color of the hierarchical facet icon to change in some circumstances, but the rest should be the same.

Note that since this is a bug fix targeted for release 9.1.1, it should be compared against the release-9.1 branch instead of dev.

Here are the [notes on setting up hierarchical facet tests](https://vufind.org/wiki/development:testing:manual_testing#hierarchical_facets) in case you don't already have them handy.

sturkel89 (commented):
I tested the PR with and without OR facets turned on. 

The test branch is behaving as designed, consistently in every theme, in that the black "play button" is replaced by an open triangle of about the same size. The down arrows for expanded levels of the hierarchy have the same appearance in test and release-9.1.

Personal opinion: I think the black triangles look better and are easier to see and parse than the open triangles. Also, when OR facets are enabled, I find the display to be slightly less legible in the test branch. Examples:

image

vs. 
image

I don't have a good solution to this; since the checkboxes need to be to the left of the facet label, hierarchical facet display is always going to be a bit busy. 

Some web interfaces use a big plus symbol when there's a list that can be expanded, but I don't know that we are considering a redesign!  

In any case - the test branch is working as intended, consistently across themes! I would prefer that all of the arrows be visually stronger in some way, to make them stand out next to the OR checkboxes, but I don't have a specific idea of how to accomplish that.

demiankatz:
Thanks, @sturkel89. For my part, I sort of appreciate the consistency of all the triangles looking the same, though I think my personal preference would be for all of them to be filled. Sadly that's not possible because of one web browser misrepresenting one character -- the joys of web development! Might be interesting to see if @crhallberg has any alternative ideas, but the good news is that the way this is set up, it's very easy to customize the icons, so, for example, if somebody wanted plus / minus icons instead of the triangles, that could be accomplished with a single configuration file edit. Maybe it's even worth including comments in the existing configuration to suggest this alternative, if @EreMaijala thinks it's worth the effort.

EreMaijala:
Well, we can always circle back to the FontAwesome icons (now here in the last commit)... What do you think? They are just a bit small without font size adjustment, but one could argue that the smaller ones make the display easier to the eye.

Regarding the busy look, maybe we could get rid of the checkboxes and indicate OR facets some other way. They are just a visual element anyway, but I'm not sure how to indicate a selected facet clearly. That's a different task, however..

EreMaijala:
Another option would be a custom SVG font..

sturkel89 (approved):
I checked the new version in all three themes and compared with release-9.1. 

I like the new, smaller filled triangles! They don't compete for my visual attention with the OR boxes the way the arrows in previous iteration did. 

The size of the triangles works for me - as @EreMaijala says, the smaller triangles are a bit easier on the eye. I think it's ready to merge!

Quick comparison:

**release-9.1**
image

**test branch**
image

demiankatz:
Thanks, @EreMaijala and @sturkel89!

PR #3229, opened 2023-11-17
Databases: Use LibGuides alternate names field

maccabeelevine:
LibGuides AZ has an "alternate names" field described as

> Alternate names associated with the database. If a user searches for these words, this database will appear in the results.

If the LibGuides option is enabled for the Databases recommendation module, we should (by default) search for URLs against the alternate names in addition to the primary name.

An example we ran into locally is Oxford Music Online often referred to as Grove Music Online by long-time researchers.  So both "Oxford" and "Grove" should match that database URL.

### DIFF
--- a/module/VuFind/src/VuFind/Recommend/Databases.php
+++ b/module/VuFind/src/VuFind/Recommend/Databases.php
@@ -332,6 +332,7 @@ class Databases implements RecommendInterface, \Laminas\Log\LoggerAwareInterface
             $nameToDatabase = [];
             foreach ($databases as $database) {
                 $nameToDatabase[$database->name] = (array)$database;
+                // The alt_names field is single-valued free text
                 if ($this->useLibGuidesAlternateNames && ($database->alt_names ?? false)) {
                     $nameToDatabase[$database->alt_names] = (array)$database;
                 }
--- a/module/VuFind/src/VuFind/Recommend/Databases.php
+++ b/module/VuFind/src/VuFind/Recommend/Databases.php
@@ -172,8 +172,8 @@ class Databases implements RecommendInterface, \Laminas\Log\LoggerAwareInterface
         if (!$databasesConfig) {
             throw new \Exception("Databases config file $databasesConfigFile must have section 'Databases'.");
         }
-        $this->configFileDatabases = isset($databasesConfig->url) ? $databasesConfig->url->toArray()
-            : $this->configFileDatabases;
+        $this->configFileDatabases = $databasesConfig->url?->toArray()
+            ?? $this->configFileDatabases;
         array_walk($this->configFileDatabases, function (&$value, $name) {
             $value = [
                 'name' => $name,
@@ -181,8 +181,7 @@ class Databases implements RecommendInterface, \Laminas\Log\LoggerAwareInterface
             ];
         });
 
-        $this->resultFacet = isset($databasesConfig->resultFacet)
-            ? $databasesConfig->resultFacet->toArray() : $this->resultFacet;
+        $this->resultFacet = $databasesConfig->resultFacet?->toArray() ?? $this->resultFacet;
         $this->resultFacetNameKey = $databasesConfig->resultFacetNameKey
             ?? $this->resultFacetNameKey;
 
--- a/module/VuFind/src/VuFind/Recommend/Databases.php
+++ b/module/VuFind/src/VuFind/Recommend/Databases.php
@@ -74,14 +74,7 @@ class Databases implements RecommendInterface, \Laminas\Log\LoggerAwareInterface
      *
      * @var int
      */
-    protected $limit;
-
-    /**
-     * Name of the configuration file for databases config (minus ".ini").
-     *
-     * @var string
-     */
-    protected $databasesConfigFile;
+    protected $limit = 5;
 
     /**
      * The result facet with the list of databases.  Each value in the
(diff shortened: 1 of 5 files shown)


module/VuFind/src/VuFind/Recommend/Databases.php
```diff
@@ -316,6 +332,9 @@ protected function getLibGuidesDatabases()
             $nameToDatabase = [];
             foreach ($databases as $database) {
                 $nameToDatabase[$database->name] = (array)$database;
+                if ($this->useLibGuidesAlternateNames && ($database->alt_names ?? false)) {
+                    $nameToDatabase[$database->alt_names] = (array)$database;
```
demiankatz, line 335: I notice that the API field is plural (alt_names) but we're treating it as a single flat string. What happens when there are multiple values here? Is there a standard delimiter, or is this just a free-text field that's used inconsistently?

I realize that we're doing a str_contains search against this value, so it should work even if the field contains multiple names... but if there is a standard and reliable delimiter, it feels like it might be worth splitting it up to make our index a little more accurate -- that might eliminate a certain rare subset of false positive hits and/or provide an opportunity for smarter relevance ranking logic in the future.

demiankatz (commented):
I don't have a problem with merging this as-is, but I wanted to raise one point before doing it in case there's room for improvement here.

Even if we change nothing, maybe it's worth adding a comment to address the point raised, so I don't forget and ask it again in the future. ;-)

module/VuFind/src/VuFind/Recommend/Databases.php
maccabeelevine: Yeah, unfortunately it is a plain single-line string field.  I have seen a value like "Used to be called XX" which is probably not what was intended and of course would cause problems if someone actually searched for "used to" for some ^$E@#%! reason.  So I went with a naive algorithm.  In any case that's what LibGuides would be doing itself to search against that field.

demiankatz: Makes sense! As I noted, it's then probably worth adding a comment like:

```
// The alt_names field is single-valued free text; we'll do the best we can with it
```

:-)

module/VuFind/src/VuFind/Recommend/Databases.php
```diff
@@ -126,6 +126,14 @@ class Databases implements RecommendInterface, \Laminas\Log\LoggerAwareInterface
      */
     protected $useLibGuides;
 
+    /**
+     * Configuration of whether to match on the alt_names field in LibGuides
+     * in addition to the primary name
+     *
+     * @var bool
+     */
+    protected $useLibGuidesAlternateNames;
```
demiankatz: Since this property isn't set until `setConfig` is called, it might be a good idea to give it a default value. Of course, `setConfig` should always be called in normal circumstances, but it might be good to avoid errors if the class is used in a different way in the future. (Same probably goes for a couple of other existing properties).

demiankatz (changes requested):
Thanks for the clarification, @maccabeelevine -- just a couple of small comments.

module/VuFind/src/VuFind/Recommend/Databases.php
maccabeelevine: Agreed, sorry I missed the comment suggestion the first time around.

module/VuFind/src/VuFind/Recommend/Databases.php
maccabeelevine: I agree in theory with the point about protecting if setConfig is not called in the future, but I don't like the idea of declaring the same default value (for each property) in two different places.  That seems to create a maintenance issue  if the default changes or we deprecate something, and it makes the code less clear.  Of course declaring a different default value in the property declaration is no better.  The best approach might be declare defaults and then to change setConfig() to only set the values if configured, i.e.  

if (isset($databasesConfig->useQuery) { $this->useQuery = $databasesConfig->useQuery; }

which is verbose, or

$this->useQuery = $databasesConfig->useQuery ?? $this->useQuery;

which is shorter but looks silly.  And neither seems to be the norm in this codebase generally, or in recommendation modules in particular; i.e. no defaults set during property declaration in AbstractSearchObject.  So I agree with the future-proofing problem but I think the alternatives are worse, unless we want to start a much larger conversation?

demiankatz: I actually wouldn't mind either of these approaches. As you say, it may also be a good idea to check other Recommendation modules for problems that fit this pattern (it's an easy thing to overlook, and probably has been overlooked elsewhere), but I don't think we have to solve the problem everywhere in order to address it here. Both of those patterns can be found elsewhere in the codebase (particularly in the Search\*\Options classes), though you might not find such precedents in the Recommendation code, as it is generally not very uniform, since each module is designed to be fairly stand-alone, and many come from different developers.

There's also a third potential approach: create a getter for the property, and have it throw an exception if the getter is called before the value has been set.

I think of the three options, I slightly prefer the null coalescing approach because of its brevity, in spite of the fact that as you say, it looks a bit silly. But I could live with any of them!

maccabeelevine: Ok, as you say I did find examples of both approaches I had suggested, even if they look a bit odd.  I went with the null coalescing approach, like you I think it's best.  As long as there is a comment explaining it, which I adapted from Base/Options.

demiankatz (approved):
Thanks, @maccabeelevine!

One small point that I'll leave to your discretion. With new syntax introduced in PHP 8 (and thus not yet used widely in the codebase), I'm pretty sure you can simplify:

```
        $this->configFileDatabases = isset($databasesConfig->url) ? $databasesConfig->url->toArray()
            : $this->configFileDatabases;
```

to:

```
        $this->configFileDatabases = $databasesConfig?->url?->toArray() ?? $this->configFileDatabases;
```

I've approved this PR but won't merge it right away, so I can leave this up to you: if you want to try that simplification and it works, feel free to make the change and merge (I trust you!). If you've spent enough time on this and don't want to bother, then just merge as-is and we can work on applying that optimization more globally in the future.

maccabeelevine:
> new syntax introduced in PHP 8

@demiankatz Nice!  I am fan of the same in javascript, glad to see it is in PHP now as well.  I made those changes.  To be clear I did not also null-safe the $databasesConfig variable since that is checked earlier on line 172.

PR #3232, opened 2023-11-17
Combined: Fix column widths in grid-layout

maccabeelevine:
**Fixed from #3220.**

This addresses two issues with the CSS grid layout for combined search added in #3077 that occur when one column happens to have a value that can't be wrapped, and is longer than the intended width of the column.  (The intention is that each column have equal width, as with the other combined column modes.)  This happens in practice with EDS data that has a long, visible URL.

1) When one column is forced longer than expected, but the combined width of all columns is still small enough to fit in .mainbody.  The result is the columns are uneven.

image

2) When once column is forced so large that the total columns can't fit in .mainbody.  Then the sidebar overlaps the main content.  This example is fake but I've seen it happen for real. 

image


themes/bootstrap3/less/components/search.less
```diff
@@ -427,6 +427,10 @@ body.rtl {
 .combined-search-container.grid {
   margin-left: 1rem;
   margin-right: 1rem;
+
+  .combined-list {
+    overflow-wrap: anywhere;
```
maccabeelevine, line 432: > I also agree that the overflow scrollbar is not good. I'd prefer hidden but there's more we can do, I think. Since this happens when long words come into play, we can allow for more breaks in words:
> 
> ```css
> .combined-list {
>     word-break: break-word;
>     hyphens: auto;
> }
> ```
> 
> We can also do [what we do locally at Villanova](https://library.villanova.edu/Find/Combined/Results?type=&lookfor=behold&limit=20&sort=relevance) and make the container on the combined screen full width.

@crhallberg Yes this is so much better than a scrollbar, thanks for the idea.  MDN said word-break: break-word is deprecated so I used its suggested replacement.  Overflow-wrap: break-word didn't work either but :anywhere did.  The hyphens rule didn't affect it, but actually given that what is wrapping (mostly) are URLs, I'm happier not hyphenating them as I think that confuses things further.

crhallberg: Oh! I hadn't realized that! Thank you for the heads up!

crhallberg (approved):

PR #3233, opened 2023-11-22
Combined: Fix jump links and grid with zero results

maccabeelevine:
Two small fixes to combined search when there are zero results in a particular bento box and `hide_if_empty = true`:

1. Hide that particular jump link.  And if all bentos are hidden, hide the jump link row entirely.
2. In Grid stack placement, fix hiding the bento box.


themes/bootstrap3/js/combined-search.js
```diff
@@ -5,6 +5,11 @@ VuFind.combinedSearch = (function CombinedSearch() {
     VuFind.loadHtml(container, url, '', function containerLoad(responseText) {
       if (responseText.length === 0) {
         container.hide();
+        container = container.parent();
+        while (container.hasClass('hide-if-empty')) {
+          container.hide();
+          container = container.parent();
+        }
```
maccabeelevine: I know we are moving to eliminate JQuery from the codebase.  I used it here because this function is called with a JQuery element, and VuFind.initResultScripts below also expects a JQuery element, so with input and output determined it seemed a bad place to fix.

themes/bootstrap3/templates/combined/jump-links.phtml
```diff
@@ -1,8 +1,8 @@
-<div class="combined-jump-links">
+<div class="combined-jump-links" style="visibility: hidden;">
```
maccabeelevine: Using visibility instead of display: none to avoid the layout jerk when it becomes visible.

demiankatz:
Thanks, @maccabeelevine! Do you think we should backport this to the release-9.1 branch? If so, let me know and I'm happy to see if I can easily rebase it.

maccabeelevine:
> Thanks, @maccabeelevine! Do you think we should backport this to the release-9.1 branch? If so, let me know and I'm happy to see if I can easily rebase it.

Yeah I considered that, and I guess it's a risk tolerance issue on how sure we are I didn't break anything else with this fix. :)  Also minor UI implications so @crhallberg may have a thought on that.  It doesn't matter to me one way or the other.

themes/bootstrap3/templates/combined/jump-links.phtml
crhallberg: Can we move this to the CSS instead of hard-coding in the HTML?

themes/bootstrap3/templates/combined/jump-links.phtml
```diff
@@ -1,8 +1,8 @@
...
   <span><?=$this->translate('combined_jump_links_intro')?></span>
   <ul>
     <?php foreach ($combinedResults as $section) : ?>
-      <li class="<?= $this->escapeHtmlAttr($section['domId']) ?>">
+      <li class="<?= $this->escapeHtmlAttr($section['domId']) ?>" style="display: none;">
```
crhallberg: Same here.

themes/bootstrap3/templates/combined/stack-grid.phtml
```diff
@@ -27,7 +27,7 @@
       $columnIndex ++;
     ?>
     <?php // This empty div prevents margin collapse, as the otherwise unneeded div.combined-column does elsewhere. ?>
-    <div>
+    <div class="hide-if-empty">
```
crhallberg: Can we add a `js-` prefix to this class? It's a new convention for classes that are JS triggers or flags.

themes/bootstrap3/js/combined-search.js
```diff
@@ -5,6 +5,11 @@ VuFind.combinedSearch = (function CombinedSearch() {
     VuFind.loadHtml(container, url, '', function containerLoad(responseText) {
       if (responseText.length === 0) {
         container.hide();
+        let parent = container.parent();
```
crhallberg, line 8: Would [`.closest()`](https://api.jquery.com/closest/) or [`.parents()`](https://api.jquery.com/parents/) be a more flexible option? It would cover a case where not every element up the chain has the class.

crhallberg (changes requested):

themes/bootstrap3/js/combined-search.js
maccabeelevine: My intention (in this use case) was to stop as soon as the traversal encounters an element that does hot have the class.  For most of the layout modes, none of the parents would have that class so it would stop immediately; for stack-grid it would go up one level before stopping.  I get your point that this could be more flexibly implemented, but it's not needed for this use case and I think we'd have to think more about how it would be used in those other scenarios.

themes/bootstrap3/templates/combined/jump-links.phtml
maccabeelevine: Sure, I hate inline HTML, but I thought it was appropriate here to imply the temporary condition of that CSS since it will be overridden via JS (and thus a net in the DOM of no inline styles).  That's just a stylistic choice so if you prefer the original styles in .less files I can do that.

themes/bootstrap3/templates/combined/stack-grid.phtml
maccabeelevine: Done.

demiankatz:
@maccabeelevine, I'm going to pin this to the 10.0 milestone while you and @crhallberg work through this conversation, just to ensure it doesn't get lost track of. When it's done, we can evaluate whether or not to backport to the release-9.1 branch based on the final results.

themes/bootstrap3/js/combined-search.js
crhallberg: That makes sense. It just looks like you set this up to handle multiple layers of nesting within the loop, and I thought those utilities might save us a headache in the future.

crhallberg (commented):
Will review again once the inline CSS is... outlined?

maccabeelevine:
Hi @crhallberg I replaced the inline CSS with declarations in search.less and (updated) overrides in the JS.

crhallberg (approved):

crhallberg:
Thank you for making those changes! I've switched my review to approval. Is there anything else on your radar, @maccabeelevine, that this PR needs?

maccabeelevine:
> Thank you for making those changes! I've switched my review to approval. Is there anything else on your radar, @maccabeelevine, that this PR needs?

@crhallberg No it's good on my end!

demiankatz (approved):
Thanks, @crhallberg and @maccabeelevine -- merging now!

demiankatz:
(Since this is a somewhat significant change, I decided not to backport to 9.1 in case of unwanted side effects... but I'm happy to revisit that decision if anyone has a reason to want the fix released sooner).

PR #3234, opened 2023-11-23
Disable null overrides in RecordDataFormatter options

ThoWagen:
In the RecordDataFormatter's `addOptions` method options from different sources get merged. Currently, if null values are set they also override other options. This means that for example if you change `renderType` in the `[Global]` section of the config it is not applied when you have something like the following.
```
$spec->setLine(
          'New Title',
          'getNewerTitles',
          null,
          ['recordLink' => 'title']
);
```
Therefore, null values get filtered before merging the options now.


demiankatz (approved):
Thanks, @ThoWagen!

PR #3235, opened 2023-11-23
Removed prioritizeAlt default from SpecBuilder

ThoWagen:
Now, that there are configurable defaults we don't need this anymore to set the default for the prioritizeAlt option.


demiankatz (approved):
Thanks, @ThoWagen!

PR #3236, opened 2023-11-27
MySQL port configurable in build.xml

ThoWagen:
This PR adds the option to set a Non-standard MySQl port in build.xml. It also changes the host name component of users to '%' since the IP address of the system that runs the apache is not necessarily the same as the one that runs the database. E.g. if you use containers.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -30,6 +30,7 @@
   <property name="vufinddbpass" value="vufindtestpass" />
   <property name="dbtype" value="mysql" /><!-- use pgsql for PostgreSQL -->
   <property name="mysqlhost" value="localhost" />
+  <property name="mysqlport" value="3306" />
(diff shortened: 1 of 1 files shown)


demiankatz:
This looks good to me, and I plan to do a hands-on test as soon as time permits (currently catching up on vacation backlog). However, I'm also requesting a review from @EreMaijala, just in case he has any concerns that I'm not thinking of!

EreMaijala (approved):
The only issue I can think of is that the old vufinddbuser@localhost may be left dangling if startup is done with an older version, but that's fine for me.

demiankatz (approved):
Thanks, @EreMaijala -- I agree that it's fairly unlikely that someone would startup with one commit and then shutdown with this one, so I'm willing to take that risk. All tests are passing for me, so I'm merging this now!

PR #3237, opened 2023-11-27
Adding hidden sort options

ThoWagen:
One might want to allow more sort options than those that can be selected in the result list. This PR adds the option to add hidden sort options. They can also be regexes. This can be useful if the sorting depends on a record id for example.

Note: I'm using the nullsafe operator introduced in PHP 8. If I recall correctly this will be the minimum required version in 10.0. Please let me know if this is not correct or if the compatibility break is not wanted anyway.


demiankatz (commented):
Thanks, @ThoWagen, this makes sense to me, though I haven't studied it extremely closely yet. In the meantime, two comments:

1.) Using the nullsafe operator is fine. We have required PHP 8 since release 9.1, so you're welcome to use all PHP 8 language features now.

2.) It doesn't seem like this needs to be a Solr-only feature. Should we add the setting of hiddenSortOptions to the base class instead of the Solr class? I can't think of a reason why this would be harmful, and it would allow us to easily extend this setting to various other backend-specific .ini files.

ThoWagen:
The feature was already in the base Options and Params classes. But I added it to all of the other backends that have sorting and included it in the same places where the normal sorting is applied. I tested it for EDS and Seach2 but not the other backends due to lacking configs for them. But it should work for them too.

I also added a Mink test for it.

demiankatz:
Thanks, @ThoWagen -- my only question is whether it's best to move the configuration loading into the Base/Options class so it's not necessary to touch so many other classes. I realize that this may enable the setting to be configured for a backend that otherwise does not support sorting, but I can't think of any particularly serious implications of doing this. If the parameter isn't wired up to anything anyway, allowing values to be set in the parameter shouldn't really have any effect.

If you prefer to keep this as-is, I'll finish reviewing and will merge -- but if you'd like to simplify, please go ahead.

Either way, let me know and I'll hopefully be able to finish this up as soon as I hear back from you.

ThoWagen:
Now, I understand what you mean. I agree. This is a better solution and it's not necessary to do this in each sub class.
I changed it.

demiankatz (approved):
This looks great, @ThoWagen, and all tests are passing. Merging now!

PR #3238, opened 2023-11-28
EPF: Add embargo description

maccabeelevine:
For EBSCO Publication Finder results, each holding's coverage statement should also display any related embargo.  See "Business Source Complete" example:

image

This formatting (the entire holding info in a single line) mirrors what the native EPF UI does.  However I added two classes in case someone wants to add local CSS to format it.  Similarly someone could hide it with `display: none;` if they really want to.


demiankatz (approved):
Makes sense to me! Thanks, @maccabeelevine!

PR #3239, opened 2023-11-29
Configurable bulk action limits

ThoWagen:
This PR adds the option to disable or set limits for bulk actions.

It is based on the discussion in #3142

- [x] Fix errors in lightbox
- [x] Translate error messages
- [x] Find a solution for BulkActionTrait
- [x] Add Mink tests
- [x] refactor translations (wait for #3249)
- [x] Note BC breaks in changelog when merging (language file changes, CartController/MyResearchController constructor signature changes).
- Find and set proper limits for the actions (Task for a followup PR)


ThoWagen:
@demiankatz This is a first version for this. There is still something to do (see todo list). I am not sure what the best solution is for the following problem.

Currently by using `redirectToSource` when the action limit is exceeded we may return to the result or favorite list inside of the lightbox. 

Some ideas on how we could fix this:
1. Additionally catching an exceeded action limit by using JS like we do when no items are selected. Advantage: We don't have to fist make a request and it's probably the user friendliest solution. Disadvantage: It might make the code less maintainable and more error prone.
2. We could create a new action that only renders the flashmessages. This could for example be added to the index controller. If we are in a lightbox we can set `$target` to this action `redirectToSource`. Advantage: It could also be used for similar problems. Disadvantage: It feels a bit hacky.
3. Catching the exceeded limit after the forms inside of the lightbox is submitted and just returning the form with the error message. Advantage: Probably the simplest solution. Disadvantage: It is not very user friendly since error is displayed rather late.

What do you think. Do you have a better idea?

demiankatz:
Thanks for the progress on this, @ThoWagen -- I'll look at this more closely when I'm back in the office. In the meantime, I agree that the most user-friendly and efficient solution would be to check in the Javascript. Maybe we can do this generically with a `data-max-selected-items` element on the form (or something similar). As you say, we'll need to be careful to implement checks in both the front end and the back end to avoid inconsistencies and gaps, and maybe we should figure out a way to bypass the Javascript check in tests so we can add test coverage to the backend logic. If you'd like to start exploring this route, I certainly wouldn't object! Though if you prefer to do the option 3 "simplest approach" first, that's fine too -- because we can always do it that way initially and add the extra JS checks as a later enhancement. Option 2 is interesting, but I agree that it feels a bit hacky.

config/vufind/config.ini
```diff
@@ -1838,6 +1838,17 @@ era             = true      ; allow browsing of era subdivisions
 ; <result_limit> most popular entries -- it only affects display order.
 ;alphabetical_order = true
 
+; This section is used to configure bulk actions.
+[BulkActions]
+; Set the limit of items for each action. Set 0 to disable the action and hide the button.
+; Depending on you setup you may be able to increase the limits or have to decrease them
```
demiankatz: Typo: "you setup" --> "your setup"

config/vufind/config.ini
```diff
@@ -1868,6 +1879,18 @@ RIS = false
 ; export.ini for more details on available options.
 ;defaultType = download
 
+; This overrides the limit of exports in the BulkActions section per method. Note that they
+; can not exceed that limit and can only be used to further restrict the number of items that
+; can be used per export.
+limits["RefWorks"] = 100
```
demiankatz: There might be practical reasons to leave this here, but did you consider putting these settings in Export.ini instead? It seems like it might be more desirable to put these configurations in a more specific place, to reduce noise in config.ini and to remind people to set this value when adding new export methods.

module/VuFind/src/VuFind/Controller/CartController.php
```diff
@@ -246,6 +257,12 @@ public function emailAction()
             return $this->redirectToSource('error', 'bulk_noitems_advice');
         }
 
+        // Check if id limit is exceeded
+        $actionLimit = $this->config?->BulkActions?->limits?->email ?? 0;
```
demiankatz: Maybe a `checkBulkActionLimit($action)` support method would make this code less repetitive.

I'm also a little concerned about defaulting this to 0, since users who fail to update your config.ini when upgrading will suddenly lose all bulk options. Maybe we should add a separate config setting for "default bulk limit" and default the default to 100. What do you think?

module/VuFind/src/VuFind/Controller/CartController.php
```diff
@@ -55,16 +57,25 @@ class CartController extends AbstractBase
...
      * @param ServiceLocatorInterface $sm        Service manager
      * @param Container               $container Session container
+     * @param Config                  $config    VuFind configuration
      */
-    public function __construct(ServiceLocatorInterface $sm, Container $container)
+    public function __construct(ServiceLocatorInterface $sm, Container $container, Config $config)
     {
         parent::__construct($sm);
         $this->session = $container;
+        $this->config = $config;
```
demiankatz: Might be better to pass the Config Plugin Manager here so we can access multiple configurations (i.e. if you want to take my suggestion about moving settings to export.ini).

themes/bootstrap3/templates/myresearch/bulk-action-buttons.phtml
```diff
@@ -2,23 +2,29 @@
   <input type="hidden" name="listID" value="<?=$this->escapeHtmlAttr($list->id)?>">
   <input type="hidden" name="listName" value="<?=$this->escapeHtmlAttr($list->title)?>">
 <?php endif; ?>
-<?php $user = $this->auth()->isLoggedIn(); ?>
+<?php
+$user = $this->auth()->isLoggedIn();
+$actionLimits = $this->config()->get('config')?->BulkActions?->limits;
```
demiankatz: If we add more smarts to limits as suggested above, we'll probably need more view helper logic somewhere to apply equivalent behavior here. If you want to go that route, I can give it a little more thought and propose a location, if you don't already have ideas.

demiankatz (changes requested):
Thanks, @ThoWagen -- I was able to give this a deeper review and had a few more thoughts...

config/vufind/config.ini
ThoWagen: fixed

config/vufind/config.ini
ThoWagen: Moved it. But just realized that the explanation is missing in the comment of export.ini. I will add it.

module/VuFind/src/VuFind/Controller/CartController.php
ThoWagen: > Maybe a checkBulkActionLimit($action) support method would make this code less repetitive.

Did this and added it to a new BulkActionTrait.

>I'm also a little concerned about defaulting this to 0, since users who fail to update your config.ini when upgrading will suddenly lose all bulk options. Maybe we should add a separate config setting for "default bulk limit" and default the default to 100.

Good Idea. Did this.

module/VuFind/src/VuFind/Controller/CartController.php
ThoWagen: Done

themes/bootstrap3/templates/myresearch/bulk-action-buttons.phtml
ThoWagen: I added a view helper for rendering bulk action buttons that checks the limits.

ThoWagen:
@demiankatz I implemented your suggestions.
There is a new BulkAction trait that is used by the CartController, MyResearchController and the new BulkAction view helper. But there is one problem with it. 

I also moved the `redirectToSource` method inside. But it is not used by the BulkAction view helper. To pass the tests I added an empty `redirectToSource` method to it. But this is not a good design choice I'm afraid. Maybe I just have to remove the method from the trait again with the drawback that the `deleteAction` inside of the MyResearchController has to implement this again. Do you have a better idea for this?

I also think the templates could be simplified by creating a mapping of the actions to the icons and translation key inside of the view helper. And I am not sure if we need all of the different translations for the titles. The email button for example has the title "email_selected" in the favorites, "bookbag_email_selected" in the search results and "bookbag_email". What do you think?

config/vufind/config.ini
```diff
@@ -1838,6 +1838,18 @@ era             = true      ; allow browsing of era subdivisions
 ; <result_limit> most popular entries -- it only affects display order.
 ;alphabetical_order = true
 
+; This section is used to configure bulk actions.
+[BulkActions]
+; Set the limit of items for each action. Set 0 to disable the action and hide the button.
+; Depending on your setup you may be able to increase the limits or have to decrease them
+; for the actions to work properly.
+limits["default"] = 100
```
demiankatz, line 1864: It might be worth adding a comment to highlight the fact that this default value will be used for any action lacking a more specific setting.

module/VuFind/src/VuFind/Controller/MyResearchControllerFactory.php
```diff
@@ -0,0 +1,76 @@
...
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        if (!empty($options)) {
+            throw new \Exception('Unexpected options sent to factory.');
+        }
+        $configLoader = $container->get(\VuFind\Config\PluginManager::class);
+        return $this->applyPermissions(
```
demiankatz: You can simplify this last line to `return parent::__invoke($container, $requestedName, [$configLoader]);` -- the AbstractBaseFactory is designed to accept additional constructor parameters through the `$options` array, which then allows us to reduce redundancy in its children.

module/VuFind/src/VuFind/Controller/CartControllerFactory.php
```diff
@@ -71,9 +71,10 @@ public function __invoke(
             'cart_followup',
             $container->get(\Laminas\Session\SessionManager::class)
         );
+        $configLoader = $container->get(\VuFind\Config\PluginManager::class);
         return $this->applyPermissions(
             $container,
-            new $requestedName($container, $session)
+            new $requestedName($container, $session, $configLoader)
```
demiankatz: The same factory simplification can be applied here (and I see that somebody, possibly me, neglected to do that in the first place :-) ).

module/VuFind/src/VuFind/Controller/Feature/BulkActionTrait.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Get the limit of a bulk action.
+     *
+     * @param string $action Name of the bulk action
+     *
+     * @return int
+     */
+    public function getBulkActionLimit($action)
+    {
+        return $this->configLoader->get('config')?->BulkActions?->limits?->$action
```
demiankatz: Might be worth doing `$config = $this->configLoader->get('config');` so we don't have to call the configLoader multiple times when checking the default... though in practice it's a very fast lookup and shouldn't make much difference. I'll leave it up to your discretion. :-)

module/VuFind/src/VuFind/Controller/Feature/BulkActionTrait.php
```diff
@@ -0,0 +1,109 @@
...
+            && !$redirectInLightbox
+            && $this->inLightbox()
+        ) {
+            return false;
+        }
+
+        // If we entered the controller in the expected way (i.e. via the
+        // myresearchbulk action), we should have a source set in the followup
+        // memory. If that's missing for some reason, just forward to MyResearch.
+        if (isset($this->session->url)) {
```
demiankatz: This trait depends on having a specific session container set up in `$this->session`. We should document that, and possibly also add a check here that throws an exception if it's missing. I realize that this particular method is only going to be used in the context of a controller, but since this trait is shared with a view helper, it seems like a potential source of problems or confusion. Should we consider splitting to two traits or moving some functionality back into the controller?

module/VuFind/src/VuFind/View/Helper/Bootstrap3/BulkAction.php
```diff
@@ -0,0 +1,106 @@
...
+            return '';
+        }
+        $attributes['class'] = 'toolbar-btn';
+        $attributes['value'] = '1';
+        $attributes['type'] = 'submit';
+        $attributes['name'] = $action;
+        $attributes['data-item-limit'] = $limit;
+        return $this->getView()->render(
+            'Helpers/bulk-action-button.phtml',
+            [
```
demiankatz: You could reduce this array to `compact('action', 'icon', 'content', 'limit', 'attributes')`.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/BulkAction.php
```diff
@@ -0,0 +1,106 @@
...
+            ]
+        );
+    }
+
+    /**
+     * Disable trait method
+     *
+     * @return mixed
+     */
+    public function redirectToSource()
```
demiankatz: I guess this was how you decided to address my concerns about redirectToSource -- but I think I would prefer to handle this on the trait side with an `isset($this->session)` type of check instead of relying on the client to stub out problematic methods. But maybe I misunderstand your intent here or am overlooking some important detail!

themes/bootstrap3/js/common.js
```diff
@@ -537,10 +537,23 @@ function resetCaptcha($form) {
 }
 
 function bulkFormHandler(event, data) {
-  if ($('.checkbox-select-item:checked,checkbox-select-all:checked').length === 0) {
+  let numberOfSelected = $('.checkbox-select-item:checked').length;
```
demiankatz: Maybe worth changing to `document.querySelector` to reduce jQuery usage while we're changing the line anyway.

themes/bootstrap3/templates/search/bulk-action-buttons.phtml
```diff
@@ -6,17 +9,82 @@
...
+              array_merge(
+                  [
+                    'id' => $this->idPrefix . 'ribbon-print',
+                    'title' => $this->translate('bookbag_print_selected'),
+                  ],
+                  $this->formAttr ? ['form' => $this->formAttr] : []
+              )
+          )
+        ?>
+        <?php if (($actionLimits?->saveCart ?? 0) > 0 && $this->userlist()->getMode() !== 'disabled'): ?>
```
demiankatz: $actionLimits is still being defined without being used in several templates... and it's used here, but I don't think it needs to be because the view helper now incorporates the limit check.

demiankatz (changes requested):
Thanks, @ThoWagen, this is shaping up very nicely! I haven't done any hands-on testing yet, but the below comments (and a couple of recent trivial commits) are the result of a fairly thorough read-through. Once these are addressed, I'll work with @sturkel89 to actually exercise all the code!

demiankatz:
@ThoWagen, my apologies -- I should have more carefully read your comments before writing my review, as I see you already raised the issue of the BulkActionTrait. One option might be to have a BulkActionControllerTrait which incorporates BulkActionTrait and adds the controller-specific methods. There might be some other repetitive logic which could be moved into such a trait (e.g. something like a `getSelectedIdsFromRequest()` method to encapsulate the repeated logic that picks the right posted ID values.

I also agree that moving more of the button settings into the view helper might make sense, though we should be careful that we don't make anything too hard to customize or override (e.g. if a user wants to add a local custom button, it shouldn't be too difficult to do). It does appear that we can clean up some of the redundant translation strings as well. Maybe that task should wait until we finish #3249, since doing so would enable us to set up aliases for some of the redundant strings -- then perhaps the view helper could include a "context-prefix" argument that would be applied to translation strings, and this would allow users to differentiate labels for bulk vs. cart if they wanted to, without creating a heavier burden in language file maintenance.

config/vufind/config.ini
ThoWagen: done

module/VuFind/src/VuFind/Controller/MyResearchControllerFactory.php
ThoWagen: done

module/VuFind/src/VuFind/Controller/CartControllerFactory.php
ThoWagen: done

module/VuFind/src/VuFind/Controller/Feature/BulkActionTrait.php
ThoWagen: done

module/VuFind/src/VuFind/Controller/Feature/BulkActionTrait.php
ThoWagen: I split the trait into two as you suggested and added the dependency in the doc comment.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/BulkAction.php
ThoWagen: done

module/VuFind/src/VuFind/View/Helper/Bootstrap3/BulkAction.php
ThoWagen: removed it, since it is not necessary anymore.

themes/bootstrap3/js/common.js
ThoWagen: done

themes/bootstrap3/templates/search/bulk-action-buttons.phtml
ThoWagen: cleaned up the templates.

ThoWagen:
I split the trait into two traits and moved the "main" trait into a new "Feature" directory in the VuFind module. But maybe there is a better place for it. 

Now, refactoring the translations and figuring out proper limit is still todo. For the translations I wait as suggested for #3249. The limits should be tested on multiple systems. Maybe it's better to stick with the limits of 100 for now and do it as a followup of this and #3142.

And I agree that it is better not to move to much logic into the view helper to keep adjustments simpler.

demiankatz:
Thanks, @ThoWagen, the changes so far look good! I'll try to make some more progress on #3249 later this week (hopefully on Friday) and we can resume work here when that is done. I agree that it probably makes sense to start with default limits of 100, since that essentially mirrors the existing functionality, and we can investigate tweaking defaults as a separate project after the basic functionality is in place.

demiankatz:
@ThoWagen, as noted elsewhere, #3249 is now merged, so you should be free to make further progress here. Please let me know if I can do more to help and/or when you're ready for me to do another round of tests and reviewing.

ThoWagen:
@demiankatz I refactored the translations using aliases. All tasks should be completed now.

languages/aliases.ini
```diff
@@ -0,0 +1,13 @@
+bulk_delete = "delete_selected"
+bulk_delete_myresearch = "bulk_delete"
```
demiankatz: Would it be better to alias this directly to delete_selected instead of aliasing an alias? Or do you prefer it this way?

config/vufind/export.ini
```diff
@@ -14,6 +14,8 @@
 ;     specify the appropriate export format.
 ; label - the label to display to the user when referring to this format; if omitted,
 ;     the section name will be used by default.
+; limit - limits the number of items that can be exported at once. Overrides the export
```
demiankatz, line 17: It's not exactly accurate to say that this limit overrides the [BulkActions] limit, since it only overrides if it is LOWER. If I set a limit of 5 in config.ini, exports will be blocked even if everything is set to 100 in this file. Not sure if we need to clarify this comment or change some behavior; I don't have strong feelings and am happy to use whatever approach is easiest, at least to begin with!

demiankatz (changes requested):
Thanks, @ThoWagen! This looks good to me and all tests are passing. Just a couple of small things to discuss before it is merged.

languages/aliases.ini
ThoWagen: As long as we only have a bulk delete button in the favorites we can do this. Done.

config/vufind/export.ini
ThoWagen: The export actually isn't blocked but you get a message that it is. To fix this we could either use the maximum of all export limits as the limit that is used for the message in the general case or we can implement that the general export limit in config.ini is an upper limit of the limit in export.ini and update this comment. I'm not sure what the better approach is.

demiankatz: I could be persuaded to pick either of the approaches you propose.

I think having the export settings truly override the config.ini settings, whether they go higher or lower, seems like the most intuitive and flexible approach... but the disadvantage is that I suspect calculating the correct limit value may become complicated and require special-case logic somewhere in the code. If this proves too burdensome, I think the compromise of treating the config.ini setting as a hard maximum is acceptable.

I also think that regardless of which approach we take, it may be a good idea to comment out more of the export configurations by default. Right now, we have lots of settings all set to the same value of 100. If somebody wants to change it, they'll potentially need to override a lot of places. But if we just set the default to 100 and comment everything else out, we'll get the same effect, and it will be easier for people to customize. (Obviously, we might decide to uncomment more of these settings in future once we figure out appropriate individual defaults for each situation... but as a starting point, I think just having a default set will make things easier).

ThoWagen: I implement the first approach and commented the limits in export.ini out.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/BulkActionFactory.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+/**
+ * BulkAction helper factory.
+ *
+ * PHP version 8
+ *
+ * Copyright (C) Villanova University 2018.
```
demiankatz: Might be worth updating this copyright line.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/BulkAction.php
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Constructor
+     *
+     * @param ServiceLocatorInterface      $sm           Service Locator
+     * @param \VuFind\Config\PluginManager $configLoader Configuration loader
+     */
+    public function __construct(ServiceLocatorInterface $sm, \VuFind\Config\PluginManager $configLoader)
+    {
+        $this->serviceLocator = $sm;
```
demiankatz: If the only reason for adding the serviceLocator is so it can be used in `getBulkActionExport()`, maybe it would be better to directly inject and set `$this->bulkActionExport` instead. Exposing the service locator is generally a bad practice, and I'm trying to reduce that pattern as much as possible -- if we can avoid introducing it here, I think that would be preferable.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/BulkAction.php
```diff
@@ -0,0 +1,99 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * @category VuFind
+ * @package  View_Helpers
+ * @author   Thomas Wagener <wagener@hebis.uni-frankfurt.de>
+ * @license  http://opensource.org/licenses/gpl-2.0.php GNU General Public License
+ * @link     https://vufind.org/wiki/development Wiki
+ */
+
+namespace VuFind\View\Helper\Bootstrap3;
```
demiankatz, line 30: Is there a strong reason to put this in the Bootstrap3 namespace rather than the Root namespace? I realize that there might be theme-specific class names involved, but even if that's the case, it might make sense to have a Root version that gets extended/overridden in specific theme namespaces. I don't insist on changing this if you think it's best as-is, but since we're currently working on migrating from bootstrap3 to bootstrap5, I think it might reduce future work if we can keep things in the Root namespace.

demiankatz (changes requested):
Thanks, @ThoWagen. This seems to be working as expected, and all tests are still passing.

See below for a couple of final thoughts related to the view helper; I think we can merge once we've sorted these out.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/BulkActionFactory.php
ThoWagen: Done

module/VuFind/src/VuFind/View/Helper/Bootstrap3/BulkAction.php
ThoWagen: The BulkActionTrait does depend of `$export` instead now.

module/VuFind/src/VuFind/View/Helper/Bootstrap3/BulkAction.php
ThoWagen: Moved most of the logic to Root and only setting the Bootstrap3 button class here.

module/VuFind/src/VuFind/Controller/CartController.php
```diff
@@ -48,23 +49,45 @@
...
+     * @var \VuFind\Config\PluginManager
+     */
+    protected $configLoader;
+
+    /**
+     * Export support class
+     *
+     * @var \VuFind\Export
+     */
+    protected $export;
```
demiankatz, line 73: Would it make more sense to move this property definition into the trait that uses it, rather than duplicating it in multiple places?

module/VuFind/src/VuFind/Feature/BulkActionTrait.php
```diff
@@ -113,7 +93,7 @@ protected function getBulkActionExportConfig()
     public function getBulkActionLimit($action)
     {
         if ($action == 'export') {
-            $formats = $this->getBulkActionExport()->getActiveFormats('bulk');
+            $formats = $this->export->getActiveFormats('bulk');
```
demiankatz, line 96: Maybe worth throwing an exception if export is unset at this stage?

demiankatz (changes requested):
Thanks, @ThoWagen -- a couple final suggestions on the latest refactoring!

module/VuFind/src/VuFind/Controller/CartController.php
ThoWagen: \VuFind\Export is actually also used in CartController.php. I removed the getExport method since we set it in the constructor now.

I checked how other traits handle dependencies and those that I checked state the dependency in the doc comment but don't define the properties themselves. And since they get set in the constructors of the classes I think it is better to define them in the classes as well. But this more based on a feeling than on knowledge of trait design concepts. If you prefer to define `$export` in the trait I will do so.

module/VuFind/src/VuFind/Feature/BulkActionTrait.php
ThoWagen: As with the comment above I checked other traits. They only state the dependency in the doc comment and do not check them. And in this case we get an exception anyway if we want to call the method on null. A custom exception might only have a better description. I will still add it if you want.

module/VuFind/src/VuFind/Controller/CartController.php
demiankatz: It's fine to leave it as-is. This is the part of traits that I simply do not like -- I wish that traits could formally define requirements and/or have some kind of constructor equivalent to initialize them... but the reality is that traits are a very simplistic mechanism, and these types of things don't really have a place there. We just have to do our best to come up with useful conventions to avoid problems.

Perhaps there is an argument for moving the export property definition into the trait and also adding an `initializeExportTrait` method which could be called from the class constructors to provide the required value(s). But that feels like inventing a whole new mechanism that doesn't already have a precedent, and now is probably not the time to do it. :-)

module/VuFind/src/VuFind/Feature/BulkActionTrait.php
demiankatz: Similarly, it's fine to leave this as-is. I think in my mind, there's a difference between properties that the trait expects the composing class to already have, vs. things that are used primarily by the trait. But in reality this distinction is not usually very clear, so it does not make sense to overcomplicate things based on my feelings here.

demiankatz (approved):
Thanks, @ThoWagen, I think this is in good shape now. I fixed a couple of small typos in comments and made some updates to a recently-added language file, and I think it's now ready to merge. When you have a chance, I think we're now ready to resume work on #3142. In the meantime, I hope you have a good weekend!

PR #3240, opened 2023-11-29
Sierra: Fix batch retrieval of items and bibs.

EreMaijala:
Turns out Sierra will limit results to 50 records by default even if you ask for a set of identifiers. This will override the default limit with the number of identifiers requested. There'll be another PR to further tweak the mechanism so this is a minimal one to fix the issue.


demiankatz (approved):
Thanks, @EreMaijala!

demiankatz:
Just noticed that in spite of the branch name, this was targeted against dev instead of release-9.1. I've cherry-picked the commit onto release-9.1 to compensate, so this is now in both release-9.1 and dev.

EreMaijala:
@demiankatz Oops, thanks!

PR #3241, opened 2023-11-29
Disabled session writes for Ajax tabs and getResultCount

ThoWagen:
As discussed in #2941 enabling `loadInitialTabWithAjax` or `show_result_counts` in the config.ini can lead to session conflicts when also using SSO login for bulk actions. This PR fixes this by disabling session writes for these Ajax calls.

### DIFF
--- a/module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
+++ b/module/VuFind/src/VuFind/AjaxHandler/GetResultCount.php
@@ -34,6 +34,7 @@ namespace VuFind\AjaxHandler;
 use Laminas\Mvc\Controller\Plugin\Params;
 use Laminas\Stdlib\Parameters;
 use VuFind\Search\Results\PluginManager as ResultsManager;
+use VuFind\Session\Settings as SessionSettings;
 
 /**
  * "Get Result Counts" AJAX Handler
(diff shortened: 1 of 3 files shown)


demiankatz:
Thanks, @ThoWagen! I'll give this a closer look when I'm back tomorrow, but I think it should be fine. As I mentioned in #2941, I had some concerns about the CollectionList RecordTab handler needing access to the session, but that handler already disables AJAX functionality, so it shouldn't be an issue. I'll review the rest to be sure there are no other edge cases I'm forgetting, but I doubt I'll find any.

demiankatz (approved):
I've reviewed this a little more thoroughly and run the full test suite, and all seems well, so I'm going to merge this.

However, I had a conversation with @EreMaijala about this, and he made a good point that if we ever have forms inside tabs that use CSRF, disabling the session will break them due to the current mechanism we are using for tracking CSRF tokens. He has opened [VUFIND-1652](https://openlibraryfoundation.atlassian.net/jira/software/c/projects/VUFIND/issues/VUFIND-1652) to propose a new mechanism for storing session-adjacent information, which should solve this problem and hopefully help alleviate other potential race conditions. In the meantime, though, the changes here shouldn't impact existing core tabs, and the option always exists to disable AJAX loading for specific tabs via the `supportsAjax()` method of the tab classes.

I think we _could_ theoretically backport this change to the release-9.1 branch, but I'm not going to do that for now just in case we're missing any subtle side effects.

PR #3242, opened 2023-11-30
Sierra: load items and bibs in batches for holds and fines.

EreMaijala:
Consolidates all bib and item caching and retrieval into a single method and use it for all cases (apart from `getItemsForBibRecord` which fetches items by bib id). This makes the code cleaner and improves performance when retrieving holds and fines.


demiankatz (approved):
Looks reasonable to me, though I don't have the ability to test it directly. Unit tests might be nice, but probably not a priority, especially given the potentially finite lifetime of Sierra itself.

PR #3243, opened 2023-11-30
FOLIO: Fix getStatus() after getHolding() change

maccabeelevine:
In #3094 I changed the FOLIO driver getHolding() function to return an array instead of just $items.  Holds.php is fine either way but I didn't notice that getStatus() is picky and still expects just $items.  This broke the availability check in GetItemStatuses.


demiankatz (approved):
Thanks for the fix!

PR #3253, opened 2023-12-08
VUFIND-1342 Update browse commands for new browse project layout

marktriggs:
This is the companion commit to the changes in https://github.com/vufind-org/vufind-browse-handler/pull/48

TODO
- [x] Finish and merge vufind-browse-handler PR, and then add the resulting binaries here before merging.
- [x] Update changelog when merging.


demiankatz (approved):
Looking good! I've updated the binaries here to include the changes from https://github.com/vufind-org/vufind-browse-handler/pull/48 and https://github.com/vufind-org/vufind-browse-handler/pull/49, and I have confirmed that the index script works correctly in Windows and that the full test suite passes in Linux.

PR #3256, opened 2023-12-11
JS: HTML escape update

crhallberg:
I was looking to improve the [htmlEncode method](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/js/common.js#L364-L370) we've been using, since it requires jQuery and the creation of an element. Seemed overkill. After looking at [Lodash's solution](https://lodash.com/docs/4.17.15#escape) and contributing it to a [plain-JS alternative](https://thescottyjam.github.io/snap.js/#!/nolodash), I was directed to the XSS resources below, which I based this change on.

- [Cross Site Scripting Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html) (especially the Safe Sinks section)
- [Snap.js' escape function and explanation](https://thescottyjam.github.io/snap.js/#!/nolodash/escape).

---

Todo:
- [x] ~~Remove or~~ update [htmlEncode](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/js/common.js#L364-L370)?
- ~~Remove [escapeHtmlAttr](https://github.com/vufind-org/vufind/blob/dev/themes/bootstrap3/js/common.js#L390-L430)?~~
- In the future, create more elements instead of HTML strings.


demiankatz:
Thanks, @crhallberg, this looks like it's off to a good start. Could you let me know when you're ready for thorough testing? I assume it's work in progress for the moment, based on the TODO list.

For my part, I would be inclined not to remove the escape functions right away -- I know that they're still getting used in a few more places (and I'm pretty sure I just added another place recently in one of my open pull requests). It's probably worth the effort to de-jQuery-ize them for short-term use... and if we want to get rid of them in the longer term, maybe the first step would be to start `console.log`-ing deprecation warnings when they are used, so we can slowly phase them out.

demiankatz:
Thanks, @crhallberg, this makes sense to me, but before I do deeper testing, I'm interested in @EreMaijala's opinion as well. Is this good enough, or should we consider incorporating [he](https://github.com/mathiasbynens/he) for a more robust approach?

EreMaijala (changes requested):
I think this is fine otherwise, but `String.replaceAll()` is a relatively recent addition, and there are at least iOS versions that are still in active use on older devices [that don't support it](https://caniuse.com/?search=String.replaceAll), and while using a few year old browser is not recommendable, I wouldn't want to knowingly exclude them just yet.

demiankatz:
> I think this is fine otherwise, but `String.replaceAll()` is a relatively recent addition, and there are at least iOS versions that are still in active use on older devices [that don't support it](https://caniuse.com/?search=String.replaceAll), and while using a few year old browser is not recommendable, I wouldn't want to knowingly exclude them just yet.

Is the solution something like this?

```
  return String(value)
    .replace(/&/g, "&amp;")
    .replace(/</g, "&lt;")
    .replace(/>/g, "&gt;")
    .replace(/"/g, "&quot;")
    .replace(/'/g, "&#39;");
```

That might actually be preferable on general principle because it seems like fewer characters to do the same thing. :-)

EreMaijala:
> > I think this is fine otherwise, but `String.replaceAll()` is a relatively recent addition, and there are at least iOS versions that are still in active use on older devices [that don't support it](https://caniuse.com/?search=String.replaceAll), and while using a few year old browser is not recommendable, I wouldn't want to knowingly exclude them just yet.
> 
> Is the solution something like this?
> 
> ```
>   return String(value)
>     .replace(/&/g, "&amp;")
>     .replace(/</g, "&lt;")
>     .replace(/>/g, "&gt;")
>     .replace(/"/g, "&quot;")
>     .replace(/'/g, "&#39;");
> ```
> 
> That might actually be preferable on general principle because it seems like fewer characters to do the same thing. :-)

Yes, that'd do it. Fewer characters, yes, but at least in theory using a regular expression is slower... But anyway, I think this is the best compromise. :)

demiankatz (approved):
Thanks, @crhallberg, this looks good to me now, and the full test suite is passing. I think we can merge this as soon as @EreMaijala approves.

EreMaijala (approved):

PR #3257, opened 2023-12-12
Update website content

demiankatz:
This PR is designed to update some outdated language and improve organization of the vufind.org website.

### DIFF
--- a/about.html
+++ b/about.html
@@ -32,17 +32,17 @@ layout: default
       support when you need it.
     </p>
 
-    <h2>Power</h2>
-    VuFind® runs on Solr Energy. Apache Solr, an open source search engine, offers amazing performance and scalability to allow for VuFind® to respond to search queries in milliseconds time. It has the ability to be distributed if you need to spread the load of the catalog over many servers or in a server farm environment.
+    <h2>Power and Flexibility</h2>
+    VuFind®'s search index runs on Apache Solr, an open source search engine, offering amazing performance and scalability. It has the ability to be distributed if you need to spread the load of the catalog over many servers or in a cloud environment. If you don't need to run your own local index, or if you want to complement it with third-party content, VuFind® can also talk to <a href="https://vufind.org/wiki/configuration:third-party_content">a variety of other tools and services</a>.
 
     <h2>Licensing</h2>
     VuFind® is offered for free through the GPL open source license. This means that you can use the software for free. You can modify the software and share your successes with the community!
 
-    <h2>VuFind® in Action</h2>
-    Take a look at our <a href="https://vufind.org/wiki/community:installations">VuFind® Installations Wiki page</a> to see how a variety of organizations have taken advantage of VuFind®'s flexibility. If you are already using VuFind®, feel free to edit the page and share your accomplishments.
   </div>
   <div class="col-sm-6">
     <img src="https://vufind.org/images/magnify.png"/>
+    <h2>VuFind® in Action</h2>
+    Take a look at our <a href="https://vufind.org/wiki/community:installations">VuFind® Installations Wiki page</a> to see how a variety of organizations have taken advantage of VuFind®'s flexibility. If you are already using VuFind®, tell us about it through our <a href="https://vufind.org/advanced_demo/Feedback/Form/InstallationReport">Installation Report form</a>.
     <h2>Praise for VuFind®</h2>
     <quote>
       <b>"This completely rules, as you kids say."</b>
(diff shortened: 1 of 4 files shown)


features.html
```diff
@@ -3,25 +3,24 @@
 ---
 <h2>Features</h2>
 <div class="video">
-  <iframe src="//player.vimeo.com/video/1239618" width="500" height="333" frameborder="0" webkitallowfullscreen mozallowfullscreen allowfullscreen></iframe>
-  <a href="http://www.vimeo.com/1239618?pg=embed&amp;sec=1239618">VuFind®: Social Features</a> from
-  <a href="http://www.vimeo.com/user433393?pg=embed&amp;sec=1239618">Falvey Memorial Library</a> on
-  <a href="http://vimeo.com?pg=embed&amp;sec=1239618">Vimeo</a>.
+  <iframe width="560" height="315" src="https://www.youtube.com/embed/6Nf6sDSlQqg?si=p00c27j_OD-z5BL8" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>
```
demiankatz, line 6: Replacing 2008-era video with more recent webinar. I kept a link to the old video in the wiki on the videos page, under "Historical."

features.html
```diff
@@ -30,58 +29,61 @@ <h2>Features</h2>
...
-      And if you want complete access to your indexed data, you can interact with Solr,
+      You can syndicate your record data with other institutions via an
+      <a href="https://vufind.org/wiki/indexing:tracking_record_changes#oai-pmh_server_functionality">OAI server</a>.
+      You can search using VuFind®'s algorithms via its <a href="https://vufind.org/wiki/development:apis">built-in APIs</a>.
+      And if you want complete access to your indexed data, you can interact directly with Solr,
       VuFind®'s backend search and index engine.
     </quote>
+
+    <quote>
+      <b>Integration with Content Providers</b>
```
demiankatz: I don't think highlighting author biographies specifically makes a lot of sense, so I generalized it and moved it down here.

support.html
```diff
@@ -57,21 +57,19 @@ <h2 id="community">Community Support</h2>
 
       <li><a href="https://lists.sourceforge.net/mailman/listinfo/vufind-admins">VuFind&reg; Admins</a> - A mailing list for discussing management of the VuFind&reg; project.</li>
 
-      <li><a href="https://lists.lehigh.edu/mailman/listinfo/vufind-unicorn-l">VuFind&reg; Unicorn Users</a> - A mailing list for SirsiDynix Unicorn users who are using VuFind&reg;.</li>
```
demiankatz, line 60: This list appears to be dead.

support.html
```diff
@@ -57,21 +57,19 @@ <h2 id="community">Community Support</h2>
...
-
       <li><a href="http://vufind.slack.com/">VuFind&reg; on Slack</a> - Email info@vufind.org for an invitation to the Slack if you cannot sign up directly.</li>
 
-      <li><strong>VuFind&reg; on IRC</strong> - For real-time discussion of VuFind&reg;, try joining the vufind channel at <a href="http://webchat.freenode.net/?channels=vufind">freenode</a>.</li>
```
demiankatz, line 64: I never check IRC, and I think Slack has replaced it, so I'm removing it. Also adding in GitHub Discussions since that's a new option that we might as well mention here.

features.html
dltj: Suggest as a heading: _Enhance Record Displays with Content Providers_

...gives the heading more of a why-would-you-want-to-do-this phrasing.

about.html
```diff
@@ -32,17 +32,17 @@ <h2>About VuFind®</h2>
       support when you need it.
     </p>
 
-    <h2>Power</h2>
-    VuFind® runs on Solr Energy. Apache Solr, an open source search engine, offers amazing performance and scalability to allow for VuFind® to respond to search queries in milliseconds time. It has the ability to be distributed if you need to spread the load of the catalog over many servers or in a server farm environment.
+    <h2>Power and Flexibility</h2>
```
dltj: Suggest as heading: _Powerful and Flexible Indexing_

sturkel89 (approved):
All of these changes look good to me!

about.html
demiankatz: I adjusted the about headings in 0c1e2ababccbc7b0e2a8d5d798bf6a2034d4bd0c -- I used "Search Capabilities" instead of "Indexing" since I also want to account for scenarios where no local index is used.

features.html
demiankatz: Good idea -- edit made.

PR #3258, opened 2023-12-12
Combined: Display noresults recommendations in bento box or sidebar

maccabeelevine:
"noresults" recommendations appear when the primary search backend returns zero results.  This PR allows those recommendations to appear in combined mode, either in the results box itself, or in the sidebar.

Displaying them in the sidebar allows the same recommendation to appear in the same place regardless if the primary backend returned any results or not, by enabling both include_recommendations_side and include_recommendations_noresults_side.  This seems to be good UX for recommendation modules that are not tightly coupled to their primary searches, like all of the AbstractSearchObject recommendations.

Displaying them in the box itself might make more sense if the recommendation is more relevant in a noresults scenario, or is otherwise tightly coupled to that box.


module/VuFind/src/VuFind/Controller/CombinedController.php
```diff
@@ -339,21 +339,32 @@ protected function adjustQueryForSettings($settings, $searchType = null)
         // Override the search type:
         $query->type = $searchType;
 
-        // Always leave noresults active (useful for 0-hit searches).
```
maccabeelevine, line 342: Despite this comment, it doesn't seem like noresults recommendations have any way to appear without this PR.  I may have been missing something.

module/VuFind/src/VuFind/Controller/CombinedController.php
```diff
@@ -339,21 +339,32 @@ protected function adjustQueryForSettings($settings, $searchType = null)
...
             $recommendOverride['top'] = $settings['include_recommendations'];
         } else {
             $noRecommend[] = 'top';
         }
+
+        // Display or hide side based on include_recommendations_side setting.
         if (is_array($settings['include_recommendations_side'] ?? false)) {
             $recommendOverride['side'] = $settings['include_recommendations_side'];
         } else {
             $noRecommend[] = 'side';
```
maccabeelevine, line 359: I am not adding anything to $noRecommend[] for noresults or noresults_side, because (per the comment above) I think I am missing something about how they used to be active.

themes/bootstrap3/templates/combined/results-list.phtml
```diff
@@ -130,6 +130,40 @@
...
+            recommendationContainer.append(recommendation);
+            recommendation.style.display = 'inherit';
+          }
+
+          if (document.readyState != 'loading') {
+            addRecommendation_$recommendationClass();
+          } else {
+            document.addEventListener('DOMContentLoaded', addRecommendation_$recommendationClass);
+          }
+          JS;
```
maccabeelevine: This whole JS block is a direct copy of the one at line 75, which I hate.  But it's a pretty short function, and with PHP writing $recommendationClass into the JS dynamically, including function names, I think predefining this in one place would get tricky and lead to less readable code.  But I can give it a shot if someone prefers.

demiankatz: Would it make sense to create a PHP function to generate this Javascript? There's nothing stopping you from just defining an anonymous function at the top of the template file, like:

`$generateMoveToSidebarJavascript = function ($recommendationClass) { ... };`

That would reduce redundancy without significantly harming clarity, unless I'm overlooking something!

module/VuFind/src/VuFind/Controller/CombinedController.php
demiankatz: I'm seeing norecommend settings in the EDS column of Villanova's Bento box when I search for gibberish:

https://library.villanova.edu/Find/Combined/Results?dfApplied=1&lookfor=zdarearsar&type=AllFields

Must be coming from somewhere! I don't have time to dig into the details today, but if you want me to find time for it later, let me know. Is it possible that AJAX-loaded columns are doing something different from non-AJAX-loaded columns? (I ask just because we load EDS by AJAX, and it's showing recommendations, so maybe that's a factor).

demiankatz (changes requested):
I didn't have time for anything approaching a thorough review today, but a couple of comments to perhaps help keep the ball rolling. :-)

module/VuFind/src/VuFind/Controller/CombinedController.php
maccabeelevine: Ok I think I found the problem.  In AbstractSearch->getSeachSetupCallback(), any `recommendOverride` settings in combined.ini overrride all of the default recommendation modules from searches.ini / EDS.ini etc.  So setting even a top recommendation in combined.ini overrides the noresults recommendations from EDS.ini.  

I think that's actually the intended behavior, override being global to all recommendations?  If so then we do need something like the include_recommendations_noresults setting in combined.ini.

demiankatz: Thanks for digging into this! It does indeed seem like there's a problem here that needs to be addressed. I suspect that different bits of development at different points in time led us to this situation, and it was not a conscious decision to suppress noresults recommendations when top/side overrides were present... but it does seem like the most straightforward solution would be to add a new include setting to address this. Let me know if you need any help on the details or implementation there. Maybe it's worth doing that as a separate PR and then circling back to this one after it is merged.

demiankatz: (Or maybe it makes more sense to do it all here -- I don't have a strong preference one way or the other!)

maccabeelevine: I decided it was easier to resolve separately.  #3264

themes/bootstrap3/templates/combined/results-list.phtml
maccabeelevine: Good call -- done!

config/vufind/combined.ini
```diff
@@ -33,6 +33,14 @@
 ;         the side column. If set to false, side recommendations will be suppressed
 ;         (default). Adding the same recommendation module to multiple sections is
 ;         unsupported.
+; include_recommendations_noresults = If set to an array of recommendation settings
+;         (as per searches.ini), the specified recommendations will be loaded into
+;         the results column. If set to false, noresults recommendations will be
```
demiankatz: I think we should probably document a "true" setting here equivalent to the "true" value of include_recommendations, and make that the default behavior for better backward compatibility.

demiankatz (changes requested):
Thanks for the progress on this! One more thought below, based on my recent work on #3264 and #3272. I'll give this a more thorough test after we've gotten those two merged and sorted out any resulting conflicts.

config/vufind/combined.ini
maccabeelevine: Once #3264 is merged, I think I should be able to remove the `include_recommendations_noresults` option entirely as redundant.  But I would still introduce `include_recommendations_noresults_side`.

And in that case, I'm not sure if we need a `true` option.  I didn't include one for `include_recommendations_side`.  I see why for full compatibility they both should have it.  But I wrote in combined.ini's explanation of `include_recommendations_side` that  `Adding the same recommendation module to multiple sections is ... unsupported.` because it would lead to problems -- for the JS code moving things to the sidebar, I'm keying IDs on the recommendation module name and not the source module name, so there would be name conflicts if the same recommendation was added more than once.  And because of that I thought automatically inheriting the combined side recommendations from the parents is a bad idea, you would create conflicts without easily seeing why.  So I would rather not offer it as an option for `include_recommendations_noresults_side` either.

demiankatz: #3264 is now merged, so please feel free to merge dev into here and make further adjustments -- just let me know when you're ready for the next (and hopefully final) review. Thanks!

I agree that it makes sense to omit the `true` option for `include_recommendations_noresults_side` for the reasons you state. I think there might still be some value in having an `include_recommendations_noresults` setting in case users want to disable the noresults recommendations in certain columns for some reason. This seems less likely to be changed than other types of recommendations, but if it costs us very little to make it configurable, it doesn't seem to hurt to do it while we're here anyway. :-)

maccabeelevine: That's fair, I will leave in `include_recommendations_noresults`.  In which case I think this is ready for review again.

demiankatz (changes requested):
I haven't re-tested this yet, but I think now that we've fixed the bugs that made this confusing, some more work is needed here to fully implement the new settings. As noted previously, I think it's worth documenting the "true" setting for include_recommendations_noresults, and I suspect that we do need to leverage noRecommend[] if we want to disable recommendations with false settings. If I'm mistaken and this is more functionally complete than I think it is, please let me know and I'll do hands-on testing... but I'm pretty sure it needs one more minor pass before it's ready.

maccabeelevine:
> I haven't re-tested this yet, but I think now that we've fixed the bugs that made this confusing, some more work is needed here to fully implement the new settings. As noted previously, I think it's worth documenting the "true" setting for include_recommendations_noresults,

Agreed - done.

> and I suspect that we do need to leverage noRecommend[] if we want to disable recommendations with false settings. If I'm mistaken and this is more functionally complete than I think it is, please let me know and I'll do hands-on testing... but I'm pretty sure it needs one more minor pass before it's ready.

Yes.  I think(?) I have this working correctly now.

config/vufind/combined.ini
```diff
@@ -33,6 +33,15 @@
 ;         the side column. If set to false, side recommendations will be suppressed
 ;         (default). Adding the same recommendation module to multiple sections is
 ;         unsupported.
+; include_recommendations_noresults = If set to true, standard 'top'
```
demiankatz: This should say 'norecommend' instead of 'top', shouldn't it?

config/vufind/combined.ini
```diff
@@ -33,6 +33,15 @@
...
+;         recommendations will be displayed at the top of the column. If set to an
+;         array of recommendation settings (as per searches.ini), the specified
+;         recommendations will be loaded into the results column. If set to false,
+;         noresults recommendations will be suppressed (default). Adding the same
```
demiankatz: Does the "Adding the same recommendation module to multiple sections is unsupported." note really only apply if using include_recommendations_noresults_side, or is the restriction broader than that? I'm wondering if that note should be moved down.

module/VuFind/src/VuFind/Controller/CombinedController.php
```diff
@@ -339,9 +342,6 @@ protected function adjustQueryForSettings($settings, $searchType = null)
         // Override the search type:
         $query->type = $searchType;
 
-        // Always leave noresults active (useful for 0-hit searches).
-        // Display or hide top based on include_recommendations setting.
```
demiankatz, line 343: I think you might want to keep the comment about include_recommendations -- that's still accurate here, isn't it?

module/VuFind/src/VuFind/Controller/CombinedController.php
```diff
@@ -350,11 +350,30 @@ protected function adjustQueryForSettings($settings, $searchType = null)
...
+        if (is_array($includeRecommendNoResultsSetting)) {
+            $recommendOverride['noresults'] = $settings['include_recommendations_noresults'];
+        } elseif (!$includeRecommendNoResultsSetting) {
+            $noRecommend[] = 'noresults';
+        }
+
+        if (is_array($settings['include_recommendations_noresults_side'] ?? false)) {
+            $recommendOverride['noresults_side'] = $settings['include_recommendations_noresults_side'];
+        } else {
+            $noRecommend[] = 'noresults_side';
```
demiankatz, line 375: I don't think you strictly need to add `noresults_side` here, since `noresults_side` is not a standard location that will have configuration set under circumstances other than combined search... but it may be worth leaving this anyway just for consistency, and in case we add a noresults sidebar to non-combined search in the future.

demiankatz (changes requested):
Thanks, @maccabeelevine. This seems to be working correctly now, though I have a few comments about comments and documentation.

Also, I wonder if the include_recommendations_noresults_side documentation would benefit from further clarification. At first, I expected that if I set `include_recommendations_noresults_side = true`, then I would get all the default noresults recommendations, only have them moved to the side bar. This is, of course, not actually the case. The only format that works is `include_recommendations_noresults_side[] = RecommendationModuleName`. That's probably as it should be, but this means that saying that `include_recommendations_noresults_side` is the same as `include_recommendations_noresults` in all regards except the location of the output is perhaps misleading.

config/vufind/combined.ini
maccabeelevine: D'oh.  Fixed.

config/vufind/combined.ini
maccabeelevine: You're right, that was another editing error.  Removed.

maccabeelevine:
> Also, I wonder if the include_recommendations_noresults_side documentation would benefit from further clarification. At first, I expected that if I set include_recommendations_noresults_side = true, then I would get all the default noresults recommendations, only have them moved to the side bar. This is, of course, not actually the case. The only format that works is include_recommendations_noresults_side[] = RecommendationModuleName. That's probably as it should be, but this means that saying that include_recommendations_noresults_side is the same as include_recommendations_noresults in all regards except the location of the output is perhaps misleading.

@demiankatz You are right, this was very misleading.  I rewrote the explanation from scratch instead of saying it's a copy of another one.

module/VuFind/src/VuFind/Controller/CombinedController.php
maccabeelevine: Thanks, accidental delete.  Fixed.

module/VuFind/src/VuFind/Controller/CombinedController.php
maccabeelevine: > and in case we add a noresults sidebar to non-combined search in the future

Yeah I think it's useful to leave for that reason.  There are a number of recommendation modules that work fine whether or not there are any search results -- just using the search query.  I.e. the various AbstractSearchObjects, or LibGuides & AZ.

demiankatz (approved):
Thanks, @maccabeelevine, I think this is good to go now!

PR #3259, opened 2023-12-12
Support Shibboleth in KohaRest driver

PerplexedTheta:
This patch adds support for Shibboleth authentication in the KohaRest driver, by allowing the sysadmin to bypass password authentication during the login process

In the KohaILSDI driver, you don't need to log in with a password if the dontValidatePasswords variable is set to true. I hope to reproduce this behaviour in this patch.

This is my first patch to VuFind. Please let me know if I've not met your standards, or if there are additional things I can help with, to get this code into the product.

Thanks!

### DIFF
--- a/themes/bootstrap3/templates/search/history-table.phtml
+++ b/themes/bootstrap3/templates/search/history-table.phtml
@@ -18,11 +18,13 @@
       <tr class="<?=$iteration % 2 == 1 ? 'even' : 'odd'?>row">
         <td class="history_time" data-label="<?=$this->transEscAttr('history_time')?>"><?=$this->escapeHtml($this->dateTime()->convertToDisplayDateAndTime('U', $info->getStartTime()))?></td>
         <td class="history_search" data-label="<?=$this->transEscAttr('history_search')?>">
-          <?=$this->historylabel($info->getParams()->getSearchClassId())?>
-          <a href="<?=$this->url($info->getOptions()->getSearchAction()) . $info->getUrlQuery()->getParams()?>"><?php
-            $desc = $info->getParams()->getDisplayQuery();
-            echo empty($desc) ? $this->transEsc('history_empty_search') : $this->escapeHtml($desc);
-          ?></a>
+          <span>
+            <?=$this->historylabel($info->getParams()->getSearchClassId())?>
+            <a href="<?=$this->url($info->getOptions()->getSearchAction()) . $info->getUrlQuery()->getParams()?>"><?php
+              $desc = $info->getParams()->getDisplayQuery();
+              echo empty($desc) ? $this->transEsc('history_empty_search') : $this->escapeHtml($desc);
+            ?></a>
+          </span>
         </td>
         <td class="history_limits" data-label="<?=$this->transEscAttr('history_limits')?>">
           <span class="history_limits_field">


config/vufind/KohaRest.ini
```diff
@@ -63,6 +63,11 @@ clientSecret = ""
 ; end. Leave this commented out to omit OPAC links.
 ;opacURL = "http://koha-server/cgi-bin/koha/opac-detail.pl?biblionumber=%%id%%"
 
+; If we trust our authentication source and know it to be the same as the one used by
+; Koha then we can choose to not validate our patron's passwords (Useful if you are
+; using SAML/Shibboleth for authentication for both VuFind and Koha)
```
demiankatz: It is probably worth adding a note here like:

```
; It is strongly recommended that you set allowUserLogin to false in the [Catalog] section
; of config.ini when setting this value to true.
```

...additionally, I assume the default value here should be false rather than true.

demiankatz (changes requested):
Thanks for this contribution, @Pilvedore! Just a couple of suggestions regarding the .ini file.

Since I am not a Koha user, I'm also going to request a review from @EreMaijala in case he has further feedback.

config/vufind/KohaRest.ini
PerplexedTheta: I've just tested this, and it doesn't seem to have any adverse effects. Will push a commit shortly :-)

PerplexedTheta:
I have now pushed two more commits, which hopefully address comments & resolve build errors.

Thanks! :-)

config/vufind/KohaRest.ini
demiankatz: Thanks! And just for context: if you leave allowUserLogin set to true, and the credentials provided by Shibboleth/SAML are somehow incorrect, there's a possibility that VuFind will prompt the user to enter ILS credentials... and if password validation is disabled, it would be easy to spoof other users through this form. By disabling the setting, you ensure that users can ONLY access the ILS functionality by receiving valid attributes via SSO.

demiankatz (changes requested):
It looks like something here is causing the test suite to fail now (see https://github.com/vufind-org/vufind/actions/runs/7185411362/job/19569532912 for details). Please let me know if you need any help troubleshooting that problem, or if you're not already familiar with how to run tests locally, and I'll be happy to assist as needed!

demiankatz:
Ha, thanks, @Pilvedore, it looks like you were fixing the problem at exactly the same time I was reporting it. :-)

demiankatz:
Now it just looks like it needs a quick run of `vendor/bin/phing phpcbf` to clean up some whitespace issues.

PerplexedTheta:
Hi again @demiankatz - woah, really nice build suite! Great that it catches my IDE's poor configuration. I've just pushed a fix for those stray whitespaces.

Third build is the charm? Fingers crossed!

Thanks :-)

demiankatz (approved):
All checks are passing now, and this looks good to me. I'll hold off on merging until @EreMaijala has a chance to comment, in case he spots any Koha-specific concerns.

Thanks for working through the process, @Pilvedore; I'm glad you found the CI tools helpful. :-)

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
```diff
@@ -106,6 +106,13 @@ class KohaRest extends \VuFind\ILS\Driver\AbstractBase implements
      */
     protected $sessionCache;
 
+    /**
+     * Validate passwords
+     *
+     * @var string
+     */
+    protected $validatePasswords;
```
EreMaijala: This should be set to true by default just in case `init()` doesn't get called properly. And the type should be `@var bool`.

EreMaijala (changes requested):

module/VuFind/src/VuFind/ILS/Driver/KohaRest.php
PerplexedTheta: Yes, I agree, actually. Good spot — I will add this in a new commit

EreMaijala:
@Pilvedore Something odd happened here, and now I can't see any changes in this pull request anymore.

demiankatz:
@Pilvedore, did you create this PR using your dev branch, and did you also set up pull-bot to synchronize your local dev branch with the upstream dev branch? If so, I think that explains what @EreMaijala is seeing -- the pull-bot has overwritten all of your work with the latest upstream changes.

The solution is, before pulling from your repo, you should just switch to a new branch in your local code and push that up to GitHub, then close this PR and open a new one using the new branch. That should restore the lost code and prevent pull-bot from messing things up in future.

demiankatz:
This PR was accidentally closed by an automated process; it has been replaced by #3296.

PR #3261, opened 2023-12-13
Make minor tweaks to styles for improved Bootstrap 5 compatibility.

EreMaijala:
From #3222.

Notes:
- bootprint3 pagination is changed to flex as that's what Bootstrap 5 uses as well, and the table style clashed with that.
- bootprint3 font size base setting was unnecessary, and could cause trouble with relative sizes.
- sandal navbar selector is just a bit more specific to ensure it overrides Bootstrap styles.


demiankatz:
This looks fine to me based on a quick skim, but @crhallberg is better positioned to comment on style-specific concerns.

@sturkel89, do you mind giving this a quick test? It should look exactly the same as the dev branch -- it's just improving forward-compatibility with the newer version of bootstrap. The areas primarily impacted are the drop-down buttons used for theme/language selection in the top bar, and the pagination controls used by search results. If those look the same in all three themes, it's probably fine. (There's also a general font size simplification which shouldn't have any effect, but if you notice weird font sizes anywhere, that would be the other thing to report).

sturkel89 (approved):
I looked at this branch in all three themes and compared it with dev. Everything looks the same in test vs. dev! This PR is ready to merge.

-------------

I have a couple of observations re: the design of elements related to pagination controls and other elements in that section of the results view. I'll leave up to you whether to create another PR for either of these. Just thought I would provide my wish list!

- Bootstrap3 and Bootprint3 have outlining and formatting in the pagination controls; I think that looks better than the un-outlined design of the pagination controls in Sandal:
image
image
VS.
image

-------
I also think we could do something different with alignment of top controls than the current design. As it is, in each theme the alignment gets wonky when the screen is a bit narrowed. It would work better if "Results per page" was not allowed to have a line break. 

(In connection with that, perhaps the "Showing n - n for search ""; query time 1.00s" line could be forced to break after the comma.)
image
image

VS.
image
image

demiankatz (approved):
Thanks, @EreMaijala and @sturkel89! This looks good to me, so I'm going to go ahead and merge it to keep things moving forward.

Regarding @sturkel89's more general feedback, I'm interested in whether @crhallberg or others have thoughts on that! It would definitely be nice to clean up that control wrapping, but I'm not sure how feasible it is within the technical constraints of the framework.

PR #3262, opened 2023-12-13
INN-Reach Account Integration

oharacj:
INN-Reach Checked out and Held items will display in the appropriate sections of the patron account.

Updates to SierraRest ILS Driver and Configuration file. the Bootstrap3 theme was updated as well; a small template update and a new image for checked out items that don't have a record listed.

This was a request from Marmot academics who use INN-Reach extensively to provide resources to patrons. They felt that not displaying title information would be confusing to their patrons. 

As access to Sierra DNA is currently necessary to get title and author information it remains a SierraRest only change.


demiankatz:
Thanks, @oharacj! I'm asking @EreMaijala to take a look at this, since he's the primary developer/user of the SierraRest driver at this point.

demiankatz (changes requested):
Looks like there are some style issues that need to be addressed -- see failing build at https://github.com/vufind-org/vufind/actions/runs/7199562620/job/19620085433 for details.

Many of these can be automatically fixed by running `vendor/bin/phing phpcbf`, but the missing comments and long lines need to be addressed manually. If you need help with shortening long lines, let me know and I'm happy to assist! :-)

oharacj:
Demian,
  I have cleaned up the code and pushed it again. I also found a block of my code that was accomplishing nothing and removed it. Hopefully this makes it through the automated tests this time.

demiankatz:
Thanks, @oharacj -- it looks like you've fixed all the things requiring manual adjustment, but you still need to run `vendor/bin/phing phpcbf` and commit the results to get the automatic fixes applied.

demiankatz:
Okay, phpcbf is happy now... next step: `vendor/bin/phing php-cs-fixer`, and commit those results too. :-)

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -429,6 +436,21 @@ public function setConfig($config)
      */
     public function init()
     {
+        if (empty($this->config)) {
+            throw new ILSException('Configuration needs to be set.');
+        }
+        if ($this->config['InnReach']['enabled']) {
```
demiankatz: I would make this `if ($this->config['InnReach']['enabled'] ?? false) {` to fall back to a default behavior if the configuration is entirely missing.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3415,6 +3467,14 @@ protected function getItemsWithBibsForTransactions(
                 $bibIdsToItems[$this->extractId($bibId)][$itemId] = true;
             }
         }
+        if ($this->config['InnReach']['enabled']) {
```
demiankatz: Probably worth either adding the same `?? false` here as above, or creating an `innReachEnabled()` method to encapsulate the logic if you prefer.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -1179,6 +1215,21 @@ public function getMyHolds($patron)
                 'updateDetails' => $updateDetails,
             ];
         }
+        $n = 0;
+        foreach ($holds as $hold) {
+            if ($this->config['InnReach']['enabled']) {
```
demiankatz: I would move the config check outside the foreach loop so you don't iterate at all when the feature is disabled. Also, same need for fallback to default false.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -837,6 +859,20 @@ public function getMyTransactions($patron, $params = [])
             }
             $transactions[] = $transaction;
         }
+        if ($this->config['InnReach']['enabled']) {
```
demiankatz: Default value needed here too.

demiankatz (changes requested):
Hooray -- the build is passing now!

I took a closer look at the code, made a couple of small stylistic edits to save you time, and left a few comments below (though really it's mostly just one comment repeated in several places :-) ).

oharacj:
running the cleanup locally again, sorry about my memory.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3487,4 +3548,88 @@ protected function checkTitleHoldRules(array $bib, array $patron): bool
...
+                        sierra_view.bib_record_property
+                    WHERE
+                        hold.id = $1
+                    AND hold.is_ir=true
+                    AND hold.record_id = bib_record_item_record_link.item_record_id
+                    AND bib_record_item_record_link.bib_record_id = bib_record_property.bib_record_id';
+            pg_prepare($this->db, 'prep_query', $query);
+            $results = pg_execute($this->db, 'prep_query', [$holdId]);
+            while ($resultArray = pg_fetch_row($results)) {
+                $titleInfo = [
```
EreMaijala: This loops through all results and overrides `$titleInfo` on each iteration. If you only expect to return a single result, I'd get rid of the loop.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3487,4 +3548,88 @@ protected function checkTitleHoldRules(array $bib, array $patron): bool
         }
         return false;
     }
+
+    /**
+     * Gets title information for holds placed in an INN-Reach system
+     *
+     * @param $holdId the id of the hold from Sierra
+     * @param $itemId the id of the item or bib from Sierra
```
EreMaijala: This is always a bib id, right?
```suggestion
     * @param $bibId the id of the bib from Sierra
```

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -1179,6 +1215,22 @@ public function getMyHolds($patron)
...
+
+        if ($this->config['InnReach']['enabled'] ?? false) {
+            $n = 0;
+            foreach ($holds as $hold) {
+                if (!empty($hold['item_id']) && strstr($hold['item_id'], $this->config['InnReach']['identifier'])) {
+                    $id = $hold['id'];
+                    $volume = $hold['volume'];
+
+                    $innReach = $this->getInnReachHoldTitleInfoFromId($hold['reqnum'], $hold['id']);
+                    $holds[$n]['id'] = $innReach['id'];
```
EreMaijala: As far as I can see, getInnReachHoldTitleInfoFromId may return an empty array as well, so the result should be checked before accessing the array elements.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3487,4 +3548,88 @@ protected function checkTitleHoldRules(array $bib, array $patron): bool
...
+  sierra_view.checkout,
+  sierra_view.bib_record_item_record_link,
+  sierra_view.bib_record_property
+WHERE 
+  checkout.id = $1
+  AND checkout.item_record_id = bib_record_item_record_link.item_record_id
+  AND bib_record_item_record_link.bib_record_id = bib_record_property.bib_record_id';
+            pg_prepare($this->db, 'prep_query', $query);
+            $results = pg_execute($this->db, 'prep_query', [$checkOutId]);
+            while ($resultArray = pg_fetch_row($results)) {
```
EreMaijala: I think `if` would be more appropriate when you don't intend to loop the results.

themes/bootstrap3/templates/myresearch/checkedout.phtml
```diff
@@ -111,7 +111,11 @@
           <?php endif; ?>
 
           <?php
-            $coverDetails = $this->record($resource)->getCoverDetails('checkedout', 'small', $this->recordLinker()->getUrl($resource));
+          if (!empty($resource->getExtraDetail('ils_details')['id'])) {
+              $coverDetails = $this->record($resource)->getCoverDetails('checkedout', 'small', $this->recordLinker()->getUrl($resource));
+          } else {
+              $coverDetails['html'] = "<img src='" . $this->imageLink('InnReachCover.png') . "' alt='book cover -" . $resource->getTitle() . "' />";
```
EreMaijala: I don't think this can be part of the generic template here.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -429,6 +436,21 @@ public function setConfig($config)
...
+            throw new ILSException('Configuration needs to be set.');
+        }
+        if ($this->config['InnReach']['enabled'] ?? false) {
+            try {
+                $conn_string = 'host=' . $this->config['InnReach']['dna_url']
+                    . ' port=' . $this->config['InnReach']['dna_port']
+                    . ' dbname=' . $this->config['InnReach']['dna_db']
+                    . ' user=' . $this->config['InnReach']['dna_user']
+                    . ' password=' . $this->config['InnReach']['dna_password'];
+                $this->db = pg_connect($conn_string);
```
EreMaijala: Since the connection is needed in only certain situations, it'd be better to defer opening it before it's actually needed. Add something like `protected method getInnReachDb()` that initialized the member variable as necessary and returns it. I'd also rename `$this->db` to `$this->innReachDb` to make it clear that it's not used for Sierra.

Also the return value should be checked and any errors handled appropriately (at least errors should be logged, see [pg_last_error](https://www.php.net/manual/en/function.pg-last-error.php)). I'm not sure whether that means logging the error and continuing without INN-Reach support, or giving up and throwing an exception.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -837,6 +859,20 @@ public function getMyTransactions($patron, $params = [])
...
             $transactions[] = $transaction;
         }
+        if ($this->config['InnReach']['enabled'] ?? false) {
+            $n = 0;
+            foreach ($transactions as $transaction) {
+                $irIdentifier = $this->config['InnReach']['identifier'];
+                if ($transaction['item_id'] && strstr($transaction['item_id'], $irIdentifier)) {
+                    $irCheckoutId = $transaction['checkout_id'];
+                    $irItemId = $transaction['item_id'];
+                    $innReach = $this->getInnReachCheckoutTitleInfoFromId($irCheckoutId, $irItemId);
```
EreMaijala, line 877: As far as I can see, getInnReachCheckoutTitleInfoFromId return value should be checked before accessing the array elements.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3487,4 +3548,88 @@ protected function checkTitleHoldRules(array $bib, array $patron): bool
...
+        return $titleInfo;
+    }
+
+    /**
+     * Gets title information for checked out items from INN-Reach systems
+     *
+     * @param $checkOutId the id of the checkout from Sierra
+     * @param $itemId     the id of the item or bib from Sierra
+     *
+     * @return array|void
```
EreMaijala: Make it return an empty array if nothing is found to align with getInnReachHoldTitleInfoFromId.

config/vufind/SierraRest.ini
```diff
@@ -268,3 +268,18 @@ title_hold_excluded_item_codes = "e"
 ;display_orders = false
 ; Whether to check holdings records in search results (default is true):
 ;check_holdings_in_results = true
+
+; This section can be configured to allow display of held and checked out items from
+; INN-Reach systems
+[InnReach]
+; In order to get title information from INN-Reach we need to connect to Sierra DNA
+dna_url = [DNA address]
```
EreMaijala: Why not make the configuration a connection string instead of the separate settings that you need to parse together? That would enable any connection string parameters like `sslmode` without additional effort.

config/vufind/SierraRest.ini
```diff
@@ -268,3 +268,18 @@ title_hold_excluded_item_codes = "e"
 ;display_orders = false
 ; Whether to check holdings records in search results (default is true):
 ;check_holdings_in_results = true
+
+; This section can be configured to allow display of held and checked out items from
+; INN-Reach systems
+[InnReach]
+; In order to get title information from INN-Reach we need to connect to Sierra DNA
```
EreMaijala: Maybe add a note that this requires the PostgreSQL extension in PHP?

EreMaijala (changes requested):
I don't know INN-Reach, so I can't really comment on the way things are done, but this is my best effort. What I find a bit confusing is that it looks like the SQL statements are retrieving data that the API calls should have returned but didn't. Perhaps a couple of additional comments for the methods could clarify the workflow.

From what I gathered, INN-Reach has some sort of an integration API (mentioned in [Ex Libris documents](https://knowledge.exlibrisgroup.com/Alma/Implementation_and_Migration/Implementation_Guides/InnReach_API_Implementation_Guide)) too. Would it be possible to use it instead of direct database access?

Regarding the database access: if you don't store the prepared statements, it'd be simpler to use `pg_query_params` instead. But since the statements can be executed several times, it'd be more efficient to store them (results of pg_prepare) in member variables and reuse them.

themes/bootstrap3/templates/myresearch/checkedout.phtml
demiankatz: Agreed. This just appears to be a generic "cover not available" image, so I'm not sure if it's needed at all. You could locally adjust the no-cover image if you want to use this version instead of the default one. Taking it out should simplify things a little bit, and if you really think InnReach needs its own cover image, we could create a more generic mechanism to support that in a separate PR after the basic functionality is merged.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3487,4 +3548,88 @@ protected function checkTitleHoldRules(array $bib, array $patron): bool
...
+     *
+     * @throws ILSException
+     */
+    private function getInnReachCheckoutTitleInfoFromId($checkOutId, $itemId)
```
demiankatz: Please change private methods to protected -- this makes it easier to locally extend/override the code. (We don't have any private methods in the code base for this reason).

demiankatz (changes requested):
Thanks, @oharacj, for working through all the CI stuff. No need to apologize -- this stuff is here so that we don't HAVE to have good memories. :-) Once your first contribution gets merged, then CI will run automatically for subsequent PRs and the whole process will get smoother. It's just a little slow for this first one because another committer needs to manually approve the CI run for each update. (I don't mind doing this, but it does slow down the feedback loop for you).

Thanks, @EreMaijala, for the helpful review. I've added a couple of small comments of my own to expand upon it.

config/vufind/SierraRest.ini
```diff
@@ -268,3 +268,15 @@ title_hold_excluded_item_codes = "e"
...
 ;check_holdings_in_results = true
+
+; This section can be configured to allow display of held and checked out items from
+; INN-Reach systems
+[InnReach]
+; Toggle enabled to true to turn on InnReach Integration
+enabled = false
+; In order to get title information from INN-Reach we need to connect to Sierra DNA
+;This requires the PostgreSQL extension to be enabled in PHP
+sierra_db = postgresql://[DNA URL]:[DNA Port]/[DNA Database]?user=[DNA User]&password=[DNA password]&sslmode=prefer
```
demiankatz: This line seems to be causing a parse error and breaking some tests. You should probably put the value inside double quotes and/or comment it out by default.

config/vufind/SierraRest.ini
```diff
@@ -268,3 +268,15 @@ title_hold_excluded_item_codes = "e"
...
 ; Whether to check holdings records in search results (default is true):
 ;check_holdings_in_results = true
+
+; This section can be configured to allow display of held and checked out items from
+; INN-Reach systems
+[InnReach]
+; Toggle enabled to true to turn on InnReach Integration
+enabled = false
+; In order to get title information from INN-Reach we need to connect to Sierra DNA
+;This requires the PostgreSQL extension to be enabled in PHP
```
demiankatz: Minor style note: please put a space after the `;` for informational comments (our convention is to use single spaces for human-readable comments, and no spaces for commented-out example settings).

themes/bootstrap3/images/InnReachCover.png
demiankatz, line 1: I think this file can be deleted, as it is no longer being used.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -276,7 +283,7 @@ class SierraRest extends AbstractBase implements
      *
      * @var int
      */
-    protected $apiVersion = 5;
+    protected $apiVersion = 5.1;
```
demiankatz: Is there a reason to change the default API version as part of this PR? This may impact backward compatibility for users with older versions of Sierra (not sure how common that is at this point, though).

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -509,6 +520,16 @@ public function init()
         $this->sessionCache = $factory($namespace);
     }
 
+    protected function getInnReachDb()
```
demiankatz, line 528: This function will need a doc comment to avoid style complaints. Maybe something like:

```
/**
 * Establish INN-Reach database connection.
 *
 * @return void
 */
```

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -837,6 +858,23 @@ public function getMyTransactions($patron, $params = [])
             }
             $transactions[] = $transaction;
         }
+        if ($this->config['InnReach']['enabled'] ?? false) {
+            $n = 0;
+            foreach ($transactions as $transaction) {
```
demiankatz: Will this work the same if you do `foreach ($transactions as $n => $transaction) {` instead of separately managing the $n counter? It looks like that might be a worthwhile simplification.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -1179,6 +1217,24 @@ public function getMyHolds($patron)
                 'updateDetails' => $updateDetails,
             ];
         }
+
+        if ($this->config['InnReach']['enabled'] ?? false) {
+            $n = 0;
+            foreach ($holds as $hold) {
```
demiankatz: Same loop simplification is likely possible here.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3405,6 +3461,7 @@ protected function getItemsWithBibsForTransactions(
         }
         // Fetch items and collect bib id mappings if available:
         $itemIds = [];
+        $innReachIndices = [];
```
demiankatz: This array gets populated but never seems to be used for anything. Is something missing, or can something be simplified/removed?

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3487,4 +3552,91 @@ protected function checkTitleHoldRules(array $bib, array $patron): bool
...
+     * @param $bibId  the id of the bib from Sierra
+     *
+     * @return array|void
+     *
+     * @throws ILSException
+     */
+    protected function getInnReachHoldTitleInfoFromId($holdId, $bibId)
+    {
+        if (!isset($this->innReachDb)) {
+            $this->getInnReachDb();
```
demiankatz: Maybe rather than making getInnReachDb a void function, it would be better to have it actually return the database connection... then it could contain logic like:

```
if (!isset($this->innReachDb)) {
    // ... establish connection here ...
}
return $this->innReachDb;
```

And then the database code below can use the returned connection. I think that slightly clarifies the flow of everything.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
demiankatz: I think this feedback still needs to be addressed, unless you intentionally decided not to do things this way.

demiankatz (changes requested):
Thanks, @oharacj, this continues to improve! Please see below for the next round of feedback, questions and suggestions.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
oharacj: I was having problems connecting to our Sierra instance and tried this as a fix. It didn't work and I accidentally forgot to switch it back. Changed back now.

oharacj:
> I think this feedback still needs to be addressed, unless you intentionally decided not to do things this way.

I've made these changes except for the one quoted above. I'm not seeing the same thing you are. Both of the params in the code I'm looking at seem to say bibId and not itemId.

demiankatz:
> > I think this feedback still needs to be addressed, unless you intentionally decided not to do things this way.
> > I've made these changes except for the one quoted above. I'm not seeing the same thing you are. Both of the params in the code I'm looking at seem to say bibId and not itemId.

The feedback I was referring to was about changing the function to consistently return an array, instead of array|void (i.e. return an empty array instead of failing to return any kind of value). It looks like the GitHub code review interface displayed my comment without a lot of context, so sorry for the lack of clarity there! :-)

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -509,6 +520,24 @@ public function init()
         $this->sessionCache = $factory($namespace);
     }
 
+    /**
+     * Establish INN-Reach database connection
+     *
+     * @return PgSql\Connection|void
```
demiankatz: Should this be `@return resource` for consistency with the property definition above?

Also, I don't think disabling the InnReach configuration when an exception is thrown is necessarily going to help, because there's code that's assuming `$this->innReachDb` is valid after calling this method without checking anything else. I think it might be better for the methods that use this to actually capture and check the return value, so they can act appropriately if a connection is not available for whatever reason.

demiankatz (changes requested):
Thanks for the quick turnaround on all this, @oharacj -- looking good. Just a couple remaining questions about handling of the database connection.

demiankatz:
Also note that I pushed up a small fix (commit [3a6c950](https://github.com/vufind-org/vufind/pull/3262/commits/3a6c950973aee5937bbd3eac699e32e937911d54)) to eliminate an unnecessary change, since it seemed easier to do it than to waste your time with a review comment. ;-)

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
oharacj: I wondered about that, too. I'll get on it.

oharacj:
@EreMaijala, I forgot your INN-Reach questions. I can add comments but I'll answer them the best I can here and you can tell me if the comments are needed.

Holds and Checked out items do not exist in the catalog and will not be returned using the Sierra API. I met with someone who runs our INN-Reach collection (The Colorado Alliance - Prospector), and was informed that we do not have access to the API. I'm not sure why. She told me that'd we'd have to use the temporary bibs created in the Sierra database in order to get title or author for the items.

I have no answer for the way that the calls are happening other than lack of knowledge on my end. I can definitely look into storing them or using variables for them.

Thanks.

demiankatz:
Thanks once again, @oharacj! I'll wait and see what @EreMaijala has to say before commenting further, but hopefully we're getting close to wrapping this up.

EreMaijala:
@oharacj Thanks for the background information! This makes sense to me.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -77,6 +77,13 @@ class SierraRest extends AbstractBase implements
      */
     public const HOLDINGS_LOCATION_FIELD = '40';
 
+    /**
+     * Database connection
+     *
+     * @var resource
+     */
+    protected $innReachDb;
```
EreMaijala: I'd initialize this to null here. And make type `@var ?resource`.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -509,6 +520,27 @@ public function init()
...
 
+    /**
+     * Establish INN-Reach database connection
+     *
+     * @return resource|void
+     */
+    protected function getInnReachDb()
+    {
+        try {
+            if (!isset($this->innReachDb)) {
```
EreMaijala: When initialized to null above you can just compare with null here: `if (null === $this->innReachDb) {`. 
I'd also check this outside of the try-catch block, since exception handling always adds a bit of overhead.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3487,4 +3557,103 @@ protected function checkTitleHoldRules(array $bib, array $patron): bool
...
+     * @param $bibId  the id of the bib from Sierra
+     *
+     * @return array
+     *
+     * @throws ILSException
+     */
+    protected function getInnReachHoldTitleInfoFromId($holdId, $bibId): array
+    {
+
+        $this->getInnReachDb();
```
EreMaijala: Assign the return value to a variable and use it here. Like:

```
$db = $this->getInnReachDb();
...
if ($db) ...

```

...and remove the extra blank line above. :)

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -3487,4 +3557,103 @@ protected function checkTitleHoldRules(array $bib, array $patron): bool
...
+     * @param $bibId      the id of the bib from Sierra
+     *
+     * @return array
+     *
+     * @throws ILSException
+     */
+    protected function getInnReachCheckoutTitleInfoFromId($checkOutId, $bibId): array
+    {
+
+        $this->getInnReachDb();
```
EreMaijala: Same as in `getInnReachHoldTitleInfoFromId` here.

EreMaijala (changes requested):
Looks pretty good to me! I had just a few mostly nitpicky notes.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -509,6 +520,27 @@ public function init()
...
+        if (!isset($this->innReachDb)) {
+            try {
+                $conn_string = $this->config['InnReach']['sierra_db'];
+                $connection = pg_connect($conn_string);
+                $this->innReachDb = $connection;
+            } catch (\Exception $e) {
+                $this->logWarning("INN-Reach: Could not connect to the Sierra database: {$e}");
+                $this->innReachDb = null;
+            } finally {
+                return $this->innReachDb;
```
demiankatz: If you call this function more than once, it won't return anything after the first call, because it currently only returns innReachDb when it was not previously set. I think the return statement should be moved to the very bottom of the function, outside of the if statement, to guarantee that it is returned whether or not it had to be initialized on the current call.

module/VuFind/src/VuFind/ILS/Driver/SierraRest.php
```diff
@@ -509,6 +520,27 @@ public function init()
         $this->sessionCache = $factory($namespace);
     }
 
+    /**
+     * Establish INN-Reach database connection
+     *
+     * @return resource|void
```
demiankatz: I think it would be best to use `?resource` here instead of `resource|void` for consistency with the property definition above (and brevity).

demiankatz (changes requested):
Thanks once again, @oharacj. Looks like you still have a few more of @EreMaijala's comments to work through before this gets merged, but while I was reviewing the work in progress, I noticed a couple more small things related to the database initialization. I think as soon as we get all this wiring connected up optimally, we'll be ready to merge. :-)

demiankatz:
Thanks, @oharacj -- I think all that's left is making use of the return value of $this->getInnReachDb(). No rush if you simply haven't gotten to it yet, but if our comments have been too vague, let me know and I'd be happy to push up a change to clarify things.

oharacj:
@demiankatz, Apologies, but yeah, I'm not sure what you mean by "making use of the return value of $this->getInnReachDb()" a further explaination would be very helpful

demiankatz:
> @demiankatz, Apologies, but yeah, I'm not sure what you mean by "making use of the return value of $this->getInnReachDb()" a further explaination would be very helpful

Ha, I'm the one who should be apologizing, because you've already done what we suggested, and I somehow missed when it happened. Everything looks good to me now. :-)

demiankatz (approved):
I believe this is ready to merge, but I'll wait for @EreMaijala's final approval as well.

EreMaijala (approved):

PR #3263, opened 2023-12-14
Enclose search history label in span to avoid collapsing of white space.

EreMaijala:
Otherwise the history labels and search terms would lose the white space between them on narrow screens.


demiankatz (approved):
Thanks, @EreMaijala!

PR #3255, opened 2023-12-11
Add support for Primo REST API.

EreMaijala:
Deprecates the legacy brief search API support and adds support for the current REST API for Primo search. Also includes a fix for exclude filters clashing with normal filters for the same field. Processing of the format field is moved to record driver to avoid duplication (and doing it in the right place).

The legacy brief search API has issues with multi-term searches providing illogical results, and has been suffering from some other issues as well lately. To make the transition easier, support for the old API is maintained so that switching to the new REST API can be done seamlessly when the required preparation is complete.

The new RestConnector is modeled after the old one, though it has went through a bit of refactoring.

PR #3265, opened 2023-12-15
Improve Primo advanced search form layout.

EreMaijala:
From #3255. Note that this still keeps the right side empty for possible help columns like in Solr advanced search.


demiankatz (approved):
Thanks, @EreMaijala! Since these changes are minor and Primo-focused, they pose very little risk of side effects, so I'm going to go ahead and merge this right away so it can get merged back into #3255, and further refinement/testing/conversation can occur over there.

PR #3266, opened 2023-12-15
Initialize cluster properly in GetSideFacets.

EreMaijala:
At least hierarchical facets expect `$this->cluster` to contain the `list` element.


demiankatz (approved):
Thanks for the fix!

PR #3267, opened 2023-12-15
Trigger errors from AJAX methods in non-production mode.

EreMaijala:
Errors in AJAX request could easily go undetected (e.g. #3266) because they were only in the `php_errors` element in AJAX responses. This change makes it so that in non-production mode any errors go through the normal triggers so that they can cause the request to fail and the error to be triggered in debugger.


EreMaijala:
Note that I intentionally chose to compare with 'production' instead of 'development' so that you can run Mink tests with e.g. 'preproduction' to enable error triggers but keep Whoops disabled.

module/VuFind/src/VuFind/Controller/AjaxResponseTrait.php
```diff
@@ -183,12 +183,12 @@ protected function callAjaxMethod($method, $type = 'application/json')
      * @param string $errfile File where error occurred
      * @param string $errline Line number of error
      *
-     * @return bool           Always true to cancel default error handling
+     * @return bool           True in production to cancel default error handling, false otherwise
      */
     public static function storeError($errno, $errstr, $errfile, $errline)
     {
         self::$php_errors[] = "ERROR [$errno] - " . $errstr . "<br>\n"
```
demiankatz: While we're changing this, I wonder if we should also change the content of the error message to be more generic in production mode, to prevent leaking of any sensitive information -- i.e. for production just use the 'An error has occurred.' string. Since we do that for regular exceptions, I see no obvious reason to behave differently here.

Of course, that then raises the question of whether we want to include errors in the AJAX response without causing a fatal result in some circumstances but not others. I'm not really sure what's best in that regard.

Also, if you think having a "preproduction" environment would be valuable for certain situations, that might be something worth beginning to document somewhere... and/or maybe we should add a new Phing parameter to set the level of the started environment. (Obviously, these details probably fall outside the scope of this PR).

demiankatz (changes requested):
A few thoughts...

module/VuFind/src/VuFind/Controller/AjaxResponseTrait.php
EreMaijala: I'm not sure either what would be the best course of action, but I think I'd prefer failure instead of just reporting the errors (silently). We'd just need to revamp our error handling so that an AJAX request wouldn't receive the full-blown error page as result. I'll think about this a bit and will change this accordingly later.

If we decide to fail requests with errors, we won't need the "preproduction" environment, so I'll leave that for later.

demiankatz: Thanks! I think the current design intends three levels: development, testing, production. One approach could be to switch to a generic message in production, use a full message in testing, and cause a full-blown failure in development. (That seems like the best approach if we want to hold on to the existing logic in some form... but I'm not opposed to totally rethinking this if you have a better overall idea).

EreMaijala: Well, what do I know... I just noticed that `formatContent()` only outputs the errors in development mode, so that part is already taken care of regardless of what we do in `storeError()`.

The problem with just outputting a message in testing is that Mink tests won't catch it. Here's the latest iteration that enables the `storeError()` handler only in production mode. I think this is the most straight-forward way, though with the default setup the stored errors won't be used in any way. But they could be used if a specific handler wants to. What do you think?

demiankatz: If we're only storing the messages in production mode, and we're only outputting stored messages in development mode, doesn't that just mean that we'll never do anything? If that's the case, it's probably better to strip the code out entirely rather than leaving potentially confusing vestiges.

Might a better approach be to write a different storeError handler that writes to the logger instead of messing around with AJAX response content? That would probably improve the visibility of errors in a production environment (for the right people).

EreMaijala:
@demiankatz How about this? Real errors would prevent a response from being generated. Warnings and notices would be logged normally while prevented from being displayed. This would ensure that PHP settings are otherwise in effect just like in all other controllers.

demiankatz (approved):
Thanks, @EreMaijala, this seems like a reasonable solution to me! I'll make a note about it in the changelog.

PR #3269, opened 2023-12-15
New items improvements

demiankatz:
This PR makes three distinct improvements to New Items functionality:

1.) It fixes a bug that was causing missing whitespace in the search results display (this bug was introduced post-9.1 and does not impact the release-9.1 branch, so it does not need to be backported).

2.) It adds the ability to specify a separate default sort for new items.

3.) It adds the ability to configure facets for pre-filtering of new item results, similar to the controls found on the advanced search screen.

There's a bit of minor refactoring to support this functionality.


module/VuFind/src/VuFind/Controller/SearchController.php
```diff
@@ -235,12 +235,20 @@ public function newitemAction()
...
+                'defaultSort' => $this->newItems()->getDefaultSort(),
                 'fundList' => $this->newItems()->getFundList(),
                 'ranges' => $this->newItems()->getRanges(),
             ]
         );
+        if ($this->newItems()->includeFacets()) {
+            $view->options = $this->serviceLocator
+                ->get(\VuFind\Search\Options\PluginManager::class)
+                ->get($this->searchClassId);
+            $this->addFacetDetailsToView($view, 'NewItems');
```
maccabeelevine, line 249: I'm not clear what code pathway leads to this line being executed, since the only way I know to view the results page is by selecting a range at the New Items start page, and then this function short-circuits at line 235.

maccabeelevine (commented):
I took a look at this functionality; keeping in mind that I haven't used New Items before.  The first two parts worked fine.  I can't get the third-part working, see code comment.

module/VuFind/src/VuFind/Controller/SearchController.php
demiankatz: The idea here is that if you go to NewItems page without any parameters, you get displayed the form where you can select range, filters, etc. That page then submits back to itself with parameters set, and it is at that point that the short-circuit redirect to the results page occurs. If you go to the `/vufind/Search/NewItem` URL without any parameters set, this form should render. Are you seeing different results?

maccabeelevine: Thanks for explaining (and sorry for the delay).  Yup this works fine.

demiankatz: Great! Fine enough that you're willing to approve the PR, or do you want me to seek more input from others?

maccabeelevine (approved):
Looks solid

